PR #62, opened 2012-03-30
Added Disk I/O stats from /proc/bc/<CT ID>/ioacct files

aiurlano:
Hello.

This patch adds Disk I/O stats monitoring for OpenVZ containers.

The meaning of the ioacct files fields is documented here: http://wiki.openvz.org/IO_accounting


mrunge:
Closing this, as there has not been any action for the last 3 years.

PR #277, opened 2013-03-12
Define exactly what RRA you need

ymettier:
Hello,

This patch lets you choose what RRA to select.
Default is AVERAGE MIN and MAX.
With this patch, you can select the ones you want, for example AVERAGE, or AVERAGE and MIN...

This saves your disk space if you do not need all the RRA.

Regards,
Yves

Edit : old title was "Added RRA option to specify what RRA to use".


ymettier:

PR #278, opened 2013-03-12
Rename Solaris disks as cXtXdX instead of sdX or ssdX

ymettier:
Hello,

This patch defines a new config key "UseSolariscXtXdXName" to use the Solaris notation of disks cXtXdX instead of sdX or ssdX.

It uses /etc/path_to_inst and the links in /dev/dsk to find the cXtXdX name from the kstat name.

Regards,
Yves


dago:
Having the ctds-notation on Solaris would definitely be useful. However, the patch is way too complicated and not always correct.

Here is how this really works: Disks are managed by the `sd` and `ssd` (for Fibrechannel) drivers, but they stem from similar code and behave similarly. Drivers in Solaris are created as _instances_ from the kernel, what that means is driver-specific. For the `sd`-driver an instance manages a disk or LUN. These driver instances can be accessed by device files, which mainly consist of a major and minor number. The major number is unique for the driver and assigned during system boot by looking in `/etc/name_to_major`. The minor number specified which instance to access and which slice (this is what partitions are called) on the disk. The magic formula which is hardcoded in the driver is `<minor> = 8 * <instance> + <slice>`. The referenced `/etc/path_to_inst` maps hardware paths to driver instances and is also only read during boot, a hardware path directly maps to a slot, physical socket etc.

The ctds-numbering is solely kept under `/dev/dsk/` and `/dev/rdsk/` which contains symlinks from the ctds-name to the devicefile below `/devices`. To get the `sd`-number for a ctds-device just get the minor-number from following the symlink and divide by 8. In the opposite direction all device files below `/dev/dsk` should be read with their minor nodes and the instance can then easily be looked up in a reverse lookup table.

dago:
Traversing the device tree with `libdevinfo(3lib)` to find the `sd`-instance for a major/minor pair would be even better.

PR #338, opened 2013-05-31
added possibility to choose between holding and input registers

jan-vitek:
I added 'Function' parameter to modbus plugin config. It's value can be 'holding' or 'input'. Plugin then uses modbus_read_registers or modbus_read_input_registers function from libmodbus. Default value is holding. This does not work with legacy libmodbus.

Something like this should be added to modbus section of collectd.conf documentation:
Function holding|input
Specifies register type to be collected from device. Works only with libmodbus 2.9.2 or higher. Defaults to holding.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -2267,6 +2267,7 @@ Synopsis:
  <Data "voltage-input-1">
    RegisterBase 0
    RegisterType float
+   ModbusRegisterType holding
    Type voltage
    Instance "input-1"
  </Data>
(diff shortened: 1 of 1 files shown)


sandeen:
So, I know I'm late to the party, but as I was looking over my config files, and "ModbusRegisterType" sticks out like a sore thumb.

RegisterCmd perhaps?  What I had in my patch was:

    RegisterCmd ReadHolding
or
    RegisterCmd ReadInput
which fits better with RegisterBase and RegisterType, imho.

PR #365, opened 2013-07-02
Adding support for IPMI sensor aliases and IgnoreSensorFail option

marinazapater:
Sensor Aliases
----------------------
The IPMI plugin sets a default naming policy for sensors. This feature allows the user to override the default naming on a per-sensor basis.

To override the default sensor name simply add the custom name after the default name in the config, as follows:

    Sensor "default_sensor_name;custom_sensor_name"

Note that the `default_sensor_name` is the default name given by the IPMI plugin. At the moment, there is no easy way to know it beforehand, other than by running the IPMI plugin and checking the names it assigns to the sensors.

In the future, a helper tool should exist to list all of the available sensors with their default IPMI `default_sensor_name`.

IgnoreSensorFail option
----------------------------------
Before, the IPMI plugin would remove a sensor when a single read failure occurred. With this option set to `true`, the sensor won't be removed after failing (i.e., the IPMI plugin will retry reading from the sensor the next time it gathers data).

By default, it's set to `false`, leaving the default behavior as it was before the change (i.e. removing sensors even if they fail one reading).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -130,6 +130,10 @@ Marco Chiappero <marco at absence.it>
  - ip6tables support in the iptables plugin.
  - openvpn plugin (support for more status file formats)
 
+Marina Zapater <marina at die.upm.es>
+ - Custom sensor naming
+ - Avoid removing sensors on failure
+
 Michael Hanselmann <public at hansmi.ch>
  - md plugin.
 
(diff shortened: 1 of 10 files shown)


octo:
Closing PR: stale for over one month.

PR #390, opened 2013-08-07
libperl's ldopts include libaries and therefore should go in LIBS not LDFLAGS

radford:
Without this configure fails compile its libperl test program because -lperl comes before the test program on the compiler command line.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3219,12 +3219,12 @@ if test "x$with_libperl" = "xyes" \
 	&& test -n "$perl_interpreter"
 then
   SAVE_CFLAGS="$CFLAGS"
-  SAVE_LDFLAGS="$LDFLAGS"
+  SAVE_LIBS="$LIBS"
 dnl ARCHFLAGS="" -> disable multi -arch on OSX (see Config_heavy.pl:fetch_string)
   PERL_CFLAGS=`ARCHFLAGS="" $perl_interpreter -MExtUtils::Embed -e ccopts`
-  PERL_LDFLAGS=`ARCHFLAGS="" $perl_interpreter -MExtUtils::Embed -e ldopts`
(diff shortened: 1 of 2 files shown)


mahmoudimus:
@radford can you rebase this again and remove the typo changes to reduce the changes? I think all you really modified was `configure.am` and `Makefile.am`

radford:
@mahmoudimus, the typo changes where not mine.  There should only be my two commits now that I have rebased against master again.

mahmoudimus:
@radford sweet thank you! /cc @pyr and @mfournier - anything I can do to help expedite this patch?

radford:
> the ubuntu package maintainers do some magic to get it to build, which leads to a massive dpatch (I'd love to know what is needed to generate this patch, btw)

@mfournier, I just noticed this part of your comment.  Sorry for the delay.   The massive patch is mostly noise due to the differences of ```auto*``` versions.  My patch was originally based off the fundamental technique used in the debian / ubuntu patch and should replace it.  Given that, it has effectively been in production for years.

mfournier:
@radford I've successfully built the plugin with your patch on centos 5/6/7, ubuntu 12.04/14.04, and debian testing/stable/oldstable.

For the record a previous fix for this issue broke the build for debian. I can testify this is not the case here.

I really want to see this fixed, because it annoys a lot of people. I just want to be extra careful, because @tokkee seemed reluctant (but this was with a previous version of the patch) and he's already lost a fair amount of hair on this plugin. I'll try to get hold of him on IRC if he doesn't answer to the mention above.

mfournier:
@tokkee ACKed this patch on IRC a couple of minutes ago. He was reluctant because he would have preferred it to be fixed upstream in libperl.

So @radford, you're going to make ubuntu users happy, many many thanks !!

mfournier:
I also cherry-picked these patches to the 5.x release branches. So the fix will be available in the next bugfix releases.

PR #431, opened 2013-09-19
support for user defined queries to the mysql plugin

marekbecka:
This patch adds support for query statements to the mysql plugin. It's inspired by dbi plugin and uses auxiliary functions from utils_db_query. Any comments and suggestions are welcomed.


marekbecka:
I'm closing request in favor of a new version https://github.com/collectd/collectd/pull/3964

PR #481, opened 2013-11-21
Add support for types with multiple ds in the aggregation plugin 

marekbecka:
This patch adds support for types with multiple data sources. It enables to compute aggregations on disk io, network traffic, etc.

### DIFF
--- a/src/aggregation.c
+++ b/src/aggregation.c
@@ -72,7 +72,6 @@ struct agg_ds_s /* {{{ */
   gauge_t squares_sum;
   gauge_t min;
   gauge_t max;
-
 }; /* }}} */
 
 struct agg_instance_s;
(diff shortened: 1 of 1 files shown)


LuizOz:
This patch seems to be working fine.

Although I had to change one line of code 331:
-        "data sources. Expecting %d found %d.", ds->type, inst->ds_num, ds->ds_num);
+        "data sources. Expecting %zu found %d.", ds->type, inst->ds_num, ds->ds_num);

inst->ds_num is a size_t not a int

I'm going to put this in a few servers and see how it goes.

rpv-tomsk:
> I'm going to put this in a few servers and see how it goes.

Hi @LuizOz 

Please give us feedback about this PR.

PR #488, opened 2013-11-22
Tcpconns add all ports summary

marekbecka:
This patch adds global ports statistic to the tcpconns plugin.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1034,6 +1034,7 @@
 
 #<Plugin tcpconns>
 #	ListeningPorts false
+#	AllPortsSummary false
 #	LocalPort "25"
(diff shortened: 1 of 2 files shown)


dllhlx:
if the tcp connections nums bigger than 30000, can this patch lead to perf problems ?

manuelluis:
I think that the patch doesn't add new performance problems to the tcpconns plugin.
To do the accounting it's only do a  ``count_total[state]++``.

mfournier:
@dllhlx I've been using this feature for a few months now. I have machines peaking around 10k simultaneous tcp connections and I didn't notice any correlation with collectd's resource consumption (as reported by pointing the processes plugin to the collectd daemon itself).

PR #520, opened 2014-01-21
src/network.c: Allow automatic flushing of "old" data.

daniel-thompson:
Introduces a MaxAge configuration option to the network plugin. This
results in the accumulated data being evicted if it is older than
the maximum age.

This is useful for "slow" data sources.

Change-Id: I0af319d7d9e7350bed3e218c7e0a69bd099815c0
Signed-off-by: Daniel THOMPSON daniel.thompson@st.com

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -660,6 +660,7 @@
 #		AuthFile "/etc/collectd/passwd"
 #		Interface "eth0"
 #	</Listen>
+#	MaxAge 0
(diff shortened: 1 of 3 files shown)


rpv-tomsk:
This feature already implemented as a per-plugin flush interval.
Now anybody can configure Collectd like following:
```
<LoadPlugin network>
  FlushInterval 10
</LoadPlugin>
```

This issue should be closed as resolved.

PR #538, opened 2014-02-08
Support specifying file owner/group when creating RRD files with rrdcached

jktjkt:
This is needed to support the use case where rrdcached runs as a dedicated user while collectd runs as root.

Nope, this doesn't handle the created directories. There's no sane way to do this without modifying rrd, AFAIK.


eero-t:
@jktjkt Is this still relevant?   If yes, please please handle the CI issues (merge conflict, missing ChangeLog entry).

PR #539, opened 2014-02-08
configfile: Correctly set up the plugin context when using AutoLoadPlugin.

tokkee:
collectd does a strict checking if the plugin context has been initialized
before accessing it. In case of auto-loaded plugins, this was missing and,
thus, collectd complained about being "unable to determine interval from
context".

Thanks to @faxm0dem for reporting this in GH#535

### DIFF
--- a/src/configfile.c
+++ b/src/configfile.c
@@ -196,6 +196,26 @@ static int cf_dispatch (const char *type, const char *orig_key,
 	return (ret);
 } /* int cf_dispatch */
 
+static int cf_load_plugin (const char *name, cdtime_t interval, uint32_t flags)
+{
+	plugin_ctx_t old_ctx;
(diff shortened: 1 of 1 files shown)


tokkee:
I don't think those two issues are related. I could reproduce the problem even when using 'LoadPlugin aggregation' but in neither case was that behavior consistent. Sometimes it works and sometimes it doesn't but I haven't found a pattern yet.

That said, my patch should be applied anyway as it fixes a real issue.

tokkee:
Oh, right … the "value too old" issue should be fixed in 9f6f901889d4c9f594b5ae1fd0f449fcdd2d8fe3

mfournier:
Closing as this has been supersed (and fixed) by f596915.

PR #547, opened 2014-02-08
Discuss the possibility of naming threads.

manuelluis:
This pull request is to debate the possibility of naming threads using  `pthread_setname_np` or   `pthread_set_name_np` in BSDs.

In Linux we need to use  `#define _GNU_SOURCE` to use  `pthread_setname_np`. I don't know If the best approach is to do in configure.ac for all the files or put in plugin.c.

The advantage of naming threads is that in a top you can see by the name of the thread if some plugin that create threads is consuming to much cpu, or see the behaviour of readers and writers.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1349,6 +1349,48 @@ AC_CHECK_MEMBERS([kstat_io_t.nwritten, kstat_io_t.writes, kstat_io_t.nwrites, ks
 #endif
 	])
 
+# check for pthread_setname_np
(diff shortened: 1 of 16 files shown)


mfournier:
Just an update to my report about frequent crashes above. This patch definetely wasn't the culprit. I've been using it since then against various snapshots of the collectd source tree and didn't experience any other problem.

octo:
Superseded by  #2040.

PR #576, opened 2014-03-10
added new apachelog plugin

toni-moreno:
# collectd-apachelog-plugin
## Description

A plugin for collectd which efficiently parses apache log files to get global and extended ( per HTTP CODE)  metrics. It defines a http_perf ( performance ) new type, which defines the new metrics.

**hit_rate** ( hits /second)
**hit_x_interval** ( #hits in the collectd interval time=> will be used to count hits across time )
**rt_avg**  ( average response time in the elapsed interval time in "ms" )
**rt_max** ( max response time in the elapsed interval time in "ms")
**rt_min** ( min response time in the elapsed interval time in "ms")

In extended mode gather performance statistics  for :
- **global** : all requests
- **1XX** : all HTTP 1XX code request
- **2XX** : all HTTP 2XX code request
- **3XX** : all HTTP 3XX code request
- **4XX**: all HTTP 4XX code request
- **5XX**: all HTTP 5XX code request
## Limits

MAX FIELDS per line =100 
if one apache log line has a greater number of  100 FIELDS ( space as FIELD separator) this line will be discarded.

MAX LINE SIZE =  16384 bytes.
if one apache log line is greater than 16384 bytes this line will be discarded
## Customized configuration for Graphite/Collectd users 

Add these storage-aggregation rules the (/opt/graphite/conf/storage-aggregation.conf) file.

```
[http_perf_hits]
pattern = \.hit_x_interval$
xFilesFactor = 0.5
aggregationMethod = sum

[http_perf_rt_avg]
pattern = \.rt_avg$
xFilesFactor = 0
aggregationMethod = avg

[http_perf_rt_min]
pattern = \.rt_min$
xFilesFactor = 0
aggregationMethod = min

[http_perf_rt_max]
pattern = \.rt_max$
xFilesFactor = 0
aggregationMethod = max
```

It can parse apache logs and supports for rotatelogs tool ( by checking the last created file with a pattern)
# Build 

You can now rebuild collecd project.

```
# ./build.sh
# ./configure  --enable-apachelog [other_configure_options]
# ./make
# ./make install
```
# Configure The Plugin
- Instance: Apache instance name (mandatory)
- RenamePluginAs: (default:none)Used to put apachelog metrics beside de apache metrics in case of organize metrics by product better than by plugin source.
- UseApacheRotatedLogs: (default:false) when true , apachelog plugin tracks always the last modified file with pattern in <File ""> section.  
  See: http://httpd.apache.org/docs/2.2/programs/rotatelogs.html
[...]

PR #577, opened 2014-03-10
Added SeparateLevelCharacter/SeparatePluginInstances/ SeparateTypeInstances parameters to graphite plugin to do more flexible naming scheme

toni-moreno:
All these changes are backwards compatible !!

**SeparatePluginInstances and SeparateTypeInstances**

With  SeparateInstances parameter set to true

The collectd-naming schema:

```
host "/" plugin ["-" plugin instance] "/" type ["-" type instance]
```

becomes ( once sended to graphite)

```
host / plugin / plugin instance / type / type instance
```

now we can choose separate plugin and type instance separately ( or together with old parameter)

if SeparatePluginInstances=true

```
host / plugin / plugin instance / type - type instance
```

will be send to graphite 

if SeparateTypeInstances= true

```
host / plugin - plugin instance / type / type instance
```

will be send.

**SeparateLevelCharacter**

Added also SeparateLevelCharacter, that combined with  Prefix/Suffix config parameters  on some plugins ( like GenericJMX) and combined also with  replace rules ( as in the example)  , helps to re-organize metrics in a more flexible way ( by product  by example system/apache/jvm )   ( see picture below) 

```
  <Rule "rename_memory">
     <Target "replace">
        Plugin  "\\<memory\\>" "system@memory"
     </Target>
   </Rule>
```

with this graphite config file...

```
LoadPlugin write_graphite
<Plugin write_graphite>
  <Node "example">
    Host "localhost"
    Port "2003"
    Protocol "tcp"
    LogSendErrors true
    Prefix "XXX.c1."
    Postfix ""
    StoreRates true
    AlwaysAppendDS false
    SeparatePluginInstances true
    SeparateTypeInstances false
    SeparateLevelCharacter "@"
  </Node>
</Plugin>
```

You will 

image

image

**EXAMPLES**

1) Config oracle without SeparateLevelCharacter

```
<Query "q_db_wait">
  Statement "XXXXXX"
 <Result>
    Type "db_wait"
    InstancesFrom "event"
    ValuesFrom "TIME_WAITED"
  </Result>
</Query>

<Query "q_db_io">
  Statement "XXXXX"
  <Result>
    Type "db_io"
    InstancesFrom "i_prefix" "TABLESPACE_NAME"
    ValuesFrom "PHY_BLK_R" "PHY_BLK_W"
  </Result>
</Query>

```

 **result**

image

2) Config with SeparateLevelCharacter

[...]

PR #585, opened 2014-03-17
Added RenamePluginAs in general Purpose Plugin collectors and fix tail csv bug

toni-moreno:
**IMPORTANT** this change es completely Backward compatible.

To make more flexible metric "naming" scheme, I've added a "RenamePluginAs" parameter on each general purpose  plugin  collectors (ie. not closed to a only technology/product).
- tail
- tail_csv
- curl 
- curl_json
- curl_xml
- GenerixJMX.

so you can select "product/plugin" name for each "(File/Instance)" In this way you can  get metrics with tail from system ( and name it as <host>-<systemlog> , and also get metrics from postfix and name it as <host>-<mail>).

NOTE1: we are not able to get this behaviour  with any "replace" Rule , because of a Replace Rule applies to **ALL** instances / files and you can not organize metrics we have gotten from differents sources  with the same  plugin name.

NOTE2: I've also fixed a bug on the tail_csv plugin when more than one metric defined by File.

NOTE3: I've also added this same parameter in the new "apachelog" plugin   released on ( https://github.com/collectd/collectd/pull/576 ) as apachelog format are used by other server products ( WebLogic/HTTP IbmServer/ and other like apache web servers).

### DIFF
--- a/src/collectd-java.pod
+++ b/src/collectd-java.pod
@@ -682,6 +682,13 @@ to it.
 Configures which of the I<MBean> blocks to use with this connection. May be
 repeated to collect multiple I<MBeans> from this server. 
 
+=item B<RenamePluginAs>
+
+Changes the default plugin name  "GenericJVM" by any other string. It lets us to reorganize
+metric tree by an arbitrary category as you need (by example product or technology)  better than by the origin JMX of the data.
+You can change it on each Connection block, so you can use this parameter to put "GenericJMX"  metrics under different products.
+
+
 =back
 
 =head1 SEE ALSO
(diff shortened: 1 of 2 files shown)


toni-moreno:
Hi @pyr @octo. almost 1 year after this  PR was closed  , I'm still working on production with this patch.

 I  have been following daily  the collectd evolution , and I haven't saw any new PR that let me configure collectd metrics by "product" rather than by "origin". (that's was the  main goal of  this PR). 

I would like to know  if you have plans to add this needed (https://github.com/collectd/collectd/pull/577#issuecomment-141605250) functionality in some other way  , if not I would like to create a new PR in current master to add this functionality  again.

Remember that we are not able to get this behavior with any "replace" Rule , because of a Replace Rule applies to ALL instances / files and I don't want to organize metrics we have gotten from different products with the same plugin name.

Thanks a lot for your patience.

PR #598, opened 2014-04-04
Add ceph plugin

daryder:
The ceph plugin uses libjson-c (https://github.com/json-c/json-c) for parsing JSON from ceph daemon admin sockets. This plugin collects all output from a ceph daemon admin socket "perf dump".

### DIFF
--- a/src/ceph.c
+++ b/src/ceph.c
@@ -114,6 +114,14 @@ struct ceph_daemon
     uint32_t *ds_types;
     /** Track ds names to match with types */
     char **ds_names;
+    
+    /**
+     * Keep track of last data for latency values so we can calculate rate
+     * since last poll.
+     */
+    struct last_data **last_poll_data;
+    /** index of last poll data */
+    int last_idx;
 };
 
 /******* JSON parsing *******/
@@ -131,14 +139,6 @@ struct yajl_struct
     int depth;
 };
 typedef struct yajl_struct yajl_struct;
-
-/**
- * Keep track of last data for latency values so we can calculate rate
- * since last poll.
- */
-struct last_data **last_poll_data = NULL;
-/** index of last poll data */
-int last_idx = 0;
 
 enum perfcounter_type_d
 {
@@ -199,7 +199,6 @@ struct last_data
     double last_sum;
     uint64_t last_count;
 };
-
 
 /******* network I/O *******/
 enum cstate_t
@@ -404,7 +403,14 @@ static void ceph_daemons_print(void)
 static void ceph_daemon_free(struct ceph_daemon *d)
 {
     int i = 0;
-    for(; i < d->ds_num; i++)
+    for(; i < d->last_idx; i++)
+    {
+        sfree(d->last_poll_data[i]);
+    }
+    sfree(d->last_poll_data);
+    d->last_poll_data = NULL;
+    d->last_idx = 0;
+    for(i = 0; i < d->ds_num; i++)
     {
         sfree(d->ds_names[i]);
     }
@@ -533,6 +539,7 @@ static int parse_keys(const char *key_str, char *ds_name)
         memcpy(tmp_ds_name, key_str, max_str_len - 1);
         goto compact;
     }
+    
     ds_name_len = (rptr - ptr) > max_str_len ? max_str_len : (rptr - ptr);
     if((ds_name_len == 0) || strncmp(rptr + 1, "type", 4))
     { /** copy whole key **/
@@ -700,6 +707,7 @@ static int cc_add_daemon_config(oconfig_item_t *ci)
                 "with '/' or './' Can't parse: '%s'\n", cd.name, cd.asok_path);
         return -EINVAL;
     }
+    
     array = realloc(g_daemons,
                     sizeof(struct ceph_daemon *) * (g_num_daemons + 1));
     if(array == NULL)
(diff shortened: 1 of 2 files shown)


daryder:
Hi @mfournier, I've been running the latest version of the ceph plugin for a while now and I haven't run into any problems. Could you take a look at the changes I've made and see if this meets satisfies the concerns you had in the previous reply?

Thanks!

mfournier:
Thanks for the update @daryder ! Sorry, I wasn't able to spend much time on collectd lately.

So I confirm the new naming scheme plays well with graphite, thanks for the adaptations you made !

I'm now facing an awkward problem. I have a cluster with 2 nodes having 15 OSDs and 1 MON each. Same OS, same software versions (ceph 0.80.5 on ubuntu 12.04), same ceph config, etc. On one of them collectd runs fine. On the other one there's a contant memory usage increase which rapidely leads collectd to consume all the system memory.

Here's the output of valgrind/ms_print on both systems + the collectd config used to track down the issue (same collectd config on both systems, except for the sockets paths & `Daemon` names):

https://gist.github.com/mfournier/26f531f7b2429d02e6f2

I tried enabling each `Daemon` one by one, trying to find a possible rogue one, but no luck.

I didn't spend a lot of time investigating this yet, so maybe I'm missing something obvious. Let me know if this rings a bell or if I can provide more info !

daryder:
Thanks for taking a look so quickly, @mfournier. A couple of things: please check to make sure this is the latest version of the plugin which has my commits from https://github.com/daryder/collectd/commit/6f2dbf2f32c2d0a290b081b7141e824514494312. I noticed a problem while monitoring multiple daemons which could potentially result in the problem you are seeing.  The reason I ask to check the version is I noticed the line numbers in massif output didn't seem to match correctly to the latest plugin source. The latest commits I mention above contain some important bug fixes for monitoring multiple daemons. 

Were any logs generated which could give us more of a clue what is going on?

I haven't seen any memory leaks in my test cluster (running on 4-5 nodes). However, the cluster I'm testing with is a smaller test cluster - for most nodes only 1 OSD daemon so it could be an issue with the larger number of daemons you are monitoring which I haven't seen. I'll work on setting up collectd on our "production" Ceph cluster which is a much better test case.

Thanks!

mfournier:
Bummer, I indeed missed this last commit from early january ! I rebuilt with this last patch and the memory usage is now perfectly stable. Thanks, sorry for the noise !

crdc-ppm-os:
Hi, @mfournier, any update on this one?

daryder:
Hi, @mfournier, any update on this one?

mfournier:
@daryder, Thanks for being so patient ! I ran the latest version for a couple of days, and all looks good !

I have a couple of questions/feedbacks:

- running with debug enabled is extremely verbose. Are all the `DEBUG()` statements really useful to you ? If you don't mind, I'd suggest removing at least those which simply print out values just before passing them to `plugin_dispatch_values()`.

- I failed to find a reference in ceph's documentation of what all the collected values represent exactly. Are you aware of such a documentation ? Would you mind adding a pointer to it in the manpage in this case ?

- I noticed some graphs that were filled with empty values. Apparently "NaN" values get submitted to collectd for some reason (example below). Is this symptomatic of an of issue ? Or does ceph return some sort of non-numerical data here ? In this case, would it be an idea to filter them out ? What do you think ?

```
collectdctl getval storage202/ceph-osd.9/ceph_latency-RecoverystatePerf.startedLatency
send:    --> GETVAL "storage202/ceph-osd.9/ceph_latency-RecoverystatePerf.startedLatency"
receive: <-- 1 Value found
receive: <-- value=NaN
value=nan
```

pyr:
LGTM, :+1:

daryder:
Hi @mfournier,

In regards to your queries:

1. I removed a bunch of extra debug statements.

2. I've struggled with this as well - currently I don't think there is a clear set of documentation on these values. I've been able to find out what most of them are by searching around online and by looking into the ceph source code in the worst case. I've asked on the ceph mailer list a couple of times and apparently this is on the "to-do" list but not sure when something will come of it. Currently http://ceph.com/docs/master/dev/perf_counters/ is the only official doc on the performance counters - but that doesn't explain what each value represents.

3. This is purposeful; as in the example you shared this should only happen for: latency values for which there is no new data, and recovery state values for which there is no new data. The initial version of the plugin calculated latency values as a long-run average (average latency since the Ceph daemon started). In my case, and I think the typical case, this isn't very useful. This being the case, I added a config "LongRunAvgLatency" to allow users to utilize the initial long-run calculation. By default, this is set to "False", which means the plugin calculates the latency based on the last poll data (the plugin keeps track of the last poll and takes the delta between the current poll values and the last) which gives us more useful data (in my use case). As such, sometimes when the plugin runs there is no new ceph value for latency data (collecting intervals differ, some latencies are only calculated when certain ceph operations are run, etc.). In this case, the plugin will write the "NaN" value to collectd instead of writing a "0" (which I believe would be inaccurate).

The other counters this may show up for are recovery state / error state values. These may result in "NaN" because they will only collect meaningful data when the cluster is in an error state or recovering from some failure.

mfournier:
@daryder thanks for the last updates & answers !

Also @pyr had a look to the code (he has a much better eye for C than I do), so I'm pleased to make this an official plugin :-)

So many, many thanks to you @daryder, and also to all the other people who iterated over this piece of code !

One complain I suspect we'll get pretty quickly is that the plugin generates a lot of metrics. IMHO this isn't a huge issue as the filter chain system can be used to skip stuff users don't care about. But maybe sooner or later, we'll feel the need to select/discard metrics in the plugin itself, like the df or protocols plugins do. But let's wait for feedback on the current implementation first...

daryder:
@mfournier , @pyr thanks a lot for having a look and approving this plugin - very excited to hear it!

A few comments about your concern: Ceph is constantly changing, new releases are out fairly often. Since I have begun work on the plugin, a few official releases have been completed as well as many, many dev releases. As a part of this continuous release cycle, the performance counters are also changing (bug fixes, new counters added, old counters removed, etc). The plugin is coded the way it is to allow for the most flexibility when it comes to counters available from Ceph. If we get into selecting which metrics we want in the plugin itself, we may get caught having to keep track of a list of valid metrics for each major Ceph release.

mfournier:
Yes, I'm aware of this, @daryder. Kudos for making the plugin that flexible
by the way !

What I had in mind was something along the lines of what the "protocols"
plugin does: config options allowing the user to pick which metrics he
cares about, by matching the names against a set of regular expressions.

This way no actual metric name would be hardcoded in `ceph.c`. The
responsibility of keeping track of which counters are added/removed as ceph
moves forward is left to the user.

But as said, let's get 5.5 out and leave the dust settle first.

rzerres:
Kodos to @daryder, @mfournier,@pyr and all other parties makeing this plugin fly.

I have upgraded my collectd instance to make use of  performance_counters and forwarded the informations with network/rrdtool plugin. Tested on ArchLinux and  stable ubuntu (14.04) with success. Wonderful.

Now, as @mfournier already stated [ https://github.com/collectd/collectd/pull/598#issuecomment-66484443 ], I'd like to reduce the overwhelming information. Per default all ceph generated stats are pulled from the admin socket. Sincerely, most of us admins do just need an excerpt of that fields. Considering storage capacity and bandwith usage when piping with collectd  "network" plugin is working out. But collection3.cgi will have a hard time to render all this rrd-graphs.  

Therfore, I'd like to kindly request a proposal to implement a filter logic. When dumping  [ ceph daemon osd.<num> perf schema ], we do get the available ceph-tuples ( "instance": { "instance_field:  {value} }. GIven this output, we might store just "instance_name": {instance_field, instance_field} in a line based file. This file is  used as the input-filter, collecting just given tuples .... ??? That way a generic file could be dumped and user/admin just comments out non needed tuples or might delete non needed instance_fields.

Whatever might be practical/usefull. Looking forward for you thoughts.

daryder:
Hi @rzerres, thanks for the comments!

I think it's a good idea to implement a filter similar to what you proposed. However, I personally am not sure I'll be able to fulfill that request any time soon. It doesn't look to be a whole lot of work, but I believe I will remain busy for the near future.

If you want to go ahead and code this proposal that would be great - or we can see if anyone else steps up. Otherwise, I don't mind adding the feature but it may take some time until I get around to it.

octo:
@rzerres sounds reasonable. Could you please file a feature request in a separate issue for this? Comments on an already merged pull request are prone to being overlooked.

Thanks,
—octo

PR #631, opened 2014-06-03
hddtemp: Support more than 32 drives

bgilbert:
The hddtemp plugin imposes a 32-drive and 1024-byte limit on the response from the hddtemp daemon.  On my system, the 1024-byte limit truncates the response after 29 drives.

Drop the drive limit and increase the buffer limit to 1 MB.  (This is hopefully larger than anyone will need, and prevents collectd from being DoSed by the hddtemp daemon.)

### DIFF
--- a/src/hddtemp.c
+++ b/src/hddtemp.c
@@ -3,6 +3,7 @@
  * Copyright (C) 2005,2006  Vincent Stehlé
  * Copyright (C) 2006-2010  Florian octo Forster
  * Copyright (C) 2008       Sebastian Harl
+ * Copyright (C) 2014       Carnegie Mellon University
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the
@@ -22,6 +23,7 @@
  *   Vincent Stehlé <vincent.stehle at free.fr>
  *   Florian octo Forster <octo at collectd.org>
  *   Sebastian Harl <sh at tokkee.org>
+ *   Benjamin Gilbert <bgilbert at cs.cmu.edu>
  *
  * TODO:
  *   Do a pass, some day, and spare some memory. We consume too much for now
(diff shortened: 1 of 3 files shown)


src/hddtemp.c
```diff
@@ -247,55 +275,46 @@ static void hddtemp_submit (char *type_instance, double value)
...
 
 	/* NB: strtok_r will eat up "||" and leading "|"'s */
-	num_fields = 0;
 	ptr = buf;
 	saveptr = NULL;
-	while ((fields[num_fields] = strtok_r (ptr, "|", &saveptr)) != NULL)
+	while ((name = strtok_r (ptr, "|", &saveptr)) != NULL &&
+	       (model = strtok_r (NULL, "|", &saveptr)) != NULL &&
+	       (temperature = strtok_r (NULL, "|", &saveptr)) != NULL &&
+	       (mode = strtok_r (NULL, "|", &saveptr)) != NULL)
```
tokkee: Since you're touching this anyway, it should use strsplit() from common.h instead.

src/hddtemp.c
```diff
@@ -158,16 +166,41 @@ static int hddtemp_query_daemon (char *buffer, int buffer_size)
...
-	memset (buffer, '\0', buffer_size);
-
+	buffer = NULL;
+	buffer_size = 0;
 	buffer_fill = 0;
-	while ((status = read (fd, buffer + buffer_fill, buffer_size - buffer_fill)) != 0)
+	while (1)
 	{
-		if (status == -1)
+		if (buffer_fill >= buffer_size - 1)
```
tokkee: This will stop working if somebody decides to use size_t for buffer_fill and buffer_size (which sounds like a good choice to me in general). Then, if buffer_size == 0, buffer_size -1 will underflow. To fix this, either handle this case separately here or pre-allocate 1024 bytes before entering the loop and then drop the buffer_size == 0 case below.

tokkee:
The general approach sounds fine to me. I've added two comments on the code but otherwise didn't look at all details yet.

src/hddtemp.c
bgilbert: strsplit() has a hard-coded set of field separators and won't split on `|`.

src/hddtemp.c
bgilbert: Fixed.

octo:
Thanks for the patch @bgilbert. Could you please rebase on master so I can merge this? Thanks!

bgilbert:
Done.

octo:
Thank you so much @bgilbert!

PR #637, opened 2014-06-07
Statsd flexible histogram

### LINKED ISSUE #401: statsd plugin: percentile calculations don't seem to work reliably
On a system I have which is recieving real-world statsd data, the plugin sporadically emits values which seem to make sense compared to the average, but most of the time only emits zero (although avg is != 0 at that time). On my test system, I send the following data in a loop with netcat, and only 0s are sent to the write plugin:

```
glork:320|ms
glork:123|ms
glork:456|ms
glork:999|ms
glork:0|ms
glork:447|ms
glork:5632|ms
glork:732|ms
glork:34732|ms
glork:34|ms
glork:34|ms
glork:344|ms
```

And here is the relevant part of the config file:
```
<Plugin statsd>
  DeleteCounters false
  DeleteTimers   false
  DeleteGauges   false
  DeleteSets     false
  TimerPercentile 90.0
  TimerPercentile 95.0
</Plugin>
```

dothebart:
The collectd Statsd implementation currently will only accept values in the range from 0 - 1000 in the histogram it uses to calculate the percentile. In case of of values > 1000 these will be ignored; the Histogram is to be compared to an over exposure photo. 
To be able to adopt to other usecases the range of the histogramm has to become flexible so the user can configure the range of values his data series generate.

(fixes #401)

### DIFF
--- a/src/statsd.c
+++ b/src/statsd.c
@@ -84,7 +84,7 @@ static _Bool conf_delete_sets     = 0;
 static double *conf_timer_percentile = NULL;
 static size_t  conf_timer_percentile_num = 0;
 
-static int conf_timer_percentile_resolution_ms = 20;
+static int conf_timer_percentile_bucket_width_ms = 20;
 
 static _Bool conf_timer_lower     = 0;
 static _Bool conf_timer_upper     = 0;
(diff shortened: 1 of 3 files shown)


pyr:
closing in favor of #838 which supports dynamic bucket widths

PR #640, opened 2014-06-09
Limit the number of rrd_create threads

ymettier:
Hello,

PR #262 is a nice feature (create asynchronous threads for rrd files creation).

However, when too many new metrics come, there is still an issue : too many rrdcreate threads (I can have more than 1000 at the same time). And more there are, slower they become.

PR #244 has already proven to me that it was useful, but it does not help when the problem comes from too many RRD files to create.

So we should limit the number or create threads.

Check the attached patch and the new option `CreateFilesAsyncLimit`, that defaults to 0 (no limit, current behaviour) but that can be set to limit the number of threads.

```
CreateFilesAsync true
CreateFilesAsyncLimit 10
```

__Description__
There is a list (`async_creation_list`) of rrd files being created. I added a var (`async_creation_list_size`) with the size of this list.
Before the thread is created, `async_creation_list_size` is checked quickly (no lock) to see if the rrd file can be created or if there are already too many threads.
If there are too many threads, the metric is dropped and we hope that it will be sent again in 60 seconds (or whatever the interval is) and file can be created then.
This check is done fast, outside a pthread_locked section, so it is reliable only if there are too many threads.

If `async_creation_list_size` is small enough, the thread is created. A new check is done on `async_creation_list_size` because many threads can start and launch file creation at the same time (1st check was out of the pthread_locked section). The second check is reliable because it is done inside the locked section.

Note : 1st check is reliable when there are already too many threads running. This is why I implemented it : no need to start a new thread to see that it was useless.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -894,12 +894,14 @@
 #	DataDir "@localstatedir@/lib/@PACKAGE_NAME@/rrd"
 #	CreateFiles true
 #	CreateFilesAsync false
+#	CreateFilesAsyncLimit 0
 #	CollectStatistics true
 #</Plugin>
 
 #<Plugin rrdtool>
 #	DataDir "@localstatedir@/lib/@PACKAGE_NAME@/rrd"
 #	CreateFilesAsync false
+#	CreateFilesAsyncLimit 0
 #	CacheTimeout 120
 #	CacheFlush   900
 #	WritesPerSecond 50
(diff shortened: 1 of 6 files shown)


ymettier:
Hello,

It seems that this feature is implemented since Collectd-5.3 and comit 115ac52.

So I close this PR.

Thanks for implementing this feature.

Yves

PR #650, opened 2014-06-19
Aid the user in setting up filterchains by better logmessages

dothebart:
The builtin filterchain writer module requires to know the name of the writer module to send the data to. I didn't find any documentation about how the writers names were created; some of them are dynamicaly generated depending on the configuration. 
This patch outputs the name of all configured writers right next to the errormessage of being unable to choose one so the user can succeed in generating his config without source reading and documentation nightmares.

### DIFF
--- a/src/plugin.c
+++ b/src/plugin.c
@@ -299,6 +299,7 @@ static void log_list_callbacks (llist_t **list, /* {{{ */
 	}
 	INFO(str);
 	free(str);
+	free(keys);
 	
 }
 
--- a/src/plugin.c
+++ b/src/plugin.c
@@ -261,9 +261,17 @@ static void log_list_callbacks (llist_t **list, /* {{{ */
 	int len;
 	llentry_t *le;
 	int i;
-	int n = llist_size(*list);
-	char **keys = malloc(sizeof(char* *) * n);
-		
+	int n;
+	char **keys;
+
+	n = llist_size(*list);
+	keys = calloc(n, sizeof(char*));
+
+	if (keys == NULL)
+	{
+		ERROR("failed to allocate memory for list of callbacks");
+		return;
+	}
 
 	for (le = llist_head (*list), i = 0, len = 0;
 	     le != NULL;
@@ -300,8 +308,7 @@ static void log_list_callbacks (llist_t **list, /* {{{ */
 	INFO(str);
 	free(str);
 	free(keys);
-	
-}
+} /* }}} int log_list_callbacks */
 
 static int create_register_callback (llist_t **list, /* {{{ */
 		const char *name, void *callback, user_data_t *ud)
--- a/src/plugin.c
+++ b/src/plugin.c
@@ -253,6 +253,55 @@ static int register_callback (llist_t **list, /* {{{ */
 	return (0);
 } /* }}} int register_callback */
 
+static void log_list_callbacks (llist_t **list, /* {{{ */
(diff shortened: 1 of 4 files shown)


mfournier:
Reopened because I really believe filter chains could be a bit more user-friendly.

Let's at least discuss if this this patch would be helpful or not. @dothebart are you able to check if rebasing this against current master can be easily done ? Thanks !

src/utils_llist.h
```diff
@@ -60,4 +60,6 @@ llentry_t *llist_search_custom (llist_t *l,
 llentry_t *llist_head (llist_t *l);
 llentry_t *llist_tail (llist_t *l);
 
+
+
```
mfournier: Can you remove these unnecessary blank lines BTW, please ?

src/plugin.c
```diff
@@ -253,6 +253,56 @@ static int register_callback (llist_t **list, /* {{{ */
...
+	}
+	else
+	{
+		*str = '\0';
+		strjoin(str, len, keys, n, "', '");
+		INFO("%s ['%s']", comment, str);
+		free(str);
+	}
+	free(keys);
+} /* }}} int log_list_callbacks */
```
mfournier: `s/int/static void/` I assume ?

src/plugin.h
```diff
@@ -312,6 +312,7 @@ int plugin_unregister_data_set (const char *name);
 int plugin_unregister_log (const char *name);
 int plugin_unregister_notification (const char *name);
 
+void plugin_log_available_writers ();
```
mfournier: Would it be possible to write a couple of lines of doc here ? Thanks !

alexandruionica:
Just came here to say how much of PITA was to figure out the write_graphite callback name for a "Target" expression. I had to both read the source code of write_graphite.c and also do a gdb heap dump and examine it . But before this, first I had to actually figure out the source of the problems.

Please fix the error message so it also actually mentions possible targets.

mfournier:
FWIW c6a43371 (part of 5.3.2 and 5.4.2) adds a bit of doc & an example for this.

But I agree a bit of insight on what's happening, using the logs, would be even better !

src/plugin.c
```diff
@@ -253,6 +253,56 @@ static int register_callback (llist_t **list, /* {{{ */
...
+				const char *comment)
+{
+	char *str;
+	int len;
+	llentry_t *le;
+	int i;
+	int n;
+	char **keys;
+
+	n = llist_size(*list);
```
octo: I would simplify this function considerably, essentially doing:

```c
INFO ("Available write targets:");
for (le = llist_head (*list), i = 0, len = 0;
    le != NULL;
    le = le->next, i++)
{
  INFO ("* \"%s\"", le->key);
}
```

My reasoning is: This is not printed very often. If it is printed, there will be only a handful of lines, usually less then half a dozen. So all the allocating / prettyprinting is adding more complexity than I feel is justified.

octo:
Great idea, thank you!

src/plugin.c
dothebart: Since I developed this from a situation where I needed this Information, I can tell you it will get logged often; Essentially collectd isn't in a workable state then.
I first had implemented it the way you sugest, and it wouldn't deliver any usefull information anymore, but simply flood the log.

mfournier:
Superseded by #1043, so closing.

PR #651, opened 2014-06-19
New plugin: turbostat (CPU frequency and C-state residency)

Feandil:
This plugin is based on https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/tools/power/x86/turbostat/turbostat.c

It reads CPU frequency and C-state residency on modern Intel turbo-capable processors.

Here are examples of produced data (using CGP, see https://github.com/Feandil/collectd-turbostat/tree/master/CGP-plugin for the CGP plugin).
Example 1 (Small load):
collectd-turbostat_cgp
Example 2 (Big compilation near the end):
collectd-turbostat_cgp-2

### DIFF
--- a/src/turbostat.c
+++ b/src/turbostat.c
@@ -1168,14 +1168,14 @@ mark_cpu_present(int cpu)
 }
 
 static int __attribute__((warn_unused_result))
-allocate_cpu_set(cpu_set_t * set, size_t * size) {
-	set = CPU_ALLOC(topology.max_cpu_id  + 1);
-	if (set == NULL) {
+allocate_cpu_set(cpu_set_t ** set, size_t * size) {
+	*set = CPU_ALLOC(topology.max_cpu_id  + 1);
+	if (*set == NULL) {
 		ERROR("Unable to allocate CPU state");
 		return -ERR_CPU_ALLOC;
 	}
 	*size = CPU_ALLOC_SIZE(topology.max_cpu_id  + 1);
-	CPU_ZERO_S(*size, set);
+	CPU_ZERO_S(*size, *set);
 	return 0;
 }
 
@@ -1204,13 +1204,13 @@ topology_probe()
 		return -ERR_CALLOC;
 	}
 
-	ret = allocate_cpu_set(cpu_present_set, &cpu_present_setsize);
+	ret = allocate_cpu_set(&cpu_present_set, &cpu_present_setsize);
 	if (ret != 0)
 		goto err;
-	ret = allocate_cpu_set(cpu_affinity_set, &cpu_affinity_setsize);
+	ret = allocate_cpu_set(&cpu_affinity_set, &cpu_affinity_setsize);
 	if (ret != 0)
 		goto err;
-	ret = allocate_cpu_set(cpu_saved_affinity_set, &cpu_saved_affinity_setsize);
+	ret = allocate_cpu_set(&cpu_saved_affinity_set, &cpu_saved_affinity_setsize);
 	if (ret != 0)
 		goto err;
 
--- a/src/turbostat.c
+++ b/src/turbostat.c
@@ -69,7 +69,6 @@ static unsigned int do_rapl;
 static unsigned int do_dts;
 static unsigned int do_ptm;
 static unsigned int tcc_activation_temp;
-static unsigned int tcc_activation_temp_override;
 static double rapl_energy_units;
 
 #define RAPL_PKG		(1 << 0)
@@ -141,8 +140,8 @@ struct pkg_data {
 	unsigned int energy_gfx;	/* MSR_PP1_ENERGY_STATUS */
 	unsigned int rapl_pkg_perf_status;	/* MSR_PKG_PERF_STATUS */
 	unsigned int rapl_dram_perf_status;	/* MSR_DRAM_PERF_STATUS */
+	unsigned int tcc_activation_temp;
 	unsigned int pkg_temp_c;
-
 } *package_even, *package_odd;
 
 #define ODD_COUNTERS thread_odd, core_odd, package_odd
@@ -482,7 +481,7 @@ do {						\
 
 	if (do_dts) {
 		READ_MSR(MSR_IA32_THERM_STATUS, &msr);
-		c->core_temp_c = tcc_activation_temp - ((msr >> 16) & 0x7F);
+		c->core_temp_c = p->tcc_activation_temp - ((msr >> 16) & 0x7F);
 	}
 
 	/* collect package counters only for 1st core in package */
@@ -532,7 +531,7 @@ do {						\
 	}
 	if (do_ptm) {
 		READ_MSR(MSR_IA32_PACKAGE_THERM_STATUS, &msr);
-		p->pkg_temp_c = tcc_activation_temp - ((msr >> 16) & 0x7F);
+		p->pkg_temp_c = p->tcc_activation_temp - ((msr >> 16) & 0x7F);
 	}
 
 out:
@@ -980,10 +979,8 @@ set_temperature_target(struct thread_data *t, struct core_data *c, struct pkg_da
 	if (!(t->flags & CPU_IS_FIRST_THREAD_IN_CORE) || !(t->flags & CPU_IS_FIRST_CORE_IN_PACKAGE))
 		return 0;
 
-	if (tcc_activation_temp_override != 0) {
-		tcc_activation_temp = tcc_activation_temp_override;
-		ERROR("cpu%d: Using cmdline TCC Target (%d C)\n",
-			cpu, tcc_activation_temp);
+	if (tcc_activation_temp != 0) {
+		p->tcc_activation_temp = tcc_activation_temp;
 		return 0;
 	}
 
@@ -995,14 +992,14 @@ set_temperature_target(struct thread_data *t, struct core_data *c, struct pkg_da
 	if (target_c_local < 85 || target_c_local > 127)
 		goto guess;
 
-	tcc_activation_temp = target_c_local;
+	p->tcc_activation_temp = target_c_local;
 
 	return 0;
 
 guess:
-	tcc_activation_temp = TJMAX_DEFAULT;
-	WARNING("cpu%d: Guessing tjMax %d C, Please use -T to specify",
-		t->cpu_id, tcc_activation_temp);
+	p->tcc_activation_temp = TJMAX_DEFAULT;
+	WARNING("cpu%d: Guessing tjMax %d C, Please use TCCActivationTemp to specify",
+		t->cpu_id, p->tcc_activation_temp);
 
 	return 0;
 }
@@ -1454,6 +1451,7 @@ turbostat_init(void)
 	DO_OR_GOTO_ERR(check_dev_msr());
 	DO_OR_GOTO_ERR(setup_all_buffers());
 	DO_OR_GOTO_ERR(for_all_cpus(set_temperature_target, EVEN_COUNTERS));
+	DO_OR_GOTO_ERR(for_all_cpus(set_temperature_target, ODD_COUNTERS));
 
 	plugin_register_complex_read(NULL, PLUGIN_NAME, turbostat_read, NULL, NULL);
 
@@ -1463,8 +1461,32 @@ err:
 	return ret;
 }
 
+static const char *config_keys[] =
+{
+	"TCCActivationTemp",
+};
+static const int config_keys_num = STATIC_ARRAY_SIZE (config_keys);
+
+static int
+turbostat_config(const char *key, const char *value)
+{
+	long unsigned int tmp_val;
+	char *end;
+
+	if (strcasecmp("TCCActivationTemp", key) == 0) {
+		tmp_val = strtoul(value, &end, 0);
+		if (*end != '\0' || tmp_val > UINT_MAX)
+			return -1;
+		tcc_activation_temp = (unsigned int) tmp_val;
+	} else {
+		return -1;
+	}
+	return 0;
+}
+
 void module_register(void);
 void module_register(void)
 {
 	plugin_register_init(PLUGIN_NAME, turbostat_init);
+	plugin_register_config(PLUGIN_NAME, turbostat_config, config_keys, config_keys_num);
 }
--- a/src/turbostat.c
+++ b/src/turbostat.c
@@ -1249,7 +1249,7 @@ topology_probe()
 		ret = parse_int_file("/sys/devices/system/cpu/cpu%d/topology/core_siblings_list", i);
 		if (ret < 0)
 			goto err;
-		else if (ret == cpu->core_id)
+		else if (ret == i)
 			cpu->first_core_in_package = 1;
 
 		ret = get_threads_on_core(i);
@@ -1259,13 +1259,12 @@ topology_probe()
 			num_threads = ret;
 		if (num_threads > max_thread_id)
 			max_thread_id = num_threads;
-		if (num_threads > 1) {
-			ret = parse_int_file("/sys/devices/system/cpu/cpu%d/topology/thread_siblings_list", i);
-			if (ret < 0)
-				goto err;
-			else if (ret == num_threads)
-				cpu->first_thread_in_core = 1;
-		}
+		ret = parse_int_file("/sys/devices/system/cpu/cpu%d/topology/thread_siblings_list", i);
+		if (ret < 0)
+			goto err;
+		else if (ret == i)
+			cpu->first_thread_in_core = 1;
+
 		DEBUG("cpu %d pkg %d core %d\n",
 			i, cpu->package_id, cpu->core_id);
 	}
--- a/src/turbostat.c
+++ b/src/turbostat.c
@@ -1490,6 +1490,14 @@ turbostat_init(void)
 	struct stat sb;
 	int ret;
 
+	if (stat("/dev/cpu/0/msr", &sb)) {
+		ERROR("Turbostat plugin: Initialization failed: /dev/cpu/0/msr"
+		      " does not exist while the CPU supports MSR. You may be "
+		      "missing the corresponding kernel module, please try '# "
+		      "modprobe msr'");
+		return -1;
+	}
+
 	if (getuid() != 0) {
 		ERROR("Turbostat plugin: Initialization failed: this plugin "
 		      "requires collectd to run as root in order to read "
@@ -1498,14 +1506,6 @@ turbostat_init(void)
 	}
 
 	DO_OR_GOTO_ERR(probe_cpu());
-
-	if (stat("/dev/cpu/0/msr", &sb)) {
-		ERROR("Turbostat plugin: Initialization failed: /dev/cpu/0/msr"
-		      " does not exist while the CPU supports MSR. You may be "
-		      "missing the corresponding kernel module, please try '# "
-		      "modprobe msr'");
-		return -1;
-	}
 
 	DO_OR_GOTO_ERR(setup_all_buffers());
 
--- a/src/turbostat.c
+++ b/src/turbostat.c
@@ -787,255 +787,9 @@ for_all_cpus_delta(const struct thread_data *thread_new_base, const struct core_
 }
 
 
-static void
-free_all_buffers(void)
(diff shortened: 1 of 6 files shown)


Feandil:
Hi @octo,
I hope that I've resolved all your issues.

Unfortunately, this code now needs a rebase on the master branch, plus small changes to be able to merge it.

Feandil:
I've added a rebased branch to my repo, which contains the modifications to Makefile.am required by 4de560d and small modification to the AUTHORS files (there was a conflict): https://github.com/Feandil/collectd/commits/fastforward

Should I push-force it here?

Feandil:
Hi, I just rebased it on master in order to make it mergeable again.
Tell me if there is anything missing/broken in this module

Feandil:
Is there anything I can do to help you proceed with this request ?

mfournier:
Hi @Feandil ! I wonder why this hasn't been included yet, as @octo seems to have done quite a thorough review already and that you adressed his requests in later commits.

I tried out your patch and here are a few (minor) issues I noticed:

* When building on a 32bit architecture, gcc fails with this error:
```
DEBUG: turbostat.c: In function 'delta_package':
DEBUG: turbostat.c:458: error: integer constant is too large for 'long' type
DEBUG: turbostat.c:459: error: integer constant is too large for 'long' type
DEBUG: turbostat.c:460: error: integer constant is too large for 'long' type
DEBUG: turbostat.c:461: error: integer constant is too large for 'long' type
```
This will be annoying for integration in linux distribution supporting this architecture such as debian & fedora.

* The plugin includes `<asm/msr-index.h>` and `<cpuid.h>` which seem to be only in modern kernels/gcc versions. Would it be possible to check for their existence in the configure script ?

* It also includes `<sys/capability.h>` which comes with the libcap-devel/libcap-dev package. So a check on this is also needed + a mention of the dependence on this lib in the README would be nice.

Feandil:
Hi @mfournier, could you check again? These commit should answer your comments.

One note though, for 5.4.2, 844b714e82006e02e99af29693a7bd09fdb49884 removes too many headers: shed.h is missing (see https://github.com/Feandil/collectd/commit/246539819aefee0950a7673b60e505292edd9b31). I am not 100% sure that on all configurations (depending on what modules are selected) it will work or not

Feandil:
Sorry, 9b2419163f4a850bff9ddef9c9d50cb2b56e1fbf introduced a bug (inexisting cpus will report as cpu0, leading to multiple reports), fixed by 39755bc4eaaea9b453ed423253bc7c478fbd95e5.
It also revealed another bug fixed in 70e1e4c4fc031a971d8801903f9d11c3b1dd9bbe (which, if not for the first bug would only have a small memory and cpu impact).

mfournier:
Thanks for the updates @Feandil !

I passed this through our (prototype) continuous integration system, which checks the buildability on a bunch of different distros.

Here's a short summary:
 * centos5, freebsd: detected as unsupported, turbostat.c isn't built :+1:
 * centos6, centos7, ubuntu 14.04:
```
checking asm/msr-index.h usability... yes
checking asm/msr-index.h presence... yes
checking for asm/msr-index.h... yes
checking cpuid.h usability... yes
checking cpuid.h presence... yes
checking for cpuid.h... yes
checking sys/capability.h usability... yes
checking sys/capability.h presence... yes
checking for sys/capability.h... yes
```
... and build the plugin successfully :+1:
I noticed that `ldd src/.libs/turbostat.so` shows no mention of libcap. Not sure if it was missed out or if it's the correct way of doing in this case ?

* debian oldstable, stable and ubuntu 12.04: these ones successfully pass configure's tests (same output as above) but all fail with this error:
```
depbase=`echo turbostat.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
	/bin/bash ../libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I.  -I./daemon -DPREFIX='"/opt/collectd"' -DCONFIGFILE='"/opt/collectd/etc/collectd.conf"' -DLOCALSTATEDIR='"/opt/collectd/var"' -DPKGLOCALSTATEDIR='"/opt/collectd/var/lib/collectd"' -DPIDFILE='"/opt/collectd/var/run/collectd.pid"' -DPLUGINDIR='"/opt/collectd/lib/collectd"' -DPKGDATADIR='"/opt/collectd/share/collectd"'  -Wall -Werror -g -O2 -MT turbostat.lo -MD -MP -MF $depbase.Tpo -c -o turbostat.lo turbostat.c &&\
	mv -f $depbase.Tpo $depbase.Plo
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I./daemon -DPREFIX=\"/opt/collectd\" -DCONFIGFILE=\"/opt/collectd/etc/collectd.conf\" -DLOCALSTATEDIR=\"/opt/collectd/var\" -DPKGLOCALSTATEDIR=\"/opt/collectd/var/lib/collectd\" -DPIDFILE=\"/opt/collectd/var/run/collectd.pid\" -DPLUGINDIR=\"/opt/collectd/lib/collectd\" -DPKGDATADIR=\"/opt/collectd/share/collectd\" -Wall -Werror -g -O2 -MT turbostat.lo -MD -MP -MF .deps/turbostat.Tpo -c turbostat.c  -fPIC -DPIC -o .libs/turbostat.o
turbostat.c: In function 'get_counters':
turbostat.c:345:3: error: 'MSR_SMI_COUNT' undeclared (first use in this function)
turbostat.c:345:3: note: each undeclared identifier is reported only once for each function it appears in
turbostat.c:356:3: error: 'MSR_CORE_C3_RESIDENCY' undeclared (first use in this function)
turbostat.c:358:3: error: 'MSR_CORE_C6_RESIDENCY' undeclared (first use in this function)
turbostat.c:360:3: error: 'MSR_CORE_C7_RESIDENCY' undeclared (first use in this function)
turbostat.c:374:3: error: 'MSR_PKG_C2_RESIDENCY' undeclared (first use in this function)
turbostat.c:376:3: error: 'MSR_PKG_C3_RESIDENCY' undeclared (first use in this function)
turbostat.c:378:3: error: 'MSR_PKG_C6_RESIDENCY' undeclared (first use in this function)
turbostat.c:380:3: error: 'MSR_PKG_C7_RESIDENCY' undeclared (first use in this function)
turbostat.c:382:3: error: 'MSR_PKG_C8_RESIDENCY' undeclared (first use in this function)
turbostat.c:384:3: error: 'MSR_PKG_C9_RESIDENCY' undeclared (first use in this function)
turbostat.c:386:3: error: 'MSR_PKG_C10_RESIDENCY' undeclared (first use in this function)
turbostat.c:389:3: error: 'MSR_PKG_ENERGY_STATUS' undeclared (first use in this function)
turbostat.c:393:3: error: 'MSR_PP0_ENERGY_STATUS' undeclared (first use in this function)
turbostat.c:397:3: error: 'MSR_DRAM_ENERGY_STATUS' undeclared (first use in this function)
turbostat.c:401:3: error: 'MSR_PP1_ENERGY_STATUS' undeclared (first use in this function)
turbostat.c: In function 'probe_cpu':
turbostat.c:997:18: error: 'MSR_RAPL_POWER_UNIT' undeclared (first use in this function)
make[3]: *** [turbostat.lo] Error 1
make[3]: Leaving directory `/tmp/tmp.VOn7KAuKwL/collectd-5.4.1.842.g70e1e4c/src'
make[2]: Leaving directory `/tmp/tmp.VOn7KAuKwL/collectd-5.4.1.842.g70e1e4c/src'
make[1]: *** [all] Error 2
make[1]: Leaving directory `/tmp/tmp.VOn7KAuKwL/collectd-5.4.1.842.g70e1e4c/src'
```

* also, no more issue on 32bit architecture :+1:

mfournier:
Oh, and one thing I also wanted to mention: most new plugins are licences as MIT rather than GPL. No hard requirement, just an invitation to follow the trend ;-)

Feandil:
So:
- I've added a test for MSR_CORE_C3_RESIDENCY in configure.ac (I'm no expert in this kind of language, I hope it works on old autoconf tools
- The dependency on libcap was a recent addition, daebff6bb861acc0ed7390bcb101b6e9af85efff, allowing people to run this module without requiring collectd to run as root (ported from https://github.com/torvalds/linux/commit/98481e79b60a50d699b79292ff1b7e56e7fa8425#diff-9922d79dbe4fc0eeea2fb868dd74e8c1)
- Concerning the licence, this code comes from https://github.com/torvalds/linux/blob/master/tools/power/x86/turbostat/turbostat.c which is under GPL2. I cannot change its licencing

Feandil:
Sorry, wrong button -_-

mfournier:
Thanks a lot for these updates @Feandil !

There was a small issue with the very last patch of the serie. I just pushed a branch with a fix for this (dfe9bb8c) + 2 other details here: https://github.com/mfournier/collectd/commits/turbostat Can you take a look and import the patches into your branch if you agree with them ?

When trying out this plugin on my laptop, here are the values and identifiers I get:
```
host.turbostat-Average.frequency-cpu00 204.247941 1430479979
host.turbostat-Average.frequency-cpu01 287.644429 1430479979
host.turbostat-Average.frequency-cpu02 159.776051 1430479979
host.turbostat-Average.frequency-cpu03 330.855785 1430479979
host.turbostat-Buzy.frequency-cpu00 2003.031389 1430479979
host.turbostat-Buzy.frequency-cpu01 2134.114166 1430479979
host.turbostat-Buzy.frequency-cpu02 1936.677680 1430479979
host.turbostat-Buzy.frequency-cpu03 2151.205068 1430479979
host.turbostat-core00.percent-c3 13.948736 1430479979
host.turbostat-core00.percent-c6 50.962224 1430479979
host.turbostat-core02.percent-c3 12.602173 1430479979
host.turbostat-core02.percent-c6 48.885468 1430479979
host.turbostat-cpu00.percent-c0 10.196942 1430479979
host.turbostat-cpu00.percent-c1 24.892098 1430479979
host.turbostat-cpu01.percent-c0 13.478399 1430479979
host.turbostat-cpu01.percent-c1 21.610834 1430479979
host.turbostat-cpu02.percent-c0 8.250007 1430479979
host.turbostat-cpu02.percent-c1 30.262352 1430479979
host.turbostat-cpu03.percent-c0 15.380021 1430479979
host.turbostat-cpu03.percent-c1 23.132210 1430479979
host.turbostat.current-cpu00 0.000000 1430479979
host.turbostat.current-cpu01 0.000000 1430479979
host.turbostat.current-cpu02 0.000000 1430479979
host.turbostat.current-cpu03 0.000000 1430479979
host.turbostat-pkg00.percent-pc3 7.364878 1430479979
host.turbostat-pkg00.percent-pc6 20.668102 1430479979
host.turbostat-pkg00.percent-pc7 0.000000 1430479979
host.turbostat.temperature-core00 46.000000 1430479979
host.turbostat.temperature-core02 42.000000 1430479979
host.turbostat-TSC.gauge-cpu00 2526.951554 1430479979
host.turbostat-TSC.gauge-cpu01 2526.959081 1430479979
host.turbostat-TSC.gauge-cpu02 2526.972167 1430479979
host.turbostat-TSC.gauge-cpu03 2526.966890 1430479979
```

I find them a bit inconsistently organised. Following the logic used by the cpu plugin, what would you think of changing the names as follows ? Would it still make sense in the context of this plugin ?

```
host.turbostat-cpuNN.frequency-average
host.turbostat-cpuNN.frequency-busy # or was "buzy" intended ?
host.turbostat-cpuNN.percent-cX
host.turbostat-cpuNN.current
host.turbostat-cpuNN.gauge-tsc
host.turbostat-coreNN.percent-cX
host.turbostat-coreNN.temperature
host.turbostat-pkgNN.percent-pcX
```

One last question: from what I understood, the cpu/core numbering is taken care of by one of the support libs, right ? Can we assume that everything will go smoothly on systems with more than 99 CPUs ?

Thanks again ! Looking forward to see this merged !

Feandil:
Hi @mfournier, 

I've merged your changes, thanks.

Concerning the organization of the different values, here is what I get on my CPU:
```
turbostat-Average/frequency-cpuXX.rrd
turbostat-Buzy/frequency-cpuXX.rrd      # Yes, should be "busy"
turbostat-TSC/gauge-cpuXX.rrd
turbostat-coreXX/percent-c3.rrd
turbostat-coreXX/percent-c6.rrd
turbostat-coreXX/percent-c7.rrd
turbostat-cpuXX/percent-c0.rrd
turbostat-cpuXX/percent-c1.rrd
turbostat-pkgXX/percent-pc2.rrd
turbostat-pkgXX/percent-pc3.rrd
turbostat-pkgXX/percent-pc6.rrd
turbostat-pkgXX/percent-pc7.rrd
turbostat-pkgXX/power-Cor_W.rrd
turbostat-pkgXX/power-GFX_W.rrd
turbostat-pkgXX/power-Pkg_W.rrd
turbostat/current-cpuXX.rrd
turbostat/temperature-coreXX.rrd
turbostat/temperature-pkgXX.rrd
```
My reasoning behind this organisation is entirely linked to the visualization I use with CGP (the graphs in the first post (it changed a bit (busy frequency, TSC & SMI), but the idea is there), generated via https://github.com/Feandil/collectd-turbostat/blob/master/CGP-plugin/turbostat.json):
- Grouping all temperature in the same graph
- Grouping all average/busy frequency in their own graphs (they have a completely different scale)
- Grouping the SMI count for all cpus
- Grouping cpu states per cpu, core states per core, package state per package

I agree that you naming convensions looks cleaner, but I'm not sure if I will be able to group them as before. What do you recon ?

Concerning the cpu/core/pkg numbering:
- The numbers are given by the kernel, I hope that they scale properly ;) (see `/proc/stat`, `/sys/devices/system/cpu/cpu*/topology/core_id` and `/sys/devices/system/cpu/cpu*/topology/physical_package_id`)
- Inside my code, I should be always using ints to represent them, reading the files as %d, so I do not expect overflows there
- The returned indice is generated with a `printf("%02d")`, which means that unfortunately, on systems with >100 Cpus, the incides will be XX when < 100, XXX when < 1000 and > 100, etc
- One "potential" overflow is on `num_threads * num_cores * num_packages` which needs to be an int and `num_threads * num_cores * num_packages * sizeof(struct thread_data)` which needs to be a size_t. I can't garantee anything on system that have enough cores/packages to make this number explode...

mfournier:
Thanks for the clarifications @Feandil !

I must admit I don't use/know CGP, but it would be a shame bake one visualisation tool's assumptions/limitations into the plugin. It might seem a detail, but once this is in a release, it's a pain to change because backward compatibility must be guaranteed. So I'd really like to settle for the best/most logical naming scheme.

My suggestion is from the "consistency with the rest of the collectd plugins" point of view. That said, CPU-resource consumption definetely isn't my domain of expertise (ie: no clue what an SMI or a TSC even is :smile:), so if the specialists believe following another naming scheme makes more sense, then let's go for it !

Also, I just tried it out on my laptop. I can imagine potential users for this plugin would be folks running large computing clusters, which need an overview of resources consumption across a range of nodes. @faxm0dem, as such user, would you have an opinion about this naming question ?

Feandil:
- TSC: "Time-Stamp Counter" defined in section 17.13 of volume 3B of Intel® 64 and IA-32 Architectures Software Developer’s Manual. It is used in this plugin to evaluate the percentage of time spent in each state. Turbostat first relied on the "invariant TSC" feature of modern intel, which ensures that `The invariant TSC will run at a constant rate in all ACPI P-, C-. and T-states.`. For non-invariant TSCs, this is not ensured, so we need to log it: if it's stable, turbostat is working as expected. If not, randomness happens.

- SMI: "System Management Interrupt" defined in section 2.2 of volume 3B/section 34.3 of volume 3C of Intel® 64 and IA-32 Architectures Software Developer’s Manual. It puts the processor into "System Management Mode" when received. You might have heard about it in some recent security discussions (e.g. https://twitter.com/grsecurity/status/578558748496568322)

I'm not sure this module is for people running large computer clusers: I would not expect them to have C-states enabled. C-states are a power consumption reduction feature (which can 'hurt' perfomances), so I believe that this plugin is more for people who want to understand their power consumption (and why their cores/packages don't go into low power states). That's at least why I used turbostat in the first place.

I quoted CGP because it's the visualisation I'm using. I'm not sure when and how other visualisation tools "merge" RRDs into a single graph by default. AFAIK CGP merges RRDs with the same (host, plugin, plugin-instance, type), i-e only merges values that differ by their type-instance only. Do you know how other tools do it by default?

mfournier:
Ok, so it really looks like the naming was chosen after CGP's capabilities.

I don't use rrd-based frontends myself (anymore), but as far as I can tell, graphite/grafana, influxdb, riemann, etc, have no problem aggregating/grouping on any element of the metric name path. Also, running `rrdtools graph ...` doesn't care about collectd's naming scheme. It just knows about .rrd files.

I pushed a patch to https://github.com/mfournier/collectd/tree/turbostat (eff671de) which reorganises the metrics in a consistent hierarchical way. Can you take a look ?

mfournier:
Any thoughts about my last comment @Feandil ? I'd love to find a consensus and be able to merge this soon :-)

Feandil:
Sorry @mfournier, I didn't have time to really look at CGP/other visualization tools to see if I would be able to cope with the new schema.

I don't really care in the end: as I'm using gentoo, it's trivial for me to maintain a local patch that change the names if I really need it... Your patch looks good, I'll merge it.

I see that "This pull request contains merge conflicts that must be resolved.", do you want me to rebase this branch (and maybe squash some commits?)

src/turbostat.c
```diff
@@ -0,0 +1,1611 @@
...
+
+	if ((!aperf_mperf_unstable) || (!(t->aperf > t->tsc || t->mperf > t->tsc)))
+		turbostat_submit(name, "frequency", "busy", 1.0 * t->tsc / 1000000 * t->aperf / t->mperf / interval_float);
+
+	/* Sanity check (should stay stable) */
+	turbostat_submit(name, "gauge", "TSC", 1.0 * t->tsc / 1000000 / interval_float);
+
+	/* SMI */
+	if (do_smi)
+		turbostat_submit(name, "current", NULL, t->smi_count);
```
Feandil: Speaking of cleaning, this probably needs to be changed: the SMI count has nothing to do with currents (I just used it because I didn't know what type to use): what do you recon?

mfournier:
Ok, great @Feandil !

Don't worry about the merge conflict. It's a minor detail in the rpm specfile.

About SMI, if I understand correctly, the value represents the amount of times the CPU was put in "System Management Mode", right ? In this case this would be the only metric in this plugin which is a counter, rather than a gauge, so this would require changing (or duplicating) `turbostat_submit()` to handle this case :-/

If I misunderstood, then you can just pick a more meaningful type name from `grep GAUGE src/types.db`, or simply use the generic `gauge` type.

If you want to squash some of the commits together, feel free to. But no hard requirement on my side!

Feandil:
Ok, I won't rebase nor squash in that case :)

Yes, rewriting/duplicating `turbostat_submit` is, I think, non-optimal. Using the "gauge" type would mean that both the SMI count and the TSC would have the same type. What do you think of the existing count (which is not a counter): `count                   value:GAUGE:0:U`?

mfournier:
So just to clarify a possible misunderstanding: the type chosen for gathered metrics does matter for most output plugins. When using a counter- or derive-based type, the delta between each mesurement is what will get stored. For gauge-based types, the collected value is what gets stored.

So if the chosen type doesn't match the nature of the collected value, users will have to convert them at display time, which is something we want to avoid.

Then something I initially overlooked was the `delta_thread()` function, which seems to take care of this conversion at the plugin level, right ? This would mean that all the values dispatched by this plugin gauges. If you confirm my interpretation is correct then +1 for using the `count` type.

Feandil:
Here it is.

Due to the fact that "count" comes only in 5.5 and that I'm using 5.4 myself, I was not able to test the last commit, but I don't see why it would fail.

Cheers,
Vincent

mfournier:
Thanks @Feandil for this last update! At last this plugin is merged, thanks so much for all this work and for your patience :-)

faxm0dem:
:+1: I will most certainly use this in production soon, thanks for the hard work and contribution!

PR #667, opened 2014-07-21
Enhance/copy of the curl_json plugin so it can query jolokia.org

dothebart:
Http://jolokia.org is another way to query jmx counters. It breaks the dependency chain by specifying a REST/HTTP API to access the values of http counters.

The module itself is intended to be configured by using the attached python script.
The configuration of the plugin consists of 3 parts:
- URL setup of the servlet container
- the Json Post data, which is a 'BULK-Request' in the nomenclature of jolokia
- a table of matches from the JMX names and their Attributes to the name of the metric on collectd side

The Python script uses the pyjolokia library to access your jolokia host. It queries the list interface of jolokia, which gives you all available beans. it then filters this list by patterns, so you don't query unavailable beans. 

The plugin was tested with weblogic.

Another companion product of jolokia to be mentioned is http://hawt.io

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4794,6 +4794,7 @@ plugin_contextswitch="no"
 plugin_cpu="no"
 plugin_cpufreq="no"
 plugin_curl_json="no"
+plugin_curl_jolokia="no"
 plugin_curl_xml="no"
 plugin_df="no"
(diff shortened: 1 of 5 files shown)


dothebart:
closing in favour of #2470

PR #678, opened 2014-07-28
OKAY notification in missing values broken.

manuelluis:
Hello.

This is a slightly different version of https://github.com/collectd/collectd/pull/225

There are two changes form the original patch of @mce35:

Change the state to `STATE_MISSING` in `ut_missing`
to `uc_check_timeout` after call `plugin_dispatch_missing` because
threshold is a plugin (maybe you can use another plugin to track missings...)

Also in `uc_check_timeout` instead of not call `plugin_dispatch_missing` 
if value is missing, not put in the value in the list of entries to be flushed if
it's missing.

I was using this feature in collectd 4, would be nice to recover it in collectd 5.

I've been using this patch for a couple of months in production.

### DIFF
--- a/src/aggregation.c
+++ b/src/aggregation.c
@@ -299,7 +299,7 @@ static int agg_instance_update (agg_instance_t *inst, /* {{{ */
     return (EINVAL);
   }
 
-  rate = uc_get_rate (ds, vl);
+  rate = uc_get_rate (ds, vl, /* include missing = */ 0);
   if (rate == NULL)
   {
     char ident[6 * DATA_MAX_NAME_LEN];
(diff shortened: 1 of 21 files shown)


octo:
FYI, this cannot be merged due to pending review comments.

rpv-tomsk:
Hi!

Let's recover this PR?

I think we need two versions of `uc_get_rate`, one left untouched (w/o 'include_missing' third argument), and other modified for needs. So we will have no massive changes of plugins code.

rpv-tomsk:
Also, I have a proposal. Maybe it will be better to add new field(s) to `threshold_t` instead?

PR #709, opened 2014-08-14
TSDB metadata

kev009:
This allows collectd to closely align with OpenTSDB concepts without drastic changes to either system by using the metadata and filter chains in collectd:
* http://opentsdb.net/docs/build/html/user_guide/writing.html
* https://collectd.org/wiki/index.php/Meta_Data_Interface
* https://collectd.org/wiki/index.php/Chains

We use prefixes for organizing metrics into trees, and it is particularly useful for i.e. separating systems metrics vs applications metrics:  "app.", "sys." and then making those easily browsable in UIs like http://grafana.org/ or Graphite (with the TSDB backend https://github.com/graphite-project/graphite-web/issues/693).

The tag is a central component of TSDB.  Using filter chains, complex tagging is easily available for all plugins and can be driven by config management.  Since we use the collectd metadata API, it is also a first class attribute usable by the collectd perl and python plugins.

A quite simple example to prefix all collectd metric names with "sys." and add an arbitrary "status" tag:
```
<Chain "PreCache">
  <Rule "add_sys_prefix">
    <Match "regex">
      Host "^"
    </Match>
    <Target "set">
      TSDBPrefix "sys."
      TSDBTags "status=production"
    </Target>
  </Rule>
</Chain>
```
target-set1

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -7789,7 +7789,7 @@ Example:
 
 =item B<TSDBPrefix> I<String>
 
-=item B<TSDBTags> I<Strings>
+=item B<TSDBTags> I<String>
 
 These are used by the B<write_tsdb> plugin. The prefix is prepended to the
 metric name and should include the separator, i.e. "sys.". Tags are passed as
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -7785,6 +7785,26 @@ Example:
    TypeInstance "core3"
  </Target>
 
+=over 4
+
+=item B<TSDBPrefix> I<String>
+
+=item B<TSDBTags> I<Strings>
+
+These are used by the B<write_tsdb> plugin. The prefix is prepended to the
+metric name and should include the separator, i.e. "sys.". Tags are passed as
+name=value pairs along with the metrics and aid in filtering and aggregating
+metrics in OpenTSDB.
+
+=back
+
+Example:
+
+ <Target "set">
+   TSDBPrefix "sys."
+   TSDBTags "devicerole=webserver"
+ </Target>
+
 =back
 
 =head2 Backwards compatibility
--- a/src/target_set.c
+++ b/src/target_set.c
@@ -35,6 +35,8 @@ struct ts_data_s
   char *plugin_instance;
   /* char *type; */
   char *type_instance;
+  char *tsdb_prefix;
+  char *tsdb_tags;
 };
 typedef struct ts_data_s ts_data_t;
 
@@ -108,6 +110,8 @@ static int ts_destroy (void **user_data) /* {{{ */
   free (data->plugin_instance);
   /* free (data->type); */
   free (data->type_instance);
+  free (data->tsdb_prefix);
+  free (data->tsdb_tags);
   free (data);
 
   return (0);
@@ -132,6 +136,8 @@ static int ts_create (const oconfig_item_t *ci, void **user_data) /* {{{ */
   data->plugin_instance = NULL;
   /* data->type = NULL; */
   data->type_instance = NULL;
+  data->tsdb_prefix = NULL;
+  data->tsdb_tags = NULL;
 
   status = 0;
   for (i = 0; i < ci->children_num; i++)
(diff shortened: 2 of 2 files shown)


ymettier:
Hello,

I'm adressing here this issue and the issue #887.

I think the real need for opentsdb write plugin is to redefine the UID and define tags.

Define the UID
* add a prefix
* in some case (including the df or cpu from the example of issue #887), remove PluginInstance and/or TypeInstance

Add tags
* add some static tags (like `"status=production"` here)
* add some tags with values from PluginInstance and/or TypeInstance (like in issue #887)

### Example/Suggestion :
```
<Chain "PreCache">
  <Rule "opentsdb_df">
    <Match "regex">
      Plugin "^df$"
    </Match>
    <Target "set">
      MetaDataSet "tsdb_id" "sys.df.%Type%.%TypeInstance%"
      MetaDataAdd "tsdb_tags" "mount=%PluginInstance%"
    </Target>
  </Rule>

  <Rule "opentsdb_cpu">
    <Match "regex">
      Plugin "^cpu$"
    </Match>
    <Target "set">
      MetaDataSet "tsdb_id" "sys.cpu.%Type%"
      MetaDataAdd "tsdb_tags" "cpu=%PluginInstance%"
    </Target>
  </Rule>

  <Rule "add_sys_prefix">
    <Match "regex">
      Host "^"
    </Match>
    <Target "set">
      MetaDataSet "tsdb_prefix" "sys."
      MetaDataAdd "tsdb_tags" "status=production"
    </Target>
  </Rule>
</Chain>
```
Note : this merges the need to add "sys." as a prefix and the 2 examples from issue #887 :

Having 

```
put df.df_complex.reserved 1420724763 12345678 fqdn=machine.domain.tld,mount=boot
put cpu.idle 1420724763 12345678 fqdn=machine.domain.tld,cpu=0
```

instead of 

```
put df.boot.df_complex.reserved 1420724763 12345678 fqdn=machine.domain.tld
put cpu.0.cpu.idle 1420724763 12345678 fqdn=machine.domain.tld
```

### Explanation

1/ add a `tsdb_id` metadata.

In the write_tsdb plugin, if the `tsdb_id` exists, use it and ignore p.pi.t.ti.dsname

Also ignore any tsdb_prefix ? I'm not sure. If you take my example, yes, ignore it, because it is already in the `tsdb_id`. I'm not sure because it may be a good idea to have a generic prefix.

If the `tsdb_id` does not exist, create it from p.pi.t.ti as it is already done in write_tsdb.c.

2/ add tags with values from pluginInstance and/or typeInstance

As show in the example, we need 3 variables :
* `%pluginInstance%`
* `%typeInstance%`
* `%dsname%`
I don't think we need other variables

The metadata tags will take their values from the pi or ti or dsname from the metric.




What do you think ?

Regards,
Yves

ymettier:
Hello,

I wrote a patch in order to implement `MetaDataSet`. Please see PR #1106.

While I was writing this patch, I was thinking on how to use it here.
First thing : using variables (like %pluginInstance%) will make the code a lot more complex. This is a bad idea.

Here is my new suggestion/example : 
```
<Chain "PreCache">
  <Rule "opentsdb_df">
    <Match "regex">
      Plugin "^df$"
    </Match>
    <Target "set">
# put df.df_complex.reserved 1420724763 12345678 fqdn=machine.domain.tld,mount=boot
      MetaDataSet "tsdb_prefix" "sys."
      MetaDataSet "tsdb_tag_pluginInstance" "mount"
    </Target>
  </Rule>

  <Rule "opentsdb_cpu">
    <Match "regex">
      Plugin "^cpu$"
    </Match>
    <Target "set">
# put sys.cpu.idle 1420724763 12345678 fqdn=machine.domain.tld,cpu=0
      MetaDataSet "tsdb_prefix" "sys."
      MetaDataSet "tsdb_tag_pluginInstance" "cpu"
      MetaDataSet "tsdb_tag_type" ""
    </Target>
  </Rule>

  <Rule "add_sys_prefix">
    <Match "regex">
      Host "^"
    </Match>
    <Target "set">
      MetaDataSet "tsdb_prefix" "sys."
      MetaDataAdd "tsdb_tags" "status=production"
    </Target>
  </Rule>
</Chain>
```

The write_tsdb plugin would work like this : 
* If the tsdb_prefix is defined, add it.
* if `tsdb_tag_pluginInstance` is defined and is empty, remove pluginInstance from the ID
* if `tsdb_tag_pluginInstance` is defined and is not empty, remove pluginInstance from the ID, but add it as a tag
* Same for `tsdb_tag_type`
* Same for `tsdb_tag_typeInstance`
* Same for `tsdb_tag_dsname`

Not in this example, but another meta data could be `tsdb_id` in order to force the id (but don't use variables as I suggested in my previous post).

What do you think ?

Regards,
Yves

ymettier:
Hello,

Please check if PR #1107 can be a solution.

Regards,
Yves

dlmarion:
Any ETA on getting a solution to the TSDB issue resolved? Is the Pull Request satisfactory?

Dieken:
I think this pull request is superseded by pull request #1107.

PR #721, opened 2014-08-29
adding support for using irq assignments not numbers

gilesw:
This is to give the option of a more readable key for the irq, like munin.


babs:
I've cleaned a few things and submitted it as #1373

octo:
Superseded by #1373

PR #736, opened 2014-09-17
network: initialize ai_flags of client socket

mfournier:
Bitwise OR-ing options on `ai_flags` without properly resetting it
silently enabled the `AI_PASSIVE` flag under certain conditions. So this
left collectd *clients* with an UDP socket open on 0.0.0.0.

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -2107,6 +2107,7 @@ static int sockent_client_connect (sockent_t *se) /* {{{ */
 		return (0);
 
 	memset (&ai_hints, 0, sizeof (ai_hints));
+	ai_hints.ai_flags  = 0;
 #ifdef AI_ADDRCONFIG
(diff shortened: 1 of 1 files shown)


rubenk:
I did some testing with gdb too, and we do set the ai_flags correctly, to 0x20, AI_ADDRCONFIG.
Apparently the socket showing up in netstat is normal behaviour on Linux.

From https://stackoverflow.com/questions/25647098/create-udp-client-socket-without-socket-listening:
"UDP sockets have a kernel buffer for incoming messages. This buffer is maintained by the kernel regardless whether you call recv/recvfrom/recvmsg from user-space code."

mfournier:
Yes, thanks for the explanations. I guess we can close this issue then. Sorry for the confusion this created !

PR #742, opened 2014-09-20
Update bind plugin to support the new XML v3 API.

elinorbgr:
This commit adds support for the new XML schema of Bind9 statistics-channels (version 3). This new schema is the default one in Bind9.9 (can be reverted to v2 with a `./configure` switch) and support for older versions has been dropped in Bind9.10.

Support for versions 1 and 2 remain in the collectd plugin.

Unfortunately, this new schema does not provide per-zone data any more, thus 'Zone' options like 

```
Zone "example.com/IN"
```

are ignored if schema version is 3.

Before this, a v3 schema was silently failed with only `read-function of plugin `bind' failed.`.

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -670,29 +670,29 @@ static int bind_parse_generic_name_attr_value_list (const char *xpath_expression
       if (child->type != XML_ELEMENT_NODE)
         continue;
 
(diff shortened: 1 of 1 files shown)


johntdyer:
Doesn't look like this made it into the changelog for 5.4.2.  Any reason why?

Issue #737, opened 2014-09-18
Java plugin used in high volume application causes process to grow very large


fhossain:
We are running the Java plugin that publishes thousands of metrics per minute. Over a period of a week the 500 M process grows to 18G in virtual and real memory. The problem appears to the way JNI requests are made. I refactored java.c based on JNI best practice as in http://www.ibm.com/developerworks/library/j-jni/
My refactored code is base collectd 5.4.0. Diff is attached. It would be great to have these changes reviewed and integrated to the master.

```diff
diff --git a/src/java.c b/src/java.c
index b69ca94..ab940f9 100644
--- a/src/java.c
+++ b/src/java.c
@@ -74,6 +74,78 @@ struct cjni_callback_info_s /* {{{ */
 typedef struct cjni_callback_info_s cjni_callback_info_t;
 /* }}} */
 
+struct cjni_base_md_s /* {{{ */
+{
+  jclass    class;
+  jmethodID constructor;
+};
+typedef struct cjni_base_md_s cjni_base_md_t;
+/* }}} */
+
+struct cjni_list_md_s /* {{{ */
+{
+  jclass    class;
+  jmethodID toArray;
+};
+typedef struct cjni_list_md_s cjni_list_md_t;
+/* }}} */
+
+struct cjni_dataset_md_s /* {{{ */
+{
+  jclass    class;
+  jmethodID constructor;
+  jmethodID addDataSource;
+};
+typedef struct cjni_dataset_md_s cjni_dataset_md_t;
+/* }}} */
+
+struct cjni_datasource_md_s /* {{{ */
+{
+  jclass    class;
+  jmethodID constructor;
+  jmethodID setName;
+  jmethodID setType;
+  jmethodID setMax;
+  jmethodID setMin;
+};
+typedef struct cjni_datasource_md_s cjni_datasource_md_t;
+/* }}} */
+
+struct cjni_valuelist_md_s /* {{{ */
+{
+  jclass    class;
+  jmethodID constructor;
+  jmethodID setDataSet;
+  jmethodID setHost;
+  jmethodID getHost;
+  jmethodID setPlugin;
+  jmethodID getPlugin;
+  jmethodID setPluginInstance;
+  jmethodID getPluginInstance;
+  jmethodID setType;
+  jmethodID getType;
+  jmethodID setTypeInstance;
+  jmethodID getTypeInstance;
+  jmethodID addValue;
+  jmethodID getValues;
+  jmethodID setTime;
+  jmethodID getTime;
[...]

PR #747, opened 2014-09-27
Refactored java.c to cache JNI methods and classes and skip detach for r...

fhossain:
@octo here are the changes for "Java plugin used in high volume application causes process to grow very large #737"


mahmoudimus:
ping - wanted to raise @pyr's attention to this patch as well

rubenk:
@pyr a friendly reminder, is still on your radar?

PR #783, opened 2014-10-28
Add support for varnish 4.x

### LINKED ISSUE #618: varnish plugin doesn't work with Varnish 4.0
collectd 5.4.0 in Fedora Rawhide FTBFS because it cannot find varnishapi.h

configure.ac checks for varnishapi.h but looking at the Varnish source this header got dropped somewhere in 2011.

Unfortunately I'm not familiar enough with Varnish or its api to propose a fix.

mfournier:
I built this patchset against libvarnishapi 2.0.6, 3.0.2 and 4.0.2, but I didn't have the opportunity to run it on production varnishes yet. As the changes are well confined inside `#ifdef` blocks, I'm pretty confident that any problem possibly introduced here would be limited to varnish 4.x.

Would fix #618

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6305,6 +6305,10 @@ Take the UUID from the given file (default I</etc/uuid>).
 =head2 Plugin C<varnish>
 
 The I<varnish plugin> collects information about Varnish, an HTTP accelerator.
+It collects a subset of the values displayed by L<varnishstat(1)>, and
+organizes them in categories which can be enabled or disabled. Currently only
+metrics shown in L<varnishstat(1)>'s I<MAIN> section are collected. The exact
+meaning of each metric can be found in L<varnish-counters(7)>.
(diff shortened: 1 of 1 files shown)


disassembler:
Is there any chance of this getting backported to 5.4? Updating to varnish 4 and have no collectd monitoring ability right now with it. I was hoping 5.4.2 was going to have it added.

quiffman:
:+1: 

I spent a couple of hours yesterday trying to figure out why the new 5.4.2 release wasn't compiling properly against varnish4, before I realised this patch hadn't been included.

disassembler:
FYI, I was able to get it to compile against 5.4.2 and am receiving varnish 4 data using the following:
curl https://github.com/collectd/collectd/pull/783.patch | git am
It will fail on a couple files, but they looked like they were just documentation. Just do git am --skip on the ones it fails on.

PR #790, opened 2014-11-05
Add sysctl plugin

rottenbytes:
Hi,

We have developed at Fotolia a sysctl plugin, allowing us to fetch some values only availaible from sysctl and we would like to merge it upstream.

The plugin is running smoothly on Linux and FreeBSD, but the code may require some modifications to make it portable through multiple OS (like net|openBSD) by checking the existence of function _sysctlbyname(3)_, than I don't know how to do.


rubenk:
Closing stale issue.

PR #830, opened 2014-12-01
curl_json: allow escaping the separator

radford:
`curl_json` specifies keys via a `/`-separated path through a JSON structure.  This works well until someone puts a slash in their `JSON` key, like the `HHVM` memory statistics do.  To work around this, `/`s can now be escaped with `\`s.  This is not technically backwards compatible, as old keys with `\`s will need an extra `\` to escape them now, but finding a `\` in a key is even more improbable than a `/` so it should be fine.

ChangeLog: curl_json plugin: Allow backslash escaping the key-path separator.

### DIFF
--- a/src/curl_json.c
+++ b/src/curl_json.c
@@ -439,6 +439,29 @@ static int cj_config_append_string(const char *name,
   return 0;
 } /* }}} int cj_config_append_string */
 
+static const char *cj_extract_escaped_key(const char *from, char *into,
+                                          size_t into_size) {
+  bool escaped = false;
+  int len = 0;
+  const char *cur = from;
+  for (; *cur; cur++) {
+    bool escape = *cur == '\\', sep = *cur == '/';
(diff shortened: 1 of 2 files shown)


dago:
@radford I know it has been quite some time since your contribution, but maybe you would be willing to put this forward by rebasing it and fixing the conflicts? I would like to include this PR in the upcoming 5.11 release.

radford:
@dago, It has been quite some time!  I more or less had to re-write the patch because of the conflicts.  The good thing is the patch is now more testable.  An even better thing is you now have a testing infrastructure, so I added tests.   I made this version more strict by rejecting unexpected escapes.

octo:
Hi @radford, thank you so much for picking this up even after such a long time!

Could you please run `contrib/format.sh src/curl_json_test.c` to fix the formatting of the unit test?

Thanks and best regards,
—octo

fcorneli:
Any news on merging this patch? I also stumbled upon the escaping problem of "/" within the key.

PR #833, opened 2014-12-02
AMQP connection failure leads collectd to use 100%CPU

yogeswaran:
Summary: AMQP connection failure leads collectd to use 100%CPU
Issue : https://ipm.bloomberg.com/jira/browse/SS-103
Solution : When amqp connection fails, it retries for every message, added a fix to retry only after a delay. A new configuration "ConnectionRetryDelay"  has been introduced with default value as 60 seconds.

Test Plan:
1. Tested without the new configuration
2. Tested with the new configuration

Verified connection is retried only after the delay and made sure collectd is not taking 100% cpu.

### DIFF
--- a/src/amqp.c
+++ b/src/amqp.c
@@ -80,6 +80,9 @@ struct camqp_config_s
     char   *exchange;
     char   *routing_key;
 
+    /* Number of seconds to wait before connection is retried */
+    int     connection_retry_delay;
+
     /* publish only */
     uint8_t delivery_mode;
     _Bool   store_rates;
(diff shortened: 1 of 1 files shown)


mfournier:
I took a moment to test this patch and merged it to master a couple of minutes ago. Thanks for yet another valuable contribution @yogeswaran :-)

A couple of comments:
- as it adds a configuration option, it isn't suitable for the collectd-5.3 bugfix branch you submitted the PR against. It will be part of upcoming version 5.5 though.
- from my testing, collectd does *not* consume 100% CPU when the failure condition is met. It just consumes slightly more CPU than under normal condition (that is, using librabbitmq0 0.0.1.hg216-1 on debian 7). If you have 12 metrics per minute passing through this code path, this won't be an issue. If you have 12k, it's probably another story.
- therefore, I changed the default value for `ConnectionRetryDelay` to 0, disabling the reconnection delay feature by default. I don't think we should impose this delay to everyone using this plugin, as the issue you mention probably doesn't affect most users (we would have heard about it earlier I guess). I'm of course open to discuss this if you disagree !

Thanks again !

PR #835, opened 2014-12-03
RPM specfile: switch to using upstart/systemd

mfournier:
Favor using systemd when installing on EL7, and upstart on EL6. No
changes for EL5. Uses the upstart & systemd config files from
the contrib/ directory.

The sysv initscript used on EL5 is still available in the
collectd-contrib package, for those who prefer using the traditional
init system.


mfournier:
@mspiegele, yes so this assumes you provide a tarball of the master branch, which you can obtain by cloning the git repo, then: `./build.sh && ./configure && make dist`.
You'll also need to adapt the Version field in the specfile.

PR #836, opened 2014-12-03
processes plugin: add per-process open files count

mfournier:
Only implemented on Linux for now. On other platforms no value will get
submitted.

Counts the number of files in the `/proc/<pid>/fd/` directory.
The function used to count the number of threads has been slightly
modified to work for this case too.

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -742,7 +742,7 @@ static void ps_submit_proc_list (procstat_t *ps)
 	vl.values_len = 2;
 	plugin_dispatch_values (&vl);
 
-	if (ps->num_fd != -1)
+	if (ps->num_fd > 0)
 	{
 		sstrncpy (vl.type, "files", sizeof (vl.type));
 		vl.values[0].gauge = ps->num_fd;
@@ -1135,8 +1135,8 @@ int ps_read_process (int pid, procstat_t *ps, char *state)
 		if ((ps_read_status(pid, ps)) == NULL)
 		{
 			/* No VMem data */
-			ps->vmem_data = -1;
-			ps->vmem_code = -1;
+			ps->vmem_data = 0;
+			ps->vmem_code = 0;
 			DEBUG("ps_read_process: did not get vmem data for pid %i",pid);
 		}
 		if (ps->num_lwp <= 0)
@@ -1440,14 +1440,14 @@ static int ps_read_process(long pid, procstat_t *ps, char *state)
 	 * TODO: Data and code segment calculations for Solaris
 	 */
 
-	ps->vmem_data = -1;
-	ps->vmem_code = -1;
+	ps->vmem_data = 0;
+	ps->vmem_code = 0;
 	ps->stack_size = myStatus->pr_stksize;
 
 	/*
 	 * TODO: File descriptor count for Solaris
 	 */
-	ps->num_fd = -1;
+	ps->num_fd = 0;
 
 	/*
 	 * Calculating input/ouput chars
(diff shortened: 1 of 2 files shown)


icholy:
:+1:

faxm0dem:
:+1:

mfournier:
I just rebased this on current master + changed it to fetch the information the same way threads count is reported (cf 880246ed), which makes the plugin perform better for processes which use a large number of file descriptors.

A notable difference over the last version (c4211b3f87) is that all the memory-mapped files are now included in the file count. So now the report should be roughly equivalent to how many files `lsof -p <PID>` finds.

src/processes.c
```diff
@@ -736,6 +741,14 @@ static void ps_submit_proc_list (procstat_t *ps)
 	vl.values_len = 2;
 	plugin_dispatch_values (&vl);
 
+	if (ps->num_fd != -1)
```
octo: I think zero (`0`) is a better "not valid" value, because `-1` will likely cause warnings / errors with some compilers (always true due to range of variable, implicit cast at assignment, …). A process without *any* file descriptors is unable to communicate with the outside world and is therefore … rare.

octo:
LGTM, but see inline comment. Thanks @mfournier!

mfournier:
Rebased on master to fix a merge conflict + changed what @octo suggested.

manuelluis:
The patch is using ``FDSize`` from ``/proc/<pid>/status`` to count the open files of a process.

From: ``man 5 proc``: 

    FDSize: Number of file descriptor slots currently allocated.

It's not the open files per process, it's the size of file descriptors table of the process

For example, ``bash`` has 5 open files but ``FDSize`` is 256.

    $ cat /proc/$$/status | grep FDSize
    FDSize:	256
    $  ls -ln /proc/$$/fd
    total 0
    lrwx------ 1 100001 100001 64 Aug  3 19:39 0 -> /dev/pts/1
    lrwx------ 1 100001 100001 64 Aug  3 19:39 1 -> /dev/pts/1
    lrwx------ 1 100001 100001 64 Aug  3 18:04 2 -> /dev/pts/1
    lrwx------ 1 100001 100001 64 Aug  3 19:39 255 -> /dev/pts/1
    lr-x------ 1 100001 100001 64 Aug  3 19:39 3 -> /var/lib/sss/mc/passwd

octo:
Thanks @manuelluis, that makes sense.

@mfournier I don't think the FD set size is all that interesting. To get a meaningful count we would have to walk over `/proc/${pid}/fd/`. Is there an actual need for this metric or was it just conveniently available in `/proc/${pid}/status`?

mfournier:
You're right @manuelluis, thanks !

The initial implementation (c4211b3f) used to walk through `/proc/<pid>/fd/` but I switched to `FDSize` from `/proc/<pid>/status` based on https://github.com/collectd/collectd/pull/1041#issuecomment-104761272.

It doesn't seem the information we want is exposed anywhere in `/proc/` unfortunately. So I guess the way to go is to move back to the initial implementation, and add a configuration switch for this metric, which would be disabled by default.

What do you think @octo and @manuelluis ?

octo:
+1, SGTM

manuelluis:
+1

rpv-tomsk:
Proposal:

Extend `Process` /  `ProcessMatch` to options group and move `CollectContextSwitch`/ `CollectFileDescriptor` inside to make all these configurable per process/group.

rpv-tomsk:
I'm going to reimplement  this (by using code from commits, listed in #836 / #1041 and especially c4211b3f87bb5981e7d80c94ea6170fe9d4ecb51) with adding new config style: allow child elements in `Process` / `ProcessMatch`, so that will be something like:

```
<Plugin processes>
  CollectContextSwitch true|false
  CollectFileDescriptor true|false
  <Process "name">
    CollectContextSwitch  true|false
    CollectFileDescriptor true|false
  </Process>
</Plugin>
```
Any suggestions/comments from original authors?

octo:
@rpv-tomsk thanks Pavel, the config syntax you propose looks good to me.

PR #838, opened 2014-12-03
statsd histogram to support more than 1 second

yogeswaran:
Summary: statsd histogram to support more than 1 second
Problem: Collectd/Statsd supports configurable percentiles for timers but it limits the value to be 1 to 1000ms. If the timer value is more than 1000ms, it uses for min, max, average,... and drops it for percentile computation.

Solution: Added support for increasing bin width when the value is above histogram's range.

Test Plan: Tested by sending metrics within range and out-of-range (ie above 1000 ms)

Differential Revision: https://all.phab.dev.bloomberg.com/D156454

### DIFF
--- a/src/utils_latency.c
+++ b/src/utils_latency.c
@@ -25,12 +25,17 @@
  **/
 
 #include "collectd.h"
+#include "plugin.h"
 #include "utils_latency.h"
 #include "common.h"
 
-#ifndef LATENCY_HISTOGRAM_SIZE
-# define LATENCY_HISTOGRAM_SIZE 1000
+#include <math.h>
+
+#ifndef HISTOGRAM_NUM_BINS
+# define HISTOGRAM_NUM_BINS 1000
 #endif
+
+static const int HISTOGRAM_DEFAULT_BIN_WIDTH = 1;
 
 struct latency_counter_s
 {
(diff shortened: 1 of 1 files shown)


mre:
This would be really helpful for us as well. We use this for real user monitoring and the values regularly exceed 1000ms. As a result, the metrics are invalid. Would be cool to merge this in. :+1:

xo4n:
+1 for the merge

mfournier:
So I gave this patch a try. Here are my findings:

- statsd-tg is a bit flawed in a sense that it generates random timer values within the 0-1024ms range. So assuming @octo used it to help develop the statsd plugin, it could explain why this problem went unnoticed. For the testing I did, I modified statsd-tg to submit values in the 0-1048576ms range.

- I noticed collectd is able to absorb 8-10% less timer events with this patch than without. The ratio stays the same when the amount of different timers submitted is increased/decreased (`statsd-tg -t`).

- here's the massif output running `statsd-tg -c 0 -g 0 -s 0 -t 1000` for 60s with the patch (collectd's RSS peaks at about 14 MB when running outside of valgrind):
```
    MB
12.56^                                                                     #  
     |                                                               @     #  
     |                  @                      :                     @:    #  
     |                 @@:          @:         ::          :         @::  :#: 
     |                 @@:          @@        :::         :::        @::  :#::
     |                 @@::        :@@:       ::::        :::@       @::: :#::
     |         @@@@@@@@@@:::::::::::@@@@@@@@@@:::::::::::::::@:::::::@:::::#:@
     |       :@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |       :@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |       :@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |       :@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |      ::@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |      ::@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |      ::@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |:::::@::@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |:    @::@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |:    @::@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |:    @::@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |:    @::@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
     |:    @::@@       @@:::      ::@@@       :::::      ::::@      :@::: :#:@
   0 +----------------------------------------------------------------------->Gi
     0                                                                   5.290
```

- and without (RSS around 12.5 MB):
```
    MB
12.50^                                                                    #   
     |                     @                                   :          #   
     |                     @           :           :           :          #:  
     |                     @:          ::          :           ::         #:  
     |                    :@::        ::::        ::::         :::        #:@ 
     |                    :@::        ::::        ::::::::::::::::        #:@:
     |            :::::::::@::::::::::::::::::::::::::        ::::::::::::#:@@
     |           :::      :@:::       :::::       ::::        :::::      :#:@@
     |          @:::      :@:::       :::::       ::::        :::::      :#:@@
     |          @:::      :@:::       :::::       ::::        :::::      :#:@@
     |          @:::      :@:::       :::::       ::::        :::::      :#:@@
     |          @:::      :@:::       :::::       ::::        :::::      :#:@@
     |          @:::      :@:::       :::::       ::::        :::::      :#:@@
     |          @:::      :@:::       :::::       ::::        :::::      :#:@@
     |::::::::::@:::      :@:::       :::::       ::::        :::::      :#:@@
     |:        :@:::      :@:::       :::::       ::::        :::::      :#:@@
     |:        :@:::      :@:::       :::::       ::::        :::::      :#:@@
     |:        :@:::      :@:::       :::::       ::::        :::::      :#:@@
     |:        :@:::      :@:::       :::::       ::::        :::::      :#:@@
     |:        :@:::      :@:::       :::::       ::::        :::::      :#:@@
   0 +----------------------------------------------------------------------->Gi
     0                                                                   5.086
```

- on the other hand, submitting a larger number of timers simultaneously allocates an order of magnitude more memory. `statsd-tg -t 100000` will fill up 1GB of memory in a few seconds, both with and without this patch. So I'm not sure if it's really worth to bother keeping the per-bucket memory usage "under control" (which is one of the arguments in favor of #637) while the unbounded number of buckets potentially allows almost immediate memory exhaustion.

All in all, I +1 this patch as it fixes a major usability problem, without any change from the user point of view.
I just wonder if it shouldn't also to the collectd-5.4 branch.

pyr:
A huge thank you to @mfournier for this thorough testing. Thanks @yogeswaran !

mfournier:
I also cherry-pick this patch to the 5.4 branch, as there is no visible changes for the user.

PR #840, opened 2014-12-03
Fix ifdefs around multiline strings and define for AIX machines

yogeswaran:


rubenk:
@yogeswaran what compiler were you using? These errors don't look like output from GCC.

rubenk:
Closing stale issue.

PR #847, opened 2014-12-09
bind-plugin: fix zones iteratation upper limit

mihu:
Seems like it was created due to the copy-paste. Similar loop and check is in the `bind_xml_stats_handle_view`


pyr:
thanks @mihu !

PR #850, opened 2014-12-10
mysql wsrep stats

XANi:
I've added option to gather some stats specific to MySQL Galera replication setup ( more details here http://www.percona.com/doc/percona-xtradb-cluster/5.5/wsrep-status-index.html )

as per https://github.com/collectd/collectd/pull/641 I used generic datatypes and added one, window_size as there wasn't any existing one that would fit. Also rebased on current master instead of latest stable

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -2855,6 +2855,12 @@ Memory subsystem rail (uVolts).
 
 =back
 
+=item B<WsrepStats> I<true|false>
+
+Enable the collection of wsrep plugin statistics, used in Master-Master
+replication setups like in MySQL Galera/Percona XtraDB Cluster.
+User needs only privileges to execute 'SHOW GLOBAL STATUS'
+
 =back
 
 =head2 Plugin C<memory>
--- a/src/mysql.c
+++ b/src/mysql.c
@@ -53,6 +53,7 @@ struct mysql_database_s /* {{{ */
 	_Bool master_stats;
 	_Bool slave_stats;
 	_Bool innodb_stats;
+	_Bool wsrep_stats;
 
 	_Bool slave_notif;
 	_Bool slave_io_running;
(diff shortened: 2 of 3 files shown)


XANi:
any chance of merging this ? or should I change something ?

pyr:
I'm lagging behind on PRs but will try to act quick, sorry!

src/collectd.conf.pod
```diff
@@ -2855,6 +2855,12 @@ Memory subsystem rail (uVolts).
 
 =back
 
+=item B<WsrepStats> I<true|false>
```
mfournier: I assume placing this between the mic and memory plugin sections is a mistake. Can you correct this please @XANi ?

Also, there's no obvious way to know if the option is enabled or disabled by default. Can you amend the doc to mention this detail ?

I realise none of the other boolean options to this plugin document this either. I'll add this to the bugfix branches.

src/mysql.c
```diff
@@ -691,6 +694,144 @@ static int mysql_read_innodb_stats (mysql_database_t *db, MYSQL *con)
...
+		else if (strncmp (key, "wsrep_local_send_queue", key_length ) == 0 )
+			wsrep_local_send_queue = (gauge_t) val;
+		else if (strncmp (key, "wsrep_local_send_queue_avg", key_length ) == 0 )
+			wsrep_local_send_queue_avg = (gauge_t) val;
+		else if (strncmp (key, "wsrep_commit_window", key_length ) == 0 )
+			wsrep_commit_window = (gauge_t) val;
+	}
+
+	mysql_free_result (res);
+
```
mfournier: I might be mistaken, but wouldn't the code below submit empty values in the case one or the other of the items matched above aren't found ?

Wouldn't it be an idea to move these into `*_submit()` calls into the if/else clauses above, and avoid these numerous `wsrepo_*` variables in the process ?

It would make this consitent with how `mysql_read()` works and behaves at least, wouldn't it ?

mfournier:
@XANi, could you please also add a `WsrepStats` example line to `src/collectd.conf.in` ?

Issue #867, opened 2014-12-16
Segfaults when using absolute_t with plugin_dispatch_multivalue


ifesdjeen:
Hi,

I'm using `absolute_t` with `asbolute` type with `plugin_dispatch_multivalue`: 

```
asbolute_t ram_value         = (asbolute_t) memory.ram;

plugin_dispatch_multivalue (&vl, 0,
                            "ram"         , ram_value,
                            ...
                            NULL);          
```

Although that always results into segfault. 
The minute I change `asbolute_t` to `gauge_t` everything works smoothly. Although that's not a gauge, that's an absolute value. 

Any ideas what could possibly cause that?

mfournier:
Thanks a lot @ifesdjeen ! Closing as your patch got merged the other day.

ifesdjeen:
Ah right! Totally forgot thanks @mfournier !

PR #870, opened 2014-12-17
Change "plugin_dispatch_multivalue" to accept any metric type.

ifesdjeen:
Currently, `plugin_dispatch_multivalue` works only with 
`gauge_t` metric type. 

Although this function is extremely useful for other other data types.

This commit changes it to accept a "store_type" (one of `DS_TYPE_{GAUGE|COUTNTER|ABSOLUTE|DERIVE}`).

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -2162,7 +2162,7 @@ int plugin_dispatch_values (value_list_t const *vl)
 
 __attribute__((sentinel))
 int plugin_dispatch_multivalue (value_list_t const *template, /* {{{ */
-		_Bool store_percentage, ...)
+		_Bool store_percentage, int store_type, ...)
 {
 	value_list_t *vl;
 	int failed = 0;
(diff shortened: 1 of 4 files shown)


ifesdjeen:
Hi everyone! Could you take a look at the PR? cc @mfournier @pyr

mfournier:
@ifesdjeen, thanks for this contribution! Sorry for the delay, I guess we're all still on holidays ;-)

The patch looks good to me, the use case makes sense, so +1 from me, but I'd like to also have someone else take a look as plugin.c is a pretty central file.

pyr:
Hi everyone,

Even though the code is good, this is unnecessary since values are enums containing types with the same storage size and alignment. So even though values will be stored using `vl->values[0].gauge` they can still be accessed as counters or derives later on.

Maybe to make this fact a bit more clear, the assignment could switch to a `memcpy` (gauge would still be accessed when `store_percentages` is true though).

ifesdjeen:
Thanks for response! 
I'm not sure though why I was getting segfaults if the value was absolute then :/ (see #867)

pyr:
Maybe because my logic is flawed... I'll test it.

pyr:
Hi @ifesdjeen, 

I can indeed reproduce the behavior, my understanding of `va_arg` behavior was wrong.
As it stands, I'm :+1: on this.

ifesdjeen:
:+1: <3 @pyr

ifesdjeen:
Thanks a lot @pyr !
<3

PR #873, opened 2014-12-21
Support reading per-connection TCP stats (RTT) from Linux kernel tcp_diag

arielshaqed:
The Linux tcp_diag module supports reading additional fields from the host TCP stacks. This patch adds the ability to read the per-connection ("smoothed") RTT value from the reported tcp_info struct.

RTT is typically read at short intervals. This patch also batches value_list reporting, to allow a future batched reporting API.

### DIFF
--- a/src/tcpconns.c
+++ b/src/tcpconns.c
@@ -275,12 +275,16 @@ static const char *config_keys[] =
   "ListeningPorts",
   "LocalPort",
   "RemotePort",
-  "AllPortsSummary"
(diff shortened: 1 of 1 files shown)


arielshaqed:
Added another change: (configurably) ignore unchanged connections.

arielshaqed:
Closing: Missing collectd.conf docs, want to squash commits together, etc.
Will open a new pull request soon.

PR #876, opened 2014-12-27
StatsD write plugin

stefano-pogliani:
A write plugin to allow CollectD to send values to a StatD server.

Please note that, in this implementation, `DS_TYPE_DERIVE` is forwarded as a gauge value and `DS_TYPE_ABSOLUTE` is not supported.
Any suggestions about better type mappings would be much appreciated, as would be any other comment or suggestion in general.


stefano-pogliani:
Hello everyone,

The latest CI run for this PR (https://ci.collectd.org/view/tests/job/pull-requests-github_trigger-aggregation/112/) due to the CentOS5 build step (https://ci.collectd.org/job/pull-requests-build-on-epel5-x86_64-with-default-toolchain/240/).

I was unable to find the error in the log so I don't know how I could fix it.
I would be grateful if anyone could point me in the right direction.

Cheers,
Stefano

mfournier:
Yes, sorry the output of these builds is not easy to grok, I have to improve this!

Here is the error which is making the build fail:

```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I./daemon -DPREFIX=\"/opt/collectd\" -DCONFIGFILE=\"/opt/collectd/etc/collectd.conf\" -DLOCALSTATEDIR=\"/opt/collectd/var\" -DPKGLOCALSTATEDIR=\"/opt/collectd/var/lib/collectd\" -DPIDFILE=\"/opt/collectd/var/run/collectd.pid\" -DPLUGINDIR=\"/opt/collectd/lib/collectd\" -DPKGDATADIR=\"/opt/collectd/share/collectd\" "-DLT_LAZY_OR_NOW=RTLD_LAZY|RTLD_GLOBAL" -Wall -Werror -O2 -g -m64 -mtune=generic -MT write_statsd_la-write_statsd.lo -MD -MP -MF .deps/write_statsd_la-write_statsd.Tpo -c write_statsd.c  -fPIC -DPIC -o .libs/write_statsd_la-write_statsd.o
write_statsd.c: In function 'write_statsd_config':
write_statsd.c:180: warning: comparison is always false due to limited range of data type
write_statsd.c:181: warning: comparison is always false due to limited range of data type
make[3]: *** [write_statsd_la-write_statsd.lo] Error 1
```

That said, I find puzzling that only centos5 *64bit* yields this error. FWIW, this distro comes with gcc 4.1.2 as default compiler.

stefano-pogliani:
I was unable to reproduce the error on my local machine [gcc (GCC) 4.8.3 20140911 (Red Hat 4.8.3-7) - 64bits].

I believe that the issue was caused by `write_statsd_config_s.port` being defined as `int` instead of `long int`, which is the type returned by `strtol`.
I've changed that now.

As for why only CentOS 5 seems to report the issue I could only find https://gcc.gnu.org/bugzilla/show_bug.cgi?id=12963 that could explain why newer versions of GCC do not complain.

dch:
hi @stefano-pogliani sorry for the delay!  Nice work :-) are you able to rebase this PR on master, squash it down to a single commit, & repush it so we can take a look?

dch:
@stefano-pogliani I got what looks like a clean squashed commit now https://github.com/skunkwerks/collectd/tree/feature/write-statsd if you want to pull & check it? I assume you've been using this for a while now, how has it been going?

dch:
on debian it seems the last change introduced breaks things:

```
write_statsd.c: In function 'write_statsd_init':
write_statsd.c:278:3: error: format '%i' expects argument of type 'int', but argument 4 has type 'long int' [-Werror=format=]
   DEBUG("%s Port: %i", WRITE_STATSD_NAME, configuration.port);
```

ideas on how to address this?

mfournier:
@dch, I guess you want `%li` here.

For the background, Jenkins enforces the gcc flags used when building the Debian and Fedora packages. The goal is to catch this sort of issues as early as possible, to release the burden this typically puts on distro package maintainers.

So if you use the same options you find in the Jenkins build logs, on your Mac, you should be able to trigger the same errors as the Jenkins would. HTH :-)

stefano-pogliani:
Hi all, sorry I've been unavailable.
I'll look at the error now and carefully look at the output for warnings as well as errors.

@dch, thank you for opening the squashed pull request.
Should I push the fix to this branch or the one you created, to this or a new squashed one?

@mfournier thank you for the information and the correction :smile:.

dch:
@stefano-pogliani you should add it to this PR, it's your work after all!

dch:
On Fri, 9 Oct 2015, at 12:42 PM, Stefano Pogliani wrote:
> Hi all, sorry I've been unavailable.
> I'll look at the error now and carefully look at the output for warnings
> as well as errors.
> 
> @dch, thank you for opening the squashed pull request.
> Should I push the fix to this branch or the one you created, to this or a
> new squashed one?
> 
> @mfournier thank you for the information and the correction :smile:.
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/collectd/collectd/pull/876#issuecomment-146831161

LGTM so far; Stefano can you pull in the tweaks from my
skunkwerks/write_statsd fork into yours? The only changes are:

- ensuring alphabetical ordering in a bunch of places
- a couple of spelling errors that snuck in

I think we are ready to merge, would be good to get a +1 from some
others.

Thanks Stefano for your contribution!

mfournier:
Thanks a lot @stefano-pogliani ! Apart from being an useful addition, the code is simple and easy to follow, great job ! Also thanks for taking care of this @dch !

I'm not sure I understand the generated metrics naming logic. From what I can see, only a subset of what uniquely identifies metrics in collectd is used to compose the string submitted to the server. This might work for some read plugins, but leads to duplicates with others.

Example:
```
[2015-10-13 21:42:42] [debug] plugin_dispatch_values: time = 1444765361.860; interval = 1.000; host = lonquimay; plugin = memory; plugin_instance = ; type = memory; type_instance = buffered;
[2015-10-13 21:42:42] [debug] plugin_dispatch_values: time = 1444765361.860; interval = 1.000; host = lonquimay; plugin = memory; plugin_instance = ; type = memory; type_instance = used;
[2015-10-13 21:42:42] [debug] plugin_dispatch_values: time = 1444765361.860; interval = 1.000; host = lonquimay; plugin = memory; plugin_instance = ; type = memory; type_instance = cached;
[2015-10-13 21:42:42] [debug] plugin_dispatch_values: time = 1444765361.860; interval = 1.000; host = lonquimay; plugin = memory; plugin_instance = ; type = memory; type_instance = free;
[2015-10-13 21:42:42] [debug] plugin_dispatch_values: time = 1444765361.860; interval = 1.000; host = lonquimay; plugin = memory; plugin_instance = ; type = memory; type_instance = slab_unrecl;
[2015-10-13 21:42:42] [debug] plugin_dispatch_values: time = 1444765361.860; interval = 1.000; host = lonquimay; plugin = memory; plugin_instance = ; type = memory; type_instance = slab_recl;
```
results in the following data getting sent over the wire:
```
U 127.0.0.1:40411 -> 127.0.0.1:8125
  memory.value:325328896.000000|g                                                                     
U 127.0.0.1:34260 -> 127.0.0.1:8125
  memory.value:1431060480.000000|g                                                                    
U 127.0.0.1:44934 -> 127.0.0.1:8125
  memory.value:292970496.000000|g                                                                     
U 127.0.0.1:40286 -> 127.0.0.1:8125
  memory.value:53112832.000000|g                                                                      
U 127.0.0.1:55450 -> 127.0.0.1:8125
  memory.value:2554638336.000000|g                                                                    
U 127.0.0.1:58875 -> 127.0.0.1:8125
  memory.value:3505385472.000000|g
```

I would have rather imagined `lonquimay.memory.memory.buffered`, etc, or something along these lines.

You might want to take a look at https://collectd.org/wiki/index.php/Naming_schema to make sure all the different cases are covered.

This would be the main blocker from my perspective.

A couple of other details I noticed:
* shouldn't you check the return status of the `strdup()` statements ? As an alternative, maybe replace them with calls to the `cf_util_get_*` functions ?
* IPv6 doesn't seem to work: `write_statsd plugin: unable to resolve addres - Address family for hostname not supported`. Would you mind fixing this ?
* can you please add a couple of lines of doc to the manpage ?

Thanks !

stefano-pogliani:
Thanks @dch for the corrections!

Hi @mfournier, in regard to the names I misunderstood the format.
Based on the naming schema and your example I understand your example to be `hostname.plugin[.plugin_instance].type[.type_instance]`. Correct?

As for the other issues you listed I'll be looking at them over the weekend and update this PR.

mfournier:
Yes, this is correct.

One thing that also needs to be handled is types containing more than one value. The load type (used by the load plugin) for example always contains 3 numbers: shortterm, midterm and longterm. From what I can see, this case is already properly handled. So just to make sure this doesn't get broken in the refactor.

Thanks !

stefano-pogliani:
Hello again @mfournier,

I'm looking at the name now and, in addition to the fields in `value_list_t` we discussed above, there are also two other, possibly relevant, fields from `data_set_t` and `data_source_t`:
  * `data_source_t.name` is the name of the value in a multi-valued data set.
  * `data_set_t.type` is this the same as `value_list_t.type`? If not could you explain the difference.

stefano-pogliani:
I've changed the way names are generated (as discussed above).
I've omitted the `data_set_t.type` for now but I can add it if needed for unique names, which results in the following:

    localhost.load.load.shortterm: 1.1
    localhost.load.load.midterm: 1.11
    localhost.load.load.longterm: 1.03
    localhost.memory.memory.buffered.value: 299933696
    localhost.memory.memory.slab_unrecl.value: 97218560
    localhost.memory.memory.free.value: 940601344
    localhost.memory.memory.slab_recl.value: 289783808
    localhost.memory.memory.cached.value: 1491329024
    localhost.memory.memory.used.value: 5063704576

stefano-pogliani:
Hi @mfournier, I've checked Jenkins to fix the broken build but it is marked as aborted by you.
Is this correct? Did I miss the error?

mfournier:
Hey @stefano-pogliani, sorry about that. I screwed up a jenkins update and wasn't able to allocate enough time to fix things properly yet :-( I'll trigger the job again once it's working again.

Other than that, thanks for the updates ! It all looks great at first sight ! I should be able to take a moment to test this out tomorrow. @dch can you give this a try too by any chance ? Thanks !

README
```diff
@@ -462,6 +462,9 @@ Features
     - write_riemann
       Sends data to Riemann, a stream processing and monitoring system.
 
+    - write_statsd
+      Sends data to a StatsD server.
+
     - write_sensu
```
dch: This section should come after  `write_sensu` for Great Ordered Alphabetness.

AUTHORS
```diff
@@ -267,6 +267,9 @@ Simon Kuhnle <simon at blarzwurst.de>
 Sjoerd van der Berg <harekiet at gmail.com>
  - iptables plugin.
 
+Stefano Pogliani <stefano at spogliani.net>
+ - StatsD write plugin.
+
 Stefan Hacker <stefan.hacker at web.de>
  - teamspeak2 plugin.
```
dch: Stefan comes before Stefano!

dch:
This is working brilliantly down here on OSX, I'll check the memory over the next few hours but metrics are spewing out into statsd. super!

dch:
I'm seeing a slow memory leak overnight. This may have nothing to do with the plugin, but I'm up top ~392Mb now, from a base of under 5 Mb at start time. Does anybody else see the same thing?

mfournier:
Yes, I noticed the leak too, using valgrind. I didn't dig very far but it wasn't obvious to me what was leaking. What I can tell is that the problem was introduced in 7b2c952650b.

Here's my valgrind output:
```
==1008== 
==1008== HEAP SUMMARY:
==1008==     in use at exit: 191,478 bytes in 1,674 blocks
==1008==   total heap usage: 30,113 allocs, 28,439 frees, 9,210,329 bytes allocated
==1008== 
==1008== 11,025 bytes in 185 blocks are definitely lost in loss record 46 of 50
==1008==    at 0x4C28C4F: malloc (vg_replace_malloc.c:299)
==1008==    by 0x7FF1F48: allocate (write_statsd.c:93)
==1008==    by 0x7FF1F48: write_statsd_write (write_statsd.c:337)
==1008==    by 0x40C5A1: plugin_write (plugin.c:1884)
==1008==    by 0x408A7C: fc_bit_write_invoke (filter_chain.c:724)
==1008==    by 0x40BCCC: plugin_dispatch_values_internal (plugin.c:2223)
==1008==    by 0x40D89E: plugin_write_thread (plugin.c:844)
==1008==    by 0x513E0A3: start_thread (pthread_create.c:309)
==1008== 
==1008== 37,632 bytes in 549 blocks are definitely lost in loss record 49 of 50
==1008==    at 0x4C28C4F: malloc (vg_replace_malloc.c:299)
==1008==    by 0x7FF1EEF: allocate (write_statsd.c:93)
==1008==    by 0x7FF1EEF: write_statsd_write (write_statsd.c:337)
==1008==    by 0x40C5A1: plugin_write (plugin.c:1884)
==1008==    by 0x408A7C: fc_bit_write_invoke (filter_chain.c:724)
==1008==    by 0x40BCCC: plugin_dispatch_values_internal (plugin.c:2223)
==1008==    by 0x40D89E: plugin_write_thread (plugin.c:844)
==1008==    by 0x513E0A3: start_thread (pthread_create.c:309)
==1008== 
==1008== LEAK SUMMARY:
==1008==    definitely lost: 48,657 bytes in 734 blocks
==1008==    indirectly lost: 0 bytes in 0 blocks
==1008==      possibly lost: 0 bytes in 0 blocks
==1008==    still reachable: 142,821 bytes in 940 blocks
==1008==         suppressed: 0 bytes in 0 blocks
==1008== Reachable blocks (those to which a pointer was found) are not shown.
==1008== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==1008== 
==1008== For counts of detected and suppressed errors, rerun with: -v
==1008== ERROR SUMMARY: 2 errors from 2 contexts (suppressed: 0 from 0)
```

Other than that, here are my comments from what I tested yesterday:

I think it would be wise to sanitize the various fields from the value_list which are used to build the metric names. Not sure exactly what characters a typical statsd server would choke on, but for example the `host` part would be full of dots when the FQDN is used. This might mess up things downstream from statsd. Also plugins such as snmp or ipmi sometimes return awkward caracters in the `type_instance` or `plugin_instance` fields. Not sure if this can be reused directly, but the write_graphite plugin does this sort of thing. Check out `GRAPHITE_FORBIDDEN` from `src/utils_format_graphite.c`. See also https://collectd.org/wiki/index.php/Naming_schema.

I expect 2 options users will soon ask for are `StoreRates` and `AlwaysAppendDS`. Almost all write_* plugins implement these. I wrote the `AlwaysAppendDS` part in https://github.com/mfournier/collectd/commit/f076049ab30. Probably not the smartest way to do this, so up to you to decide if you want to reuse the patch or not.

Thanks for all your work on this ! It's getting very nice :-)

stefano-pogliani:
It seems I only free `value_key` on error but not on success. Thanks for spotting that!

As far as the name is concerned a quick search seems to suggest that the key can be anything except for `:` which are used as a separator in the protocol.
At any rate I'll look more into this and think how to sanitize names without distorting them.

As for the additional options you mentioned I will defiantly look at them :smile:

stefano-pogliani:
After digging around more the only reference to the meaning of `.` in statsd names is in this StackOverflow answer http://stackoverflow.com/a/18131221 and in the graphite documentation that speaks about `metric path`s (http://graphite.readthedocs.org/en/latest/feeding-carbon.html#the-plaintext-protocol)

After testing out different names, the following characters are treated specially:
  * The `.` represents a hierarchical division (similar to '/' in paths).
  * The `:` in the name cause errors as they are divisors for the statsd protocol.
  * The `|` in the name does not error in Etsy's statsd server but it is dropped from the name so I wonder what other systems would do.
  * Whitespaces are collapsed into an `_`.

I therefore intend to escape all of the above to avoid issues but I am wondering about the following two points:
  * Should I generalise the graphite escape function to take the list of invalid characters from a `const char*` or should I copy the code to my own module/utils module?
   If I generalise the function should I also change the one in `write_graphite` or should I leave that alone?
  * Since `.` is used as a separator I would escape it in the fields that form the name but I'm tempted to leave it as is for host names.
  This would allow to reference hosts or groups of hosts when building metrics and works well with the fact that DNS is a hierarchical name itself.
  What do you guys think about this?

mfournier:
Hi, @stefano-pogliani !

I just updated my branch over at https://github.com/mfournier/collectd/tree/write_statsd with the `StoreRates` feature (enabled by default, debatable) + a couple of minor fixes. Would you mind taking a look and merging them into your branch if you agree ?

Also I've been wondering about the decision to map collectd `COUNTER` types to statsd counters. Despite the identical names, it seems they don't really have the same characteristics. Or at least collectd `DERIVE` types should map to the same statsd type as `COUNTER`s do, as they behave the same (you might want to check out https://collectd.org/wiki/index.php/Data_source).
Short example: let's say collectd reads a value which increases of 1 per read-interval (1, 2, 3, ...), after 3 intervals statsd will have sumed up 1 + 2 + 3. How to make sense of such a number ? The only case where it seems right to me is for the `ABSOLUTE` collectd types (hence 215e98f6). @dch, any thought about this ? How is the datadog integration working for you so far ?

As for your question from last week (sorry for not answering earlier!), yes feel free to make `escape_graphite_string()` more generic ! Maybe even rename & move it to `src/daemon/utils_subst.[ch]` for example ?
Regarding escaping `.`, I'd favor escaping them for the hostname part too. My reasoning being that for the DNS hierarchy thing to work, the FQDNs would have to be reversed first (ie: `org.collectd.www`).

stefano-pogliani:
Hi @mfournier,

Thank you for the changes in your branch.
I've already looked at it and I'll merge it in when I have time.

As far as the escaping is concerned I'll generalise `escape_graphite_string` by adding the set of characters to escape as an argument (thanks for the hint on where to place it).
Also I see your point about the host name and it should probably be escaped too.

Finally I am looking at the types in collectd (https://collectd.org/wiki/index.php/Data_source) versus the types in Etsy's statsd (https://github.com/etsy/statsd/blob/master/docs/metric_types.md):

  * **Gauges**
    They appear similar but have one key exception that makes them incompatible with each other: their implementation of statsd uses sign in front of gauges to make them relative to the last known value.
    So if the value is 5 and the next value is -3 collectd would think of the gauge as -3 while statsd would think the value is 2.
  * **Derives**
    My understanding is that this type is used to tell collectd to manipulate the reported value into a derivative and then output that result as a value. If that is the case derives map to statsd gauges as well.
    Does this sound correct to you?
  * **Counters** Same as derives?
  * **Absolutes**
    These seems to be what statsd calls "sampling". If this is the case where would I get the time since the last read of the value that was used to compute the absolute?

Ignoring the problem of gauges in statsd supporting relative values (signed values are relative to the last known value of the gauge) do these mappings make more sense?

I still am not sure about how to solve the discrepancy in definition of gauge between collectd and statsd but if I can access the previous value of a gauge in collectd I may be able to send a "reset to zero" metric to statsd and then sent it the actual value before any negative gauge is sent.
What do you think?

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+#define _BSD_SOURCE
+
+#include <limits.h>
+#include <stdlib.h>
+
+#include <netdb.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+
+#include "collectd.h"
```
octo: Please include `"collectd.h"` as first header. This will pull in all other headers except `<netdb.h>`.

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+#include "common.h"
+#include "plugin.h"
+
+
+/*** Compile-time constants. ***/
+#define WS_DOUBLE_FORMAT  "%f"
```
octo: Please use `GAUGE_FORMAT` from `collectd.h`.

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+#define VALUE_STR_LEN     20
+#define WRITE_STATSD_NAME "write_statsd"
+
+#define FREE_NOT_NULL(var) \
```
octo: Please use the `sfree()` macro from `common.h`.

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+  "g",  /* DS_TYPE_GAUGE    */
+  "g",  /* DS_TYPE_DERIVE   */
+  NULL  /* DS_TYPE_ABSOLUTE */
+};
+
+/*
+ * The configuration is stored in a global instance and must not
+ * be modified outside of load_config or initialise.
+ */
+static write_statsd_config_t configuration = {
```
octo: Please consider passing a `write_statsd_link_t*` to the read callback as the user_data pointer. This way we can easily support writing to multiple statsd instances in parallel.

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+  NULL, 8125, NULL, NULL, 0, 0,
+  NULL, 0, NULL, 0
+};
+
+
+/*** Module functions. ***/
+static void* allocate(size_t size) {
```
octo: Please use `maloc()` instead.

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+  return result;
+}
+
+
+static write_statsd_link_t open_socket(void) {
+  write_statsd_link_t link  = {{0}, 0};
+  struct addrinfo* resolved = NULL;
+  int error = 0;
+
+  link.sock = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
```
octo: Please rewrite this in a protocol independent manner. Look for `getaddrinfo()` in `src/zookeeper.c` for an example of opening a client connection.

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+
+  for (idx = 0; idx < conf->children_num; idx++) {
+    oconfig_item_t *child = conf->children + idx;
+    status = 0;
+
+    if (strcasecmp("Host", child->key) == 0) {
+      status = cf_util_get_string(child, &configuration.host);
+
+    } else if (strcasecmp("Port", child->key) == 0) {
+      status = cf_util_get_int(child, & configuration.port);
```
octo: Please use `cf_util_get_service()`. The returned string should be passed to `getaddrinfo()` (see above).

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+   */
+  if (configuration.prefix == NULL && configuration.postfix == NULL) {
+    /* %s.%s:%s|%s && %s:%s|%s */
+    configuration.event_line_format = strdup("%s.%s:%s|%s");
+    configuration.short_event_line_format = strdup("%s:%s|%s");
+    len = 3;
+
+  } else if (configuration.prefix == NULL) {
+    /* %s.%s.<postfix>:%s|%s && %s.<postfix>:%s|%s */
+    len = 4 + strlen(configuration.postfix);
```
octo: This is way too error prone. Please do the line formatting when writing metrics, ideally in a separate function, and don't store the format in a variable. The function could look something like this:

```c
int format_line (char *buffer, size_t buffer_size, int ds_type, value_t v)
{
  /* format v */
  if ((configuration.prefix != NULL) && (configuration.postfix != NULL))
    ssnprintf (buffer, buffer_size, "%s.%s.%s.%s:%s|%s",
        prefix, foo, bar, postfix, qux, v_str);
  else if (configuration.prefix != NULL)
    ssnprintf (buffer, buffer_size, "%s.%s.%s:%s|%s",
        prefix, foo, bar, qux, v_str);
  else
    /* … */
}
```

Yes, it's repetitive, but it's easy to read – maybe even boring – and readers don't have to wonder what the "base length" is supposed to be.

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+  /*
+   * Build the type name using all available fields in vl:
+   * hostname.plugin[.plugin_instance].type[.type_instance]
+   */
+  char*  value_key = NULL;
+  size_t value_key_len = 2;  // Dots are always present.
+  size_t plugin_instance_len = strlen(vl->plugin_instance);
+  size_t type_instance_len = strlen(vl->type_instance);
+
+  value_key_len += strlen(vl->host);
```
octo: Each part of the name is at most `DATA_MAX_NAME_LEN` bytes long – just use a `char[6 * DATA_MAX_NAME_LEN]` on the stack.

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+
+  if (result == -1) {
+    ERROR("write_statsd plugin: unable to send message.");
+  };
+
+  close(link.sock);
+  return 0;
+}
+
+static int write_statsd_write(
```
octo: Please move all the line formatting out of this function. Essentially this function should boil down to:

```c
for (i = 0; i < vl->values_len; i++)
{
  char buffer[1024];

  format_line (buffer, sizeof (buffer), ds, vl, i);

  pthread_mutex_lock (&socket_lock);
  write_statsd_send_message (buffer);
  pthread_mutex_unlock (&socket_lock);
}
```

src/write_statsd.c
```diff
@@ -0,0 +1,476 @@
...
+
+    } else {
+      message_len = configuration.short_event_line_base_len;
+    }
+
+    message_len += value_key_len;
+    message_len += strlen(value);
+    message_len += strlen(ds_type);
+
+    message = allocate(message_len + 1);
```
octo: This byte counting needs to go. Use a large enough buffer, e.g. 1 kByte, and go with that. If you're *really* **absolutely** concerned about not wasting a byte of memory, call `snprintf()` twice: once to determine the required buffer size, and a second time with the correctly allocated buffer. Beware, however, that `snprintf()` is *extremely* slow and this will have a large performance impact on this plugin.

octo:
Thank you very much for your plugin, @stefano-pogliani! I've left some code review comments inline – I hope you can forgive my brevity.

I'm very concerned about the byte counting and the dynamically computed format strings. Both are *very* error prone and don't provide an obvious benefit. If there is a benefit I'm not seeing, then we need to add unit tests to *demonstrate* that everything behaves as intended. As it is, this is a buffer overflow waiting to be discovered.

Best regards,
—octo

stefano-pogliani:
I believe I took care of all the comments.
Please let me know if I missed something or you find anything else.
Also, thank you for your patient everyone :smile:

oryband:
Can we give this write plugin another go? Could be very useful for reporting to other backends other than graphite.

stefano-pogliani:
I can probably pull in the latest master and resolve conflicts in the next few days if there is interest in it.

If you are suggesting a full rewrite however it may have to wait a bit longer, I am in the middle of house buying in London (fun times) ...

oryband:
@stefano-pogliani this is definitely an interest. Collectd is a great metric collection agent, StatsD is a great, generic metric aggregator agent, and there are plenty of new metric backends besides graphite that I would like to flush into using StatsD as a proxy.

oryband:
Note that I'm currently doing this [using our own (Rounds) fork](https://github.com/rounds/collectd-statsd) of [collectd-statsd](https://github.com/lyft/collectd-statsd) python plugin:

```apache
<LoadPlugin python>
    Globals true
</LoadPlugin>
<Plugin python>
    # this is were i installed the statsd plugin
    ModulePath "/usr/local/lib/python2.7/site-packages/statsd_writer"
    LogTraces true
    Interactive false

    Import "statsd_writer"
    <Module "statsd_writer">
        host "localhost"
        port "8125"
        prefix ""
    </Module>
</Plugin>
```

sreeix:
Is it possible to get this merged? If it's about resolving the existing conficts, I can give it a go.
We use a slightly unusual setup and use the repeater statsd backend, and would be happy to use this plugin to write to the statsd server.

PR #878, opened 2015-01-02
add MemAvailable support to the memory plugin

skottler:
A new metric called 'MemAvailable' was added to `/proc/meminfo` in the 3.14 kernel, which is an "estimate of how much memory is available for starting new applications, without swapping. Calculated from MemFree, SReclaimable, the size of the file LRU lists, and the low watermarks in each zone."

This patch adds support for this metric to the memory plugin.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -282,12 +282,14 @@ static int memory_read_internal (value_list_t *vl)
 	int numfields;
 
 	_Bool detailed_slab_info = 0;
+	_Bool mem_available_info = 0;
 
 	gauge_t mem_total = 0;
 	gauge_t mem_used = 0;
 	gauge_t mem_buffered = 0;
 	gauge_t mem_cached = 0;
 	gauge_t mem_free = 0;
+	gauge_t mem_available = 0;
 	gauge_t mem_slab_total = 0;
 	gauge_t mem_slab_reclaimable = 0;
 	gauge_t mem_slab_unreclaimable = 0;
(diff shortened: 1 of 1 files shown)


aybabtme:
lgtm

src/memory.c
```diff
@@ -348,19 +354,32 @@ static int memory_read_internal (value_list_t *vl)
...
-	else
+	if (detailed_slab_info) {
+		/* The only condition under which there will be a MemAvailable line is
+		 * on a kernel which also supports detailed slab information. */
+		if (mem_available_info) {
+			MEMORY_SUBMIT ("used",          mem_used,
+			               "buffered",      mem_buffered,
+			               "cached",        mem_cached,
+			               "free",          mem_free,
+			               "available",     mem_available,
```
katzj: Is this accounted as separate from the rest or is it a part of the overall memory?  From your description, I think the former?  In which case, I think this needs to be done in a separate submit to avoid throwing things off if ValuesPercentage is set to true.

mfournier:
Yes, @katzj's remark would definetely be a blocker IMHO.

We must also keep in mind that this plugin is supposed to be more or less feature-equivalent across platforms. So `MemAvailable` would feel pretty linux-ish to BSD/MacOS/etc users (and would have no effect, possibly confusing them).

So what about having a more generic name such as `ExtendedStats`, which would enable collecting misc platform-specific data, and which would not get accounted for in the percentage calculation ?

pyr:
Hi @skottler ! Thanks for the patch. I agree with @katz's remark, but am otherwise looking forward to this patch, if you need pointers for the separate submit let me know here or on IRC

skottler:
Thanks for the feedback guys. I'll update the PR to submit `MemAvailable` separately later today.

nelg:
+1 for this change

PaulFurtado:
Any update on this? What's currently blocking getting this in?

octo:
Hi @PaulFurtado, this PR is currently pending the resolution of @katzj's review comment.

martinhynar:
This issue feels like dead after 3 years

Issue #862, opened 2014-12-15
Master doesn't compile on Mac OS X


ifesdjeen:
Currently, when trying to build Master on Mac OS X, I get: 

```
cpu.c:235:2: error: use of undeclared identifier 'cpu_temp_retry_max'
        cpu_temp_retry_max = 86400 / CDTIME_T_TO_TIME_T (plugin_get_interval ());
        ^
```

(actually I can't really find that defined anywhere, either)...

ryan-williams:
Hi, I ran into this one building `master` () on OSX. `git checkout origin/collectd-5.3 && ./build.sh && ./configure && ./make` did get a little bit further, but failed on something else:

```
In file included from aggregation.c:32:
./common.h:255:20: error: expected ')'
unsigned long long ntohll (unsigned long long n);
                   ^
/usr/include/sys/_endian.h:140:25: note: expanded from macro 'ntohll'
#define ntohll(x)       __DARWIN_OSSwapInt64(x)
                        ^
/usr/include/libkern/_OSByteOrder.h:78:30: note: expanded from macro '__DARWIN_OSSwapInt64'
    (__builtin_constant_p(x) ? __DARWIN_OSSwapConstInt64(x) : _OSSwapInt64(x))
                             ^
./common.h:255:20: note: to match this '('
/usr/include/sys/_endian.h:140:25: note: expanded from macro 'ntohll'
#define ntohll(x)       __DARWIN_OSSwapInt64(x)
                        ^
...
```

[Here is a gist of the full output](https://gist.github.com/ryan-williams/e931712fdb144381148c). Let me know if you have any ideas for fixing this?

mfournier:
@ryan-williams, this looks unrelated to the initial issue. Could you please open a new issue with the details?

Also, I'm surprised others didn't already hit this problem. Could you give some detail about your build environment? Does this only happen with the collectd-5.3 branch? Or are collectd-5.4 and master also affected?

ifesdjeen:
@ryan-williams just re-run configure. Autoconf already knows to to handle `ntohll`. That probably happened after migrating to Yosemite, although re-configuring and compiling from scratch should help (helped me at least).

`ntohll` is a widespread issue, almost every compiled project had same problem..

ryan-williams:
@ifesdjeen hm, I tried a few permutations of re-configuring / re-build.sh-ing / wiping out the workspace and starting over. Still seeing the same error:
```
network_buffer.c:154:17: error: expected ')'
[...]

PR #882, opened 2015-01-04
declare cpu_temp_retry_max in src/cpu.c to fix comp under clang (#862)

skottler:
This may be vaguely related to #862, but this makes master compile on 10.10 for me.

### DIFF
--- a/src/cpu.c
+++ b/src/cpu.c
@@ -218,7 +218,7 @@ static int init (void)
 {
 #if PROCESSOR_CPU_LOAD_INFO
 	kern_return_t status;
-	static int cpu_temp_retry_max;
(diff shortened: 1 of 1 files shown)


skottler:
Without this patch:

```
cpu.c:160:2: error: use of undeclared identifier 'cpu_temp_retry_max'
        cpu_temp_retry_max = 86400 / CDTIME_T_TO_TIME_T (plugin_get_interval ());
        ^
1 error generated.
make[3]: *** [cpu_la-cpu.lo] Error 1
make[1]: *** [all] Error 2
```

skottler:
This actually looks like line 160 in `cpu.c` was removed. Closing.

PR #886, opened 2015-01-08
Add a write_log output plugin which logs values.

pyr:
I find this to be useful when developping new input plugins,
it allows creating a very simple configuration and combined
with `-f` can be used for a very simple workflow.

I'm proposing this, since I think it could be useful for
people wanting to debug their config on local agents as well.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -7129,6 +7129,12 @@ want to set B<metadata.broker.list> to your Kafka broker list.
 
 =back
 
+=head2 Plugin C<write_log>
+
+The I<write_log plugin> will send values to the configured log destination.
+It is meant to be used for testing purposes, to ensure that values
+are correctly reported. Values will be formatted in the I<graphite> format.
+
 =head2 Plugin C<write_redis>
 
 The I<write_redis plugin> submits values to I<Redis>, a data structure server.
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -7129,12 +7129,6 @@ want to set B<metadata.broker.list> to your Kafka broker list.
 
 =back
 
-=head2 Plugin C<write_log>
-
-The I<write_log plugin> will send values to the configured log destination.
-It is meant to be used for testing purposes, to ensure that values
-are correctly reported. Values will be formatted in the I<graphite> format.
-
 =head2 Plugin C<write_redis>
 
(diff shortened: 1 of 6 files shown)


katzj:
:+1: I tend to use the csv plugin for this but this looks nicer

src/write_log.c
```diff
@@ -0,0 +1,76 @@
...
+        return -1;
+    }
+
+    memset (buffer, 0, sizeof (buffer));
+    status = format_graphite (buffer, sizeof (buffer), ds, vl,
+                              NULL, NULL, '_', 0);
+    if (status != 0) /* error message has been printed already. */
+        return (status);
+
+    INFO ("write_log values:\n%s", buffer);
```
katzj: Info seems aggressive for this

pyr: @katzj I'm open to ideas, but as soon as you start using this, you want to see the logs. DEBUG means compiling at `-DDEBUG` which for people just wanting to check that things are operating normally using standard packages it would become unfeasible.

katzj: Hm, yeah.  I forgot that we compile our packages with -DDEBUG just because it makes it easier for me to turn it on for people.

src/write_log.c
```diff
@@ -0,0 +1,76 @@
+/**
+ * collectd - src/write_log.c
+ * Copyright (C) 2015       Pierre-Yves Ritschard
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
```
mfournier: Maybe switch to MIT ?

src/collectd.conf.pod
```diff
@@ -7129,6 +7129,12 @@ want to set B<metadata.broker.list> to your Kafka broker list.
 
 =back
 
+=head2 Plugin C<write_log>
```
mfournier: Plugins without any configure options aren't usually mentioned in the manpage (which is something that could be discussed). On the other hand, they are listed in the README, which isn't the case yet for this plugin.

mfournier:
Seems useful, indeed. Just wondering why a new plugin rather than adding an option to the existing logfile, syslog, and log_logstash plugins ?

pyr:
@mfournier I switched to MIT and added a mention in the README. As for the separate plugin, it's just because it is not implementing the log callbacks and is independent of the log module that you will be using.

Issue #505, opened 2014-01-03
Unable to compile python plugin with python-3.3


trtrmitya:
Hello!

I try to compile collected-5.4.0 on FreeBSD with python support. I want to use python version 3
(3.3.3 to be specific).

manpage for collectd-python mentions python version 3, so I assume it is supported.

But I get the following error:

libtool: compile:  cc -DHAVE_CONFIG_H -I. -DPREFIX=\"/usr/local\" -DCONFIGFILE=\
"/usr/local/etc/collectd.conf\" -DLOCALSTATEDIR=\"/var\" -DPKGLOCALSTATEDIR=\"/v
ar/db/collectd\" -DPIDFILE=\"/var/run/collectd.pid\" -DPLUGINDIR=\"/usr/local/li
b/collectd\" -DPKGDATADIR=\"/usr/local/share/collectd\" -I/usr/local/include/pyt
hon3.3m -I/usr/local/include -Wall -Werror -fno-strict-aliasing -Wno-strict-alia
sing -O2 -pipe -O2 -fno-strict-aliasing -pipe -MT python_la-python.lo -MD -MP -MF .deps/python_la-python.Tpo -c python.c  -fPIC -DPIC -o .libs/python_la-python.
o
cc1: warnings being treated as errors
python.c: In function 'cpy_init_python':
python.c:1023: warning: passing argument 2 of 'PySys_SetArgv' from incompatible pointer type
python.c: In function 'cpy_config':
python.c:1071: warning: implicit declaration of function 'PyUnicode_SetDefaultEncoding'
gmake[3]: *** [python_la-python.lo] Error 1
gmake[3]: Leaving directory `/place/WRK/ports/net-mgmt/collectd5/work/collectd-5.4.0/src'
gmake[2]: *** [all-recursive] Error 1

trtrmitya:
Hello!

Any chance python3 support could be fixed please?

trtrmitya:
Time comes, I tried with python3.4 and it compiles!  But dumps core on start:
# collectd -f
Fatal Python error: no mem for sys.argv
ValueError: character U+6f633c00 is not in range [U+0000; U+10ffff]
Abort (core dumped)

rubenk:
Fixed by 	920d5ea

@trtrmitya I just compiled collectd with python33, python34 and python35 on FreeBSD current but can't reproduce your coredump. If you can still trigger it with the latest version of collectd from git, please open a new issue, preferable with a stack trace.

Issue #864, opened 2014-12-15
Python plugin on Solaris dumps core


yogeswaran:
When python plugin is enabled in Solaris environment, collectd dumps core. Here is the core dump:

  [1] 0xffffffff7cfdcd70(0xffffffff7fffddf8, 0x4a3, 0x1, 0x6, 0x6, 0x1), at 0xffffffff7cfdcd70 
  [2] 0xffffffff7de07ca8(0x1, 0x6, 0x0, 0xffffffff7b04c0e0, 0xffffffff79700200, 0x5), at 0xffffffff7de07ca8 
  [3] raise(0x6, 0x0, 0x1001378a8, 0xffffffffffffffff, 0xffffffff7b23e000, 0x0), at 0xffffffff7b074244 
  [4] abort(0x1, 0x1d8, 0xffffffff750423dc, 0x1f202c, 0x0, 0x0), at 0xffffffff7b04c0e0 
  [5] PyEval_SaveThread(0x0, 0x0, 0x0, 0x34000, 0xffffffff79700200, 0x1), at 0xffffffff7501097c 
  [6] cpy_init(0x0, 0x5, 0x1001444e8, 0x1, 0xffffffff7fffe120, 0x0), at 0xffffffff75408834 
  [7] plugin_init_all(0x6, 0x100031ef8, 0x0, 0x5, 0x100144bd0, 0x100144c00), at 0x10001451c 
  [8] update_kstat(0x0, 0x0, 0x1001444e8, 0xffffffff79700200, 0x2c, 0x0), at 0x1000072c4 
  [9] do_loop(0x6, 0x100030bd0, 0x1523d5d6f660ae15, 0x1001444e0, 0x1523d5e5f65d81ab, 0x780000000), at 0x100007418 
  [10] main(0x4, 0x

yogeswaran:
While attempting to debug this issue we noticed some unusual behavior.  There were two calls to plugin_init_all() / cpy_init() being performed by the main thread.  In between those calls a separate thread (as shown below) calls PyThreadState_DeleteCurrent() which clears the python global thread context _PyThreadState_Current.

After _PyThreadState_Current is set to null the 2nd cpy_init() call is made and when PyEval_SaveThread() is then called it aborts as it checks/assumes that _PyThreadState_Current is NOT null.

Can plugin_init_all() be called more than once? If so, do we need to call plugin_shutdown_all() before calling plugin_init_all() ?

[...]

PR #890, opened 2015-01-09
Fix compiler warnings with Python3.

trenkel:
PyUnicode_SetDefaultEncoding was removed from the Python source between 3.1 and 3.2. It was never a documented API so I guess I have only myself to blame for this breakage.
https://github.com/collectd/collectd/issues/505


trenkel:
Added one commit to fix the fix.

mfournier:
Please take note of my comment in #864 before merging 9bce34a please.

Other than that, +1 for fixing support for python 3.3+

trenkel:
This commit has absolutely nothing to do with #864, both are completely unrelated issues.

PR #894, opened 2015-01-15
memcached: report global Hostname by default

### LINKED ISSUE #801: memcached plugin: Unexpected hostname "127.0.0.1".
Hi,

Multi-instances doesn't seem to work on collectd 5.4 :

```
LoadPlugin memcached
<Plugin "memcached">
  <Instance "default">
    Host "127.0.0.1"
    Port 11211
  </Instance>

  <Instance "session">
    Host "127.0.0.1"
    Port 11213
  </Instance>

  <Instance "storage">
    Host "127.0.0.1"
    Port 11212
  </Instance>
</Plugin>
```

Ref: http://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_memcached

Strace output seem to be ok (for each instance) : 
```
socket(PF_INET, SOCK_STREAM, IPPROTO_TCP) = 7
connect(7, {sa_family=AF_INET, sin_port=htons(11211), sin_addr=inet_addr("127.0.0.1")}, 16) = 0
write(7, "stats\r\n", 7)                = 7
recvfrom(7, "STAT pid 11499\r\nSTAT uptime 1303"..., 4096, 0, NULL, NULL) = 1130
shutdown(7, 2 /* send and receive */)   = 0
close(7)                                = 0
```

mfournier:
When the Host parameter is left unspecified, the "host" part of the
value list should be set to the global hostname instead of the default
value hardcoded in the plugin.

Thanks to @Poil and @CyBeRoni for reporting and tracking down this
problem !

Fixes #801

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -2257,7 +2257,9 @@ following options are allowed:
 
 =item B<Host> I<Hostname>
 
-Hostname to connect to. Defaults to B<127.0.0.1>.
+Hostname to connect to. Defaults to B<127.0.0.1>. Also sets the B<host> field
+of I<value lists> when dispatching values. Defaults to the global hostname
+setting.
 
 =item B<Port> I<Port>
 
--- a/src/memcached.c
+++ b/src/memcached.c
@@ -256,7 +256,7 @@ static void memcached_init_vl (value_list_t *vl, memcached_t const *st)
       sstrncpy (vl->host, hostname_g, sizeof (vl->host));
     else
       sstrncpy (vl->host,
-          (st->host != NULL) ? st->host : MEMCACHED_DEF_HOST,
+          (st->host != NULL) ? st->host : hostname_g,
           sizeof (vl->host));
     sstrncpy (vl->plugin_instance, st->name, sizeof (vl->plugin_instance));
   }


mfournier:
@CyBeRoni yes, I'm aware of this. I don't have the whole history of this plugin, but it looks like it was made so on purpose. Else the whole if/else logic in `memcached_init_vl()` doesn't make any sense.

I want to avoid breaking the plugin for people possibly relying on this feature. My idea here is to change a default value which doesn't make sense + better document the behaviour for people monitoring memcached from a remote collectd instance.

CyBeRoni:
I agree with that notion, but the current behaviour is entirely inconsistent with the other plugins. At least, I haven't seen any other plugins that use config settings for the hostname unless explicitly configured to do so.

I feel like the default here should be to be consistent with the other plugins. I understand it might be useful for some that monitor remote memcached instances to have the behaviour as implemented now, though I think this is exactly what the named instances are useful for. But I can see how you might like to have the info show up under the hostname of the machine being monitored.

Perhaps the safest way, for now at least, is to default to using hostname_g if the monitored memcached runs on localhost, and ideally make a config option that allows you to be explicit about this choice (i.e., switch between hostname_g and st->host) The default for that, I think should be hostname_g, though for backwards compatibility it can also be st->host. Perhaps the default should be switched on a .x version change.

In any case, the situation as it is now is impossible to use for people like @Poil who have multiple memcached instances to monitor on multiple hosts, and want to do it from the local host, with an identical config pushed to each host, because then you end up with 127.0.0.1 being clobbered. I only use a single instance on a bunch of hosts, so though I have the same issue I can get away with using the legacy config options to prevent it.

CyBeRoni:
I'm just now realising that the solution as implemented here should work if you want to connect to localhost and, then just leave out the 'Host' statement. I can test that. The problem remains if you're monitoring locally but don't have memcached on 127.0.0.1, or are monitoring remotely and *don't* want the data to show up under the memcached hostname.

octo:
Closing this in favor of f35f67df9ccfb80c001ed0befcce8bb90ebb2a07.

PR #895, opened 2015-01-16
fixes for perl and contrib subpackage

faxm0dem:
* perl: preserve full path of perl examples
* move collection3 perl from perl subpackage to contrib
* filter unwanted dependencies from perl scripts

tested against collectd-5.4.1.729.g21f35cd

Change-Id: I7f3144cf4a30e64f39b4ebe25108939ae5efaab1

### DIFF
--- a/contrib/redhat/collectd.spec
+++ b/contrib/redhat/collectd.spec
@@ -208,7 +208,7 @@
 Summary:	Statistics collection daemon for filling RRD files
 Name:		collectd
 Version:	5.4.0
-Release:	1%{?dist}
+Release:	2%{?dist}
 URL:		http://collectd.org
 Source:		http://collectd.org/files/%{name}-%{version}.tar.bz2
 License:	GPLv2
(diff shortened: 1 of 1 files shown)


mfournier:
Thanks @faxm0dem !

Here are my thoughts about this:

 - +1 for the "perl_default_filter" thing. I didn't know this macro, it seems to be exactly what we need. I'd rather put it at the top of the specfile though, so that it benefits all subpackages.
 - what do you think about *not* copying the perl stuff from contrib/ into collectd-perl.rpm at all ? The `find` currently doesn't match the perl code with a `.px` extension, misses the config files that come with them, flattens the directory structure. Blah ! IMHO we should just leave the stuff in contrib/ in collectd-contrib.rpm (except for collection3 and php-collection).
 - collection3 and php-collection should actually get moved, not copied, to their final locations. What is done now makes no sense, and the `-not -path '*/collection3/*'` is merely a workaround for this.

I can prepare a patch for this if you agree with my points. Just let me know !

faxm0dem:
* The reason I didn't put the macro at the top was to avoid long rpmbuild runtime when not using perl. But agreed: the fedora recommendations are to put it before the first `%description` section anyway :+1:
* :+1: for not copying contrib files at all
* :+1:

mfournier:
Replaced by a06c934c92. Thanks @faxm0dem !

PR #900, opened 2015-01-20
implement reporting of ncpu

faxm0dem:
Change-Id: Ia305b74139352f16f71b1f747875d6c7a0c376a4

### DIFF
--- a/src/cpu.c
+++ b/src/cpu.c
@@ -466,11 +466,9 @@ static void cpu_commit_one (int cpu_num, /* {{{ */
 /* Commits the number of cores */
 static void cpu_commit_num_cpu (gauge_t num_cpu) /* {{{ */
 {
-//// static void submit_value (int cpu_num, int cpu_state, const char *type, value_t value)
 	value_t values[1];
 	value_list_t vl = VALUE_LIST_INIT;
 
-	/// memcpy(&values[0], &value, sizeof(value));
 	values[0].gauge = num_cpu;
 
 	vl.values = values;
@@ -478,9 +476,7 @@ static void cpu_commit_num_cpu (gauge_t num_cpu) /* {{{ */
 
 	sstrncpy (vl.host, hostname_g, sizeof (vl.host));
 	sstrncpy (vl.plugin, "cpu", sizeof (vl.plugin));
-	sstrncpy (vl.type, "gauge", sizeof (vl.type));
-	sstrncpy (vl.type_instance, "ncpu",
-			sizeof (vl.type_instance));
+	sstrncpy (vl.type, "count", sizeof (vl.type));
 
 	plugin_dispatch_values (&vl);
 } /* }}} void cpu_commit_num_cpu */
@@ -520,9 +516,7 @@ static void cpu_commit (void) /* {{{ */
 	};
 
 	if (report_num_cpu)
-	{
 		cpu_commit_num_cpu ((gauge_t) global_cpu_num);
-	}
 
 	if (report_by_state && report_by_cpu && !report_percent)
 	{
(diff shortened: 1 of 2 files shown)


faxm0dem:
@octo please review

faxm0dem:
Just adding some context here.
The idea behind this change is that when using percentages instead of ticks, there is no way to know the number of cores anymore using collectd alone. While it might seem weird to report a usually constant number, it **has** shown to be useful in different cases to know this number, *e.g.* :

* broken cpu after a reboot
* noticing changes after a resize of a cloud instance

faxm0dem:
@octo do you think this could make it to the next release?

octo:
Hi @faxm0dem,

this looks good overall, thanks for the patch! I'll leave some small review comments inline.

Best regards,
—octo

src/cpu.c
```diff
@@ -452,6 +456,28 @@ static void cpu_commit_one (int cpu_num, /* {{{ */
 	}
 } /* }}} void cpu_commit_one */
 
+/* Commits the number of cores */
+static void cpu_commit_num_cpu (gauge_t num_cpu) /* {{{ */
+{
+//// static void submit_value (int cpu_num, int cpu_state, const char *type, value_t value)
```
octo: Please remove this comment.

src/cpu.c
```diff
@@ -452,6 +456,28 @@ static void cpu_commit_one (int cpu_num, /* {{{ */
...
+	value_t values[1];
+	value_list_t vl = VALUE_LIST_INIT;
+
+	/// memcpy(&values[0], &value, sizeof(value));
```
octo: Please remove this comment.

src/cpu.c
```diff
@@ -452,6 +456,28 @@ static void cpu_commit_one (int cpu_num, /* {{{ */
...
+	values[0].gauge = num_cpu;
+
+	vl.values = values;
+	vl.values_len = 1;
+
+	sstrncpy (vl.host, hostname_g, sizeof (vl.host));
+	sstrncpy (vl.plugin, "cpu", sizeof (vl.plugin));
+	sstrncpy (vl.type, "gauge", sizeof (vl.type));
```
octo: I suggest to use type "count" and an empty type instance for this metric, i.e. "<host>/cpu/count".

src/cpu.c
```diff
@@ -493,6 +519,11 @@ static void cpu_commit (void) /* {{{ */
 	};
 	size_t cpu_num;
 
+	if (report_num_cpu)
+	{
```
octo: Please remove the curly braces around this single-line block.

octo:
Good to go once the review comments are resolved.

faxm0dem:
fixed and rebased!

src/cpu.c
```diff
@@ -459,6 +463,24 @@ static void cpu_commit_one (int cpu_num, /* {{{ */
 	}
 } /* }}} void cpu_commit_one */
 
+static void cpu_commit_num_cpu (gauge_t num_cpu) /* {{{ */
+{
+//// static void submit_value (int cpu_num, int cpu_state, const char *type, value_t value)
```
rubenk: could you please remove this?

octo: That's the comment I mentioned earlier. I think you mistakenly removed this comment instead:

```c
/* Commits the number of cores */
```

rubenk: Good thing that we have two people reviewing ;)

faxm0dem:
sorry about that - I hope I got it right this time

rubenk:
Thanks @faxm0dem!

PR #902, opened 2015-01-20
write_http: deprecate <URL> blocks in favor of <Node>

mfournier:
This makes the plugin use `<Node>` blocks like most other write plugins,
while maintaining backwards compatibility with `<URL>` blocks.

It's a follow up to #899, which was merely a fix for the release
branches.

It should also make using write_http with filter chains more convenient.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1277,7 +1277,8 @@
 #</Plugin>
 
 #<Plugin write_http>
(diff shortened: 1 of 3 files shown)


pyr:
works great, :+1:

mfournier:
Thanks for testing & giving feedback @pyr !

PR #903, opened 2015-01-21
Fix memory leak

forsyth:
A check was missing in one path that wrote to the queue, allowing it to creep up beyond the boundary.
The original code would force a message once a second as the queue filled up (just what's needed when
running behind). I've reduced the frequency, limited messages to appear when the loss level
has changed, and ensured that if it later goes wrong again after recovering, it will again produce messages.

Separately, in error cases that were probably rare but possible, putval could leak value lists.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5097,6 +5097,7 @@ AC_COLLECTD([getifaddrs],[enable],  [feature], [getifaddrs under Linux])
 dependency_warning="no"
 dependency_error="no"
 
+plugin_aoe_stats="no"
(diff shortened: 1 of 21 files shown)


pyr:
@forsyth Hi!, there are some unrelated commits in the PR, can you clean-it up and re-push ?

forsyth:
yes I am sorry about that. It was meant for a local fork initially, and I didn't notice the default destination for the pull request was to the master. I was just looking round for the way to cancel/delete it.

PR #905, opened 2015-01-22
memcached plugin: Implement the "Alias" option.

mre:
We have two ip addresses for memcached:
One backbone address and one internal address.

Our hostfile looks a bit like this:
```
1.2.3.1 mc0 mc0.example.com # Internal address (only available from company office)
1.2.3.2 mc0 mc0.example.com # Backbone address (only available from datacenter)
```

We need the backbone address for security reasons here to avoid that somebody
from the office can accidentally flush the cache from the office :wink: 

In our collectd.conf we have the following settings:
```
LoadPlugin memcached
<Plugin memcached>
    <Instance "default">
        Host "1.2.3.2"
        Port "11211"
    </Instance>
    <Instance "pool_one">
        Host "1.2.3.2"
        Port "11212"
    </Instance>
    <Instance "pool_two">
        Host "1.2.3.2"
        Port "11213"
    </Instance>
</Plugin>
```

We can’t use the fqdn in our config file because it would resolve to the internal address (which comes first in the host file) instead of the backbone address.
The end result is, that our series names contain the IP address instead of the fqdn. 
So 
`datacenter.memcached.1.2.3.2.memcached-data.memcached_command-get.mean`
instead of
`datacenter.memcached.mc0.memcached-data.memcached_command-get.mean`

To fix this, I’ve added an `alias` parameter to the memcached plugin.
If it is given, the alias will appear in the series name.
If not, the normal behavior (using IP or fqdn as before) will be used.

This can also be useful if you have multiple memcached instances on one machine which are completely
separate from each other. They can all have a separate alias, as if they were running on different machines.

It would be nice to have this functionality upstream. What do you think?

### DIFF
--- a/src/memcached.c
+++ b/src/memcached.c
@@ -269,6 +269,7 @@ static void memcached_init_vl (value_list_t *vl, memcached_t const *st)
           sizeof (vl->host));
       }
       sstrncpy (vl->plugin_instance, st->name, sizeof (vl->plugin_instance));
+    }
   }
 }
 
--- a/src/memcached.c
+++ b/src/memcached.c
@@ -63,6 +63,7 @@ static void memcached_free (memcached_t *st)
   sfree (st->socket);
   sfree (st->host);
   sfree (st->port);
+  sfree (st->alias);
 }
 
 static int memcached_connect_unix (memcached_t *st)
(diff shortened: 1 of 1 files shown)


MakaHost:
+1

mre:
Any update on this?

mfournier:
@mre, this is related to the discussions in #801 and #894. From my pov there isn't a consensus on what should be done yet.

I'm personnally not convinced adding an Alias option is the right way to fix this problem, but I'd like to have @Poil and @CyBeRoni's opinion too, as they seem to use the multi-instance feature too.

CyBeRoni:
I would expect any data source to, unless explicitly configured otherwise, show up under the hostname of the machine on which it runs, such as 'prefix.collectdhostname.memcached.instancename.data'. This 'Alias' directive would be a way to explicitly configure it otherwise if that suits your needs.

As per @Poil's list (https://github.com/collectd/collectd/issues/801#issuecomment-703615440), most plugins work that way already and I think it'd be great if they all did.

mre:
So can we merge this?

mre:
Any updates on this guys? It would be nice to have this integrated upstream as we currently need to maintain our own fork for this.

octo:
I have to admit I'm not a huge fan of the *Alias* name for this option. I'd prefer to mimick other plugins and use the *Host* config option to set the *host* field of the identifier. The address used for network connections would default to *Host* but could be overridden by the *Address* option. For example:

```
<Instance "905">
  Host "example.com"
  Address "memcache.mgmt.internal.example.com"
</Instance>
```

The *modbus plugin* uses exactly these names, for example, and many other plugins use *Host* for the identifier, so there's enough precedence IMHO.

Best regards,
—octo

octo:
Superseded by #1975.

PR #908, opened 2015-01-23
libstatgrab: fix detection of libstatgrab >= 0.90

vincentbernat:
Ensure linking is done after compilation of the test program, otherwise
symbols cannot be found. `LDFLAGS` and `CFLAGS` are placed before the
test program on gcc command line while `LIBS` is placed after. Even if
in our case, `LDFLAGS` also contain the lib, being put before the
program while the linking phase is done with the same command leads to
`sg_init()` never found which is misinterpreted as requiring an
argument.

This has been tested with both libstatgrab 0.90 and 0.17.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4139,10 +4139,11 @@ fi
 if test "x$with_libstatgrab" = "xyes"
 then
   SAVE_CFLAGS="$CFLAGS"
-  SAVE_LDFLAGS="$LDFLAGS"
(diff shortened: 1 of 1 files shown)


pyr:
works here, :+1:

mfournier:
Thanks a lot @vincentbernat !

I just tested your patch on EL6 (0.17) and debian oldstable (0.16) and all looks good.

PR #909, opened 2015-01-23
added doc CollectStatistics to section rrdcached

michakrause:


mfournier:
Thanks a lot @michakrause ! Cherry-picked as 8fa66d769 to the earliest bugfix branch, so it will be part of all subsequent bugfix & feature releases.

Issue #869, opened 2014-12-17
ping_init() called multiple times on Solaris


julianrct:
On Solaris when the kstat chain is updated it looks like ping_init() (and possibly other plugin _init() functions) are called again. In the case of ping_init() this causes the plugin to fail, since it tries for a second time to create the read thread.

The following simple patch fixes the issue:

```
--- src/ping.c.orig     2014-01-26 08:09:14.000000000 +0000
+++ src/ping.c  2014-12-17 09:52:19.363630000 +0000
@@ -437,7 +437,10 @@
         "Will use a timeout of %gs.", ping_timeout);
   }

-  if (start_thread () != 0)
+  /* ping_init() can be called more than once if the kstat chain is updated,
+     so guard against creating multiple read threads. */
+
+  if (ping_thread_loop == 0 && start_thread () != 0)
     return (-1);

   return (0);
```

Regards,
Julian

octo:
When called again, *start_thread()* returns non-zero, causing the init function to signal failure, which causes *plugin_unregister_read()* to be called for that plugin.

The simplest possible fix is to return gracefully, i.e. a zero, in that case.

PR #911, opened 2015-01-25
Rework Solaris kstat handling, avoiding repeated plugin_init_all() calls

baryonix:
It has been noted in some issues such as #864 and #869 that on Solaris, plugin_init_all() gets called repeatedly, namely every time kstat_chain_update() indicates that the kstat chain has been updated.

It was also noted that on many Solaris systems, kstat_chain_update() will _very_ frequently indicate a kstat chain update (basically, this will happen on every collectd interval), so even if plugins had a "reload" callback to be called instead of repeated calls to the init callback, we would still end up calling this reload callback way too frequently.

This patch attempts to solve this in a somewhat more complex way, maintaining a hash table that reflects the current state of the kstat chain, and on chain updates, it will compare the new chain against the saved state in the hash table to figure out what kstat items exactly have appeared and disappeared.

Plugins can register a callback function to be notified about added/removed kstats. They can specify filters to ensure plugin's callback function only gets invoked for kstats which the plugin is actually interested in. Additionally, a standard callback function is provided which maintains a "set" of relevant kstats in a data structure that a plugin's read function can simply iterate over.

This adds some complexity to core, but in turn makes it slightly simpler to write plugins making use of kstats that may change dynamically.

I've also converted those plugins for which it makes sense to use this new kstat infrastructure. This affects the cpu, disk, interface, processes and tape plugins. There are some more plugins using kstat, but those can stay as they are, as they don't utilize any kstats which would be expected to dynamically appear or disappear at run time.

### DIFF
--- a/src/collectd.c
+++ b/src/collectd.c
@@ -44,6 +44,171 @@ cdtime_t interval_g;
 int  timeout_g;
 #if HAVE_LIBKSTAT
 kstat_ctl_t *kc;
+
+
+/* FNV-1a hash algorithm (32-bit) */
+typedef uint32_t hash_t;
+static const hash_t hash_init = 2166136261UL;
+static const hash_t hash_mult = 16777619UL;
+
+/* for 64-bit hash use:
+typedef uint64_t hash_t;
+static const hash_t hash_init = 14695981039346656037ULL;
+static const hash_t hash_mult = 1099511628211ULL;
+*/
+
+static inline hash_t
+hash_update (hash_t h, unsigned char input)
+{
+	return (h ^ input) * hash_mult;
+}
+
+static inline hash_t
+hash_update_str (hash_t h, const char *str)
+{
+	for (; *str != 0; str++)
+		h = hash_update (h, (unsigned char) *str);
+	return h;
+}
+
+static inline hash_t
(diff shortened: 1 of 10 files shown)


mfournier:
Thanks a lot for working on this tricky problem @baryonix !

2 blockers from my point of view:
 - This work needs to go to the master branch, as it's a significant change in the way collectd behaves on solaris + it changes the API. Only bugfixes go to the collectd-5.x branches.
 - I'm not sure anyone on the team currently works with collectd on this platform, so we'll need to rely on solaris users to review and test your work. @ymettier, @yogeswaran, @julianrct ?

baryonix:
Okay. Rebasing this to master should be trivial (already checked that), I'll do that and push an update as soon as I can.

Regarding testing, these changes have of course been verified in a lab environment, but it would certainly make sense to do some more real-world testing. I run a couple of Solaris servers at work and would be able to contribute in theory - though this would require me to upgrade some boxes to run the master branch instead of the 5.4.1 release they're running currently.

xtra-be:
i will be able to test on solaris 10 & 11.2

baryonix:
Closing this pull request, superseded by #915.

PR #915, opened 2015-01-26
Rework Solaris kstat handling, avoiding repeated plugin_init_all() calls [rebased to master]

baryonix:
This is the same patch as #911, only rebased to the master branch. Copy of description from #911 follows:

It has been noted in some issues such as #864 and #869 that on Solaris, `plugin_init_all()` gets called repeatedly, namely every time `kstat_chain_update()` indicates that the kstat chain has been updated.

It was also noted that on many Solaris systems, `kstat_chain_update()` will very frequently indicate a kstat chain update (basically, this will happen on every collectd interval), so even if plugins had a "reload" callback to be called instead of repeated calls to the init callback, we would still end up calling this reload callback way too frequently.

This patch attempts to solve this in a somewhat more complex way, maintaining a hash table that reflects the current state of the kstat chain, and on chain updates, it will compare the new chain against the saved state in the hash table to figure out what kstat items exactly have appeared and disappeared.

Plugins can register a callback function to be notified about added/removed kstats. They can specify filters to ensure plugin's callback function only gets invoked for kstats which the plugin is actually interested in. Additionally, a standard callback function is provided which maintains a "set" of relevant kstats in a data structure that a plugin's read function can simply iterate over.

This adds some complexity to core, but in turn makes it slightly simpler to write plugins making use of kstats that may change dynamically.

I've also converted those plugins for which it makes sense to use this new kstat infrastructure. This affects the cpu, disk, interface, processes and tape plugins. There are some more plugins using kstat, but those can stay as they are, as they don't utilize any kstats which would be expected to dynamically appear or disappear at run time.


baryonix:
There is one thing I noticed during today's first round of real-world testing. It looked much like there was a performance regression as a version built with these changes would consistently use about 2-3 times more CPU than one without.

Turns out the abundance of interface-related kstats has grown really wild in Solaris 11 (especially when zones are involved), and on the system I was testing on, the interface plugin would pick up some 1800-something interface kstats to report upon. The old code put a hard limit at 256 kstats in the list, but the new code doesn't have such a limitation, and this was the reason for the increased CPU usage.

Of course most of these kstats are really bogus, and the old code wouldn't make anything useful out of them either. Most of them ended up being reported as (nonexistent) interfaces with names such as "link", "phys", or "mac_tx_hwlane0".

collectd needs to be much more selective about which kstats are actually worth reporting as interface statistics. I would like to include such a fix in this pull request. Right now, I'm still busy figuring out how to best make sense out of this.

ymettier:
Hello,

I read some code. Nice job !
I cannot remember of an issue on Solaris that it can collect only from 256 CPU max or only from 1024 disks max, but if there was, it is fixed here :
```
-#define MAX_NUMDISK 1024
extern kstat_ctl_t *kc;
-static kstat_t *ksp[MAX_NUMDISK];
```

Sorry, I cannot test this patch soon. But it looks as a nice implementation. There were the "init", "config", "read", "write", "log", "flush", "shutdown" events (and registered callbacks). Now there is a new one (specific to Solaris) for kstat. Having a new event is the best solution in my mind.

One more suggestion : prepare an update to https://collectd.org/wiki/index.php/Plugin_architecture.

Regards,
Yves

baryonix:
I've been testing this (combined with the changes proposed in #919) for about a week now, on two busy servers, one on Solaris 10 and one on Solaris 11.1. So far I haven't had any issues. So if anyone wants to help testing, feel free to join in - it seems to be relatively safe. :-)

poolski:
Please merge this soon!

mfournier:
@poolski, have you been able to test this patchset (ideally together with #919) in a solaris environment ? It would be a great way to help this move forward !

Let me know if you need some help to build collectd from the git repo !

poolski:
@mfournier, I would love to get help with building in Solaris - I'm mostly a Debian guy.

mfournier:
@poolski, if you're familiar with debian, I'd recommend starting by building a release tarball on a debian box. Here would be the steps:

```
git clone https://github.com/collectd/collectd
git remote add baryonix https://github.com/baryonix/collectd
git remote update
git branch foo origin/master
git checkout foo
git merge baryonix/kstat-fixes
git merge baryonix/solaris-misc-fixes
sudo apt-get install autoconf automake bison cpp flex gcc libc6-dev libtool m4 make pkg-config
./build.sh
./configure
make dist-gz
```

This should output a tarball you can extract on the solaris box and then run the build with the usual `./configure && make && make install`. Not sure of the specific things you'd need to install on solaris, but I guess the basic toolchain should be suffiscient.

baryonix:
I've pushed an updated version of this. The main change is that it now uses the generic hash table implementation provided in #1318 instead of rolling its own. The commits are also rebased to the current state of the master branch.

octo:
I would actually like to get as much of the *kstat* handling out of the daemon and into the plugins that need it. This patch appears to go the opposite direction, i.e. there is now much *more* kstat related code in the daemon.

Which parts of kstat absolutely *have* to be in the daemon? Can we get rid of all of it?

PR #919, opened 2015-01-28
Fixes for zfs_arc and processes plugins on Solaris

baryonix:
- zfs_arc: Silence errors about failure to read the "l2_size" value from the zfs:0:arcstats kstat. This value is missing on several versions of Solaris.

- processes: Several fixes to make the plugin work on Solaris in 32-bit and 64-bit builds. Enable the plugin by default.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5228,6 +5228,7 @@ fi
 if test "x$with_kstat" = "xyes"
 then
 	plugin_nfs="yes"
+	plugin_processes="yes"
 	plugin_uptime="yes"
 	plugin_zfs_arc="yes"
 fi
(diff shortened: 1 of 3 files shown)


baryonix:
I've pushed an updated (and rebased) version which will make the processes plugin work on Solaris in both 32-bit and 64-bit builds. Building of the plugin is now enabled by default for all Solaris builds.

mfournier:
I confirm this builds & works on a solaris11.2 box. Also, no side effects would be expected on other platforms as the changes are all confined inside `#if KERNEL_SOLARIS` blocks or equivalent. So +1 to this patchset.

octo:
LGTM

mfournier:
PS: *many thanks* @baryonix !! It's awesome to have people with the skills to improve support for solaris !

PR #922, opened 2015-01-29
write_graphite plugin: Add SendTimeout

lucasmo:
When sending metrics to remote networks, occasionally the write_graphite plugin will hang in a socket write call waiting for the write to succeed. This can hang indefinitely, causing metrics to not be reported until the service is forcibly killed.

This pull request does not change the default behavior, but adds a SendTimeout to the write_graphite plugin configuration to abort sending metrics after a certain amount of time has passed while stuck in a write. All unwritten data is lost and this is treated as a failure.

### DIFF
--- a/src/write_graphite.c
+++ b/src/write_graphite.c
@@ -98,6 +98,7 @@ struct wg_callback
     char    *node;
     char    *service;
     char    *protocol;
+    int      send_timeout;
     _Bool   log_send_errors;
     char    *prefix;
     char    *postfix;
@@ -251,6 +252,25 @@ static int wg_callback_init (struct wg_callback *cb)
             close (cb->sock_fd);
             cb->sock_fd = -1;
             continue;
+        }
+
+        if (cb->send_timeout > 0)
+        {
+            struct timeval timeout;
+            timeout.tv_sec = cb->send_timeout;
+            timeout.tv_usec = 0;
+
+            status = setsockopt (cb->sock_fd, SOL_SOCKET, SO_SNDTIMEO,
+                    (char *)&timeout, sizeof (timeout));
+            if (status != 0)
+            {
+                char errbuf[1024];
+                snprintf (connerr, sizeof (connerr), "failed to set send timeout: %s",
+                        sstrerror (errno, errbuf, sizeof (errbuf)));
+                close (cb->sock_fd);
+                cb->sock_fd = -1;
+                continue;
+            }
         }
 
         break;
@@ -489,6 +509,7 @@ static int wg_config_node (oconfig_item_t *ci)
     cb->node = NULL;
     cb->service = NULL;
     cb->protocol = NULL;
+    cb->send_timeout = -1;
     cb->log_send_errors = WG_DEFAULT_LOG_SEND_ERRORS;
     cb->prefix = NULL;
     cb->postfix = NULL;
@@ -529,6 +550,8 @@ static int wg_config_node (oconfig_item_t *ci)
                 status = -1;
             }
         }
+        else if (strcasecmp ("SendTimeout", child->key) == 0)
+            cf_util_get_int (child, &cb->send_timeout);
         else if (strcasecmp ("LogSendErrors", child->key) == 0)
             cf_util_get_boolean (child, &cb->log_send_errors);
         else if (strcasecmp ("Prefix", child->key) == 0)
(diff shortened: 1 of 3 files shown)


DaveBlooman:
Any idea when this will get merged?

mfournier:
On Thu, Jan 29, 2015 at 02:34:19PM -0800, lucasmo wrote:
> When sending metrics to remote networks, occasionally the write_graphite
> plugin will hang in a socket write call waiting for the write to succeed.
> This can hang indefinitely, causing metrics to not be reported until the
> service is forcibly killed.

@lucasmo, any hint on how to get into this plugin hang situation ? Were you
able to reliably reproduce this error condition to test your patch ?

Several different "write_graphite stalls" issues have been reported, so I
just want to make sure this patch wouldn't just hide the symptom of another
issue.

rtkmhart:
I saw something similar a while ago but haven't had time to back and try figure it out. Configuration was a node connecting to a network with OpenVPN. If the OpenVPN tunnel was torn down and rebuilt connecting to a different VPN concentrator, collectd wouldn't notice. Effectively the IP on the interface that collectd was using changed, and it just hung until restarted. I suspect but have no proof that it sent a packet, never got back an ACK and then just waited forever for it. I may have a bit of time this afternoon to try track it down.

rubenk:
Perhaps it's an idea to enable keepalive on the socket?

lucasmo:
I've simulated this by using an iptables rule to block traffic coming back from graphite (just the ACK packets). This seems to work in this condition.

However, I wrote this patch to address a problem in which the connection would become backed up with a huge sendqueue, which eventually would percolate back to making the send() call take minutes at a time. In this case, I want the connection to abort, drop the sendqueue, and try again. This would not be addressed by keepalives.

Keepalives are sent when no data has been sent on the channel for a period of time, which appears to be 2 hours by default. I don't see any harm in enabling them and setting the timeout to a lower period of time as well.

mfournier:
Ok, thanks for the details, @lucasmo! I now see what you mean.

When I block the traffic using iptables, I can see data piling up in the send-Q, and collectd's memory usage growing once the send-Q is full. But soon after I unblock iptables, all the buffered data gets flushed to graphite (netcat actually).

Unfortuately, I failed to see any difference when using `SentTimeout`. I'm certainly missing something :(

Assuming the connection recovers by itself, I'd rather recommend people to use the `WriteQueueLimit*` options to take care of dropping data when write_graphite stalls. Of course if there is a situation where the connection is unable to recover, we definitely need to figure that out.

mfournier:
In the meantime #1104 got merged in. I'm wondering if it might fix the initial problem which led to this patch ?

If by any chance someone suffering from this problem had a chance to test, it would be great ! Thanks !

src/write_graphite.c
```diff
@@ -98,6 +98,7 @@ struct wg_callback
     char    *node;
     char    *service;
     char    *protocol;
+    int      send_timeout;
```
octo: Please use the `cdtime_t` type for time values. (also see below)

src/write_graphite.c
```diff
@@ -253,6 +254,25 @@ static int wg_callback_init (struct wg_callback *cb)
             continue;
         }
 
+        if (cb->send_timeout > 0)
+        {
+            struct timeval timeout;
+            timeout.tv_sec = cb->send_timeout;
```
octo:
```c
CDTIME_T_TO_TIMEVAL(cb->send_timeout, &timeout);
```

src/write_graphite.c
```diff
@@ -529,6 +550,8 @@ static int wg_config_node (oconfig_item_t *ci)
                 status = -1;
             }
         }
+        else if (strcasecmp ("SendTimeout", child->key) == 0)
+            cf_util_get_int (child, &cb->send_timeout);
```
octo:
```c
cf_util_get_cdtime (child, &cb->send_timeout);
```

octo:
I'm happy to merge this, but I'm intrigued if this and #1104 solve the same problem. Can anybody comment?

lucasmo:
This never went in to testing on my end so I can't verify it works. Sorry for dropping off, other things got more pressing.

It's possible that #1104 addresses this condition, but I would think that the send() call would return an error when the remote peer was disconnected instead of hanging indefinitely. The real robust solution here would be to use nonblocking I/O entirely, but that involves maintaining and managing buffers.

lucasmo:
This is ancient, closing.

PR #923, opened 2015-02-01
TCPConns support for reporting per-connection statistics (RTT) on Linux

arielshaqed:
See collectd.conf.pod for details.

Adds the ability to report per-connection statistics on TCP connections when reading Linux tcp_diag module. Other read methods are NOT supported.

### DIFF
--- a/src/tcpconns.c
+++ b/src/tcpconns.c
@@ -77,6 +77,8 @@
 #if HAVE_LINUX_INET_DIAG_H
 # include <linux/inet_diag.h>
 #endif
+#include <linux/rtnetlink.h>
+# include <netinet/tcp.h>
 # include <sys/socket.h>
 # include <arpa/inet.h>
 /* #endif KERNEL_LINUX */
@@ -273,12 +275,18 @@ static const char *config_keys[] =
   "ListeningPorts",
   "LocalPort",
   "RemotePort",
-  "AllPortsSummary"
+  "AllPortsSummary",
+  "ReportByPorts",
+  "ReportByConnections",
+  "ConnectionsAgeLimitSecs",
 };
 static int config_keys_num = STATIC_ARRAY_SIZE (config_keys);
 
 static int port_collect_listening = 0;
 static int port_collect_total = 0;
+static int report_by_connections = 0;
+static int report_by_ports = 1;
+static int connections_age_limit_msecs = -1;
 static port_entry_t *port_list_head = NULL;
 static uint32_t count_total[TCP_STATE_MAX + 1];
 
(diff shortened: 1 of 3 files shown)


arielshaqed:
Ping?
Am I missing some step to progress on this patch (or am I missing just patience)?
Thanks,
-- 
Ariel.

mfournier:
Sorry for the lack of reactivity @arielshaqed :-(

This patch you have looks interesting, thanks a lot !

It's a significant change to an already complex plugin (if we take the number of supported platforms in account). We won't focus on this at the moment, as we're trying to get the 5.5 release out.

But please keep on updating the PR. If other people show interest, or even better test out the patch and report back, it will be a good indicator that we should work to get this in quickly after the release.

Thanks for your patience !

arielshaqed:
Hi @mfournier! I can see you're busy with lots of other patches.
I understand your primary concern is with increasing complexity of TCPConns. I could rework the path to create an entirely new plugin -- would that be easier to incorporate?
Thanks!

octo:
Hi @arielshaqed and @yosih, I'm very sorry for the late reply.  The code looks good overall, but I'll leave some review comments inline.

Did you do any benchmarking to see if batching calls to `plugin_dispatch_values()` actually has a performance gain? I would have expected the bottleneck elsewhere.

I have some minor cleanup changes in https://github.com/octo/collectd/tree/pr/923, could you please merge them into your branch?

src/tcpconns.c
```diff
@@ -273,12 +275,18 @@ static const char *config_keys[] =
...
+  "AllPortsSummary",
+  "ReportByPorts",
+  "ReportByConnections",
+  "ConnectionsAgeLimitSecs",
 };
 static int config_keys_num = STATIC_ARRAY_SIZE (config_keys);
 
 static int port_collect_listening = 0;
 static int port_collect_total = 0;
+static int report_by_connections = 0;
```
octo: Nit: It'd be cleaner if the `report_by_*` variables were of type `_Bool` (a weirdly named C99 type).

src/tcpconns.c
```diff
@@ -486,6 +494,133 @@ static int conn_handle_ports (uint16_t port_local, uint16_t port_remote, uint8_t
...
+static void value_list_batch_flush(value_list_batch_t *batch)
+{
+    size_t i;
+    /* TODO(arielshaqed): Use new batched reporting API! */
+    for (i = 0; i < batch->cur; i++) {
+        plugin_dispatch_values (&batch->buffer[i]);
+    }
+    for (i = 0; i < batch->cur; i++) {
+        free(batch->buffer[i].values);
+        /* Also free meta? */
```
octo: `plugin_dispatch_values()` gets a `const` pointer, so it will not allocate a meta object that you would have to free here.

octo:
I've done some very basic profiling, but could only test with a normal box (i.e. not that many TCP connections). In that case, `conn_read_netlink()` spends 75% of it's time in `value_list_batch_create()`.

I have pushed a commit to my `pr/923` branch which replaces the batching implementation with a single `value_list_t` which is reused for each connection, which reduces the initialization cost to nearly zero: octo@7aee5cd5253ccf50ef5975f10432e90e22806dfa

One problem I'm seeing however is that the *plugin instance* is not large enough to hold two IPv6 addresses (plus two ports, plus state), so I'm seeing quite a few truncated connections. This is likely to cause duplicate metric warnings and corrupted data on a busy system, e.g. when talking to two devices on the same IPv6 subnet. I don't have a clever idea right now how to work around this. :(

octo:
FYI, version 5.6 increases the sizes for the identifier, so two IPv6 addresses should now fit.

PR #925, opened 2015-02-03
Adds IPC Plugin with documentation

ajdiaz:
The IPC module allows collectd to gather information about process IPC
mechanims, like used semaphores, size of shared memory segments and so
on...

This is an old module which was missing during the big refactor waiting for
rebasing the current master. I hope that this new PR can be merged :smiley:


ajdiaz:
By the way, the old PR was #463

src/types.db
```diff
@@ -176,7 +179,13 @@ response_code		value:GAUGE:0:U
 route_etx		value:GAUGE:0:U
 route_metric		value:GAUGE:0:U
 routes			value:GAUGE:0:U
+sem_used_arrays    value:GAUGE:0:65535
+sem_used    value:GAUGE:0:65535
 serial_octets		rx:DERIVE:0:U, tx:DERIVE:0:U
+shm_bytes_rss    value:GAUGE:0:65535
+shm_bytes_swapped  value:GAUGE:0:65535
+shm_bytes_total    value:GAUGE:0:65535
+shm_segments    value:GAUGE:0:65535
```
mfournier: @ajdiaz we're trying to avoid infinite multiplication of the amount of entries in types.db. Would it be possible in some of the cases above to use type-instances to existing types (or to a reduced amount of new types) ? Thanks !

ajdiaz: Sure, I will refactor the code to use less types and and will update the PR soon :)

ajdiaz:
@mfournier, changes done :)

Now uses a combination of plugin_instance and type_instance to avoid create a number of types in types.db, but "segments" one still necessary (I can't find another with similar meaning).

configure.ac
```diff
@@ -5158,6 +5158,7 @@ then
 	plugin_entropy="yes"
 	plugin_fscache="yes"
 	plugin_interface="yes"
+	plugin_ipc="yes"
```
mfournier: As this plugin seems to also work in AIX, would you like to add it below (around line 5200) to build it by default on this platform ?

src/collectd.conf.pod
```diff
@@ -2326,6 +2326,12 @@ other interfaces are collected.
 
 =back
 
+=head2 Plugin C<ipc>
```
mfournier: FYI, we usually don't mention plugins which have no configuration options in the manpage.

src/ipc.c
```diff
@@ -0,0 +1,314 @@
+/**
+ * collectd - src/ipc.c, based on src/memcached.c
+ * Copyright (C) 2010       Andres J. Diaz <ajdiaz@connectical.com>
+ * Copyright (C) 2010       Manuel L. Sanmartin <manuel.luis@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
```
mfournier: Most of the new plugins use the MIT licence. Would you consider switching to follow the trend ?
You're of course totally free to stick to the GPL if you care about the implications of this.

src/ipc.c
```diff
@@ -0,0 +1,314 @@
...
+  x |= ipc_read_msg();
+
+  return (x);
+}
+/* }}} */
+
+#ifdef KERNEL_LINUX
+static int ipc_init (void) /* {{{ */
+{
+  pagesize_g = sysconf(_SC_PAGESIZE);
```
mfournier: I don't really understand the point of this (and also the reference to `pagesize_g` on line 82), as the same statement is found on line 206.

mfournier:
Globally I find this plugin nice and useful. Postgresql users will love it!

:+1: for inclusion (barring the couple of open questions).

Can someone else take a look please ?

mfournier:
oops @ajdiaz I realize I didn't hilight you in my comments last week. Just wanted to make sure you've seen them.

src/collectd.conf.pod
```diff
@@ -2326,6 +2326,12 @@ other interfaces are collected.
...
+
+The I<IPC plugin> allows to B<collectd> to gather information about
```
jamtur01: Extra "to" - The IPC plugin allows collectd to gather...

mfournier:
Thanks a lot @ajdiaz ! The ipc plugin will be part of collectd 5.5.

ajdiaz:
Thanks for the merge @mfournier, sorry to not reply before, I was on vacation without much internet access :)

mfournier:
Thanks to you for your patience and for this nice plugin @ajdiaz :-)

FYI the small changes I added on top of your patchet: https://github.com/collectd/collectd/compare/cfe801cf0...a4cd2c372
Please complain loudly if you disagree !

I was also wondering if you'd care switching the licence to MIT (like most of the new plugins) ?

ajdiaz:
@mfournier The changes are fine for me :)
About the license, I've no problems to change the plugin license to MIT, but part of the plugin is based to busybox ipc.c, which is GPLv2, so I think that it could be better if we keep the GPL license. What do you think about that?

PR #943, opened 2015-02-17
Added support for SSL multithread to collectd (fixes #858)

### LINKED ISSUE #858: apache plugin HTTPS - multithread crash on RHES 5.7
I've deployed recently collectd  to read mod_status apache info through collectd apache plugin ( over https protocol).

Collectd was compiled  and built directly from 25 May 2014 master brach ( d76d25155399943bc9d317c6fb6d1981ca6e9b00).
Libcurl (libcurl.so.4.3.0)  was also compiled and built from official curl-7.35.0 sources.

**System:** RHEL 5.7 (Linux  2.6.18-274.el5 #1 SMP Fri Jul 8 17:36:59 EDT 2011 x86_64 x86_64 x86_64 GNU/Linux)
**OpenSSL:**  openssl-0.9.8e-27.el5_10.3

After some days running ok the process suddenly crashes  ( core dump shows some kind of error with ssl lib). 

```
Core was generated by `/opt/collectd/sbin/collectd -C /opt/collectd/etc/collectd.conf'.
Program terminated with signal 11, Segmentation fault.
#0  0x0000003d5787c39b in ?? ()
(gdb) backtrace
#0  0x0000003d5787c39b in ?? ()
#1  0x00000036efcdeb5e in SHA1_Update () from /lib64/libcrypto.so.6
#2  0x00000036efcdbd0f in ?? () from /lib64/libcrypto.so.6
#3  0x00000036efcdb5a6 in ?? () from /lib64/libcrypto.so.6
#4  0x00000036f001d351 in ssl3_client_hello () from /lib64/libssl.so.6
#5  0x00000036f001e7c9 in ssl3_connect () from /lib64/libssl.so.6
#6  0x00002aed7c3d180d in ossl_connect_common () from /opt/collectd/lib/libcurl.so.4
#7  0x00002aed7c3d273d in Curl_ssl_connect_nonblocking () from /opt/collectd/lib/libcurl.so.4
#8  0x00002aed7c398cde in https_connecting () from /opt/collectd/lib/libcurl.so.4
#9  0x00002aed7c3a5a3e in Curl_protocol_connect () from /opt/collectd/lib/libcurl.so.4
#10 0x00002aed7c3b888f in multi_runsingle () from /opt/collectd/lib/libcurl.so.4
#11 0x00002aed7c3b9895 in curl_multi_perform () from /opt/collectd/lib/libcurl.so.4
#12 0x00002aed7c3b33f4 in curl_easy_perform () from /opt/collectd/lib/libcurl.so.4
#13 0x00002aed7c189521 in apache_read_host (user_data=0xb96a768) at apache.c:615
#14 0x000000000041017f in plugin_read_thread (args=0x0) at plugin.c:462
#15 0x0000003d5840673d in ?? ()
#16 0x0000000000000000 in ?? ()
```

Anyone knows what is the origin of the problem? what versions are known to run ok with collectd (5.4.1) + apache plugin (https) + libcurl 7.35  on RHEL 5.7?

toni-moreno:
Added support for SSL multithread to collectd and initialized on the apache plugin, this PR should fix this issue (https://github.com/collectd/collectd/issues/858)


src/daemon/utils_ssl.c
```diff
@@ -0,0 +1,104 @@
...
+ **/
+
+#if HAVE_CONFIG_H
+# include "config.h"
+#endif
+
+#ifndef HAVE_OPENSSL
+int cd_ssl_multithread_setup(void)
+{
+	return 0;
```
tokkee: This should at least log a warning and/or return non-zero (in which case the calling plugin can decide what do do).

src/daemon/utils_ssl.c
```diff
@@ -0,0 +1,104 @@
...
+#include <pthread.h>
+#include <openssl/err.h>
+#include <openssl/crypto.h>
+
+#define MUTEX_TYPE       pthread_mutex_t
+#define MUTEX_SETUP(x)   pthread_mutex_init(&(x), NULL)
+#define MUTEX_CLEANUP(x) pthread_mutex_destroy(&(x))
+#define MUTEX_LOCK(x)    pthread_mutex_lock(&(x))
+#define MUTEX_UNLOCK(x)  pthread_mutex_unlock(&(x))
+#define THREAD_ID        pthread_self(  )
```
tokkee: None of these defines are needed here (they would be useful if they would have different values on certain platforms).

src/daemon/utils_ssl.c
```diff
@@ -0,0 +1,104 @@
...
+  if (!mutex_buf)
+    return 0;
+  for (i = 0;  i < CRYPTO_num_locks(  );  i++)
+    MUTEX_SETUP(mutex_buf[i]);
+  CRYPTO_set_id_callback(id_function);
+  CRYPTO_set_locking_callback(locking_function);
+  return 1;
+}
+
+int cd_ssl_multithread_cleanup(void)
```
tokkee: The first call to this function would do all the cleanup while multiple plugins might use SSL. This is not a problem if the setup and cleanup functions are only used in init and shutdown callbacks but it'll cause problems if that's not the case.

Instead, you could add a usage counter. setup() would increment it every time it's called (but then return early if the counter is greater than one) and cleanup() would decrement it and only perform the cleanup if the counter is zero. And, to be really pedantic, access to the counter should be protected by a mutex.

src/daemon/utils_ssl.c
```diff
@@ -0,0 +1,104 @@
+/**
+ * collectd - src/utils_ssl.c
+ * Copyright (C) 2007-2010  Florian octo Forster
```
tokkee: The copyright holder for this file is either you and/or Jeremy Brown.

src/daemon/utils_ssl.c
toni-moreno: Ok, I will remove ASAP.

src/daemon/utils_ssl.c
toni-moreno: OK, I will fix this issue ASAP

src/daemon/utils_ssl.c
toni-moreno: OK, I will fix ... ASAP..

src/daemon/utils_ssl.c
toni-moreno: Ok, I can change but my goal is to help collectd developers to improve this wonderful tool,

toni-moreno: fixed!

src/daemon/utils_ssl.c
toni-moreno: Fixed

src/daemon/utils_ssl.c
toni-moreno: Fixed

src/daemon/utils_ssl.c
toni-moreno: I've been thinking, and you are right  that should be properly defined where to call these two functions:

cd_ssl_multithread_setup()
cd_ssl_multithread_cleanup()

Once defined that should be always called on "init" and "shutdown"   plugin callbacks. 

It is strictly necessary to add more complexity to that? I can not imagine some situation where these two functions should be called in a multithread environment  ( in plugin_read_cb by example ?).

@tokkee  if you think it is absolutely necessary I will do, but I don't think so. 

I will wait  your answer before any other change. 

Thank you !!

octo:
Hi @toni-moreno, thank you very much for your patch! I'll leave some review comments inline …

Best regards,
—octo

src/apache.c
```diff
@@ -586,14 +587,30 @@ static int apache_init (void) /* {{{ */
 {
 	/* Call this while collectd is still single-threaded to avoid
 	 * initialization issues in libgcrypt. */
+	if(cd_ssl_multithread_setup() == SSL_MT_OK){
```
octo: If I understand the problem correctly, this needs to be called from every plugin using libcurl. I'd therefore try to keep this as short as possible, i.e. just `cd_ssl_multithread_setup()` — the function will print a warning itself and since we don't abort, we don't change the behavior by ignoring the return value.

src/daemon/utils_ssl.c
```diff
@@ -0,0 +1,100 @@
+/**
+ * collectd - src/utils_ssl.c
+ * Copyright (C) 2015  Toni Moreno
+ *
+ * This program is free software; you can redistribute it and/or modify it
```
octo: libcurl and [the example code](http://curl.haxx.se/libcurl/c/opensslthreadlock.html) are licensed under the [MIT license](https://collectd.org/wiki/index.php/Category:MIT_License). Would you mind using the same licensing terms here, too?

src/daemon/utils_ssl.c
```diff
@@ -0,0 +1,100 @@
+/**
+ * collectd - src/utils_ssl.c
+ * Copyright (C) 2015  Toni Moreno
```
octo: Please add a line with:

```
Copyright (C) 1998 - 2011, Daniel Stenberg, <daniel@haxx.se>, et al.
```

src/daemon/utils_ssl.c
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * Author:
+ *   Toni Moreno <toni.moreno at gmail.com>
+ */
+
+ /* Code taken from The example code from Jeremy Brown:
+ * http://curl.haxx.se/libcurl/c/opensslthreadlock.html
+ **/
+
+#if HAVE_CONFIG_H
```
octo: Please simply include:

```c
#include "collectd.h"
```

This header will take care of including `config.h` and a bunch of other stuff.

src/daemon/utils_ssl.c
```diff
@@ -0,0 +1,100 @@
...
+# include "config.h"
+#endif
+
+# include "plugin.h"
```
octo: It doesn't look like you're using `plugin.h` in here. Am I missing something?

src/daemon/utils_ssl.h
```diff
@@ -0,0 +1,32 @@
+/**
+ * collectd - src/utils_ssl.h
+ * Copyright (C) 2015  Florian octo Forster
```
octo: Please use the same copyright / licensing header as for `src/utils_ssl.c`.

src/apache.c
toni-moreno: Hi @octo  .

Indeed this function have to be called from every plugin that uses libgcrypt , ie openssl . 

do you prefer to insert the INFO() and WARNING() messages on the    cd_ssl_multithread_setup() ?

src/daemon/utils_ssl.c
toni-moreno: Ok . I will add in a few days

src/daemon/utils_ssl.c
toni-moreno: An an example code I think we can use the licensing that you prefer. But I'm not an expert on open licences , so I prefer delegate this question on you and select the licence that you think is better for collectd.
What do you thing ? Could you give me a hint on this?

src/daemon/utils_ssl.c
toni-moreno: ok! I will fix this.

src/daemon/utils_ssl.c
toni-moreno: ups, you are right. I can not remember what I did here.. I will review ASAP and remove it if not needed

src/daemon/utils_ssl.h
toni-moreno: Ok I will fix

toni-moreno:
Hi @octo I've done your  suggested changes and also rebased to simplify merge.

Thank you for your patience.

toni-moreno:
Hi @octo . It seems that you forgot this PR... :) no matter I'm still here to remember you.

any news about this merge?

octo:
LGTM, I think this should go in the 5.4 branch. I want to test-compile before merging though.

PR #944, opened 2015-02-17
new plugin: logtail 

darashenka:
this plugin allows processing of logfiles in "tail -f" mode with much easier configuration vs "tail" plugin.


jktjkt:
You might want to squash these commits into a usable history. There's no point in having e.g. "typos fix" commits, or commit messages such as "working version". If the history is not important, my suggestion would be to actually squash to a single commit, and do not include any merges.

darashenka:
@jktjkt: I do not know how to squash PR on github and I cannot remove this PR.
Could you close this PR and use PR#1274 instead ? 
https://github.com/collectd/collectd/pull/1274

PR #946, opened 2015-02-24
Added SSLCiphers into Instance section to select ciphers from client

toni-moreno:
Libcurl and curl itself have disabled by default some ciphers by example RC4 from libcurl > 7.34 as you can see  in the following link.

http://curl.haxx.se/mail/tracker-2014-11/0046.html

"curl" tool client can force ciphers with --ciphers option.

With this new option  "SSLCiphers" you can force also any cipher suite from the apache plugin configuration.

take as example:

````
LoadPlugin apache
<Plugin apache>

  <Instance "test_0">
    URL "https://localhost:440/server-status?auto"
    CACert "/opt/collectd/etc/ca-bundle.crt"
    SSLCiphers "RC4-SHA"
    VerifyPeer false
    VerifyHost false
  </Instance>

</Plugin apache>
````

### DIFF
--- a/src/apache.c
+++ b/src/apache.c
@@ -48,6 +48,7 @@ struct apache_s
 	_Bool verify_peer;
 	_Bool verify_host;
 	char *cacert;
+	char *ssl_ciphers;
 	char *server; /* user specific server type */
 	char *apache_buffer;
 	char apache_curl_error[CURL_ERROR_SIZE];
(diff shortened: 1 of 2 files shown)


mfournier:
LGTM :+1:

For the record, `CURLOPT_SSL_CIPHER_LIST` was added to curl 7.9 (released in september 2001).

PR #949, opened 2015-02-25
Added StoreRates support for network write plugin.

edsonmarquezani:
This is a patch I've written some time ago and posted on the mailing list, but didn't receive any feedback, so I'm sending it through the official way.

http://mailman.verplant.org/pipermail/collectd/2014-April/006164.html

I hope it suits you and help other people.

Thank you.

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -329,6 +329,8 @@ static derive_t stats_values_sent = 0;
 static derive_t stats_values_not_sent = 0;
 static pthread_mutex_t stats_lock = PTHREAD_MUTEX_INITIALIZER;
 
+static int store_rates = 0;
+
 /*
  * Private functions
  */
@@ -608,9 +610,11 @@ static int write_part_values (char **ret_buffer, int *ret_buffer_len,
 	uint16_t      pkg_num_values;
 	uint8_t      *pkg_values_types;
 	value_t      *pkg_values;
+	gauge_t      *rates = NULL;
 
 	int offset;
 	int i;
+	
 
 	num_values = vl->values_len;
 	packet_len = sizeof (part_header_t) + sizeof (uint16_t)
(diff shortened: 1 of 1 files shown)


mfournier:
This need got mentioned on IRC no later than this afternoon :)

The use case is when collectd sends metrics to other tools able to decode the network protocol (influxdb, logstash), they complain about "values constantly growing" in their graphs. Having a way to make the network plugin do the rate calculation, like all the other write plugins do by default, would be useful to them.

edsonmarquezani:
Indeed, I needed this because of Logstash. To be honest, I'm not even using it anymore, but it was motivated by Logstash.

robodendron:
Thanks for creating the PR so prompty, Edson! :+1:

@mfournier: That was me in IRC. :)

edsonmarquezani:
I should probably have included the option on the default conf file's example, but I forgot. If the patches really makes it into the upstream, I think the option should be there for documentation.

mfournier:
@untergeek, @corylanou I was wondering if you had an opinion about this, as you seem to be the ones in charge of the collectd decoder in both of these tools ? Thanks !

corylanou:
@pauldix @otoolep thoughts?

untergeek:
I don't think it's a bad idea.  If you're planning on using Logstash + Kibana for more than viewing the logs, this change makes even more sense.  Kibana already has a mode that correctly plots "constantly growing" values, so that was not an issue.

src/network.c
```diff
@@ -643,32 +647,47 @@ static int write_part_values (char **ret_buffer, int *ret_buffer_len,
...
+		{
+			pkg_values[i].gauge = htond (vl->values[i].gauge);
+		}
+		else if (store_rates == 1)
+		{
+			if (rates == NULL)
+				rates = uc_get_rate (ds, vl);
+			if (rates == NULL)
+			{
+				WARNING ("csv plugin: "
```
mfournier: This looks a copy paste mistake, can you correct this please ?

src/network.c
```diff
@@ -643,32 +647,47 @@ static int write_part_values (char **ret_buffer, int *ret_buffer_len,
...
-
-			default:
-				free (pkg_values_types);
-				free (pkg_values);
-				ERROR ("network plugin: write_part_values: "
-						"Unknown data source type: %i",
-						ds->ds[i].type);
+		if (ds->ds[i].type == DS_TYPE_GAUGE)
+		{
+			pkg_values[i].gauge = htond (vl->values[i].gauge);
```
mfournier: Nitpick: by convention, blocks containing only one statement don't take brackets (like you did on line 658 for example).

Could you please correct this one as well as those below ? Thanks !

src/network.c
```diff
@@ -3307,6 +3327,10 @@ static int network_config (oconfig_item_t *ci) /* {{{ */
       network_config_set_boolean (child, &network_config_forward);
     else if (strcasecmp ("ReportStats", child->key) == 0)
       network_config_set_boolean (child, &network_config_stats);
+    else if (strcasecmp ("StoreRates", child->key) == 0)
+    {
+      network_config_set_boolean (child, &store_rates);
```
mfournier: ditto

src/network.c
```diff
@@ -329,6 +329,8 @@ static derive_t stats_values_sent = 0;
 static derive_t stats_values_not_sent = 0;
 static pthread_mutex_t stats_lock = PTHREAD_MUTEX_INITIALIZER;
 
+static int store_rates = 0;
```
mfournier: Can you please move this to line 288, next to the other variables pulled out of the configuration file ?

mfournier:
Another comment about this: I don't think it's wise to configure that globally, for the whole plugin. It would be better to be able to toggle this setting independently, in each `<Server>` block.

octo:
@edsonmarquezani Thank you very much for your patch! Are you still working on this or should we look for someone to take this on?

@mfournier I'd be all for making this a per-server option, but at the moment the `send_buffer` is a global variable, making this a bigger change than one might expect. It'd go with the current solution for the time being and tackle the per-server restructuring separately.

src/network.c
octo: Please use `cf_util_get_boolean()` here. See also: ac73c75aed7c134c6db377a5a306d51e20356fbb

src/network.c
octo: s/int/_Bool/

daniel-ruess:
Getting this merged in would be great.

edsonmarquezani:
Hello, folks. I did this quite some time ago, more than a year by now. I actualy kept it just with me for some time before submitting it. To be honest, I can't give much attention to it right now, so, if you, guys, could manage to get someone else to take this on, I think it's better.

Thanks, and I'm sorry for that.

octo:
@edsonmarquezani thanks for your feedback, and sorry again for reacting so late. I'll see if I can find some time next week to finish this up.

Note to self: receiving end needs yet to be done.

alekseyp:
:+1:

ThomasADavis:
I've taken the liberty to create a new pull request that is updated to current git master.

See pull request #1215.

jeanpralo:
:+1:

Issue #893, opened 2015-01-14
Why are the debian patches not merged upstream?


mahmoudimus:
For example,

 https://github.com/collectd/pkg-debian/blob/master/debian/patches/bts559801_plugin_find_fix.dpatch#L36-L37

This patches collectd, hardening it against vulnerable libltdl libraries (http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2009-3736).

Here is the unpatched code on master: https://github.com/collectd/collectd/blob/3ed71d2774fcacdf0c88b3a6d81bd826a6893b32/src/daemon/plugin.c#L992-L1005 

For my own knowledge and curiousity, is this a deliberate decision or an oversight?

Happy to port the patches upstream.

mfournier:
Good question! @tokkee do you have an explanation? Thanks!

tokkee:
Sounds like an oversight on my side. I'll send a PR to merge the patch.

PR #951, opened 2015-02-27
plugin: Use strcasecmp rather than strncasecmp to find a plugin.

tokkee:
This makes collectd resistant against vulnerable copies of libltdl
(CVE-2009-3736). See <http://bugs.debian.org/559801#15> for further
details.

Thanks to @mahmoudimus for reporting this in #893


mfournier:
As this has been part of the debian package for ages, I don't think there's much point discussing it too long. It's really a matter of reducing the divergence at this point IMHO.

Thanks @tokkee !

PR #953, opened 2015-02-28
Powerdns plugin: fix logging on failed lookup

rubenk:
Don't look past the end of the array but check the boundary first.


mfournier:
Thanks a lot @rubenk ! Also cherry-picked to the collectd-5.3 bugfix branch as 7096c4c6.

PR #954, opened 2015-03-01
Setting Globals to true is not needed anymore

rubenk:
003c6e86b480 made this happen by default


mfournier:
Thanks @rubenk !

I also cherry-picked the patch to the collectd-5.4 branch as add4647f5.

PR #957, opened 2015-03-04
Fix an autoreconf warning

rubenk:
Check in the 2 empty dirs autoreconf needs.

Fixes:
+ autoreconf -vif
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I libltdl/m4
aclocal: warning: couldn't open directory 'libltdl/m4': No such file or
directory
autoreconf: configure.ac: tracing
autoreconf: configure.ac: subdirectory libltdl not present
autoreconf: configure.ac: creating directory libltdl/config
autoreconf: cannot create libltdl/config: No such file or directory

### DIFF
--- a/libltdl/config/.gitignore
+++ b/libltdl/config/.gitignore
@@ -1,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
--- a/libltdl/m4/.gitignore
+++ b/libltdl/m4/.gitignore
@@ -1,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore


mfournier:
-1 to this: the whole of `libltdl/` is out of the git tree. It's created by
`./build.sh` and removed by `./clean.sh`.

What are you trying to achieve exactly here ?

rubenk:
Yes, I wasn't planning on checking it in :)
I run autoreconf all the time, instead of ./build.sh, since that's the canonical way to bootstrap an autotools project.

mfournier:
Ok, so closing this. Please complain loudly if I misunderstood your intent !

rubenk:
Sorry if I wasn't clear. By checking 'it' in I meant the whole libltdl directory, not this patch.

mfournier:
@rubenk I wasn't clear neither I guess ;)

Having files under git control in this directory is a bad idea because of: https://github.com/collectd/collectd/blob/master/clean.sh#L18

rubenk:
Ah, understood! Never mind then.

Issue #861, opened 2014-12-15
collectd.conf still contains ${prefix} after make install


ymettier:
Hello,

```
git pull
./clean.sh
./build.sh
./configure --prefix=/some/path
make -j
make install
```

Here are some buggy lines in collectd.conf : 

```
BaseDir     "${prefix}/var/lib/collectd"
PIDFile     "${prefix}/var/run/collectd.pid"
PluginDir   "${exec_prefix}/lib/collectd"
```

Should't I see the prefix and exec_prefix variables expanded ?

(or maybe having variables in the config file can be a nice new feature - however I do not need it myself)

Tested with branch master, at commit 83f88a178b310d844db10057b0f17a5039ed6ddc.

Regards,
Yves

ymettier:
Hello,

Please merge PR #964 and close this issue.

Regards,
Yves

PR #964, opened 2015-03-09
Fixed issue #861 (collectd.conf still contains ${prefix} after make install)

ymettier:
Hello,

Please read issue #861 for the description of the problem.

Here is how this fix works : 
0/ before we do the substitions (in 2 steps), let's rename `collectd.conf.in` as `collectd.conf.in.in`.
First step : `collectd.conf.in.in` -> `collectd.conf.in`
2nd step : `collectd.conf.in` -> `collectd.conf`

1/ do the substitutions with `make` as http://www.gnu.org/software/automake/manual/html_node/Scripts.html#Scripts suggests. In fact, do only the substitutions for variables that contains paths.

```
do_subst_conf = sed -e 's,[@]datadir[@],$(datadir),g' \
       -e 's,[@]libdir[@],$(libdir),g' \
       -e 's,[@]localstatedir[@],$(localstatedir),g' \
       -e 's,[@]prefix[@],$(prefix),g' \
       -e 's,[@]exec_prefix[@],$(exec_prefix),g'

collectd.conf.in: collectd.conf.in.in Makefile
       $(do_subst_conf) < $(srcdir)/collectd.conf.in.in > collectd.conf.in
```

In the future, if you need, you can add lines to `sed` above.

2/ because there are too many substitutions to do with make (and I do not like redundancy, so I will not copy/paste all the variables), do the other substitutions with autoconf. Wait... autoconf has already ran, so we cannot do it with autoconf directly. So we do it with the generated `config.status` script, that does exactly what we want.

```
collectd.conf: collectd.conf.in Makefile
       $(top_srcdir)/config.status --file=collectd.conf
```

More details in the patch/pull request.


ymettier:

PR #965, opened 2015-03-09
powerdns: update plugin with stats for pdns 3.4.3

rubenk:


mfournier:
Thanks a lot @rubenk !

PR #967, opened 2015-03-12
Postgresql plugin: Add application name option

jakubjedelsky:
You can specify application name with AppName option for better
differentiation of your monitoring app in postgresql.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -5084,6 +5084,10 @@ Skip expired values in query output.
 Specify whether to use an SSL connection when contacting the server. The
 following modes are supported:
 
+=item B<AppName> I<application name>
+
+Specify the name of application that is connected to postgresql.
+
 =over 4
 
 =item I<disable>
--- a/src/postgresql.c
+++ b/src/postgresql.c
@@ -142,6 +142,7 @@ typedef struct {
 	char *database;
 	char *user;
 	char *password;
+	char *application_name;
 
 	char *instance;
 
@@ -265,6 +266,7 @@ static c_psql_database_t *c_psql_database_new (const char *name)
 	db->port       = NULL;
 	db->user       = NULL;
 	db->password   = NULL;
+	db->application_name = NULL;
 
 	db->instance   = sstrdup (name);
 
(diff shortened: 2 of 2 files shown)


mfournier:
Good idea! Do you know if this would be compatible with ancient versions of postgresql/libpq?

mfournier:
So it appears this parameter is only supported since postgresql 9.x. When building against libpq4 (postgresql client library shipped with 8.x), collectd crashed as soon as I mentioned `AppName` in the configuration.

```
#0  0x00007ffff67b61de in connectDBStart (conn=0x670e00) at fe-connect.c:798
#1  0x00007ffff67b63f6 in PQreset (conn=0x670e00) at fe-connect.c:2039
#2  0x00007ffff69cdc3d in c_psql_check_connection (db=0x65f490) at postgresql.c:385
#3  0x00007ffff69cf10e in c_psql_read (ud=<value optimized out>) at postgresql.c:637
#4  0x000000000040f5e3 in plugin_read_thread (args=<value optimized out>) at plugin.c:525
#5  0x00007ffff773f83d in start_thread (arg=<value optimized out>) at pthread_create.c:301
#6  0x00007ffff72b0fcd in clone () from /lib64/libc.so.6
```

mfournier:
On the other hand, it looks like libpq will set the application name to the value defined in the `PGAPPNAME` environment variable: http://www.postgresql.org/docs/9.4/static/libpq-envars.html
This would obviously be ignored by older versions of libpq.

Would this be sufficient for your use-case @jakubjedelsky ?

jakubjedelsky:
Well, I didn't check the backward compatibility :/ Env variable is perfect solution I think. Is there any file where this variables can be placed and load? Or do we need to edit the postgresql.c source again?

mfournier:
@jakubjedelsky, somewhere in the collectd initscript seems the right place for this.

Using the debian/ubuntu package, I was able to set the appname in postgresql by adding `export PGAPPNAME=foobar` to `/etc/default/collectd` (so without touching anything in the collectd source code).

Does this solution seem good enough to you ? Do you agree we can close this issue ?

PR #973, opened 2015-03-17
zfs_arc: derive hit/miss stats for ratio calculations

rkojedzinszky:
The ratio statistics collected for arc hits/misses are useless now, they should be derived.


mfournier:
@rkojedzinszky does this relate to some change in zfs ? Which version would that be ? What about older versions ?

Thanks for your patch, but please give us a bit of context :)

mfournier:
Would this by any chance be the same issue also adressed by 964b45d from #919 ? /cc @baryonix

baryonix:
As far as I can see, these are two unrelated issues. The one addressed in #919 is about the *size* of the cache, while this one corrects the way in which the hit/miss ratio is calculated.

rkojedzinszky:
Dear Marc,

The counters read by collectd are just increasing statistics counters, 
they dont get reset when being read. And as two values get divided, the 
ratio gives the ratio of the overall performance since the system has 
booted up, not since the last read. I suppose that would have more sense 
to display hit percentage for the elapsed period as opposed to the total 
system uptime.

Did i clarify my intents?

2015-03-17 15:44 időpontban Marc Fournier ezt írta:
> @rkojedzinszky [1] does this relate to some change in zfs ? Which
> version would that be ? What about older versions ?
> 
> Thanks for your patch, but please give us a bit of context :)
> 
> --
> Reply to this email directly or view it on GitHub [2].
> 
> 	*
> 
> Links:
> ------
> [1] https://github.com/rkojedzinszky
> [2] https://github.com/collectd/collectd/pull/973#issuecomment-82389692

-- 
Richard Kojedzinszky

inthecloud247:
Agreed that the current zfs arc ratio stats are unfortunately pretty useless. a derived stat would definitely make a lot more sense.

octo:
+1, but please use value_to_rate() from common.h

octo:
Err, I mean rate_to_value() of course.

octo:
Thank you very much for updating this, @rkojedzinszky! I've cherry-picked your patch into *master* as 42c95164f7c6bcc5fe831c3d9844aca0e69faf4d.

Best regards,
—octo

PR #976, opened 2015-03-19
ntpd: Don't treat normal peers as refclocks.

bjh21:
Specifically, don't try to interpret the srcadr field as a refclock
identifier unless it actually is a refclock identifier.  Now collectd
logs all stats for peers rather than skipping delay for most of them
and offset for a few.

### DIFF
--- a/src/ntpd.c
+++ b/src/ntpd.c
@@ -965,8 +965,12 @@ static int ntpd_read (void)
 
 		char peername[NI_MAXHOST];
 		uint32_t refclock_id;
+		int is_refclock;
 
 		ptr = ps + i;
+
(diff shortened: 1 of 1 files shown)


dsanader:
Hello, just stambled on this problem too.

The ntpd plugin in its current state doesn't indeed record the delay for regular peers (refclock_id is != 0 for peers in my setup).

I don't understand the bit wizardry in the ntpd_get_refclock_id function, but It seems to be just wrong.

rpv-tomsk:
I just created new PR based on yours, and it will be merged to master branch soon.

If needed, patch can be applied to collectd-5.8 too.

Thanks for reporting and working on Collectd!

PR #980, opened 2015-03-23
Upgrade spec to 5.4.2

Poil:


mfournier:
I cherry-picked your patch to the collectd-5.4 branch as 73c94f851, and also made the equivalent change to the collectd-5.3 branch. Thanks a lot !

PR #752, opened 2014-09-30
Adding Low Speed Limit 

loginator17:
By default libcurl will not abort stuck/slow connections. If stuck (for whatever reason) connection was not terminated from outside, plugin will be stuck with it. This addition expected to abort slow connections and allow plugin  to reconnect.

PR #982, opened 2015-03-25
Support specifying connection timeouts for curl and curl_json

jktjkt:
Would be nice to do the same for curl_xml, but I don't currently use that plugin.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1190,6 +1190,7 @@ Sets the plugin instance to I<Instance>.
 =item B<CACert> I<file>
 =item B<Header> I<Header>
 =item B<Post> I<Body>
+=item B<Timeout> I<Timeout in miliseconds>
 
 These options behave exactly equivalent to the appropriate options of the
 I<cURL> plugin. Please see there for a detailed description.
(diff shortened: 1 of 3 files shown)


katzj:
It's worth thinking about this also vis-a-vis #752

mfournier:
I actually had this on my todo-list, so I find very convenient that you already took care of it, thanks a lot @jktjkt :-)

As @octo mentions in #752, I think it would be event better to have the default value set to the plugin's interval. What do you think ?

If you're up for it, the apache, ascent, bind, curl_xml and nginx plugins would benefit from the equivalent change !

@katzj, FYI I've started cleaning up #752 no later than yesterday (not finished yet though). The fact it's a write plugin implementing buffering makes it a bit different than the other curl-based read-plugins, wrt to timeouts.

mfournier:
Oh, as the patch adds configuration options, it needs to go to the master branch. Can you rebase and "push -f" to update the PR please @jktjkt ? Thanks !

jktjkt:
Would you be OK with a "by default, set timeout to interval" submitted to 5.4, and a configurable timeout to master? If not, no problem, I'll be happy to retarget this just to master.

mfournier:
I'm a bit reluctant to this idea, as patches to the release branches are only supposed to fix bugs. We can reasonably argue that having plugins not timeout at all is a problem, but on the other hand setting the timeout to a pretty low value without any way to increase it could bite some users too.

Let's try to have this part of the next feature release, see if any unexpected side-effects get reported back, and consider backporting the change once we're confortable with it. Sounds reasonable ?

jktjkt:
Seems that GitHub won't allow me to change the target branch. I've undone my push now and will open a new pull request.

PR #983, opened 2015-03-26
Support specifying connection timeouts for curl and curl_json

jktjkt:
Continuing from #982.

TODO: by default, use the configured interval rather than "no timeout".

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1562,6 +1562,7 @@ URL. By default the global B<Interval> setting will be used.
 =item B<Header> I<Header>
 
 =item B<Post> I<Body>
+=item B<Timeout> I<Timeout in miliseconds>
 
 These options behave exactly equivalent to the appropriate options of the
 I<cURL> plugin. Please see there for a detailed description.
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1469,6 +1469,14 @@ plugin below on how matches are defined. If the B<MeasureResponseTime> or
 B<MeasureResponseCode> options are set to B<true>, B<Match> blocks are
 optional.
 
+=item B<Timeout> I<Timeout in miliseconds>
+
+The B<Timeout> option sets the overall timeout for each request. Make sure that
+collectd is configured with enough C<ReadThreads>, otherwise an overly long
+timeout could block other plugins. By default or when set to B<0>, a timeout
+equal to the B<Interval> is used. Prior to version 5.5.0, there was no timeout
+and requests might hang indefinitely.
+
 =back
 
 =head2 Plugin C<curl_json>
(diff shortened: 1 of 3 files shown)


mfournier:
Thanks @jktjkt !

Are you planning to work on the default timeout feature & the other curl-based plugins ? If not, I might take over. Just let me know :)

FWIW, I just submitted #985 (which should supersede #752). Similar intent, but for write_http. /cc @katzj

jktjkt:
v2:
- we're now using the interval as a default value for timeout
- fixed broken logic of the curl_json plugin where `Timeout` would only take place if specified after an `URL`

One optional suggestion is maybe using a default value of, say, `3 * interval`. That way, people can get some data from their monitoring even in case of network problems. But I don't really know what's the usual way of doing this in collectd.

jktjkt:
v3: `curl_json` actually supports adding custom intervals, so default to this per-plugin-instance interval and only use the global default when there's no override.

The `curl` plugin apparently doesn't have this `Interval` settings, so that part remains unchanged.

Sorry for noise.

mfournier:
@jktjkt thanks for the updates !

I opened another PR, adding a couple of changes on top of yours: #993. I'd love to have your feedback on the changes !

The curl_json is a bit tricky, as it's `Interval` can be set from 3 different places. I hope the way I handled the different cases makes sense.

About your `3 * Interval` suggestion, I first took a moment to properly understand the failure conditions of these curl-based plugins (without any of these timeout patches): basically as long as an HTTP request in flight, the plugins will not fire another one for the same plugin/plugin_instance, even when "it's time" according to the value of `Interval` (this will "block" just one of the `ReadThreads`, no more).

This means that, assuming you're writing to a tsdb which requires regular updates (rrdtools, graphite), you'll have "holes" in your graphs when a connection is slow or stalled, no matter what.

If it's just a slowdown, a couple of `Intervals` would get skipped, but as long as the network connection recovers, the plugin will continue to fire requests every `Interval`. Now if the connection is broken and never recovers, the plugin stupidly wait forever. This is the situation we wanted to avoid with this patchset.

So as we can see, as soon as we don't get the value back within an `Interval` we start losing datapoints, I believe the right thing to do by default is to trigger the timeout as soon as the `Interval` expires. This way in the best case we lose only one datapoint, not 3. And in the worst case, we lose n datapoints, not n+3.

Does this make sense ?

jktjkt:
@mfournier , on the other hand, suppose your interval is 10 and you're monitoring a service which suddenly tooks "long time" to respond, but manages to reply, eventually. Now the tradeoff is different; a long timeout (longer than interval) indeed means that your graphs will have holes in them, but on the other hand you'll e.g. see that at least some responses arrived and that their RTT was, say, 25 seconds.

Tradeoffs :). I can live with it both ways, this is just bikeshedding about default behavior from my side.

mfournier:
@jktjkt good point! This new default could indeed break the `MeasureResponseTime` option (which only the "curl" plugin has). People relying on it should be careful about this. I'm going to add a short warning to the docs.

mfournier:
Closing this, as #993 supersedes it now. Thanks for triggering this @jktjkt !

PR #990, opened 2015-04-01
Add stenographer plugin: https://github.com/google/stenographer

openloop:


octo:
Hey @openloop, thank you very much for your patch! I'll leave some review comments inline.

Best regards,
—octo

src/stenographer.c
```diff
@@ -0,0 +1,372 @@
+/**
+ * collectd - src/stenographer.c
+ * Copyright (C) 2015       Google
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
```
octo: The collectd code base is moving towards the ISC license (https://en.wikipedia.org/wiki/ISC_license). Would that be an option for you as well?

src/stenographer.c
```diff
@@ -0,0 +1,372 @@
...
+    sfree (st->name);
+    sfree (st->host);
+    sfree (st->cert);
+    sfree (st->key);
+    sfree (st->cacert);
+    sfree (st->stenographer_buffer);
+    if (st->curl) {
+        curl_easy_cleanup(st->curl);
+        st->curl = NULL;
+    }
```
octo: I think `st` is leaking here, i.e. the `free (st);` line is missing.

src/stenographer.c
```diff
@@ -0,0 +1,372 @@
...
+
+        memset (&ud, 0, sizeof (ud));
+        ud.data = st;
+        ud.free_func = (void *) stenographer_free;
+
+        memset (callback_name, 0, sizeof (callback_name));
+        ssnprintf (callback_name, sizeof (callback_name),
+                "stenographer/%s/%s",
+                (st->host != NULL) ? st->host : hostname_g,
+                (st->name != NULL) ? st->name : "default"),
```
octo: `st->name` is asserted to be not-NULL above.

src/stenographer.c
```diff
@@ -0,0 +1,372 @@
...
+    if (status == 0)
+    {
+        user_data_t ud;
+        char callback_name[3*DATA_MAX_NAME_LEN];
+
+        memset (&ud, 0, sizeof (ud));
+        ud.data = st;
+        ud.free_func = (void *) stenographer_free;
+
+        memset (callback_name, 0, sizeof (callback_name));
```
octo: You don't need to do this: `ssnprintf()` will make sure that the buffer is always null-terminated.

src/stenographer.c
```diff
@@ -0,0 +1,372 @@
...
+    for (i = 0; i < ci->children_num; i++)
+    {
+        oconfig_item_t *child = ci->children + i;
+
+        if (strcasecmp ("Instance", child->key) == 0)
+            config_add (child);
+        else
+            WARNING ("stenographer plugin: The configuration option "
+                    "\"%s\" is not allowed here. Did you "
+                    "forget to add an <Instance /> block "
```
octo: Kudos for a well-written error message!

src/stenographer.c
```diff
@@ -0,0 +1,372 @@
...
+    value_list_t vl = VALUE_LIST_INIT;
+
+    vl.values = &value;
+    vl.values_len = 1;
+
+    sstrncpy (vl.host, (st->host != NULL) ? st->host : hostname_g,
+            sizeof (vl.host));
+
+    sstrncpy (vl.plugin, "stenographer", sizeof (vl.plugin));
+    if (st->name != NULL)
```
octo: This is guaranteed by the config. Please change to an `assert()` if you want to demonstrate that this can't be NULL.

src/stenographer.c
```diff
@@ -0,0 +1,372 @@
...
+{
+    value_t v;
+    v.counter = c;
+    submit_value (type, type_instance, v, st);
+} /* void submit_counter */
+
+static int stenographer_read_host (user_data_t *user_data) /* {{{ */
+{
+    int  lines_num = 0;
+    int  curl_lines = 21; // number of steno stats in types.db
```
octo: You're mixing comment styles; can you please convert this to `/* … */`?

src/stenographer.c
```diff
@@ -0,0 +1,372 @@
...
+
+        char *prefix = "stenographer_";
+        int titleSize = strlen(prefix) + strlen(stat[0]) + 1;
+        char *title = (char *)malloc(titleSize);
+        if (title == NULL)
+        {
+            ERROR("stenographer plugin: prefix malloc failed.");
+            return (-1);
+        }
+        strcpy(title, prefix);
```
octo: Please use `ssnprintf (title, title_size, "%s%s", prefix, stat[0]);` instead of *strcpy(3)* / *strcat(3)*.

src/stenographer.c
```diff
@@ -0,0 +1,372 @@
...
+        {
+           lines[lines_num] = NULL;
+
+           stat_num++;
+           if (stat_num >= num_stat_fields)
+               break;
+        }
+
+        char *prefix = "stenographer_";
+        int titleSize = strlen(prefix) + strlen(stat[0]) + 1;
```
octo: s/titleSize/title_size/

src/types.db
```diff
@@ -192,6 +192,27 @@ snr			value:GAUGE:0:U
 spam_check		value:GAUGE:0:U
 spam_score		value:GAUGE:U:U
 spl			value:GAUGE:U:U
+stenographer_aged_files value:COUNTER:0:U
```
octo: This needs some work: You need to have one "type" for each data type you're handling, so good types would be "bytes", "blocks", "queries", … I'll leave some suggestions below.

src/types.db
```diff
@@ -192,6 +192,27 @@ snr			value:GAUGE:0:U
 spam_check		value:GAUGE:0:U
 spam_score		value:GAUGE:U:U
 spl			value:GAUGE:U:U
+stenographer_aged_files value:COUNTER:0:U 
+stenographer_current_files value:COUNTER:0:U
+stenographer_http_request_query_POST_bytes value:COUNTER:0:U
+stenographer_http_request_query_POST_completed value:COUNTER:0:U
+stenographer_http_request_query_POST_nanos value:COUNTER:0:U
```
octo: There is already a type `total_time_in_ms`. I would add a type `total_time_in_ns` and use it for all `…_nanos` metrics:

```
<host>/stenographer-<name>/total_time_in_ns-http_request_query_POST
<host>/stenographer-<name>/total_time_in_ns-index_base_lookup
<host>/stenographer-<name>/total_time_in_ns-index_set_lookup
<host>/stenographer-<name>/total_time_in_ns-indexfile_read
<host>/stenographer-<name>/total_time_in_ns-packet_read
<host>/stenographer-<name>/total_time_in_ns-packet_scan
```

src/types.db
```diff
@@ -192,6 +192,27 @@ snr			value:GAUGE:0:U
...
+stenographer_index_base_lookup_nanos value:COUNTER:0:U
+stenographer_index_base_lookups_finished value:COUNTER:0:U
+stenographer_index_base_lookups_started value:COUNTER:0:U
+stenographer_index_set_lookup_nanos value:COUNTER:0:U
+stenographer_index_set_lookups_finished value:COUNTER:0:U
+stenographer_index_set_lookups_started value:COUNTER:0:U
+stenographer_indexfile_current_reads value:COUNTER:0:U
```
octo: If I read `indexfile.go` correctly, this is a *gauge* value:

```go
indexCurrentReads.IncrementBy(-1)
```

I would use the `pending_operations` type for this.

src/types.db
```diff
@@ -192,6 +192,27 @@ snr			value:GAUGE:0:U
 spam_check		value:GAUGE:0:U
 spam_score		value:GAUGE:U:U
 spl			value:GAUGE:U:U
+stenographer_aged_files value:COUNTER:0:U 
+stenographer_current_files value:COUNTER:0:U
+stenographer_http_request_query_POST_bytes value:COUNTER:0:U
+stenographer_http_request_query_POST_completed value:COUNTER:0:U
+stenographer_http_request_query_POST_nanos value:COUNTER:0:U
+stenographer_index_base_lookup_nanos value:COUNTER:0:U
+stenographer_index_base_lookups_finished value:COUNTER:0:U
```
octo: For `index_{base,set}_lookups_{finished,started}` I would use the existing `total_operations` type.

src/types.db
```diff
@@ -192,6 +192,27 @@ snr			value:GAUGE:0:U
...
+stenographer_index_set_lookup_nanos value:COUNTER:0:U
+stenographer_index_set_lookups_finished value:COUNTER:0:U
+stenographer_index_set_lookups_started value:COUNTER:0:U
+stenographer_indexfile_current_reads value:COUNTER:0:U
+stenographer_indexfile_read_nanos value:COUNTER:0:U
+stenographer_indexfile_reads value:COUNTER:0:U
+stenographer_packet_read_nanos value:COUNTER:0:U
+stenographer_packet_scan_nanos value:COUNTER:0:U
+stenographer_packets_blocks_read value:COUNTER:0:U
+stenographer_packets_read value:COUNTER:0:U
```
octo: There's a `packets` type which I would use as `packets-read`.

src/types.db
```diff
@@ -192,6 +192,27 @@ snr			value:GAUGE:0:U
...
+stenographer_packets_scanned value:COUNTER:0:U
```
octo: Dito: `packets-scanned`.

src/types.db
octo: Please use `DERIVE` instead of `COUNTER`: With 64bit counters overflows are so rare that counter-resets are a much larger problem. Especially when the maximum value is unset. When using `COUNTER`, you will get *huge* spikes every time that stenographer is restarted.

src/stenographer.c
mfournier: +1

This actually yields this error:
```
stenographer.c: In function 'stenographer_read_host':
stenographer.c:345:9: error: attempt to use poisoned "strcpy"
         strcpy(title, prefix);
         ^
stenographer.c:346:9: error: attempt to use poisoned "strcat"
         strcat(title, stat[0]);
         ^
Makefile:4614: recipe for target 'stenographer_la-stenographer.lo' failed
make[3]: *** [stenographer_la-stenographer.lo] Error 1
```

... when building packages on debian/ubuntu, which default to building with `CFLAGS=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security` (cf `dpkg-buildflags(1)`)

octo:
FYI, the build currently fails with the following errors:

```
stenographer.c: In function 'config_add':
stenographer.c:185:35: error: passing argument 4 of 'plugin_register_complex_read' makes integer from pointer without a cast [-Werror]
                 /* interval  = */ NULL,
                                   ^
In file included from ./daemon/common.h:32:0,
                 from stenographer.c:21:
./daemon/plugin.h:296:5: note: expected 'cdtime_t' but argument is of type 'void *'
 int plugin_register_complex_read (const char *group, const char *name,
     ^
stenographer.c: In function 'stenographer_read_host':
stenographer.c:345:9: error: attempt to use poisoned "strcpy"
         strcpy(title, prefix);
         ^
stenographer.c:346:9: error: attempt to use poisoned "strcat"
         strcat(title, stat[0]);
         ^
```

The first is due to cce136946b879557f91183e4de58e92b81e138c8 — just change the `NULL` to `0` and you should be good. The *strcpy()* / *strcat()* uses [I already picked upon](https://github.com/collectd/collectd/pull/990#discussion_r30582149).

octo:
Stale for over one month; closing.

PR #992, opened 2015-04-02
Fix type instance sanitization in bind plugin

tkho:
Noticed this copy-paste bug. Happy to send this PR, but I'm not readily equipped to build or test this... found it while browsing the code to determine the behavior of another plugin.


mfournier:
Nice catch @tkho ! Thanks a lot !

Cherry-picked as ba51433e to the oldest release branch.

PR #993, opened 2015-04-02
Curl plugins connection timeout

mfournier:
This is a followup to #983. It changes/improves 2 things:
 - the default timeout is the effective plugin/plugin instance interval, not the global configured default.
 - it accepts `Timeout 0` which disables the timeout completely.

### DIFF
--- a/src/apache.c
+++ b/src/apache.c
@@ -53,6 +53,7 @@ struct apache_s
 	char apache_curl_error[CURL_ERROR_SIZE];
 	size_t apache_buffer_size;
 	size_t apache_buffer_fill;
+	int timeout;
 	CURL *curl;
 }; /* apache_s */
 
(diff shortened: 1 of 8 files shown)


mfournier:
@jktjkt, I already started to have a look at the other curl-based plugins. If you agree with the approach on these 2, I'll try to make the others behave the same, for the sake of consistency.

jktjkt:
I don't see a use case for deliberately configuring the monitoring to use no timeouts whatsoever. This isn't about "hey, get the result no matter what, I'm OK with any possible delay, restart as needed". It's rather about "hey curl, just pretend that a reply to that TCP packet you sent last week is going to arrive, eventually".

If you see some reason to support this behavior, I would love to know what that reason is; maybe I'm missing something obvious. If you still want this change in depite these objections, though, well, these commits appear to do what the documentation says, so +1 from me.

mfournier:
Thanks for the review @jktjkt !

I can't see any reasonable need for "no timeout" either, but you can be sure that if we completely remove this possibilty, someone will complain about it. And as it is as trivial as testing on `-1` rather than `0`, I though we might as well go for it.

PR #999, opened 2015-04-11
CGroups Plugin - Memory accounting feature

MisterG:
This commit adds the memory accounting of a cgroup.
Some custom types are needed too (types.db).


MisterG:
Please refer to #1213

PR #1001, opened 2015-04-12
write_sensu: pass literal string to ERROR() macro

mfournier:
This allows building with gcc's `-Werror=format-security` flag, which is
enforced when building debian packages.

See https://wiki.debian.org/Hardening

### DIFF
--- a/src/write_sensu.c
+++ b/src/write_sensu.c
@@ -116,7 +116,6 @@ struct sensu_host {
 static char	*sensu_tags;
 static char	**sensu_attrs;
 static size_t sensu_attrs_num;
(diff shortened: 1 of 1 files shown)


mfournier:
Not sure this is the correct fix though. I'm a bit puzzled that gcc complained in this case, as it's supposed to abort if it encounters an attempt to modify a `const`.

/cc @fabricemarie

tokkee:
This looks like the right fix to me. The other options would be ERROR(%s, alloc_err) or making alloc_err a define both of which don't look better to me.

mfournier:
Thanks for confirming this @tokkee!

Issue #996, opened 2015-04-10
write_kafka plugin only working with -f flag.


utkarshcmu:
write_kafka plugin is not working without -f flag. -f is for "Don't fork to the background.", this plugin is not working in background.

mfournier:
Thanks for reporting this @utkarshcmu. I'm a bit surprised. Can you define "not working" more precisely please? Thanks!

utkarshcmu:
When I am using the "-f" flag for running CollectD Daemon with "write_kafka" plugin enabled and configured properly, the CollectD Daemon is able to write to Kafka. But when I am not using "-f" flag, I do not see any writes happening to Kafka keeping the same configuration and running the same CollectD Daemon.

mfournier:
OK, can you try enabling the syslog plugin to see of there's any error message? Also maybe try using tcpdump/tshark/ngrep to see if there's any network traffic between collectd and kafka.

mfournier:
I can confirm this issue. Setting up a basic kafka service following https://kafka.apache.org/documentation.html#quickstart, building collectd from current master (against librdkafka 0.8.5) and simply uncommenting the sample configuration block for write_kafka is sufficient to reproduce the problem.

Using kafka's `bin/kafka-console-consumer.sh` we can see collectd data flow in when the daemon is run in the foreground, but nothing when it's run in the background. In both cases no error show up in collectd logs, only plugin: `plugin_write: Writing values via write_kafka/test.`

One difference that I noticed:

* when running in the foreground:
```
$ lsof -p $(pidof collectd) -nn
lsof: WARNING: can't stat() ext4 file system /vm/docker/aufs
      Output information may be incomplete.
COMMAND    PID USER   FD   TYPE             DEVICE SIZE/OFF     NODE NAME
collectd 22224 marc  cwd    DIR              254,0     4096   107956 /opt/collectd/var/lib/collectd
collectd 22224 marc  rtd    DIR              254,0     4096        2 /
collectd 22224 marc  txt    REG              254,0   586608     3796 /opt/collectd/sbin/collectd
collectd 22224 marc  mem    REG              254,0    47712  1364038 /lib/x86_64-linux-gnu/libnss_files-2.19.so
collectd 22224 marc  mem    REG              254,0    31784  1364069 /lib/x86_64-linux-gnu/librt-2.19.so
collectd 22224 marc  mem    REG              254,0   109144  1315765 /lib/x86_64-linux-gnu/libz.so.1.2.8
[...]

PR #1002, opened 2015-04-15
write_kafka plugin : Create kafka handles from kafka_write callback rather than the config callback

ciomaire:
This is a potential fix for issue 996. Added a new function : kafka_handle() which is called by kafka_write() which invokes rd_kafka_new() and rd_kafka_topic_new(). Access to kafka_handle() is controlled by a mutex. 

This patch also fixes the double-free error mentioned @mfournier in the patch posted in 996 yesterday.

### DIFF
--- a/src/write_kafka.c
+++ b/src/write_kafka.c
@@ -38,6 +38,7 @@
 #include <librdkafka/rdkafka.h>
 #include <pthread.h>
 #include <zlib.h>
+#include <errno.h>
 
 struct kafka_topic_context {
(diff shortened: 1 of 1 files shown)


utkarshcmu:
Are we merging this pull request sometime soon?

utkarshcmu:
@ciomaire & @mfournier , I verified that "write_kafka" works fine in the background. Issue #996 is fixed I guess then.

pyr:
:+1:

pyr:
Sorry I took so long and thanks everyone involved.

Issue #887, opened 2015-01-08
write_tsdb metric names and tags


trixpan:
Not sure if other agree, but it seems to me that the naming pattern adopted for the metrics is not making full use of OpenTSDB's tagging capabilities. 

Wouldn't it be better that instead of using:

`put df.boot.df_complex.reserved 1420724763 12345678 fqdn=machine.domain.tld`
`put cpu.0.cpu.idle 1420724763 12345678 fqdn=machine.domain.tld`

we used:

`put df.df_complex.reserved 1420724763 12345678 fqdn=machine.domain.tld mount=boot`
`put cpu.idle 1420724763 12345678 fqdn=machine.domain.tld cpu=0`

Cheers

mschenck:
It definitely would be nice to not need a _static_ set of tags, for sure.

mfournier:
@kev009 can you comment on this ?

kev009:
See https://github.com/collectd/collectd/pull/709 for a partial solution.  I had a future thought of a PCRE extractor to use filter chains to pull out a match group and stuff that in a tag but the static rule sets have served us pretty well at $dayjob (and can be generated with config management).

I'm unsure how to reconcile the two systems beyond that.  If there is rough consensus for a better method, I will implement it.

ymettier:
Hello,

I'm new to OpenTSDB, so I hope I will not write too many stupid things.

A metric in Collectd is identified as `h/p-pi/t-ti/ds` (with optional `pi` and `ti`. And usually `ds="value"` with notable exceptions of `rx` and `tx` for IO metrics).
If you want to aggregate, you will aggregate on `pi` and/or `ti` and/or on `ds`. And of course on `h`.
So, in my mind, the metric name is `p.t` with mandatory tag `fqdn=h` and optional tags `pi`, `ti` and `ds`.

Now I'm thinking as a possible implementor of this feature. Here is my problem : how to name the optional tags `pi`, `ti` and `ds` ?

I definitely think that `pi`, `ti` and `ds` are bad names. `plugin-instance`, `type-instance` and `datasource` are also bad ideas.
As @trixpan show in the issue description, for df plugin, the plugin instance should be named `mount`. For cpu, it should be named `id` or something similar.

[...]

PR #1003, opened 2015-04-17
Addon for JSON TSDB

cdesbouche-acipia:
Add new format for write_http plugin.
With this format, you can push metrics to TSDB using JSON instead of telnet.

```
<Plugin write_http>
    <Node "example">
...
        URL "opentsdb"
        Format "JSONTSDB"
...
    </Node>
</Plugin>

```

Very usefull when OpenTSDB is not on local network.

Feel free to ask questions.

### DIFF
--- a/src/write_http.c
+++ b/src/write_http.c
@@ -28,6 +28,7 @@
 #include "common.h"
 #include "utils_cache.h"
 #include "utils_format_json.h"
+#include "utils_subst.h"
 
 #if HAVE_PTHREAD_H
 # include <pthread.h>
@@ -65,6 +66,7 @@ struct wh_callback_s
 
 #define WH_FORMAT_COMMAND 0
 #define WH_FORMAT_JSON    1
+#define WH_FORMAT_JSONTSDB 2
         int format;
 
         CURL *curl;
@@ -98,6 +100,9 @@ static void wh_reset_buffer (wh_callback_t *cb)  /* {{{ */
 static int wh_send_buffer (wh_callback_t *cb) /* {{{ */
 {
         int status = 0;
+        long return_code = 0;
+
+        DEBUG ("write_http: send buffer with %i bytes", cb->send_buffer_fill);
 
         curl_easy_setopt (cb->curl, CURLOPT_POSTFIELDS, cb->send_buffer);
         status = curl_easy_perform (cb->curl);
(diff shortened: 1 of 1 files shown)


octo:
Hello Christophe, thank you very much for your patch!

I'm a bit concerned that the format you chose differs from the format used by the new [Write TSDB plugin](https://collectd.org/wiki/index.php/Plugin:Write_TSDB). Ideally, users should be able to switch from one method to the other without having to rename all of their existing data. I think the best way to do this is probably to move the `wt_format_name()` function to a separate file, `src/utils_format_tsdb.c`. Potentially, the JSON encoding could live there, just like `utils_format_json.c` and `utils_format_graphite.c`.

What do you think?

Best regards,
—octo

cdesbouche-acipia:
Hello.

This code was originally created for collectd 5.2 (we still use this branch internally) and I changed some parts of the code to match the 5.5 branch without watching new features/config files structure (should pay attention to this). We never had to rename data if we switch from telnet to http.

If I correctly understand what you wrote, the best way to follow the new config format should be to create a complete new plugin or (maybe simplier) it follows the new config file format but this new format doesn't have Format option.

Could you give me more informations about what you want?

Best regards.
C.

mfournier:
I'm not sure write_http is the best place to add support for opentsdb over http: it would be confusing for users to have to use different plugins for writing to opentsdb, depending on the protocol used. 

Also one thing I anticipate is the need for tags, as discussed in #887, which would be a mess to implement in write_http, as it's specific to opentsdb.

On the other hand, it would be a shame to reimplement/duplicate the rather mature and well tested curl/HTTP specific parts of write_http (buffering, error handling, etc).

These are really just random thought, posing more questions than anwsering the exiting ones, sorry !

(for the record, opentsdb v2/json format is described here: http://opentsdb.net/docs/build/html/api_http/put.html)

octo:
@cdesbouche-acipia I'm a bit on the fence between adding this to the *Write TSDB* plugin or to the *Write HTTP* plugin. As @mfournier points out, the former makes more sense from a user's point of view, the latter lets us reuse more existing code. I think ultimately the user experience should win. A completely new plugin is probably not the best way forward.

There is a new *Write TSDB* plugin in *collectd 5.5*. It writes values to OpenTSDB using a simple plain text protocol which looks something like this:
```
put <key> <timestamp> <values> fqdn=example.com <tags> <host tags>
```
What I'm concerned about is that the way this plugin choses the values for `<key>`, `<tags>` and `<host tags>` differs from the way your code does it. This means that people using the existing plugin and people using your code will get different entries in OpenTSDB, even if the original data was the same. That's what I mean when talking about a "different format" above.

Does this clear it up?
—octo

PR #1004, opened 2015-04-18
Add support for detailed cURL statistics

tokkee:


mfournier:
Nice ! Thanks @tokkee :)

I wonder if `MeasureResponseTime` shouldn't be flagged as deprecated (ie: raise a warning) as it looks like it's equivalent to `total_time`.

I noticed this error when building on centos5:
```
utils_curl_stats.c:138: error: 'CURLINFO_APPCONNECT_TIME' undeclared here (not in a function)
make[3]: *** [curl_la-utils_curl_stats.lo] Error 1
make[3]: Leaving directory `/tmp/tmp.tyROUZ1663/collectd-5.4.2.781.g9019586/src'
make[2]: Leaving directory `/tmp/tmp.tyROUZ1663/collectd-5.4.2.781.g9019586/src'
make[1]: *** [all] Error 2
make[1]: Leaving directory `/tmp/tmp.tyROUZ1663/collectd-5.4.2.781.g9019586/src'
```
This symbol seems to have been introduced in libcurl 7.19.0. I wouldn't spend too much energy for such ancient systems, but it would be nice that these plugins continue to build on this distro, as it's an LTS.

src/collectd.conf.pod
```diff
@@ -1415,6 +1415,92 @@ number.
...
+each page or URL queried from the curl, curl_json, or curl_xml plugins. See
+the documentation of those plugins for specific information. This section
+describes the available metrics that can be configured with each plugin. All
+options are disabled by default.
+
+See L<http://curl.haxx.se/libcurl/c/curl_easy_getinfo.html> for more details.
+
+=over 4
+
+=item B<total_time> B<true|false>
```
mfournier: Shouldn't these options rather be named `TotalTime`, etc ? To be consistent with the way options are usually named.

bzed: If you want to be able to point people to the library docs to read details about the timing informations, using a naming similar to curl makes sense imho.
CURLINFO_TOTAL_TIME being total_time is pretty obvious.

octo: Agreed, with both of you: Following cURL's naming is good, but so is following collectd's typical CamelCase convention. The two don't conflict with each other, imho.

src/utils_curl_stats.c
```diff
@@ -0,0 +1,245 @@
...
+				break;
+		if (field >= STATIC_ARRAY_SIZE (field_specs))
+		{
+			ERROR ("curl stats: Unknown field name %s", c->key);
+			free (s);
+			return NULL;
+		}
+
+		if ((c->values_num != 1)
+				|| ((c->values[0].type != OCONFIG_TYPE_STRING)
```
octo: Can we just use `cf_util_get_boolean()` here, too? I don't see the need for allowing users to use strings as well – we don't allow this in many other cases.

src/utils_curl_stats.c
```diff
@@ -0,0 +1,245 @@
...
+	size_t field;
+
+	if (s == NULL)
+		return 0;
+	if (curl == NULL)
+		return -1;
+
+	if (hostname != NULL)
+		sstrncpy (vl.host, hostname, sizeof (vl.host));
+	if (plugin != NULL)
```
octo: I don't think there is a legitimate case in which *plugin* may be `NULL`. Log and return error?

src/utils_curl_stats.c
```diff
@@ -0,0 +1,245 @@
...
+
+void curl_stats_destroy (curl_stats_t *s)
+{
+	if (s != NULL)
+		free (s);
+} /* curl_stats_destroy */
+
+int curl_stats_dispatch (curl_stats_t *s, CURL *curl,
+		const char *hostname, const char *plugin, const char *plugin_instance,
+		const char *instance_prefix)
```
octo: *instance_prefix* is currently unused. Do you have a use-case in mind or can we keep it simple and remove it?

octo:
Looks great, thanks @tokkee!

octo:
FYI, this breaks on EPEL 5 (curl-7.15.5-17.el5_9) with the following error:

```
utils_curl_stats.c:138: error: 'CURLINFO_APPCONNECT_TIME' undeclared here (not in a function)
```

src/collectd.conf.pod
tokkee: Agreed. Changed to CamelCase.

src/utils_curl_stats.c
tokkee: Ack. Done.

src/utils_curl_stats.c
tokkee: Ack. Ditto for hostname.

src/utils_curl_stats.c
tokkee: I may have had a use-case in mind but we don't need it now. Hence, dropped.

tokkee:
Thanks for the review.

I added a check for `CURLINFO_APPCONNECT_TIME` and clarified the difference between `MeasureResponseTime` and `TotalTime`.

configure.ac
```diff
@@ -1949,6 +1949,36 @@ then
 fi
 if test "x$with_libcurl" = "xyes"
 then
+	SAVE_CPPFLAGS="$CPPFLAGS"
+	SAVE_LDFLAGS="$LDFLAGS"
+	CPPFLAGS="$CPPFLAGS $with_curl_cflags"
+	LDFLAGS="$LDFLAGS $curl_curl_libs"
```
rubenk: This should be $with_curl_libs.

configure.ac
```diff
@@ -1949,6 +1949,36 @@ then
...
+	AC_CACHE_CHECK([for CURLINFO_APPCONNECT_TIME],
+		[c_cv_have_curlinfo_appconnect_time],
+		AC_LINK_IFELSE([AC_LANG_PROGRAM(
+[[[
```
rubenk: I think double quotes should suffice here.

configure.ac
```diff
@@ -1949,6 +1949,36 @@ then
...
+]]]
+			)],
+			[c_cv_have_curlinfo_appconnect_time="yes"],
+			[c_cv_have_curlinfo_appconnect_time="no"]
+		)
+	)
+	CPPFLAGS="$SAVE_CPPFLAGS"
+	LDFLAGS="$SAVE_LDFLAGS"
+fi
+AM_CONDITIONAL(BUILD_WITH_LIBCURL, test "x$with_libcurl" = "xyes")
```
rubenk: please quote the arguments.

src/collectd.conf.pod
```diff
@@ -1451,6 +1451,92 @@ number.
 
 =back
 
+=head2 cURL Statistics
+
+All cURL-based plugins support collection of generic, request-based
+statistics. These are disabled by default and can be enabled selectively for
+each page or URL queried from the curl, curl_json, or curl_xml plugins. See
+the documentation of those plugins for specific information. This section
+describes the available metrics that can be configured with each plugin. All
```
rubenk: perhaps "for each plugin" instead of "with each plugin"?

src/collectd.conf.pod
```diff
@@ -1451,6 +1451,92 @@ number.
...
+was completed.
+
+=item B<AppconnectTime> B<true|false>
+
+Time it took from the start until the SSL/SSH connect/handshake to the remote
+host was completed.
+
+=item B<PretransferTime> B<true|false>
+
+Time it took from the start until the file transfer is just about to begin.
```
rubenk: until the file transfer began.

configure.ac
tokkee: Oops. Thanks.

configure.ac
tokkee: You mean the AM_CONDITIONAL arguments? The current format is consistent with the rest of the file. If we think this should be changed, we should do it globally.

configure.ac
tokkee: Ack, done.

src/collectd.conf.pod
tokkee: Done.

src/collectd.conf.pod
tokkee: It's until just before it begins. Rephrased accordingly.

rubenk:
Thanks @tokkee!

bzed:
Thanks @tokkee & @rubenk!

PR #1006, opened 2015-04-20
Update meta_data.c

hshopeful:
In the function meta_data_get_string(), when the type mismatchs,the ERROR statement should be 
ERROR ("meta_data_get_string: Type mismatch for key `%s'", e->key);
not
ERROR ("meta_data_get_signed_int: Type mismatch for key `%s'", e->key);


mfournier:
Thanks a lot @hshopeful !

I applied your patch to the oldest bugfix branch as 15037c9f4, so that it eventually becomes part of all the other bugfix and feature branches.

PR #1008, opened 2015-04-21
fix build with librabbitmq 0.6.0

remicollet:
See
* https://github.com/alanxz/rabbitmq-c/releases
* https://github.com/alanxz/rabbitmq-c/commit/9626dd5cd5f78894f1416a1afd2d624ddd4904ae

Just a trivial fix (but perhaps not the best)

### DIFF
--- a/src/amqp.c
+++ b/src/amqp.c
@@ -304,6 +304,10 @@ static int camqp_create_exchange (camqp_config_t *conf) /* {{{ */
             /* type        = */ amqp_cstring_bytes (conf->exchange_type),
             /* passive     = */ 0,
             /* durable     = */ 0,
+#if AMQP_VERSION >= 0x00060000
+            /* auto delete = */ 0,
+            /* internal    = */ 0,
+#endif
             /* arguments   = */ argument_table);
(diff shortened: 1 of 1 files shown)


rubenk:
Looks good to me. This fixes building on Fedora Rawhide.

mfournier:
Thanks a lot @remicollet !

This doesn't introduce any regression from what I can tell.

It looks like `AMQP_VERSION` was added in rabbitmq-c v0.4.0. Current this change is part of an macro which prevents it from getting built when an older version of rabbitmq-c is found. But to be on the safe side (in case the part of the code gets refactored at some point), could you please change this to read:

```#if defined(AMQP_VERSION) && AMQP_VERSION >= 0x00060000```

Thanks !

remicollet:
done

mfournier:
Thanks a lot @remicollet ! I also cherry-picked your patchset to the collectd-5.3 branch.

remicollet:
@mfournier thanks.

> also cherry-picked your patchset to the collectd-5.3 branch.

Not in 5.4 ?

mfournier:
@remicollet, yes it will eventually get merged to 5.4 too.

PR #1014, opened 2015-04-28
improve java detection/defaults in configure script

mfournier:
This should make building the java plugin work out of the box at least on current linux distros.

Hopefully this sort of workarounds won't be necessary anymore:
* https://github.com/collectd/pkg-debian/blob/master/debian/rules#L41-L52
* http://pkgs.fedoraproject.org/cgit/collectd.git/tree/collectd.spec#n457

/cc @rubenk as this fishes up one old patch of yours.

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -2004,6 +2004,10 @@ fi
 
 # --with-java {{{
 with_java_home="$JAVA_HOME"
+if test "x$with_java_home" = "x"
+then
+	with_java_home="/usr/lib/jvm"
+fi
 with_java_vmtype="client"
 with_java_cflags=""
 with_java_libs=""
--- a/configure.in
+++ b/configure.in
@@ -2032,7 +2032,7 @@ then
 	if test -d "$with_java_home"
 	then
 		AC_MSG_CHECKING([for jni.h])
-		TMPVAR=`find "$with_java_home" -name jni.h -type f -exec 'dirname' '{}' ';' 2>/dev/null | head -n 1`
+		TMPVAR=`find -L "$with_java_home" -name jni.h -type f -exec 'dirname' '{}' ';' 2>/dev/null | head -n 1`
 		if test "x$TMPVAR" != "x"
 		then
 			AC_MSG_RESULT([found in $TMPVAR])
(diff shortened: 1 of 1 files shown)


rubenk:
Great, thanks @mfournier.

Did you test if the java plugin works after this change? I got a report that it's broken, not sure if it's related: https://bugzilla.redhat.com/1212151

mfournier:
@rubenk I didn't try running collectd with this plugin loaded, but it doesn't look like this `LD_LIBRARY_PATH` problem appears in the test/build environment I use. Here's the output of ldd on el7 when built with this patchset:

```
collectd-5.4.2.786.g5a87b2c/src/.libs/java.so:
	linux-vdso.so.1 =>  (0x00007fffa9dcc000)
	libjvm.so => /usr/lib/jvm/java/jre-abrt/lib/amd64/server/libjvm.so (0x00007f5e395b5000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f5e393a4000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f5e38fe3000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f5e38cdc000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f5e389d9000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f5e387bd000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f5e3a5f5000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f5e385a7000)
```

And the list of java-related packages installed on this platform:
```
java-1.7.0-openjdk-1.7.0.79-2.5.5.1.el7_1.x86_64
java-1.7.0-openjdk-devel-1.7.0.79-2.5.5.1.el7_1.x86_64
java-1.7.0-openjdk-headless-1.7.0.79-2.5.5.1.el7_1.x86_64
javapackages-tools-3.4.1-6.el7_0.noarch
```

rubenk:
Thanks for checking, that looks good. It might just be an issue in either Fedora or openjdk 1.8 then.

PR #1018, opened 2015-05-01
Demote read_file_contents fread failure from error to debug. Add strerror

### LINKED ISSUE #1017: Error due to race hazard: read_file_contents: Reading file "/proc/5842/stat" failed
We get these errors reported fairly regularly. This looks like a simple race-hazard: collectd opens the /proc/$pid/stats file ok but by the time it gets to read it the process has exited.

Since failure to open the file is handled _silently_ (by returning -1) I suggest that failure to read is handled the same way, e.g., just downgrade the ERROR to a DEBUG. (I'd be happy enough with downgrading to WARNING but that still seems like overkill.)

I'd also like to change the message so it _includes the actual underlying error message_.

Let me know what you'd prefer and I'll submit a PR for you.

timbunce:
Closes #1017

### DIFF
--- a/src/daemon/common.c
+++ b/src/daemon/common.c
@@ -1339,8 +1339,8 @@ ssize_t read_file_contents (const char *filename, char *buf, size_t bufsize)
 	ret = (ssize_t) fread (buf, 1, bufsize, fh);
 	if ((ret == 0) && (ferror (fh) != 0))
 	{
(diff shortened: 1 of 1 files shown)


timbunce:
Oops. Wrong branch. I'll close and redo. Sorry.

PR #1019, opened 2015-05-01
FreeBSD jail fixes

biancalana:
Currently the process and swap plugins does not work inside Jail on FreeBSD, this patch fixes that.

### DIFF
--- a/src/swap.c
+++ b/src/swap.c
@@ -157,6 +157,8 @@ static int swap_init (void) /* {{{ */
 /* #endif defined(VM_SWAPUSAGE) */
 
 #elif HAVE_LIBKVM_GETSWAPINFO
+	char errbuf[1024];
+
 	if (kvm_obj != NULL)
 	{
 		kvm_close (kvm_obj);
@@ -165,14 +167,11 @@ static int swap_init (void) /* {{{ */
 
 	kvm_pagesize = getpagesize ();
 
-	if ((kvm_obj = kvm_open (NULL, /* execfile */
-					NULL, /* corefile */
-					NULL, /* swapfile */
-					O_RDONLY, /* flags */
-					NULL)) /* errstr */
-			== NULL)
+	kvm_obj = kvm_openfiles (NULL, "/dev/null", NULL, O_RDONLY, errbuf);
+
+	if (kvm_obj == NULL)
 	{
-		ERROR ("swap plugin: kvm_open failed.");
+		ERROR ("swap plugin: kvm_open failed, %s", errbuf);
 		return (-1);
 	}
 /* #endif HAVE_LIBKVM_GETSWAPINFO */
--- a/src/swap.c
+++ b/src/swap.c
@@ -157,7 +157,7 @@ static int swap_init (void) /* {{{ */
 /* #endif defined(VM_SWAPUSAGE) */
 
 #elif HAVE_LIBKVM_GETSWAPINFO
-	char errbuf[1024];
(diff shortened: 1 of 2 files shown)


mfournier:
Thanks a lot @biancalana !

I'm not too familiar with this, can you comment on the implications of changing the 2nd parameter to `/dev/null` on non-jailed instances of collectd ? Are you sure all the subsequent uses of the kvm descriptor don't do anything which requires access to the kernel memory ?

Also, I noticed the tcpconns plugin invokes `kvm_open()` the same way the processes and swap plugins do. Do you know if this plugin suffers from the same problem ? If it does, would you mind also updating this plugin ? Thanks !

src/swap.c
```diff
@@ -165,14 +167,11 @@ static int swap_init (void) /* {{{ */
...
-					O_RDONLY, /* flags */
-					NULL)) /* errstr */
-			== NULL)
+	kvm_obj = kvm_openfiles (NULL, "/dev/null", NULL, O_RDONLY, errbuf);
+
+	if (kvm_obj == NULL)
 	{
-		ERROR ("swap plugin: kvm_open failed.");
+		ERROR ("swap plugin: kvm_open failed, %s", errbuf);
 		return (-1);
```
mfournier: Please change the message so it reads `[...] kvm_openfiles [...]`. Thanks !

src/swap.c
```diff
@@ -157,6 +157,8 @@ static int swap_init (void) /* {{{ */
 /* #endif defined(VM_SWAPUSAGE) */
 
 #elif HAVE_LIBKVM_GETSWAPINFO
+	char errbuf[1024];
```
mfournier: According to http://www.unix.com/man-page/FreeBSD/3/kvm_openfiles/ the size should be `_POSIX2_LINE_MAX` rather than 1024. Could you please correct this here as well as in the processes plugin ? tcpconns seems to already have this correct.

biancalana:
Hi @mfournier !

I have based my changes on FreeBSD ps code, that use kvm_openfiles and works inside and outside jails... Also I have collectd with that changes working on FreeBSD with and without jails, but I can speak about other BSDs.

I make additional changes as you requested.

Regards,
Alexandre

mfournier:
Great @biancalana, thanks !

I just pushed your patchset to the collectd-5.3 branch (the oldest relevant bugfix branch). It will eventually get merged to the more recent branches, and be part of the next bugfix & feature releases.

I also realised meanwhile that the tcpconns plugin doesn't support the freebsd platform. The `kvm_openfiles()` in there is for openbsd & netbsd if I'm correct. Sorry for the confusion !

biancalana:
@mfournier,

 I'm using tcpconns plugin on FreeBSD and it works inside and outside Jails. As far as I can say it has  always worked. :joy: 

Do you know when the next version will be released ?

mfournier:
@biancalana by bad, it does indeed ! :confused: 

Thanks for checking, good to know all 3 of them work inside jails !

There's no fixed date for the next releases, although the next feature release (5.5) should get out pretty soon !

biancalana:
@mfournier do you plan to merge this fixes ?

biancalana:
Actually I've seen that it was merged, but 5.5 changelog doesn't mention anything about fixes

mfournier:
@biancalana yes, as it's actually a bugfix it will be part of the next relevant bugfix release. It's in the wiki page from which the changelog is generated at release time though: https://collectd.org/wiki/index.php/Version_5.3

This is a bit confusing admitedly :-/

PR #1023, opened 2015-05-08
when --with-librdkafka arg is not a path searched by linker

ciomaire:
Building write_kafka plugin  - bugfix : the path passed via the --with-librdkafka arg was not assigned to CPPFLAGS or LDFLAGS.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3722,6 +3722,7 @@ AC_ARG_WITH(librdkafka, [AS_HELP_STRING([--with-librdkafka@<:@=PREFIX@:>@], [Pat
   then
     with_librdkafka_cppflags="-I$withval/include"
     with_librdkafka_ldflags="-L$withval/lib"
+    with_librdkafka_rpath="$withval/lib"
     with_librdkafka="yes"
   else
     with_librdkafka="$withval"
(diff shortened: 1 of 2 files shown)


mfournier:
```
$ ./configure --with-librdkafka=/opt/librdkafka-master/
[...]
$ ldd src/.libs/write_kafka.so 
        linux-vdso.so.1 (0x00007fffccaf5000)
        librdkafka.so.1 => /opt/librdkafka-master//lib/librdkafka.so.1 (0x00007fa4e8421000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fa4e81d6000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa4e7e2c000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fa4e7c0f000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fa4e79f4000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fa4e77eb000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fa4e885c000)
```

Awesome, thanks a lot @ciomaire !

PR #1024, opened 2015-05-09
/bin/sh on OSX doesn't understand -n

rubenk:
This breaks PACKAGEVERSION since the -n is left verbatim in the string.
Use the more portable printf instead


mfournier:
Nice, all 3 build shell scripts make checkbashisms happy now. Thanks a lot @rubenk !

I'll also cherry-pick this patch to the release branches.

rubenk:
Thanks!

PR #1025, opened 2015-05-10
Plug a small leak, found with AddressSanitizer

rubenk:


mfournier:
Thanks for figuring this out @rubenk !

I wonder though if this shouldn't rather be done inside the `oconfig_free()` function itself, at the bottom of `src/liboconfig/oconfig.c`.

I grepped through current master, and all calls to this function are done against local variables, most of the time just before a `return()` statement.

rubenk:
That was my first thought as well. The problem is that you'll get a double-free since that function calls itself recursively.

octo:
I've merged #1027 instead. Thanks everybody!

PR #1027, opened 2015-05-12
oconfig: fix oconfig_free to free all elements

mfournier:
The recursive nature of this function made it difficult to free the root
node of the config tree. Splitting it in 2 allows to work around this
problem.


mfournier:
This is a counter-proposition to #1025. @rubenk, can you take a look ? Thanks !

rubenk:
This is much better, thanks!

octo:
Thanks @rubenk and @mfournier!

PR #1030, opened 2015-05-20
write_redis: Correct and improve connection failure detection and logging

brianpkelly:
Previously, the plugin did not actually check if it failed to connect to the redis server and would not print an error to the log. This corrects that and also prints the error responses from the server.

### DIFF
--- a/src/write_redis.c
+++ b/src/write_redis.c
@@ -125,11 +125,11 @@ static int wr_write (const data_set_t *ds, /* {{{ */
   assert (node->conn != NULL);
   rr = redisCommand (node->conn, "ZADD %s %s %s", key, time, value);
   if (rr==NULL)
-    WARNING("ZADD command error. key:%s", key);
+    WARNING("ZADD command error. key:%s message:%s", key, node->conn->errstr);
 
   rr = redisCommand (node->conn, "SADD collectd/values %s", ident);
   if (rr==NULL)
-    WARNING("SADD command error. ident:%s", ident);
+    WARNING("SADD command error. ident:%s message:%s", ident, node->conn->errstr);
 
   pthread_mutex_unlock (&node->lock);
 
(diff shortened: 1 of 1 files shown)


mfournier:
Thanks a lot for the patch @brianpkelly ! Seems a good idea !

I don't think it's wise to skip the initial check for `NULL` though, as the various `redisConnect*` functions will all return `NULL` in some cases: https://github.com/redis/hiredis/blob/master/hiredis.c#L690-L691

I wonder if we don't need 2 checks here. One to test if `redisConnectWithTimeout()` returned `NULL` (probably some memory allocation error in libhiredis) and a second one below to check if `node->conn->err` is non-zero (actual connection problem, with a useful error message set). What do you think ? The assertion on line 125 would become superfluous in this case I guess.

brianpkelly:
@mfournier My original thought was to have the assert handle it, but checking for ```NULL``` does seem better. How about something like this?

```C
    node->conn = redisConnectWithTimeout ((char *)node->host, node->port, node->timeout);
    if (node->conn != NULL && node->conn->err)
    {
      ERROR ("write_redis plugin: Connecting to host \"%s\" (port %i) failed: %s",
          (node->host != NULL) ? node->host : "localhost",
          (node->port != 0) ? node->port : 6379,
          node->conn->errstr);
      pthread_mutex_unlock (&node->lock);
      return (-1);
    } 
    else if (conn == NULL) 
    {
      ERROR ("write_redis plugin: Connecting to host \"%s\" (port %i) failed: Reason unknown",
          (node->host != NULL) ? node->host : "localhost",
          (node->port != 0) ? node->port : 6379);
    }
```

octo:
Sounds good, though I'd slightly prefer:

```c
if (node->conn == NULL) {
  …
} else if (node->conn->err) {
  …
}
```

brianpkelly:
OK, I added a commit to reflect these changes.

mfournier:
Thanks a lot @brianpkelly ! I just pulled these changes to the collectd-5.5 branch (so the commits have different sha1's).

PR #1031, opened 2015-05-20
Plugin rrd: in synchronous file creation use lock_file

manuelluis:
To prevent the creation of the same file at the same time multiple times and obtain a corrupt rrd file.

If you delete some rrd files or have a lot of new hosts and the rrd files are not created in 10 seconds (the default interval), you may get two threads creating the same file.


octo:
Thanks for fixing this, @manuelluis! I've committed this to *collectd-5.4* as 84f5f5923d8109366b2c3caf66ad7c2d89be38bd.

PR #1033, opened 2015-05-20
Plugin process: in Linux read the number of threads from   /proc/pid/status

manuelluis:
In Linux read the number of threads from  /proc/pid/status instead of counter the number of files in
/proc/pid/task/*.

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -805,8 +805,8 @@ static int ps_read_tasks (int pid)
 	return ((count >= 1) ? count : 1);
 } /* int *ps_read_tasks */
 
-/* Read advanced virtual memory data from /proc/pid/status */
-static procstat_t *ps_read_vmem (int pid, procstat_t *ps)
+/* Read data from /proc/pid/status */
+static procstat_t *ps_read_status (int pid, procstat_t *ps)
 {
 	FILE *fh;
 	char buffer[1024];
@@ -814,6 +814,7 @@ static procstat_t *ps_read_vmem (int pid, procstat_t *ps)
 	unsigned long long lib = 0;
 	unsigned long long exe = 0;
 	unsigned long long data = 0;
+	unsigned long long threads = 0;
 	char *fields[8];
 	int numfields;
 
@@ -826,7 +827,8 @@ static procstat_t *ps_read_vmem (int pid, procstat_t *ps)
 		long long tmp;
 		char *endptr;
 
-		if (strncmp (buffer, "Vm", 2) != 0)
+		if (strncmp (buffer, "Vm", 2) != 0
+				|| strncmp (buffer, "Threads", 7) != 0)
 			continue;
 
 		numfields = strsplit (buffer, fields,
@@ -852,6 +854,10 @@ static procstat_t *ps_read_vmem (int pid, procstat_t *ps)
 			{
 				exe = tmp;
 			}
+			else if  (strncmp(buffer, "Threads", 7) == 0)
+			{
+				threads = tmp;
+			}
 		}
 	} /* while (fgets) */
 
@@ -864,6 +870,7 @@ static procstat_t *ps_read_vmem (int pid, procstat_t *ps)
 
 	ps->vmem_data = data * 1024;
 	ps->vmem_code = (exe + lib) * 1024;
+	ps->num_lwp = threads;
 
 	return (ps);
 } /* procstat_t *ps_read_vmem */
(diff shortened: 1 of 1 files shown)


src/processes.c
```diff
@@ -826,7 +827,8 @@ static procstat_t *ps_read_vmem (int pid, procstat_t *ps)
 		long long tmp;
 		char *endptr;
 
-		if (strncmp (buffer, "Vm", 2) != 0)
+		if (strncmp (buffer, "Vm", 2) != 0
+				|| strncmp (buffer, "Threads", 7) != 0)
```
octo: This should be `&&`:

> If it doesn't start with *Vm* **and** is doesn't start with *Threads*, then skip.

src/processes.c
```diff
@@ -864,6 +870,7 @@ static procstat_t *ps_read_vmem (int pid, procstat_t *ps)
 
 	ps->vmem_data = data * 1024;
 	ps->vmem_code = (exe + lib) * 1024;
+	ps->num_lwp = threads;
```
octo: Type mismatch: `unsigned long` vs. `unsigned long long`. Please change `data`, `exe`, `lib` and `threads` to `unsigned long` and change `strtoll` above to `strtoul`.

src/processes.c
```diff
@@ -1006,10 +1013,21 @@ int ps_read_process (int pid, procstat_t *ps, char *state)
...
-			ps->num_lwp = 1;
+			/* No VMem data */
+			ps->vmem_data = -1;
+			ps->vmem_code = -1;
+			ps->num_lwp  = 0;
+			DEBUG("ps_read_process: did not get vmem data for pid %i",pid);
+		}
+		if ( ps->num_lwp <= 0)
+		{
+			if ( (ps->num_lwp = ps_read_tasks (pid)) == -1 )
```
octo: Let's just remove `ps_read_tasks()` entirely.

octo:
Thanks @manuelluis! :-)

manuelluis:
I've noticed that in ``/proc/[pid]/stat`` the 20th field is the number of threads.

From ``man 5 proc``:

    (20) num_threads  %ld
              Number of threads in this process (since Linux 2.6).
              Before kernel 2.6, this field was hard coded to 0 as
              a placeholder for an earlier removed field.

The changes are more simple:

    diff --git a/src/processes.c b/src/processes.c
    index 0649eab..d2e54e0 100644
    --- a/src/processes.c
    +++ b/src/processes.c
    @@ -778,33 +778,6 @@ static void ps_submit_fork_rate (derive_t value)
     
     /* ------- additional functions for KERNEL_LINUX/HAVE_THREAD_INFO ------- */
     #if KERNEL_LINUX
    -static int ps_read_tasks (int pid)
    -{
    -	char           dirname[64];
    -	DIR           *dh;
    -	struct dirent *ent;
    -	int count = 0;
    -
    -	ssnprintf (dirname, sizeof (dirname), "/proc/%i/task", pid);
    -
    -	if ((dh = opendir (dirname)) == NULL)
    -	{
    -		DEBUG ("Failed to open directory `%s'", dirname);
    -		return (-1);
    -	}
    -
    -	while ((ent = readdir (dh)) != NULL)
    -	{
    -		if (!isdigit ((int) ent->d_name[0]))
    -			continue;
    -		else
    -			count++;
    -	}
    -	closedir (dh);
    -
    -	return ((count >= 1) ? count : 1);
    -} /* int *ps_read_tasks */
    -
     /* Read advanced virtual memory data from /proc/pid/status */
     static procstat_t *ps_read_vmem (int pid, procstat_t *ps)
     {
    @@ -1006,9 +979,10 @@ int ps_read_process (int pid, procstat_t *ps, char *state)
     	}
     	else
     	{
    -		if ( (ps->num_lwp = ps_read_tasks (pid)) == -1 )
    +		ps->num_lwp = strtoul (fields[17], NULL, /* base = */ 10);
    +		if (errno != 0 || ps->num_lwp == 0)
     		{
    -			/* returns -1 => kernel 2.4 */
    +			/* 0 => kernel 2.4 */
     			ps->num_lwp = 1;
     		}
     		ps->num_proc = 1;

octo:
@manuelluis Thank you so much for your follow-up!

That field in `stat` is used for this purpose since Linux 2.6. The *proc(5)* manpage doesn't say anything about when the *Threads* field in `status` was added. It may seem odd, but I would actually like to keep 2.4 support around for a bit. How about we try our luck with `stat` first, and use the value from `status` if it is available. Something like this:

```c
ps->num_lwp = strtoul (fields[17], NULL, /* base = */ 10);
if (ps->num_lwp == 0)
  ps->num_lwp = 1;
/* […] in ps_read_status */
if (threads != 0)
  ps->num_lwp = threads;
```

manuelluis:
Sorry for the delay, I was busy this week. 

In the 2.4 kernel it seems not  be the field ``Threads`` in ``status``: http://lxr.free-electrons.com/source/fs/proc/array.c?v=2.4.37

I don't remember if Red Hat or other distributions backported NPTL to 2.4.

If I have time I will try a RHEL4.

PR #1035, opened 2015-05-20
Allow empty Plugin blocks.

manuelluis:
If you use templates for the config file you can get a empty block plugin (by the conditions) and if you use autoload the plugin  is not loaded.


octo:
Thanks @manuelluis! I've committed this to *collectd-5.5* as b2e0cf2ea4b094a96e8096c31897bdd98e798282.

PR #1037, opened 2015-05-20
Plugin network: flush the write buffer at the specified timeout.

manuelluis:
This patch create a thread that run at the interval specified in ``ReadTime`` and flushed the network buffer if the data is older that ``MaxAge``.

This solve the following problems:
- if you have a high interval some data may remain in the buffer and the network plugin will send in the next interval if the buffer is full or until the buffer is full.
- If you send few indicators they are not sent until the buffer is full
- if you have in time (``t1``) some data of ``host/plugin-plugin_instance/type...`` (``d1``) that remains in the buffer because it's not full, in the next interval ``t2``  you may get data for the same  ``d1`` in the same packet. If in destination the data of the second interval is processed  before data of ``t1`` you lost this data.
- The same previous problem if you get data in different packets and they arrive out of order. That happen to me

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -769,6 +769,8 @@
 #		AuthFile "/etc/collectd/passwd"
 #		Interface "eth0"
 #	</Listen>
+#	ReapTime 0
+#	MaxAge 0
 #	MaxPacketSize 1452
 #
 #	# proxy setup (client and server as above):
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -4100,6 +4100,18 @@ multicast, and IPv4 and IPv6 packets. The default is to not change this value.
 That means that multicast packets will be sent with a TTL of C<1> (one) on most
 operating systems.
 
+=item B<ReapTime> I<Milliseconds>
+
+Specifies the the interval, in milliseconds, between runs of the thread to
+check if data in send buffer is older than B<MaxAge>. Set to zero to disable
+time based flushing (this is the default).
+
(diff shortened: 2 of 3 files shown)


octo:
Thanks for your patch, Manuel! I've had this problem a few times, mostly when testing / debugging and activating only a single plugin.

What do you think about making this a bit more generic: Instead of adding this to the network plugin, add it to the daemon itself? When configured, call all registered *flush* callbacks every `FlushInterval` seconds. The flush callbacks get a timeout value, which should also be configurable, e.g. `FlushTimeout`. This way we wouldn't have to re-implement the same functionality for write_http, write_graphite, …

Please let the configuration options be seconds. You can parse them with `cf_util_get_cdtime()` and users can specify `0.2` if they want the flush called every 200 milliseconds.

mfournier:
+1 for @octo's suggestion. The write_riemann plugin could definitely make use of such a feature: delivery of events before they exipre cannot be guaranteed when batching is enabled.

manuelluis:
I agree,  I'll do it the way you say.

manuelluis:
I've noticed that if ``FlushInterval`` and ``FlushTimeout`` are global variables you may want that  write_http plugin flush the data but the rrdtool plugin not (It would be the return of the I/O hell :scream:  )

Maybe ``FlushInterval`` and ``FlushTimeout`` should be options per plugin set in the LoadPlugin block. What do you think?

manuelluis:
Done in: https://github.com/collectd/collectd/pull/1044

PR #1040, opened 2015-05-22
Normalize interval

manuelluis:
ChangeLog: Add NormalizeInterval option per plugin to round the time of the collection of the metrics to the interval.

For example if you are counting the number of orders per minute you can get some strange number when you store the data in rrd, for example 6.5 orders  in the last minute ¿?

The rrdtool do a normalization of the data if the time are not normalize with the interval. This is explained here:

http://www.vandenbogaerdt.nl/rrdtool/process.php

With this patch you execute the plugins in a time that is multiple of the interval, for example with a `Interval` of `10` if the plugin will read at `1432308122.088` the patch change this to  `1432308130.000`.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -139,6 +139,12 @@ defined in this plugin. By default, this is disabled.
 
 Specifies the value of the timeout argument of the flush callback.
 
+=item B<NormalizeInterval> B<false>|B<true>
+
+When set to B<true> will normalize the time in which collect metrics for this
+plugin and the timestamp of resulting metrics as a multiple of the interval.
+The default value is B<false>.
+
 =back
 
 =item B<AutoLoadPlugin> B<false>|B<true>
@@ -264,11 +270,6 @@ to get data.
 
 This options limits the maximum value of the interval. The default value is
 B<86400>.
-
-=item B<NormalizeInterval> B<false>|B<true>
-
-When set to B<true> normalize the time in which query the read plugins
-to be a multiple of the interval. The default value is B<false>.
 
 =item B<Timeout> I<Iterations>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -265,6 +265,11 @@ to get data.
 This options limits the maximum value of the interval. The default value is
 B<86400>.
 
+=item B<NormalizeInterval> B<false>|B<true>
+
+When set to B<true> normalize the time in which query the read plugins
+to be a multiple of the interval. The default value is B<false>.
+
 =item B<Timeout> I<Iterations>
 
 Consider a value list "missing" when no update has been read or received for
--- a/src/daemon/configfile.c
+++ b/src/daemon/configfile.c
@@ -115,7 +115,8 @@ static cf_global_option_t cf_global_options[] = {
     {"CollectInternalStats", NULL, 0, "false"},
     {"PreCacheChain", NULL, 0, "PreCache"},
     {"PostCacheChain", NULL, 0, "PostCache"},
-    {"MaxReadInterval", NULL, 0, "86400"}};
+    {"MaxReadInterval", NULL, 0, "86400"}
+	{"NormalizeInterval", NULL, "false"}};
 static int cf_global_options_num = STATIC_ARRAY_SIZE(cf_global_options);
 
 static int cf_default_typesdb = 1;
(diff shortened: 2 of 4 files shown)


octo:
Hi Manuel, thanks for your patch! One problem I'm seeing with this approach is that many plugins leave the time and interval set to zero, letting `plugin_value_list_clone()` (called via `plugin_dispatch_values()`) fill in the current time.

What do you think about adding this feature to the *rrdtool* plugin? I don't know how many other systems use fixed time periods, like RRDtool, so maybe this would be the best approach. What do you think?

—octo

manuelluis:
If we modify the time in the rrdtool plugin we would be lying on the dates: If we got ``1432308122.088`` we need to change this date to the past or to the future, I'm not sure if it's a good idea touch all dates.

With this patch I'm trying to modify the time when the read function of the plugin is called. 
I'm not considering if the plugin takes much time to generate the data or the execution is delayed by other plugins. 

Maybe instead of a global option make it an option per plugin in the LoadPlugin block and change the name to reflect  that is the time when the read function of the plugin is called.

octo:
@manuelluis I'm not sure I understand your concern. You current patch would, if activated, change the timestamp on all metrics and pass the modified timestamp to all write plugins. By putting this into the RRDtool plugin, you can let the *network plugin*, for example, send the unmodified times and still get the "normalization" feature in the one plugin that actually has this behavior.

I would still like this to be optional (and disabled by default), even when implemented in the *RRDtool plugin*, if that's our misunderstanding …?

manuelluis:
My concern to change the timestamp in RRDtool plugin is that it will apply to all data even if you only want to do this for some data. Maybe we will break something like that someone is sending multiple values in the same range with different timestamps (not sure if this works now) for example.

With the patch I don't  change the timestamp of the data, I only call the read function of the plugin on this timestamp. The advantage I see in this is approach is that we can do this per plugin (in RRDtool plugin we need to add some filter for this) and It can coordinate several collectd instances to read some data at the same time. 

The problem I see with my patch is that it can be confusing to the user, because if you activate the functionality it does not mean that data with get the data with this timestamp, you get something near
and if the read function take some time to get the data you don't fixed the normalization problem.

octo:
Good point. I don't know how well this would work, though, since some plugins will talk to external daemons, which may introduce a delay. Also, reading *all* plugins at discreet intervals will likely result in thread contention, amplifying such effects.

I currently see three options:

* Add a per-plugin parameter for `LoadPlugin`, like we did for the flush intervals, to control this. If enabled, 
 * read calls will be scheduled at discreet intervals, and
 * the dispatch function will enforce a discreet *time* of metrics coming from that plugin.
* Add this to the *RRDtool* plugin, as it is the only one really needing this. Perhaps we can add a configure option to limit this behavior to a list of specific "types", e.g. "only do this for `latency` metrics".
* Add a *Match* which is doing the normalization.
 * **Pro:** user gets full control over what get's normalized and what doesn't. You can, for example, only normalize `latency` metrics from the `foo` plugin sent to `RRDtool`, everything else, e.g. all plugins sent via the network, are left unmodified.
 * **Contra:** doesn't provide the "synchronize several daemons" effect.

I'm totally undecided which of these I would like best. The first tries to minimize the changes to the *time* field, but feels a bit over-eager, the second fits the *intended* use-case very well, but is not very flexible, and the third is very flexible, but very cumbersome to configure. What's your take?

As for user-confusion: If this is a setting of the *RRDtool plugin* which is disabled by default, I think users would find this via the manpage or at the very least, if they're wondering what this option might do, they should be able to find the documentation in the manpage.

manuelluis:
I like the *Match* option, at least for my case: a query to a database that will work.

rdietric:
It is a reasonable feature for many write plugins, not only for RRDtool. I would like to have it, since it makes my pull request (#3327) almost obsolete and is the more elegant solution. In Telegraf this feature is named "round_interval". 

If the author does not want to, I could pick it up and make it a per-plugin option.

rpv-tomsk:
Please stop touching things what are simply working.
Nobody still argue me what issue you try to solve in these PRs.

As described on referenced page http://www.vandenbogaerdt.nl/rrdtool/process.php:
`On the long term it doesn't make a difference`.

manuelluis:
@rpv-tomsk  Yes, in the long term it doesn't matter, but if you have a graph with sales in the last hour quering the database in a one minute interval you didn't get real sales numbers in this intervals with rrdtool. I have this problem.

@rdietric No problem in doing it by plugin.

manuelluis:
Change the NormalizeInterval option to a plugin config option.

```
<LoadPlugin load>
  NormalizeInterval true
</LoadPlugin>
```

Also normalize the value list timestamp.

PR #1041, opened 2015-05-22
Plugin process:  in Linux count open file descriptors.

manuelluis:
Add the option CollectFileDescriptor since it's count the files in ``/proc/pid/fd/*`` and it's add some overhead.

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -180,6 +180,8 @@ typedef struct procstat_entry_s
 	derive_t io_syscr;
 	derive_t io_syscw;
 
+	unsigned long nfd;
+
 	struct procstat_entry_s *next;
 } procstat_entry_t;
 
@@ -211,11 +213,15 @@ typedef struct procstat
 	derive_t io_syscr;
 	derive_t io_syscw;
 
+	unsigned long nfd;
+
 	struct procstat   *next;
 	struct procstat_entry_s *instances;
 } procstat_t;
 
 static procstat_t *list_head_g = NULL;
+
+static _Bool ps_fd = 0;
 
 #if HAVE_THREAD_INFO
 static mach_port_t port_host_self;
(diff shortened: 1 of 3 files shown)


src/processes.c
```diff
@@ -592,6 +602,18 @@ static int ps_config (oconfig_item_t *ci)
 			ps_list_register (c->values[0].value.string,
 					c->values[1].value.string);
 		}
+		else if (strcasecmp (c->key, "CollectFileDescriptor") == 0)
+		{
+
+			if ((c->values_num != 1)
```
octo: Can you please use `cf_util_get_boolean()` here? It will do the validation and logging for you.

src/processes.c
```diff
@@ -211,12 +213,16 @@ typedef struct procstat
...
 
+	unsigned long nfd;
+
 	struct procstat   *next;
 	struct procstat_entry_s *instances;
 } procstat_t;
 
 static procstat_t *list_head_g = NULL;
 
+static _Bool ps_fd = 0;
```
octo: The use of this variable was not obvious to me. Can you please rename to `collect_nfd`? We use `collect_*` in many other plugins for variables of this purpose.

src/types.db
```diff
@@ -166,6 +166,7 @@ ps_cputime		user:DERIVE:0:U, syst:DERIVE:0:U
 ps_data			value:GAUGE:0:9223372036854775807
 ps_disk_octets		read:DERIVE:0:U, write:DERIVE:0:U
 ps_disk_ops		read:DERIVE:0:U, write:DERIVE:0:U
+ps_fd		value:GAUGE:0:9223372036854775807
```
octo: Please reuse the `file_handles` type of the (brand-new) *fhcount* plugin.

octo:
Thank you very much for your patch, Manuel! I've added a few nits, apart from this this is good to go, I think :)

—octo

mfournier:
See also #836 which does basically the same thing (maybe in a more simplistic way). I've used it for a couple of months now, and it has already been pretty useful.

I don't have a strong opinion about one vs the other implementation. But definetely +1 for such a feature !

octo:
Both look fine to me. I like the `ps_count_dir_entries()` solution in @mfournier's approach :)

manuelluis:
Sorry , I had not seen #836. It's looks fine to me.

My concern is that is enabled by default, since it's a expensive operation if you have a lot of process and open files or connections.

PR #1043, opened 2015-05-23
Filterchain: if we don't find a writer, output the available writers to Syslog

dothebart:
this superseedes https://github.com/collectd/collectd/pull/650 - reviews fixed, merges again with main.


mfournier:
As there seems to be consensus on this in #650, I included it in the collectd-5.5 branch as 809a8973 (+ added 68923fc). Many thanks @dothebart !

@octo FYI I updated the changelog in the wiki but not in the git repo.

PR #1044, opened 2015-05-23
Support for call the flush callback at regular intervals.

manuelluis:
Only configurable per plugin, for  example:

    <LoadPlugin network>
        FlushInterval 2
        FlushTimeout 1
    </LoadPlugin>

Also add support in network plugin for call the flush callback with a timeout.

It's the generalization of: https://github.com/collectd/collectd/pull/1037

### DIFF
--- a/src/daemon/configfile.c
+++ b/src/daemon/configfile.c
@@ -287,12 +287,26 @@ static int dispatch_loadplugin (const oconfig_item_t *ci)
 	/* default to the global interval set before loading this plugin */
 	memset (&ctx, 0, sizeof (ctx));
 	ctx.interval = cf_get_default_interval ();
+	ctx.flush_interval = 0;
+	ctx.flush_timeout = 0;
 
 	for (i = 0; i < ci->children_num; ++i) {
 		if (strcasecmp("Globals", ci->children[i].key) == 0)
 			cf_util_get_flag (ci->children + i, &flags, PLUGIN_FLAGS_GLOBAL);
 		else if (strcasecmp ("Interval", ci->children[i].key) == 0) {
 			if (cf_util_get_cdtime (ci->children + i, &ctx.interval) != 0) {
+				/* cf_util_get_cdtime will log an error */
+				continue;
+			}
+		}
+		else if (strcasecmp ("FlushInterval", ci->children[i].key) == 0) {
+			if (cf_util_get_cdtime (ci->children + i, &ctx.flush_interval) != 0) {
+				/* cf_util_get_cdtime will log an error */
+				continue;
+			}
+		}
+		else if (strcasecmp ("FlushTimeout", ci->children[i].key) == 0) {
+			if (cf_util_get_cdtime (ci->children + i, &ctx.flush_timeout) != 0) {
 				/* cf_util_get_cdtime will log an error */
 				continue;
 			}
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -80,6 +80,23 @@ struct write_queue_s
 	write_queue_t *next;
 };
 
+#define FF_SIMPLE  0
+#define FF_REMOVE  65535
+struct flush_func_s
+{
+	/* `flush_func_t' "inherits" from `callback_func_t'.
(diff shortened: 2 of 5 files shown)


cep21:
Hi,

What's the status of getting this into a release version?

Thanks!

src/daemon/configfile.c
```diff
@@ -287,6 +287,8 @@ static int dispatch_loadplugin (const oconfig_item_t *ci)
 	/* default to the global interval set before loading this plugin */
 	memset (&ctx, 0, sizeof (ctx));
 	ctx.interval = cf_get_default_interval ();
+	ctx.flush_interval = 0;
+	ctx.flush_timeout = 0;
 
 	for (i = 0; i < ci->children_num; ++i) {
```
octo: Can you please add the following variable here and use that instead of `ci->children[i]` below? This is how we do it almost everywhere else and this seems a good time to clean up this legacy :)

```c
oconfig_item_t *child = ci->children + i;
```

src/daemon/plugin.c
```diff
@@ -80,6 +80,23 @@ struct write_queue_s
...
+struct flush_func_s
+{
+	/* `flush_func_t' "inherits" from `callback_func_t'.
+	 * The `ff_super' member MUST be the first one in this structure! */
+#define ff_callback ff_super.cf_callback
+#define ff_udata ff_super.cf_udata
+#define ff_ctx ff_super.cf_ctx
+	callback_func_t ff_super;
+	char *ff_name;
+	int ff_type;
```
octo: It appears, that setting this to `FF_SIMPLE` is never checked for and only `FF_REMOVE` is used. Please change this to `_Bool ff_remove;` which is initially `0` (false) and set to `1` by `plugin_unregister_flush()`.

src/daemon/plugin.c
```diff
@@ -636,6 +659,167 @@ static void stop_read_threads (void)
...
+		/* Get the flush function that needs to be flushed next.
+		 * We don't need to hold "flush_lock" for the heap, but we need
+		 * to call c_heap_get_root() and pthread_cond_wait() in the
+		 * same protected block. */
+		pthread_mutex_lock (&flush_lock);
+		ff = c_heap_get_root (flush_heap);
+		if (ff == NULL)
+		{
+			pthread_cond_wait (&flush_cond, &flush_lock);
+                        pthread_mutex_unlock (&flush_lock);
```
octo: Nit: Whitespace error.

octo:
Hi @manuelluis,

thank you very much for your patch! I'm a bit concerned about copying the logic scheduling read threads – having such complicated code twice makes it quite hard to maintain.

How about (ab)using the read scheduler to schedule the flush calls as well. Something along these lines (pseudo-code):

```c
struct flush_callback_s {
  char *name;
  cdtime_t timeout;
}
typedef struct flush_callback_s flush_callback_t;

static int flush_callback (user_data_t *ud)
{
  flush_callback_t *cb = ud->data;
  return plugin_flush (cb->name, cb->timeout, /* identifier = */ NULL);
  /* if plugin_flush fails with ENOENT (or so), call unregister_read on self. */
}

int plugin_register_flush (const char *name, plugin_flush_cb callback, user_data_t *ud)
{
  create_register_callback (&list_flush, name, (void *) callback, ud));
  if (ctx.flush_interval != 0)
    plugin_register_complex_read (
        /* group     = */ "flush",
        /* name      = */ "flush/" + name,
        /* callback  = */ flush_callback,
        /* interval  = */ ctx.flush_interval,
        /* user data = */ flush_callback_t{ .name = name, .timeout = ctx.timeout; });
}
```

octo:
@manuelluis Can you create a separate CL for the changes to the *network plugin*? Thanks!

manuelluis:
The changes of the network plugin are in: https://github.com/collectd/collectd/pull/1056

manuelluis:
The suggestion of @octo done in: https://github.com/collectd/collectd/pull/1067

octo:
Thank you so much @manuelluis! Is there anything in this PR that you would like to talk about or can I close it?

manuelluis:
You can close the PR.

PR #1047, opened 2015-05-25
Fix make uninstallcheck in java bindings

rubenk:


rubenk:
Ah, damn, this is a dupe of 2cbd37356 as well.

I must say I find the model of committing straight to branches instead of to master first fairly confusing.
If these fixes would have been on master it would have saved me from doing duplicate work.

PR #1052, opened 2015-05-27
update changelog + bump release dates

### LINKED ISSUE #880: 5.5 release timing
First off thanks so much for all your hard work. We run collectd as our primary metrics collection agent on many thousands of machines at DigitalOcean and have found it be reliable and extensible.

Currently we run a custom build which is 5.4.1 plus a relatively large number of patches from master. We've also used our build as a testing ground for patches like adding I/O to the disk plugin, or support for `MemAvailable` in the memory plugin. That being said, we'd like to completely get rid of our custom build and instead use the work which has been upstreamed since the 5.4.1 release. Is there a timeline for the 5.5 release?

Additionally, DigitalOcean would be happy to sponsor any test machines you need. I personally would also be willing to help setup some kind of CI pipeline for collectd if you all are interested in that.

Thanks again!

mfournier:
Fixes #880 ;-)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-2015-05-22, Version 5.5.0
+2015-05-27, Version 5.5.0
 	* Build system: Ability to make out-of-tree builds has been fixed.
 	  Thanks to Vincent Bernat. #792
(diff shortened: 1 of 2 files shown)


cruatta:
:tada:

PR #1053, opened 2015-05-27
Added TCP plugin and tunneling to collectd-nagios

bbczeuz:
Hi,
We have a split setup (collectd push receiver on one host, Nagios-style monitoring system on another host), so we cannot easily integrate collectd results into the monitoring system. (ssh version <6.7 (as on RHEL 5-7) does not support unix socket tunneling).
Therefore I added support for tcp sockets to collectd-nagios (-t) and wrote a collectd plugin (tcpsock) that opens a listening (tcp)socket (mostly based on the unixsock plugin code).

Test (on the monitoring host):
ssh -nNTL 5576:localhost:5576 collect-host &
collectd-nagios -s 127.0.0.1:5576 -t -H <SOME_HOST> -n load/load

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5630,6 +5630,7 @@ AC_PLUGIN([target_scale],[yes],                [The scale target])
 AC_PLUGIN([target_set],  [yes],                [The set target])
 AC_PLUGIN([target_v5upgrade], [yes],           [The v5upgrade target])
 AC_PLUGIN([tcpconns],    [$plugin_tcpconns],   [TCP connection statistics])
+AC_PLUGIN([tcpsock],     [yes],                [Tcpsock communication plugin])
 AC_PLUGIN([teamspeak2],  [yes],                [TeamSpeak2 server statistics])
 AC_PLUGIN([ted],         [$plugin_ted],        [Read The Energy Detective values])
 AC_PLUGIN([thermal],     [$plugin_thermal],    [Linux ACPI thermal zone statistics])
@@ -6007,6 +6008,7 @@ Configuration:
     target_set  . . . . . $enable_target_set
     target_v5upgrade  . . $enable_target_v5upgrade
     tcpconns  . . . . . . $enable_tcpconns
+    tcpsock   . . . . . . $enable_tcpsock
     teamspeak2  . . . . . $enable_teamspeak2
     ted . . . . . . . . . $enable_ted
     thermal . . . . . . . $enable_thermal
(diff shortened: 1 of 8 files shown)


octo:
Hi @bbczeuz, thank you very much for your patch and sorry for the late reply.

The idea of having a network aware plugin that does the same job as the *Unixsock plugin* is very good. The problem I'm seeing is network security. With UNIX domain sockets, you essentially inherit the file systems permissions, making it reasonably simple to strip down permissions. For networked interaction, where even the identity of your peer is hard to establish, this is an entirely different can of worms.

I've been working on this a while back but unfortunately never got around to finish it up. The idea is to use TLS certificates to authenticate and authorize the user(s). The code currently lives in the [ff/netcmd](https://github.com/octo/collectd/commits/ff/netcmd) branch in my personal repository. Would you be interested in picking up that code and getting into a shape that can be merged? I think it mostly works, but to be perfectly honest I'm not sure after such a long time :(

Best regards,
—octo

bbczeuz:
I'm not so much into merging the TLS code, as for my current job, the machine running collectd is only running collectd and sshd and local users can be trusted.

Well, I agree, that it's easier to limit access to unix sockets, but I think that should not be collectd's job. If you need user/group based access restrictions to the socket, iptables could do the job:

iptables -A OUTPUT -m state --state ESTABLISHED,RELATED -j ACCEPT
iptables -A OUTPUT -o lo -p tcp --syn -d 127.0.0.1 --dport 5576 -m owner --gid-owner collectd-reader-group -j ACCEPT
iptables -A OUTPUT -o lo -p tcp --syn -d 127.0.0.1 --dport 5576 -j DROP

(limits connections to the socket to programs running as GID collectd-reader-group)

Downside: Needs iptables and its 'owner' module

We had some issues where iptables could not reliably detect a packet's owner. It seems to default to 'root', which catches these cases as long as the connecting process (here: ssh tunnel to collectd) is not running as root. In doubt, iptables then drops the packet.

Maybe we could implement something similar to the network plugin (server-side passwd list). But I expect that to be a far from a trivial task -> at a later stage ;)

My dream implementation:
- Use openssl to establish a TLS connection
- Server uses certificate, client checks certificate (chain+common name)
- Both use some pre shared key similar to the current solution of the network plugin (but I'd outsource all the encryption stuff to openssl (including psk))
- Use ssl session cache on both sides for a quick connection setup.

bbczeuz:
bump
(The code is running without any issues since May 2015 connecting >100 collectd and SSCServ clients to Icinga2 performing a check per metric per minute)

dago:
Maybe it would be acceptable to use this plugin as is without SSL for now as it may be useful in network restricted environments, also something like stunnel could be used in the meantime as the full implementation of certificate authentication and checks will be quite some additional work to implement.

PR #1054, opened 2015-05-27
Zone support for Solaris, forward-port of https://www.mail-archive.co…

dago:
…m/collectd@verplant.org/msg01232.html

### DIFF
--- a/src/zone.c
+++ b/src/zone.c
@@ -65,10 +65,8 @@ zone_read_procfile(char *pidstr, char *file, void *buf, size_t bufsize)
 
 	char procfile[MAX_PROCFS_PATH];
 	(void)snprintf(procfile, sizeof(procfile), "/proc/%s/%s", pidstr, file);
-	while ((fd = open(procfile, O_RDONLY)) == -1) {
-		if ((errno != EMFILE) || (errno != ENFILE)) {
-			return(1);
-		}
+	if ((fd = open(procfile, O_RDONLY)) == -1) {
+		return (1);
 	}
 
 	if (pread(fd, buf, bufsize, 0) != bufsize) {
--- a/src/zone.c
+++ b/src/zone.c
@@ -144,8 +144,7 @@ zone_scandir(DIR *procdir)
 	zone_stats_t *stats;
 /*	size_t    physmem = sysconf(_SC_PHYS_PAGES) * pagesize;*/
 
-	if (!(tree=c_avl_create((int (*)
-				 (const void *, const void *))zone_compare))) {
+	if (!(tree=c_avl_create((void *) zone_compare))) {
 		WARNING("Failed to create tree");
 		return(NULL);
 	}
--- a/src/zone.c
+++ b/src/zone.c
@@ -17,6 +17,7 @@
  *
  * Authors:
  *   Mathijs Mohlmann
+ *   Dagobert Michelsen (forward-porting)
  **/
 
 #define _BSD_SOURCE
--- a/src/zone.c
+++ b/src/zone.c
@@ -155,7 +155,8 @@ zone_scandir(DIR *procdir)
 		return(NULL);
 	}
 
-	for (rewinddir(procdir); (direntp = readdir(procdir)); ) {
+	rewinddir(procdir);
+	while ((direntp = readdir(procdir))) {
 		pidstr = direntp->d_name;
 		if (pidstr[0] == '.')	/* skip "." and ".."  */
 			continue;
--- a/src/zone.c
+++ b/src/zone.c
@@ -100,16 +100,16 @@ zone_find_stats(c_avl_tree_t *tree, zoneid_t zoneid)
 
 	if (c_avl_get(tree, (void **)&zoneid, (void **)&ret)) {
 		if (!(ret = malloc(sizeof(zone_stats_t)))) {
-			WARNING("no memory");
+			WARNING("zone plugin: no memory");
 			return(NULL);
 		}
 		if (!(key = malloc(sizeof(zoneid_t)))) {
-			WARNING("no memory");
+			WARNING("zone plugin: no memory");
 			return(NULL);
 		}
 		*key = zoneid;
 		if (c_avl_insert(tree, key, ret)) {
-			WARNING("error inserting into tree");
+			WARNING("zone plugin: error inserting into tree");
 			return(NULL);
 		}
 	}
@@ -145,7 +145,7 @@ zone_scandir(DIR *procdir)
 /*	size_t    physmem = sysconf(_SC_PHYS_PAGES) * pagesize;*/
 
 	if (!(tree=c_avl_create((void *) zone_compare))) {
-		WARNING("Failed to create tree");
+		WARNING("zone plugin: Failed to create tree");
 		return(NULL);
 	}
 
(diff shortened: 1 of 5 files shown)


src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+	value_list_t vl = VALUE_LIST_INIT;
+	value_t      values[1];
+
+	values[0].gauge = value;
+
+	vl.values = values;
+	vl.values_len = 1; /*STATIC_ARRAY_SIZE (values);*/
+	sstrncpy (vl.host, hostname_g, sizeof (vl.host));
+	sstrncpy (vl.plugin, "zone", sizeof (vl.plugin));
+ 	sstrncpy (vl.type, "zone_cpu", sizeof (vl.type));
```
octo: I'm assuming that the `FRC2PCT()` macro results in a percentage. Please use the `percent` type here if that is so. Otherwise, please use the `vcpu` type.

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+{
+	char         *pidstr;
+	pid_t         pid;
+	dirent_t     *direntp;
+	psinfo_t      psinfo;
+	c_avl_tree_t *tree;
+	zone_stats_t *stats;
+/*	size_t    physmem = sysconf(_SC_PHYS_PAGES) * pagesize;*/
+
+	if (!(tree=c_avl_create((int (*)
```
octo: Please simply cast to a `void *` instead of a function pointer cast, i.e.:

```c
c_avl_create((void *) zone_compare)
```

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+
+static zone_stats_t *
+zone_find_stats(c_avl_tree_t *tree, zoneid_t zoneid)
+{
+	zone_stats_t *ret = NULL;
+	zoneid_t     *key = NULL;
+
+	if (c_avl_get(tree, (void **)&zoneid, (void **)&ret)) {
+		if (!(ret = malloc(sizeof(zone_stats_t)))) {
+			WARNING("no memory");
```
octo: Please add a `zone plugin: ` prefix to all warning and error messages.

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+static void
+zone_submit_values(c_avl_tree_t *tree)
+{
+	char          zonename[MAX_ZONE_NAME];
+	zoneid_t     *zoneid = NULL;
+	zone_stats_t *stats  = NULL;
+
+	while (c_avl_pick (tree, (void **)&zoneid, (void **)&stats) == 0)
+	{
+		getzonenamebyid(*zoneid, zonename, MAX_ZONE_NAME-1);
```
octo: Please use `sizeof (zonename) - 1` instead of repeating the define.

I'm assuming that this will write at most `MAX_ZONE_NAME-1` bytes to `zonename`, meaning that `zonename[MAX_ZONE_NAME-1]` is still uninitialized.

* If `getzonenamebyid()` guarantees nul-termination, please pass `sizeof (zonename)` instead.
* else, please initialize either all of `zonename` or at least `zonename[sizeof (zonename) - 1]` to zero.

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+	zone_stats_t *stats;
+/*	size_t    physmem = sysconf(_SC_PHYS_PAGES) * pagesize;*/
+
+	if (!(tree=c_avl_create((int (*)
+				 (const void *, const void *))zone_compare))) {
+		WARNING("Failed to create tree");
+		return(NULL);
+	}
+
+	for (rewinddir(procdir); (direntp = readdir(procdir)); ) {
```
octo: How does this loop ever end? I see no `break` and there is no third condition …

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+		pidstr = direntp->d_name;
+		if (pidstr[0] == '.')	/* skip "." and ".."  */
+			continue;
+		pid = atoi(pidstr);
+		if (pid == 0 || pid == 2 || pid == 3)
+			continue;	/* skip sched, pageout and fsflush */
+		if (zone_read_procfile(pidstr, "psinfo", &psinfo, 
+				  sizeof(psinfo_t)) != 0)
+			continue;
+		stats = zone_find_stats(tree, psinfo.pr_zoneid);
```
octo: `zone_find_stats()` returns `NULL` on error. You have to handle this error here otherwise the following lines will crash the daemon.

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+	return(tree);
+}
+
+
+static int zone_read (void)
+{
+	DIR          *procdir;
+	c_avl_tree_t *tree;
+
+	pagesize = sysconf(_SC_PAGESIZE);
```
octo: Nit: could be done in an *init* callback.

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+	if ((procdir = opendir("/proc")) == NULL) {
+		ERROR("cannot open /proc directory\n");
+		exit(1);
```
octo: `return (-1);` – if this function returns non-zero it will be called with an exponential backoff, so the logs won't get flooded.

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	tree=zone_scandir(procdir);
+	closedir(procdir);
```
octo: Add after this line:

```c
if (tree == NULL)
  return (-1);
```

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+	return(1);
+}
+
+static int
+zone_read_procfile(char *pidstr, char *file, void *buf, size_t bufsize)
+{
+	int fd;
+
+	char procfile[MAX_PROCFS_PATH];
+	(void)snprintf(procfile, MAX_PROCFS_PATH, "/proc/%s/%s", pidstr, file);
```
octo: s/MAX_PROCFS_PATH/sizeof (procfile)/

src/zone.c
```diff
@@ -0,0 +1,191 @@
...
+	while ((fd = open(procfile, O_RDONLY)) == -1) {
+		if ((errno != EMFILE) || (errno != ENFILE)) {
```
octo: Uhm, I don't think this is the best approach: If the process or system have run out of file handles, busy-looping around an *open()* is not the best thing to do.

octo:
Hi @dago, thank you very much for picking this up! I've left some review comments inline. It's all relatively small issues that should be easily resolvable.

Best regards,
—octo

src/zone.c
dago: `readdir()` returns `NULL` if all entries have been processed, so it is comparison and iteration at the same time.

src/zone.c
dago: This is handled in 5a7b308

dago:
Hi Octo,

I think I changed all issues you mentioned. Thanks a lot for the thorough review!

Best regards -- Dago

src/zone.c
octo: Heh, you're right. Maybe a while loop would make this easier to read:

```c
rewinddir(procdir);
while ((direntp = readdir(procdir)) != NULL) {
  …
}
```

octo:
Thanks for your work @dago! :-)

dago:
This is not in 5.5.2, would it be eligible for inclusion in 5.5.3 or is another 5.6 planned soon?

rubenk:
Yes, this will be in 5.6 which we're planning to do in a month or two.

PR #1061, opened 2015-06-04
Fix header condition in users.c

dago:
The dependency is either UTMPX or UTMP or Statgrab. The inclusion above hasn't statgrab.h in the elif-chain, that means compilation fails when only libstatgrab is available. The line can safely be taken out as the same condition is checked in #L120.


octo:
Thanks @dago! I've pulled this into the *collectd-5.5* branch as fc67dcb59e0f7fa1fa73518ff9abc2f7c7e33615.

PR #1063, opened 2015-06-05
Fix compile issue when doing make install before make dist after a clean.

nberkley:
If you do a clean, configure and make you get a compile error since PACKAGE and VERSION aren't defined. Should be using PACKAGE_NAME and PACKAGE_VERSION so they're correctly defined by AC_INIT.


octo:
Thanks @nberkley! I've pulled this into the *collectd-5.5* branch as dc691c71359a11632fd0ad3882ba0cdb6d8f9cee.

PR #1067, opened 2015-06-07
Support for call the flush callback at regular intervals using the  plugin read callback.

manuelluis:
It's the version of https://github.com/collectd/collectd/pull/1044 using the  plugin read callback.

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -1625,11 +1625,11 @@ int plugin_unregister_flush (const char *name)
 		char *flush_name;
 
 		flush_name = plugin_flush_callback_name (name);
-		if (flush_name == NULL)
-			return (-1);
-
-		plugin_unregister_read(flush_name);
-		sfree(flush_name);
+		if (flush_name != NULL)
+		{
+			plugin_unregister_read(flush_name);
+			sfree(flush_name);
+		}
 	}
 
 	return plugin_unregister (list_flush, name);
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -1318,6 +1318,29 @@ static void plugin_flush_timeout_callback_free (void *data)
 	sfree(cb);
 } /* static void plugin_flush_callback_free */
 
+static char *plugin_flush_callback_name (const char *name)
+{
+	char *flush_prefix = "flush/";
+	size_t prefix_size;
+	char *flush_name;
+	size_t name_size;
+
+	prefix_size = strlen(flush_prefix);
+	name_size = strlen(name);
+
+	flush_name = malloc (sizeof(char) * (name_size + prefix_size + 1));
+	if (flush_name == NULL)
+	{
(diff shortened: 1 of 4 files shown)


manuelluis:
TODO / debate:
- ``plugin_flush`` not check for errors in the callback or missing names.
- the use of malloc and free to create: "flush/name"

octo:
@manuelluis Looks awesome, thank you! I'll give others a day or so to comment before merging though.

src/daemon/plugin.c
```diff
@@ -1295,11 +1301,99 @@ int plugin_register_write (const char *name,
...
+		flush_name = (char *) malloc (sizeof (char) *
+			(name_size + prefix_size + 1));
+		if (flush_name == NULL)
+		{
+			ERROR ("plugin_register_flush: malloc failed.");
+			plugin_unregister (list_flush, name);
+			return (-1);
+		}
+
+		sstrncpy (flush_name, flush_prefix, prefix_size + 1);
```
octo: These 12 or so lines constructing `flush_name` appear again in `plugin_unregister_flush()`. How about a function like this:?

```c
static char *plugin_flush_callback_name (char const *base_name);
```

I see the following benefits:

* Less code duplication.
* Makes it harder to break by partial code changes.
* `if { … }` blocks become shorter.
* `ERROR()` reporting when `malloc()` fails can be done in the function.

I'm not super happy with the above name, as "callback" is not a very good name for what I try to describe, but I can't think of a better alternative atm.

src/daemon/plugin.c
```diff
@@ -1295,11 +1301,99 @@ int plugin_register_write (const char *name,
...
+		sstrncpy (flush_name + prefix_size, name, name_size + 1);
+
+		cb = (flush_callback_t *)malloc(sizeof(flush_callback_t));
```
octo: Nit: the preferred style of `{m,re}alloc()` calls is:

```c
cb = malloc (sizeof (*cb));
```

I.e. remove all type information and let the compiler infer it. This makes refactorings easier and the line less cluttered.

src/daemon/plugin.c
```diff
@@ -1295,11 +1301,99 @@ int plugin_register_write (const char *name,
...
+			/* callback  = */ plugin_flush_timeout_callback,
+			/* interval  = */ ctx.flush_interval,
+			/* user data = */ &ud);
+
+		sfree(flush_name);
+		if (status != 0)
+		{
+			sfree(cb->name);
+			sfree(cb);
+			plugin_unregister (list_flush, name);
```
octo: Not sure if that's the most intuitive failure mode. If the installing the regular flush fails, letting the flush callback itself register seems more intuitive IMHO.

pyr:
Hi @manuelluis. Thanks for this, this will be of great help for `write_riemann` :-) :+1:

octo:
Thanks @manuelluis

PR #1070, opened 2015-06-09
write_redis: Increase parsability of multi-valued keys by insterting …

brianpkelly:
This PR modifies the write_redis plugin so that it inserts delimiting `|` characters into values that have multiple parts. Previously, certain values were difficult to parse (such as those collected by the load plugin) or impossible to parse (such as those collected by the disk plugin).

For instance, a value for load could look like this ` "1433876308.933760881:00.010"` corresponding to load values of `0 0.01 0`. This PR changes the plugin to produce values like this: `"1433876308.933760881:0|0.01|0"`

### DIFF
--- a/src/write_redis.c
+++ b/src/write_redis.c
@@ -62,7 +62,6 @@ static int wr_write (const data_set_t *ds, /* {{{ */
   char *value_ptr;
   int status;
   redisReply   *rr;
-  int i;
 
   status = FORMAT_VL (ident, sizeof (ident), vl);
   if (status != 0)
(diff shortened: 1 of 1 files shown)


octo:
Hi @brianpkelly, thank you very much for your patch! Would it be feasible (on the Redis side) if we used the `format_values()` function from `src/daemon/common.h` for this? It will use colons to separate "data sources", e.g. `1433876308.934:0:0.01:0` rather than pipes, which is what essentially all other plugins, including the `PUTVAL` command, are using. That would make it therefore a lot more consistent.

brianpkelly:
@octo OK, I replaced the `APPEND` macro method with `format_values()`. Looks much cleaner now.

octo:
Thanks @brianpkelly! I've merged the first two commits (5c416628a579131d57e75f217bfb7845a66fe62f, 1fc8ef7784730f6da37b3648fdeb4772d16cb4a9) but solved the third slightly different to avoid using `assert()` for return value checking (7c04f775898bef164b834e036f1698f752b12f74).

PR #1071, opened 2015-06-09
report swap sizes in bytes

marekbecka:
Information provided about swap sizes in /proc/swaps and /proc/meminfo is in kilobytes. This patch converts this values to bytes as is used to be in previous version.

### DIFF
--- a/src/swap.c
+++ b/src/swap.c
@@ -281,7 +281,8 @@ static int swap_read_separate (void) /* {{{ */
 		if (total < used)
 			continue;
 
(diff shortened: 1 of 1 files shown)


octo:
Appears to be a regression in 4f033001b6a4df92f7f0d14bd443acc2c9b9c496. I'll submit this to *collectd-5.5*.

octo:
Committed as 871b643dc18d7ffe0ba8334dc5fb222741433774.

PR #1072, opened 2015-06-09
fix types for nr_dirtied and nr_written

marekbecka:
Values in nr_dirtied and nr_written are counters. Report them as a vmpage_action rather than a vmpage_number.


octo:
Thanks @marekbecka! Pushed this to *collectd-5.4* as 39f51b526f3801ff1dd212ae431f0b22193ddb06.

PR #1076, opened 2015-06-10
Update smart.c

schuellerf:
Are you sure this doesn't memory-leak?
(I didn't have time to run valgrind but is it possible that this command is missing?)


vincentbernat:
Good catch. LGTM (missing space before parenthesis).

pyr:
Thanks @schuellerf and @vincentbernat !

mfournier:
Cherry-picked to the collectd-5.5 branch too. Thanks !

PR #1079, opened 2015-06-11
contrib/redhat/collectd.spec: added bison and flex

nfubh:
Current .spec does not work due to missing bison and flex dependencies.
This fix lets install bison and flex into the build environment as well as
define YACC and YFLAGS environment variables.

### DIFF
--- a/contrib/redhat/collectd.spec
+++ b/contrib/redhat/collectd.spec
@@ -226,7 +226,7 @@ Source:		http://collectd.org/files/%{name}-%{version}.tar.bz2
 License:	GPLv2
 Group:		System Environment/Daemons
 BuildRoot:	%{_tmppath}/%{name}-%{version}-root
-BuildRequires:	libgcrypt-devel, kernel-headers, libtool-ltdl-devel, libcap-devel
+BuildRequires:	libgcrypt-devel, kernel-headers, libtool-ltdl-devel, libcap-devel, flex, bison
 Vendor:		collectd development team <collectd@verplant.org>
 
 %if 0%{?el7:1}
@@ -1584,6 +1584,9 @@ Collectd utilities
 %else
 %define _with_zookeeper --disable-zookeeper
 %endif
+
+export YACC=bison
+export YFLAGS=-y
 
 %configure CFLAGS="%{optflags} -DLT_LAZY_OR_NOW=\"RTLD_LAZY|RTLD_GLOBAL\"" \
 	--disable-static \


rubenk:
Afaik Bison and flex are not needed when building from a tarball, only from a git checkout.

octo:
Committed as 38203b1c8d8a8e5e567d7f6783dd8b88a2487070, thanks @nfubh!

@rubenk Yep, that's how it should work. I'm assuming that you run into the mentioned issue when trying to package *master* or so … @nfubh can you confirm this?

nfubh:
Just today I've built a 5.5.0.el6 RPM from a downloaded tarball. It only
worked after applying the above changes.

nfubh:
Can attach some confirming logs tomorrow if needed.

octo:
@nfubh the exact error messages would be quite useful in figuring out what's going wrong there, thanks

mfournier:
Mmmh, this specfile is what is used for the nightly builds and I can testify it works fine with tarballs from master as well as from the release branches.

I think I'd rather have the build fail upfront than having autoconf silently run behind the scenes for some obscure reason.

@nfubh, what for a tarball did you use exactly ? Did you generate it yourself?

nfubh:
Sorry, guys, my bad, mixed up the download link. Yesterday I build the
collectd package downloaded from
https://github.com/collectd/collectd/archive/collectd-5.5.0.tar.gz and
thought it's a release version..

Now I tried again with the _real_distribution (
https://collectd.org/files/collectd-5.5.0.tar.bz2) and it worked absolutely
fine.

Anyways, I think the patch does not hurt in general and may help building
the "raw" Git releases.

mfournier:
> Anyways, I think the patch does not hurt in general and may help building the "raw" Git releases.

@rubenk what is your opinion about this ?

rubenk:
I'm not sure how this patch helps. Attaching build logs that show the failure are much appreciated.
Bison and Flex are not the only tools needed to build from a git checkout. Think autoconf, automake, libtool etc. In general it is expected of developers to have a complete build environment. Perhaps documenting the requirements for this in README or perhaps a HACKING document might help with this.

As for the patch itself, first it exports the variables instead of passing it on the configure command line. Second autoconf does a reasonable job of detecting bison, and if it fails at that we should improve the detection. Third, the YFLAGS might clash with the AM_YFLAGS in Makefile.am.

nfubh:
Ok, let's abandon this patch then. I'll try to attach the build logs if I
have time.

mfournier:
Thanks for your feedback @nfubh and @rubenk !

How to build packages out of this specfile is explained in a comment at the top of the file, including what tarball should be used. So I assume this was overlooked.

Anyways I interpret this problem as "lacking/insuffiscent documentation". I'll try to find a moment to make a "how to build custom deb/rpm packages" page in the wiki. A lot of people have been asking about packaging recently.

rubenk:
Great, thanks @mfournier

PR #1080, opened 2015-06-11
fix openldap segfault when ldapi is used

marekbecka:
If LDAPI protocol is used in configuration, e.g. URL "ldapi://", ldap_url_parse returns NULL in `ludpp->lud_host`. Strdup has undefined behaviour if NULL argument is used and segmentation fault can also occur.

### DIFF
--- a/src/openldap.c
+++ b/src/openldap.c
@@ -604,7 +604,8 @@ static int cldap_config_add (oconfig_item_t *ci) /* {{{ */
 				st->name, st->url);
 			status = -1;
 		}
-		else if (ludpp->lud_host != NULL)
+
+		if ((status == 0) && (ludpp->lud_host != NULL))
 		{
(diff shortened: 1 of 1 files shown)


octo:
LGTM, though I'd prefer to lose the `else`, i.e.:

```c
if (ldap_url_parse ()) != 0) {
  // error handling
}

st->host = (ludpp->lud_host != NULL)
    ? strdup (ludpp->lud_host)
    : NULL;
```

marekbecka:
I have updated patch. I'm not sure if conditional assignment should be used, a cldap_t structure is zeroed at cldap_config_add.

octo:
LGTM, thanks @marekbecka!

@mfournier since you assigned yourself, do you have any further comments?

mfournier:
Thanks @marekbecka ! I just merged this now to the collectd-5.5 branch,

PR #1082, opened 2015-06-15
Added Prefix Parameter to write_redis plugin

eryx12o45:
At the moment it is not possible to add further prefixes to the key when you use write_redis plugin. I've added an option "Prefix" so you can add custom prefixes if you need them.

### DIFF
--- a/src/write_redis.c
+++ b/src/write_redis.c
@@ -72,7 +72,7 @@ static int wr_write (const data_set_t *ds, /* {{{ */
     ssnprintf (key, sizeof (key), "collectd/%s", ident);
   }
   else {
-    ssnprintf (key, sizeof (key), "%s/collectd/%s", node->prefix, ident);
+    ssnprintf (key, sizeof (key), "%s/%s", node->prefix, ident);
   }
   ssnprintf (time, sizeof (time), "%.9f", CDTIME_T_TO_DOUBLE(vl->time));
 
--- a/src/write_redis.c
+++ b/src/write_redis.c
@@ -40,6 +40,7 @@ struct wr_node_s
   char *host;
   int port;
   struct timeval timeout;
+  char *prefix;
 
   redisContext *conn;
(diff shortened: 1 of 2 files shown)


src/write_redis.c
```diff
@@ -67,7 +68,12 @@ static int wr_write (const data_set_t *ds, /* {{{ */
   status = FORMAT_VL (ident, sizeof (ident), vl);
   if (status != 0)
     return (status);
-  ssnprintf (key, sizeof (key), "collectd/%s", ident);
+  if (node->prefix == NULL) {
+    ssnprintf (key, sizeof (key), "collectd/%s", ident);
+  }
+  else {
+    ssnprintf (key, sizeof (key), "%s/collectd/%s", node->prefix, ident);
```
octo: Other plugins, for example the *write_graphite plugin*, do **not** add a separator after the prefix, so users have more flexibility. I'd prefer if this would follow the same style.

octo:
Thank you very much for your patch, @eryx12o45!

Can you please add documentation for the new *Prefix* option to the *collectd.conf(5)* manual page? It is generated from `src/collectd.conf.pod`. The syntax is quite simple and documented in *perlpod(1)*.

eryx12o45:
I've added the docu and also removed the seperator between the prefix and the rest of the key as suggested.

octo:
Hi @eryx12o45 :)

Thank you very much for updating your patch, it looks great! I'd have one more request (sorry I'm mentioning this so late – I had this epiphany just now …): Let the `Prefix` default to `collectd/` and remove that string from the format in `wr_write()`. That way the default behavior doesn't change, yet users can "configure away" that part if they so wish.

Thanks and best regards,
—octo

eryx12o45:
Hi,

I changed the code as suggested. Hope this is like you've had in mind.

Best regards
- eryx

octo:
Awesome, thank you @eryx12o45!

Issue #1077, opened 2015-06-11
zone plugin: Build fails on Solaris 11


octo:
Looks like `procfs.h` is unhappy about us using a large `off_t` or something …

```
  CC       zone_la-zone.lo
In file included from /usr/include/procfs.h:26:0,
                 from zone.c:31:
/usr/include/sys/procfs.h:45:2: error: #error "Cannot use procfs in the large file compilation environment"
 #error "Cannot use procfs in the large file compilation environment"
  ^
Makefile:4576: recipe for target 'zone_la-zone.lo' failed
```

octo:
cc @dago

octo:
A quick Google search seems to suggest that this is due to `_FILE_OFFSET_BITS=64` while compiling on a 32-bit architecture.

octo:
We can likely steal this from `src/swap.c`:

```c
/* avoid swap.h error "Cannot use swapctl in the large files compilation environment" */
#if HAVE_SYS_SWAP_H && !defined(_LP64) && _FILE_OFFSET_BITS == 64
#  undef _FILE_OFFSET_BITS
#  undef _LARGEFILE64_SOURCE
#endif
```

dago:
I noticed that also yesterday and wanted to recommend to use the same construct as for swap:
  https://github.com/collectd/collectd/blob/master/configure.ac#L397-L407

octo:
SGTM, but I have a feeling that notice never gets printed – the configure script should have printed this, right? The previous `AC_CHECK_HEADERS` already contains `#undef _FILE_OFFSET_BITS` which prevents the check from failing …

octo:
Let's see how this commit works out …

dago:
We should see that the notice gets printed as 32 bit applications cannot inspect procfs for 64 bit applications: http://docs.oracle.com/cd/E18752_01/html/816-5138/advanced-10.html

octo:
Are there 64bit processes on a x86 platform?

octo:
FYI, it compiles now past the zone plugin but failes ultimately due to #978.

dago:
Yes, since Solaris 10 the kernel is 64 bit on x86 and Solaris 11 does not even offer 32 bit, but as Solaris has a 32 bit ABI for the 64 bit kernel it is still customary to run 32 bit applications.

octo:
[...]

PR #1083, opened 2015-06-15
Fix 64 bit check issue in issue #1077

dago:
Just running the preprocessor does not honor CFLAGS which is important, as CFLAGS contains the relevant -m64.


dago:
I meant, this fixes the issue from my last comment in #1077, sorry for my mixup.

octo:
Committed to *collectd-5.5* as 65ca69925947aad1307e6ab2bff8501d7c49b249. Thanks @dago!

PR #1085, opened 2015-06-16
use counter type for Innodb_buffer_pool_pages_flushed

marekbecka:
The number of buffer pool page-flush requests is provided as a counter.


octo:
Thanks @marekbecka!

PR #1087, opened 2015-06-16
add support for simple bind authentication

marekbecka:
This patch adds support for simple bind authentication, other SASL mechanisms are not supported now.

### DIFF
--- a/src/openldap.c
+++ b/src/openldap.c
@@ -38,6 +38,8 @@ struct cldap_s /* {{{ */
 {
 	char *name;
 
+	char *binddn;
+	char *password;
 	char *cacert;
 	char *host;
 	int   state;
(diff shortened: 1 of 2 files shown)


src/openldap.c
```diff
@@ -110,10 +114,19 @@ static int cldap_init_host (cldap_t *st) /* {{{ */
...
+		cred.bv_len = strlen (st->password);
+	}
+	else
+	{
+		cred.bv_val = "";
+		cred.bv_len = 0;
+	}
 
-	rc = ldap_sasl_bind_s (st->ld, NULL, NULL, &cred, NULL, NULL, NULL);
+	rc = ldap_sasl_bind_s (st->ld, st->binddn, LDAP_SASL_SIMPLE, &cred,
```
octo: Does `NULL, LDAP_SASL_SIMPLE` behave in the same way as `NULL, NULL`?

octo:
LGTM

Thanks @marekbecka! I'll leave this open for a day or so to give others a chance to comment.

src/openldap.c
marekbecka: #define LDAP_SASL_SIMPLE	((char*)0)

PR #1096, opened 2015-06-24
df plugin fails on btrfs if ReportInodes and ValuesPercentage are enabled

marekbecka:
fstatfs called on btrfs returns with undefined fields f_files and f_ffree. Read function for df plugin returns -1 if f_files is equal or below 0 and plugin stops to report values. This patch adds additional check for filesystems which don't provide inodes info.

### DIFF
--- a/src/df.c
+++ b/src/df.c
@@ -320,7 +320,7 @@ static int df_read (void)
 		}
 
 		/* inode handling */
-		if (report_inodes)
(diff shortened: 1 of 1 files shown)


octo:
Thanks @marekbecka! I've pulled this into the *collectd-5.4* branch as 305fccf949726b6dcaea65919c9adcbd342919f6.

PR #1098, opened 2015-06-25
[configure.ac] allow percona to provide libmysql

ChrisLundquist:
This allows the mysql_config tool provided by percona
to satisfy the build dependencies.
E.G.
mysql_config --libs_r
-L/usr/lib/x86_64-linux-gnu -lperconaserverclient_r -lpthread -lm -lssl -lcrypto -ldl

Notice perconaserverclient_r instead of mysqlclient_r

Previously the explicit name of libmysqlclient would prevent this
alternative from building.

I suspect this approach will also allow other alternatives such as
MariaDB to work as well.

I have tested this *only* with percona. Further testing with vanilla mysql (and perhaps even with MariaDB) is suggested.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2717,15 +2717,15 @@ then
 	then
 		with_libmysql="no ($with_mysql_config failed)"
 	else
-		AC_CHECK_LIB(mysqlclient, mysql_init,
+		AC_CHECK_LIB(c, mysql_init, # Hack to allow percona
 		 [with_libmysql="yes"],
 		 [with_libmysql="no (symbol 'mysql_init' not found)"],
-		 [$with_mysql_libs])
+		 [$with_mysql_libs]) # The appropriate percona or mysql libs are here
 
-		AC_CHECK_LIB(mysqlclient, mysql_get_server_version,
+		AC_CHECK_LIB(c, mysql_get_server_version, # Hack to allow percona
 		 [with_libmysql="yes"],
 		 [with_libmysql="no (symbol 'mysql_get_server_version' not found)"],
-		 [$with_mysql_libs])
+		 [$with_mysql_libs]) # The appropriate percona or mysql libs are here
 	fi
 fi
 if test "x$with_libmysql" = "xyes"


configure.ac
```diff
@@ -2717,15 +2717,15 @@ then
...
 	else
-		AC_CHECK_LIB(mysqlclient, mysql_init,
+		AC_CHECK_LIB(c, mysql_init, # Hack to allow percona
 		 [with_libmysql="yes"],
 		 [with_libmysql="no (symbol 'mysql_init' not found)"],
-		 [$with_mysql_libs])
+		 [$with_mysql_libs]) # The appropriate percona or mysql libs are here
 
-		AC_CHECK_LIB(mysqlclient, mysql_get_server_version,
+		AC_CHECK_LIB(c, mysql_get_server_version, # Hack to allow percona
```
ChrisLundquist: If someone knows a better way to do this please let me know!

configure.ac
```diff
@@ -2717,15 +2717,15 @@ then
 	then
 		with_libmysql="no ($with_mysql_config failed)"
 	else
-		AC_CHECK_LIB(mysqlclient, mysql_init,
+		AC_CHECK_LIB(c, mysql_init, # Hack to allow percona
```
ChrisLundquist: Looking at this today http://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Libraries.html mentions `AC_SEARCH_LIBS` which might be a better option here.

rubenk:
What OS are you on? On my CentOS box I have the Percona-Server-shared-compat rpm installed which contains libmysqlclient.so libraries.

ChrisLundquist:
Ubuntu 14.04. I didn't see a Debian package like that, but I wasn't looking
for it.

ChrisLundquist:
Yeah, https://www.percona.com/doc/percona-server/5.6/installation.html doesn't mention a similar debian library. To me, it still makes sense to use the configuration given to us by mysql_config as it's a tool to solve just this problem.

rubenk:
Could you give the patches in #1184 a try?

mfournier:
Superseded and fixed by #1184. Thanks for contributing to fix this @ChrisLundquist ! Please let us know if you still encounter issues building against percona.

ChrisLundquist:
Thanks!

PR #1104, opened 2015-06-30
swrite: Check for closed peer connection before send

sfthomas:
The issue is described here: http://stackoverflow.com/a/11552492/1824548
And I got the solution from here: http://stefan.buettcher.org/cs/conn_closed.html

I have been running two docker containers, one with Graphite and another with collectd on two separate vbox machines (using TCP). When restarting the Graphite instance (docker stop && docker start) I noticed that collectd would stop sending data - it just sits there for like an hour before it realizes (if it realizes) that the connection is gone.
Weirdly when using netcat instead of the Graphite server (outside of the Graphite docker container), collectd worked as expected (detected the disconnect), so it must be something with docker - not closing the connected sockets, just killing the peer process silently.
But even then, on the collectd container the connnection goes into CLOSE_WAIT, indicating that it's the client (collectd) that should close the connection, so I searched around a bit, and came up with a solution to detect and resolve this scenario.
This PR captures the idea at least - given that I am no C coder, it is likely lacking style/error handling/etc...

### DIFF
--- a/src/daemon/common.c
+++ b/src/daemon/common.c
@@ -49,6 +49,8 @@
 #include <sys/socket.h>
 #include <netdb.h>
 
+#include <poll.h>
+
 #if HAVE_NETINET_IN_H
 # include <netinet/in.h>
 #endif
@@ -270,9 +272,23 @@ ssize_t swrite (int fd, const void *buf, size_t count)
 	const char *ptr;
 	size_t      nleft;
 	ssize_t     status;
+	struct      pollfd pfd;
 
 	ptr   = (const char *) buf;
 	nleft = count;
+	
+	/* checking for closed peer connection */
+	pfd.fd = fd;
+	pfd.events = POLLIN | POLLHUP | POLLRDNORM;
+	pfd.revents = 0;
+	if (poll(&pfd, 1, 0) > 0) {
+		char buffer[32];
+		if (recv(fd, buffer, sizeof(buffer), MSG_PEEK | MSG_DONTWAIT) == 0) {
+			// if recv returns zero (even though poll() said there is data to be read),
+			// that means the connection has been closed
+			return -1;
+		}
+	}
 
 	while (nleft > 0)
 	{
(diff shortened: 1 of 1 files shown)


mfournier:
FWIW, it looks like `POLLRDNORM` isn't supported on older linux distros:
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../src -I../../src -DPREFIX=\"/opt/collectd\" -DCONFIGFILE=\"/opt/collectd/etc/collectd.conf\" -DLOCALSTATEDIR=\"/opt/collectd/var\" -DPKGLOCALSTATEDIR=\"/opt/collectd/var/lib/collectd\" -DPIDFILE=\"/opt/collectd/var/run/collectd.pid\" -DPLUGINDIR=\"/opt/collectd/lib/collectd\" -DPKGDATADIR=\"/opt/collectd/share/collectd\" "-DLT_LAZY_OR_NOW=RTLD_LAZY|RTLD_GLOBAL" -Wall -Werror -O2 -g -m64 -mtune=generic -MT common.lo -MD -MP -MF .deps/common.Tpo -c common.c  -fPIC -DPIC -o .libs/common.o
common.c: In function 'swrite':
common.c:282: error: 'POLLRDNORM' undeclared (first use in this function)
common.c:282: error: (Each undeclared identifier is reported only once
common.c:282: error: for each function it appears in.)
make[3]: *** [common.lo] Error 1
```

Other than this detail, the error you describe reminds me of #922.

sfthomas:
Right, accoring to poll(2) manpage, it is some kind of an extra - and equivalent to POLLIN anyway, so I removed it.
It does resemble #922, although when I was checking the return codes from 'write', they were 0 - that as I understand means no error, as if the system did not realized that the socket is (half-) closed + collectd did not hang for me. Hence I am not entirely sure if that timeout would solve this problem.

bwhaley:
+1, I was about halfway in to implementing this same fix when I came across this PR. @sfthomas's fix worked for me.

octo:
Hi @sfthomas, thank you very much for your fix and sorry for the late reply! I've pulled your changes into the *collectd-5.4* branch as 42a62db99c96d77fd6dc676b76705d8d90156c23 and de6a0cf70cf56af705d63bdc02e9c9d0dfb14b21.

Best regards,
—octo

PR #1105, opened 2015-06-30
Add notifications to write http

mdubbyap:
Added notifications to write_http, they're behind NotifyLevel with the default being off and you can select which level of notification you want to send along.  Format for command is standard plaintext PUTNOTIF command, format for json is as below:

``` json
{
  "time": 1435698639.932,
  "severity": "FAILURE",
  "host": "localhost",
  "plugin": "exec",
  "plugin_instance": "exec_instance",
  "type": "example",
  "type_instance": "example_instance",
  "message": "my message"
}
```


mfournier:
Many thanks for your patch @mdubbyap ! This would be a valuable addition to this plugin I think !

Currently it fails to build with the following error when building with the default build options on debian/ubuntu (cf `dpkg-buildflags`):
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I./daemon -DPREFIX=\"/opt/collectd\" -DCONFIGFILE=\"/opt/collectd/etc/collectd.conf\" -DLOCALSTATEDIR=\"/opt/collectd/var\" -DPKGLOCALSTATEDIR=\"/opt/collectd/var/lib/collectd\" -DPIDFILE=\"/opt/collectd/var/run/collectd.pid\" -DPLUGINDIR=\"/opt/collectd/lib/collectd\" -DPKGDATADIR=\"/opt/collectd/share/collectd\" -Wall -Werror -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wall -Wno-error=deprecated-declarations -MT write_http_la-write_http.lo -MD -MP -MF .deps/write_http_la-write_http.Tpo -c write_http.c  -fPIC -DPIC -o .libs/write_http_la-write_http.o
write_http.c: In function 'wh_notification_command':
write_http.c:683:3: error: attempt to use poisoned "strcpy"
   strcpy(temp, value); \
   ^
write_http.c:714:3: error: attempt to use poisoned "strcpy"
   strcpy(temp, meta->nm_value.nm_string); \
   ^
Makefile:4491: recipe for target 'write_http_la-write_http.lo' failed
make[3]: *** [write_http_la-write_http.lo] Error 1
```

This is usually easily fixable by using `sstrncpy()` from `common.h` instead.

mdubbyap:
Weird, i didn't run into that but i'll fix it up.  Thanks!

matthew.

mdubbyap:
I fixed your concern last week and brought the branch up to date with master today. Let me know if there are any other concerns.

matthew.

BCWilsonDotCom:
+1 This is a great addition. Let's get this merged.

mdubbyap:
Looks like bringing it up-to-date from master has brought with it build failures... will continue to sync till it goes green again...

src/write_http.c
```diff
@@ -489,6 +490,376 @@ static int wh_write (const data_set_t *ds, const value_list_t *vl, /* {{{ */
         return (status);
 } /* }}} int wh_write */
 
+static int notification_meta_data_to_json(char *buffer, size_t buffer_size, /* {{{ */
```
octo: Can you please move this code to `src/utils_format_json.c` so other plugins can benefit from it?

octo:
@mdubbyap Thank you very much for your patch!

I've pushed a new test to *master* which surfaces a bug on some i386 systems and I haven't been able to get to the bottom of this yet (#1139). I think you're running into that issue, too. I'll likely roll back the test for now until I find a way to work around the issue.

Best regards,
—octo

src/write_http.c
mdubbyap: @octo done.

BCWilsonDotCom:
With @octo 's test now removed, does this PR build successfully again?

mdubbyap:
Does for me, just waiting for their CI system to build it after i made the
refactoring changes @octo wanted.

octo:
Hi @mdubbyap, thank you so much for your work and sorry this is moving so slow!

The code looks good overall. What I'm a bit concerned about is how parsers of the JSON output can (reliably) figure out whether an entry in the array is a *value list* vs. a *notification*. Does it make sense to mix *value lists* and *notifications* in the first place? Maybe it would be better to fire them off in a separate POST request (other plugins usually don't buffer *notifications*), potentially with a `X-Collectd: Notification` header … Last but not least: it might be useful to configure separate URLs for *value lists* and *notifications*, similar to the *exec plugin's* `NotificationExec` config option.

What do you think? I'm honestly undecided on these issues and would love to hear about people's use cases.

mdubbyap:
Thanks for the feedback.

Our downstream from this parses the JSON in such a way that we were
ignoring the new notifications in the stream before we started parsing them
for this change and then doing "the right thing" with each type.  I didn't
have any problem differentiating since the mandatory fields are different,
but i see what you're saying.  Separate urls are probably the right way to
go so people like us can use the same url if we want, but if others want to
send to different endpoints they're able to.  I'll go ahead and make this
change.

I'm on the fence about single posts for each notification (no buffering).
I agree it's a bit different paradigm from existing plugins, but i feel
like the existing egress endpoints like syslog, network or exec are very
lightweight while http each time (particularly https) seems like perhaps
too much overhead since it's all going to be happening on the edge of the
interval not spread out during it.  Thoughts?   @BCWilsonDotCom as someone
else who is looking to use this mod, do you have an opinion here?

mdubbyap:
Hrm, configuration for a separate EventURL seems like it will be unwieldy.
Do all the other arguments then apply to both, or do we duplicate them?
Perhaps specifying NotifyLevel means that that Node only sends events?  And
one configures two Nodes as endpoints? If you want more complicated setups
you can use chains?  Your input is appreciated.

BCWilsonDotCom:
I think single posts for each notification is a good idea. I wouldn't want any chance of a critical event notification sitting in the buffer.

mdubbyap:
Ok, @octo  i've updated the pull request with what i put above.  No longer buffering (a big deal when you separate the channels) and specifying a NotifyLevel makes that Node a notification sender not a metric sender.  Gives people the opportunity to send both to the same endpoint, or different.  I agree this is better and i hope you think the implementation is correct.  Thanks!

BCWilsonDotCom:
Just touching base on this. Any further issues, suggestions, or objections?

mdubbyap:
@octo  any word on this?  Any changes you would like i'm happy to make.  Thanks!

BCWilsonDotCom:
@octo any thoughts? thanks.

bviswana101:
@mdubbyap, very nice addition, will test it out. Would like to see this merged into master.

ciomaire:
Looks like invalid json is been generated when there are nan values for WarningMax/FailureMin, etc fields in the notification metadata.

eg: when they're unset in the threshold plugin configuration:

```
"meta":{"DataSource":"value","CurrentValue":1.524927e-01,"WarningMin":1.000000e+04,"WarningMax":nan,"FailureMin":nan,"FailureMax":nan}}]
```

Should the nan values be quoted or represented as null or just not included in the json?

This patch just to illustrate a workaround I used....

```
diff --git a/src/utils_format_json.c b/src/utils_format_json.c
index da27a65..8b043e5 100644
--- a/src/utils_format_json.c
+++ b/src/utils_format_json.c
@@ -513,7 +513,11 @@ static int notification_meta_data_to_json(char *buffer, size_t buffer_size, /* {
  } else if (meta->type == NM_TYPE_UNSIGNED_INT) { \
    BUFFER_ADD ( "%"PRIu64, meta->nm_value.nm_unsigned_int); \
  } else if (meta->type == NM_TYPE_DOUBLE) { \
-   BUFFER_ADD ( "%e", meta->nm_value.nm_double); \
+   if ( isnan ( meta->nm_value.nm_double ) ) { \
+     BUFFER_ADD ( "null"); \
+   } else { \
+     BUFFER_ADD ( "%e", meta->nm_value.nm_double); \
+   } \
  } else if (meta->type == NM_TYPE_BOOLEAN) { \
    BUFFER_ADD ( "%s", meta->nm_value.nm_boolean?"true":"false"); \
   } \
```

mdubbyap:
Nice find! Thanks!  My gut is not including them at all, but what do others think? @BCWilsonDotCom @bviswana101 @octo @mfournier thoughts?

BCWilsonDotCom:
Yeah, I agree. They should probably not be included at all.

bviswana101:
Yes, Makes sense not to include it as JSON/ECMA does not define Nan or Infinity.

mdubbyap:
@ciomaire i've uploaded a patch, can you verify your problem is indeed fixed?  Sorry it took so long.

ciomaire:
@mdubbyap - just verified its working, thanks.

mfournier:
Uh oh, it looks like this last commit breaks the build on freebsd, @mdubbyap :-/ Would you mind taking a look at the error linked above ? Thanks !

mdubbyap:
Drat! Thanks @mfournier , will take a look.

mdubbyap:
I guess this is moot since https://github.com/collectd/collectd/pull/1396 covers the same thing (although written 5 months after this one). Any context on why this PR was ignored and another equivalent format chosen?

mdubbyap:
@octo could you comment on if this is worth bring up to date and fixing conflicts?

PR #1106, opened 2015-07-01
Added MetaDataSet to target_set

ymettier:
Hello,

This patch allows to set metadata with filters and `target_set` target.
It adds a new keyword `MetaDataSet`.
Example : 
```
 <Target "set">
   MetaDataSet "key1" "value1"
   MetaDataSet "key2" "value2"
 </Target>
```

Note : this patch adds some functions to the daemon/meta_data.c in order to clone_and_merge meta data. I prefered to share my new functions instead of implementing them in target_set.c.

Regards,
Yves

### DIFF
--- a/src/daemon/meta_data.c
+++ b/src/daemon/meta_data.c
@@ -106,12 +106,15 @@ static meta_entry_t *md_entry_alloc (const char *key) /* {{{ */
   return (e);
 } /* }}} meta_entry_t *md_entry_alloc */
 
-static meta_entry_t *md_entry_clone (const meta_entry_t *orig) /* {{{ */
+/* XXX: The lock on md must be held while calling this function! */
+static meta_entry_t *md_entry_clone_contents (const meta_entry_t *orig) /* {{{ */
 {
   meta_entry_t *copy;
 
-  if (orig == NULL)
-    return (NULL);
+  /* WARNINGS :
+   *  - we do not check that orig != NULL here. You should have done it before.
+   *  - we do not set copy->next. DO NOT FORGET TO SET copy->next IN YOUR FUNCTION
+   */
 
   copy = md_entry_alloc (orig->key);
   copy->type = orig->type;
@@ -119,6 +122,18 @@ static meta_entry_t *md_entry_clone (const meta_entry_t *orig) /* {{{ */
     copy->value.mv_string = strdup (orig->value.mv_string);
   else
     copy->value = orig->value;
+
+  return (copy);
+} /* }}} meta_entry_t *md_entry_clone_contents */
+
+static meta_entry_t *md_entry_clone (const meta_entry_t *orig) /* {{{ */
+{
+  meta_entry_t *copy;
+
+  if (orig == NULL)
+    return (NULL);
+
+  copy = md_entry_clone_contents(orig);
 
   copy->next = md_entry_clone (orig->next);
   return (copy);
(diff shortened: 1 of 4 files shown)


nathanielc:
I have been looking for something like this in collectd for a while. Glad to see some work is being done. I would like to see this get merged. I'll patch this locally and see how it works for my needs and post back here.

nathanielc:
See my comment here https://github.com/collectd/collectd/pull/1107#issuecomment-120129828 on my findings. Turns out I needed both PRs not just this one.

ciomaire:
@ymettier - this patch restores support for TypeInstance in ts_create()

```
diff --git a/src/target_set.c b/src/target_set.c
index 5e9d087..da2d307 100644
--- a/src/target_set.c
+++ b/src/target_set.c
@@ -177,6 +177,9 @@ static int ts_create (const oconfig_item_t *ci, void **user_data) /* {{{ */
       status = ts_config_add_string (&data->type, child,
           /* may be empty = */ 0);
 #endif
+    else if (strcasecmp ("TypeInstance", child->key) == 0)
+      status = ts_config_add_string (&data->type_instance, child,
+          /* may be empty = */ 1);
     else if (strcasecmp ("MetaDataSet", child->key) == 0)
       status = ts_config_add_meta (&data->meta, child,
           /* may be empty = */ 1);
```

ymettier:
*ooops*

Thanks @ciomaire . Fixed in commit a89fc92.
Also merged in PR #1107.

Regards,
Yves

shanson7:
This is a very nice feature, fairly crucial for deployments to many machines with various identifying characteristics (ie.g. datacenter, rack, cluster, etc.). Without additional metadata, slicing and dicing metrics gets messy and difficult.

yogeswaran:
:+1:  this feature will be very useful for us. This PR is 6 months old, is there any reason for not being merged?

ymettier:
@yogeswaran : no reason as far as I know.
@octo or @mfournier  : what do you think ?

src/daemon/meta_data.c
```diff
@@ -250,6 +322,27 @@ meta_data_t *meta_data_clone (meta_data_t *orig) /* {{{ */
...
 } /* }}} meta_data_t *meta_data_clone */
 
+int meta_data_clone_merge (meta_data_t **dest, meta_data_t *orig) /* {{{ */
+{
+  meta_entry_t *e;
+
+  if (orig == NULL)
+    return (0);
+
+  if(NULL == *dest) {
```
rubenk: *dest = = NULL please, and space after if.

src/daemon/meta_data.c
```diff
@@ -250,6 +322,27 @@ meta_data_t *meta_data_clone (meta_data_t *orig) /* {{{ */
...
+    *dest = meta_data_clone(orig);
+    return(0);
+  }
+
+  pthread_mutex_lock (&orig->lock);
+  for(e=orig->head; NULL != e; e = e->next) {
```
rubenk: e != NULL, and space after for

src/target_set.c
```diff
@@ -60,6 +78,42 @@ static int ts_config_add_string (char **dest, /* {{{ */
...
+  }
+
+  if (!may_be_empty && (strlen (string) == 0))
+  {
+    ERROR ("Target `set': The `%s' option does not accept empty string as second argument.",
+        ci->key);
+    return (-1);
+  }
+
+  if(NULL == (*dest)) {
```
rubenk: *dest = = NULL please, and space after if.

src/target_set.c
```diff
@@ -60,6 +78,42 @@ static int ts_config_add_string (char **dest, /* {{{ */
...
+    // Create a new meta_data_t
+    if( NULL == (*dest = meta_data_create())) {
```
rubenk: Same here.

src/target_set.c
```diff
@@ -60,6 +78,42 @@ static int ts_config_add_string (char **dest, /* {{{ */
...
+      ERROR ("Target `set': failed to create a meta data for `%s'.", ci->key);
+      return (-1);
+    }
+  }
+
+  return(meta_data_add_string (*dest, key, string));
```
rubenk: Space after return.

src/target_set.c
```diff
@@ -177,6 +235,10 @@ static int ts_invoke (const data_set_t *ds, value_list_t *vl, /* {{{ */
     return (-EINVAL);
   }
 
+  if(NULL != data->meta) {
```
rubenk: Space after if, data->meta != NULL

wpengcheng:
@ymettier This URL:https://github.com/collectd/collectd/pull/1107
image
Where is not
<Rule "opentsdb_load">
     <Match "regex">
       Plugin "^load$"
     </Match>
     <Target "set">
       MetaDataSet "tsdb_tag_pluginInstance" "load"
       MetaDataSet "tsdb_prefix" "sys."
     </Target>
   </Rule>
   <Rule "opentsdb_swap">
     <Match "regex">
       Plugin "^swap$"
     </Match>
     <Target "set">
       MetaDataSet "tsdb_tag_pluginInstance" "swap"
       MetaDataSet "tsdb_prefix" "sys."
     </Target>
   </Rule>


I have a doubt：The following few write where？



tsdb_prefix	Will prefix the OpenTSDB <metric> (also prefix tsdb_id if defined)
tsdb_id	Replace the metric with this tag
tsdb_tag_plugin
tsdb_tag_pluginInstance
tsdb_tag_type
tsdb_tag_typeInstance
tsdb_tag_dsname	When defined, tsdb_tag_* removes the related item from metric id.
If it is not empty, it will be the key of an opentsdb tag (the value is the item itself)
If it is empty, no tag is defined.

PR #1108, opened 2015-07-02
attempt at adding support for ssl for amqp plugin

paulczar:
this /should/ work but it the connection fails and I'm
having a hard time debugging.  No more time to mess with
it but adding it in case somebody else wants to take it on.


dago:
Closing as this is further tracked in #3391

PR #188, opened 2012-11-16
Add statistics from /proc/self/mountstats for nfs clients (like iostat -n)

ymettier:
Hello,

On a nfs client, run "iostat -n" and  have a look on /proc/self/mountstats...

This patch parses /proc/self/mountstats and get statistics about all NFS mountpoints.

Note : patch against release 5.1.1 is in perfwatcher/collectd-pw in branch patch-nfs_stats_with_network_plugin/5.1.1 if you are also interested in it.

Regards,
Yves

PR #1109, opened 2015-07-03
Add statistics from /proc/self/mountstats from nfs clients (like iostat -n)

ymettier:
Hello,

This is the same as PR #188, but in one commit.
Rebasing on master was not working well, and mergins messed up my branch.

So this is a brand new commint, with exactly the same patch. If you merge into master, it will be easier to read.

Regards,
Yves

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -4154,6 +4154,62 @@ statistics available. Defaults to B<false>.
 
 =back
 
+=head2 Plugin C<nfs>
+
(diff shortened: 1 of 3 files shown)


screentan:
This is the final feature that we need to be able to move from pcp to collectd for our site.  (Which would then allow us to move to influxdb.)  Please add this.

kristian-lesko:
Is there any change to get this merged into upstream?

ymettier:

PR #1111, opened 2015-07-04
Remove double reported innodb metrics in mysql plugin

tmm1:
Lots of innodb metrics are reported into a clean namespace with `InnodbStats true`. This PR removes duplicated metrics into legacy namespace outside the new innodb feature.


rubenk:
@tmm1 Apologies for now merging this sooner. Your patch does obviously the right thing. Would you mind rebasing so we can merge this?

octo:
Superseded by #1845

PR #1118, opened 2015-07-04
DO NOT MERGE: fix leak on exit

rubenk:


rubenk:
Please don't merge this yet. The leak is there, but this is not the right fix.

octo:
Cherry picked as 35e6b9d366016300178a55b163ac0173d7be5874, then reverted in 57d95740f47c862be0c08841fecf28aeb97ecb94

rubenk:
The correct fix is in #1159

Issue #966, opened 2015-03-12
write_http: context-dependent callback names limited to 63 characters, causes problems


coreykosak:
Hello,

I'm using a collectd built from tag `collectd-5.4.2` (7f3cfaf93d1dec3eece0be995b10202f7f61bb50 )

Consider this collectd.conf fragment:

    LoadPlugin write_http
    <Plugin write_http>
        <URL "http://example.com/aaaaaaaa/bbbbbbbb/cccccccc/dddddddd/eeeeeeee/ffffffff/1">
            User "zamboni"
            Password "ilikepie"
        </URL>
    
        <URL "http://example.com/aaaaaaaa/bbbbbbbb/cccccccc/dddddddd/eeeeeeee/ffffffff/2">
            User "zamboni"
            Password "ilikepie"
        </URL>
    </Plugin>

    PostCacheChain "WriteOut"
    <Chain "WriteOut">
        <Target "write">
            Plugin "write_http/http://example.com/aaaaaaaa/bbbbbbbb/cccccccc/dddddddd/eeeeeeee/ffffffff/1"
        </Target>
    </Chain>

There are two issues with this.

First, callback names are limited to DATA_MAX_NAME_LEN (== 64 including trailing NUL), so the two URLs are considered equivalent. This results in an error message like `plugin: register_callback: a callback named 'write_http/http://example.com/aaaaaaaa/bbbbbbbb/cccccccc/dddddd' already exists - overwriting the old entry!`. (Also, 63 characters is not much for a URL -- see Note 1)

Second, the "Chain" clause can't find *either* callback, because the Plugin reference in the Chain clause is *not* limited to DATA_MAX_NAME_LEN, resulting in a confusing situation where the conf file looks right but isn't. The result is an error like: `Filter subsystem: Built-in target 'write': Dispatching value to the 'write_http/http://example.com/aaaaaaaa/bbbbbbbb/cccccccc/dddddddd/eeeeeeee/ffffffff/1' plugin failed with status 2.`

In order to work around the first issue, the user would need to make sure that his/her URLs were distinct within the first 52 characters (because the prefix "write_http/" takes up 11 of the 63). In order to work around the second issue, the user would have to know that the Plugin reference (including write_http/) needs to be truncated to 63 characters. Both workarounds feel a little awkward to me.

[...]

Issue #872, opened 2014-12-18
Statsd + Network: instance type is truncate


duylong:
Hi, 

I use statsd and network plugins with collectd (5.4.1) and I have a issue.

I send this instance "logs.aaa-aaaaaa-aaa01_aa_aa-aaaaa-aaaa_aa.collectd.severity.informational" to collectd, but when it receive the data, I get to the output this instance "logs.aaa-aaaaaa-aaa01_aa_aa-aaaaa-aaaa_aa.collectd.severity.inf" (63 chars)

Do you know this issue ?

My config:

```
<Plugin statsd>
  Host "::"
  Port "8125"
  DeleteCounters false
  DeleteTimers   false
  DeleteGauges   false
  DeleteSets     false
  TimerPercentile 90.0
</Plugin>

<Plugin network>
    <Server "XXXX" "25826">
    </Server>
    MaxPacketSize 1452
</Plugin>
```

Cordialy,

duylong:
This limit is short when we have a FQDN with 40/50 characters. Associated with the program name, the limit is quickly exceeded.

Dinoshauer:
What is the reasoning for limiting these things? Would it be safe to build collectd myself with an increased limit?
I'm running into a similar wall with the StatsD plugin and it gets pretty hard to know what you're looking at when you don't have control of what the name of the metrics are.

tokkee:
I suppose the original reason for a static limit was to avoid lots of temporary memory allocations. Ultimately, I think we should change that.

Anyway, for now, it *should* be safe to increase the size but, afaik, nobody tried it yet, so there's some chance that things may break because parts of the code could assume the current limit.

Dinoshauer:
Thank you for your quick response, I might give it a try. Some of these names are several hundred characters long it seems

throck:
@Dinoshauer were you able to try increasing DATA_MAX_NAME_LEN, and if so, any issues?

Dinoshauer:
Yes, we did, we haven't run into any issues so far. It's been running on a bunch of machines for the past few months.

falzm:
I ran into this "issue" a few days ago, too. It would indeed be nice to raise the metric name length.

mre:
+1 This would be very useful for us as well. See https://github.com/collectd/collectd/pull/1120

rubenk:
[...]

Issue #210, opened 2012-12-03
64 character limit in aggregation regex


faxm0dem:
When using regex in aggregation, e.g.

    <Plugin aggregation>
      <Aggregation>
        [...]
        Host "/^(host1|host2|host3|...)/"
        [...]
      </Aggregation>
    </Plugin>
 

The key "Host" if ending in "/" is being interpreted as a regex. Only if the regex exceeds DATA_MAX_NAME_LEN which is currently 64, it doesn't get identified as such because it gets truncated, therefore there's no regex parsing error and no error at all. In the user perspective, no error at all, but the aggregation doesn't work.

Possible solution:

1. Add real Regex key, e.g.

         <Host>
          Regex "/everybody stand back this is a regular expression that is longer than 64 chars/"
         </Host>

2. have Regex increased maximum lenght, as 64 is pretty short for a reasonably complex regex.

rubenk:
Let's track this in #1120

PR #1120, opened 2015-07-06
enlarge the DATA_MAX_NAME_LEN for plugins

Amy-Lin:
For application running as docker container and also part of AWS ESC service, the metric folder name follows the naming conversion defined by AWS ECS, which is like `ecs-<ecs cluster name>-<service name>-<task definition name>-<container name>-<20 random characters>`. As a result, long name tends to be a reasonable name to be handled.

### DIFF
--- a/src/daemon/plugin.h
+++ b/src/daemon/plugin.h
@@ -39,7 +39,7 @@
 
 #define PLUGIN_FLAGS_GLOBAL 0x0001
 
-#define DATA_MAX_NAME_LEN 64
+#define DATA_MAX_NAME_LEN 128
 
 #define DS_TYPE_COUNTER  0
 #define DS_TYPE_GAUGE    1


mre:
+1

msiebeneicher:
+1

Dinoshauer:
We bumped the limit to `1024` and haven't had any problems. Just a heads up :)

mfournier:
This is a recurring question on IRC and the mailing-list. And #210, #872 and #966 are related to this here on github. So I think we really should do something about it.

@tokkee mentioned at some point this should ideally be dynamically calculated rather than a hard-coded limit. I don't know how complicated that would be to implement though.

Assuming a first step would be to simply bump up this limit, by how much would be reasonable ? What sort of side-effects could this possibly have ? @octo, any thoughts ?

tbenade:
+1

davidmcgregor:
+1

davidduffett:
+1

grillp:
+1

travisblakeney:
+1

lantrix:
+1

IngaFeick:
+1

MaxBosse:
+1

xo4n:
+1

hendrikmaus:
+1

huangered:
+1

linlynn:
+1

fanyeren:
+1

mschenck:
Was this tested with the `statsd` plug-in with metric names great than 64 characters?

I've run into issues with collectd re-compiled with `DATA_MAX_NAME_LEN` set to 128.  The metrics would work for a random period of time, then suddenly stop working (no data being written).  I was using `write_grahite` so I'm not sure if that's where the problem was, but since only the statsd metrics stopped showing up I was planning to start the investigation there.

I'm hoping to get some time to re-test, as the last time I discovered this issue, I did not have time to actually investigate where the metric loss was occurring.

tokkee:
Doing this dynamically would require to change `char[DATA_MAX_NAME_LEN]` to `char *` and update all plugins to provide their own buffers, either on the stack or using something like strdup. Any sort of dynamic memory allocation on the heap (e.g. strdup) will also require cleanup.

This will be a lot of work but an intermediate solution could look like this:
(1) Use dynamic buffers in the core (plugin.c) and let it handle memory management.
(2) Add a second variant of `value_list_t` and `plugin_dispatch_values` that uses dynamic buffers.
(3) Rewrite plugins to use the new, dynamic interface.

That'll allow to focus on plugins first that are more likely to hit the problem.

octo:
I had splitting out an `identifier_t` on my mental todo list for a while. What we could do, is:

 *  Create an `identifier_t` which uses dynamically allocated `char*`.
 *  Add `identifier_t *id` to `value_list_t` and `notification_t`.
 *  In `plugin_dispatch_values()`, if `id == NULL`, allocate a new `identifier_t*` from the fixed-sized fields.

This is very close to @tokkee's suggestion, but re-uses the existing `plugin_dispatch_values()`.

tokkee:
SGTM

mimrock:
What do you think about merging this PR as it is, with simply increasing the DATA_MAX_NAME_LEN constant since it would spare a lot of headache to many?

protomouse:
@mschenck We're experiencing similar issues (data intermittently not being written through statsd after applying this fix). Did you figure something out?

(Also, apologies for cluttering this with my commits)

dennisosimon:
+1

rmrf:
+1

octo:
@mfournier, @rubenk and I met yesterday and discussed this issue. We decided to raise the default of `DATA_MAX_NAME_LEN` to 128 (i.e. double it) for the upcoming 5.6 feature release. Breaking out an `identifier_t` is still an option for the future.

octo:
Rebased and committed as 6fd0d2574fc386c2724c8307d33b12529b30048b. Thanks @Amy-Lin!

rubenk:
This broke the test-suite, the failure is at https://ci.collectd.org/job/master-build-on-jessie-amd64-with-default-toolchain/416/artifact/collectd-5.5.2.1107.g33cffce/src/test-suite.log
 @octo could you have another look please?

octo:
@rubenk will do

octo:
Issues with the ceph tests are tracked in #1834.

PR #1123, opened 2015-07-06
bind: fix zone count check

mihu:
fixing silly mistake.


mfournier:
Excellent @mihu, thanks for paying attention to this details !

I merged your patch to the collectd-5.5 branch as 9424954f, so it will be part of the next 5.5 bugfix release.

PR #1124, opened 2015-07-06
mqtt plugin: New plugin for publishing and subscribing to MQTT topics.

octo:
This new plugin allows to publish and subscribe to topics handled by an MQTT broker. The configuration is intentionally similar to the *AMQP plugin* and looks like this:

**Publisher:**
```
<Plugin mqtt>
  <Publish "foo">
    Host "mqtt.example.com"
  </Publish>
</Plugin>
```

**Subscriber:**
```
<Plugin mqtt>
  <Subscribe "bar">
    Host "mqtt.example.com"
  </Subscribe>
</Plugin>
```

This plugin is based on work by @falzm – thank you so much! CC @jpmens


mfournier:
Thanks for taking the time to put this back on tracks @octo \o/

FYI I updated the jenkins slaves to include libmosquitto (for those distros which ship it), so the build error currently reported is genuine. Basically the plugin doesn't build against the 1.x version of the library, only the 0.x.

Here's a census of the libmosquitto versions available in the distros we currently have in jenkins:
 - debian 7 (previous stable), ubuntu 12.04 (previous LTS) and 14.04 (current LTS): libmosquitto0 0.15
 - debian 8 (current stable): libmosquitto1 1.3.4
 - freebsd 10 (current stable): mosquitto-1.4.2
 - EL 7 (current stable): mosquitto-devel 1.4.2

jpmens:
It's difficult for me to follow all this on a tiny mobile device (and hence have neither yet said "thank you" nor tested), but "plugin builds with libmosquitto 0.x" is not bad news IMO as 0.x is ancient. The Mosquitto project provides newer PPAs etc. can't those be used?

(We're always recommending people to NOT use the old distro offerings.)

octo:
Thanks @mfournier, that was extremely helpful.

@jpmens I added support for 0.15 *and* 1.* versions, so that we don't get too many complaints from users on old (usually LTS) distributions.

mfournier:
@octo great ! nb: it looks like libmosquitto failed to get detected on freebsd. Maybe we'll need to add a `--with-libmosquitto` flag on this platform.

@jpmens, plugins which aren't available by default with the "official" collectd packages shipped in distros are used much less. So less feedback, less bug reports and less patches. Thus the effort to make as much plugins as possible fit well into the most popular distros distros.

mfournier:
@octo, indeed `./configure --with-libmosquitto=/usr/local` does the trick on freebsd 10. Another small issue showed up though: the `-I/usr/local/include` parameter isn't set when compiling `mqtt.c`. https://github.com/octo/collectd/pull/36 corrects this small detail.

After that, the plugin builds fine on this platform:

```
src/.libs/mqtt.so:
        libmosquitto.so.1 => /usr/local/lib/libmosquitto.so.1 (0x801603000)
        libc.so.7 => /lib/libc.so.7 (0x80081f000)
        libssl.so.7 => /usr/lib/libssl.so.7 (0x801814000)
        libcrypto.so.7 => /lib/libcrypto.so.7 (0x801a7f000)
        libthr.so.3 => /lib/libthr.so.3 (0x801e72000)
        librt.so.1 => /usr/lib/librt.so.1 (0x802097000)
        libcares.so.2 => /usr/local/lib/libcares.so.2 (0x80229d000)
```

rubenk:
I'm planning on doing some testing of this plugin, but am I right in that the plugin doesn't support ssl/tls yet?
If so, should I open a separate issue for this?

jpmens:
There is no TLS-support in the code. If I had a spare hour or two, I'd do it..

mfournier:
@rubenk, yes please, open a new issue for this. Thanks!

njh:
Created issue for possible race condition bug in #1262

sbistline:
Hi All,

I have seen the config info for MQTT, but no actual "plugin". I was hoping to be able to publish collectd data via MQTT, but having searched the web everywhere all I see are conversations about a plugin, a WIKI page with the config info but no actual plugin.

Any help would be REALLY appreciated.

Thanks

njh:
The first version with MQTT support will be version 5.6 (https://collectd.org/wiki/index.php/Plugin:MQTT).

Version 5.6 has not been released yet:
https://collectd.org/download.shtml

There is some information about how to build the development version here:
https://collectd.org/dev-info.shtml

sbistline:
Thank you for the response.... when is 5.6 expected to be available?

PR #1125, opened 2015-07-06
build: add libavltree, libcommon & libheap dependencies

gustavoz:
Otherwise it can break on very parallel builds since collectd link time
arrives before one or more of these were built.

Signed-off-by: Gustavo Zacarias <gustavo@zacarias.com.ar>


mfournier:
Thanks @gustavoz, nice catch !

I applies the patch to the collectd-5.5 branch as 780e6a7.

PR #1133, opened 2015-07-13
Statsd port inuse

michaelsalmon:
This patch primarily fixes the error that is sent if getaddrinfo returns both the IPv4 & IPv6 wildcard address. If a bind has been successful then the error is downgraded to an info. The only case that I can think of where this will occur is just multiple wildcards.

It also removes the call to getnameinfo if debug is not enabled and does not add AI_ADDRCONFIG to hints if there won't be any DNS lookup.


mfournier:
@michaelsalmon thanks for your patch !

It looks like something went wrong with the pull-request though: Commits 112 Files changed 93

Could you please rebase your patchset against the collectd-5.4 branch, and resubmit against this branch please ? Thanks !

michaelsalmon:
I have created a new pull request that only changes 1 file this time.

PR #1134, opened 2015-07-13
smart: optionally query sleeping disks

pattop:
Most disks these days don't wake from sleep when querying SMART data but still report updated information such as temperature.


pattop:
The remaining build failures are unrelated to this change:

utils_time_mock.c: In function 'cdtime':
utils_time_mock.c:31: error: integer constant is too large for 'long' type
make[3]: *** [utils_time_mock.lo] Error 1

rubenk:
I think this is fixed by f45054 in a similar way.

PR #1135, opened 2015-07-13
Add missing kstat definitions

dago:
Without these an undefined `*kc` is reported and kstat symbols are missing.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -34,6 +34,9 @@ check_PROGRAMS += test_utils_vl_lookup
 TESTS += test_utils_vl_lookup
 test_utils_vl_lookup_SOURCES = utils_vl_lookup_test.c testing.h
 test_utils_vl_lookup_LDADD = liblookup.la daemon/libcommon.la daemon/libplugin_mock.la
+if BUILD_WITH_LIBKSTAT
+test_utils_vl_lookup_LDADD += -lkstat
+endif
 
 noinst_LTLIBRARIES += libmount.la
 libmount_la_SOURCES = utils_mount.c utils_mount.h
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -44,7 +44,9 @@ check_PROGRAMS += test_utils_mount
 TESTS += test_utils_mount
 test_utils_mount_SOURCES = utils_mount_test.c testing.h
 test_utils_mount_LDADD = libmount.la daemon/libcommon.la daemon/libplugin_mock.la
-
+if BUILD_WITH_LIBKSTAT
+test_utils_mount_LDADD += -lkstat
+endif
 
 sbin_PROGRAMS = collectdmon
 bin_PROGRAMS = collectd-nagios collectdctl collectd-tg
--- a/src/daemon/common_test.c
+++ b/src/daemon/common_test.c
@@ -27,6 +27,10 @@
 #include "testing.h"
 #include "common.h"
 
+#if HAVE_LIBKSTAT
+kstat_ctl_t *kc;
+#endif /* HAVE_LIBKSTAT */
+
 DEF_TEST(sstrncpy)
 {
   char buffer[16] = "";
--- a/src/daemon/utils_subst_test.c
+++ b/src/daemon/utils_subst_test.c
@@ -29,6 +29,10 @@
 #include "common.h" /* for STATIC_ARRAY_SIZE */
 #include "utils_subst.h"
 
+#if HAVE_LIBKSTAT
+kstat_ctl_t *kc;
+#endif /* HAVE_LIBKSTAT */
+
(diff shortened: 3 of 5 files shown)


mfournier:
Thanks a lot @dago ! It seems your patch was missed out, sorry about that !

mfournier:
@dago I cherry-picked these patches to the collectd-5.5 branch, in an attempt to make https://buildfarm.opencsw.org/buildbot/builders/collectd-55-solaris10-sparc pass. But there's a puzzling error still left. Any idea ? Thanks !

dago:
Yes, that's easy: them missing symbols are all in `-lsocket  -lnsl`. There is just linkage missing in the test suite for that executable.

mfournier:
Thanks for taking a look @dago ! So what keeps me back from simply adding `-lsocket -lnsl` somewhere in the Makefile is that the mostly identical Makefile from the master branch is able to build and run these tests without explicitly declaring these 2 symbols. Worse than that, there's not a single reference to `nsl` in master, according to `git grep`.

Would you mind running `./configure --disable-silent-rules` and then pasting the output of `make check` ? Or even better, submit a patch fixing the problem :-) I don't have any solaris environment to test this on, so your help would be greatly appreciated !

dago:
I enabled the configure switch now, the builds are failing at a different position now:
  https://buildfarm.opencsw.org/buildbot/waterfall?category=collectd

rubenk:
Ah, we're missing the definition of MAXCOMLEN. A quick grep shows me it is in <sys/user.h> on Solaris 10.

rubenk:
But we do include that at https://github.com/collectd/collectd/blob/master/src/processes.c#L131, so that's not it.

mfournier:
MAXCOMLEN usage was introduced in fa517c8, which includes `<sys/user.h>` and built fine up to now. weird.

rubenk:
I have Solaris 10 x86_64 running in VirtualBox, and can't reproduce it (yet)

dago:
Not really, from what I see it is 9407ee and it indeed failed:
  https://buildfarm.opencsw.org/buildbot/builders/collectd-solaris10-i386

rubenk:
I'm not seeing the `--disable-silent-rules`, are you sure that got applied?

dago:
You can dive into the configure task:
  https://buildfarm.opencsw.org/buildbot/builders/collectd-solaris10-i386/builds/281/steps/shell_1

rubenk:
Ah, now I see it, I was looking at build 277, sorry.

rubenk:
And I can reproduce it now too. @dago would you mind also adding `--disable-dependency-tracking`? That will shorten the build output a lot.

dago:
Sure, done. Rebuilds running now.

rubenk:
Thanks. I see the same issue on fa517c8 so at least it doesn't seem to be a regression.

rubenk:
MAXCOMLEN is behind `#if defined(_KERNEL) || defined(_KMEMUSER)` so that might explain a few things ;)

mfournier:
Ok, so if I understand this isn't exposed to the user, only part of the kernel API ? Curiously, this seems to build fine on FreeBSD.

#1508 for the discussion which led us here.

rubenk:
Yeah, it doesn't seem to be exposed to userland. I think we're pretty safe just hardcoding it to 16 on Solaris.

rubenk:
Fix at #1725

mfournier:
I just pushed the fix to the collectd-5.5 branch, so we'll now be able to focus back on our initial problem: making `make check` pass for collectd-5.5 on Solaris ;-)

rubenk:
Great :)
I'll go and sprinkle some BUILD_WITH_LIBSOCKET over our testsuite.

mfournier:
@rubenk I'm just a bit wary. The tests already pass on master without any BUILD_WITH_LIBSOCKET sprinkling. My initial intention here was to cherry-pick a couple of kstat-related patches from master (cf. this issue's title), to follow @dago's initiative of also pointing the solaris buildbot to the latest release branch.

rubenk:
Ok, the problem is that we only link collectd and a few test binaries with $(COMMON_LIBS) as you can see in https://github.com/collectd/collectd/blob/master/src/daemon/Makefile.am#L13

The proper fix is to link the convenience libs to $(COMMON_LIBS) so all binaries profit, but this might need some restructuring.

rubenk:
I had to backport a few changes, the most important one being 7e23de8, but 5.5 now builds too.

mfournier:
So this looks all good now ! Hats off @rubenk, well done !
Thanks also @dago for all your support !

dago:
This looks good, thanks @rubenk and @mfournier ! If you now make a release I'll happily update the package at OpenCSW :-)

PR #1137, opened 2015-07-13
Don't send warning if the old and new callback is the same

michaelsalmon:
This patch removes the warning generated if a plugins registers itself several time as threshold does. If the old and new callback are the same then the warning is suppressed.

I have also replaced a call to sfree() with free() as this is, I believe more appropriate.


michaelsalmon:
This may be breaking because I based it on collected-5.5. I will redo it.

octo:
Hi @michaelsalmon and thank you very much for your patch.

This is very likely due to a misconfiguration and I think printing a warning to the user is the right thing to do here.

PR #1142, opened 2015-07-15
implement optional reporting of active state

ymettier:
Hello,

I want to report CPU metrics
* by state
* by cpu
* aggregated by state as "active".

However, reportbystate and aggregated as "active" are mutually exclusive in current code. I'm wondering why.
So I added a new option : `ReportStateActive`

Here is the doc (from collectd.conf.pod) : 
```
       ReportStateActive false|true
           When set to true, aggregates (sums) all non-idle states into one "active" metric.

           When ReportByState is set to false, it is already the case and this option is ignored (redundant with ReportByState).

           When ReportByState is set to true, "active" metric is not reported by default. You need to enable it explicitely with
           ReportStateActive.
```

The idea is that when `reportByState` is true (e.g. "active" state is not reported), we can force it to be reported with `ReportStateActive true`.

Regards,
Yves

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1393,6 +1393,17 @@ When set to B<true>, the default, reports per-state metrics, e.g. "system",
 When set to B<false>, aggregates (sums) all I<non-idle> states into one
 "active" metric.
 
+=item B<ReportStateActive> B<false>|B<true>
+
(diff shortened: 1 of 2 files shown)


toni-moreno:
I really need this PR !! Thanks a lot @ymettier.

:+1

PR #1143, opened 2015-07-15
write_redis plugin: Add StoreRates support and option to limit size of sorted sets

brianpkelly:
This PR has two parts. The first fixes a bug that causes added values to look like `"1436983727.331209660:0.920.860.521436983727.331:0.92:0.86:0.52"` instead of `"1436983717.723:1.09:0.89:0.53"`. I think this was caused by a mixed up merge as the `APPEND` macro and associated code should have been removed in  #1070. The second part gives the user the option to limit the size of the sorted sets where collectd writes. For example, with the configuration `MaxSetSize 10`, only the ten latest values will be stored by redis. The limit is disabled with any negative value.

Edit (7/30): This PR now also adds support for the StoreRates option.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -7661,6 +7661,10 @@ is recommended but not required to include a trailing slash in I<Prefix>.
 
 This index selects the redis database to use for writing operations. Defaults to C<0>.
 
+=item B<MaxSetSize> I<Items>
+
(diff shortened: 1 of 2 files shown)


mfournier:
Many thanks @brianpkelly ! FYI the bugfix commit was merged to the collect-5.5 branch as 8f7b3f2f73 and I rebased & merged the rest into current master.

Issue #551, opened 2014-02-10
Allow to define threshold configuration block in several files


adrianlzt:
I'm using puppet to configure collectd, and my idea is to have in the same file plugin  and threshold configuracion.
Something like:
``foo.conf:``
```
<Plugin foo>
  param bla
</Plugin>
<Plugin threshold>
  <Type "abc">
    FailureMax 1
  </Type>
</Plugin>
```

But if two different blocks of "threshold" are defined, this error is shown when collectd is restarted:
``Starting collectd: plugin: register_callback: a callback named `threshold' already exists - overwriting the old entry!``

Using the same file to define all thresholds is complicated to manage.
Any chance to implement this?

michaelsalmon:
I think that it is more appropriate to fix this in plugin.c. It should quietly ignore updating a callback with the same value.

My reasoning is that plugin.c knows already that a callback has been registered and if it is fixed here then it will never have to be fixed elsewhere.

On the other side it wastes cycles..

I think that I could have come to grips with the problem faster if I had known which callback was being overwritten. Perhaps the error message could include that information.

octo:
@michaelsalmon I'd argue that the duplicate check is a start-up cost and therefore negligible for the overall runtime performance.

michaelsalmon:
I don't want to remove the check, just the error when the callback is the same.

I have submitted pull requests for both plugin.c & threshold.c that address this problem.

PR #1144, opened 2015-07-16
Only register callbacks once

michaelsalmon:
This patch prevents threshold from registering it's callbacks more than once. At present it registers them every time that a config block is found.

I have presumed that threshold_tree will only be modified in threshold.c. If that changes then threshold may stop working.

This fixes issue #551.

### DIFF
--- a/src/threshold.c
+++ b/src/threshold.c
@@ -872,6 +872,7 @@ int ut_config (oconfig_item_t *ci)
 { /* {{{ */
   int i;
   int status = 0;
+  int old_size = c_avl_size (threshold_tree);
 
   threshold_t th;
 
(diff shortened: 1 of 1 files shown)


src/threshold.c
```diff
@@ -915,7 +916,8 @@ int ut_config (oconfig_item_t *ci)
       break;
   }
 
-  if (c_avl_size (threshold_tree) > 0) {
+  /* register callbacks if this is the first time we see a valid config */
+  if (old_size == 0 && c_avl_size (threshold_tree) > 0) {
```
tokkee: Nit: Common collectd style adds parenthesis around each case in an if statement:

    if ((old_size == 0) && ...)

tokkee:
LGTM. Thanks!

octo:
LGTM, thanks @michaelsalmon! This should be cherry-picked to *collectd-5.4* rather than merged.

octo:
Committed (with style changes) as 74fa9577b1acbb965488aa0a48c140d8ece437bf. Thanks @michaelsalmon!

PR #1153, opened 2015-07-21
Fix Unixsock.pm

mbethke:
I just noticed a small problem with the Unixsock code I wrote a few months ago:
[On Perls older than 5.14](http://search.cpan.org/~rjbs/perl-5.22.0/pod/perlsub.pod#Localization_of_special_variables), if you called `getval()` while having `$_` aliased to a variable marked Readonly (using Readonly.pm that does some tie magic behind the scenes), `_escape_argument()` would error out with "modification of read-only value attempted". This was due to `local $_` not being thorough enough in stripping the magic off `$_` so the readonly-ness would stay while the variable was localized. The fix is trivial and just makes the little helper sub a bit more verbose.

While I was at it, I also fixed an autoderef in one of the tests that should never have been there in the first place; it causes warnings on recent Perls and will probably fail on 5.22.

### DIFF
--- a/bindings/perl/lib/Collectd/Unixsock.pm
+++ b/bindings/perl/lib/Collectd/Unixsock.pm
@@ -137,13 +137,13 @@ sub _parse_identifier
 
 sub _escape_argument
 {
-	local $_ = shift;
+    my $arg = shift;
 
-	return $_ if /^\w+$/;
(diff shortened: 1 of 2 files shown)


pyr:
Thanks @mbethke !

PR #1154, opened 2015-07-21
Apache plugin: Remove undocumented Host option

rubenk:
If it is not set, we access hostname_g, and that global
variable isn't set yet at that point.

### DIFF
--- a/src/apache.c
+++ b/src/apache.c
@@ -41,7 +41,6 @@ struct apache_s
 {
 	int server_type;
 	char *name;
-	char *host;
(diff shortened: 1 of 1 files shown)


pyr:
@rubenk I think it would make more sense to just document the option. Shadowing the global hostname with a user supplied one could make sense for that plugin.

rubenk:
Ok, makes sense. I'll have to look into the underlying issue that config_add access hostname_g but global variables are set after the config handlers are run.

PR #1157, opened 2015-07-24
process monitoring changes to processes plugin

deepix:
We tried to use processes.c as a general process monitoring plugin, but faced couple problems:

* It used config-time regex patterns to pre-select processes, which is not flexible enough
* It aggregated metrics over these patterns, so we'd have to create many patterns to get metrics for individual processes

So we've made some changes to allow us to:
* send metrics for individual processes
* select processes based on min. CPU or memory %
* send a few other metrics: CPU %, memory %, PID, PPID, runtime

so that we can have a top-like dashboard.

Tested on Linux (Ubuntu Trusty).

Please take a look and let us know what you think!


deepix:
I don't understand this error.  It may not be related to my changes per se.  Can the admin please check?

`ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.
Setting commit status on GitHub for https://github.com/collectd/collectd/commit/b9cb82e3241040cca2a721ae53951a2f9e52b448
Warning: you have no plugins providing access control for builds, so falling back to legacy behavior of permitting any downstream builds to be triggered
Finished: FAILURE`

deepix:
Just checking if there's any feedback on this?  Thanks.

deepix:
Pinging again ... :-)

kbrock:
This looks helpful, any suggestions on pushing this through?

deepix:
Yes, still waiting ...

lawre:
+1 Hi deepix and mfournier, I'm very interested in this enhancement as well.

sreenip:
Hi Deepix  i would like to use this plugin and tried. but i dont see any change in process metrics that are reported. Can you help me with the plugin details that we need to put in the conf file.

PR #1158, opened 2015-07-25
collectd -T: fix memory leak

rubenk:


mfournier:
Cherry-picked as 51f45d6b2, thanks !

PR #1159, opened 2015-07-25
Fix a memory leak on shutdown

rubenk:


pyr:
:+1:

mfournier:
Merged as d089d2691a24, thanks for tracking this down @rubenk ! All your recent PRs are very much appreciated !

rubenk:
You're welcome, and thanks for merging them so quickly!

PR #1160, opened 2015-07-25
memcached: plug leak

rubenk:


mfournier:
This patch has been commited on 10. July as 76c004d1.

rubenk:
Ah, sorry for the extra work. I missed this since this commit went to 5.4 and not to master yet, and I was testing on the master branch.

PR #1161, opened 2015-07-25
gmond: plug a small leak on exit

rubenk:


mfournier:
This patch has been commited on 10. July as 26a4e6f1.

PR #1164, opened 2015-07-25
Write graphite fixes

rubenk:


pyr:
This makes it easier on the eyes, :+1:

pyr:
@mfournier I'm inclined to go ahead and commit if you see no problem with it.

PR #1165, opened 2015-07-25
Fix some small leaks on error paths

rubenk:


mfournier:
Thanks a lot @rubenk ! Did you also find these using ASAN ?

Cherry-picked as 00b6ce0313.

rubenk:
@mfournier yes, I found them all with asan. I did have to patch collectd a little to shutdown after a few hundred iterations, since asan doesn't find runtime leaks, only at exit.

PR #1166, opened 2015-07-25
liboconfig: really fix lex warnings

rubenk:
Flex has two options to disable input and unput.
Turn them on.


mfournier:
Cherry-picked as 24667b83, awesome finding @rubenk, thanks !

PR #1173, opened 2015-07-29
write_atsd plugin

rmakulov:
This new write plugin allows to send data into the [Axibase Time-Series Database](https://axibase.com/products/axibase-time-series-database/). The configuration is simple and contains basic parameters used to store data in ATSD.


rmakulov:
What should I do with `unknown plugin` error?

mfournier:
@rmakulov yes, the output of jenkins is a bit confusing, it's still a work in progress. Sorry about that !

The `unknown plugin` message you mention can be safely ignored. The problem which marks your last commit as failed is a configution error in jenkins I made a couple of days ago.

rmakulov:
Could you please tell what should I do to pass the check?

rmakulov:
What else should I do?

rubenk:
At least rebase. You're now asking us to merge 986 commits.

rmakulov:
Is that better?

rubenk:
It is. Your pull contains unrelated changes to src/df.c however.

rmakulov:
@rubenk something else?

rmakulov:
@octo is there anything to be done on this PR?

rmakulov:
@rubenk  is there anything to be done on this PR?

src/utils_format_atsd.c
```diff
@@ -0,0 +1,169 @@
...
+            n--;
+        }
+        s++;
+    }
+    *d = '\0';
+
+    return (dlen + (s - src));        /* count does not include NUL */
+}
+
+int format_value(char *ret, size_t ret_len, int i, const data_set_t *ds, const value_list_t *vl,
```
octo: Please use the variable name `i` for loop indexes only. That is so common that other uses are confusing. Other plugins use `ds_index` instead.

src/utils_format_atsd.c
```diff
@@ -0,0 +1,169 @@
...
+#include "collectd.h"
+#include "plugin.h"
+#include "common.h"
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include "utils_format_atsd.h"
+
+/* strlcat based on OpenBSDs strlcat */
```
octo: How is the original code licensed? You need to add license information here.

Unless, of course, by "based on" you really mean "is mimicking its behavior", i.e. this is a new re-implementation of the same interface.

src/write_atsd.c
```diff
@@ -0,0 +1,1273 @@
...
+ *  Normalize read plugin signatures according to the ATSD scheme of data
+ *  to prepare series commands from Network API.
+ *  https://github.com/axibase/atsd/tree/master/api/network
+ *
+ *	In general use prefix with plugin as metric name.
+ *	For some plugins calculate additional statistics,
+ *	use types in metric name.
+ *	Build tags from plugin and type instances with small changes.
+ */
+    if (strcasecmp(vl->plugin, "cpu") == 0) {
```
octo: This huge if-else not something we can merge. This is way too specialized and we can't possibly keep this list up to date. For example, the code below assumes that the "entropy" plugin reports a single metric, which is a very brittle assumption. It's just too much of a maintenance burden.

It appears that atsd is using key value pairs to identify metrics. I recommend to look at what plugins for other such systems do and mimic their behavior, for example [the write_prometheus plugin](https://github.com/collectd/collectd/blob/master/src/write_prometheus.c#L359-L373).

If you *have* to map collectd metrics to predefined atsd metrics, write that mapping to a file and have the plugin load that. It at all feasible, try to avoid that though.

Example syntax for such a file:

```
entropy/entropy  entropy.available
swap/swap-*      memory.swap_%{type_instance}
```

This way these mappings are not fixed at compile time and the user has control over what happens.

rmakulov:
Hi @octo, thanks for the feedback!
We are performing refactoring.

rmakulov:
Hi @octo!
Refactoring performed. Also we improved plugin df to add unescaped plugin_instance (disk_name) into meta information in #2567.

vtols:
Hi @octo, we have been waiting for your response for a long time, could you have a look at refactored code, please

PR #1175, opened 2015-07-31
Enable travis builds

pyr:
This is not as featured as the collectd-ci infrastructure but can give a quick additional data point on commits.
What do you think @mfournier ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: c
+script: ./configure && make
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,3 @@
+sudo: false
 language: c
 script: sh build.sh && ./configure && make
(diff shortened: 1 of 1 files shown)


pyr:
@rubenk good idea

mfournier:
Why not ? Although I think this is quite basic and simplistic compared to the stuff done with jenkins (and redundant obviously). The biggest criticism I have towards travis in this case, is that ubuntu 12.04 is the only platform available to run tests.

What is the context behind this @pyr ? Does "quick data point" imply you need a faster feedback on commits than what jenkins offers ?

pyr:
@mfournier it's more to have feedback on my branches by default before pull-requests, really.

mfournier:
Ok, I see! I had this use-case in mind actually, but didn't get around to implement anything in jenkins yet.

Meanwhile I noticed that travis added MacOSX support recently. As this is an annoying platform to integrate into jenkins, we might want to divert the .travis.yml for this at some point.

PR #1177, opened 2015-08-01
Fix libmodbus detection on FreeBSD

rubenk:
We look for modbus/modbus.h in /usr/local/include/modbus
but we should look for modbus.h

This is only an issue on FreeBSD since /usr/local/include is not
in the default search path.


mfournier:
Merged as 6124da7a4, thanks @rubenk !

PR #1179, opened 2015-08-01
amqp: look for librabbitmq in /usr/local as well

rubenk:
This detects it automatically on FreeBSD.
Ideally we should just use pkg-config but that is
left as an exercise for the reader ;)


mfournier:
Merged as c1c5087fd5, thanks !

PR #1180, opened 2015-08-01
core: sys/types is already included in collectd.h

rubenk:


pyr:
@rubenk is this fixing any error ? `<sys/types.h>` is usually protected against multiple includes.

rubenk:
@pyr, no just good practice. We check for <sys/types.h> with ./configure but don't wrap it in HAVE_SYS_TYPES_H everywhere. If we only include it from collectd.h there's less change for plugin authors to do it wrong.

pyr:
@rubenk I think the check for `sys/types.h` can be killed, we're not going to be in luck on a system where that doesn't exist :-) If you really want to work on a PR in that space, you can kill the `sys/types.h` check and remove the `#ifdefs`.

rubenk:
Ok, I'll do that in a separate PR, thanks.

PR #1183, opened 2015-08-01
autotools: create separate macro for each library

rubenk:
configure.ac is getting a bit big.
Move the checks for libraries to m4/
and create one check per library.


mfournier:
Fundamentally, big +1 for such a reorganisation. I'm a bit reluctant by the fact this would make it difficult to track down previous changes with git though (ie: the "history" of each block would get reset).

rubenk:
Yes, good point. But to make an omelet you have to break some eggs ;)

rubenk:
I'll take a different approach, in smaller steps.

PR #1185, opened 2015-08-01
configure.ac: remove some redundant x'es

rubenk:


mfournier:
Merged as 32bf93fec7, thanks !

PR #1186, opened 2015-08-01
configure.ac: fix indentation

rubenk:


mfournier:
Merged as 9dad816224e2, thanks !

PR #1189, opened 2015-08-02
irq: skip FIQ line in /proc/interrupts

### LINKED ISSUE #971: IRQ plugin sending a null value
My install of collectd is spamming syslog with messages of the form:

    collectd[]: parse_value: Failed to parse string as derive: (null).

By process of elimination I determined it is coming from the IRQ plugin. I can't seem to find anyone else with the same issue. If it helps, this system is a Raspberry Pi.

rubenk:
/proc/interrupts on my Raspberry PI contains:
FIQ:              usb_fiq

This line doesn't contain any per cpu counters
but we try to parse it anyway, resulting in:
parse_value: Failed to parse string as derive: usb_fiq.

Fixes #971

### DIFF
--- a/src/irq.c
+++ b/src/irq.c
@@ -155,6 +155,10 @@ static int irq_read (void)
 		if (irq_name[irq_name_len - 1] != ':')
 			continue;
 
+		/* Is it the the ARM fast interrupt (FIQ)? */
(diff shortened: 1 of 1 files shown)


mfournier:
Merged as 91fbf283d2c, thanks !

PR #1190, opened 2015-08-03
logstash: initialize conf struct to 0

vincentbernat:
With YAJL 1 (at least on Ubuntu Precise), if `conf.indentString` is not
initialized correctly, we would get a segfault even when `conf.beautify`
is set to 0. We avoid this case by initializing the whole structure to
0. `conf.beautify = 0` is kept for explicitness.

Using C99 designated initializer. Already used at some other places, including when empty (in `curl_json` plugin for example).

### DIFF
--- a/src/log_logstash.c
+++ b/src/log_logstash.c
@@ -208,7 +208,7 @@ static void log_logstash_log (int severity, const char *msg,
 {
 	yajl_gen g;
 #if !defined(HAVE_YAJL_V2)
-	yajl_gen_config conf;
+	yajl_gen_config conf = {};
 
 	conf.beautify = 0;
 #endif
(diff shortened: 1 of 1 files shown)


pyr:
:+1:

mfournier:
Merged to the collectd-5.5 branch as a9b65b73fc. Many thanks @vincentbernat !

vincentbernat:
You also need to merge it to master? Or will collectd-5.5 branch be merged?

mfournier:
Yes, the release branches always eventually get merged to master. As a matter of fact, I merged the current release branches to master just a couple of minutes after my previous message. Sorry for the confusion !

PR #1191, opened 2015-08-03
Use environmental variables from /etc/sysconfig/collectd

faxm0dem:
This may help in migrating from sysvinit to systemd
FWIW We use it to set `PYTHONDONTWRITEBYTECODE 1` so python plugins won't write `.pyc` files after puppet removes them


pyr:
:+1:

rubenk:
:+1:

mfournier:
Thanks a lot @faxm0dem ! FYI I added e7ed65fd to this.

PR #1195, opened 2015-08-04
fhcount: this plugin only works on Linux

rubenk:


mfournier:
merged as bae4500a00862, thanks again @rubenk !

PR #1197, opened 2015-08-04
src/daemon/plugin.c: delay first plugin read

rezib:
This commit delays the first plugin read to the beginning of the next
read interval. This way, the plugin reads are more or less synchronized
accross multiple nodes running collectd with the same read interval.
This helps comparing metrics shared by multiple nodes, such as network
bandwidth and so on.

Please note that it only works if the system clocks of these nodes are
properly synchronized using NTP (or similar) but this is generally (and
hopefully) the case on large server infrastructures.


mfournier:
Thanks for this patch @rezib !

I can imagine the opposite use case could be desirable too: randomly spread reads among the nodes (which is more or less what happens now). Would it make sense to have a global config option to let users control which behaviour they prefer ? What do you think ?

rezib:
That's right! I can rework my patch to add a global parameter (any name suggestion?) to control this.

I guess the default behaviour should be the current one (spread), right?

rezib:
I just updated my commit with new configuration parameter `SyncNodesReads true|false` (not really happy with this name but I suck at name) with default `false` value. I also rebased it on top of current master branch. Feel free to tell if anything is wrong :)

faxm0dem:
we have a cluster of servers where this behaviour is desired. Any news on what keeps this from being merged?

PR #1198, opened 2015-08-04
New slurm read plugins (do not necessarily merge)

rezib:
[Slurm](https://github.com/SchedMD/slurm) is a job scheduler and workload manager that is very popular among the HPC community.

This pull request proposes 2 new read plugins to get metrics out of Slurm:
- The `slurmctld` plugin which sends RPC to `slurmctld` daemon to get the number of CPU and nodes allocated to jobs. It relies on the `libslurm` library. Please note this plugin requires the `Globals true` parameter since the `libslurm` library also load a 2nd row of plugins with `dlopen()`. This plugin is designed to run on Slurm job scheduler server of the supercomputer.
- The `slurmd` plugin which browse Slurm job specifics (cpuset) cgroups to get the jiffies of the CPU allocated to jobs and do the sum of memory PSS of all PIDs of the job. This plugin requires the `TaskPlugin=task/cgroup` to be enable in Slurm configuration file. This plugin is desgined to run on all compute nodes of a supercomputer.

My intent is not necessarily to merge these plugins in upstream collectd repository (unless you would be happy to do so), but I would be pleased to get code review from peers and people used to develop such collectd plugins :)

I've tried to follow the [development guidelines](https://collectd.org/wiki/index.php/Coding_style) and re-use existing functions as much as possible. As I am not an experienced C developer very familiar with portability considerations (only tested on Linux and amd64), I would be pleased to get feedback on this matter too!


rezib:
I just fixed my commit for x86_32 and rebased it on top on current master branch.

lukeyeager:
Can probably close this since https://github.com/collectd/collectd/pull/3037 was merged.

mrunge:
Yes, thank you all involved here. I'm closing this, since #3037 was merged.

PR #1200, opened 2015-08-04
Only warn of callback overwrite if data changes

michaelsalmon:
This patch removes the warning issued when a callback is updated if the new callback and data are the same as the old ones.

### DIFF
--- a/src/plugin.c
+++ b/src/plugin.c
@@ -228,7 +228,7 @@ static int register_callback (llist_t **list, /* {{{ */
 		{
 			ERROR ("plugin: register_callback: "
 					"llentry_create failed.");
-			free (key);
+			sfree (key);
 			destroy_callback (cf);
 			return (-1);
 		}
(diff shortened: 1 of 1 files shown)


src/plugin.c
```diff
@@ -242,9 +242,15 @@ static int register_callback (llist_t **list, /* {{{ */
 		old_cf = le->value;
 		le->value = cf;
 
-		WARNING ("plugin: register_callback: "
-				"a callback named `%s' already exists - "
-				"overwriting the old entry!", name);
+		if (
+			old_cf->cf_callback != cf->cf_callback ||
+			old_cf->cf_udata.data != cf->cf_udata.data
+		)
```
tokkee: We don't usually have linebreaks after the opening or before the closing parenthesis:

    if ((old_cf->cf_callback != cf->cf_callback) || (...))

tokkee:
So, I'm unsure if that's what we want. Generally, we want to know about such cases and fix the caller. However, at the same time, the check doesn't hurt much and lets us not worry about this in the first place. I slightly tend towards the latter. Other opinions?

octo:
I'm leaning towards "if this happens, we want to know about it", i.e. print the warning unconditionally.

rpv-tomsk:
Hi all,

@octo, let's close this with 'wontfix' ?

octo:
Agreed.

PR #1201, opened 2015-08-04
yajl: look for libyajl in /usr/local/lib as well

rubenk:


pyr:
When libraries are in non-standard locations, they should be referenced with flags, otherwise we will end-up hardcoding non-standard locations everywhere.

rubenk:
You make a good point,. I think /usr/local/lib can be considered a standard location though, at least on FreeBSD since it's the only place ports and packages install their libs.

What I can do instead is add /usr/local/include to the default CFLAGS on FreeBSD and /usr/local/lib to LDFLAGS, this would solve it for all plugins. Let me know what you think.

pyr:
@rubenk I think this is more sensible, yes.

rubenk:
Specifying the flags globally doesn't work properly since we override them here and there. Let's drop this for now.

PR #1202, opened 2015-08-04
mqtt: look for libmosquitto in /usr/local/lib as well

rubenk:


pyr:
Same comment than for #1201

PR #1205, opened 2015-08-06
Solaris zfs changes

btoneill:
This gives more ZFS information that is useful

### DIFF
--- a/src/types.db
+++ b/src/types.db
@@ -14,7 +14,7 @@ cache_eviction		value:DERIVE:0:U
 cache_operation		value:DERIVE:0:U
 cache_ratio		value:GAUGE:0:100
 cache_result		value:DERIVE:0:U
-cache_size		value:GAUGE:0:U
+cache_size		value:GAUGE:0:1125899906842623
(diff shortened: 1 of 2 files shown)


src/types.db
```diff
@@ -14,7 +14,7 @@ cache_eviction		value:DERIVE:0:U
 cache_operation		value:DERIVE:0:U
 cache_ratio		value:GAUGE:0:100
 cache_result		value:DERIVE:0:U
-cache_size		value:GAUGE:0:U
+cache_size		value:GAUGE:0:1125899906842623
```
mfournier: Why change this ?

btoneill: The cache size value of U is a 32bit unsigned int. If you have a large
memory system, you overrun it and get no values. We ran into that with a
system that had 256G of memory.

btoneill: I picked that value as it was used in other places where you need a 64bit
counter....

mfournier:
Many thanks @btoneill ! I just double-checked these values are also present on FreeBSD and ZFS-on-Linux, so hopefully we're safe regarding portability.

PR #1206, opened 2015-08-06
added ZFS ARC cache to memory, treating it like Linux cached memory

btoneill:
Makes the memory used more useful by putting in ARC size

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -76,6 +76,7 @@ static vm_size_t pagesize;
 #elif HAVE_LIBKSTAT
 static int pagesize;
 static kstat_t *ksp;
+static kstat_t *ksz;
 /* #endif HAVE_LIBKSTAT */
 
 #elif HAVE_SYSCTL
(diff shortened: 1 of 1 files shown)


mfournier:
@btoneill the idea is that all the values passed to `MEMORY_SUBMIT ()` sum up to 100% of the memory available on the system. Does this still hold with this change ?

Also, there seems to be a couple of indentation inconsistencies introduced by your patch.

btoneill:
Yeah, it all sums up to 100%. It simply replaces the part of "kernel"
memory that is used by the ZFS arc cache and displays it separately.
Otherwise, you end up with graphs that have the kernel using 160G of memory
(which happens to us on ZFS NFS fileservers that have 196G of memory).

mfournier:
Thanks a lot @btoneill ! I merged your patch as d13ba64e, it will be part of version 5.6.

@dago, as I'm aware you've been taking good care of collectd on Solaris lately, could you please let us know if there's any problem with this change ? Thanks !

dago:
Sure. Unfortunately I am currently busy with other things. I'll see what I can do.

PR #1208, opened 2015-08-06
Fix void* arithmetic warning (-Wpointer-arith)

jy2wong:
https://gcc.gnu.org/onlinedocs/gcc/Pointer-Arith.html
gcc will assume sizeof(void) == 1 when doing pointer arithmetic on
void*s, but other compilers may not.


rubenk:
Thanks @jy2wong !

PR #1211, opened 2015-08-09
ltdl: only compile included libltdl when needed

rubenk:
If configure finds an external libltdl, it's going to
use it, so there's no use in compiling the shipped libltdl.


mfournier:
Merged as 2a8e9257d. Thanks @rubenk !

PR #855, opened 2014-12-10
Callback split: read_cpuacct_procs

MisterG:
The callback "read_cpuacct_procs" used to read the ".stat" file by itself.

This commit splits this callback into two parts:
* read_cpuacct_procs: creates the full path of the file to read
* process_cgroup_file: reads the given file and processes the metrics

It will allow people to add other metrics than CPU accounting. For example:
* disk I/O
* memory

More pull requests coming soon...

PR #1213, opened 2015-08-10
cgroups plugin: add memory accounting

mgrzybek:
# History

My previous work with branches and pull requests was a big mess (#855 and #999).
I started it from scratch to be « pull request » ready.
# Features

This commit patches the cgroup plugin to allow both CPU and memory accounting.
It also adds the required types to the database.

### DIFF
--- a/README
+++ b/README
@@ -49,7 +49,7 @@ Features
       Statistics from the Ceph distributed storage system.
 
     - cgroups
-      CPU accounting information for process groups under Linux.
+      CPU and memory accounting information for process groups under Linux.
 
     - conntrack
       Number of nf_conntrack entries.
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1304,9 +1304,10 @@ Specifies the path to the UNIX admin socket of the ceph daemon.
 
 =head2 Plugin C<cgroups>
 
-This plugin collects the CPU user/system time for each I<cgroup> by reading the
-F<cpuacct.stat> files in the first cpuacct-mountpoint (typically
-F</sys/fs/cgroup/cpu.cpuacct> on machines using systemd).
+This plugin collects the CPU user/system time and/or memory usage for each
+I<cgroup> by reading the statistics file, ie. F<cpuacct.stat> or
+F<memory.stat>. These files are typically read from the first cgroup mountpoint
+(F</sys/fs/cgroup/> on machines using systemd).
 
 =over 4
 
@@ -1322,6 +1323,11 @@ Invert the selection: If set to true, all cgroups I<except> the ones that
 match any one of the criteria are collected. By default only selected
 cgroups are collected if a selection is made. If no selection is configured
 at all, B<all> cgroups are selected.
+
+=item B<Metric>
+
+Select the metric to monitor: B<cpu> or B<memory>. You can use this keyword
+several times.
 
 =back
 


rubenk:
Can you please change the commit message from "Branch rebuilding" to something more descriptive?

MisterG:
That's Done.

rubenk:
Not the title of the Pull Request, the commit message.

MisterG:
This time the commit title has been changed.
Sorry for the inconvenience...

rubenk:
Thank you. Can you update the collectd.conf manpage as well?

MisterG:
POD file and README updated.

rpv-tomsk:
This PR is one I will have interest in nearest time.

rpv-tomsk:
I'm unable to understand why `src/types.db` needs to be patched. As I understand, that types are not required.

rpv-tomsk:
The issue found:  current code does not support hierarhy.
I haven't used cpu* cgroups yet, but for `memory` there is hierarhy support exists.

How to deal?

rubenk:
@rpv-tomsk do you mean a single hierarchy, or multiple? cgroups are moving away from multiple hierarchies (https://www.kernel.org/doc/Documentation/cgroup-v2.txt)

rpv-tomsk:
On my test server I'm using `memory.use_hierarchy` on, and the hierarchy is the following:

/ 
/daemons/
/mysql/
/users/  (with common limit set for this cgroup and its descendants)
/users/user1/
/users/user2/
/users/user3/  (some users have their own limits)
....

As I understand, this is single heararchy, right ?

rubenk:
I don't know. It's probably explained in the document I linked to.

rpv-tomsk:
Another question.  Let`s look into example of `memory.stat` file:

`
# cat /sys/fs/cgroup/memory/memory.stat
cache 132964352
rss 443363328
mapped_file 63127552
pgpgin 2887140
pgpgout 2746435
swap 341966848
pgfault 4782593
pgmajfault 8505
inactive_anon 18022400
active_anon 458600448
inactive_file 38305792
active_file 61394944
unevictable 0
hierarchical_memory_limit 9223372036854775807
hierarchical_memsw_limit 9223372036854775807
total_cache 132964352
total_rss 443363328
total_mapped_file 63127552
total_pgpgin 2887140
total_pgpgout 2746435
total_swap 341966848
total_pgfault 4782593
total_pgmajfault 8505
total_inactive_anon 18022400
total_active_anon 458600448
total_inactive_file 38305792
total_active_file 61394944
total_unevictable 0
`

I don't think each of this lines must be submitted as type instance of single type. 
And not each line must be submitted at all.

I think basic output must be similar to `memory` plugin `memory` type and type instances, with (may be) additional types/metrics.

mgrzybek:
Two questions:
- Do you want me to patch the code to support cgroup v2?
- Do you want me to deal with @rpv-tomsk comments?

octo:
Hi @mgrzybek,

thank you very much for the patch! The biggest problem with this change is the way metrics are reported: all metric values are in one "type" that has very many "data sources". Our best practice is to use "types" with *one* "data source" each. That's what @rpv-tomsk points out above and needs to be fixed before we can merge.

What we need to do here is to go through the metrics available and find a good "type" for each. A "type" is similar to a *unit* of measurement, so there can be multiple metrics of the same type, for example "cache" and "rss" appear to both report bytes (or pages) in memory. When there are multiple metrics with the same "type", use the "type instance" to distinguish the two, for example:

* cache → `{ .type = "memory", .type_instance = "cache" }`
* rss → `{ .type = "memory", .type_instance = "rss" }`

I'd propose to use *Google Sheets* to come up with a good mapping but Github's Markdown tables would also work.

Best regards,
—octo

PR #1221, opened 2015-08-14
Fix harmless build warnings with newer binutils

rubenk:
I see a lot of those on Rawhide:
ar: `u' modifier ignored since `D' is the default (see `U')

It's going to take a while to get this fixed upstream.
Meanwhile, silence the warnings.

Macro copied from libvirt.


mfournier:
Thanks @rubenk ! Merged as 9ab0a540f

PR #1222, opened 2015-08-14
Make building with -Werror optional.

rubenk:
I noticed various downstream distros patch out -Werror
for one reason or another. This adds an option to make that
easier. It is still on by default.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5910,6 +5910,20 @@ AC_SUBST(LCC_VERSION_STRING)
 
 AC_CONFIG_FILES(src/libcollectdclient/collectd/lcc_features.h)
 
+AM_CFLAGS="-Wall"
+AC_ARG_ENABLE([werror],
+        AS_HELP_STRING(
+                [--disable-werror],
+                [Disable -Werror @<:@default=no@:>@)]
+        )
+)
+
+if test "x$enable_werror" != "xno"
+then
+        AM_CFLAGS="$AM_CFLAGS -Werror"
+fi
+AC_SUBST([AM_CFLAGS])
+
 AC_CONFIG_FILES([Makefile src/Makefile src/daemon/Makefile src/collectd.conf src/libcollectdclient/Makefile src/libcollectdclient/libcollectdclient.pc src/liboconfig/Makefile bindings/Makefile bindings/java/Makefile])
 AC_OUTPUT
 
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -6,10 +6,6 @@ SUBDIRS += daemon
 
 PLUGIN_LDFLAGS = -module -avoid-version -export-symbols-regex '\<module_register\>'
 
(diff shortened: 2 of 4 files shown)


mfournier:
@rubenk I held back this patch because I wanted to doube-check a couple of things.

One important change this patch introduces is that `-Werror` is now the default for every compiler, not only gcc. So this will warrant at least a mention in the changelog. That said, building current master with clang and gcc pass with `-Werror` on major distros, so this should go unnoticed. I'm worried about solaris users though, but at least the `--disable-werror` would be available for them.

Moving the `CFLAGS` definition to configure.ac was another concern I had. I don't want the configure checks to depend on `-Werror`, as the code it builds is out of our control. But it doesn't seem to be the case: `AM_CFLAGS` doesn't get expanded at this stage.

Please also take a look at the changes I added in #1260.

That said, I'd rather see bugreports from people encoutering build failures due to `-Werror` than having everybody build with `--disable-werror` and move on with life. As already discussed, one task of our jenkins could be to test master on debian-sid and fedora-rawhide so we can anticipate changes to the toolchain, etc, before releases.

rubenk:
Thanks for the review @mfournier!.

I was wondering why -Werror was the default for GCC only.
I suspect that when @octo wrote this check clang wasn't around yet, and he didn't want to enable it for Sun Studio.

If this causes problems for specific compilers, we can always toggle the default for those, but I'd like to find out which ones they are :)

AM_CFLAGS indeed doesn't effect the configure checks.

The problem with -Werror by default is that this is ok to catch issues on a CI platform, but in practice for downstream distros it is more trouble than it's worth and they patch it out. I've seen collectd builds fail due to warnings in perl, gcrypt and other libs for instance which were turned into errors, something which is out of collectds control. I look at those build logs and report bugs upstream but collectd has to keep on building. As you know it has always been possible for users to disable Werror by adjusting their CFLAGS, if they really wanted to. This change makes it just a bit more convenient and saves them from having to run autoreconf etc.

I even thought about disabling -Werror by default but I suspected I'd get some resistance ;)
For some interesting perspective on this see https://blog.flameeyes.eu/2009/02/future-proof-your-code-dont-use-werror

As for catching issues on the CI infra with -Werror, I noticed that they are running `make --keep-going` anyway. I hope you have plans to change that?

mfournier:
> As for catching issues on the CI infra with -Werror, I noticed that they are running make --keep-going anyway. I hope you have plans to change that?

What makes you frown upon this @rubenk ?

My reasoning was that `make -k` will allow jenkins to show us all the build errors, not only the first one. The exit status will be non-zero even with the -k switch, in case of errors, and the resulting binaries will never get run, so we don't really care if they're half broken.

mfournier:
Merged via #1260. Thanks !

rubenk:
@mfournier ah sorry, I missed the fact that make, even if it keeps going, sets the exit-status non-zero.

PR #1223, opened 2015-08-16
liboconfig: Fix implicit declaration warning

rubenk:
strlen declaration is in string.h


mfournier:
Merged as 4cee263e4f.

PR #1226, opened 2015-08-16
pf: fix compilation on OpenBSD

rubenk:


mfournier:
Merged as 79032bd1b8a.

PR #1228, opened 2015-08-16
dns: fix compilation on OpenBSD

rubenk:
OpenBSD doesn't have pcap-bpf.h

pcap.h has been including pcap/bpf.h since 2006.
Since we require a pcap which has PCAP_ERROR_IFACE_NOT_UP, introduced in 2008
this shouldn't break anything.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3292,11 +3292,6 @@ then
 fi
 if test "x$with_libpcap" = "xyes"
 then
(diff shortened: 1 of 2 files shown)


mfournier:
Merged as f5b909a469.

Thanks again @rubenk for all this small but valuables details you take care about !

PR #1229, opened 2015-08-17
Allow to specify per database interval for fetching stats

mbebjak:


src/dbi.c
```diff
@@ -407,7 +414,7 @@ static int cdbi_config_add_database (oconfig_item_t *ci) /* {{{ */
       plugin_register_complex_read (/* group = */ NULL,
           /* name = */ name ? name : db->name,
           /* callback = */ cdbi_read_database,
-          /* interval = */ 0,
+          /* interval = */ (db->interval > 0) ? &interval : NULL,
```
mfournier: Shouldn't this read as:
`(db->interval > 0) ? &interval : 0,`
instead ? ie: keep the default 0.

mfournier:
Thanks a lot for this patch @mbebjak ! This seems pretty sane. Could you please mention the new parameter in the collectd.conf manpage: `src/collectd.conf.pod` ? Thanks !

mfournier:
Mmmh, actually the following error gets reported at build time.
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I./daemon -DPREFIX=\"/opt/collectd\" -DCONFIGFILE=\"/opt/collectd/etc/collectd.conf\" -DLOCALSTATEDIR=\"/opt/collectd/var\" -DPKGLOCALSTATEDIR=\"/opt/collectd/var/lib/collectd\" -DPIDFILE=\"/opt/collectd/var/run/collectd.pid\" -DPLUGINDIR=\"/opt/collectd/lib/collectd\" -DPKGDATADIR=\"/opt/collectd/share/collectd\" -Wall -Werror -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wall -Wno-error=deprecated-declarations -MT dbi_la-dbi.lo -MD -MP -MF .deps/dbi_la-dbi.Tpo -c dbi.c  -fPIC -DPIC -o .libs/dbi_la-dbi.o
dbi.c: In function 'cdbi_config_add_database':
dbi.c:417:28: error: passing argument 4 of 'plugin_register_complex_read' makes integer from pointer without a cast [-Werror]
           /* interval = */ (db->interval > 0) ? &interval : NULL,
                            ^
In file included from ./daemon/common.h:32:0,
                 from dbi.c:28:
./daemon/plugin.h:296:5: note: expected 'cdtime_t' but argument is of type 'struct timespec *'
 int plugin_register_complex_read (const char *group, const char *name,
     ^
Makefile:4107: recipe for target 'dbi_la-dbi.lo' failed
make[3]: *** [dbi_la-dbi.lo] Error 1
```

mbebjak:
@mfournier thanks for the review. I've fixed the issues you pointed out. and added documentation to the `src/collectd.conf.pod`

btw: problem was that I initially created this for the collectd-5.5 branch which I use now and did not adjust it for the master which has different `plugin_register_complex_read` method

mfournier:
Thanks for the update @mbebjak ! Your patch was merged and this new feature will be part of upcoming 5.6 release. Many thanks !

PR #1245, opened 2015-08-26
src/daemon/utils_time_test.c: add ULL suffix to large literals

mfournier:
This should prevent some 32bit compilers from complaining about:
`integer constant is too large for 'long' type`


octo:
Thanks for taking care of this Marc!

PR #1246, opened 2015-08-28
Log load plugin

mdubbyap:
Log plugins as they get loaded. Debugging what is and isn't loaded is sometimes difficult, this solves that.

### DIFF
--- a/src/daemon/configfile.c
+++ b/src/daemon/configfile.c
@@ -286,6 +286,7 @@ static int dispatch_loadplugin (oconfig_item_t *ci)
 	name = ci->values[0].value.string;
 	if (strcmp ("libvirt", name) == 0)
 		name = "virt";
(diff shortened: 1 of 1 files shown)


mdubbyap:
Any reason this hasn't had a build?

mfournier:
I also find this would be useful, thanks for coming forward with this suggestion @mdubbyap !

This implementation doesn't handle the plugins loaded automatically though (`AutoLoadPlugin true`), and unconditionnally emits a log message, even if the plugin fails to load (NB: each condition in which plugins fail to load already emits a log message).

Please take a look at #1254: same intent, but directly in the function that loads the plugins, rather than in the function handling the configuration.

mdubbyap:
Closing as @mfournier submitted a better solution.

PR #1254, opened 2015-09-03
plugin.c: emit log message when plugin loading succeeds

mfournier:


rubenk:
:+1:

mfournier:
So this is simply a counterproposal to #1246, which has a couple of issues.

Collectd startup with the logfile plugin setup to send messages to STDOUT looks like this:
```
$ /opt/collectd/sbin/collectd -t -T                                                                           
option = ReadThreads; value = -1;
Created new plugin context.
[2015-09-03 19:20:57] type = logfile, key = Timestamp, value = true
[2015-09-03 19:20:57] type = logfile, key = PrintSeverity, value = true
[2015-09-03 19:20:57] [info] plugin_load: plugin "cpu" successfully loaded.
[2015-09-03 19:20:57] [error] plugin_load: Could not find plugin "foobar" in /opt/collectd/lib/collectd
[2015-09-03 19:20:57] [info] plugin_load: plugin "csv" successfully loaded.
[2015-09-03 19:20:57] [info] plugin_load: plugin "interface" successfully loaded.
[2015-09-03 19:20:57] [info] plugin_load: plugin "load" successfully loaded.
[2015-09-03 19:20:57] [info] plugin_load: plugin "memory" successfully loaded.
[2015-09-03 19:20:57] [error] lt_dlopen ("/opt/collectd/lib/collectd/dummy.so") failed: file not found. The most common cause for this problem is missing dependencies. Use ldd(1) to check the dependencies of the plugin / shared object.
ERROR: lt_dlopen ("/opt/collectd/lib/collectd/dummy.so") failed: file not found. The most common cause for this problem is missing dependencies. Use ldd(1) to check the dependencies of the plugin / shared object.
[2015-09-03 19:20:57] [error] plugin_load: Load plugin "dummy" failed with status 1.
```
(a couple of errors introduced on purpose)

BCWilsonDotCom:
+1

mdubbyap:
Love it.  Thanks @mfournier ! I'll close the original.

PR #1256, opened 2015-09-03
Allow MySQL to use SSL connections

blalor:
This adds configuration options to allow the MySQL library to make SSL connections to the server.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -702,6 +702,11 @@
 #		User "db_user"
 #		Password "secret"
 #		Database "db_name"
+#		SSLKey "/path/to/key.pem"
+#		SSLCert "/path/to/cert.pem"
+#		SSLCA "/path/to/ca.pem"
+#		SSLCAPath "/path/to/cas/"
+#		SSLCipher "DHE-RSA-AES256-SHA"
 #		MasterStats true
 #		ConnectTimeout 10
 #		InnodbStats true
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3376,6 +3376,11 @@ Synopsis:
       Port "3306"
       MasterStats true
       ConnectTimeout 10
+      SSLKey "/path/to/key.pem"
+      SSLCert "/path/to/cert.pem"
+      SSLCA "/path/to/ca.pem"
+      SSLCAPath "/path/to/cas/"
+      SSLCipher "DHE-RSA-AES256-SHA"
     </Database>
 
     <Database bar>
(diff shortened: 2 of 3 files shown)


octo:
Thank you very much for your patch @blalor, it looks great! Especially thanks for taking care of the documentation, too!

One minor thought: could we re-use Apache's names for these settings? SSL setups profit especially when users are already familiar with the available options. The names would be:

*   SSLCACertificateFile
*   SSLCACertificatePath
*   SSLCertificateFile
*   SSLCertificateKeyFile
*   SSLCipherSuite

What do you think?

blalor:
The reason I used these parameter names is because they map directly (with capitalization normalization) to the command-line flags for the `mysql` binary as well as the arguments to the C `mysql_ssl_set params` function.  I agree with consistency, but I think keeping these consistent with the MySQL world will make it easier for people trying to port working `mysql` commands over to collectd.  That's how I approached it, anyway. :-)

rubenk:
Not sure if we already support it or not, but it would also be great if we could just use these settings from my.cnf.

blalor:
@rubenk I think that would greatly increase the scope and complexity of this change.  Are there other places where collectd reads another application's config?

rubenk:
The mysql api supports reading defaults from my.cnf / .my.cnf so it shouldn't be too hard.
I think that would make sense instead of duplicating a lot of mysql client config settings in collectd.

blalor:
I don’t think it’s safe to assume that file will always exist.  My current use case is monitoring Google’s CloudSQL and Amazon’s RDS from a host (or container) that only runs collectd.

blalor:
@octo thoughts on my reply above?

rubenk:
@blalor if .my.cnf doesn't exist on your host or container, why can't you create it?

blalor:
Having collectd depend on another application's configuration file feels awkward, and I don't see any other plugins that delegate their configuration this way.  Additionally, this plugin currently supports multiple database connections, so at the very least it would have to be extended to load its configuration from an explicit path.

The intent of this PR is very narrow in scope: allow the existing plugin to work with secured connections.  I don't want to fall down the rabbit hole of making changes to the plugin that go beyond this scope.

@octo you're assigned; what's the process for getting this merged in?  I don't want to maintain my fork forever.

blalor:
@mfournier I see you changed the milestone.  What's that mean for the inclusion of this PR into a release?

mfournier:
@blalor I've just done a bit of triaging, as I'm currently focusing on getting a bugfix release ready. As your patch *adds* a new feature, it's not a bugfix ;-) It will hopefully make it into version 5.6 though.

octo:
We're using `mysql_init()`, which appears to be reading configuration from a couple of places:

http://dev.mysql.com/doc/refman/5.7/en/mysql-library-init.html:
> For convenience, if the groups argument itself is `NULL`, the [server] and [embedded] groups are used by default.

We should verify that this is actually happening and then add a note to the config.

octo:
Rebased on *master* and committed as 680c818999f50daa1ad3b80317b101bf0205c441

PR #1259, opened 2015-09-07
Feature/mongodb plugin/kapil

srikanthops:
Mongo Support  with libmongo


rubenk:
This PR is very hard to review, since it contains so many unrelated changes. Vendoring libmongoc is also not a good idea. If you still like to persue this, please open a new PR.

PR #1260, opened 2015-09-08
-Werror build flag handling + build options printout

mfournier:
This builds on top of #1222.


rubenk:
Thanks for those @mfournier, they look good to me.

PR #1263, opened 2015-09-08
Fix copy-paste mistake in debug message in snmp plugin

berkovskyy:


mfournier:
Merged as 3e4e90b, thanks a lot @berkovskyy !

PR #1264, opened 2015-09-08
Systemd

rubenk:


src/daemon/collectd.c
```diff
@@ -487,11 +484,6 @@ int notify_systemd (void)
         su_size = sizeof (sa_family_t) + strlen (notifysocket);
         if (su_size > sizeof (su))
             su_size = sizeof (su);
-#else
-	ERROR ("Systemd socket uses Linux abstract namespace notation (\"%s\"), "
```
octo: What's the reason for removing this fallback? ("systemd is runs only on Linux anyway" is not a valid answer ;)

mfournier: @octo 543a57f8aff added a more global way of preventing non-linux systems from running through the systemd-specific code path. So this is indeed dead code. I agree with @rubenk it can be removed.

octo: Right, we don't need two nested `#if KERNEL_LINUX` …

src/daemon/collectd.c
```diff
@@ -457,9 +457,7 @@ int notify_systemd (void)
     }
     NOTICE ("Systemd detected, trying to signal readyness.");
 
-    unsetenv ("NOTIFY_SOCKET");
-
-    fd = socket (AF_UNIX, SOCK_DGRAM, /* protocol = */ 0);
+    fd = socket (AF_UNIX, SOCK_DGRAM|SOCK_CLOEXEC, /* protocol = */ 0);
```
octo: This fails on EPEL5 (both i386 and x86-64) with:

```
collectd.c:460: error: 'SOCK_CLOEXEC' undeclared (first use in this function)
```

Annoying but I guess we need to `#ifdef` around this …

rubenk: Ah damn, EPEL 5 really is getting old... I'll fix this up.

mfournier:
Thanks a lot @rubenk !

FYI I also squashed all the changes to systemd.collectd.service and commited them to 5.5 (4fd7396fb).

PR #1287, opened 2015-10-03
powerdns plugin: normalize dns_question name for udp4/6

m8t:
In order to normalize the views in CGP, rename queries-udp[46] to udp[46]

collectd-powerdns-dns_question-dns_answer


mfournier:
Well this issue here IMHO is backwards compatibility for all those *not* using CGP.

rubenk:
I agree with @mfournier, this will cause issues for everyone using a different graphing tool.
Perhaps you can use the replace target to change the name (https://collectd.org/wiki/index.php/Target:Replace)

PR #1295, opened 2015-10-07
'Modification of a read-only value attempted'  in Collectd::Unixsock::getthreshold()

ciomaire:
Store the first match in $key and trim whitespace from that rather than $1


mfournier:
Nice catch @ciomaire, thanks a lot !

Commited as 159083153.

PR #1298, opened 2015-10-08
processes: correctly count processes on {free,open}BSD

vr:
because actually we count threads instead of processes

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -2027,18 +2027,18 @@ static int ps_read (void)
 			pse.io_syscw = -1;
 
 			ps_list_add (procs[i].ki_comm, have_cmdline ? cmdline : NULL, &pse);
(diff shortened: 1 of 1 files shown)


mfournier:
So this fixes an issue which seems to date back in 4fdf004a58f, and which was copied in the openbsd code block too.

@brd, @landryb would you mind taking a look please ? Thanks !

landryb:
I havent tested it at runtime but this seems right to me. According to http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man3/kvm_getargv.3 kvm_getprocs() returns all threads, and if we'd want to only count processes we could filter on thread id being -1 (but that's another story)

```
kvm_getprocs() sets the thread ID field accordingly for each thread
except for the process (main thread) which has it set to -1.
```

mfournier:
This fix has been merged as 85c31a6 (and backported for the 5.4 line as 15eefc2d).

Many thanks for figuring this out @vr !

PR #1303, opened 2015-10-14
disk: convert multi DS to type_instance

LuizOz:
Converting disk plugin from multi DataSources values into type_instances.

Aggregation plugin now works with the disk plugin.

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -269,26 +269,27 @@ static int disk_init (void)
 } /* int disk_init */
 
 static void disk_submit (const char *plugin_instance,
-		const char *type,
-		derive_t read, derive_t write)
+               const char *type,
+               const char *operation,
+               derive_t val)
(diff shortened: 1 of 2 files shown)


mfournier:
Well the big issue with this patch is backwards compatibility, of course.

What about modifying the aggregation plugin to support multi-datasource values instead ?

In any case, thanks for bringing this forward @LuizOz !

mfournier:
@LuizOz actually #481 seems to do exactly that. Maybe would you like to give it a try ? If you do, please leave some feedback !

LuizOz:
@mfournier Nice, I'm going to take a look at https://github.com/collectd/collectd/pull/481. Aggregation to all plugins, much better

rpv-tomsk:
Propose to close this PR in favour of #481 PR.

PR #1309, opened 2015-10-17
Fix TIMESPEC_TO_CDTIME_T() on 32-bit arch.

baryonix:
Add ULL suffix to constant to ensure we get a 64-bit unsigned multiply.

Currently, on 32-bit x86 at least we get incorrect results from cdtime() resulting in an infinite loop, calling the read plugins over and over with no delay inbetween.

### DIFF
--- a/src/daemon/utils_time.h
+++ b/src/daemon/utils_time.h
@@ -72,7 +72,7 @@
   (tsp)->tv_sec = CDTIME_T_TO_TIME_T (cdt);                                  \
   (tsp)->tv_nsec = (long) CDTIME_T_TO_NS ((cdt) & 0x3fffffff);               \
 } while (0)
-#define TIMESPEC_TO_CDTIME_T(ts) NS_TO_CDTIME_T(1000000000 * (ts)->tv_sec + (ts)->tv_nsec)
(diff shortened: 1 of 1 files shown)


mfournier:
Thanks for figuring this out @baryonix !

So you noticed this using current master, right ? This problem would have been introduced in db1391aaa ?

While we're on this topic, would it make sense to also suffix those in `NS_TO_CDTIME_T()` and `CDTIME_T_TO_NS()` a couple of lines above ?

baryonix:
Yes, I found this in current master and I think it was introduced in the commit you mention.

As far as I can see `NS_TO_CDTIME_T()` is fine as it uses explicit casts to `cdtime_t` (which is known to be a 64-bit type), and `CDTIME_T_TO_NS()` and friends were already fixed in 45e631cf6134de4a5dd3bcf01794c9fdd539c412 by explicitly casting to `uint64_t`, so we should be safe. (A quick empirical test seems to confirm this.)

mfournier:
Thanks so much for fixing this @baryonix ! I guess this wasn't an easy bug to track down :-/ Hats off !

Issue #1282, opened 2015-09-26
Statsd plugin only sends counter rate not the actual count


jamtur01:
If I look at https://github.com/collectd/collectd/blob/master/src/statsd.c#L264 then it appears that the statsd plugin only sends the rate metric when sending a counter.

If I send:

```bash
for i in {1..1000000}; do echo "productiona.app1-web1.counter:$i|c" | nc -u -w0 127.0.0.1 8125; done
```

Then I'd expect to see in statsd:

```json
Flushing stats at  Sat Sep 26 2015 10:55:41 GMT-0400 (EDT)
{ counters:
   { 'statsd.bad_lines_seen': 0,
     'statsd.packets_received': 1247,
     'statsd.metrics_received': 1247,
     'productiona.app1-web1.counter': 26624697 },
  timers: {},
  gauges: { 'statsd.timestamp_lag': 0 },
  timer_data: {},
  counter_rates:
   { 'statsd.bad_lines_seen': 0,
     'statsd.packets_received': 124.7,
     'statsd.metrics_received': 124.7,
     'productiona.app1-web1.counter': 2662469.7 },
  sets: {},
  pctThreshold: [ 90 ] }
```

With the statsd plugin I see:

```json
:host app1-web1, :service statsd/derive-productiona.app1-web1.counter, :state ok, :description nil, :metric 2662469.7, :tags [collectd], :time 1443279411, :ttl 60.0, :ds_index 0, :ds_name value, :ds_type derive:rate, :type_instance productiona.app1-web1.counter, :type derive, :plugin statsd}
```

Which is just the rate and NOT the actual count. 

According to the StatsD docs for counters: `Statsd will send both the rate as well as the count at each flush.` (https://github.com/etsy/statsd/blob/master/docs/metric_types.md)

Am I missing something here?

pyr:
If you use `StoreRates false` in your riemann configuration, the metrics emitted will be absolute counts, as requested initially. If for any reason you want to mix rates and absolute counters, the only way would be to have a gauge emitted as well.

More generally, collectd is moving away from counters and towards only rates, gauges and timers.
Can you let me know if the code works for you, then @octo can weigh-in on his view of #1311.

My take on is is that in "standard" read plugins this should be avoided. In this case, since there are cases where collectd is used as a vessel between different systems, it might make sense to forward the absolute counters. I'm thinking of the scenario which you seem to be working with:

```
<app> -> <local statsd> -> <local collectd> -> <global riemann>
```

[...]

PR #1311, opened 2015-10-19
statsd: provide a gauge for counters, holding the absolute count.

pyr:
When not using StoreRates false, this allows statsd aggregators
to hold on to the absolute count, if needed. This provides a way of dealing with use-cases such as #1282

### DIFF
--- a/src/statsd.c
+++ b/src/statsd.c
@@ -850,8 +850,19 @@ static int statsd_metric_submit_unsafe (char const *name, /* {{{ */
     else
       values[0].gauge = (gauge_t) c_avl_size (metric->set);
   }
-  else
-    values[0].derive = (derive_t) metric->value;
+  else { /* STATSD_COUNTER */
+      /*
+       * Expand a single value to two metrics:
+       *
+       * - The absolute counter, as a gauge
+       * - A derived rate for this counter
+       */
+      values[0].derive = (derive_t) metric->value;
+      plugin_dispatch_values(&vl);
+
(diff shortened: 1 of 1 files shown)


mfournier:
Hi @pyr ! Don't worry about the failing test above. Currently moving jenkins to digitalocean's v2 API and it's not going as well as I'd like... I'll trigger the job manually once everything is working properly again.

pyr:
Hi @mfournier,

Thanks for the heads-up! No worries :-) I'll watch this space.

pyr:
@octo @mfournier any thoughts on the code and rationale behind it? (#1282 expands on it). To reiterate, I know we are moving away from sending raw counters as much as possible. @jamtur01's use case could warrant making an exception in this case.

octo:
Sending the count as a `GAUGE` doesn't make any sense for *collectd*.

Here is what etsy's statsd is doing:
 *  Calculates a sum and the number of counters received.
 *  On flush, it calculates `rate = sum / num` and writes `rate` and `num`.
 *  **Then it resets sum and num to zero.**

With `DERIVE` metrics, you need to keep some state (i.e. the old value) but then the same information is available:
 *  `num = value_new - value_old`
 *  `rate = (value_new - value_old) / (time_new - time_old)`

What this patch is publishing is `value_new` (as a `GAUGE`), what it should publish for compatibility with etsy's statsd is `num`. Conveniently, this is already tracked in `(statsd_metric_t *)->updates_num` in collectd's statsd implementation to implement *DeleteCounters*. I'm going to send a patch for this in a minute.

PR #1314, opened 2015-10-20
Fix NTP kerninfo offset/estimated error scaling

### LINKED ISSUE #1300: NTPD time_offset-loop is wrong
With collectd 5.4 and 5.5, the time_offset-loop does not match value from ntpq/ntpdc.

For example, right now:

```
# I'm using graphite output. Here is ntpd releated measure from REQ_GET_KERNINFO
hostname.ntpd.frequency_offset-loop -6.302734375 1444644419
hostname.ntpd.time_offset-loop 86.6073455810547 1444644419
hostname.ntpd.time_offset-error 0.201217651367188 1444644419

# And output of ntpdc -c kerninfo run at the same time
pll offset:           0.0056759 s
pll frequency:        -6.303 ppm
maximum error:        0.421666 s
estimated error:      0.013187 s
status:               2001  pll nano
pll time constant:    8
precision:            1e-09 s
frequency tolerance:  500 ppm
```

After few digging in both collectd and ntpdc, It seem that NTPD changed it's factor from 65536.0 to 1 000 000, which explain the error:

86.6073455810547 (collectd value of offset-loop) * 65536.0 / 1 000 000 = 5.675899000000001

ntpdc factor 1 000 000 (us) comes from ntpdc/ntpdc_ops.c file, function kerninfo. Note that in this function, the scale factor could be 1 000 000 000 (ns) if STA_NANO is defined :(

I'm using ntp version 4.2.6.p5+dfsg-3ubuntu6

PierreF:
This pull requests fix #1300 

The scaling was taken from ntp-4.2.6, from ntpdc/ntpdc_ops.c , function kerninfo

Tested with ntp from Ubuntu vivid (1:4.2.6.p5+dfsg-3ubuntu6), value are the correct one with this patch.

Code from ntpdc_ops.c looks like:
```C
  [...]
  struct info_kernel *ik;
  double tscale = 1e-6;
  [...]
#ifdef STA_NANO                                                                 
    if (status & STA_NANO)                                                      
        tscale = 1e-9;                                                          
#endif
    (void)fprintf(fp, "pll offset:           %g s\n",
        (int32)ntohl(ik->offset) * tscale);
    (void)fprintf(fp, "pll frequency:        %s ppm\n",
        fptoa((s_fp)ntohl(ik->freq), 3));
    (void)fprintf(fp, "estimated error:      %g s\n",
        (u_long)ntohl(ik->esterror) * 1e-6);
```

Note: on my proposed patch tscale is always 1e-6 ... I didn't find where/how STA_NANO could be defined. I've assumed it could only be defined with customized/patched build of ntp.

### DIFF
--- a/src/ntpd.c
+++ b/src/ntpd.c
@@ -904,6 +904,7 @@ static int ntpd_read (void)
 
 	int status;
 	int i;
+	double tscale = 1e-6;
 
 	ik      = NULL;
 	ik_num  = 0;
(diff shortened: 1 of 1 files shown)


octo:
`STA_NANO` appears to come from [the Linux kernel](http://lxr.free-electrons.com/source/include/linux/timex.h?v=2.6.28#L182), usable via the Linux specific `<sys/timex.h>`. I'll add an appropriate comment.

mfournier:
FYI I cherry-picked this patch as 53438c510. Thanks a lot @PierreF !

PR #1315, opened 2015-10-20
Fix a few typos in collectd-exec(5)

douglaswth:


mfournier:
Merged as f8de61198, thanks a lot @douglaswth !

Issue #1227, opened 2015-08-16
DNS plugin in 5.5.0 segfaults under Linux 4.1.x


hnyman:
I have built identical builds for Openwrt router (ar71xx / mips32r2 / WNDR3700) with Linux kernel 3.8.x and 4.1.x, and collectd 5.5.0 segfaults quickly with Linux 4.1.x if the dns plugin is enabled. If DNS plugin is disabled, collectd works otherwise normally also with 4.1.x. The plugin works ok with Linux 3.8.x.

Segfault:

```
[  104.435977] 
[  104.435977] do_page_fault(): sending SIGSEGV to collectd for invalid write access to 76db3ae8
[  104.444585] epc = 76f6ef30 in dns.so[76f6e000+14000]
[  104.449597] ra  = 0040f084 in collectd[400000+30000]
[  104.454602] 
```

I have opened an issue at the Openwrt bug tracker ( https://github.com/openwrt/packages/issues/1660 ), but I am opening also this issue here as this sounds more like an collectd problem due to some change at Linux.

rubenk:
Would it be possible for you to get a backtrace?
https://collectd.org/wiki/index.php/Core_file

hnyman:
Looks difficult. I tried compiling with debug support and used gdb remote debugging to fetch info.
It doesn't seem to find any usable info for stack trace.
```
(gdb) c
Continuing.
[New Thread 3217.3231]
warning: GDB can't find the start of the function at 0x77ce4f38.

    GDB is unable to find the start of the function at 0x77ce4f38
and thus can't determine the size of that function's stack frame.
This means that GDB may be unable to access that stack frame, or
the frames below it.
    This problem is most likely caused by an invalid program counter or
stack pointer.
    However, if you think GDB should simply search farther back
from 0x77ce4f38 for code which looks like the beginning of a
function, you can increase the range of the search using the `set
heuristic-fence-post' command.

Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 3217.3231]
0x77ce4f38 in ?? ()
(gdb) bt
#0  0x77ce4f38 in ?? ()
(gdb) c
Continuing.

Program terminated with signal SIGSEGV, Segmentation fault.
The program no longer exists.
(gdb) 


********************************
On the router the output (from collectd -f) stops at the start of the dns functionality:

pe = ping_droprate; type_instance = 82.181.192.1;
[...]

PR #1316, opened 2015-10-21
Start write threads after plugin initialization.

baryonix:
Some plugins such as "network" create own threads from within their
init callbacks which can then start submitting data to the queue
right away, even if the read threads haven't been started yet.

If write threads are started before plugin initialization, this can
result in a race where a plugin's write callback gets called before
that plugin's init callback has completed.

To fix this, delay starting the write threads until after all plugins
have been initialized.

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -1725,8 +1725,6 @@ void plugin_init_all (void)
 		write_threads_num = 5;
 	}
 
-	start_write_threads ((size_t) write_threads_num);
-
 	if ((list_init == NULL) && (read_heap == NULL))
 		return;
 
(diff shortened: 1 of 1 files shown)


mfournier:
Thanks a lot @baryonix ! This seems sensible, I think it should go in.

@hnyman I'm actually wondering if this wouldn't be a fix for #1227. Do you still have this openwrt environment available ?

hnyman:
I created a test build for Openwrt, but this change did not help. There were still crashes with the DNS plugin.

baryonix:
Quite clearly #1227 seems to happen at a later point as there we can see read threads already running, and those don't get started before plugin initialization finishes even in the current state of affairs. So I think it's most likely an issue unrelated to this.

(Actually #1227 is giving me a headache. I figured it might be some kind of race between the PCAP thread's initialization and the read callback, but I checked the code up and down and can't seem to spot it. Sorry I cannot give any more help with that one right now.)

mfournier:
This fix has been merged as b888aac60. It will be part of future bugfix and feature releases. Thanks a lot @baryonix !

PR #1317, opened 2015-10-22
Make connectivity with redis more reliable/robust.

kriegerse:
If redis fails try to reconnect, instead of asume connectivity is always available.


src/write_redis.c
```diff
@@ -55,6 +55,39 @@ typedef struct wr_node_s wr_node_t;
...
+     return NULL;
+    }
+   else if (node->conn->err)
+   {
+     ERROR ("write_redis plugin: Connecting to host \"%s\" (port %i) failed: %s",
+             (node->host != NULL) ? node->host : "localhost",
+             (node->port != 0) ? node->port : 6379,
+             node->conn->errstr);
+     pthread_mutex_unlock (&node->lock);
+     return NULL;
```
mfournier: Shouldn't you free `node->conn` here ?

src/write_redis.c
```diff
@@ -55,6 +55,39 @@ typedef struct wr_node_s wr_node_t;
 /*
  * Functions
  */
+static redisContext * wr_connect ( wr_node_t *node ) /* {{{ */
+{
+  redisReply   *rr;
+
+  node->conn = redisConnectWithTimeout ((char *)node->host, node->port, node->timeout);
+   if (node->conn == NULL)
```
mfournier: There's a little indentation issue here

src/write_redis.c
kriegerse: should be fixed with commit 51dee33f9c27b3030d5f8625f70977e41bb0e021

src/write_redis.c
kriegerse: should be fixed with commit 51dee33f9c27b3030d5f8625f70977e41bb0e021

kriegerse:
write_redis and collectd now survives, if redis is not available or was restarted since the last write.

PR #1107, opened 2015-07-02
write_tsdb with metadata

ymettier:
Hello,

First of all, PR #1106 is a pre-requisite for this PR. This is why I based my work on that branch.

This patch allows use metadata to completely rewrite some metrics for OpenTSDB.
The problem is best explained in issue #709 and #887 : some metric names should not include the instance in their name but in a tag. For example, `cpu.1.cpu.user` should become `cpu.user` with an additional tag `cpu=1`

This patch is a proposition of a solution for this problem.

I added some metadata : 

<table>
<tr><th>metadata</th><th>explanation</th></tr>
<tr>
<td>tsdb_prefix</td><td>Will prefix the OpenTSDB &lt;metric&gt; (also prefix tsdb_id if defined)</td></tr>
<tr><td>tsdb_id</td><td>Replace the metric with this tag</td></tr>

<tr><td>tsdb_tag_plugin<br/>
tsdb_tag_pluginInstance<br/>
tsdb_tag_type<br/>
tsdb_tag_typeInstance<br/>
tsdb_tag_dsname</td>
<td>When defined, tsdb_tag_* removes the related item from metric id.<br />
If it is not empty, it will be the key of an opentsdb tag (the value is the item itself)<br />
If it is empty, no tag is defined.</td>
</tr></table>

Here is how I use them for commonly used plugins cpu, df, disk, interface, load and swap : 
```
<Chain "PreCache">
   <Rule "opentsdb_cpu">
     <Match "regex">
       Plugin "^cpu$"
     </Match>
     <Target "set">
       MetaDataSet "tsdb_tag_pluginInstance" "cpu"
       MetaDataSet "tsdb_tag_type" ""
       MetaDataSet "tsdb_prefix" "sys."
     </Target>
   </Rule>
   <Rule "opentsdb_df">
     <Match "regex">
       Plugin "^df$"
     </Match>
     <Target "set">
       MetaDataSet "tsdb_tag_pluginInstance" "mount"
       MetaDataSet "tsdb_tag_type" ""
       MetaDataSet "tsdb_prefix" "sys."
     </Target>
   </Rule>
   <Rule "opentsdb_disk">
     <Match "regex">
       Plugin "^disk$"
     </Match>
     <Target "set">
       MetaDataSet "tsdb_tag_pluginInstance" "disk"
       MetaDataSet "tsdb_prefix" "sys."
     </Target>
   </Rule>
   <Rule "opentsdb_interface">
     <Match "regex">
       Plugin "^interface$"
[...]

PR #1319, opened 2015-10-25
New plugin: write_influxdb

baryonix:
This plugin writes to InfluxDB 0.9.x servers.
    
It offers a higher level of configurability than the collectd receiver integrated in InfluxDB. The naming of the time series and tags written into InfluxDB can be fully customized.
    
The plugin also supports a "StoreRates" parameter.

Note that this PR includes the changes proposed in #1318 as the plugin depends on them.


baryonix:
I'm a bit confused by that Jenkins build failure. It failed because it couldn't find `utils_iomux.h`, and that file obviously wasn't in the tar that Jenkins extracted, but it does exist in the branch. Did Jenkins mess up? Maybe you could check and give the build another try? Thanks!

mfournier:
Yes, I was actually looking into this one too ;)

So what jenkins does is more or less follow the release process: `./build.sh && ./configure && make dist`, and copies the resulting tarball to the different test environments.

It looks like the `dist` make target builds `src/utils_iomux.o` and this file get's included in the tarball rather than `src/utils_iomux.h`. I see it is part of the elements in `write_influxdb_la_SOURCES`, in `src/Makefile.am`. Probably this is what's causing this problem.

baryonix:
Oh, my bad. A typo. Thanks for spotting this.

baryonix:
Sorry for the mess, I wasn't very familiar with how the build process works. I did test this on my box now and can build successfully from the "make build" generated tarball, so this should be it hopefully.

ymettier:
Hello,

I have questions about DNS, load balancing...
* If you change the IP of your InfluxDB server (and keep the name), will your plugin know about the IP change ? Or will it stop sending data because of connection timeouts ?

* I you have a clustered InfluxDB, how does the plugin work when one node is offline (broken, maintenance or whatever) ? BTW, how would you configure Collectd and your plugin to load-balance to InfluxDB ?

* When InfluxDB is down, how many DNS requests does your plugin do until InfluxDB is back online again ? `sudo tcpdump -n dst port 53` may help you answer this question. This is an important point because you don't want 1000 Collectd requestiing 100 times (one per value) per interval when InfluxDB is down. Notice that this is currently a bug of write_tsdb.c.

Regards,
Yves

ymettier:
Hello,

I have suggestions about Collectd meta data and InfluxDB tags.
First of all, please have a look on PR #1106 that allows to add meta data in Collectd values.

* How could users add tags to the values sent to InfluxDB ? For example, I would like to add a tag "unit" for each metric. I know this is hard work, but I can do it in a meta data thanks to PR #1106.

* How can you redefine the names of the metrics ? For example, for CPU, I would prefer `cpu.wait` and tag `cpu=1` instead of `cpu.1.wait`. Same for processes, tcpconns... What about a prefix (`sys` for most common Collectd plugins, `app` for application plugins...)
Please read issue #887 for more details on this question.

You may want to also read my PR #1107 and see how I implemented those suggestions for write_tsdb plugin.

Note about PR #1107, your plugin and InfluxDB. Of course PR #1107 can help sending metrics to InfluxDB. However, write_tsdb uses "telnet" style, not http protocol. Your plugin is also interesting for its use of http protocol, which allows to solve problems with common http solutions, like load balancing, proxy, authentication, ... And everything that InfluxDB implements natively for its http input and not in its OpenTSDB input plugin. So, go on, please take the best of PR #1107 in your plugin !

Regards,
Yves

baryonix:
Regarding load balancing and failover, the plugin currently leaves this to CURL and doesn't do any of it on its own. You'd achieve this by pointing the plugin at a DNS entry that resolves to multiple IP addresses. What behavior you'll get exactly probably depends on what version of CURL you're using. I did some tests with CURL 7.40.0 and it shows the following behavior:

* It doesn't really do "round robin". Once it does a DNS lookup it seems to choose some ordering of the entries at random and then stick to that. So a single instance of collectd will keep sending traffic to the same server as long as it's available. Multiple collectd's will balance over all servers.

* Failover does seem to work quite well. If CURL fails to get a connection to the "primary" server it chose, it will fail over to others as soon as the connection times out. If a server is unavailable, you may have a couple of connection attempts piling up for timeout, but this doesn't really hurt as all CURL processing is handled by a single collectd thread. As soon as CURL does manage to get a connection, it will send out all pending requests over that one.

This may not be the optimal behavior when it comes to load balancing over multiple InfluxDB nodes, but I think it should do. If anyone has an environment so demanding that this gives a performance bottleneck, they can hire me (or anyone else) to implement a more sophisticated load balancing scheme in the plugin on top of what CURL does, but I don't think this is going to happen anytime soon.

Regarding DNS, the plugin also leaves this to CURL and doesn't do any lookups of its own. CURL has a DNS cache, so it will not overload resolvers but should pick up an address change eventually when its cache entry times out.

baryonix:
Regarding your second comment, there is some flexibility about how the plugin names its series and what tags it attaches to them. Both the series name and any tags can be configured in terms of format strings that can involve data from the value_list_t being processed such as host name, plugin name, plugin instance etc. Please see the man page for details.

#1106 looks very reasonable to me, I think it makes a lot of sense to be able to generate additional tag/value pairs using the filtering logic already present in collectd, instead of having to replicate this in specific plugins.

I will add the capability of generating additional InfluxDB tags from collectd metadata when my time allows.

baryonix:
I've added some metadata support as suggested by @ymettier. By default, measurement name and tags are still generated as configured in the `<Format>` block, but metadata can now be used to
* override the measurement name;
* add a prefix to the measurement name;
* add additional InfluxDB tags;
* override or delete tags generated from the `<Format>` block.

There are some details on this in the man page.

(Note that I decided to use `:` as a separator in the metadata attribute names instead of `_`, to be consistent with other plugins already using metadata such as `network` and `aggregation`).

octo:
Hi @baryonix, thank you very much for your work on this! I'll leave some code review comments inline.

Best regards,
—octo

src/daemon/utils_hashtable.c
```diff
@@ -0,0 +1,474 @@
+#include "utils_hashtable.h"
```
octo: Please include `"collectd.h"` as first header.

src/daemon/utils_hashtable.c
```diff
@@ -0,0 +1,474 @@
...
+#include "plugin.h"
+
+#include <stdio.h>
```
octo: Most of these are already loaded by `"collectd.h"`.

src/daemon/utils_hashtable.c
```diff
@@ -0,0 +1,474 @@
...
+
+struct bucket_s {
+    hash_t hash;
+    bucket_status_t status;
+};
+
+typedef struct bucket_s bucket_t;
+
+
+static inline uintptr_t do_align (uintptr_t base, uintptr_t alignment);
```
octo: Please don't use the `inline` keyword. Also try to avoid prototype declarations.

src/daemon/utils_hashtable.c
```diff
@@ -0,0 +1,474 @@
...
+static inline void *bucket_to_user (const hashtable_t *table, bucket_t *bucket);
+static inline bucket_t *user_to_bucket (const hashtable_t *table, void *user);
+static int rehash (hashtable_t *table, unsigned size_exp);
+static int check_grow (hashtable_t *table);
+static int check_shrink (hashtable_t *table, bool bulk);
+static int hashtable_lookup_internal (hashtable_t *table, hash_t hash,
+        bool (*match) (void *, void *), void *match_arg, void **data);
+
+
+static inline uintptr_t
```
octo: Please avoid types that are not portable. Use `unsigned int *` instead.

src/daemon/utils_hashtable.c
```diff
@@ -0,0 +1,474 @@
...
+do_align (uintptr_t base, uintptr_t alignment)
```
octo: Please don't hand-roll alignment code. Please use the C11 function(s) instead: `aligned_alloc(size_t alignment, size_t size)`

src/daemon/utils_hashtable.c
```diff
@@ -0,0 +1,474 @@
...
+{
+    return (bucket_t *) ((char *) user - table->bucket_doff);
+}
+
+
+static int
+rehash (hashtable_t *table, unsigned size_exp)
+{
+    const hash_t old_size = table->size;
+    char *const old_arena = table->arena;
```
octo: The `const` here is more confusing than helpful, please remove it.

src/daemon/utils_hashtable.c
```diff
@@ -0,0 +1,474 @@
...
+    void *const old_handle = table->arena_handle;
+
+    int rc = alloc_arena (table, size_exp);
```
octo: s/rc/status/

This is the default name for this variable thorough collectd.

src/daemon/utils_hashtable.c
```diff
@@ -0,0 +1,474 @@
...
+
+static inline bucket_t *
+get_bucket2 (char *arena, hash_t stride, hash_t index)
+{
+    return (bucket_t *) (arena + index * stride);
+}
+
+
+static int
+alloc_arena (hashtable_t *table, unsigned size_exp)
```
octo: Please use `size_t` for variables holding "a number of bytes"

src/daemon/utils_hashtable.h
```diff
@@ -0,0 +1,218 @@
+#ifndef UTILS_HASHTABLE_H
+#define UTILS_HASHTABLE_H
+
+#include "config.h"
+
+#include <stdbool.h>
```
octo: Please don't use `<stdbool.h>`. Instead, use `_Bool`, `1` and `0`.

src/daemon/utils_hashtable.h
```diff
@@ -0,0 +1,218 @@
...
+ * Equivalent to chaining hash_update() calls for all bytes of "str",
+ * not including the trailing NUL.
+ */
+static inline hash_t hash_update_str (hash_t h, const char *str);
+
+/*
+ * Equivalent to chaining hash_update() calls for "len" bytes of memory
+ * starting at "data".
+ */
+static inline hash_t hash_update_mem (hash_t h, const void *data, unsigned len);
```
octo: unsigned → size_t

octo:
Hi @baryonix,

I'm skeptical about the overhead this plugin produces: it uses an abstraction layer over cURL with an unclear benefit: other plugins register one write callback per socket / cURL object – the formatting overhead they have is negligible compared to their network and TLS overhead. I don't think this plugin is fundamentally different.

The cURL abstraction then uses a iomux layer with unclear benefit: the *write_influxdb plugin* will typically have one to three destinations configured. I.e. the code establishes a small number of TCP connections and keeps them open. I can't see the standard POSIX *poll(2)* to become the bottleneck, even with hundreds of connections.

The iomux layer then uses a hash table to manage its file descriptors. At typical configurations (1 or 2 endpoints), I'd expect a simple list to actually be *more* efficient than a hash table.

I propose the following way forward:
 *  Remove the dependency on the cURL reaktor, iomux and hashtable for the time being. Simplify the plugin as much as possible, i.e. register one *write callback* per endpoint with one cURL object each. Look at `wh_config_node()` in `src/write_http.c` for an example.
 *  Move the hash table implementation into a separate PR. Look at lookup-intensive users of `src/daemon/utils_avltree.h` and see if the hashtable offers a performance benefit for them. Good candidates are `src/daemon/utils_cache.c` and `src/utils_vl_lookup.c`.
 *  I don't think the *iomux* implementation is a very good fit for collectd: we simply don't have to handle the number of connections to justify the complexity. The *SNMP plugin* routinely has thousands of connections configured, but unfortunately the connection multiplexing there is handled within the Net-SNMP library. Maybe another plugin may have a good use for it, but the write plugins don't strike me as particularly promising.
 *  Switching to a more centralized cURL infrastructure is certainly interesting – there is way too much code duplication in the various plugins using libcurl. It would be very interesting to have all the cURL setup stuff in a central place and not having to set `CURLOPT_NOSIGNAL` for the n+1st time. I'm not sure that switching to the *multi* API provides much benefit though.

What do you think?
—octo

faxm0dem:
I'm getting multiple error messages when collectd dispatches values:
```
plugin_dispatch_values: plugin_write_enqueue failed with status 12 (Cannot allocate memory).
```

kev009:
I would recommend write_tsdb which can talk to InfluxDB's tsdb listener.  InfluxDB sucks, until it can not fall over, not sure the weight of this is worthwhile.

faxm0dem:
I can recommend write_tsdb too. The issue we had was restrictions in the tags (was it unallowed space character?)

PR #1321, opened 2015-10-26
apache: Extended the number of lines to parse from mod_status

YmrDtnJu:
mod_status of Apache 2.4.17 changed the layout of its output. The collectd module only parsed the first 16 lines. It ignored all the rest of the output and therefore ignored some values. Extending the number of lines to parse to 64 fixes this. It may be necessary to extend the number to a higher value in the future or rewrite the code to parse the output.

### DIFF
--- a/src/apache.c
+++ b/src/apache.c
@@ -517,13 +517,14 @@ static void submit_scoreboard (char *buf, apache_t *st)
 	}
 }
 
+#define _MAX_LINES 64
 static int apache_read_host (user_data_t *user_data) /* {{{ */
 {
 	int i;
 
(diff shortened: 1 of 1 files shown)


mfournier:
As discussed on IRC, we just need to make sure 64 is large enough to cover all cases.

According to http://www.apache.org/dist/httpd/CHANGES_2.4 `mod_proxy, mod_ssl, mod_cache_socache, mod_socache_*: Support machine readable server-status produced when using the "?auto" query string.` for version 2.4.13 onwards.

mfournier:
Superseded by #1343

Issue #1323, opened 2015-10-27
utils_ignorelist missing in collectd plugin API


bdrung:
commit 216c6246b73645ac093de15b87aedc9abc6ebc80 moves the collectd daemon files into the src/daemon/ directory, but also silently drops utils_ignorelist. Dropping utils_ignorelist from the daemon has the drawback that it cannot be used by plugins any more and that 16 plugins need to bundle ignorelist.

mfournier:
Hi @bdrung, thanks for reporting this and working on a patch, much appreciated !
I have trouble understanding what the issue actually is though (running 5.5 myself, using ignorelist-related config with some of these 16 plugins, and didn't notice any side-effect). Could you describe the problem from the user perspective please ? A config snippet, which collectd version have you installed, etc, would be helpful too. Thanks !

bdrung:
The problem is that utils_ignorelist is not available for **external** plugins. The shipped plugins bundle utils_ignorelist and therefore work around the problem, but external plugins cannot bundle the source code of utils_ignorelist. We have an inhouse plugin that uses ignorelist, which breaks with collectd 5.5.

The collectd daemon 5.4.1 comes with ignorelist functions, but collectd 5.5.0 drop it.

mfournier:
Ok, I get it, thanks for the explanations @bdrung !

It seems @octo bundled utils_ignorelist in the plugins using it on purpose (cf 216c6246b). So although I agree this is broken, I'll leave your PR open a few more days to give him a chance to comment on this.
Also, a couple of others seem to not be bundled with the daemon anymore, namely: utils_avltree utils_heap and utils_parse_option. Not sure if they should or not.

mfournier:
Closed by 817e080

dswarbrick:
:+1::+1:

PR #1324, opened 2015-10-27
Add utils_ignorelist back to daemon (#1323)

bdrung:
commit 216c6246b73645ac093de15b87aedc9abc6ebc80 moves the collectd
daemon files into the src/daemon/ directory, but also silently drops
utils_ignorelist. Dropping utils_ignorelist from the daemon has the
drawback that it cannot be used by plugins any more and that 16 plugins
need to bundle ignorelist.

Thus add utils_ignorelist back to the collectd daemon.


mfournier:
See discussion in #1323.

mfournier:
Merged as 817e080279, thanks a lot for tracking this down @bdrung ! The fix will be part of the 5.5.1 release.

PR #1325, opened 2015-10-28
Set SNMP ASN_GAUGE values to unsigned.

### LINKED ISSUE #50: SNMP Plugin making wrong interpretation data type
Hello, i am using collectd (4.8.2-1ubuntu0.1) to graph informations from my HP Procurve switches using collectd snmp plugin. One of the data i am collecting is the uptime.

```
<Plugin snmp>
        <Data "procurve_uptime">
                Type "uptime"
                Table false
                Values ".1.3.6.1.2.1.1.3.0"
                Scale 0.01
        </Data>
        <Host "HP5406zl-2">
                Address "xxx.xxx.xxx.xxx"
                Version 2
                Community "public"
                Collect "procurve_uptime" 
        </Host>
</Plugin>
```

SysUpTime (.1.3.6.1.2.1.1.3.0) is a 32 bit value and snmp plugin should be able to get value from 0 to 4294967295 (uptime data type is defined like this).

I can get the correct values from most of Procurve Switches (using csv plugini can see positive values) but on somes switches with highest uptime (using snmpwalk i get 2367115384 which is far less than 4294967295) collectd snmp plugin is making a wrong interpretation of the data type and csv plugin shows negatives values.

may this be a bug?.........

thanks

nward:
SNMP-SMIv2 specifies Gauge32 as unsigned, however the change to fix #50, which was this exact problem, set both Integer and Gauge to signed, when it should have only set Integer.

This patch sets Gauge32 to be unsigned.

### DIFF
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -924,8 +924,7 @@ static value_t csnmp_value_list_to_value (struct variable_list *vl, int type,
     tmp_unsigned = (uint32_t) *vl->val.integer;
     tmp_signed = (int32_t) *vl->val.integer;
 
-    if ((vl->type == ASN_INTEGER)
-        || (vl->type == ASN_GAUGE))
+    if ((vl->type == ASN_INTEGER))
       prefer_signed = 1;
 
     DEBUG ("snmp plugin: Parsed int32 value is %"PRIu64".", tmp_unsigned);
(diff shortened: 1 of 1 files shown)


nward:
Updated the PR to fix clang complaining about extraneous parenthesis.

mfournier:
Thanks for figuring this out @nward !

The reasoning seems to match what one can read on the topic in rfc1902.

I'll leave this open for a few more days so that others can review, but AFAIC it's good to go.

nward:
As a data point, I've been running this in production for a few days now, no problems with any other stats - and we have.. lots.

mfournier:
Many thanks @nward ! I've just merged your patches (30e6ef69 + 6784becb) to the bugfix branches, so they will be part of every future release.

Issue #1291, opened 2015-10-04
Suggestion (with code patch): Support custom PluginName attribute of a <Value> in an <MBean>


davidc-donorschoose:
This idea is copied from the FastJMX project, and allows segmentation of reported 
metrics into different plugin buckets rather than everything being reported as "GenericJMX".

The code patch I will provide can be applied to the collectd-5.5.0 source download after applying my patch for issue #1290 -- called "Suggestion (with code patch): Support TabularData for java.lang:type=GarbageCollector". It's the same source file GenericJMXConfValue.java, and I assume you would be more interested in #1290. I apologize for not forking your repo so I can submit a proper pull request.

This isn't the best example, but notice that I shortened the metric names from "GenericJMX-..." to "jmx-..." with the PluginName setting, which I need to put into every \<Value\> of every \<MBean\>. This is very useful for me, since long, mixed case metric names can be confusing. This is based on my example for #1290:

```
    <MBean "jmx-gc/parnew">
      # The ParNew copy collector copies from Eden to Survivor to Old Gen.
      ObjectName "java.lang:type=GarbageCollector,name=ParNew"
      <Value>
        # jmx-gc/memory_pool-old_after_parnew/{committed,max,used}
        Attribute "LastGcInfo.memoryUsageAfterGc.CMS Old Gen.committed"
        Attribute "LastGcInfo.memoryUsageAfterGc.CMS Old Gen.max"
        Attribute "LastGcInfo.memoryUsageAfterGc.CMS Old Gen.used"
        PluginName "jmx"
        Type "memory_pool" # committed:GAUGE:0:U, max:GAUGE:0:U, used:GAUGE:0:U
        InstancePrefix "old_after_parnew"
      </Value>
    </MBean>

    <MBean "jmx-gc/*cms">
      # The ConcurrentMarkSweep collector frees dead objects in Old Gen.
      ObjectName "java.lang:type=GarbageCollector,name=ConcurrentMarkSweep"
      InstancePrefix "gc"
      <Value>
        # jmx-gc/memory_pool-old_after_cms/{committed,max,used}
        Attribute "LastGcInfo.memoryUsageAfterGc.CMS Old Gen.committed"
        Attribute "LastGcInfo.memoryUsageAfterGc.CMS Old Gen.max"
        Attribute "LastGcInfo.memoryUsageAfterGc.CMS Old Gen.used"
        PluginName "jmx"
[...]

PR #1330, opened 2015-10-30
adding patch from #1291

lusis:
This adds the `PluginName` support from #1291 (without the previous tabular data patch) to support reducing length of metric names. We're currently suffering with some truncated data as well due to insanely long names that we can't really control.

Note that I didn't bother to fire up the IDE for this. I'm waiting on a build to finish locally but I wanted to get the patch out there. The original author in #1291 did all the work, I just made it into a PR that might have more traction.

### DIFF
--- a/bindings/java/org/collectd/java/GenericJMXConfValue.java
+++ b/bindings/java/org/collectd/java/GenericJMXConfValue.java
@@ -68,6 +68,7 @@ class GenericJMXConfValue
   private List<String> _attributes;
   private String _instance_prefix;
   private List<String> _instance_from;
+  private String _plugin_name;
   private boolean _is_table;
 
   /**
@@ -436,6 +437,7 @@ class GenericJMXConfValue
     this._attributes = new ArrayList<String> ();
     this._instance_prefix = null;
     this._instance_from = new ArrayList<String> ();
+    this._plugin_name = null;
     this._is_table = false;
 
     /*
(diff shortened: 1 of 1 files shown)


lusis:
cc @davidc-donorschoose

mfournier:
Thanks for working on this @davidc-donorschoose & @lusis !

Actually building generic-jmx.jar is fairly easy to do manually (so you can check everything works as expected, etc). cd to the `bindings/java` dir from the source tree, then run `javac org/collectd/java/*.java && jar cf /tmp/generic-jmx.jar org/collectd/java/*.class`.

Apart from the error jenkins pointed out above, would you mind amending the `collectd-java(5)` manpage with a short description of this new option ? This is in `src/collectd-java.pod`. Thanks !

toni-moreno:
Hi @lusis and @davidc-donorschoose . 

I'm  whaching this patch and I think what you need Is a PR I did some time ago( and unfortunately closed without merge). Let me know if the main goal  for both PR's are the same , in that case I would like to update /rebase a new PR with a generic "PluginName"  or "RenamePluginAs" parameter as a new proposal for any generic purpose plugin.

Thank you very much for your work.

octo:
#1291 has been resolved.

Thanks for pointing out the missing manual page changes, @mfournier, I would have missed it. They're in 9c446022a632d5fbcf6a1089b05e2dda9419146b.

PR #1332, opened 2015-11-02
curl_json: handle boolean values in a more useful way

mfournier:
Currently curl_json will barely ignore boolean values in a non erroneous
way, so that if you have an array like this: `{ foo: [false, 123] }`
you will be able to access the 123 number like `foo[1]`

We now rather call our number parser with "0" for false, "1" for true.

### DIFF
--- a/src/curl_json.c
+++ b/src/curl_json.c
@@ -203,12 +203,6 @@ static void cj_cb_inc_array_index (void *ctx, _Bool update_key)
 #define CJ_CB_ABORT    0
 #define CJ_CB_CONTINUE 1
 
-static int cj_cb_boolean (void * ctx, int boolVal)
-{
-  cj_cb_inc_array_index (ctx, /* update_key = */ 0);
-  return (CJ_CB_CONTINUE);
-}
-
 static int cj_cb_null (void * ctx)
 {
   cj_cb_inc_array_index (ctx, /* update_key = */ 0);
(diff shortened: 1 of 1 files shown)


pyr:
This seems reasonable to me. :+1:

src/curl_json.c
```diff
@@ -310,6 +304,11 @@ static int cj_cb_string (void *ctx, const unsigned char *val,
   return (cj_cb_number (ctx, (const char *) val, len));
 } /* int cj_cb_string */
 
+static int cj_cb_boolean (void * ctx, int boolVal)
+{
+  return (cj_cb_number (ctx, (boolVal)? "1":"0", 1));
```
octo: Nit, at your option: I'd slightly prefer to assign to a variable and call `strlen()` on that instead of hard-coding the `1`:

```c
char const *n = boolVal ? "1" : "0";
return (cj_cb_number (ctx, n, strlen (n));
```

octo:
LGTM

src/curl_json.c
TerryHowe: Not sure I follow that comment since wouldn't strlen always return 1?

TerryHowe:
What pending action is on this PR?

src/curl_json.c
dothebart: exactly - you're doing a function call, which does actual work, which is costly.  Thus only putting in 0 is a safe guess here. 

maybe not making it an implicit If, and writing `return cj_cb_number (ctx, "1", 1); else ...` would make it more readable...

dothebart:
proposed a better way of doing this with #2471 one can close this.

src/curl_json.c
octo: Yes, it would always return 1, but it would make it clear to the reader that the length of the string is passed and not, for example, a boolean "true" argument.

I'd be quite surprised if this would result in a function call. I'm fairly sure that compilers figure out that it's always invoked with a string literal.

Anyway, it's an optional comment, so there's no point in stressing this ;)

Best regards,
—octo

PR #1333, opened 2015-11-03
Implement kafka custom output format

ymettier:
Hello,

I want to be able to customize my own format to send to Kafka. How can I do it ?
* hack the current `src/utils_format_json.c` ? Bad idea because somebody else may want to customize another way.
* use a 3rd party tool that would push to Kafka. Use a compatible write plugin to push to that 3rd party tool ? No. Too complex.
* improve `src/write_kafka.c` and implement a language to describe the output ? Yes, but too complex, and that would break the perfs of Collectd. If you care about perfs, you cannot use an interpretation layer.
* fork `src/write_kafka.c` and maintain my own plugin (which would be unpublished).No, fork is bad here.

This PR is a mix of 1st, 3rd and 4th ideas above. The idea is to provide a source file similar to `src/utils_format_json.c` to hack.
* If you don't need it, it will not be compiled. You will never hear of it.
* If you want to customize your output, you can work on `src/utils_format_kafka_custom.c`. This file is provided specificallly for users who want to code their own format. You need to define `IMPLEMENT_KAFKA_FORMAT_CUSTOM_1` constant to take your work into account.

This PR does not provide my own customization of the format. Probably nobody cares about it.

This PR provides the "framework" in order to let  users develop their own format :
* define `IMPLEMENT_KAFKA_FORMAT_CUSTOM_1` constant to take your work into account. See `src/utils_format_kafka_custom.h` to know how to use it.
* `src/utils_format_kafka_custom.c` (and `src/utils_format_kafka_custom.h`) template file
* new format `CUSTOM1` defined in `write_kafka.c`. This is because nobody should hack `write_kafka.c` to enable their own format. Note : I named the new format `CUSTOM1` in order to be able to have more than one custom format later.

About documentation : I chose to write it in `src/utils_format_kafka_custom.h` because I think that only users with C development skills may use this feature. Such users will have a look in `write_kafka.c` and easily find the framework.

Regards,
Yves

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1239,6 +1239,7 @@ pkglib_LTLIBRARIES += write_kafka.la
 write_kafka_la_SOURCES = write_kafka.c \
                         utils_format_graphite.c utils_format_graphite.h \
                         utils_format_json.c utils_format_json.h \
+                        utils_format_kafka_custom.c utils_format_kafka_custom.h \
                         utils_cmd_putval.c utils_cmd_putval.h \
                         utils_crc32.c utils_crc32.h
 write_kafka_la_CPPFLAGS = $(AM_CPPFLAGS) $(BUILD_WITH_LIBRDKAFKA_CPPFLAGS)
--- a/src/utils_format_kafka_custom.c
+++ b/src/utils_format_kafka_custom.c
@@ -1,0 +1,78 @@
+/**
+ * collectd - src/utils_format_kafka_custom.c
+ * Copyright (C) 2009       Florian octo Forster
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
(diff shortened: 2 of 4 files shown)


ymettier:
Also have a look on
* PR #1106 that makes this PR very interesting (rules to add metadata to values)
* PR #1107 that gives examples on how to use PR #1106 (rules examples to rename and enrich the values)

pyr:
Hi Yves,

I agree with you on the need for better formatters. I'm not onboard with you on the idea.
What collectd needs right now is rather a concept of "formatter" plugin for all output plugins which might emit in different formats. The need is the same in RabbitMQ, HTTP and to some extent, redis as well.

In the meantime I think there are two valid courses of action to take:

- A local plugin, as you were describing (there are now compelling options for maintaining a local plugin).
- A transient kafka topic where you just rewrite to your expected destination format. This is quite easy to do and can be very low-resource wise and let you do interesting things. The upcoming stream DSL support in Kafka will even let you ship a consumer which bundles the two tasks.

ymettier:
Hello @pyr,

Do you have documentation about what you name a "local plugin" ?
How could I help on that ?

About any implementation (either with a "local plugin" or with a transient kafka topic), how could I send meta-data information ? The idea is to have static meta-data information from each Collectd (for example thanks to PR #1106) and avoid the need of an inventory database to add the information that Collectd could be aware of.

Regards,
Yves

ymettier:

PR #1339, opened 2015-11-06
collection.cgi: parse types.db to discover DS names

mgedmin:
ChangeLog: contrib collection.cgi: parse types.db to discover DS names

Fixes users, entropy, and ntpd (time_offset/frequency_offset) graphs
that were broken by the great data source name normalization of
collectd v4 -> v5.

### DIFF
--- a/contrib/collection.cgi
+++ b/contrib/collection.cgi
@@ -33,6 +33,7 @@ use RRDs ();
 use Data::Dumper ();
 
 our $Config = "/etc/collection.conf";
+our $TypesDB = "/usr/share/collectd/types.db";
 our @DataDirs = ();
 our @DontShowTypes = ();
 our $LibDir;
@@ -50,6 +51,9 @@ our @RRDDefaultArgs = ('-w', '400');
 
 our $Args = {};
 
+our $Types = {};
+read_types_db();
+
 our $GraphDefs;
 our $MetaGraphDefs = {};
 load_graph_definitions ();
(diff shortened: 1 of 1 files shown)


dago:
I would like to merge this one, but CI has not reported back and I cannot make a dummy-change to retrigger it.

mgedmin:
Rebased on master, force-pushed, let's see if that gets the CI working.

contrib/collection.cgi
```diff
@@ -61,6 +65,27 @@ for (qw(action host plugin plugin_instance type type_instance timespan))
...
+	open ($fh, "< $TypesDB") or confess ("open ($TypesDB): $!");
+	while (my $line = <$fh>)
+	{
+		chomp ($line);
+		$line =~ s/\s*(?:#.*)?$//;
+		next if (!$line);
+		my ($type, $specs) = split(m/\s+/, $line, 2);
+		my @specs = split(m/,?\s+/, $specs);
+		$Types->{$type} = [];
+		for (my $i = 0; $i < @specs; $i++)
```
dago: I suggest iterating over the elements itself, like
```perl
foreach my $spec (@specs)
{
  my ($name, $ds_type, $min, $max) = split(m/:/, $specs);
  ...
}

File: contrib/collection.cgi [hunk 2]
```diff
@@ -809,6 +834,7 @@ sub action_show_graph
 
     $file =~ s/:/\\:/g;
     s/{file}/$file/ for (@rrd_args);
+    s/{ds(\d+)}/$Types->{$gtype}[$1]/ for (@rrd_args);
```
dago: I think `for` sets `$_` instead of `$1`

mgedmin: Yes, the `$1` here is referring to the `(\d+)` capture group in the regex.  The `for` is to do the ds(\d+) replacement for each item in @rrd_args.

contrib/collection.cgi
mgedmin: I think you meant `split(..., $spec)` rather than `split(..., $specs)`.

I've made the change (but I'm not currently set up to test that it works ;).

contrib/collection.cgi
dago: Ah, ok, missed that.

rpv-tomsk:
It's funny for me to look on all these work.

This work has no sence.
Why do not simply do replacement of hardcoded values `entropy`, `ppm`, `seconds`, `users` into another hardcoded value `value`, instead of `{ds0}` which will be subsituted by `value` anyway?

It looks as work imitation much more than work.

mgedmin:
IIRC the reason I originally did this was because I had a bunch of Ubuntu LTS servers with different LTS versions (12.04 and 14.04 or something like that).  Some of them were still using collectd 4 while others were on collectd 5, and I wanted to deploy one CGI script that would work on either.

A lot of time has passed since then and it probably no longer makes sense to make this dynamic.

PR #1343, opened 2015-11-15
apache: parse whole response, not only the first 16 lines

mfournier:
mod_status recently started reporting more statistics. According to
http://www.apache.org/dist/httpd/CHANGES_2.4 `mod_proxy, mod_ssl,
mod_cache_socache, mod_socache_*: Support machine readable server-status
produced when using the "?auto" query string.` for version 2.4.13 onwards.

As we can't assume everything we need will be included in the first 16
lines of output anymore, we now simply scan through the whole output.

### DIFF
--- a/src/apache.c
+++ b/src/apache.c
@@ -585,12 +585,9 @@ static void submit_scoreboard (char *buf, apache_t *st)
 
 static int apache_read_host (user_data_t *user_data) /* {{{ */
 {
-	int i;
-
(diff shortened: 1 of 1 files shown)


mfournier:
This is a (imho) better fix than #1321. @YmrDtnJu would you mind taking a look and giving this a try ? Thanks !

octo:
LGTM

Issue #1265, opened 2015-09-08
MQTT: mqtt plugin should support TLS


rubenk:

octo:
CC @jpmens

jpmens:
@octo I'll give this a shot once #1262 is solved.

rubenk:
Thanks @jpmens!

PR #1345, opened 2015-11-17
MQTT plugin: Add TLS support

jpmens:
This patch adds TLS support for newer libMosquitto. The following plugin configuration settings are supported:

```
CAfile "ca.pem"                   Enables TLS if set
Certfile "client-cert.pem"                optional
Keyfile "client-key.pem"          optional
TLSversion "tlsv1.2"              optional
Ciphers    "" optional
```	

If `CAfile` is set it must point to a PEM-encoded CA certificate used by the broker. This enables TLS so the `Port` number must be a TLS port on the broker. Optionally, both `Certfile` and `Keyfile` may be set to the paths of a client certificate and password-less PEM-encoded key files respectively. `TLSversion` may be set to specify a specific TLS version (e.g. `tlsv1.2`), but this defaults to whatever the broker uses. Also, optionally, `Ciphers` can be configured to use only the specified list of colon-separated ciphers. Only `CAfile` is mandatory to actually enable TLS.

addresses #1265

### DIFF
--- a/src/mqtt.c
+++ b/src/mqtt.c
@@ -24,6 +24,7 @@
  * Authors:
  *   Marc Falzon <marc at baha dot mu>
  *   Florian octo Forster <octo at collectd.org>
+ *   Jan-Piet Mens <jpmens at gmail.com>
  **/
 
 // Reference: http://mosquitto.org/api/files/mosquitto-h.html
@@ -48,6 +49,9 @@
 #ifndef MQTT_KEEPALIVE
 # define MQTT_KEEPALIVE 60
 #endif
+#ifndef SSL_VERIFY_PEER
+# define SSL_VERIFY_PEER  1
+#endif
 
 
 /*
@@ -67,6 +71,11 @@ struct mqtt_client_conf
     char               *username;
     char               *password;
     int                 qos;
+    char                *cacertificatefile;
+    char                *certificatefile;
+    char                *certificatekeyfile;
+    char                *tlsprotocol;
+    char                *ciphersuite;
 
     /* For publishing */
     char               *topic_prefix;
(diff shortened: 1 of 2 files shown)


src/mqtt.c
```diff
@@ -277,6 +286,26 @@ static int mqtt_connect (mqtt_client_conf_t *conf)
         return (-1);
     }
 
+#if LIBMOSQUITTO_MAJOR != 0
+    if (conf->cafile) {
+        status = mosquitto_tls_set(conf->mosq, conf->cafile, NULL,
+	                           conf->certfile, conf->keyfile, /* pw_callback */NULL);
+        if (status != MOSQ_ERR_SUCCESS) {
+            ERROR ("mqtt plugin: cannot mosquitto_tls_set: %s", mosquitto_strerror(status));
```
octo: I think we should destroy `conf->mosq` and return non-zero when any one of these functions fails.

src/mqtt.c
```diff
@@ -494,6 +523,10 @@ static int mqtt_write (const data_set_t *ds, const value_list_t *vl,
  *   StoreRates true
  *   Retain false
  *   QoS 0
+ *   CAfile "ca.pem"			Enables TLS if set
```
octo: Please add these to `src/collectd.conf.pod`, too. Please also consider to default on [Apache's names for these settings](http://httpd.apache.org/docs/2.2/mod/mod_ssl.html) – setting up TLS is, apparently, hard enough, so I'd like to make it as easy for the users as possible.

jpmens:
I've addressed your comments and have squashed the commits.

FYI, after renaming CAcert et. at. I see that other modules use precisely those names. Be that as it may, I've named them the way Apache uses them, but I refuse to say "SSL". ;-)

octo:
Thank you @jpmens!

I've reverted *CACertificateFile* back to *CACert* for consistency with other plugins.

The *write_http plugin* has the **ClientCert** and **ClientKey** options, so changing the *CertificateFile* and *CertificateKeyFile* to those names might also be a good idea. I don't think other plugins use the original *Certfile* or *Keyfile* names. Any thoughts?

PR #1349, opened 2015-11-24
generate configure script correctly when build rpm from git work copy

Dieken:
build.sh requires extra build dependencies such as libgcrypt-devel to
copy their m4 files, and the mock build environment requires flex, bison
and libtool packages to correctly build.

My build steps:

```bash
docker run -it --privileged --name collectd_build centos:6 bash

# as root in container
yum install -y epel-release
yum install -y mock bzip2 sudo
adduser builder
usermod -a -G mock builder
usermod -a -G wheel builder
sed -i -e 's/^\s*#\s*\(%wheel.*NOPASSWD:\)/\1/' /etc/sudoers
su builder

# as builder in container
cd && mkdir -p rpmbuild/{SOURCES,SPECS}
git clone https://github.com/collectd/collectd.git collectd-5.5.0
# ... apply this patch...
tar cjf collectd-5.5.0.tar.bz2 collectd-5.5.0/

cp collectd-5.5.0/contrib/redhat/collectd.spec rpmbuild/SPECS
mv collectd-5.5.0.tar.bz2 rpmbuild/SOURCES

sudo mock -r epel-6-x86_64 --enablerepo=epel --buildsrpm --spec ~/rpmbuild/SPECS/collectd.spec --sources ~/rpmbuild/SOURCES/
sudo mock -r epel-6-x86_64 --enablerepo=epel --no-clean --rebuild /var/lib/mock/epel-6-x86_64/result/collectd-5.5.0-1.el6.src.rpm
ls /var/lib/mock/epel-6-x86_64/result/
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -8,12 +8,15 @@ check_for_application ()
 	do
 		which "$PROG" >/dev/null 2>&1
 		if test $? -ne 0; then
-			cat >&2 <<EOF
+			[ "$PROG" = "yacc" ] && which bison >/dev/null 2>&1
+			if test $? -ne 0; then
+				cat >&2 <<EOF
 WARNING: \`$PROG' not found!
     Please make sure that \`$PROG' is installed and is in one of the
     directories listed in the PATH environment variable.
 EOF
-			GLOBAL_ERROR_INDICATOR=1
+				GLOBAL_ERROR_INDICATOR=1
+			fi
 		fi
 	done
 }
--- a/contrib/redhat/collectd.spec
+++ b/contrib/redhat/collectd.spec
@@ -228,7 +228,7 @@ Source:		http://collectd.org/files/%{name}-%{version}.tar.bz2
 License:	GPLv2
 Group:		System Environment/Daemons
 BuildRoot:	%{_tmppath}/%{name}-%{version}-root
-BuildRequires:	libgcrypt-devel, kernel-headers, libtool-ltdl-devel, libcap-devel
+BuildRequires:	autoconf, automake, which, flex, bison, libgcrypt-devel, kernel-headers, libtool, libtool-ltdl-devel, libcap-devel
 Vendor:		collectd development team <collectd@verplant.org>
 
 %if 0%{?el7:1}
@@ -1618,6 +1618,8 @@ Collectd utilities
 %else
 %define _with_zookeeper --disable-zookeeper
 %endif
+
+test -e configure || ./build.sh
 
 %configure CFLAGS="%{optflags} -DLT_LAZY_OR_NOW=\"RTLD_LAZY|RTLD_GLOBAL\"" \
 	--disable-static \


Dieken:
@rubenk 

reply to https://github.com/Dieken/collectd/commit/45662cc801ed2be72b2393335264b50cc46717ee,  libtool depends on automake and autoconf, below is executed in the chroot setup by mock:
```shell
bash-4.1# rpm -q --whatrequires autoconf
automake-1.11.1-4.el6.noarch
libtool-2.2.6-15.5.el6.x86_64
bash-4.1# rpm -q --whatrequires automake
gtk2-devel-2.24.23-6.el6.x86_64
libcurl-devel-7.19.7-46.el6.x86_64
libtool-2.2.6-15.5.el6.x86_64
```
but anyway it's a good idea to explicitly declare the build dependencies,  I just committed again, please double check, thanks!

build.sh
```diff
@@ -18,7 +18,7 @@ EOF
 	done
 }
 
-check_for_application lex yacc autoheader aclocal automake autoconf
```
octo: This only checks if there is an executable named `yacc`, which *bison* should also provide. Please don't make this *bison* specific as collectd should also build with other implementations of yacc.

octo:
Thank you @Dieken!

Dieken:
@octo  Unfortunately, CentOS 6 doesn't install a symlink from bison to yacc, I have no idea,  so I add an extra check in build.sh,  pushed again, please double check, thanks!

mfournier:
Mmmh, I don't think it's a very good idea to go down this path...

IMHO the workflow to build packages should be:
- grab (or build) a release tarball
- add the packaging bits on top of it (rpm specfile/debian dir/whatever)
- build the package for each target distro/architecture

This patch attempts to merge the 2 first steps, which I think should preferrably stay separated.

You can easily get around what you're attempting to fix by modifying your build steps as follows:

```
docker run -it --privileged --name collectd_build centos:6 bash

# as root in container
yum install -y epel-release
yum install -y mock bzip2 sudo
# install the autotools & deps at this step:
yum -y install autoconf automake bison cpp flex gcc glibc-devel libgcrypt-devel libtool libtool-ltdl-devel m4 make pkgconfig redhat-rpm-config tar
adduser builder
usermod -a -G mock builder
usermod -a -G wheel builder
sed -i -e 's/^\s*#\s*\(%wheel.*NOPASSWD:\)/\1/' /etc/sudoers
su builder

# as builder in container
cd && mkdir -p rpmbuild/{SOURCES,SPECS}
git clone https://github.com/collectd/collectd.git collectd-5.5.0
# generate release tarball now
(cd collectd-5.5.0 && ./build.sh && configure && make dist)

cp collectd-5.5.0/contrib/redhat/collectd.spec rpmbuild/SPECS
mv collectd-5.5.0*.tar.bz2 rpmbuild/SOURCES

# edit the "Version:" tag in the specfile

sudo mock -r epel-6-x86_64 --enablerepo=epel --buildsrpm --spec ~/rpmbuild/SPECS/collectd.spec --sources ~/rpmbuild/SOURCES/
sudo mock -r epel-6-x86_64 --enablerepo=epel --no-clean --rebuild /var/lib/mock/epel-6-x86_64/result/collectd-5.5.0-1.el6.src.rpm
ls /var/lib/mock/epel-6-x86_64/result/
```

@rubenk any thoughts about this ?

PS @Dieken if you just want to have RPMs from current master, you can also use the nightly builds: https://github.com/collectd/collectd-ci/blob/master/README.md

mfournier:
Also, whatever change is made to the rpm specfile, please make sure they work on all of EL5, EL6 and EL7, as we depend on it working for all of these for the nightlies.

Dieken:
@mfournier,  the collectd.spec file does check whether "configure" exists,  the patch doesn't conflict with your build steps, personally I prefer to let mock setup the build environment according to BuildRequires in the collectd.spec file,  it's convenient and assure correct m4 files copied from correct package versions **even if the host OS isn't same with the mock environment**, for example, build 32bit rhel-6 pkgs on 64bit rhel-7.

I checked pkg names on centos-7, which/automake/autoconf/libtool/flex/bison keep unchanged, I guess it's true for centos-5 too but I don't verified (no centos-5 docker image locally),  so I think the spec file should be fine on EL5/6/7.

Could you add a link for the nightly build page to http://collectd.org/download.shtml#rpm ?  The collectd-ci page is very useful,  thank you very much!

build.sh
rubenk: @octo are you sure? last time I checked collectd didn't build with other yacc implementations.

rubenk:
I don't think we should do this. The spec file is meant to be used in combination with a release tarball, in which case you don't need bison, flex, autoconf etc.

When building from a git checkout, we generally assume that you are a developer, and have the right development environment setup.

Adding more BuildRequirements for the latter case will slow down the former when building in mock.

Dieken:
@rubenk,  building speed isn't a big issue here, the extra packages are small and take a very small fraction of total build time,  the spec build requires to download >200 packages.

flex and bison are required, or the spec build will fail,  they are build time requirements, not release time requirements.

which/autoconf/automake/libtool is release time requirements, to run "./build.sh" on git work copy.  **I don't think this harms anything**, if somebody already has "./configure", just waste several seconds to download four small packages,  if somebody does't have "./configure",  it makes sure correct m4 files from correct packages,  no need to realize "oh the README doesn't mention extra build requirements, I need to parse collectd.spec to install needed devel packages first, then I waste time to download those packages twice, once out of chroot, once in chroot".

So, could you kindly merge this four lines small patch for other potential contributors' convenience?

rubenk:
Flex and bison are only required when building from git, not from a release tarball.
If you can show otherwise, that is a bug, so please let us know.

Those few seconds might not seem like much, but as a downstream maintainer I rebuild release tarballs all the time, and I'd rather don't add stuff I don't need. The same goes for our CI infra which has to do this on every build.

As for each most open software, we assume developers have development tools installed. The README doesn't mention this because it is meant for end-users, not developers.

I'm not opposed to adding a HACKING document, or extending build.sh to detect more missing build requirements though.

As for figuring out which -devel packages you need to install, yum-builddep is your friend.

PR #1353, opened 2015-11-25
c_avl_pick needs to update t->size.

coreykosak:
```c_avl_pick``` needs to update the ```size``` field of the tree when it deletes a node.

### DIFF
--- a/src/daemon/utils_avltree.c
+++ b/src/daemon/utils_avltree.c
@@ -652,6 +652,7 @@ int c_avl_pick (c_avl_tree_t *t, void **key, void **value)
 	*value = n->value;
 
 	free_node (n);
+	--t->size;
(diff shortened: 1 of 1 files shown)


mfournier:
Thanks @coreykosak !

I don't fully grasp the implications of this change. What effect or behaviour change would this patch have on the plugins using the avl tree ?

octo:
@coreykosak Thank you very much for finding and fixing this issue! I've pushed your patch as 007d96fe10beb6c6d4f20c79851404be18a3418f.

@mfournier Subsequent calls to `c_avl_size()` will return the correct number of elements currently in the tree. If this is a problem for any plugin, it needs to be fixed.

I've also extended the unit test in the *master* branch to test `c_avl_pick()` which demonstrated this problem.

PR #1354, opened 2015-11-25
Remove duplicate call to pthread_mutex_destroy.

coreykosak:
Destroying the same mutex twice was probably not intentional.


mfournier:
Thanks a lot @coreykosak ! Cherry-picked as b2c062a99.

PR #1355, opened 2015-11-25
Fix for subscribing to MQTT

njh:
Hello,

I have been testing subscribing to MQTT topics but was initially unable to get it to work. After some investigation, I found that it was checking that the MQTT payload was NULL-terminated. However it is not necessary, or typical, to NULL-terminate the payload because there is a length field instead.

Applying this patch fixes it for me.

nick.

### DIFF
--- a/src/mqtt.c
+++ b/src/mqtt.c
@@ -176,9 +176,10 @@ static void on_message (
     char *payload;
     int status;
 
-    if ((msg->payloadlen <= 0)
-            || (((uint8_t *) msg->payload)[msg->payloadlen - 1] != 0))
+    if (msg->payloadlen <= 0) {
+        DEBUG ("mqtt plugin: message has empty payload");
         return;
+    }
 
     topic = strdup (msg->topic);
     name = strip_prefix (topic);
@@ -207,7 +208,15 @@ static void on_message (
     }
     vl.values_len = ds->ds_num;
 
-    payload = strdup ((void *) msg->payload);
+    payload = malloc (msg->payloadlen+1);
+    if (payload == NULL)
+    {
+        ERROR ("mqtt plugin: malloc for payload buffer failed.");
+        return;
+    }
+    memmove (payload, msg->payload, msg->payloadlen);
+    payload[msg->payloadlen] = 0;
+
     DEBUG ("mqtt plugin: payload = \"%s\"", payload);
     status = parse_values (payload, &vl, ds);
     if (status != 0)


mfournier:
Uhoh, it looks like some gcc versions choke on this change:

```
mqtt.c: In function 'on_message':
mqtt.c:217:5: error: pointer targets in passing argument 2 of 'strncpy' differ in signedness [-Werror=pointer-sign]
/usr/include/x86_64-linux-gnu/bits/string3.h:118:1: note: expected 'const char * __restrict__' but argument is of type 'uint8_t * const'
make[3]: *** [mqtt_la-mqtt.lo] Error 1
```

Any clue ?

Thanks @njh !

njh:
Hm! Not sure, I will take a look later with gcc. May be that it just needs to be casted.

I suspect it is because a the MQTT payload doesn't have to be a string, so isn't a char*

src/mqtt.c
```diff
@@ -207,7 +208,14 @@ static void on_message (
...
     vl.values_len = ds->ds_num;
 
-    payload = strdup ((void *) msg->payload);
+    payload = calloc (msg->payloadlen+1, sizeof (*payload));
+    if (payload == NULL)
+    {
+        ERROR ("mqtt plugin: calloc for payload buffer failed.");
+        return;
+    }
+    strncpy (payload, msg->payload, msg->payloadlen);
```
octo: Please use `memmove()` instead of `strncpy()` if the payload is not necessarily null terminated. Also, enforce null termination explicitly (relying on `calloc()` is too subtle to be easily readable):

```c
memmove (payload, msg->payload, msg->payloadlen);
payload[msg->payloadlen] = 0;
```

This should also fix the implicit `unsigned char *` to `char *` casting that @mfournier pointed out.

src/mqtt.c
```diff
@@ -207,7 +208,14 @@ static void on_message (
     }
     vl.values_len = ds->ds_num;
 
-    payload = strdup ((void *) msg->payload);
+    payload = calloc (msg->payloadlen+1, sizeof (*payload));
```
octo: Please unify the way this size is specified. Either simply use `msg->payloadlen` (preferred, essentially this will mean to revert to `malloc(msg->payloadlen+1)`) or `msg->payloadlen * sizeof (*payload)` (i.e. adapt `memmove()` below to do that multiplication, too).

octo:
Thank you very much for your patch @njh. I agree that we should probably not assume the payload to be null terminated, especially since this appears to be somewhat unusual with MQTT.

njh:
New patch pushed taking into account @octo 's comments

octo:
Thank you so much @njh!

PR #1356, opened 2015-11-25
memcached_connections is of type GAUGE, needs to be submitted as a gauge.

coreykosak:
Put another way, line 442 needs to be consistent with line 438.

(credit: Jason Lawrence (https://github.com/chroniton) for finding this).

### DIFF
--- a/src/memcached.c
+++ b/src/memcached.c
@@ -439,7 +439,7 @@ static int memcached_read (user_data_t *user_data)
     }
     else if (FIELD_IS ("listen_disabled_num"))
     {
-      submit_derive ("memcached_connections", "listen_disabled", atof (fields[2]), st);
+      submit_gauge ("memcached_connections", "listen_disabled", atof (fields[2]), st);
     }
 
     /*


mfournier:
Thanks a lot @coreykosak !

I just want to be sure this is the right fix. According to fa91b25bcb, this value is "a counter which increments when the connection limit to memcached is reached and a socket listen is disabled."

If this is correct, submitting it as a derive is right, but the chosen data type is not.

What do you think ?

octo:
Thanks @coreykosak! Agreed, lines 438 and 442 contradict. If this is indeed a counter, another *type* has to be used.

octo:
According to https://code.google.com/p/memcached/wiki/NewServerMaint and https://code.google.com/p/memcached/wiki/Timeouts, `listen_disabled_num` is a counter. I think this line should therefore be:
```c
submit_derive ("connections", "listen_disabled", atof (fields[2]), st);
/*              ^^^^^^^^^^^ */
```

rpv-tomsk:
> If this is correct, submitting it as a derive is right, but the chosen data type is not.
> 
> What do you think ?

Yes, choosen data type is incorrect. That is not 'connections', but a 'counter how many times limit was reached'. We are unable to know, how many connections was lost because of that.
We need to register another issue about this.

PR #1357, opened 2015-11-26
table plugin: ignore comment lines

fanyeren:
Signed-off-by: feng xiahou <xiahoufeng@yahoo.com>

### DIFF
--- a/src/table.c
+++ b/src/table.c
@@ -492,8 +492,8 @@ static int tbl_read_table (tbl_t *tbl)
 			buf[sizeof (buf) - 1] = '\0';
 			log_err ("Table %s: Truncated line: %s", tbl->file, buf);
 		}
-
-		if (0 != tbl_parse_line (tbl, buf, sizeof (buf))) {
+		# ignore comment lines
+		if ('#' != buf[0] && 0 != tbl_parse_line (tbl, buf, sizeof (buf))) {
 			log_err ("Table %s: Failed to parse line: %s", tbl->file, buf);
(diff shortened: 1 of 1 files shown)


src/table.c
```diff
@@ -492,8 +492,8 @@ static int tbl_read_table (tbl_t *tbl)
 			buf[sizeof (buf) - 1] = '\0';
 			log_err ("Table %s: Truncated line: %s", tbl->file, buf);
 		}
-
-		if (0 != tbl_parse_line (tbl, buf, sizeof (buf))) {
+		# ignore comment lines
```
octo: This is not a valid C comment. Please use:
```c
/* ignore comment lines */
```

octo:
Hi @fanyeren and thank you very much for your patch! I think this is a good thing to add to the *table plugin*, but it should go into `tbl_parse_line()` and be configurable, similar to `Separator`:

 *  Add a `Comment` configuration option which one can set to `"#"`, `"//"` or whatever the file is using.
 *  In `tbl_parse_line()`, after the calls to `strtok_r()`, add something along the lines of:
```c
if ((strlen (tbl->comment) != 0)
    && (strncmp (fields[0], tbl->comment, strlen (tbl->comment)) == 0))
  return 0;
```

This will make sure that leading spaces / separator characters are stripped before checking for the comment character.

Best regards,
—octo

fanyeren:
@octo I'm not familiar with c-stuff languages，can you fix this？tks

PR #1360, opened 2015-11-26
table plugin: log_warn is better than log_err here

fanyeren:


octo:
Hi @fanyeren, thanks for your patch!

A *warning* should be used when there is a problem but we can continue. Line 493 fits this I think.

An *error* should be used when there is a problem and we need to abort. Lines 461 and 497 fit this and should therefore continue to use `log_err()`. Can you please revert these two?

Thanks and best regards,
—octo

fanyeren:
@octo one file may have many lines，if there is one line that contain a bad format，that file is still general correct。I think it should be a warning，not an error，in this situation.

octo:
Fair enough, I didn't realize that `tbl_parse_line()` was itself the called from within a loop over the file. Looks good to me.

Issue #929, opened 2015-02-06
Statsd: counters always incremented


duylong:
Hi, 

I want to count the error in httpd. I use "increment" method in order to aggregate data with statsd.

For each event, I send this output ```syslog.httpd.error:1|c```

In collectd.conf, I use "DeleteCounters" to true, but it seem that my counter is never reset. My collectd output:

```
[{"values":[1179],"dstypes":["derive"],"dsnames":["value"],"time":1423215320.585,"interval":10.000,"host":"plop","plugin":"statsd","plugin_instance":"","type":"derive","type_instance":"syslog.httpd.error"}]
...
[{"values":[1016],"dstypes":["derive"],"dsnames":["value"],"time":1423215000.585,"interval":10.000,"host":"plop","plugin":"statsd","plugin_instance":"","type":"derive","type_instance":"syslog.httpd.error"}]
...
[{"values":[816],"dstypes":["derive"],"dsnames":["value"],"time":1423214600.585,"interval":10.000,"host":"plop","plugin":"statsd","plugin_instance":"","type":"derive","type_instance":"syslog.httpd.error"}]
...
```
Do you know this issue?

Thank you for your help,

octo:
I think this is due to a misconception what `DeleteCounters` does. The manpage reads:

> These options control what happens if metrics are not updated in an interval. If set to **False**, the default, metrics are dispatched unchanged, i.e. the rate of counters and size of sets will be zero, timers report `NaN` and gauges are unchanged. If set to **True**, the such metrics are not dispatched and removed from the internal cache.

So, if a counter was *not* incremented in the last 10 seconds (or whatever your interval is set to), and this option is enabled, then the metric will be purged to free memory.

Do you have any suggestions how the manpage entry could be improved? Thanks!
—octo

duylong:
You can add a example in order to improve the manpage ;)

What happens if metrics are updated in an interval ? "DeleteCounter" is ignored ? 
I tried the 2 choices, my value is always incremented (I have a sum). Is this normal behavior ?

duylong:
After some tests, no problem with statsd (nodejs app). There a a problem somewhere...

Nobody has this problem using Logstash as client?

mikechack:
Statsd plugin does not behave the same way as the etsy/statsd daemon. I think the true behavior for most statsd implementations for counters is to reset the counter after each sampling period. This is not how the current plugin works. As noted even if the DeleteCounters is set, the value is retained as long as there is a value samled for the current period.

[...]

PR #1363, opened 2015-11-27
statsd plugin: Implement the "CounterSum" option.

octo:
CC: @jamtur01, @pyr, @duylong
Issues: #929, #1282, #1311

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1132,6 +1132,7 @@
 #  DeleteTimers   false
 #  DeleteGauges   false
 #  DeleteSets     false
+#  CounterSum     false
 #  TimerPercentile 90.0
 #  TimerPercentile 95.0
 #  TimerPercentile 99.0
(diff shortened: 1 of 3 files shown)


mfournier:
@octo, would this patch be merged in, keep in mind that @pyr's patch from #1311 was cherry-picked to collectd-5.4 & collectd-5.5 (but not part of any release yet).

The reasoning we had was that it qualified as a bugfix as it intended to make collectd's statsd compatible with the reference implementation.

So either this one would need to get backported too, or @pyr's patch reverted, for the sake of a consistent upgrade from 5.4/5.5 to 5.6.

octo:
@mfournier Thanks for the heads up! I'm a bit on the fence, to be honest: on the one hand #1311 is already in the 5.4 branch, on the other hand I don't really want to push new config options to that branch …

jamtur01:
I'm not fussed about the 5.4 branch. I asked @pyr to back port to 5.5 because that's the version where our collectd is likely to sit for a while.

jamtur01:
And thanks for this fix!

rj33:
What was the final decision on merging this into 5.5.x? Are we waiting for 5.6.0 at this stage to see this fix in a release?

PR #1364, opened 2015-11-27
the swrite interface is 'non-zero'

sampointer:
In many other implementations status is checked to be "non-zero", as per the interface definition to `swrite`.

### DIFF
--- a/src/write_graphite.c
+++ b/src/write_graphite.c
@@ -113,7 +113,7 @@ static int wg_send_buffer (struct wg_callback *cb)
     ssize_t status = 0;
 
     status = swrite (cb->sock_fd, cb->send_buf, strlen (cb->send_buf));
-    if (status < 0)
+    if (status != 0)
     {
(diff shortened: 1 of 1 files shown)


octo:
Hi @sampointer, thank you very much for your patch!

You're right, checking for non-zero is probably better. However, in practice only negative values are returned by `swrite()` so I don't quite understand how this patch can fix #1365. Could you give some more details?

Thanks,
—octo

sampointer:
I've not tested that it does, I'll be honest. I could not get a local build
out of the build chain. However, given that that check gates `close()` and
ending up in CLOSE_WAIT is a failure to call `close()` it seems like a good
punt. Even so, pedantry makes this a good change regardless.

octo:
Ack, I've cherry-picked this into *collectd-5.4* as ea4e6f8f62e5ec51b95387d1d111476468f8a7b1. I'm fairly certain that this doesn't fix an issue, though.

PR #1367, opened 2015-11-27
processes.c: plug memory leak in error path

rubenk:
Found with Infer (http://fbinfer.com/)


octo:
Thanks @rubenk! I've pulled this into the *collectd-5.4* branch as e01209268a710341765879eb32412027fb60257d.

PR #1370, opened 2015-11-28
write_kafka: fix indentation

rubenk:
It was all over the place.
I went with 4 spaces since that's what the majority
of code used.


octo:
Thanks @rubenk!

PR #1385, opened 2015-11-29
configure.ac: remove wrong gcrypt check

rubenk:
After all the effort to detect GCRYPT_CPPFLAGS and GCRYPT_LIBS
in previous checks, this check blindly overwrites them, even
if the AM_PATH_LIBGCRYPT macro isn't found.

Also, I think it's better to check for specific functionality in a library
than an explicit minimum version, which we already did.


octo:
Makes sense, LGTM

rubenk:
Thanks for the review @octo :)

PR #1387, opened 2015-11-29
redis plugin: make sure buffer is null-terminated

rubenk:
CID 37938


octo:
LGTM, thanks @rubenk!

PR #1388, opened 2015-11-29
curl_json: remove dead code

rubenk:
CID 37963


octo:
LGTM. At your options: move the call to `yajl_get_error()` on the same line as the declaration of `msg`.

rubenk:
That looks much better indeed. Done.

PR #1396, opened 2015-12-02
write_http plugin: Implement sending of notifications.

octo:
This patch implements POSTing of notifications to a configurable URL.

It implements the alert format used by [prometheus/alertmanager](https://github.com/prometheus/alertmanager) (the *dev* branch) which looks like this:

```json
[{
  "labels": {
    "alertname": "collectd_cpu",
    "instance":  "host.example.com",
    "severity":  "FAILURE",
    "service":   "collectd",
    "cpu":       "0",
    "type":      "wait"
  },
  "annotations": {
    "summary": "..."
  },
  "startsAt": "2015-12-02T10:33:08.451491682+0100"
}]
```

### DIFF
--- a/src/utils_format_json.c
+++ b/src/utils_format_json.c
@@ -1,6 +1,6 @@
 /**
  * collectd - src/utils_format_json.c
- * Copyright (C) 2009       Florian octo Forster
+ * Copyright (C) 2009-2015  Florian octo Forster
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
@@ -26,11 +26,15 @@
 
 #include "collectd.h"
 
+#include "utils_format_json.h"
+
 #include "plugin.h"
 #include "common.h"
-
 #include "utils_cache.h"
-#include "utils_format_json.h"
+
+#if HAVE_LIBYAJL
+#include <yajl/yajl_gen.h>
+#endif
 
 static int json_escape_string (char *buffer, size_t buffer_size, /* {{{ */
     const char *string)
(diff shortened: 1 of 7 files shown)


src/utils_format_json.c
```diff
@@ -484,4 +495,215 @@ int format_json_value_list (char *buffer, /* {{{ */
         store_rates, (*ret_buffer_free) - 2));
 } /* }}} int format_json_value_list */
 
+#if HAVE_LIBYAJL
+static int json_add_string (yajl_gen g, char const *str) /* {{{ */
```
rubenk: Minor nit: const char *
I don't mind personally but I noticed this confuses less experienced contributors who think the order does matter.

src/utils_format_json_test.c
```diff
@@ -0,0 +1,171 @@
...
+#include "collectd.h"
+#include "utils_format_json.h"
+#include "common.h" /* for STATIC_ARRAY_SIZE */
+
+#include <yajl/yajl_common.h>
+#include <yajl/yajl_parse.h>
+#if HAVE_YAJL_YAJL_VERSION_H
+# include <yajl/yajl_version.h>
+#endif
+#if defined(YAJL_MAJOR) && (YAJL_MAJOR > 1)
```
rubenk: undefined evaluates to 0 so I think you can skip the left part safely.

src/utils_format_json_test.c
```diff
@@ -0,0 +1,171 @@
...
+# define HAVE_YAJL_V2 1
+#endif
+
+struct label_s
+{
+  char *key;
+  char *value;
+};
+typedef struct label_s label_t;
```
rubenk: I've always wondered why you tag the structs and not do:
```
typedef struct
{
 char *key;
  char *value;
} label_t;
```
since you're not using the tag?

src/utils_format_json_test.c
```diff
@@ -0,0 +1,171 @@
...
+# include <yajl/yajl_version.h>
+#endif
+#if defined(YAJL_MAJOR) && (YAJL_MAJOR > 1)
+# define HAVE_YAJL_V2 1
+#endif
+
+struct label_s
+{
+  char *key;
+  char *value;
```
rubenk: can you make these const?

rubenk:
I left a few minor comments inline, otherwise LGTM :)

src/utils_format_json.c
octo: The order *does* matter, which is exactly why I prefer this style: `const` always effects the keyword to its left: `char const *` is a variable pointer to a constant char, `char * const` is a constant pointer to a variable char.

When `const` is the first keyword, and only then, a special case kicks in and it effects the keyword to the right. I slightly prefer to avoid that special case because then constructs like `char const * const *ptr` become much easier to read (`const char * const *ptr` is just confusing).

src/utils_format_json_test.c
octo: The style you propose is cleaner in this case. When using opaque types you have to declare (but not define) the struct before the typedef:

```c
struct foo_s;
typedef struct foo_s foo_t;
```

That's why I sometimes fall into that style … I'm going to clean that up.

octo:
Comments fixed, not yet tested though.

src/utils_format_json_test.c
octo: Done.

src/utils_format_json_test.c
octo: I think so. Done.

src/utils_format_json.c
rubenk: Sorry, I should have been more specific, I was referring to the difference between `const char *` and `char const *`

I agree that `char const * const *ptr` is indeed much easier to read :)

src/utils_format_json_test.c
rubenk: Got it, thanks.

rubenk:
Looks good, still needs a rebase though. If you merge, can you also add a blank line below collectd.h?

octo:
Rebased on *master*.

rubenk:
This fails the tests under Valgrind on 32-bit platforms:
https://ci.collectd.org/job/master-build-on-jessie-i386-with-default-toolchain/430/artifact/collectd-5.5.2.1191.gf089ccb/src/test-suite.log

Could you have a looksie?

octo:
Will do

rubenk:
It does only seem to fail on 32-bit platforms that have libyajl 2.

octo:
I think 4f4ad5dd93776f330a0fb28fbece7157108e0c4c fixes this.

rubenk:
Seems like it did, thanks!

PR #1397, opened 2015-12-02
snmp: accept to use IpAddress for instances

vincentbernat:
Some interesting MIB are indexed with an IP address (for example,
BGP4-MIB). Make collectd able to use those indexes as instances. Format
the IP as string. IPv6 are usually encoded as Octet String, so, they
should be already supported.

Is the use of dot problematic? When used with graphite, it is a separator. Should I replace with a special dot (unicode) or a dash?

### DIFF
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -1106,6 +1106,14 @@ static int csnmp_strvbcopy (char *dst, /* {{{ */
     src = (char *) vb->val.string;
   else if (vb->type == ASN_BIT_STR)
     src = (char *) vb->val.bitstring;
+  else if (vb->type == ASN_IPADDRESS)
+  {
+    return ssnprintf (dst, dst_size, "%d.%d.%d.%d",
+          (uint8_t)vb->val.string[0],
+          (uint8_t)vb->val.string[1],
+          (uint8_t)vb->val.string[2],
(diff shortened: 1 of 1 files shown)


pyr:
I'm biased because we work next to each other, but :+1: from me.

pyr:
@vincentbernat the dots will be handled by the escape character pragma IIRC

src/snmp.c
```diff
@@ -1106,6 +1106,14 @@ static int csnmp_strvbcopy (char *dst, /* {{{ */
     src = (char *) vb->val.string;
   else if (vb->type == ASN_BIT_STR)
     src = (char *) vb->val.bitstring;
+  else if (vb->type == ASN_IPADDRESS)
+  {
+    return snprintf(dst, dst_size, "%d.%d.%d.%d",
```
octo: Please use `ssnprintf()` (note the extra "s"). The difference is that `ssnprintf()` guarantees null termination, `snprintf()` does not.

octo:
Thanks @vincentbernat, LGTM once the above comment is resolved :)

vincentbernat:
OK, PR updated. I have also fixed indentation and spaces.

octo:
Thanks @vincentbernat!

PR #1398, opened 2015-12-02
GenericJMX plugin: memory leak, indent and compile warn fixes

frapex:
- call close() on JMXConnector if JMX connection fails; fixes memory leak
- fix indentation errors
- fix "unchecked" compile warnings
- some refactoring

The memory leak was observed when a previously established org.jboss.remotingjmx connection fails. jstack shows a lot of threads not being cleaned-up. Calling close() on the JMXConnector solves this.


octo:
LGTM, thanks @frapex!

PR #1400, opened 2015-12-04
ping plugin: Help!

nitreason:
Hi ALL, I configured the `collectd.conf` with opening the ping plugin. But it seems that there is no data sent to the influxdb database. How to use the ping plugin?  My configurations as following:

collectd.conf:

```
LoadPlugin ping
<Plugin ping>
        Host "TEST"
</Plugin>
```

Please help. Thanks.


octo:
**DO NOT SUBMIT** (this should be an *issue*, not a *pull request*)

Hi @nitreason, the *ping plugin* requires root privileges to run, otherwise you should see error messages. What do the logs say?

Best regards,
—octo

PR #1403, opened 2015-12-04
git ignore src/daemon/test_utils_time

chicha:
Add src/daemon/test_utils_time to .gitignore.
Probably forgotten when introducing db1391aaa66b8b8fad82219494f61f3452441f62


octo:
Thanks @chicha!

PR #1405, opened 2015-12-04
teamspeak2: fix potential use after close

rubenk:
If connecting to the last host fails, we exit the loop
with a closed fd, which we try to fdopen() later on.

CID #38038


octo:
Cherry-picked into *collectd-5.4* as be7860449c290ec30458a514d36f3bee2a56991b

PR #1406, opened 2015-12-04
snmp plugin: don't read from uninitialized pointer

rubenk:
if the first snmp_pdu_create inside the loop returns an error,
we exit the loop with res uninitialized and then call snmp_free_pdu on it.

CID #38037


octo:
LGTM, thank you very much :-)

rubenk:
Only a hundred to go, but I'll keep them coming ;)

PR #1407, opened 2015-12-04
readlink: null-terminate result of readlink

rubenk:
readlink doesn't do it for us.

CID #38027

### DIFF
--- a/src/madwifi.c
+++ b/src/madwifi.c
@@ -821,14 +821,15 @@ check_devname (const char *dev)
 
 	if (dev[0] == '.')
 		return 0;
-	
-	ssnprintf (buf, sizeof (buf), "/sys/class/net/%s/device/driver", dev);
-	buf[sizeof (buf) - 1] = 0;
 
-	memset (buf2, 0, sizeof (buf2));
+	ssnprintf (buf, sizeof (buf), "/sys/class/net/%s/device/driver", dev);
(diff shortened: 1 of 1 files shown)


src/madwifi.c
```diff
@@ -821,15 +821,16 @@ check_devname (const char *dev)
 
 	if (dev[0] == '.')
 		return 0;
-	
+
 	ssnprintf (buf, sizeof (buf), "/sys/class/net/%s/device/driver", dev);
-	buf[sizeof (buf) - 1] = 0;
+	buf[sizeof (buf) - 1] = '\0';
 
-	memset (buf2, 0, sizeof (buf2));
```
octo: I don't understand why the old code is broken: `buf2` is initialized to zero, then `readlink()` will write at most `sizeof(buf2)-1` bytes, so the trailing null byte should not be overwritten. It's not the best style, but not broken, right?

rubenk: You're right of course, in this case Coverity is not smart enough to see that, and neither was I last night ;)

I still like this, but as a cleanup, to prevent the same thing from happening again.

rubenk:
I changed the commit message so it's clear now this is a cleanup patch. Thanks @octo!

octo:
Cherry-picked into *collectd-5.4* as 51f407eeaf373508098a06b9e352b7ed794d6335

PR #1409, opened 2015-12-04
utils_mount.c: check for null before dereferencing

rubenk:
CID #38025


octo:
Cherry-picked into *collectd-5.4* as 27a47fc62d90a115d7a00753a06ffab71e92f487

PR #1410, opened 2015-12-04
table plugin: fix check for realloc

rubenk:
We were checking the wrong thing.

CID #38024


octo:
Cherry-picked into *collectd-5.4* as 6a69ca987e1c3774428c9ab2778624ee80f343c6

PR #1414, opened 2015-12-04
multimeter plugin: fix fd leak

rubenk:
If open returns 0 we leak fd.

CID #38016


octo:
Cherry-picked into *collectd-5.4* as b8913bc047474aa0a3bfbf03ecbb3b5050605ed1

PR #1416, opened 2015-12-04
Fixed duplicate interface metrics for Solaris

yogeswaran:
On Solaris, interface plugin publishes duplicate metrics, ie for the same time interval (with the same epoch) it publishes multiple data points for the same interface name.

Collectd interface plugin uses kstat on Solaris. As per kstat documents (http://docs.oracle.com/cd/E23824_01/html/821-1468/kstat-3kstat.html#REFMAN3Ekstat-3kstat), interface name is unique ONLY within a module/instance. Following tuple is considered unique:

```  
<ks_module, ks_instance, ks_name>
```

This fix publishes above tuple as plugin_instance instead of just ks_name and avoids duplicate metrics.

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -92,13 +92,13 @@ static const char *config_keys[] =
 static int config_keys_num = 2;
 
 static ignorelist_t *ignorelist = NULL;
-static _Bool unique_name = 0;
 
 #ifdef HAVE_LIBKSTAT
 #define MAX_NUMIF 256
 extern kstat_ctl_t *kc;
 static kstat_t *ksp[MAX_NUMIF];
 static int numif = 0;
+static _Bool unique_name = 0;
 #endif /* HAVE_LIBKSTAT */
 
 static int interface_config (const char *key, const char *value)
@@ -119,8 +119,12 @@ static int interface_config (const char *key, const char *value)
 	}
 	else if (strcasecmp (key, "UniqueName") == 0)
 	{
+		#ifdef HAVE_LIBKSTAT
 		if (IS_TRUE (value))
 			unique_name = 1;
+		#else
+			WARNING ("interface plugin: the \"UniqueName\" option is only valid on Solaris.");
+		#endif /* HAVE_LIBKSTAT */
 	}
 	else
 	{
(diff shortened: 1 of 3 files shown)


octo:
Thank you very much for your patch, @yogeswaran!

Unfortunately, this patch changes the metric names of all interface metrics, not just those that result in duplicates. We consider this to be backwards incompatible, so we can't accept the patch in its current form.

The usual way for such changes is to "hide" them behind a configuration option so that users who continue to use their existing configuration don't run into changed metric names. Would you be interested in adding such a config option?

Best regards,
—octo

yogeswaran:
sure, I can add a configuration option, thanks for quickly looking into this.

yogeswaran:
@octo - I have added "UniqueName" configuration option as you suggested.

src/interface.c
```diff
@@ -116,6 +117,11 @@ static int interface_config (const char *key, const char *value)
 			invert = 0;
 		ignorelist_set_invert (ignorelist, invert);
 	}
+	else if (strcasecmp (key, "UniqueName") == 0)
+	{
+		if (IS_TRUE (value))
```
octo:
```c
#if HAVE_LIBKSTAT
unique_name = IS_TRUE (value);
#else
WARNING ("interface plugin: the \"UniqueName\" option is only valid on Solaris.");
#endif
```

src/interface.c
```diff
@@ -297,6 +304,11 @@ static int interface_read (void)
 		if (kstat_read (kc, ksp[i], NULL) == -1)
 			continue;
 
+		if (unique_name)
+			snprintf(iname, sizeof(iname), "%s_%d_%s", ksp[i]->ks_module, ksp[i]->ks_instance, ksp[i]->ks_name);
```
octo: Please use `ssnprintf()` (note the extra "s"): that version will ensure that `iname` is null-terminated.

src/interface.c
```diff
@@ -297,6 +304,11 @@ static int interface_read (void)
...
+		else
+			snprintf(iname, sizeof(iname), "%s", ksp[i]->ks_name);
```
octo: Please use `sstrncpy()` instead.

octo:
Hi @yogeswaran, thank you very much for your update! This looks good overall and I think we're close to being able to get this merged. There are only small details missing now:

 *  The code review comments I left inline just now.
 *  This option should be added as a comment to `src/collectd.conf.in`, from which the default config file is generated.
 *  This option should be added to `src/collectd.conf.pod` from which the *collectd.conf(5)* manual page is generated.

Thank you for you work and best regards,
—octo

yogeswaran:
@octo : I made all the changes suggested by you, can you review now?

octo:
Thanks for the work you put into this and thanks for bearing with me :)

I've rebased the changes to the *master* branch and pushed them.

yogeswaran:
Thanks @octo

PR #1421, opened 2015-12-05
write_sensu plugin: fix two resource leaks on error

rubenk:
CID #37999


octo:
Cherry-picked into *collectd-5.5* as a850c51972ef63605de7b3dc9a95cf2a766c7621

PR #1424, opened 2015-12-05
varnish plugin: fix two resource leaks on error

rubenk:
CID #42774


octo:
Cherry-picked into *collectd-5.4* as ce65e9a3072aff269b2caf29e7f1450a10882aba

PR #1428, opened 2015-12-05
curl_xml plugin: plug a few leaks on error

rubenk:
CID #37993


octo:
Cherry-picked into *collectd-5.4* as a2c425d3a1669c6baf06ba27ce2008122fb0ed16

PR #1430, opened 2015-12-05
zfs-arc plugin: check return value of malloc

rubenk:
CID #37989


octo:
Cherry-picked into *collectd-5.5* as 438651f58b05cbbc5d5ddf3d465da62e0d8bde40

PR #1434, opened 2015-12-05
iptables plugin: remove dead code

rubenk:
ip version is either IPv4 or IPv6

CID #37962

### DIFF
--- a/src/iptables.c
+++ b/src/iptables.c
@@ -109,132 +109,127 @@ static int iptables_config (const char *key, const char *value)
 		ip_version = IPV4;
 	else if (strcasecmp (key, "Chain6") == 0)
 		ip_version = IPV6;
+	else
+		return (1);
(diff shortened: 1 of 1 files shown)


octo:
Can you please remove the

```c
if (( ip_version == IPV4 ) || ( ip_version == IPV6 ))
```

condition and out-dent the large block?

Also, it looks as if `ip_version` might be left at zero, in which case I'd argue to add something like:

```c
if (( ip_version != IPV4 ) && ( ip_version != IPV6 ))
  return -1;
```

rubenk:
Is this what you had in mind?

octo:
Exactly, thank you!

octo:
Cherry-picked into *collectd-5.4* as 7462900ca1749634a7fa9325c2e71f49e13d0095

PR #1436, opened 2015-12-05
configure.ac: fix libjvm detection on OSX

rubenk:
Shared libraries have the extension .dylib instead of .so

Ideally we look at shrext_cmds from libtool to find the shared library
extension but this works too.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2227,7 +2227,7 @@ then
 		fi
 
 		AC_MSG_CHECKING([for libjvm.so])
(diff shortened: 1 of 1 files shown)


octo:
Cherry-picked into *collectd-5.4* as 4d04d3b7e32b8777f7c7b04110b74194b991fbea

rubenk:
This breaks the java plugin. I clearly don't know how find works, sorry. Fix coming up.

octo:
Meh, it looked so reasonable ;)

rubenk:
This should be better :)

octo:
Cherry picked as b1cad5912a658356bdcb49aa208e15307b3a61a0

PR #1441, opened 2015-12-06
.gitignore: ignore test_plugin_ceph

chicha:
I did not noticed it in my previous #1403 since i do not always compile the whole collectd source tree, but it sounds like ``test_plugin_ceph``should be added to .gitignore too.

Cheers,
CH


rubenk:
Thanks @chicha!

PR #1450, opened 2015-12-10
swap plugin: don't exit when malloc fails

rubenk:


octo:
Cherry-picked into *collectd-5.4* as 7368bf2a94337b18b67a0c3573ddb2d71534ae8d

PR #1459, opened 2015-12-11
Remove check for ip_compat.h

rubenk:
Solaris needed this at one point but since
a621ced we check for ip6_ext without including this
header and that seems to succeed to.


rubenk:
I'm not entirely sure about this since I don't have a Solaris box to test, but as far as I can tell this shouldn't  be a problem. If it is, we need to extend our autoconf check for ip6_ext.

PR #1461, opened 2015-12-11
java: link with libpthread on FreeBSD

rubenk:


mfournier:
Great, this seems to allow the java plugin to build on freebsd :clap: :clap:, thanks !

That said, according to https://ci.collectd.org/job/pull-requests-build-on-freebsd10-amd64-with-default-toolchain/439//console a whole lot of plugins (unrelated to java) didn't detect their support libs: libyajl, libdbi, protobuf, etc.

As it seems related to https://github.com/collectd/collectd-ci/commit/b6b5cb0e I'm going to revert it. I was under the impression one of your recent changes to the build system made this work out of the box, but I can't find which one now. Any clue ?

rubenk:
Great! I'll need to do some more work on the libpthread detection stuff thougt, since technically we can just compile with cc -pthread, not -lpthread on most platforms.

As for detecting support libs on FreeBSD, that change works, you can leave out the --with-$lib=/usr/local if you do ./configure CPPFLAGS="-I/usr/local/include" LDFLAGS="-L/usr/local/lib". At least that's what I do and I believe the ports makefiles do that as well.

PR #1464, opened 2015-12-12
configure.ac: warn when pkg-config is not installed

rubenk:
It hasn't been optional for a while now.

Fixes:
checking for rrd.h... no
./configure: 25157: Syntax error: word unexpected (expecting ")")


octo:
Cherry-picked into *collectd-5.4* as b1582672709204fd2342a50e1d25467d99568199

PR #1465, opened 2015-12-12
configure.ac: detect bison

rubenk:
We use some bison-specific extensions so we really need bison, not yacc.
This has bitten a few people before, so make it easier for them.

Bison is only needed when compiling from git, since we ship the generated file
in the tarball. We take this into account by checking for the generated file first.


rubenk:
Closing this. Will open a new PR against 5.4

PR #1467, opened 2015-12-13
Multiple plugins: Add support for NetBSD

he32:
Fix to not constantly close & re-open /dev/urandom -- doing so will depelete
the kernel's entropy estimate because a separate PRNG is seeded on each
/dev/urandom open.


rubenk:
Thanks for your contribution @he32!. However there are now 2 pull requests open, what should happen to #1250?

Since this is a fairly large change, and we don't have many NetBSD hackers around to review it, it would help if you:
- rebased your changes to current master
- split them out in one commit per plugin for easier review.

Would that work for you?

he32:
Sigh, well, OK; I'll try to do that.
It's a bit of a pain since a number of the plugins need to touch configure.ac.
But I'll give it a try.  It'll take a while, though.

octo:
Thanks for all that work, @he32! I'd love to merge this, but please break it up into one patch per plugin so we have a chance to properly review the changes! Ideally open a separate pull request for each plugin. You can do this incrementally, i.e. plugin by plugin – there's no need to everything at once :)

Thanks and best regards,
—octo

he32:
I'm a little unfamiliar with how to structure these pull requests, primarly I guess because I'm a git newbie.  The primary problem I have is that there's more than one plugin which depends on one and the same change to configure.ac and config.h.in.  Where then do I put the change to configure.ac and config.h.in?  Certainly I can't put the same change into multiple pull requests? Won't you then get conflicts when you merge two pull requests which touch the same file with the same change? But also I can't first do one pull request based on upstream/master and have the other pull requests which are based on the same upstream/master depend on this other pull request -- I don't know if git has a method to say "this pull request depends on that pull request over there".

he32:
Sigh, it looks like I did the silly mistake of committing my initial changes on my own master branch (if I'm interpreting git correctly, which in itself is an uncertain), and I don't know git well enough to know how to get myself out of that position.

For what it's worth, I've already committed these changes in NetBSD's pkgsrc system, so they're not entirely untested.

rubenk:
I've now merged the prerequisite autoconf changes into master. I'll close this PR, please open separate PR's per plugin.

PR #1479, opened 2015-12-19
netlink plugin: metrics for drops, requeues, overlimits, qlen, backlog

ti-mo:
This feature was implemented by @mcq8 as part of a TC monitoring setup.

We're not sure if this is a useful feature or whether or not it needs to be hidden behind VerboseInterface, let us know what you think! (also, what about HAVE_TCA_STATS / tc_stats?)

### DIFF
--- a/src/netlink.c
+++ b/src/netlink.c
@@ -536,6 +536,10 @@ static int qos_filter_cb(const struct nlmsghdr *nlh, void *args) {
       }
       if (q_stats.qs != NULL) {
         submit_one(dev, "if_tx_dropped", type_instance, q_stats.qs->drops);
+        submit_one(dev, "queue_length", type_instance, q_stats.qs->qlen);
+        submit_one(dev, "ipt_backlog", type_instance, q_stats.qs->backlog);
+        submit_one(dev, "ipt_requeues", type_instance, q_stats.qs->requeues);
+        submit_one(dev, "ipt_overlimits", type_instance, q_stats.qs->overlimits);
       }
     }
 
@@ -563,8 +567,10 @@ static int qos_filter_cb(const struct nlmsghdr *nlh, void *args) {
       ssnprintf(type_instance, sizeof(type_instance), "%s-%s", tc_type,
                 tc_inst);
 
+      submit_one(dev, "if_tx_dropped", type_instance, ts->drops);
       submit_one(dev, "ipt_bytes", type_instance, ts->bytes);
       submit_one(dev, "ipt_packets", type_instance, ts->packets);
+      submit_one(dev, "ipt_overlimits", type_instance, ts->overlimits);
     }
 
     break;
--- a/src/types.db
+++ b/src/types.db
@@ -118,6 +118,9 @@ io_packets              rx:DERIVE:0:U, tx:DERIVE:0:U
 ipc                     value:GAUGE:0:U
 ipt_bytes               value:DERIVE:0:U
 ipt_packets             value:DERIVE:0:U
+ipt_backlog             value:GAUGE:0:U
+ipt_requeues            value:DERIVE:0:U
+ipt_overlimits          value:DERIVE:0:U
 irq                     value:DERIVE:0:U
 latency                 value:GAUGE:0:U
 links                   value:GAUGE:0:U


src/netlink.c
```diff
@@ -387,20 +392,31 @@ static int link_filter_cb(const struct nlmsghdr *nlh,
...
 
   if (mnl_attr_get_type(attr) == TCA_STATS_BASIC) {
-    if (mnl_attr_validate2(attr, MNL_TYPE_UNSPEC, sizeof(**bs)) < 0) {
+    if (mnl_attr_validate2(attr, MNL_TYPE_UNSPEC, sizeof(*(stats->bstats))) < 0) {
+      ERROR("netlink plugin: qos_attr_cb: TCA_STATS_BASIC mnl_attr_validate2 "
+            "failed.");
+      return MNL_CB_ERROR;
+    }
+    stats->bstats = mnl_attr_get_payload(attr);
+    if (stats->qstats != NULL)
```
octo: This is a bit hard to read, since "bstats" and "qstats" look so similar. How about doing this instead:

```c
if (mnl_attr_get_type(attr) == TCA_STATS_BASIC) {
  // …
  stats->bstats = mnl_attr_get_payload(attr);
} else if (mnl_attr_get_type(attr) == TCA_STATS_QUEUE) {
  // …
  stats->qstats = mnl_attr_get_payload(attr);
}

if ((stats->bstats != NULL) && (stats->qstats != NULL)) {
  return MNL_CB_STOP;
}
return MNL_CB_OK;
```

src/netlink.c
```diff
@@ -492,7 +508,9 @@ static int qos_filter_cb(const struct nlmsghdr *nlh, void *args) {
 
 #if HAVE_TCA_STATS2
   mnl_attr_for_each(attr, nlh, sizeof(*tm)) {
-    struct gnet_stats_basic *bs = NULL;
+    struct gnet_stats stats;
```
octo: Please use the `{0}` initializer to initialize the struct to nil:

```c
struct gnet_stats stats = {0};
```

src/types.db
```diff
@@ -118,6 +118,11 @@ io_packets              rx:DERIVE:0:U, tx:DERIVE:0:U
 ipc                     value:GAUGE:0:U
 ipt_bytes               value:DERIVE:0:U
 ipt_packets             value:DERIVE:0:U
+ipt_qlen                value:GAUGE:0:U
```
octo: Please use the existing *queue_length* type.

src/types.db
```diff
@@ -118,6 +118,11 @@ io_packets              rx:DERIVE:0:U, tx:DERIVE:0:U
...
+ipt_backlog             value:GAUGE:0:U
+ipt_drops               value:DERIVE:0:U
```
octo: This should use *if_tx_dropped*, as is done in #2053.

src/types.db
```diff
@@ -118,6 +118,11 @@ io_packets              rx:DERIVE:0:U, tx:DERIVE:0:U
...
+ipt_requeues            value:DERIVE:0:U
```
octo: Please use the existing *if_tx_packets* type and set the *type instance* to "requeued".

src/netlink.c
```diff
@@ -503,18 +521,26 @@ static int qos_filter_cb(const struct nlmsghdr *nlh, void *args) {
...
-
-      submit_one(dev, "ipt_bytes", type_instance, bs->bytes);
-      submit_one(dev, "ipt_packets", type_instance, bs->packets);
+      if (stats.bstats != NULL) {
+        submit_one(dev, "ipt_bytes", type_instance, stats.bstats->bytes);
+        submit_one(dev, "ipt_packets", type_instance, stats.bstats->packets);
+      }
+      if (stats.qstats != NULL) {
+        submit_one(dev, "ipt_qlen", type_instance, stats.qstats->qlen);
+        submit_one(dev, "ipt_backlog", type_instance, stats.qstats->backlog);
```
octo: Do I understand this correctly that "qlen" is the length of the queue and "backlog" the number of packets in the queue? If so, I think it'd make most sense to do *avail = qlen - backlog* and then submit *avail* and *backlog* as *queue_length-free* and *queue_length-used*.

src/netlink.c
```diff
@@ -543,6 +569,8 @@ static int qos_filter_cb(const struct nlmsghdr *nlh, void *args) {
 
       submit_one(dev, "ipt_bytes", type_instance, ts->bytes);
       submit_one(dev, "ipt_packets", type_instance, ts->packets);
+      submit_one(dev, "ipt_drops", type_instance, ts->drops);
+      submit_one(dev, "ipt_overlimits", type_instance, ts->overlimits);
```
octo: What's the difference between "drops" and "overlimits"? One is due to the network the other due to policy?

src/types.db
ti-mo: @octo Any pointers on how to do this?

src/netlink.c
ti-mo: @octo As far as I understand, the significance of qlen and backlog differs between queueing algorithms. _qlen_ is not synonymous with maximum buffer length, in some types of qdiscs, it is the length of the standing queue (packets), it is a gauge, not a delta. Reading the source of the queue implementation is required to fully understand its details.

We use primarily HFSC, and as far as I can tell, _qlen_ seems to be used as an internal metric only, while _backlog_ seems to be a value that is exported to statistics only. (even though both seem to be readable through netlink)

src/netlink.c
ti-mo: @octo Same as above, I assume not every qdisc has the concept of overlimits, in HFSC this _should be_ the amount of time/packets it steals from sibling classes. Drops are packets that are gone, but drops might only occur in leaf nodes in some setups.

ti-mo:
@octo #2053 rendered nearly all of this obsolete, it was almost an exact copy of this PR. Not sure why it was put on the backburner for so long for another patch to be re-invented and merged before this one. I've trimmed all conflicting code and responded to some of your in-line comments.

Cheers

PR #1480, opened 2015-12-19
Provide the configuration tests for NetBSD.

he32:
Absent guidance to the contrary, getting this applied is a
precondition for the individual plugin updates I'll subsequently
submit pull requests for.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -99,6 +99,10 @@ case $host_os in
 	AC_DEFINE([KERNEL_OPENBSD], 1, [True if program is to be compiled for an OpenBSD kernel])
 	ac_system="OpenBSD"
 	;;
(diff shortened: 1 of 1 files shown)


rubenk:
Thanks @he32. Is netstat_udp going to be a new plugin? If so my preference to do a separate PR for that later, including its configure.ac changes. I have some questions if it would be useful for other platforms as well, and if it could be extended for other protocols, in which case 'netstat' would be a better name.

he32:
OK...  I did this primarily to monitor UDP socket buffer overflows.

rubenk:
Thanks! I just squashed those 3 commits together and merged them in to master as f0446c10f23af1fd088535ac6cfe119c8f3c0efe

PR #1483, opened 2015-12-22
Extend collected metrics in redis plugin

falzm:
This change adds support for other interesting Redis metrics:

* `expired_keys`
* `keyspace_hits`, `keyspace_misses`
* `total_net_input_bytes`, `total_net_output_bytes`

It also fixes the DS type for `expired_keys`, which was incorrectly expressed as gauge instead of derive.


haomingz:
+1

falzm:
Hi, any news on this?

mfournier:
Thanks a lot @falzm ! All good as far as I can tell.

I split your patch and merged it as a6ba92e and d3f8bb5.

falzm:
Thanks!

PR #1486, opened 2015-12-23
fixed seg fault due to null datadir to strlen

vmahedia:


bbczeuz:
I made a similar patch #1500 fixing the same problem and another one on the same lines. I think your code looses memory if datadir was not non-NULL before line 1020

mfournier:
Closing in favour of #1500. Thanks @vmahedia !

PR #1489, opened 2015-12-25
openldap plugin: fix build on OSX El Capitan

rubenk:
OpenLDAP is deprecated in OSX El Capitan.
Apple would like us to use the OpenDirectory Framework instead.

Treat these deprecation warnings as warnings, not errors.


mfournier:
Thanks a lot @rubenk !

Merged as 44acfc8.

Issue #1501, opened 2016-01-06
netapp plugin doesn't compile anymore with recent gcc and -Werror


llaurent:
```
/bin/bash ../libtool  --tag=CC   --mode=compile x86_64-linux-gnu-gcc -DHAVE_CONFIG_H -I.  -I./daemon -DPREFIX='"/usr"' -DCONFIGFILE='"/etc/collectd.conf"' -DLOCALSTATEDIR='"/var"' -DPKGLOCALSTATEDIR='"/var/lib/collectd"' -DPIDFILE='"/var/run/collectd.pid"' -DPLUGINDIR='"/usr/lib/collectd"' -DPKGDATADIR='"/usr/share/collectd"'  -D_FORTIFY_SOURCE=2 -I/usr/src/collectd-5.5.0/debian/include -DLT_LAZY_OR_NOW='RTLD_LAZY|RTLD_GLOBAL' -UCONFIGFILE -DCONFIGFILE='"/etc/collectd/collectd.conf"' -Wall -Werror -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wall -Wno-error=deprecated-declarations -MT netapp_la-netapp.lo -MD -MP -MF .deps/netapp_la-netapp.Tpo -c -o netapp_la-netapp.lo `test -f 'netapp.c' || echo './'`netapp.c
[...]

PR #1502, opened 2016-01-06
remove the addressof in the sizeof

llaurent:
quick fix for issue #1501 
the new warning for pointer access introduced in gcc 4.8 helped to find this one.


mfournier:
Thanks a lot @llaurent !

Merged as e4908c8.

We don't hear about this plugin very often... Let us know if you encounter any other issues !

PR #1506, opened 2016-01-12
apache: warn about possible misconfiguration

### LINKED ISSUE #1170: apache plugin does not log errors on invalid MIME type/ content-type.
# Synopsis
When the Apache plugin uses libcurl to retrieve a copy of the server status page, it doesn't log any errors if the page format or content-type is invalid.

### Example
If a user incorrectly configures the 'URL' (e.g. without '?auto' at the end of the URL) within the plugin configuration, Apache doesn't return a 'Machine Readable Status File'. As a result, instead of returning a document with content type 'text/plain', it returns 'text/html'. I believe that collectd should log an error in this instance, yet it does not (even with LogLevel of 'debug').

### Example _broken_ configuration:
* File "/etc/collectd.d/apache.conf":
```        
LoadPlugin "apache"
<Plugin "apache">
    <Instance "httpd">
        URL "http://localhost/server-status/"
    </Instance>
</Plugin>
```

### Example _working_ configuration:
* File "/etc/collectd.d/apache.conf":
```        
LoadPlugin "apache"
<Plugin "apache">
    <Instance "httpd">
        URL "http://localhost/server-status/?auto"
    </Instance>
</Plugin>
```
### Details
* I have tried the broken configuration above in both collectd 5.4.2 and 5.5.0.
* Both builds of collectd were compiled with '--enable-debug', yet no debug information was produced in either case.
* When the plugin receives a status page with an encoding of 'text/html' (and html rendering), no RRDs are generated. Appending '?auto' to the end of the URL fixes the problem, however, it was not trivial to debug.
* I have updated the collectd wiki page for the Apache plugin to make note of the fact that a 'Machine Readable Status File' is required for this plugin to work.

mfournier:
When the server doesn't return `Content-Type: text-plain`, it's a strong
hint that we're not querying the right URL.

Fixes #1170

### DIFF
--- a/src/apache.c
+++ b/src/apache.c
@@ -596,13 +596,16 @@ static int apache_read_host (user_data_t *user_data) /* {{{ */
 
 	st = user_data->data;
 
+	int status;
+
+	char *content_type;
+	static const char *text_plain = "text/plain";
+
 	assert (st->url != NULL);
 	/* (Assured by `config_add') */
 
 	if (st->curl == NULL)
 	{
-		int status;
-
 		status = init_host (st);
 		if (status != 0)
 			return (-1);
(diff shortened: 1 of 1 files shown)


rubenk:
Thanks @mfournier!

PR #1508, opened 2016-01-12
processes: warn about process names above OS limit

### LINKED ISSUE #1284: The processes plugin should warn that the length of the process title is limited.
On linux, at least, the process title can be up to 15 characters (plus a terminating `\0`). I spent some time confused as to why my processes weren't being detected by the plugin.

Or, better, would be if the plugin would correctly detect longer-named processes (based on the basename of the program in the command-line maybe?)

mfournier:
Fixes #1284

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3751,6 +3751,18 @@ Default: C<Collectd notify: %s@%s>
 
 =head2 Plugin C<ntpd>
 
+The C<ntpd> plugin collects per-peer ntpd data such as time offset and time
+dispersion.
+
+For talking to B<ntpd>, it mimics what the B<ntpdc> control program does on
+wire - using B<mode 7> specific requests. This mode is deprecated with
+newer B<ntpd> releases (4.2.7p230 and later). For the C<ntpd> plugin to work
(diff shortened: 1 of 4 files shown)


mfournier:
This emits this warning once at startup time:

```
[2016-01-12 22:25:10] [warning] processes plugin: this platform has a 15 character limit to process names. The `Process "a very very long process name"' option will not work as expected.
```

mfournier:
@dago, @brd would you by any chance know if Solaris & FreeBSD also have a limit to process names length ? Thanks !

dago:
I just verified that the limit on Solaris is also 15 characters. This can be reviewed in `/usr/include/sys/user.h`:
`
#define MAXCOMLEN       16      /* <= MAXNAMLEN, >= sizeof (ac_comm) */
`
and I also tested with my super long command:
`
Process name: 023456789112345
`

So it would be good to add at least `KERNEL_SOLARIS` to the condition.

mfournier:
@dago I just added a patch to the serie enabling the warning on Solaris too. Could you please take a look ? Thanks !

mfournier:
Closing this messed up PR which I really should have submitted against the collectd-5.4 branch.

PR #1511, opened 2016-01-18
df: remove legacy code skipping "rootfs" monitoring

### LINKED ISSUE #1402: df plugin: Devices mounted more than once with "ReportByDevice true" sends duplicate data
When a device is mounted more than once, it's sent multiple times but all of the keys are the same so you can't really distinguish between them.  So it's probably worth de-duping before sending.

OpenSuSE has had a patch for this for a whlie (https://build.opensuse.org/package/view_file/openSUSE:Factory/collectd/collectd-df-remove-duplicates.patch?rev=4)

mfournier:
3512bb1 added code to skip duplicate reporting of `rootfs` mounted on `/`.
f9c1c5b and f0398d0 added generic code to skip any volume mounted twice.

Depending on the order of the entries in `/etc/mtab`, reporting for `/` was
entirely skipped.

This patch basically reverts the first, non-generic patch, as it's
superseded by the 2 others.

Fixes #1402

### DIFF
--- a/src/df.c
+++ b/src/df.c
@@ -271,11 +271,7 @@ static int df_read (void)
 		else
 		{
 			if (strcmp (mnt_ptr->dir, "/") == 0)
-			{
-				if (strcmp (mnt_ptr->type, "rootfs") == 0)
-					continue;
 				sstrncpy (disk_name, "root", sizeof (disk_name));
-			}
 			else
 			{
 				int i, len;


octo:
Cherry-picked into *collectd-5.4* as 573db6cb7d8904468546746117efbd989d4a5843

PR #1516, opened 2016-01-21
docs: mention regexps for the interface plugin

shastah:
It is very useful to collect data (or ignore) from a group of interfaces that are similarly named - especially if the list is dynamic and it's not feasible to update configuration each time an interface shows up or disappears. It's possible to use regexps to specify interfaces to collect data from (or ignore), but it's not clearly stated anywhere in the documentation.

This is an attempt to improve documentation in this area.


mfournier:
Heh, I wasn't even aware of that, shame on me :-/

Thanks a lot @shastah, very much appreciated !

I merged your patch as 01bb122.

PR #1517, opened 2016-01-21
systemd.collectd.service: improve systemd & capabilities explanations

mfournier:
Any comments @rubenk ? Thanks !


rubenk:
Looks good, do it! :)

mfournier:
@rubenk also FYI, I started working on a bit of code which checks the capabilities and tries to emit a useful warning for plugins which won't work due to dropped caps or not running as root.

rubenk:
@mfournier that's great! And it would be even better if collectd could drop its capabilities itself instead of depending on systemd for this.

Issue #932, opened 2015-02-08
ntpd plugin: private mode (7) is now deprecated


shastah:
After upgrading my ntpd from 4.2.6p5 to 4.2.8, ntpd plugin stopped collecting any data:

    Feb  7 16:58:44 hostname collectd[11344]: ntpd plugin: ntpd_do_query returned unexpected data. (ik = (nil); ik_num = 0; ik_size = 0)
    Feb  7 16:58:44 hostname collectd[11344]: read-function of plugin `ntpd' failed. Will suspend it for 20.000 seconds.

Turns out ntpdc (which collectd mimics on the wire) is now deprecated, and "mode7" must be explicitly enabled in ntp.conf to keep it working:

      mode7  Enables  processing  of NTP mode 7 implementation-specific requests
             which are used by the deprecated ntpdc(1) program. The default for this flag
             is disable.  This flag is excluded from runtime configuration using ntpq(1).
             The ntpq(1) program provides the same capabilities as ntpdc(1) using standard 
             mode 6 requests.

It would probably be the best to rework collectd ntpd plugin to talk to ntpd using mode 6, like ntpq, but if that proves tricky and/or time consuming, at least it's worth mentioning the issue in ntpd plugin documentation, I think.

jettero:
It took me quite a while to figure out how to enable this and what the implications might be. I'm still not totally sure about the implications, but I wanted to post a bunch of URLs here for the next chump in danger of spending an hour trying to figure out what's going on.

From my ntp.conf —

```
# see http://doc.ntp.org/4.2.8p5/miscopt.html
# and see http://questions.ntp.narkive.com/ebjq59Qq/ntp-questions-ntpdc-is-not-working-in-ntp-4-2-8p1
# and also http://support.ntp.org/bin/view/Dev/DeprecatingNtpdate
# and lastly: https://github.com/collectd/collectd/issues/932
# (we need this depricated bit for collectd ntpd stats)
enable mode7
```

matwey:
@shastah what is the difference between mode 7 and mode 6? How can collectd be reimplement to mimic ntpq behaviour?

shastah:
@matwey mode 6 is described in RFC 1305 as "reserved for NTP control messages", whereas mode 7 is "reserved for private use", so I guess might be implementation-specific. The most striking difference between them in terms of parsing responses is that mode 6 is supposed to be human-readable, so it's mostly plaintext.

[...]

PR #1518, opened 2016-01-21
docs: Add info about mode7 requirement for ntpd

shastah:
Since ntpd-4.2.7p230 "mode 7" requests are ignored by default, and that's what ntpd plugin uses currently. The ntp daemon must be explicitly configured to enable mode7 requests. This patch adds short information about that to collectd.conf manpage.

Related to #932

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -4483,6 +4483,18 @@ Sets the I<command file> to write to. Defaults to F</usr/local/nagios/var/rw/nag
 
 =head2 Plugin C<ntpd>
 
+The C<ntpd> plugin collects per-peer ntpd data such as time offset and time
+dispersion.
+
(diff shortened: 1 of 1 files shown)


mfournier:
Thanks again @shastah !

I merged your patch as e68993d (I corrected a couple of typos on the fly).

Let's leave #932 open, as it would be better to update the plugin to match whatever mode the current ntp releases use by default.

jcberthon:
I know this PR is closed but the added info does not tell how to enable mode7, and sadly on Ubuntu 16.04, which ships ntp 4.2.8, the man page does not mention how to set it up.

I understand that this could be seen as a bug for the ntp project. But as the information is not in this ntp.conf man page, the information in the collectd.conf man page is not helping.

In addition, shouldn't the [wiki](https://collectd.org/wiki/index.php/Plugin:NTPd) be updated?

PR #1519, opened 2016-01-22
Don't divide by 0 when doing percentages (e.g. with the swap plugin).

coreykosak:
Prior to this change, if all the values being sent sum to zero, then the
calculated percentage is a division by 0 which is either Inf or NaN
depending on the numerator.

This is not a theoretical concern: it can easily happen if all the input
gauges are 0.

This change forces the output to be zero if the denominator is zero.


mfournier:
Thanks a lot for tracking this down @coreykosak ! Merged as f1552fd, so that it gets part of the next 5.5.x bugfix release.

PR #1520, opened 2016-01-22
Protect reference to field 're' with an #if guard.

coreykosak:
The field 're' only exists if HAVE_REGEX_H is defined (see definition at line 194). So, consistent with all other usages, this should be protected with an #if guard.


mfournier:
Nice catch, thanks ! Cherry-picked as 5c19a9f.

PR #1521, opened 2016-01-22
Factor out some common functionality

coreykosak:
I (personally) found this code to be repetitive and too long for what it does. We can improve readability
(IMO) by factoring it out into a subroutine. There is no change in semantics.


mfournier:
Great, many thanks @coreykosak !

PR #1522, opened 2016-01-23
Provide a unified signature for ps_get_cmdline.

coreykosak:
Prior to this change, ps_get_cmdline had different signatures for
KERNEL_LINUX vs. KERNEL_SOLARIS. This means that callers who want to
call this function would have to have an #if..#else that controlled which
variant of the function to call.

By giving them the same signature, callers don't have to worry about
that.

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -1326,8 +1326,8 @@ static int read_fork_rate ()
 #endif /*KERNEL_LINUX */
 
 #if KERNEL_SOLARIS
-static const char *ps_get_cmdline (long pid, /* {{{ */
-		char *buffer, size_t buffer_size)
+static char *ps_get_cmdline (pid_t pid, char *name __attribute__((unused)), /* {{{ */
+    char *buffer, size_t buffer_size)
 {
 	char path[PATH_MAX];
 	psinfo_t info;
@@ -2440,7 +2440,7 @@ static int ps_read (void)
 
 
 		ps_list_add (ps.name,
-				ps_get_cmdline (pid, cmdline, sizeof (cmdline)),
+				ps_get_cmdline (pid, ps.name, cmdline, sizeof (cmdline)),
 				&pse);
 	} /* while(readdir) */
 	closedir (proc);


mfournier:
Thanks again @coreykosak ! Merged as bb978c1.

mfournier:
@coreykosak, reopening as this breaks the build on solaris. See https://github.com/collectd/collectd/commit/00f9279d9711c9882b65d54fd72c2f25a3f403fc#commitcomment-15999309 and https://buildfarm.opencsw.org/buildbot/builders/collectd-solaris10-sparcv9/builds/135/steps/shell_2/logs/stdio

Thanks for the notification @dago !

Do you think we can safely change this format string to `%i` ? It's confined inside a KERNEL_SOLARIS block, so it should make the compiler happy. But would this still work on 32bit solaris ?

And what's the max process count on this platform ? Aren't we at risk of hitting this limit by downgrading to an int ?

dago:
In 32 bit Solaris the max PID is `pid32_t` which is essentially `int32_t`, so it should be safe.

mfournier:
@dago ok, but what if the system runs more processes than what an `int` can hold (ie: some PIDs larger than the size of an `int`) ? On the other hand, what's the point of the `pid_t` type if it's not exactly this sort of use case ?

I just want to be sure we're not introducing a bug here...

dago:
According to the [Oracle Solaris Tunable Parameters Reference Manual] (https://docs.oracle.com/cd/E26505_01/html/E37386/chapter2-4.html) `pidmax` is the maximum allowed PID which is 999.999 for Solaris 10. In the kernel the value can be determined with
`
echo pidmax/D | mdb -k
`
where `D` is defined in `mdb(1)` as
`
     D         decimal signed int (4 bytes)
`
At my system this defaults to 30.000.

coreykosak:
Yikes. So sorry about that. Some types have special format specifier macros, but I'm not aware of any such macro for pid_t (though we could define one I guess).

Rather than the suggestion to downgrade to an int, what if we decided to go in the opposite direction... coerce to a long or even long long? e.g.

```
snprintf(path, sizeof (path), "/proc/%lli/psinfo", (long long)pid);
```

Or maybe even "unsigned long long".

mfournier:
Thanks for the advices @coreykosak !

I digged a bit further into this and finally came up with #1550.

It appears the maximum number of processes supported on Linux and Solaris is way below the max size of a `long`, so IMHO nothing justifies using a `long long`. `long` already potentially implies scanning a `/proc` dir with 2M entries, so I guess this will be the first bottleneck ;)

Regarding signedness, I never heared of negative PIDs, but as `pid_t` doesn't seem to be unsigned, in doubt, I left it signed.

PR #1524, opened 2016-01-25
Adding 'used' memory to submitted AIX metrics

jquinn47:
This change that lets us track the memory in use on AIX boxes. The value was already available in the perfstat memory total structure, I've just added the it to the list of submitted metrics, keeping naming consistent with other platforms.


mfournier:
Same question here: would this break or fix the fact all reported metrics sum up to 100% of available memory ?

In any case, thanks for all your work @jquinn47 ! Nice to have someone give the AIX bits some love :)

jquinn47:
In this case, the sum of the metrics will not add to 100%.

The perfstat documentation is pretty sparse, but from looking at metrics we've been gathering for the past few months:

```

PR #1525, opened 2016-01-25
Track process syscalls for AIX, solaris, and linux. Add Fork rate to AIX

jquinn47:
This change lets us track the number of syscalls for AIX, solaris, and linux using perfstat, kstat, and proc/stat respectively.

In addition, I've added 'fork_rate' to AIX, which is already being tracked by other platforms.


jquinn47:
I intended to merge this onto the collectd-5.5 branch

PR #1528, opened 2016-01-25
Report* Flags added to DF configuration, to disable undesired metrics

jquinn47:
I've added three new configuration flags to the DF plugin, which allows users to disable undesired metrics if they want to. Default behavior is to send all metrics, for backwards compatibility.

### DIFF
--- a/src/df.c
+++ b/src/df.c
@@ -350,30 +350,30 @@ static int df_read (void)
 		if (values_absolute)
 		{
 			if (report_free) 
-                df_submit_one (disk_name, "df_complex", "free",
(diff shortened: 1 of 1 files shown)


mfournier:
The `ReportReserved` option existed at the time of 4.x, but has been removed. The idea is that all the metrics reported sum up to 100% of the available capacity.

Also, the filter chains mechanism give the user a generic way of renaming/dropping metrics before they get submitted to whatever storage backend is used.

jquinn47:
Thank you for the update. We'll look into using filter chains moving forward.

PR #1530, opened 2016-01-26
Add capabilities checking to plugins needing privileges

mfournier:
A few plugins historically required collectd to be run as root. Capabilities offer a way to control more precisely the privileges given to the collectd process, depending on which plugins are loaded.

My goal to start with, is simply to help the user configure his system to run collectd with reduced privileges. Once we've got a bit more experience with this, it would be nice to have collectd drop the unneeded capabilities itself, depending on which plugins are loaded.

So this hasn't been thoroughly tested yet, I just wanted to pass it through jenkins, and get some feedback about the idea.

Also, I'm sure more plugins than just these 6 require privileges.

### DIFF
--- a/src/ceph.c
+++ b/src/ceph.c
@@ -1584,7 +1584,7 @@ static int ceph_init(void)
       WARNING ("ceph plugin: Running collectd as root, but the "
           "CAP_DAC_OVERRIDE capability is missing. The plugin's read "
           "function will probably fail. Is your init system dropping "
-          "capabilities ?");
+          "capabilities?");
     else
       WARNING ("ceph plugin: collectd doesn't have the CAP_DAC_OVERRIDE "
           "capability. If you don't want to run collectd as root, try running "
--- a/src/ceph.c
+++ b/src/ceph.c
@@ -1577,7 +1577,7 @@ static int ceph_init(void)
 {
     int ret;
 
-#ifdef HAVE_SYS_CAPABILITY_H
+#if defined(HAVE_SYS_CAPABILITY_H) && defined(CAP_DAC_OVERRIDE)
   if (check_capability (CAP_DAC_OVERRIDE) != 0)
   {
     if (getuid () == 0)
--- a/src/ceph.c
+++ b/src/ceph.c
@@ -38,6 +38,9 @@
 #if HAVE_YAJL_YAJL_VERSION_H
 #include <yajl/yajl_version.h>
 #endif
+#ifdef HAVE_SYS_CAPABILITY_H
+# include <sys/capability.h>
+#endif
 
 #include <limits.h>
 #include <poll.h>
@@ -1573,6 +1576,22 @@ static int ceph_read(void)
 static int ceph_init(void)
 {
     int ret;
+
+#ifdef HAVE_SYS_CAPABILITY_H
+  if (check_capability (CAP_DAC_OVERRIDE) != 0)
+  {
+    if (getuid () == 0)
+      WARNING ("ceph plugin: Running collectd as root, but the "
+          "CAP_DAC_OVERRIDE capability is missing. The plugin's read "
+          "function will probably fail. Is your init system dropping "
+          "capabilities ?");
+    else
+      WARNING ("ceph plugin: collectd doesn't have the CAP_DAC_OVERRIDE "
+          "capability. If you don't want to run collectd as root, try running "
+          "\"setcap cap_dac_override=ep\" on the collectd binary.");
+  }
+#endif
+
     ceph_daemons_print();
 
     ret = cconn_main_loop(ASOK_REQ_VERSION);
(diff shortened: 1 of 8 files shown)


rubenk:
This looks great @mfournier, nice work!

mfournier:
Just rebased this on current master, to make it merge again.

octo:
Thank @mfournier, that looks like an excellent addition. I'll leave some review comments inline.

src/ceph.c
```diff
@@ -1573,6 +1576,22 @@ static int ceph_read(void)
...
     int ret;
+
+#if defined(HAVE_SYS_CAPABILITY_H) && defined(CAP_DAC_OVERRIDE)
+  if (check_capability (CAP_DAC_OVERRIDE) != 0)
+  {
+    if (getuid () == 0)
+      WARNING ("ceph plugin: Running collectd as root, but the "
+          "CAP_DAC_OVERRIDE capability is missing. The plugin's read "
+          "function will probably fail. Is your init system dropping "
+          "capabilities ?");
```
octo: Great warning message, it tells the user exactly what's going on. One minor nit: please remove the space before the question mark.

src/turbostat.c
```diff
@@ -1474,35 +1474,22 @@ turbostat_read(void)
 static int
 check_permissions(void)
 {
-#ifdef HAVE_SYS_CAPABILITY_H
-	struct __user_cap_header_struct cap_header_data;
-	cap_user_header_t cap_header = &cap_header_data;
-	struct __user_cap_data_struct cap_data_data;
-	cap_user_data_t cap_data = &cap_data_data;
+
 	int ret = 0;
```
octo: `ret` is declared and not used if `HAVE_SYS_CAPABILITY_H` or `CAP_SYS_RAWIO` is undefined.

src/daemon/common.c
```diff
@@ -1668,3 +1672,32 @@ void strarray_free (char **array, size_t array_len) /* {{{ */
 		sfree (array[i]);
 	sfree (array);
 } /* }}} void strarray_free */
+
+#ifdef HAVE_SYS_CAPABILITY_H
+int check_capability (int capability) /* {{{ */
+{
+#ifdef _LINUX_CAPABILITY_VERSION_3
+	struct __user_cap_header_struct cap_header_data;
```
octo: I'm feeling very uncomfortable using private system structures. I think we have two options:

1.  Allocate memory on the heap, using `calloc(sizeof (*cap_header), 1)`. Unfortunately, I'm not aware of a stack-only way of initializing pointer types :\
2.  Use `cap_get_proc()` and `cap_get_flag()` (and `cap_free()`) instead of the Linux-only interface. This seems to be the nicer interface to me.

mfournier:
> I'm feeling very uncomfortable using private system structures

Sorry, I didn't really pay attention to this. I basically copied these lines from `src/turbostat.c`, which is the first plugin checking these capabilities (all the credits  to @Feandil for this idea, thanks again!).

I went for the `calloc()` variant despite your advice, @octo: while working on this patchset, I took a look at the different capabiliies implementations, and initially aimed for the POSIX-defined functions. But unfortunately this is a "withdrawn POSIX.1e draft" (see `capabilities(7)` and `capget(2)`). Apart from Linux, only FreeBSD has "some initial work" towards this standard, according to their handbook. But it seems to have stalled somewhere around 2006. As for Linux itself, `_LINUX_CAPABILITY_VERSION_3` is the only reasonable implementation to use.

So this patchset is currently organised to run on this Linux-specific, post-2.6.26 implementation only, avoiding the POSIX.1e standard. But keeping the door open for an (unlikely) concurrent implementation (ie: the somewhat redundant `#ifdef`s on `HAVE_SYS_CAPABILITY_H` and `_LINUX_CAPABILITY_VERSION_3`).

Further than that, using these `cap_get_*()` POSIX.1e functions require linking against libcap, which adds a runtime dependency just to be able to emit a warning message.

Thanks for the idea of using `calloc()` though! It looks like it just works :-)

src/daemon/common.c
```diff
@@ -1668,3 +1672,51 @@ void strarray_free (char **array, size_t array_len) /* {{{ */
...
+	{
+		ERROR("check_capability: calloc failed");
+		return (-1);
+	}
+
+	cap_user_data_t cap_data = calloc(sizeof (*cap_data), 1);
+	if (cap_data == NULL)
+	{
+		ERROR("check_capability: calloc failed");
+		return (-1);
```
octo: this leaks cap_header

octo:
LGTM, but please see the one inline comment

src/daemon/common.c
mfournier: doh, nice catch, thanks !

octo:
LGTM, thanks Marc!

Feandil:
Sorry for bringing these troubles into collectd.
@mfournier: You are crediting me for the idea, but this is wrong ;)
I mostly ported the turbostat utility (part of the linux kernel) into collectd: the addition of capability checks, https://github.com/Feandil/collectd/commit/daebff6bb861acc0ed7390bcb101b6e9af85efff, was a "backport" of https://github.com/torvalds/linux/commit/98481e79b60a50d699b79292ff1b7e56e7fa8425

mfournier:
Ok, so this would explain why we ended up accessing these private system structures directly.

No trouble at all, all the opposite. Don't worry ;-)

PR #1540, opened 2016-02-05
write_sensu: avoid calling strlen() on variable set to NULL

mfournier:
This fixes a segfault when no `Tag` option is set.


rubenk:
lgtm.

PR #1542, opened 2016-02-07
generic-jmx: add AtomicLong and AtomicInteger

liry:
Some programs expose numbers of AtomicLong or
AtomicInteger types (e.g. Eureka from Netflix).


mfournier:
Thanks a lot @liry, especially for bootstrapping tests for this part ! Would you think it would be possible to alter the Makefiles so these tests run when the "make check" target gets called ?

liry:
@mfournier I've added tests to Makefiles. I also had to add binaries of required libraries to the repo. The other option would be to use e.g. Maven to handle dependencies, but this would require to have Maven installed to be able to build the project.

liry:
@rubenk could you please help me with adding java tests to single Makefile you prepared for this project?

I've added `bindings/java-tests/` with source code of tests and also `libs` necessary to run these tests. If I understand it correctly, all of it needs to be part of tarball created by `make -s dist-gzip`, then those tests need to be compiled and executed by `make -sk check`. But none of it should be installed to end user's computer.

I can change directory layout if there's a better solution.

PR #1546, opened 2016-02-09
processes: warn about process names above OS limit

### LINKED ISSUE #1284: The processes plugin should warn that the length of the process title is limited.

mfournier:
Fixes #1284. Rebased version of #1508.

This emits this warning once at startup time:

```
[2016-01-12 22:25:10] [warning] processes plugin: this platform has a 15 character limit to process names. The `Process "a very very long process name"' option will not work as expected.
```

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -4796,6 +4796,9 @@ collected for these selected processes are size of the resident segment size
 (RSS), user- and system-time used, number of processes and number of threads,
 io data (where available) and minor and major pagefaults.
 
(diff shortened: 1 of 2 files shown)


rubenk:
Thanks @mfournier!

PR #1548, opened 2016-02-10
[plugin] Chrony NTP service check

bbczeuz:
Hi,
I wrote a new plugin that collects statistics from chronyd. Chronyd is an implementation of NTP. It's the default NTP service on RHEL7, but packages for various other platforms exist (like Debian).

Tested on RHEL7 (x86_64), FreeBSD 10.2 (x86_64) and Debian 8.2.


rubenk:
Can you please rebase on current master?

bbczeuz:
@rubenk Rebased to my fork of collectd/master which I rebased to the upstream... The operation itself didn't work very smooth (had to skip many patches, added/removed files from other branches,...). Can you use it or do you prefer a plain diff? Build works on all platforms at jenkins :)

rubenk:
@bbczeuz first of all thanks for your contribution! These are a lot of changes which makes it somewhat hard to review. Not all of them seem to be needed and introduce unneccesary churn, for instance 9830d2a is a merge commit which just changes the file mode of configure.ac, and 4f36893 changes the collectd-nagios code which is probably an accident? I also see references to a tcpsock plugin and a wxt plugin?

I'd prefer if you squash your commits down to one, or a few commits that introduce the changes one by one.

bbczeuz:
Ok, I start liking ```git rebase -i``` ==> History cleaned. Do you want it even more compressed? I messed my branches in the past (I'm working on a wxt plugin (wxt520 weather sensor) and an extension to collectd-nagios to support TCP (#1053, open since 9 months...))

- I use ```/dev/urandom``` to avoid seeding the RNG via time(NULL)
- I use ```__attribute__((packed))``` to enforce/_hint_ the compiler to pack structs

rubenk:
Much better, thanks!

I'll leave the in-depth review to others more equipped to do that, but I'll leave some comments inline.

bbczeuz:
@rubenk: Thanks for the review. I implemented most of your comments. Except:
- "return (1)": I used that form it in the past (a now squashed commit), but removed the braces again as they look ugly to me.
- Indentation: I used vim's "gg=G" to enforce it's default which is tabbed. If there /is/ a strong opposition against tabs, I can change it to spaces.

rubenk:
Thanks for the update.

As for the coding style, we do have an existing coding style for a reason, so yes, there's s strong preference for new code to adhere to that style, as I think goes for most projects.

bbczeuz:
@rubenk, @mfournier 
- Rebased to current head.
- Fixed intendation to two spaces (as style guide suggests)

src/Makefile.am
```diff
@@ -216,6 +216,13 @@ cgroups_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 cgroups_la_LIBADD = libmount.la
 endif
 
+if BUILD_PLUGIN_CHRONY
+pkglib_LTLIBRARIES += chrony.la
+chrony_la_SOURCES = chrony.c
+chrony_la_LDFLAGS = $(PLUGIN_LDFLAGS)
+chrony_la_LIBADD =
```
rubenk: there's no need to set this.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
+/* chrony plugin for collectd (monitoring of chrony time server daemon)
+ **********************************************************************
+ * Copyright (C) Claudius M Zingerli, ZSeng, 2015-2016 
+ *
+ * Internas roughly based on the ntpd plugin
```
rubenk: Internals

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ * 
+ **********************************************************************
+*/
+/* Formated using 'GNU indent -bli0 -ce -ppi 2' and some manual alignment
```
rubenk: There's no use for this.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ * 
+ **********************************************************************
```
rubenk: Please remove all these starts and just continue the comment. Other plugins don't do this either.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+ * TODO:
+ *	- More robust udp parsing (using offsets instead of structs?)
+ *	  -> Currently chrony parses its data the same way as we do (using structs)
+ *	- Plausibility checks on values received
+ *	  -> Done at higher levels
+ */
+
+#include "config.h"
+
+#if HAVE_SYS_TYPES_H
```
rubenk: this is already included by collectd.h below

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+#  include <sys/types.h>        /* getaddrinfo */
+#endif
+#if HAVE_SYS_SOCKET_H
+#  include <sys/socket.h>
```
rubenk: same here.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+#define RAND_DEVICE_PATH   "/dev/random"        /* Used to initialize seq nr generator (fall back) */
+
+static const char *g_config_keys[] = {
+  CONFIG_KEY_HOST,
+  CONFIG_KEY_PORT,
+  CONFIG_KEY_TIMEOUT
+};
+
+static int g_config_keys_num = STATIC_ARRAY_SIZE(g_config_keys);
+static int g_is_connected = 0;
```
rubenk: no need to zero-initialize static variables.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+static int g_chrony_socket = -1;
+static time_t g_chrony_timeout = -1;
+static char *g_chrony_host = NULL;
```
rubenk: same here.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+static char *g_chrony_port = NULL;
+static char *g_plugin_instance = NULL;
+static uint32_t g_chrony_rand = 1;
+static uint32_t g_chrony_seq_is_initialized = 0;
```
rubenk: and here.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+  RPY_N_SOURCES        = 2,
+  RPY_SOURCE_DATA      = 3,
+  RPY_MANUAL_TIMESTAMP = 4,
+  RPY_TRACKING         = 5,
+  RPY_SOURCE_STATS     = 6,
+  RPY_RTC              = 7
+} eDaemonReplies;
+
+#define ATTRIB_PACKED __attribute__((packed))
+typedef struct ATTRIB_PACKED
```
rubenk: I think it's more clear if you used __attribute__((packed)) directly.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+    a = (ip >> 24) & 0xff;
+    b = (ip >> 16) & 0xff;
+    c = (ip >> 8) & 0xff;
+    d = (ip >> 0) & 0xff;
+    rc = snprintf(p_buf, p_buf_size, "%ld.%ld.%ld.%ld", a, b, c, d);
+    break;
+  case IPADDR_INET6:
+    ip6 = addr->addr.ip6;
+
+#ifdef FEAT_IPV6
```
rubenk: where does this define come from?

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+  }
+  assert(rc > 0);
+  return p_buf;
+}
+
+
+static int
+chrony_set_timeout()
+{
+  /*Set the socket's  timeout to g_chrony_timeout; a value of 0 signals infinite timeout */
```
rubenk: Please add a space after the /*

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+chrony_connect()
+{
+  /*Connects to the chrony daemon */
+  /*Returns 0 on success, !0 on error (check errno) */
+  int socket;
+
+  if (g_chrony_host == NULL)
+  {
+    g_chrony_host = strdup(CHRONY_DEFAULT_HOST);
+    assert(g_chrony_host);
```
rubenk: Please handle the error, instead of asserting.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+chrony_recv_response(tChrony_Response * p_resp, size_t p_resp_max_size,
+                     size_t * p_resp_size)
+{
+  ssize_t rc = recv(g_chrony_socket, p_resp, p_resp_max_size, 0);
+  if (rc <= 0)
+  {
+    ERROR(PLUGIN_NAME ": Error receiving packet: %s (%d)", strerror(errno),
+          errno);
+    return CHRONY_RC_FAIL;
+  } else
```
rubenk: put the else on the next line.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+  assert(p_req);
+  assert(p_resp);
+  assert(p_resp_size);
+
+  if (g_is_connected == 0)
+  {
+    if (chrony_connect() == 0)
+    {
+      g_is_connected = 1;
+    } else
```
rubenk: put the else on the next line.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+      resp_code = RPY_SOURCE_STATS;
+      valid_command = 1;
+      break;
+    default:
+      ERROR(PLUGIN_NAME ": Unknown request command (Was: %d)", p_command);
+      break;
+    }
+
+    if (valid_command == 0)
+    {
```
rubenk: we tend not to use braces for single-line if statements.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+    uint32_t seq_nr = rand_r(&g_chrony_rand);
+    p_req->header.f_cmd = htons(p_command);
+    p_req->header.f_cmd_try = 0;
+    p_req->header.f_seq = seq_nr;
+
+    DEBUG(PLUGIN_NAME ": Sending request (.cmd = %d, .seq = %d)", p_command,
+          seq_nr);
+    if (chrony_send_request(p_req, req_size) != 0)
+    {
+      break;
```
rubenk: same here

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+    }
+
+    DEBUG(PLUGIN_NAME ": Waiting for response");
+    if (chrony_recv_response(p_resp, resp_size, p_resp_size) != 0)
+    {
+      break;
```
rubenk: and here.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+  if (exp >= 1 << (FLOAT_EXP_BITS - 1))
+  {
+    exp -= 1 << FLOAT_EXP_BITS;
+  }
+
+  /* coef = (x << FLOAT_EXP_BITS) >> FLOAT_EXP_BITS; */
+  coef = uval % (1U << FLOAT_COEF_BITS);
+  if (coef >= 1 << (FLOAT_COEF_BITS - 1))
+  {
+    coef -= 1 << FLOAT_COEF_BITS;
```
rubenk: and here.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+#define FLOAT_COEF_MAX (-FLOAT_COEF_MIN - 1)
+
+  int32_t exp, coef;
+  uint32_t uval;
+
+  uval = ntohl(p_float.value);
+  exp = (uval >> FLOAT_COEF_BITS) - FLOAT_COEF_BITS;
+  if (exp >= 1 << (FLOAT_EXP_BITS - 1))
+  {
+    exp -= 1 << FLOAT_EXP_BITS;
```
rubenk: and here.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+    if (g_chrony_host != NULL)
+    {
+      free(g_chrony_host);
+    }
+    if ((g_chrony_host = strdup(p_value)) == NULL)
+    {
+      ERROR(PLUGIN_NAME ": Error duplicating host name");
+      return CHRONY_RC_FAIL;
+    }
+  } else if (strcasecmp(p_key, CONFIG_KEY_PORT) == 0)
```
rubenk: put the else on a new line please.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+    if (g_chrony_port != NULL)
+    {
+      free(g_chrony_port);
+    }
+    if ((g_chrony_port = strdup(p_value)) == NULL)
+    {
+      ERROR(PLUGIN_NAME ": Error duplicating port name");
+      return CHRONY_RC_FAIL;
+    }
+  } else if (strcasecmp(p_key, CONFIG_KEY_TIMEOUT) == 0)
```
rubenk: same here.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+  {
+    time_t tosec = strtol(p_value, NULL, 0);
+    g_chrony_timeout = tosec;
+  } else
```
rubenk: and here.

src/chrony.c
```diff
@@ -0,0 +1,1147 @@
...
+{
+  /* Collectd shutdown callback: Free mem */
+  if (g_is_connected != 0)
+  {
+    close(g_chrony_socket);
+    g_is_connected = 0;
+  }
+  if (g_chrony_host != NULL)
+  {
+    free(g_chrony_host);
```
rubenk: please use sfree() which sets the pointer to NULL too.

src/Makefile.am
bbczeuz: You mean the empty LIBADD? There are several other plugins defining an empty LIBADD, too.

src/chrony.c
bbczeuz: Sorry, I don't understand what you mean. HAVE_SYS_TYPES_H is usually defined in collectd.h, but there are no includes in config.h.

src/Makefile.am
rubenk: Yes. The other empty LIBADDs are extended later on, which is not the case here.

src/chrony.c
rubenk: You already include collectd.h below, which includes `<sys/types.h>`

src/chrony.c
bbczeuz: I use this at 13 places in the code. As this is a gcc/clang extension (=not c standard AFAIK), I prefer to define this at one place so it's easier to adapt to other compilers. Shall I test for GCC and issue a warning? Might break compilation due to ```-Werror```. Just stay quiet?
```


#if defined(__GNUC__) or defined(__clang__)
#  define ATTRIB_PACKED __attribute__((packed))
#else
#  warning Not defining packed attribute (unknown compiler)
#  define ATTRIB_PACKED 
#endif
```

src/chrony.c
bbczeuz: Is this negotiable? I once spent more than a day fixing some strange error in some other code that only happened due to missing braces, since then I consequently use braces everywhere.

src/chrony.c
bbczeuz: It was in the original code from chronyd. There it was statically defined in ```configure```. I removed the ```#if```, assuming inet_ntop is always defined (Didn't find any feature test macro). I'll push an updated code today evening after some testing on my systems.

rubenk: I think you're reasonably safe assuming inet_ntop exists on all the platforms we build on.
If not, we can always add a check to configure.ac later.

src/chrony.c
rubenk: We've had a similar discussion already. I personally don't disagree with your preference for bracing, but you want your code included in an existing project with an existing style, so you should adhere to it. If each plugin used a different style the codebase would quickly become a mess.

src/chrony.c
rubenk: Fair enough. Ideally we detect if __attribute__((packed)) is supported by the compiler (Sun Studio has it too iirc). We should do that for the other attributes we use in the codebase to, but let's leave that for another day :)

It's probably good to break compilation, you might even turn the warning into an error, so we get reports for other compilers.

One last thing, clang also defines __GNUC__ so there's no need to test for that seperately.

rubenk:
Thanks @bbczeuz!

eyalzek:
Hey, is there a way to use this plugin currently? I'm on collectd 5.5.2.git and trying to load this plugin I'm getting:
`plugin_load: Could not find plugin "chrony" in /usr/lib/collectd`

PR #1549, opened 2016-02-13
Enable TCP socket keepalive on write plugins

mfournier:
Up to now, collectd fails to notice when a remote server dies without properly closing the network connection, leading to more and more memory getting allocated as the values pile up in the write queue. This patch serie aims to help collectd identify this scenario and let the error handling code paths kick in.

I split this up to be able to cherry-pick bits to the release branches (I do consider collectd triggering the OOM killer as a bug :wink:)

The function name has no explicit reference to keepalive because I'd also like to add a global configure option to enable & tweak the socket close lingering (giving control on if and how collectd drops unflushed data at shutdown time). This part would be targetted to the next feature release.

TODO: check if plugins which have their network socket setup handled by an external lib (amqp, mqtt, postgresql, write_kafka, write_mongodb, write_redis) are also prone to this problem.

### DIFF
--- a/src/daemon/common.c
+++ b/src/daemon/common.c
@@ -51,6 +51,10 @@
 # include <netinet/in.h>
 #endif
 
+#if HAVE_NETINET_TCP_H
+# include <netinet/tcp.h>
+#endif
+
 /* for ntohl and htonl */
 #if HAVE_ARPA_INET_H
 # include <arpa/inet.h>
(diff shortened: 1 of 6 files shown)


rubenk:
Again nice work @mfournier! LGTM.

src/daemon/common.c
```diff
@@ -1591,6 +1595,46 @@ int service_name_to_port_number (const char *service_name)
...
+		WARNING("set_sock_opts: failed to determine socket type");
+		return;
+	}
+
+	if (socktype == SOCK_STREAM)
+	{
+		status = setsockopt(sockfd, SOL_SOCKET, SO_KEEPALIVE,
+				&so_keepalive, sizeof (so_keepalive));
+		if (status != 0)
+			WARNING("set_sock_opts: failed to set socket keepalive flag");
```
kev009: Minor; seems like preferred style is WARNING (

mfournier:
I just rebased this work against current master so it merges again, and took @kev009's feedback in account. I'll merge this once jenkins checks have passed.

I don't want to rush the "make timeout values configurable" part for 5.6. But I've been running with this patchset in production for a few months now, without any noticeable problem.

rubenk:
Sounds good.

PR #1550, opened 2016-02-16
processes: use long for pids on linux and solaris

mfournier:
Up to now, we had a mix of `int`, `long` and `pid_t` to represent the
variable part of `/proc/<PID>/status` and friends. This patch
standardizes on `long` on the Linux and Solaris platforms.

The max size of `int` is smaller than the maximum number of processes
current kernels support. `pid_t` is used to ensure portability of
functions such as `getpid()`, which aren't used in this plugin, and
apparently resolves to an `int` on Solaris.

This is a follow up to bb978c1, which triggered a format string issue on
solaris 64 bit.

/cc @coreykosak @dago

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -792,7 +792,7 @@ static void ps_submit_fork_rate (derive_t value)
 
 /* ------- additional functions for KERNEL_LINUX/HAVE_THREAD_INFO ------- */
 #if KERNEL_LINUX
-static procstat_t *ps_read_tasks_status (int pid, procstat_t *ps)
+static procstat_t *ps_read_tasks_status (long pid, procstat_t *ps)
 {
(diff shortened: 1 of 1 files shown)


mfournier:
Thanks for the feedback & suggestions @coreykosak !

I'm not really sure having a `cdpid_t` is worth the effort. Apart from the daemon code itself, only the exec and processes plugins seem to work with pids. On the other hand, `cdtime_t` is used all over the place, and is pretty critical to collectd's functioning.

My focus here was to do a bit of cleanup while fixing this build problem. So on the short term, I'll merge this so I can be sure the build passes on Solaris again. But let's keep the idea and maybe implement it if some deeper refactoring happens once.

Thanks again for your help and input on this !

Issue #664, opened 2014-07-01
many many syslog messages like "0 Success: 1 value has been dispatched"


putty12:
I´m using collectd 5.4.1 on gentoo 64bit.
I have written some exec plugins in bash.

Problem: i receive a lot of "0 Success: 1 value has been dispatched" messages which fills my syslog (systemd journal).

It seems that these message comes from "src/utils_cmd_putval.c:" but not 100% shure.

How can I disable that ?
Is it a feature ?

Thank you...
..regards putty12

kaos:
+1, would also like to see this resolved.

blysik:
Any idea how to quiet this message?

martbhell:
Same in 5.5.0-2 on CentOS7.
Getting rid of collectd in /var/log/messages by adding this to a file in /etc/rsyslog.d/: 

if $app-name contains 'collectd' then {
   stop
}

octo:
There are currently written to `STDOUT`, which should normally hooked up to `/dev/null`. Are you starting *collectd* with one of these newfangled init systems kids [*] are using these days?

[*] Get off my lawn! ;-)

octo:
I think there is a lot of overlap with #1327: in both cases logging of the `PUTVAL` commands received by the *exec plugin* are not done as expected.

gsauthof:
I can confirm this issue with CentOS 7 and Collectd 5.5.

(Where the Collectd package comes from the EPEL repository. The EPEL repository contains backports of Fedora packages to RHEL, maintained by the Fedora community.)

I can also confirm that collectd is started via systemd on such systems.

It is started like this (when using the default systemd startup file):

    /usr/sbin/collectd -C /etc/collectd.conf -f

(where `-f` is documented as 'collectd will also not close standard file descriptors')

Looking at `lsof` stdout and stderr are thus hooked up with journald:

    collectd 1322 root    1u  unix 0xffff88040842e580      0t0   5702647 socket
    collectd 1322 root    2u  unix 0xffff88040842e580      0t0   5702647 socket

Where `ss -xp` tells us:

    u_str  ESTAB 0 0 * 5702647  * 5702648 users:(("collectd",1322,2),("collectd",1322,1))
    u_str  ESTAB 0 0 /run/systemd/journal/stdout 5702648  * 5702647 users:(("systemd-journal",484,16))

aerusso:
My pull request #1553 should address this issue. I'd appreciate comments on that patch.

rubenk:
[...]

PR #1553, opened 2016-02-16
Suppress spammy debug messages of #664

aerusso:
Suppress putval responses when output is STDOUT, which in the past had been expected to be /dev/null. Addresses debug messages of #664.

This appears to be working on my system for the past day---the only place fh is set to stdout is in exec.c, where we want this message suppressed. I admit I do not understand what is going on in amqp.c, but this patch will not affect the behavior there, since fh is set to STDERR instead.

### DIFF
--- a/src/utils_cmd_putval.c
+++ b/src/utils_cmd_putval.c
@@ -217,7 +217,8 @@ int handle_putval (FILE *fh, char *buffer)
 	} /* while (*buffer != 0) */
 	/* Done parsing the options. */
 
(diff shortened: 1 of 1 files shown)


jsosic:
:+1:

aerusso:
@mfournier Could I get some feedback from you on this proposed patch?

matejzero:
Any news around this? My logs are overwhelmed by 'spam' messages and wouldn't like to custom patch every release if possible.

PR #1555, opened 2016-02-19
Report packets dropped in interface plugin

falzm:
Add `if_dropped` to the metrics reported by the interface plugin.

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -286,6 +286,10 @@ static int interface_read (void)
 		incoming = atoll (fields[2]);
 		outgoing = atoll (fields[10]);
 		if_submit (device, "if_errors", incoming, outgoing);
+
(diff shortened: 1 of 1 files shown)


rubenk:
This looks good to me, and I think this is useful to have. @mfournier what do you think?

mfournier:
Same for me, +1 for usefulness, thanks @falzm :-)

NB: quite a few of the plugins which run on a range of different platforms try to provide the same data accross all platforms. So let's add this feature, but for the sake of consistency, I think we should try to also provide the dropped packets metrics in the `HAVE_GETIFADDRS`, `HAVE_LIBKSTAT` and `HAVE_PERFSTAT` cases.

falzm:
Great, thanks!

PR #1557, opened 2016-02-24
* Collectd::Unixsock: Split messages sent by flush() to 1024 byte chunks to not overflow buffer in unixsock.c us_handle_client().

rpv-tomsk:
Collection3 scripts generates commands of huge length when many files needs to be flushed.

This causes errors in web-server logs like this:

index.cgi: FLUSH failed: Parsing options failed. at ../lib/Collectd/Graph/Common.pm

And errors from collectd like this:

collectd[1243]: unixsock plugin: failed to write to socket #8: Broken pipe

Proposed patch splits FLUSH to separate commands. We have used this in production for a long time, and I think it will be useful for others too.

### DIFF
--- a/bindings/perl/lib/Collectd/Unixsock.pm
+++ b/bindings/perl/lib/Collectd/Unixsock.pm
@@ -193,6 +193,32 @@ sub _socket_chat
 	return $cbdata;
 }
 
+# Send a raw message on a socket.
+# Returns true upon success and false otherwise.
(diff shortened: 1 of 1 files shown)


rubenk:
Please rebase and squash.

rpv-tomsk:
Updated.

rubenk:
Thanks!

PR #1561, opened 2016-02-25
modbus: avoid enabling libmodbus's debug flag by default

mfournier:
Having this enabled makes collectd's stdout and stderr very noisy. From
now on, only activate this flag when collectd is built with
--enable-debug.

Thanks to Eric Sandeen for mentioning this problem on IRC.


rubenk:
LGTM.

sandeen:
Looks good, thanks Marc!

PR #1563, opened 2016-02-25
Curl_JSON: Custom plugin name; PHP-FPM configuration example

rpv-tomsk:
Hi!

Proposed patch adds ability to set custom plugin name instead of 'curl_json'.
curl_json is a generic plugin and we found such feature very useful. I add configuration example of php-fpm status monitoring as example of custom plugin name usage.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -459,7 +459,7 @@
 #  </URL>
 ## PHP-FPM status metrics
 #<URL "http://nginx-status/php-fpm-status?json">
-#  PluginName 'phpfpm'
+#  Plugin 'phpfpm'
 #  Instance 'main'
 #  <Key "accepted conn">
 #      Type "phpfpm_requests"
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -457,6 +457,31 @@
 #      Type "bytes"
 #    </Key>
 #  </URL>
+## PHP-FPM status metrics
+#<URL "http://nginx-status/php-fpm-status?json">
+#  PluginName 'phpfpm'
+#  Instance 'main'
+#  <Key "accepted conn">
+#      Type "phpfpm_requests"
+#      Instance ""
+#  </Key>
+#  <Key "slow requests">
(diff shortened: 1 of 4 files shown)


rubenk:
Can you rebase?

toni-moreno:
Hi @rubenk @rpv-tomsk . 
I think this PR is a great idea. I would like also ask for add custom plugin name for all general purpose plugins, not also curl_xml or curl_json , as I did a couple of years ago.

https://github.com/collectd/collectd/pull/585

My PR is completely obsolete, but could be a great idea recover the main concept over all general purpose.

Thanks a lot both for your great work!!

rubenk:
Thanks for pointing me to #585 @toni-moreno. I've just read through it and I agree with @pyr in comment https://github.com/collectd/collectd/pull/585#issuecomment-65226391, in that it's best to come up with a generic solution for all collectd plugins instead of fixing this plugin for plugin.

@rpv-tomsk is this something you'd be willing to look in to?

rpv-tomsk:
I don't know how to implement this as a generic solution.

And as I understand, there is no way to do this.

For example, in 'curl_json', we want one plugin name to be reported for one URL and another plugin name - for another URL. But only curl_json plugin knows about URL's (like only 'tail' plugin knows about file paths), so only plugin can be responsible for reported plugin name. Plugin needs to know which name to report, so it must be configured in plugin configuration - i.e. implemented in plugin.

I can prepare patches for curl / tail / tail_csv like it already was done for curl_xml / curl_json, if you want.

rpv-tomsk:
Rebased.

rpv-tomsk:
Renamed 'PluginName' option to unified 'Plugin' and rebased.

rpv-tomsk:
ping

rpv-tomsk:

rpv-tomsk:

octo:
Merged as 23128c9c41d00cd2e5e3f529f1792a261affd0a4 and c5a2b1153577f3322bf14215c616326b731a5c51. Thanks @rpv-tomsk!

PR #1564, opened 2016-02-25
disk plugin: send disk_merged,disk_io_time,pending_operations only when real data exists for device

rpv-tomsk:
For Linux systems, disk plugin send 'disk_octets', 'disk_ops', 'disk_time' data only then real data exist.
There is no similar check for 'disk_merged', 'disk_io_time', 'pending_operations' metrics and their data is collected independently of real data existence.

Because of this, collectd produces empty metrics for LVM ('disk_merged' always zero) and Linux RAID MD devices ('disk_merged', 'disk_io_time', 'pending_operations' always zero). In some configurations there is too much unneeded metrics. 

Proposed patch solves this issue.

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -106,6 +106,10 @@ typedef struct diskstats
 	derive_t avg_read_time;
 	derive_t avg_write_time;
 
+	_Bool has_merged;
+	_Bool has_in_progress;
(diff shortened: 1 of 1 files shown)


rubenk:
Thanks!

PR #1572, opened 2016-02-27
Move unversioned .so to -devel subpackage

rubenk:


mfournier:
Merged as 6bf6dbc, thanks @rubenk !

PR #1573, opened 2016-02-27
Replace traffic with interface in man page

rubenk:
The traffic plugin was renamed a long time ago.


mfournier:
Nice catch ! Merged as b7f085f.

PR #1578, opened 2016-02-29
Nginx vhost

Zanop:
Add Vhost config option to allow for requesting the servers status from specific vhost. This is needed in case nginx exposes the server status only for specific vhost/s.

### DIFF
--- a/src/nginx.c
+++ b/src/nginx.c
@@ -40,8 +40,10 @@ static char *verify_peer = NULL;
 static char *verify_host = NULL;
 static char *cacert      = NULL;
 static char *timeout     = NULL;
+static char *vhost       = NULL;
 
 static CURL *curl = NULL;
+static struct curl_slist *headers = NULL;
 
 static char   nginx_buffer[16384];
 static size_t nginx_buffer_len = 0;
@@ -50,6 +52,7 @@ static char   nginx_curl_error[CURL_ERROR_SIZE];
 static const char *config_keys[] =
 {
   "URL",
+  "VHost",
   "User",
   "Password",
   "VerifyPeer",
@@ -99,6 +102,8 @@ static int config (const char *key, const char *value)
 {
   if (strcasecmp (key, "url") == 0)
     return (config_set (&url, value));
+  if (strcasecmp (key, "vhost") == 0)
+    return (config_set (&vhost, value));
   else if (strcasecmp (key, "user") == 0)
     return (config_set (&user, value));
   else if (strcasecmp (key, "password") == 0)
@@ -125,7 +130,7 @@ static int init (void)
     ERROR ("nginx plugin: curl_easy_init failed.");
     return (-1);
   }
-
+  curl_easy_setopt (curl, CURLOPT_VERBOSE, 1L);
   curl_easy_setopt (curl, CURLOPT_NOSIGNAL, 1L);
   curl_easy_setopt (curl, CURLOPT_WRITEFUNCTION, nginx_curl_callback);
   curl_easy_setopt (curl, CURLOPT_USERAGENT, COLLECTD_USERAGENT);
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Why this change is needed ?

I'm simply use the following config without any problems in my setup:

<Plugin nginx>
URL "http://nginx-status/nginx_status?auto"
</Plugin>

This sets required "Host: nginx-status"  header to requests of status from specific vhost.
All this works without any patches.

Zanop:
Hello, there are several scenarios:

1. For special purposes like api calls or for security reasons(you dont want status info seen by outsiders) you have named virtual host which does not actually resolve by the dns but instead you request that vhost to issue commands/check status.
2. You have multiple servers/IPs for A record for load balancing reasonds. You want to check specific server's status but if you call just by URL, say "vhost.domain.com" DNS resolver will send you to any one of those IPs. By having the Vhost option I can specify exactly which server I want to check with "http://XX.XX.XX.XX/nginx_status" and then the Vhost will determine which vhost on that server I will be checking.
3. You are replacing the server for site.domain.com with a new one which you are currently configuring. For the sake of having exactly the same server configuration on the new one you have vhost site.domain.com on it but your A record is not yet pointing to it because it is in dev/testing. You want collectd to pull data from the new server, so you have to specify its address which is different from site.domain.com which still points to the old one currently in production.

Probably can think of more uses but having this option adds a lot of flexebility which is very usefull for any devop.

regards,
Vladimir Smolensky

rpv-tomsk:
All you need in these cases - put record into /etc/hosts.

Zanop:
That is exactly what Im trying to avoid. Puting things in hosts afects the whole machine and Im trying to limit the chnges only to services I need. Also its an ugly hack that can lead to side effects.

rpv-tomsk:
>That is exactly what Im trying to avoid. Puting things in hosts afects the whole machine and Im trying to limit the chnges only to services I need. 

Collecting stats - is the service too. You have to do changes it to work anyway, and putting a line into /etc/hosts is no different from other changes. Also I can say what any change affects the whole machine, not only changes of /etc/hosts file.

I see no problem about that change affects whole machine. On the contrary, putting a line into /etc/hosts you can simply do "curl http://nginx-status/nginx_status?auto" and get result.

>Also its an ugly hack that can lead to side effects.

Strange opinion. File /etc/hosts is specially intended for such changes and changing it is not a hack. 
What are the side effects of such changes? I do not know such.

Poil:
:+1: @Zanop

rpv-tomsk:
Note about patch contents / proposed code changes:

    - I see no reasons to declare globally 'vhost' and 'headers' both.
    - curl_slist_append() can return NULL. curl* plugins check this, your code should check this too.
    - Freeing of 'headers' variable (curl_slist_free_all() call) is missing.

rpv-tomsk:
IMHO (I'm not member of project core team):

Project should be consistent. Nginx is not only one supported webserver, there is an 'apache' plugin exists too. The 'bind' and 'ascent' plugins also uses HTTP-requests for gathering stats. 

Currently, all them successfully works without custom 'Host' header and I see no real reasons to commit proposed patch.

curl* plugins has option 'Header', but I think such generic option would be superfluous for 'not generic' plugins.

In my view:

- If you want to add your proposed option, all plugins should be touched.
- Documentation patch should be provided too.

Zanop:
Hi, I've added NULL check although I don't think is needed, from libbcurl docs:

> Pass a NULL to this option to reset back to no custom headers.

As for freeing headers, I don't think I can since according to libcurl docs:

> When this option is passed to curl_easy_setopt, libcurl will not copy the entire list so you must keep it around until you no longer use this handle for a transfer before you call curl_slist_free_all on the list.

I can't find  plugin_deregister callback when I can do this.

Zanop:
As for the docs, Im not sure if I have to edit collectd.conf.pod collectd.conf.in or both?

rpv-tomsk:
Hi!

I believe, both files should be touched. The new option should be added as commented out line to collectd.conf.in.  

> I can't find plugin_deregister callback where I can do this.

It does not exists, but you can add it yourself, right? Don't forget about curl_easy_cleanup(curl);.

PR #1580, opened 2016-02-29
zfs_arc: 'stolen' was removed after FreeBSD 10.2

rubenk:
Patch taken from FreeBSD port


rubenk:
And indeed, on my FreeBSD-11-CURRENT system:
`[2016-03-06 11:19:36] [warning] zfs_arc plugin: Reading kstat value "stolen" failed.`

Issue #1536, opened 2016-02-03
5.5.1 release?


hnyman:
Version 5.5.1 was tagged 2 weeks ago and the .tar.gz file is also visible in http://collectd.org/files/

But as there has been no official announcement about the release, I wonder if that really is the final 5.5.1 release package, or if the 5.5.1 release is still on hold?

mfournier:
It *is* the official release, please use it ! It's the best collectd release ever ;-)

FYI the debian, ubuntu and fedora distros already have updated their packages.

@octo still needs to update the website. We've discussed putting the static pages of the website in a git repo, so others can also take care of this, but the idea didn't materialize yet (hint, hint!) ;-)

hnyman:
Thanks for info. Good to know that it is the release ;-)

In addition to the website, also here in Github the "Latest release" pointer on the "Releases" page still points to 5.5.0...
https://github.com/collectd/collectd/releases

roycec:
The spec file in the 5.5.1 tar file shows version 5.5.0

hnyman:
Any news on updates to the website and Github releases page?
They still reflect 5.5.0 as the newest release and 5.5.1 release notes are well hidden...

kev009:
This can be closed

matteocontrini:
Still not updated

rubenk:
We're working on it.

rubenk:
@octo just tagged and published 5.5.2, and the release notes for that release are on the website now.

octo:
Superseded by 5.5.2.

Issue #1531, opened 2016-01-29
Build RPM failed with notify_nagios plugins


frostynova:
OS: CentOS 6.7
Collectd: 5.5.1

rpmbuild -ba collectd.rpm

```

extracting debug info from /root/rpmbuild/BUILDROOT/collectd-5.5.1-1.el6.x86_64/usr/lib64/collectd/threshold.so
symlinked /usr/lib/debug/usr/lib64/libcollectdclient.so.1.0.0.debug to /usr/lib/debug/usr/lib64/libcollectdclient.so.debug
symlinked /usr/lib/debug/usr/lib64/libcollectdclient.so.1.0.0.debug to /usr/lib/debug/usr/lib64/libcollectdclient.so.1.debug
6157 blocks
+ /usr/lib/rpm/check-buildroot
+ /usr/lib/rpm/redhat/brp-compress
+ /usr/lib/rpm/redhat/brp-strip-static-archive /usr/bin/strip
+ /usr/lib/rpm/redhat/brp-strip-comment-note /usr/bin/strip /usr/bin/objdump
+ /usr/lib/rpm/brp-python-bytecompile
+ /usr/lib/rpm/redhat/brp-python-hardlink
+ /usr/lib/rpm/redhat/brp-java-repack-jars
Processing files: collectd-5.5.1-1.el6.x86_64
error: File not found: /root/rpmbuild/BUILDROOT/collectd-5.5.1-1.el6.x86_64/usr/lib64/collectd/notify_nagios.so
Executing(%doc): /bin/sh -e /var/tmp/rpm-tmp.X97ANL
```

rubenk:
Hi,

Where did you get the spec file from?
Please post your spec file and complete build output.

frostynova:
I got it from this git repo, what i modified is change mysql-devel to MySQL-devel (our own version)

collectd.spec:

https://gist.github.com/frostynova/1be21f03b4f051aeac4c

build.log:

https://gist.github.com/frostynova/fa7bed16f8676f43c6bc

mfournier:
It looks like you're using the RPM specfile from the `master` branch. If you want to build 5.5.x you must take it from the `collectd-5.5` branch.

frostynova:
@mfournier yes, i use the master code
Spec file from tag 5.5.1 built successfully
A little problem is the version in it is old:

Version: 5.5.0   
should be 
Version: 5.5.1

rubenk:
5.5.1 is not released yet.

[...]

PR #1583, opened 2016-03-01
Tested building the RPM for collectd 5.5.1 released a couple of

je-nunez:
weeks ago

Tested generating the RPM files with `mock -r centos-6-x86_64 ...`,
as in the comments at the beginning of './contrib/redhat/collectd.spec'

(As for the contents of the RPM history at the end of the file 
'./contrib/redhat/collectd.spec', there are no details specific for 
what is new in version 5.5.1 in respect to 5.5.0, as from

    https://collectd.org/wiki/index.php/Version_5.5

to:

    https://collectd.org/wiki/index.php/Version_5.5.1

which URL doesn't exist, as if it is not necessary because version
5.5.1 contains fixes over 5.5.0)

### DIFF
--- a/contrib/redhat/collectd.spec
+++ b/contrib/redhat/collectd.spec
@@ -221,7 +221,7 @@
 
 Summary:	statistics collection and monitoring daemon
 Name:		collectd
-Version:	5.5.0
+Version:	5.5.1
 Release:	1%{?dist}
 URL:		http://collectd.org
 Source:		http://collectd.org/files/%{name}-%{version}.tar.bz2


je-nunez:
Note:

In the `mock -r centos-6-x86_64 --rebuild ...` step it was needed to
disable some subpackages and options with:

      --without=amqp   --without=modbus  --without=nut \
      --without=pinba --without=ping  --without=redis \
      --without=varnish   --without=write_redis  \
      --without=write_riemann   --without=notify_nagios \

not by reason of any compilation issue per-se, but by reason of the
repository options: seems to be an issue with the base repositories
used in this test's `mock -r centos-6-x86_64`'s `config_opts['yum.conf']`
that didn't have the repositories to the RPM packages above. As it
has been said, this seem to be an issue with the base repositories 
listed in the `config_opts['yum.conf']` option for `mock`.

But the main RPM package `collectd-5.5.1` and most subpackages were
generated ok.

rubenk:
Did you enable the EPEL repository in your mock config as described in the comment in collectd.spec?
"Tested building the RPM for collectd 5.5.1 released a couple of
weeks ago" is not a very good commit message. Based on that description, I have no idea what your patch does.

The website indeed doesn't contain a page for 5.5.1 yet, and we haven't gotten around to updating the spec yet. This has been reported a few times already, see #1536 for example.

je-nunez:
Hi rubenk,

Yes, with the EPEL repository in the mock config, then mock builds the RPM package and subpackages for collectd version 5.5.1.

(The only issue that it happened to be encountered is with 

    --without=notify_nagios

that seems to be still necesary. This seems to be issue [1531](https://github.com/collectd/collectd/issues/1531) : a test with `mock` with the ideas there, and `--without=notify_nagios` doesn't seem to be necessary then.)

rubenk:
--without=$plugin is not valid and doesn't do anything, it is --without-$plugin.

#1531 is about building 5.5.1 with a spec from the wrong branch, that doesn't apply here.

PR #1588, opened 2016-03-02
Add LAST consolidation function RRAs in RRD files

guillemj:
This makes this available for later querying, otherwise the information gets lost.

Also fixed an issue in add_rra that was failing when I tested the changes.


clemensg:
The LAST RRA data would be nice to have!
Is this PR _pending_ due to the merge conflicts or due to other concerns?

rpv-tomsk:
Ruben, what do you think about this? I can take care if you think it's necessary.

rpv-tomsk:
Ok, will postpone this PR again )

PR #1589, opened 2016-03-02
Fix memory leak in redis.c

ajdiaz:
There was a memory leak in redis.c module when using pasword to authenticate connection. The reply object (rr) created was not removed if authentication succeed, but rr variable was override into the next if condition.

This patch fix the leak and improve some error messages when authentication fails.

### DIFF
--- a/src/redis.c
+++ b/src/redis.c
@@ -400,24 +400,26 @@ static int redis_read (void) /* {{{ */
     if (strlen (rn->passwd) > 0)
     {
       DEBUG ("redis plugin: authenticanting node `%s' passwd(%s).", rn->name, rn->passwd);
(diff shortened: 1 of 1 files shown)


rubenk:
Thanks for your contribution @ajdiaz! I left some small comments inline. Since this is a bugfix, would you mind submitting your PR against the collectd-5.5 branch?

ajdiaz:
Thanks for the comments @rubenk, I submit the PR against master to be included in next relase, but you are right, I can move the PR to collectd-5.5 branch if you prefer... So... Should we close this PR and open a new one or keep both?

rubenk:
@ajdiaz I squashed both your commits together and merged them in both the collectd-5.5 and master branch (bb13476)

ajdiaz:
Thanks @rubenk!

PR #1592, opened 2016-03-03
mysql plugin: Split 'mysql_sort' type into several types.

rpv-tomsk:
The 'mysql_sort' type needs to be splitted.

As documented in http://dev.mysql.com/doc/refman/5.6/en/server-status-variables.html#statvar_Sort_range , 'Sort_*' keys are:

 * The number of sorts:    Sort_range, Sort_scan
 * The number of sorted rows: Sort_rows
 * The number of merge passes: Sort_merge_passes

These are cannot be mixed together into one type 'mysql_sort'. 
Proposed patch solves this problem.

https://github.com/collectd/collectd/commit/7d1ee33e6e2250ab6c52be935c897ad8141fdea4

### DIFF
--- a/src/mysql.c
+++ b/src/mysql.c
@@ -900,8 +900,15 @@ static int mysql_read (user_data_t *ud)
 		}
 		else if (strncmp (key, "Sort_", strlen ("Sort_")) == 0)
 		{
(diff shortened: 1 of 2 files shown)


rubenk:
Good idea, thanks.

PR #1608, opened 2016-03-07
xencpu plugin: RFC release

rpv-tomsk:
Hi,
I wrote a new simple plugin that collects statistics about physical CPU load under XEN.

Tested on Centos5  with Xen-3.4, Debian 7 with Xen-4.1.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1277,6 +1277,13 @@ write_tsdb_la_SOURCES = write_tsdb.c
 write_tsdb_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 endif
 
+if BUILD_PLUGIN_XENCPU
+pkglib_LTLIBRARIES += xencpu.la
+xencpu_la_SOURCES = xencpu.c
+xencpu_la_LDFLAGS = $(PLUGIN_LDFLAGS)
+xencpu_la_LIBADD = -lxenctrl
+endif
+
 if BUILD_PLUGIN_XMMS
 pkglib_LTLIBRARIES += xmms.la
 xmms_la_SOURCES = xmms.c
--- a/configure.ac
+++ b/configure.ac
@@ -4709,6 +4709,46 @@ then
 fi
 # }}}
 
+# --with-libxenctrl {{{
+with_libxenctrl_cppflags=""
+with_libxenctrl_ldflags=""
+AC_ARG_WITH(libxenctrl, [AS_HELP_STRING([--with-libxenctrl@<:@=PREFIX@:>@], [Path to libxenctrl.])],
+[
+	if test "x$withval" != "xno" && test "x$withval" != "xyes"
+	then
+		with_libxenctrl_cppflags="-I$withval/include"
+		with_libxenctrl_ldflags="-L$withval/lib"
+		with_libxenctrl="yes"
+	else
+		with_libxenctrl="$withval"
+	fi
+],
+[
+	with_libxenctrl="yes"
+])
+if test "x$with_libxenctrl" = "xyes"
+then
+	SAVE_CPPFLAGS="$CPPFLAGS"
+	CPPFLAGS="$CPPFLAGS $with_libxenctrl_cppflags"
+
+	AC_CHECK_HEADERS(xenctrl.h, [with_libxenctrl="yes"], [with_libxenctrl="no (xenctrl.h not found)"])
+
+	CPPFLAGS="$SAVE_CPPFLAGS"
+fi
+if test "x$with_libxenctrl" = "xyes"
+then
+	SAVE_CPPFLAGS="$CPPFLAGS"
+	SAVE_LDFLAGS="$LDFLAGS"
+	CPPFLAGS="$CPPFLAGS $with_libxenctrl_cppflags"
+	LDFLAGS="$LDFLAGS $with_libxenctrl_ldflags"
+
+	#Xen versions older than 3.4 has no xc_getcpuinfo()
+	AC_CHECK_LIB(xenctrl, xc_getcpuinfo, [with_libxenctrl="yes"], [with_libxenctrl="no (symbol 'xc_getcpuinfo' not found)"], [])
+
+	CPPFLAGS="$SAVE_CPPFLAGS"
+	LDFLAGS="$SAVE_LDFLAGS"
+fi
+
 # --with-libxmms {{{
 with_xmms_config="xmms-config"
 with_xmms_cflags=""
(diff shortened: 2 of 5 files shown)


rubenk:
Just wondering, I'm not too familiar with xen, but it seems that everything is moving toward using libxl (libxenlite) instead of using the low-level interfaces. Have you considered that?

rpv-tomsk:
>Just wondering, I'm not too familiar with xen, but it seems that everything is moving toward
> using libxl  (libxenlite) instead of using the low-level interfaces. Have you considered that?

Sure, you are right about move directions. 'libxl' is intended to hide 'xenstore', 'xenctrl', and 'xenguest' libraries API from higher levels and to provide one unified API to them. But that does not mean what low-level interfaces will disappear, conversely, 'libxl' uses 'xenctrl' in it's implementation, so that 'xenctrl' will have a long life.  

Low-level 'xenctrl' fully suits for plugin needs. If we want to use a higher-level API, provided by 'libxl' - we are unable. The 'libxl' has no support of 'XEN_SYSCTL_getcpuinfo' hypercall in any kind currently (xen-4.4, xen-4.6 was checked), and also it is too heavy and has too many dependencies for such simple use (only one hypercall is needed by plugin).

At Debian 8:

# ldd /usr/lib/x86_64-linux-gnu/libxenlight-4.4.so
        linux-vdso.so.1 (0x00007ffd414b4000)
        libxenctrl-4.4.so => /usr/lib/x86_64-linux-gnu/libxenctrl-4.4.so (0x00007f1cbf07a000)
        libxenguest-4.4.so => /usr/lib/x86_64-linux-gnu/libxenguest-4.4.so (0x00007f1cbee4f000)
        libxenstore.so.3.0 => /usr/lib/x86_64-linux-gnu/libxenstore.so.3.0 (0x00007f1cbec44000)
        libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f1cbea41000)
        libuuid.so.1 => /lib/x86_64-linux-gnu/libuuid.so.1 (0x00007f1cbe83c000)
        libyajl.so.2 => /usr/lib/x86_64-linux-gnu/libyajl.so.2 (0x00007f1cbe631000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f1cbe414000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1cbe069000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f1cbde65000)
        liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x00007f1cbdc42000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f1cbda27000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f1cbf508000)

I think, 'xenctrl' will be accessible and can be used by plugin for a long time.
The things may change when time comes, though.

rubenk:
Ok, thanks for the clarification. And there I thought libxl was supposed to be a "lightweight" interface:
http://wiki.xen.org/wiki/Choice_of_Toolstacks#Libxenlight_.28libxl.29

I just want to avoid a scenario where Xen drops support for the low-level interface and we have to keep maintaining the plugin. Collectd versions will have to be supported for a long time too.

rubenk:
Could you also add some documentation to collectd.conf.inf and collectd.conf.pod?

rpv-tomsk:
Thank you for your review of my patch. Please take look at new version.

src/Makefile.am
```diff
@@ -1286,6 +1286,13 @@ xmms_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 xmms_la_LIBADD = $(BUILD_WITH_LIBXMMS_LIBS)
 endif
 
+if BUILD_PLUGIN_XENCPU
```
rubenk: Please keep the list sorted and move this above BUILD_PLUGIN_XMMS

rubenk:
Thanks a lot. I have just one more remark. Is there any chance that this plugin will be able to gather more statistics from Xen, like memory for instance, in the future? In that case  you might just want to call it "xen" instead of "xencpu".

rpv-tomsk:
>Is there any chance that this plugin will be able to gather more statistics from Xen,
> like memory for  instance, in the future? 

I think these tasks are successfully solved by 'virt' plugin, isn't it ?
Currently I see no reasons for duplicate its functions. But if you want I see no reasons against rename too.

src/Makefile.am
rpv-tomsk: Updated, thanks for a catch.

rubenk:
The virt plugin currently doesn't expose dom0 statistics. If I'm not mistaken libvirt has access to those so there's no reason why it couldn't.

rpv-tomsk:
>If I'm not mistaken libvirt has access to those so there's no reason why it couldn't.

I do not understand your thought. Can you please explain it more detailed?

rpv-tomsk:
>The virt plugin currently doesn't expose dom0 statistics. 

As for Xen, dom0 is a generic /Linux or other OS/ environment.
Xentop shows only CPU (used time + number of cpus) and memory limit values for Dom0.
Network RX/TX or disk (VBD - virtual block device) IO not  provided there.

Which statistics do you expect to get?

rubenk:
Can we extend the libvirt plugin to get memory and cpu values from the hypervisor?

rpv-tomsk:
As I know, libvirt provides stats only about VM, not about hardware / main OS.

The libvirt-1.3.2 (latest) has no references to XEN xc_getcpuinfo() function, so we are unable to get HW CPU load by using it.

rubenk:
I saw https://libvirt.org/html/libvirt-libvirt-host.html#virNodeGetCPUStats
But you're right, this is only implemented for the Qemu, LXC and UML drivers.

rpv-tomsk:
Yes, XEN driver has no implementation of 'nodeGetCPUStats'.

rubenk:
I think this is good to go in. @mfournier what do you think?

mfournier:
@rubenk, I'm really sorry, I didn't take much time to track what's going on with collectd lately :-(

I can't answer to your question right now (didn't really look at the code yet), but please merge it if you think it's good to go ;-)

mfournier:
... and thanks a lot for your work and recent help on the project @rpv-tomsk !! :-)

rubenk:
@mfournier no problem at all, I know the feeling all too well.

@rpv-tomsk can you rebase? Then I'll merge it.

rpv-tomsk:
Hi guys! Nice to meet you here.

I will do rebase soon (this week, ~ at Thursday).

rpv-tomsk:
Rebased.

rubenk:
Thanks @rpv-tomsk!

PR #1609, opened 2016-03-09
Nginx plugin: Incorrect metric provided

rpv-tomsk:
As documented in http://nginx.org/en/docs/http/ngx_http_stub_status_module.html , nginx provides following status information:

 accepts -   The total number of accepted client connections. 
 handled -   The total number of handled connections. 

The 'handled', as noticed, is the same as 'accepts' unless some resource limits have been reached (for example, the worker_connections limit). 

In normal state, these values are equal. When an error occurs, the difference between these values appear, but rate of this difference is too small (compared with the number of accepted connections) to see it on graphs.  So, it does not useful to provide pure 'handled' metric, but 'failed' (failed=accepts-handled) should be provided instead. Such 'failed' metric also allows to configure notifications, based on it, which is impossible in current state.

### DIFF
--- a/src/nginx.c
+++ b/src/nginx.c
@@ -256,7 +256,7 @@ static int nginx_read (void)
   /*
    * Active connections: 291
    * server accepts handled requests
-   *  16630948 16630948 31070465
+   *  101059015 100422216 347910649
    * Reading: 6 Writing: 179 Waiting: 106
    */
   for (int i = 0; i < lines_num; i++)
@@ -277,6 +277,7 @@ static int nginx_read (void)
       {
 	submit ("connections", "accepted", atoll (fields[0]));
 	submit ("connections", "handled", atoll (fields[1]));
+	submit ("connections", "failed", (atoll(fields[0]) - atoll (fields[1])));
 	submit ("nginx_requests", NULL, atoll (fields[2]));
       }
     }


tokkee:
Dropping "handled" is a backward-incompatible change, so we cannot do this in collectd 5.x. Anyway, I don't see a reason to drop it. For one, "failed" can be calculated from "accepted"-"handled" when creating the graph, so we don't have to record it separately but even if we record it separately, it can be added in addition to the existing metrics.

octo:
+1, if required, the "failed" metric needs to be added. Having "handled" and "accepted" doesn't follow our best practice, but since this has been released it takes a back seat to backwards compatibility.

rpv-tomsk:
Please change the label. PR is pending team action, not contributor.

octo:
Hi @rpv-tomsk, thanks for keeping an eye on this!

I don't think this *needs* to be configurable. If you want to add an option, I'd probably call it *ReportFailedConnections*.

If you do make it configurable, then please follow the best practice when it is enabled and report accepted and failed only, i.e. omit the total (aka. handled).

Best regards,
—octo

rpv-tomsk:
Hi!

If this needs no configuration when this patch is already here.

@octo, @tokkee ?

octo:
@rpv-tomsk Not quite: the patch, as it is, *removes* the metric for "handled" connections unconditionally. That is breaking backwards compatibility.

The easiest way forward is simply to add the line for "handled" connections again.

rpv-tomsk:
Hi!

Patch updated.

rpv-tomsk:
Hi! Could anybody take care about failed build/checks or simply commit PR?

PR #1611, opened 2016-03-10
Collection3: Graphing speed improvements with many hosts

stefanth:
At present graph.cgi is extremely slowly on servers with large numbers of
RRD files.  This is because it issues an unnecessary stat() call to _every_
RRD file on the local system before loading and rendering graph data.

Patch originally from:
http://mailman.verplant.org/pipermail/collectd/2011-May/004519.html

The patch was improved so that it correctly handles multiple hostnames.

I'm seeing a 5x improvement in speed, the original patch creator
had a 24x speedup.


rubenk:
That certainly seems like a nice improvement. Thanks!

PR #1612, opened 2016-03-12
Tail plugin: DSType Latency

rpv-tomsk:
This PR proposes new DSType "Latency" in 'tail' plugin and allows to calculate percentiles. 
Useful for webserver response time monitoring.

There is a plan to add 'request rate by range' reporting based on this DSType.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -31,6 +31,9 @@ libmount_la_LIBADD = daemon/libcommon.la
 liblookup_la_SOURCES = utils_vl_lookup.c utils_vl_lookup.h
 liblookup_la_LIBADD = daemon/libavltree.la daemon/libcommon.la
 
+liblatency_la_SOURCES = utils_latency.c utils_latency.h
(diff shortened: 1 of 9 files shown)


rpv-tomsk:
@Zanop , @Poil  - you can be interested in this patch.

rpv-tomsk:
Not only 'statsd' plugin needs to be rebuilt when utils_latency.c changes, but 'tail' plugin too.
What do you think about new revision?

Also, please, lets look into another example: src/daemon/utils_ignorelist.[ch].
Many plugins depends on this, but no one will be rebuilt when these files are changed. Is it ok?
Maybe 'collectd_SOURCES' in src/daemon/Makefile.am needs to be reorganized with 'libcommon_la_SOURCES'  (not in this PR, of course)?

rubenk:
As this PR introduces new functionality, it should go into the master branch.

There are indeed some dependency issues currently. I'm not in favor of introducing more intermediate convenience libraries for this though. This is better discussed in a separate PR indeed.

PR #1615, opened 2016-03-15
write_http: in cleanup handler free http headers

skinowski:
slist of http headers require cleanup. See libcurl
example:

https://curl.haxx.se/libcurl/c/httpcustomheader.html

### DIFF
--- a/src/write_http.c
+++ b/src/write_http.c
@@ -69,6 +69,7 @@ struct wh_callback_s
         int format;
 
         CURL *curl;
+        struct curl_slist *headers;
         char curl_errbuf[CURL_ERROR_SIZE];
 
         char  *send_buffer;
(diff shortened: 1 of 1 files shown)


skinowski:
xmms error? This sounds like a bogus failure.

rubenk:
@skinowski yes, this is bogus. The EPEL7 builders needs to be rebuild.

rubenk:
Thanks for the patch. Storing the headers in cb->headers seems unrelated to what your commit message says this PR does. What is the advantage of this? That might need a separate commit.

rpv-tomsk:
>What is the advantage of this?

There is no other correct way to pass 'headers' value to  wh_callback_free() -> curl_slist_free_all();.

IMHO: I think this should be marked as a 'Bug' and applied to all supported branches.

rubenk:
@rpv-tomsk I missed that. Github making function definitions light gray in a diff isn't helping.

@skinowski this is indeed a bug, can you please close this PR and open a new one against the 5.4 branch?

skinowski:
closing and opening a new one against 5.4.

PR #1620, opened 2016-03-20
write_http: in cleanup handler free http headers

skinowski:
slist of http headers require cleanup to avoid
memory leak. See libcurl example:

https://curl.haxx.se/libcurl/c/httpcustomheader.html

### DIFF
--- a/src/write_http.c
+++ b/src/write_http.c
@@ -56,6 +56,7 @@ struct wh_callback_s
         int format;
 
         CURL *curl;
+	struct curl_slist *headers;
         char curl_errbuf[CURL_ERROR_SIZE];
 
         char   send_buffer[4096];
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Superseedes #1615.

rpv-tomsk:
Collectd-4.10 also has this plugin. The https://collectd.org/download.shtml describes 4.10 as 'old stable' together with 5.4 and 5.5. The https://collectd.org/wiki/index.php/Main_Page has the same list of 'current versions'. 

As this is a bug, then it needs to be fixed in all supported versions, so 4.10, 5.4 and 5.5 to be affected.

Personally I see no reasons for 4.x branch support as 5.x is much better for me, but I see no messages what 4.x is out of support also. The 4.10 branch has some other unreleased changes.  I think, at least one minor release may be done before closing support, if such decision will be accepted. What do you think?

rubenk:
@rpv-tomsk At this point we're only fixing very severe issues in the 4 branch.

@skinowski thank you for your contribution!

PR #1622, opened 2016-03-20
statsd, utils_latency: allow to reduce bin width

rpv-tomsk:
Currently, utils_latency.c code allows only to increase hystogram bin width.
When any spike arrives, bin_width can raise too high and it never decreases later, unless collectd restarted.

Proposed patch allows bin width to be reduced to required level.

Submitting @octo as author of utils_latency.c code.

### DIFF
--- a/src/utils_latency.c
+++ b/src/utils_latency.c
@@ -125,8 +125,8 @@ latency_counter_t *latency_counter_create (void) /* {{{ */
   if (lc == NULL)
     return (NULL);
 
-  latency_counter_reset (lc);
   lc->bin_width = HISTOGRAM_DEFAULT_BIN_WIDTH;
+  latency_counter_reset (lc);
   return (lc);
 } /* }}} latency_counter_t *latency_counter_create */
 
@@ -175,6 +175,28 @@ void latency_counter_reset (latency_counter_t *lc) /* {{{ */
     return;
 
   cdtime_t bin_width = lc->bin_width;
+  cdtime_t max_bin = (lc->max - 1) / lc->bin_width;
+
+/*
+  If max latency is REDUCE_THRESHOLD times less than histogram's range,
+  then cut it in half. REDUCE_THRESHOLD must be >= 2.
+  Value of 4 is selected to reduce frequent changes of bin width.
+*/
+#define REDUCE_THRESHOLD 4
+  if ((lc->num > 0) && (lc->bin_width >= HISTOGRAM_DEFAULT_BIN_WIDTH * 2)
+     && (max_bin < HISTOGRAM_NUM_BINS / REDUCE_THRESHOLD))
+  {
+    /* new bin width will be the previous power of 2 */
+    bin_width = bin_width / 2;
+
+    DEBUG("utils_latency: latency_counter_reset: max_latency = %.3f; "
+          "max_bin = %"PRIu64"; old_bin_width = %.3f; new_bin_width = %.3f;",
+        CDTIME_T_TO_DOUBLE (lc->max),
+        max_bin,
+        CDTIME_T_TO_DOUBLE (lc->bin_width),
+        CDTIME_T_TO_DOUBLE (bin_width));
+  }
+
   memset (lc, 0, sizeof (*lc));
 
   /* preserve bin width */


rpv-tomsk:
without-bin-reduce

rpv-tomsk:
Hi.

@tokkee, @mfournier,  @rubenk  - let's merge this PR?

src/utils_latency.c
```diff
@@ -176,6 +176,28 @@ void latency_counter_reset (latency_counter_t *lc) /* {{{ */
...
+
+/*
+  If max latency is REDUCE_THRESHOLD times less than histogram's range,
+  then reduce bin width twice. REDUCE_THRESHOLD must be >= 2.
+  Value of 4 is selected to reduce frequent changes of bin width.
+*/
+#define REDUCE_THRESHOLD 4
+  if ((lc->bin_width >= HISTOGRAM_DEFAULT_BIN_WIDTH * 2)
+     && (max_bin < HISTOGRAM_NUM_BINS / REDUCE_THRESHOLD))
+  {
```
tokkee: I think this should also check for lc->num != 0, right? Else, we'd reduce bin_width in case we didn't receive any values (i.e. lc->max would be 0).

src/utils_latency.c
```diff
@@ -176,6 +176,28 @@ void latency_counter_reset (latency_counter_t *lc) /* {{{ */
     return;
 
   cdtime_t bin_width = lc->bin_width;
+  cdtime_t max_bin = (lc->max - 1) / lc->bin_width;
+
+/*
+  If max latency is REDUCE_THRESHOLD times less than histogram's range,
+  then reduce bin width twice. REDUCE_THRESHOLD must be >= 2.
```
tokkee: Nit: "twice" isn't right. We cut it in half (but only once).

src/utils_latency.c
```diff
@@ -126,8 +126,8 @@ latency_counter_t *latency_counter_create (void) /* {{{ */
     return (NULL);
   memset (lc, 0, sizeof (*lc));
 
-  latency_counter_reset (lc);
   lc->bin_width = HISTOGRAM_DEFAULT_BIN_WIDTH;
+  latency_counter_reset (lc);
```
tokkee: Side-note: This change is a no-op.

tokkee:
Thanks, looks mostly good to me. I could have fixed this myself but I'm sending it back to you for double-checking it.

src/utils_latency.c
rpv-tomsk: latency_counter_reset() has 'cdtime_t max_bin = (lc->max - 1) / lc->bin_width;' so it would be division by zero without this change.

src/utils_latency.c
rpv-tomsk: When we didn't receive any values then lc->bin_width will be HISTOGRAM_DEFAULT_BIN_WIDTH so first condition will fail.

tokkee: No, I mean when we didn't receive any values on later iterations. That is, first, we received values and then there's a gap (for whatever reason). Since this is a generic package, it should behave right in such cases.

src/utils_latency.c
tokkee: Oh, good point. I think that should actually be fixed in counter_reset(). Otherwise, it's very easy to accidentally break this in the future with an unrelated change.

src/utils_latency.c
rpv-tomsk: Ok, when no values was received due to gap then lc->max will be equal 0 and lc->bin_width will reduce down to HISTOGRAM_DEFAULT_BIN_WIDTH, by half of width at each latency_counter_reset() call.

I think that is ok, as we don't know which values will arrive after gap.
If you prefer to save bin width during gap - it will be ok to check lc->num !=0 as proposed by you.

Which the final decision will be about this?

tokkee: Either decision will be some sort of guess but I think it's a decent assumption that you'll get similar values after the gap, hence, we can avoid the need to size up again by adding that check.

rpv-tomsk:
Updated:
* Changed comment
* Added "lc->num > 0" check

rpv-tomsk:
How about merging this ?

octo:
LGTM; could you rebase this on *master*? I can do it, too, but I'd prefer it if PR history was intact :)

Best regards,
—octo

rpv-tomsk:
Rebased.

octo:
Thanks @rpv-tomsk :)

PR #1624, opened 2016-03-21
Disable logging option added to the GenericJMX plugin

kurtmc:
This pull request implements the option to disable logging for the GenericJMX plugin.
.

### DIFF
--- a/bindings/java/Makefile.am
+++ b/bindings/java/Makefile.am
@@ -22,7 +22,8 @@ EXTRA_DIST = org/collectd/api/CollectdConfigInterface.java \
 	     org/collectd/java/GenericJMXConfMBean.java \
 	     org/collectd/java/GenericJMXConfValue.java \
 	     org/collectd/java/GenericJMX.java \
-	     org/collectd/java/JMXMemory.java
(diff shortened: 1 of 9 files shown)


kurtmc:
I'm not sure how my last commit broke the build. I modified src/collectd-java.pod, is there something I am missing?

rubenk:
I suspect it is an issue with the build scripts. See https://ci.collectd.org/job/pull-requests-prepare-tarball/779/consoleText, at the bottom. Since you touched collectd-java.pod, make dist-gzip should remake collectd-java.5 from collectd-java.pod. For some reason it isn't.

@mfournier any ideas?

kurtmc:
I rebased my branch, will Jenkins build this again?

kurtmc:
@octo since you are the main contributor to GenericJMX, would you be able to comment on whether this is good enough to merge?

I need to disable logging because I have a config that pulls many attributes from a certain type of MBean but some of the MBeans do not have all of the attributes in the list, so GenericJMX spams the log with messages about not being able to find the particular attribute for some MBeans.

PR #1627, opened 2016-03-21
Remove broken filter syntax from collectd.conf(5)

rubenk:
fb67910d documented the new syntax.
The old syntax does no longer work, remove it.


rubenk:
@tokkee does this look OK to you?

mfournier:
All good, from what I understand. Thanks @rubenk !

PR #1625, opened 2016-03-21
Fix process plugin memory leak [solaris]

jquinn47:
On solaris, if a process is flagged as zombie, the 'myStatus', 'myInfo', and 'myUsage' structures are leaked. I've added the required frees to release the resources before the method returns.

PR #1630, opened 2016-03-22
Fix process plugin memory leak [solaris]

jquinn47:
Applying PR to collectd-5.4 branch.

Same change as https://github.com/collectd/collectd/pull/1625

On solaris, if a process is flagged as zombie, the 'myStatus', 'myInfo', and 'myUsage' structures are leaked. I've added the required frees to release the resources before the method returns.


rubenk:
Almost; This is still against the master branch. You'll need to select the collectd-5.4 branch in the dropdown when you open the PR.

rubenk:
Merged into 5.4 as f86731f4fca466e89453b02bf185730e7d547898. Thanks @jquinn47.

PR #1634, opened 2016-03-24
add http headers functionality to write_http plugin

BrandonArp:
This commit allows the addition of arbitrary headers into the POST requests.  The code is taken (and adapted) from the curl plugin.

### DIFF
--- a/src/write_http.c
+++ b/src/write_http.c
@@ -156,7 +156,6 @@ static int wh_callback_init (wh_callback_t *cb) /* {{{ */
         curl_easy_setopt (cb->curl, CURLOPT_NOSIGNAL, 1L);
         curl_easy_setopt (cb->curl, CURLOPT_USERAGENT, COLLECTD_USERAGENT);
 
-        cb->headers = NULL;
         cb->headers = curl_slist_append (cb->headers, "Accept:  */*");
         if (cb->format == WH_FORMAT_JSON)
                 cb->headers = curl_slist_append (cb->headers, "Content-Type: application/json");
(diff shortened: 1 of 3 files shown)


BrandonArp:
@rubenk Any chance you could take a look at this?  The changes here conflict with the multi-threaded HTTP rewrite in #1639, but I have a commit that solves the conflict (and can work with @skinowski to get it applied).

src/write_http.c
```diff
@@ -540,6 +539,22 @@ static int config_set_format (wh_callback_t *cb, /* {{{ */
...
+static int wh_config_append_string (const char *name, struct curl_slist **dest, /* {{{ */
+    oconfig_item_t *ci)
+{
+  if ((ci->values_num != 1) || (ci->values[0].type != OCONFIG_TYPE_STRING))
+  {
+    WARNING ("write_http plugin: `%s' needs exactly one string argument.", name);
+    return (-1);
+  }
+
+  *dest = curl_slist_append(*dest, ci->values[0].value.string);
```
skinowski: **curl_slist_append** is tricky. Here *dest will leak if it returns NULL since we clobbered it and lost the change of freeing it. I'd use a temp variable here to store and check the return value, and then overwrite *dest. (The original http_write code has these leaks as well.)

BrandonArp: Thanks for calling that out.  I fixed this on a combined branch, but didn't fix it here.  I'll make sure to update this PR.

rubenk: Ah yes. It would be great if one of you could do a separate bugfix PR first which fixes this issue throughout the codebase. From a cursory glance, the curl, curl_json and curl_xml plugins also do this wrong.

rubenk:
@BrandonArp other than the curl_slist_append issue, this PR looks good. It's only missing some documentation. Could you add something to collectd.conf.pod and collectd.conf.in?

BrandonArp:
Certainly.  I'll get this updated ASAP.

BrandonArp:
Updated with documentation and the fix for the possible memory leak.  I'll tackle the other possible leaks and have a PR for you soon.

rubenk:
Nice work, thanks!

PR #1636, opened 2016-03-25
Add alias to Ping Host config

nward:
This pull request adds an "alias" option to the ping plugin's Host configuration statement. The alias is used in all the external output of the host, i.e. in the logs, and more importantly, in stats that are written to anywhere.

In this way, we can have long term stats for something, even if its IP address changes time. Of course, collectd configuration is required to track the new IP address, but if the alias stays consistent then we will write the stats to the same place.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -930,6 +930,7 @@
 
 #<Plugin ping>
 #	Host "host.foo.bar"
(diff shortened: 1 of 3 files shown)


nward:
A minor fix to make the example match the implementation. The alias then host ordering was from a previous version that I decided to re-order.

dago:
Closing as this is now tracked in #2494

Issue #1637, opened 2016-03-26
collectd don't work with xfsprogs-4.5.0


Whissi:
When trying to compile *collectd* with *xfsprogs-4.5.0* checking for *xfs/xqm.h* fails:

```
configure:17300: checking xfs/xqm.h usability
configure:17300: x86_64-pc-linux-gnu-gcc -c -O2 -pipe -march=ivybridge -mtune=ivybridge -mno-xsaveopt  conftest.c >&5
In file included from /usr/include/xfs/xfs.h:37:0,
                 from /usr/include/xfs/xqm.h:21,
                 from conftest.c:142:
/usr/include/xfs/linux.h:145:9: error: unknown type name 'off64_t'
 typedef off64_t  xfs_off_t;
         ^
configure:17300: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "collectd"
| #define PACKAGE_TARNAME "collectd"
| #define PACKAGE_VERSION "5.5.1.git"
| #define PACKAGE_STRING "collectd 5.5.1.git"
| #define PACKAGE_BUGREPORT ""
| #define PACKAGE_URL ""
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define HAVE_DLFCN_H 1
| #define LT_OBJDIR ".libs/"
| #define LT_MODULE_EXT ".so"
| #define LT_MODULE_PATH_VAR "LD_LIBRARY_PATH"
| #define LT_DLSEARCH_PATH "/lib:/usr/lib:/lib64:/usr/lib64:/usr/local/lib64:/lib32:/usr/lib32:/usr/local/lib32:/lib:/usr/lib:/usr/local/lib:/usr/lib/gcc/x86_64-pc-linux-gnu/5.3.0/32:/usr/lib/gcc/x86_64-pc-linux-gnu/5.3.0:/opt/hp/hpacucli:/usr/lib64/oracle/11.2.0.4/client/lib64:/usr/lib/postgresql:/usr/lib64/postgresql:/usr/lib64/postgresql-9.5/lib64/"
[...]

PR #1638, opened 2016-03-26
Use AC_GNU_SOURCE macro to detect and propagate __GNU_SOURCE macro

Whissi:
*_GNU_SOURCE* is needed for *xfs/xqm.h* since xfsprogs-4.5.0 at least.

Using *AC_GNU_SOURCE* is the better way instead of patching individual files.

Fixes: https://github.com/collectd/collectd/issues/1637

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,6 @@
 dnl Process this file with autoconf to produce a configure script.
 AC_INIT([collectd],[m4_esyscmd(./version-gen.sh)])
+AC_GNU_SOURCE
 AC_CONFIG_SRCDIR(src/)
 AC_CONFIG_HEADERS(src/config.h)
 AC_CONFIG_AUX_DIR([libltdl/config])


rubenk:
Setting _GNU_SOURCE globally potentially changes behaviour for all the code. I'd rather set it in the plugin itself.

rubenk:
I fixed this in 865a6c8

Whissi:
@rubenk: If you got a minute, could you please explain why you don't like *AC_GNU_SOURCE* macro? Your argument
> potentially changes behavior of all code

isn't valid in my eyes. The macro uses a very clean way to enable *_GNU_SOURCE*. It is portable. Now we already have 3 files where we are using such a hack. Probably missing another in */src/liboconfig/parser.c*.

If you are aware of any problems caused by this macro please share. I always appreciate to learn something new.

rubenk:
Hey @Whissi, sure :)

I never said I didn't like the macro, just that there are potential pitfalls in using it.

Defining _GNU_SOURCE globally can, depending on libc version, prefer GNU versions of certain functions instead of BSD. See http://man7.org/linux/man-pages/man7/feature_test_macros.7.html and https://lwn.net/Articles/590381/ for some background. If we want to enable it throughout the whole project, we will need to very carefully check the whole codebase, on all platforms we support, with all libc's we support, that it does the right thing.

If you look at the AC_GNU_SOURCE macro itself, you'll notice that it is obsolete and an alias for AC_USE_SYSTEM_EXTENSIONS. That macro does more than just define _GNU_SOURCE.

As for the actual issue with xfsprogs, I did work around this in Collectd, but I'd rather see it fixed in xfsprogs itself.
They started to use off64_t, which is hidden behind _LARGEFILE64_SOURCE. feature_test_macros(7) says "New programs should not employ this macro; instead _FILE_OFFSET_BITS=64 should be employed." Perhaps they could have done something like
```
#ifdef _LARGEFILE64_SOURCE
typedef off64_t xfs_off_t
#elif _FILE_OFFSET_BITS=64
typedef off_t xfs_off_t
```

I also appreciate learning something new, so if you think I'm wrong on some of the details, let me know :)

rubenk:
I looked into this some more.
We use AC_SYS_LARGEFILE, which sets _FILE_OFFSET_BITS=64.
Based on this, the right off_t type is defined in /usr/include/unistd.h:
```
# ifndef __off_t_defined
#  ifndef __USE_FILE_OFFSET64
typedef __off_t off_t;
#  else
typedef __off64_t off_t;
#  endif
#  define __off_t_defined
# endif
# if defined __USE_LARGEFILE64 && !defined __off64_t_defined
typedef __off64_t off64_t;
#  define __off64_t_defined
# endif
```

PR #1641, opened 2016-03-29
'df' plugin: warn about unrecognized config keys.

coreykosak:
Change 'df' so that instead of silently ignoring keys, it writes a message to the log.

Note: it might at first seem that 'df' is returning an error code on an unrecognized key,
but as far as I can tell, that error is ignored by collectd. Evidence: the following code inside
```cf_read()``` of ```configfile.c``` ignores the result code that bubbled up to ```dispatch_value``` or ```dispatch_block```.

```
for (i = 0; i < conf->children_num; i++)
{
    if (conf->children[i].children == NULL)
        dispatch_value (conf->children + i);
    else
        dispatch_block (conf->children + i);
}
```

I intend to address this in my next PR.

### DIFF
--- a/src/df.c
+++ b/src/df.c
@@ -152,7 +152,7 @@ static int df_config (const char *key, const char *value)
 
 		return (0);
 	}
-
+	WARNING ("df plugin: Ignoring config key '%s'", key);
 	return (-1);
 }
 

### CODE REVIEW

**Comment by mfournier** [comment]
I'm a bit puzzled about this @coreykosak. Without your patch, I *do* get a warning from the configfile parser:
```
<Plugin df>                                                                     
  FooBar true                                                                 
</Plugin>
```
```
[warning] Plugin `df' did not register for value `FooBar'
```

For your warning to kick in, I had to add "FooBar" to the list of keys on line 49.

Am I missing something ?

mfournier:
Ok, after reading your comment in #1642, I now understand better :-)
So I guess you've based this on an outdated fork, or on one of the release branches, as  references to `ReportReserved` got cleaned out a couple of months ago in a870053.

Not sure if it's still worth adding this anymore ?

coreykosak:
Oops. Let's abandon this PR. I was working with code that didn't have this change to ```df.c``` ( https://github.com/collectd/collectd/commit/a8700539e4131ae20af1462d1e6c6ec38a266577 ) and I was bothered that there was no code that would complain if it saw a ```ReportReserved``` in a config file.

You are right that if ```config_keys``` and ```df_config()``` are properly kept in sync (as they are now), then the warning can't happen.

PR #1642, opened 2016-03-29
Modify the -T command line option so it watches for more errors.

coreykosak:
I have really come to appreciate the -T command line option for testing to see if my collectd is working as configured. However, I noticed that, as written, the only errors that get propagated to the exit status of collectd are errors from the ```read``` callback of the configured plugins. In my opinion it would be more valuable if it reported errors that happened in other stages as well. In this way it would be a more faithful end-to-end test.

After this change, the following kinds of errors will cause ```collectd -T``` to exit with an abnormal status:
1. errors reading the config file
2. errors reading ```types.db```
3. errors on ```plugin_init```
4. errors on ```plugin_read``` (as now)
5. errors on ```plugin_shutdown```

**Note:** this PR causes collectd to do more error checking than it did previously. Specifically it bubbles up errors from each plugin's config callback rather than ignoring those errors. It's hypothetically possible that this change could break some existing configurations. Suggestions are welcome on the question of (a) whether it is desirable for collectd to be more strict like this, and (b) if so, how to test the existing plugins to make sure they don't break.

Edit: this paragraph mooted by https://github.com/collectd/collectd/commit/a8700539e4131ae20af1462d1e6c6ec38a266577
~~One thing I found myself was that the ```df``` plugin will return an error in ```df_config()``` if the deprecated option ```ReportReserved``` is in your config, but the rest of collectd will ignore that error. (see https://github.com/collectd/collectd/pull/1641 for an explanation of why this is so). After this change, such a keyword will cause collectd to report an error. Of course it's possible that the correct fix there is to fix ```df``` so that unrecognized keywords in the config file are just warnings and not errors (in other words, maybe ```df_config``` should return 0 and not -1).~~

### DIFF
--- a/src/daemon/collectd.c
+++ b/src/daemon/collectd.c
@@ -332,9 +332,7 @@ static int do_init (void)
 	}
 #endif
 
-	plugin_init_all ();
-
-	return (0);
+	return plugin_init_all ();
 } /* int do_init () */
 
 
@@ -388,8 +386,7 @@ static int do_loop (void)
 
 static int do_shutdown (void)
 {
-	plugin_shutdown_all ();
-	return (0);
+	return plugin_shutdown_all ();
 } /* int do_shutdown */
 
 #if COLLECT_DAEMON
@@ -729,12 +726,19 @@ int main (int argc, char **argv)
 	/*
 	 * run the actual loops
 	 */
-	do_init ();
+	if (do_init () != 0)
+	{
+		ERROR ("Error: one or more plugin init callbacks failed.");
+		exit_status = 1;
+	}
 
 	if (test_readall)
 	{
 		if (plugin_read_all_once () != 0)
+		{
+			ERROR ("Error: one or more plugin read callbacks failed.");
 			exit_status = 1;
+		}
 	}
 	else
 	{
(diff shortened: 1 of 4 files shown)


mfournier:
Yes, I like the idea, thanks for the patch !

I wouldn't question if it's desirable to be more strict or not. If we do have a way to figure out if a given configuration is unfit for execution, it's the current situation that doesn't make sense, right ?

I'm not too worried about breakages neither, for the following reasons:
 - this would be part of a feature release. We can reasonably expect people to take a look at the changelog before upgrading, and maybe having to fix the odd broken config that doesn't load anymore. 
 - the configfile parsing system already enforces pretty consistent configuration files. Errors with types.db should also be a no-go. Errors when running `plugin_init` already prevent proper plugin execution. And errors in `plugin_shutdown` wouldn't block you from running collectd, only from running something like `collectd -T && collectd -f`.
 - if introduced early enough before the release, I'm sure we'll get feedback from users running on master if cases such as the one you encountered with the df plugin pop up (at least for the most popular plugins).

rubenk:
I totally agree, and I'm sure we can handle the fallout if we do this soon. Please merge :)

mfournier:
Great, thanks a lot @coreykosak !

PR #1646, opened 2016-04-01
Cleanup malloc calls

rubenk:


src/cpu.c
```diff
@@ -814,8 +814,7 @@ static int cpu_read (void)
 
 	if (pnumcpu != numcpu || perfcpu == NULL)
 	{
-		if (perfcpu != NULL)
-			free(perfcpu);
+		free(perfcpu);
```
rpv-tomsk: Hey, please review this!

Code is strange:

if (perfcpu == NULL) 
  free(perfcpu);

rubenk: Can you be more specific? calling free on a NULL pointer is perfectly safe.

rpv-tomsk: Hm, I'm looked to manual, it's really safe to pass NULL to free(), I don't know about this.

#define sfree(ptr) \
    do { \
        if((ptr) != NULL) { \
            free(ptr); \
        } \
        (ptr) = NULL; \
    } while (0)

Why when we have macro with not-null check?  Possible, to avoid function call, which is more expensive.
Sorry for confusion.

rubenk: Yeah, sfree doesn't have to do that, but it's been like that for more than 10 years. I doubt avoiding the function call has a measurable performance impact in practice.

octo: > I doubt avoiding the function call has a measurable performance impact in practice.

+1. The important thing about `sfree()` is the `(ptr) = NULL;` bit, i.e. that pointers are guaranteed to be invalid afterwards. If you feel strongly about the `NULL` check, feel free to send a PR.

PR #1650, opened 2016-04-04
Submit FreeBSD disk IO time

kev009:
@delphij I believe these values are roughly equivalent to the values collectd submits for Linux.  *io_time* is mapped to FreeBSD's **sb** and *weighted_io_time* FreeBSD's **tsvc_t** from `iostat -x -I`.

They are interesting because they are effectively monotonic counters from boot.

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -536,7 +536,7 @@ static int disk_read (void)
 	struct gident *geom_id;
 
 	const char *disk_name;
-	long double read_time, write_time;
+	long double read_time, write_time, busy_time, total_duration;
 
 	for (retry = 0, dirty = 1; retry < 5 && dirty == 1; retry++) {
 		if (snap != NULL)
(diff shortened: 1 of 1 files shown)


rubenk:
Hi kev009, thanks for the PR!
Unfortunately the diff is a bit hard to read. Would you mind splitting this out in 2 commits, one that moves submit_io_time out of the #if KERNEL_LINUX block and one that does the actual work?

kev009:
@rubenk amended

rubenk:
I'd still like to track TRIM / UNMAP requests as well, is that something you're willing to look into?

kev009:
@rubenk what are your thoughts on the KPIs to do that?  It does sound interesting to me.

rubenk:
I just recalled the comment made in @octo https://github.com/collectd/collectd/pull/341#issuecomment-19508877. Perhaps we need new types for those.

kev009:
The values @delphij did appear to be desirable counters that are relative to each polling interval, but would like that to be double checked.

Do you known what, in devstat, tracks TRIM?

kev009:
If we can get interesting TRIM data from devstat it is a simple matter of adding it to types.db..

PR #1655, opened 2016-04-06
write_tsdb with metadata

kev009:
Merge conflict resolution and fixes for @rubenk review

PR #1656, opened 2016-04-06
Added MetaDataSet to target_set

kev009:
Address review from @rubenk for https://github.com/collectd/collectd/pull/1106

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -8950,6 +8950,8 @@ Available options:
 
 =item B<TypeInstance> I<String>
 
+=item B<MetaDataSet> I<String> I<String>
+
 Set the appropriate field to the given string. The strings for plugin instance
 and type instance may be empty, the strings for host and plugin may not be
 empty. It's currently not possible to set the type of a value this way.
--- a/src/daemon/meta_data.c
+++ b/src/daemon/meta_data.c
@@ -105,12 +105,15 @@ static meta_entry_t *md_entry_alloc (const char *key) /* {{{ */
   return (e);
 } /* }}} meta_entry_t *md_entry_alloc */
 
-static meta_entry_t *md_entry_clone (const meta_entry_t *orig) /* {{{ */
+/* XXX: The lock on md must be held while calling this function! */
+static meta_entry_t *md_entry_clone_contents (const meta_entry_t *orig) /* {{{ */
 {
   meta_entry_t *copy;
 
-  if (orig == NULL)
-    return (NULL);
+  /* WARNINGS :
+   *  - we do not check that orig != NULL here. You should have done it before.
+   *  - we do not set copy->next. DO NOT FORGET TO SET copy->next IN YOUR FUNCTION
+   */
 
   copy = md_entry_alloc (orig->key);
   if (copy == NULL)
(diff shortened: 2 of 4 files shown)


mfournier:
Thanks for making this move forward @rubenk & @kev009 ! I think it's a great way to address the tagging need. Many thank @ymettier !!

I'm just a bit worried about this patchset after a discussion I had with @beorn- (which worked on the kairosdb PR) on IRC a couple of weeks ago:

```
2016-02-12 15:30:12     mfournier       beorn_: sorry for not getting back to you earlier... What was wro
ng with metadata ? AFAICT it's not used much currently (only by the network plugin I think).
2016-02-12 15:31:08     beorn_  it's racy
2016-02-12 15:31:23     beorn_  and shared globally
2016-02-12 15:31:45     mfournier       ok :-/
2016-02-12 15:31:54     beorn_  it's not "bad"
2016-02-12 15:32:03     beorn_  but it should be clear not to use as tags for metrics
2016-02-12 15:32:14     beorn_  as many TSDB are going down this path 
2016-02-12 15:32:31     beorn_  i.e. my format generates
2016-02-12 15:34:11     beorn_  collectd.<pluginname> then adds tags to it plugin_instance=bleh ds=blah host=blih type=bloh type_instance=bluh
2016-02-12 15:35:16     beorn_  since we use the vl struct
2016-02-12 15:36:29     beorn_  there's a vl->meta attached to it. but in this context you can't mix collectd meta wich are meant for an instance level , and not a datapoint level 
2016-02-12 15:36:59     beorn_  dunno if i'm clear enough :)
2016-02-12 15:41:24     mfournier       beorn_: ok, thanks ! So if I'm following, you'd consider #1106 unsafe, because prone to race conditions, right ?
2016-02-12 15:41:45     beorn_  got to re-read it now that i'm fully aware of how it works
2016-02-12 15:41:46     beorn_  :D
2016-02-12 15:41:57     mfournier       great, thanks !
2016-02-12 15:42:06     beorn_  gimme a sec
2016-02-12 15:47:39     beorn_  performance wise it's suboptimal but it should work
2016-02-12 15:51:32     beorn_  mfournier: i should write an email i think
2016-02-12 15:51:33     beorn_  :)
2016-02-12 15:53:39     mfournier       beorn_: yes, please do ! Or post a comment on the github issue- I don't think Yves's hangs around here, and it would be great to have him included in the discussion.
2016-02-12 15:55:57     mfournier       beorn_: I would really love to have better, ideally generic, arbitrary key=value support for metrics. As you said, many newer TSDBs work this way. I think it's important to have good integration with them.
2016-02-12 15:56:57     beorn_  the tricky part
2016-02-12 15:56:58     beorn_  is 
2016-02-12 15:57:02     beorn_  it worked fine 
2016-02-12 15:57:03     beorn_  BUT
2016-02-12 15:57:07     beorn_  under heavy load
2016-02-12 15:57:11     beorn_  it became racy
2016-02-12 15:57:21     beorn_  and some metrics were missing tags,
2016-02-12 15:57:26     beorn_  or bearing tags from other metrics
2016-02-12 15:57:31     beorn_  other side effect was 
2016-02-12 15:57:59     beorn_  since the wrong thread was freeing recursively data, it was not freeing all the memory 
2016-02-12 15:58:02     beorn_  hence a huge memory leak
```

What do you think ? Have you been able to benchmark collectd with this patchset @kev009 or @ymettier ?

kev009:
@mfournier I wish that was captured in the other PR :(, I will read through the code and see if I spot anything.  For my collectd usage, the client is not sampling at great frequency so performance has never been a concern.

kev009:
@mfournier I really don't see anything controversial in this PR.  If meta_data has problems, they existed prior to @ymettier's changes merged here.  Do you think the critique was aimed at usage https://github.com/collectd/collectd/pull/1655?

dc-mattj:
Any news on this getting merged ? We'd really like to use it without building our own packages

PR #1657, opened 2016-04-06
Use consistent C-style comments

dago:
For older compilers it is necessary to consistently use C-style comments


rubenk:
Thanks for the patch. I wasn't aware that we still had compilers that didn't support C99. Which compiler do you see this on?

dago:
I thought it was related to this issue:

- https://sourceforge.net/p/libpng/bugs/245/

But I figured I could also downgrade to C99 instead of C89. Sorry for the noise.

PR #1671, opened 2016-04-15
fix potential curl plugin curl_slist_append memory leak

BrandonArp:
make sure that we dont leak memory if curl_slist_append returns null

**NOTE: this does not fix the curl_slist_appends in write_http since #1639 addresses those**


BrandonArp:
@rubenk Here's the fix you requested.

rubenk:
Wow, that was fast. Merged into collectd.5.4 as 5b1caeeadcecd798ee4b386b15c17cb7591a8bfe

PR #1672, opened 2016-04-18
write_kafka: do not call depreciated functions unless necessary

pyr:
This is in preparation for allowing compiling/running against recent rdkafka releases found in recent distros.

### DIFF
--- a/src/write_kafka.c
+++ b/src/write_kafka.c
@@ -114,7 +114,8 @@ static int kafka_handle(struct kafka_topic_context *ctx) /* {{{ */
 
         INFO ("write_kafka plugin: created KAFKA handle : %s", rd_kafka_name(ctx->kafka));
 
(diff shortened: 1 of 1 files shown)


ciomaire:
Is that a an error on L118?

```
diff --git a/src/write_kafka.c b/src/write_kafka.c
index 2c3296d..37b3e14 100644
--- a/src/write_kafka.c
+++ b/src/write_kafka.c
@@ -115,7 +115,6 @@ static int kafka_handle(struct kafka_topic_context *ctx) /* {{{ */
         INFO ("write_kafka plugin: created KAFKA handle : %s", rd_kafka_name(ctx->kafka));
 
 #if defined(HAVE_LIBRDKAFKA_LOGGER) && !defined(HAVE_LIBRDKAFKA_LOG_CB)
-        if
         rd_kafka_set_logger(ctx->kafka, kafka_log);
 #endif
     }
```

PR #1491, opened 2015-12-26
sensors plugin: add support for energy and current inputs

dangowrt:
Add support for current and energy values to log INA209, LTC4151, ...

PR #1673, opened 2016-04-19
Add current values from libsensors.

sKuhne:
Four sensors are available with this patch.
Tested with axp209-isa-0000 on BananaPi-R1.

### DIFF
--- a/src/sensors.c
+++ b/src/sensors.c
@@ -132,10 +132,10 @@ static sensors_labeltypes_t known_features[] =
 	{ "2.5V", SENSOR_TYPE_VOLTAGE },
 	{ "2.0V", SENSOR_TYPE_VOLTAGE },
 	{ "12V", SENSOR_TYPE_VOLTAGE },
-	{ "power1", SENSOR_TYPE_POWER }
-	{ "curr1", SENSOR_TYPE_CURRENT }
-	{ "curr2", SENSOR_TYPE_CURRENT }
-	{ "curr3", SENSOR_TYPE_CURRENT }
+	{ "power1", SENSOR_TYPE_POWER },
+	{ "curr1", SENSOR_TYPE_CURRENT },
+	{ "curr2", SENSOR_TYPE_CURRENT },
+	{ "curr3", SENSOR_TYPE_CURRENT },
 	{ "curr4", SENSOR_TYPE_CURRENT }
 };
 static int known_features_num = STATIC_ARRAY_SIZE (known_features);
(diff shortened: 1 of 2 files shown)


src/sensors.c
```diff
@@ -131,6 +133,10 @@ static sensors_labeltypes_t known_features[] =
 	{ "2.0V", SENSOR_TYPE_VOLTAGE },
 	{ "12V", SENSOR_TYPE_VOLTAGE },
 	{ "power1", SENSOR_TYPE_POWER }
+	{ "curr1", SENSOR_TYPE_CURRENT }
```
mfournier: Older gcc versions seem to choke on this: `sensors.c:136: error: expected '}' before '{' token`. See the report from the jenkins build.

mfournier:
Thanks for the update @sKuhne !

I was taking a look at the pending patches to sensors.c, and noticed some overlap with #1491. What do you think ? Would it make sense to merge @dangowrt's patch into yours ?

Also, @rubenk mentions these symbols were added in API version 0x402. Could you please make the patchset safe for older versions ? Thanks !

sKuhne:
@mfournier:
The other patch add one value more.
I can add this to my patch and add the test for older versions. Is this okay for you?

mfournier:
That would be awesome @sKuhne !

As I don't have very interesting hardware (from libsensors' perspective) to test this, it would be great if you can check @dangowrt's patch works as expected.

sKuhne:
I've add #1491 and add the 0x402 check. But I can't test energy.

src/sensors.c
```diff
@@ -64,9 +64,7 @@ static char *sensor_type_name_map[] =
 	"temperature",
 # define SENSOR_TYPE_POWER       3
 	"power",
-# define SENSOR_TYPE_CURRENT     4
-	"current",
-# define SENSOR_TYPE_UNKNOWN     5
```
rubenk: Removing this breaks older api's that don't have SENSOR_TYPE_CURRENT.

sKuhne: I didn't check the complete "#if" part.
This code is only for API below 0x400 and this does not have this feature as I see in other comments. So I have to clean "sensors_labeltypes_t".

rubenk: But you are using SENSOR_TYPE_CURRENT unguarded in known_features[]. If it is not defined, the code breaks. Please check your code against an older version of libsensors.

rubenk: You can see this issue here: https://ci.collectd.org/job/pull-requests-build-on-epel5-x86_64-with-default-toolchain/812/console, it indeed breaks on RHEL5.

clemensg:
I created #2255

rubenk:
Should be fixed by b2b559936dd8e1c476257021add9310c414e4df4

PR #1675, opened 2016-04-21
Add metadata support for write_riemann

jeanpralo:
Hi there,

@pyr @mfournier  following up our conversation from last time on irc here is a pull request to support MetaData in write_riemann plugin.

First time in a while I am writing C so it is probably far from perfect and not sure if what I propose to access all the metadata keys will suit you. I am ready for any feedback :)

I have added the function `meta_data_list_keys()` for the meta_data access. It is because I did not want to break what is currently done and expose `meta_entry_s`. So it basically give us a list of all the meta data keys so we can then use the already existing method without breaking the thread protection around it.
It sounds a bit redundant but I could not think on a better way to do it.

I have tested it and it works fine with riemann as I can see my extra attributes, in order to activate it here is a snippet of my config:

```
<Chain "PostCache">
        <Rule "cpu-send">
               <Match "regex">
                Plugin "^cpu$"

               </Match>
               <Target "set">
                 MetaDataSet "KEY1" "VAL1"
                 MetaDataSet "janus" "pralorantus"
               </Target>
               Target "write"
        </Rule>
       Target "stop"
</Chain>
```

Let me know what you think.

### DIFF
--- a/src/daemon/meta_data.c
+++ b/src/daemon/meta_data.c
@@ -746,3 +746,41 @@ int meta_data_as_string(meta_data_t *md, /* {{{ */
 
   return (0);
 } /* }}} int meta_data_as_string */
+
+int meta_data_list_keys (meta_data_t *md, meta_data_keys_t **keys) /* {{{ */
+{
+  meta_entry_t *e;
+  meta_data_keys_t *tmp;
+
+  if (md == NULL)
+    return (-EINVAL);
+
+  pthread_mutex_lock (&md->lock);
+
+  for (e = md->head; e != NULL; e = e->next)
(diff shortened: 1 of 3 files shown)


jeanpralo:
I have rebased this PR any chance to have a look at it @mfournier and @pyr ?

jeanpralo:
Rebased again to remove conflict @octo @mfournier  do you think you guys could have a look at it ?

Cheers.

faxm0dem:
I though this was merged. What's the plan?

jeanpralo:
I am still keen to get that merge for sure :)

PR #1558, opened 2016-02-24
Curl_xml: Custom names for plugin/plugin instances

rpv-tomsk:
Hi!

Proposed patches adds ability to set custom plugin name and set plugin instance from XML data.
curl_xml is a generic plugin and we found such feature very useful.

In our case, we have XML-response like this:

```
<?xml version="1.0"?>
<statistic>
  <source id="total">
    <errors>4</errors>
    <avg-duration>900</avg-duration>
    <max-duration>1500</max-duration>
  </source>
  <source id="pool1">
    <errors>1</errors>
    <avg-duration>1000</avg-duration>
    <max-duration>1500</max-duration>
  </source>
  <source id="pool2">
    <errors>3</errors>
    <avg-duration>850</avg-duration>
    <max-duration>1000</max-duration>
  </source>
</statistic>
```

From this response and with proposed patches, our collectd produces RRDs:

service-total/service_errors.rrd
service-total/service_avgduration.rrd
service-total/service_maxduration.rrd

service-pool1/service_errors.rrd
service-pool1/service_avgduration.rrd
service-pool1/service_maxduration.rrd

service-pool2/service_errors.rrd
service-pool2/service_avgduration.rrd
service-pool2/service_maxduration.rrd

The config example for this case:

```
<Plugin "curl_xml">
<URL "http://service.example.com/xml/stat">
    Host "service.example.com"
    PluginName "service"
    Instance "all"

    <xpath "/statistic/source">
      Type "service_errors"
      PluginInstanceFrom "@id"
      ValuesFrom "errors/text()"
    </xpath>

    <xpath "/statistic/source">
      Type "service_avgduration"
      PluginInstanceFrom "@id"
      ValuesFrom "avg-duration/text()"
    </xpath>

    <xpath "/statistic/source">
      Type "service_maxduration"
      PluginInstanceFrom "@id"
      ValuesFrom "max-duration/text()"
    </xpath>
</URL>
</Plugin>
```

PR #1681, opened 2016-04-24
Allow custom plugin name for 'curl', 'memcachec', 'tail' and 'tail_csv' plugins.

rpv-tomsk:
Like added in  #1558 and #1563 , added option 'PluginName' to other Collectd plugins:

tail plugin: Allow custom plugin name for file instances
tail_csv: Allow custom plugin name for file instances
curl plugin: Allow custom plugin name for each Page block
memcachec: Allow custom plugin name for each Page block

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3542,6 +3542,7 @@ Synopsis of the configuration:
    <Page "plugin_instance">
      Server "localhost"
      Key "page_key"
+     PluginName "plugin_name"
      <Match>
        Regex "(\\d+) bytes sent"
        DSType CounterAdd
@@ -3568,6 +3569,11 @@ B<Page> block.
 =item B<Key> I<Key>
 
 When connected to the memcached server, asks for the page I<Key>.
+
+=item B<PluginName> I<PluginName>
+
+Use I<PluginName> as the plugin name when submitting values.
+Defaults to 'memcachec'.
 
 =item E<lt>B<Match>E<gt>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1611,6 +1611,7 @@ finance page and dispatch the value to collectd.
 
   <Plugin curl>
     <Page "stock_quotes">
+      PluginName "quotes"
       URL "http://finance.google.com/finance?q=NYSE%3AAMD"
       User "foo"
       Password "bar"
@@ -1641,6 +1642,11 @@ string argument to the B<Page> block is used as plugin instance.
 The following options are valid within B<Page> blocks:
 
 =over 4
+
+=item B<PluginName> I<PluginName>
+
+Use I<PluginName> as the plugin name when submitting values.
+Defaults to 'curl'.
 
 =item B<URL> I<URL>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1611,7 +1611,7 @@ finance page and dispatch the value to collectd.
 
   <Plugin curl>
     <Page "stock_quotes">
-      PluginName "quotes"
+      Plugin "quotes"
       URL "http://finance.google.com/finance?q=NYSE%3AAMD"
       User "foo"
       Password "bar"
@@ -1643,9 +1643,9 @@ The following options are valid within B<Page> blocks:
 
 =over 4
 
-=item B<PluginName> I<PluginName>
+=item B<Plugin> I<Plugin>
 
-Use I<PluginName> as the plugin name when submitting values.
+Use I<Plugin> as the plugin name when submitting values.
 Defaults to 'curl'.
 
 =item B<URL> I<URL>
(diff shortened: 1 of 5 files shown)


rpv-tomsk:
Hi @toni-moreno ,

>Could be a great idea recover the this change to use in all general purpose pugins.

Stay tuned.

----

If someone knows about other generic plugins I missed in this work, please inform me, I will update this PR.

rpv-tomsk:
Hi @toni-moreno ,

>Could be a great idea recover the this change to use in all general purpose pugins.

Stay tuned.

----

If someone knows about other general purpose plugins I missed in this work, please inform me, I will update this PR.

mfournier:
I'm a bit concerned about adding a per-plugin config option to change the plugin name. With the filter chains system, we already have a generic way to rename the host, plugin, plugin_instance and type_instance on the fly.

Agreed it's not as simple as a one-line configuration option, but it works for all current and future plugins, without adding any code.

Am I missing something ? What's the motivation behing this change ?

rpv-tomsk:
Existence of generic way for renaming is good, but for 'general purpose' plugins configuration by per-plugin option is _much_ simplier. 

As for me, as sysadm, I _never_ touched Collectd 'filters' configuration at all. Its usage (for renaming), as for me (again), is like using iptables to redirect HTTP port 80 when 'webserver' listens 65079 and that value is unchangeable.

Currently, other general-purpose plugins allows to specify PluginName: 'exec'; 'java' ('GenericJMX'), 'python', 'perl'  bindings. And there is no requirement to rename fixed 'exec' / 'java' / 'python' / 'perl' plugin name to required value via filters - that plugins just works. That is simple and simply works - that is what expected by people.

I see no problems about adding PluginName option to curl*, tail* and memcachec plugins (and to possible future general-purpose plugins too).

toni-moreno:
Hi @mfournier I agree with @rpv-tomsk . 

Anyway 2 years ago I was playing with filters and I was not able to configure filters to change plugin name for combined plugins and instanceid.  how can be done now?

I think also this code could be easily maintainable , since there is no much generic purpose plugins, and there is not much code to change.

I can help if you need. 

Thank you very much.

tokkee:
On another note, there are now a bunch of PRs all related to this topic. @rpv-tomsk could you open a separate issue as a feature request for this and collecting all information (and conversations) in one place?

octo:
I tend to agree with @mfournier here: changing the plugin name is not essential, since in all the rather generic plugins you can control large parts of the identifier already. If you have a good use case, you can change the *plugin* with generic tools (i.e. filter chains) that work for *all* metrics.

rpv-tomsk:
>changing the plugin name is not essential

Completely disagreed. Did you look into #1558 and #1563, especially into https://github.com/collectd/collectd/pull/1563/commits/0dcb6a619165d4ae16d19773399c8618b37e2fc7 ?

>If you have a good use case, you can change the plugin with generic tools (i.e. filter chains) that 
>work for all metrics.

* There is MANY good use-cases. These (with patch proposed) plugins are 'generic', so people can use them for gathering metrics from different part of their systems. That is their common use, they are implemented for this, so there is many use cases exists. For example, some metrics from curl_json can be sent as 'phpfpm' plugin name and some as another plugin name. I use  'Tail' plugin for gathering metrics which are sent as 'postfix' and 'apache' plugin names.

*  That looks too complex to use filters for these simple tasks. It is much easier to put one directive to set 'PluginName' than write complex filter with matches and sets. Filters is not convenient for this as configuration split is inconvenient in practice.  

//As for me, I do not use 'filters' in none of my servers, and I have 50+ of them. 
//I think many peoples does not use that too. 

* The possibility to set 'plugin instance' value from the values obtained by plugin is required.
And the possibility to set 'plugin name' makes things (configuration) holistic.

So, proposed possibility to set 'PluginName' is very useful, powerful and easy to use.

Also this feature is easily implemented and does not require efforts to support it in future.

octo:
You're right in that there is no "service" label in *collectd* and *plugin* can kind of fill that position.

Please use the config option `Plugin` for this purpose everywhere. Open an umbrella issue, like @tokkee suggested, to keep track of all ongoing work in this direction and then create one PR per plugin, referencing that tracking issue.

Thanks and best regards,
—octo

rpv-tomsk:
>and then create one PR per plugin, referencing that tracking issue.

Hi! I'm going to continue this work. 
Which branch should I use as PR target?

octo:
Hi @rpv-tomsk, thanks for looking into this! Please open PRs implementing new features against the *master* branch.

rpv-tomsk:
Renamed 'PluginName' option to unified 'Plugin' form.

rpv-tomsk:
Hi guys! 

There is some buildbot problem, the CI check has failed for this PR. Please take care.

Thanks!

rpv-tomsk:
Rebased to current master branch.

rpv-tomsk:
ping

rpv-tomsk:

rpv-tomsk:

PR #1686, opened 2016-04-27
Zfs arc

rubenk:


mfournier:
Merging this, at it's obviously trivial and useful. Thanks !

I'm just a bit worried the plugin would emit a lot of warning logs if legacy/future versions of ZFS don't have all the metrics we're looking for. Maybe we should revisit the plugin to make it a bit more tolerant regarding missing stats. NB: it's a general concern, not specific to this change.

rubenk:
Thanks!

You're absolutely right. In this case I tested on FreeBSD, Solaris and with ZoL and they have all these stats. It would be great if we didn't have to do fixes like ec371b03 in the future though, that one filled my logs real quickly. Also, log rate limiting ;)

Issue #1238, opened 2015-08-24
Maximum field length (63) issue


SavitaKulkarni:
Hi,

We are using GenericJMX plugin to fetch java related statistics from our application.
In our case mbeans are created dynamically considering many configurations.
Most of the times mbeans are created with name > 63 chars. In our case mbean name is the plugin type in collectd.
As in collectd there is a limit on max field length to 63, these names are truncated.
Is it possible to eliminate/increase this max limit for field names in collectd?
Or can we have option to abbreviate field names like we do in /var/log/messages? 

Regards,
Savita

druchoo:
The character limitation is global. 63 characters just isn't enough in this day an age and really needs to increased. +1

mschenck:
I was able to extend this limit by increasing [DATA_MAX_NAME_LEN](https://github.com/collectd/collectd/blob/master/src/daemon/plugin.h#L42), but after a short period of time (on that new build) metrics would stop being written.  At the time, I was using the [ write_graphite](https://www.collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_write_graphite), a very limited set of system stats collection(cpu, memory, disk, df, etc), and the [statsd](https://www.collectd.org/wiki/index.php/Plugin:StatsD) plugin.

The only metrics that were exceeding the 63 characters were those written to the statsd plugin, but all metrics would cease to submit.  So, I believe there are additional dependencies on this limit, possibly/hopefully only the write_graphite plugin, but I haven't had time to investigate any further just yet.

druchoo:
@mschenck I was tempted to do the same and change DATA_MAX_NAME_LEN, but it seemed too easy and I didn't have time to test. I guess you have confirmed my suspicions. Hopefully it's fairly simple and someone can let us know what exactly needs to be changed.

SavitaKulkarni:
Hi,

@mschenck  I did changes of increasing max field length to 264 in my setup with write_graphite plugin no issues observed so far.

Regards,
Savita

mschenck:
@SavitaKulkarni have you tried to write something that hits/exceeds this limit?  I believe that might've been happening for me. 


 When I stick with the default `DATA_MAX_NAME_LEN` then the name is just truncated and everything continues on.

My issues started when I increased `DATA_MAX_NAME_LEN` and the name of the metric submitted to the `statsd` plug-in exceeded this value.  We'd have intermittent (eventually total) loss of statsd metrics being submitted, until collectd was restarted.

[...]

PR #1690, opened 2016-04-30
make DATA_MAX_NAME_LEN a compile time tunable

fr0stbyte:
This PR addresses the issue described in #1238 , #966, #872 and ( I think ) proposes a more flexible solution than #1120 

This makes the DATA_MAX_NAME_LEN a compile time tunable by exposing it in the ```./configurate``` script

Example usage:
```
./configure --with-data-max-name-len=1024
```

The default value remains 64 chars, but allows users that need larger sizes to easily change the value without the need of patching.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1352,6 +1352,20 @@ AC_ARG_WITH(useragent, [AS_HELP_STRING([--with-useragent@<:@=AGENT@:>@], [User a
 
 # }}}
 
+# --with-data-max-name-len {{{
+AC_ARG_WITH(data-max-name-len, [AS_HELP_STRING([--with-data-max-name-len@<:@=VALUE@:>@], [Maximum length of data buffers])],
+[
+    if test "x$withval" != "x" && test $withval -gt 0
+    then
+        AC_DEFINE_UNQUOTED(DATA_MAX_NAME_LEN, [$withval], [Maximum length of data buffers])
+    else
+        AC_MSG_ERROR([DATA_MAX_NAME_LEN must be a positive integer -- $withval given])
+    fi
+],
+[   AC_DEFINE(DATA_MAX_NAME_LEN, 128, [Maximum length of data buffers])]
+)
+# }}}
+
 have_getfsstat="no"
 AC_CHECK_FUNCS(getfsstat, [have_getfsstat="yes"])
 have_getvfsstat="no"


fr0stbyte:
Updated branch to the latest changes from upstream.

@octo the configurate script will provide a default value for DATA_MAX_NAME_LEN if the flag was not used. Is the ifndef wrapper around the define necessary in that case ?

rubenk:
@fr0stbyte thanks you very much for your PR!

Indeed, if we let ./configure set this in config.h then b432968 becomes obsolete.

configure.ac
```diff
@@ -1329,6 +1329,20 @@ AC_ARG_WITH(useragent, [AS_HELP_STRING([--with-useragent@<:@=AGENT@:>@], [User a
 
 # }}}
 
+# --with-max-data-len {{{
+AC_ARG_WITH(max-data-len, [AS_HELP_STRING([--with-max-data-len@<:@=VALUE@:>@], [Maximum length of data buffers])],
+[
+    if test "x$withval" != "x" && test $withval -gt 0
```
rubenk: I haven't checked, but how portable is test -gt?
I'm fine with leaving this check out. If a user decided to pass -1 to this, it's his own fault.

configure.ac
```diff
@@ -1329,6 +1329,20 @@ AC_ARG_WITH(useragent, [AS_HELP_STRING([--with-useragent@<:@=AGENT@:>@], [User a
 
 # }}}
 
+# --with-max-data-len {{{
+AC_ARG_WITH(max-data-len, [AS_HELP_STRING([--with-max-data-len@<:@=VALUE@:>@], [Maximum length of data buffers])],
```
rubenk: minor nitpick, --with-data-max-name-len matches better what it sets.

configure.ac
octo: I think `test -gt` is in POSIX and should be very portable.

configure.ac
octo: +1

fr0stbyte:
@rubenk @octo thanks for the review, I've updated the PR

rubenk:
Thanks for the fast response.
I have just two more comments and then I think we're good to go.

- We changed the default in https://github.com/collectd/collectd/commit/6fd0d2574fc386c2724c8307d33b12529b30048b to 128, could you update accordingly?
- Could you change the commit title to something like 'make DATA_MAX_NAME_LEN configurable'?   
'configurate' doesn't mean the same thing as 'configure'. That will make it more obvious what it does in the changelog.

fr0stbyte:
@rubenk all valid points, addressed.

PR #1693, opened 2016-05-02
Processes: add config option CountThreadState with Linux support

ndiller-janestreet:
The name of the processes plugin suggests that it only counts top level processes, not individual threads, in statistics it emits.  However, for some stats, and for processes.ps_state in particular, including threads in the count can be desirable, or even "obvious".  Indeed, judging from the commit history there have been some back-and-forth changes in the past, with divergent behavior across platforms.

This patch introduces a configuration option CountThreadState, and associated changes to the Linux platform code, to toggling ps_state to count either only processes, or both processes and threads.  Since this is the second use case for walking a subset of Linux's /proc tree, I factored out the code from  common to both

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -224,6 +224,7 @@ typedef struct procstat
 static procstat_t *list_head_g = NULL;
 
 static _Bool report_ctx_switch = 0;
+static _Bool count_thread_state = 0;
 
 #if HAVE_THREAD_INFO
 static mach_port_t port_host_self;
@@ -589,6 +590,10 @@ static int ps_config (oconfig_item_t *ci)
 		{
 			cf_util_get_boolean (c, &report_ctx_switch);
 		}
+		else if (strcasecmp (c->key, "CountThreadState") == 0)
+		{
+			cf_util_get_boolean (c, &count_thread_state);
+		}
 		else
 		{
 			ERROR ("processes plugin: The `%s' configuration option is not "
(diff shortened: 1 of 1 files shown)


octo:
Hi @ndiller-janestreet,

thank you very much for your patch! Sorry for getting to this so late. Please don't rebase your branch just yet, that makes reviewing quite hard. I'll leave some review comments inline.

Best regards,
—octo

src/processes.c
```diff
@@ -1001,6 +923,126 @@ static procstat_t *ps_read_io (long pid, procstat_t *ps)
...
+
+static void ps_state_callback (void *ctx, char *buffer, int buflen)
+{
+	taskstate_counter_t *counter;
+	char *fields[2];
+	int   numfields;
+
+	counter = (taskstate_counter_t *) ctx;
+
+	if (strncmp (buffer, "State:", 6) != 0)
```
octo: Please use `strlen("State:")` instead of `6`.

src/processes.c
```diff
@@ -1001,6 +923,126 @@ static procstat_t *ps_read_io (long pid, procstat_t *ps)
...
+		&& strncmp (buffer, "nonvoluntary_ctxt_switches", 26) != 0)
+		return;
+
+	numfields = strsplit (buffer, fields, STATIC_ARRAY_SIZE (fields));
+	if (numfields < 2)
+		return;
+
+	errno = 0;
+	endptr = NULL;
+	tmp = (derive_t) strtoll (fields[1], &endptr, /* base = */ 10);
```
octo: Please use `parse_value (fields[1], &value, DS_TYPE_DERIVE)` instead of `strtoll()`.

src/processes.c
```diff
@@ -1147,14 +1189,26 @@ static int ps_read_process (long pid, procstat_t *ps, char *state)
 
 	if ( report_ctx_switch )
 	{
-		if ( (ps_read_tasks_status(pid, ps)) == NULL)
+		taskswitch_counter_t counter;
```
octo: Please use a struct initializer:

```c
taskswitch_counter_t counter = { 0 };
```

src/processes.c
```diff
@@ -1792,12 +1846,8 @@ static int ps_read (void)
 /* #endif HAVE_THREAD_INFO */
 
 #elif KERNEL_LINUX
-	int running  = 0;
-	int sleeping = 0;
-	int zombies  = 0;
-	int stopped  = 0;
-	int paging   = 0;
-	int blocked  = 0;
+	taskstate_counter_t counter;
```
octo: Dito:

```c
taskstate_counter_t counter = { 0 };
```

src/processes.c
```diff
@@ -1001,6 +923,126 @@ static procstat_t *ps_read_io (long pid, procstat_t *ps)
 	return (ps);
 } /* procstat_t *ps_read_io */
 
+static int ps_iter_dir (const char *dirname, const char *iterfile, void (*callback)(void *, char *, int), void *ctx)
+{
+	DIR           *dh;
+	struct dirent *ent;
+	char           filename[64];
```
octo: Please use `PATH_MAX` instead of `64` here.

src/processes.c
```diff
@@ -1001,6 +923,126 @@ static procstat_t *ps_read_io (long pid, procstat_t *ps)
...
+	FILE          *fh;
+	char           buffer[1024];
+
+	if ((dh = opendir (dirname)) == NULL)
+	{
+		DEBUG ("Failed to open directory `%s'", dirname);
```
octo: Should these really be `DEBUG()`? If so, please add a comment .

src/processes.c
```diff
@@ -1147,14 +1189,26 @@ static int ps_read_process (long pid, procstat_t *ps, char *state)
 
 	if ( report_ctx_switch )
 	{
-		if ( (ps_read_tasks_status(pid, ps)) == NULL)
+		taskswitch_counter_t counter;
+		char dirname[64];
```
octo: Dito: s/64/PATH_MAX/

ndiller-janestreet:
octo, i think i've done justice to your comments.  i'm not entirely sure about the change from DEBUG -> ERROR.  maybe DEBUG would be better since many times these cases represent a (inherent) race condition, rather than an issue with the code.

thanks for reviewing!

NATE

octo:
Hi @ndiller-janestreet, thanks for following up! It appears that there are merge conflicts and prevents the build bots from doing their thing. Could you rebase this on *master* please?

Thanks! —octo

ndiller-janestreet:
Sorry it took me so long, got pulled into troubleshooting some stuff.  I reconciled my changes with master, tested, and used git merge.  If you would prefer git rebase, I'd be fine doing that too.

PR #1694, opened 2016-05-03
Adding mysql Open Table stats. Fixes #979

### LINKED ISSUE #979: MySQL - Open table statistics
It would be helpful to monitor open table stats in MySQL using the plugin.

The following statistics are returned by show status, which the plugin currently calls, and could be sent:

Open_table_definitions - GAUGE
Open_tables - GAUGE
Opened_table_definitions - Counts up over time, resets on server start
Opened_tables - Counts up over time, resets on server start

Having positive values for opened_tables in particular is often indicative of a mistuned table cache.

(this is meant as a feature request FYI)

nickrobinson:
This pull request adds support for [open_tables](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Open_tables), [open_table_definitions](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Open_table_definitions), [opened_tables](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Opened_tables), and [opened_table_definitions](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Opened_table_definitions)

I have tested locally with mysql:5.7.12

Example output:
```
{"values":[105],"dstypes":["gauge"],"dsnames":["value"],"time":1462236936.270,"interval":10.000,"host":"192.168.99.100","plugin":"mysql","plugin_instance":"foo","type":"mysql_open_table_definitions","type_instance":"open"},{"values":[114],"dstypes":["gauge"],"dsnames":["value"],"time":1462236936.270,"interval":10.000,"host":"192.168.99.100","plugin":"mysql","plugin_instance":"foo","type":"mysql_open_tables","type_instance":"open"},{"values":[105],"dstypes":["derive"],"dsnames":["value"],"time":1462236936.270,"interval":10.000,"host":"192.168.99.100","plugin":"mysql","plugin_instance":"foo","type":"mysql_opened_table_definitions","type_instance":"opened"},{"values":[121],"dstypes":["derive"],"dsnames":["value"],"time":1462236936.270,"interval":10.000,"host":"192.168.99.100","plugin":"mysql","plugin_instance":"foo","type":"mysql_opened_tables","type_instance":"opened"}
```

### DIFF
--- a/src/mysql.c
+++ b/src/mysql.c
@@ -904,6 +904,20 @@ static int mysql_read (user_data_t *ud)
 				counter_submit ("mysql_sort", "scan", val, db);
 
 		}
+        else if (strncmp (key, "Open_", strlen ("Open_")) == 0)
+        {
+            if (strcmp (key, "Open_table_definitions") == 0)
+                gauge_submit ("mysql_open_table_definitions", "open", val, db);
+            else if (strcmp (key, "Open_tables") == 0)
+                gauge_submit ("mysql_open_tables", "open", val, db);
+        }
+        else if (strncmp (key, "Opened_", strlen ("Opened_")) == 0)
+        {
+            if (strcmp (key, "Opened_table_definitions") == 0)
+                counter_submit ("mysql_opened_table_definitions", "opened", val, db);
+            else if (strcmp (key, "Opened_tables") == 0)
+                counter_submit ("mysql_opened_tables", "opened", val, db);
+        }
 	}
 	mysql_free_result (res); res = NULL;
 
--- a/src/types.db
+++ b/src/types.db
@@ -126,6 +126,10 @@ mysql_innodb_row_lock   value:DERIVE:0:U
 mysql_innodb_rows       value:DERIVE:0:U
 mysql_locks             value:DERIVE:0:U
 mysql_log_position      value:DERIVE:0:U
+mysql_open_table_definitions        value:GAUGE:0:U
+mysql_open_tables                   value:GAUGE:0:U
+mysql_opened_table_definitions      value:DERIVE:0:U
+mysql_opened_tables                 value:DERIVE:0:U
 mysql_octets            rx:DERIVE:0:U, tx:DERIVE:0:U
 mysql_select            value:DERIVE:0:U
 mysql_sort              value:DERIVE:0:U


ekilby:
The values here look like what I'd expect, the indent seems inconsistent with the preceding lines.

nickrobinson:
Should I use something like 'mysql_opened_table_def'. While descriptive using 'mysql_opened_table_definitions' is going to cause me to shift every type definition which leads me to think this is a bit too long.

rubenk:
@rpv-tomsk you might be interested in reviewing this?

rpv-tomsk:
I put this patch to collectd on my production server with 3 MySQL instances. 

I have added `mysql_open_files` and `mysql_opened_files` also.

At first look, `*_opeN_*` metrics is much more useful than `*_openeD_*`, but, for example, added `mysql_open_files` also unable to help us detect `Too many open files` problem with server due to "Also, the count does not include files that storage engines open using their own internal functions rather than asking the server level to do so." (cite from https://dev.mysql.com/doc/refman/5.5/en/server-status-variables.html#statvar_Open_files).

Each metric should have some kind of practical sense, but currently I don't know how to use these new  metrics (especially `*_openeD_*`) and what should be tuned in MySQL as result of their viewing. @nickrobinson, how to use these new metrics?


The main goal of #979 is to monitor table cache.

> Having positive values for opened_tables in particular is often indicative of a mistuned table cache.

The 'show global status' output  also has metrics which is more suitable for this, and which should be added to collectd (maybe instead of proposed `*_open_*` metrics):

Table_open_cache_hits             
Table_open_cache_misses          
Table_open_cache_overflows 

What do you think?

ekilby:
@rpv-tomsk my concern with the Table_open_cache_* variables is that they are only available in 5.6 in higher as far as I can tell.  Most of the variables published via this plugin are very backwards compatible, so I'd at least leave the *_open_* metrics alongside.  I agree that those cache variables seem useful though.

rpv-tomsk:
> my concern with the Table_open_cache_* variables is that they are only available in 5.6 in higher as far as I can tell. 

That is absolutely correct, they are available since MySQL version 5.6.6.

> Most of the variables published via this plugin are very backwards compatible, so I'd at least leave the open metrics alongside. 

What is the reason to support outdated MySQL versions in new Collectd versions? :-)

>Each metric should have some kind of practical sense, but currently I don't know how to use these new metrics 
>(especially *_openeD_*) and what should be tuned in MySQL as result of their viewing. 

I don't like adding excessive and possible unnecessary metrics, but other people has other requirements, of course. These metrics can be useful for them, so I would prefer things to be configurable. :-)

As sysadm, I would like to see `cache variables` + `open files` without all rest `opeN` and `openeD` metrics.  As for me, that will be most useful combination without excessive and unnecessary (as for me, again) metrics.

I have just added metrics based on `Table_open_cache_*` variables and will put charts tomorrow to continue this discussion more specifically.

Also, please look into #1722 for another idea about adding `StatusVariable`, `StatusVariablePrefix` options to `mysql` plugin. Maybe adding these options will solve this issue?

ekilby:
See https://plumbr.eu/blog/io/most-popular-relational-databases as of Dec 2015 (couldn't find something comparable and newer) but 5.5 and 5.6 versions were still quite popular.

PR #1696, opened 2016-05-03
exec plugin : kill correct pid on fdopen failure

ciomaire:
Hi there,

hopefully I havn't misunderstood somthing here - but it looks like the wrong pid is been killed during error handling in the NotificationExec code.

Al far as I can see, pl->pid is always 0 at this line : https://github.com/collectd/collectd/blob/master/src/exec.c#L745

which means SIGTERM "is sent to every process in the process group of the calling process"  (man 2 kill)

Thanks.

### DIFF
--- a/src/exec.c
+++ b/src/exec.c
@@ -742,8 +742,7 @@ static void *exec_notification_one (void *arg) /* {{{ */
     char errbuf[1024];
     ERROR ("exec plugin: fdopen (%i) failed: %s", fd,
         sstrerror (errno, errbuf, sizeof (errbuf)));
-    kill (pl->pid, SIGTERM);
-    pl->pid = 0;
+    kill (pid, SIGTERM);
     close (fd);
     sfree (arg);
     pthread_exit ((void *) 1);


mfournier:
This looks legit, but I'm incapable of drawing a conclusion just by reading the code :-/
Do you have any evidence ? Are you able to add a statement such as `DEBUG("%i %i", pid, pl->pid);` and force collectd to go through this error code path ?

In any case, thanks for tracking this down, @ciomaire !

ciomaire:
@mfournier I've done some more tests with that extra debug enabled, pl->pid is 0 in all the testing I've done.

A comment earlier in exec.c indicates that that field is unused when handling notifications : https://github.com/collectd/collectd/blob/master/src/exec.c#L55

Also fork_child() called a few lines before this error handling block would have returned -1 if pl->pid is not 0
https://github.com/collectd/collectd/blob/master/src/exec.c#L404

mfournier:
Great, thanks for checking and for the additional explanations @ciomaire ! I merged your patch as fbb3081.

PR #1700, opened 2016-05-07
 Tail plugin: DSType Latency

rpv-tomsk:
This PR proposes new DSType "Latency" in 'tail' plugin and allows to calculate percentiles and 'rate by response time'. Useful for webserver response time monitoring.

Superseedes #1612.

example

```
<Plugin "tail">
    <File "/var/log/nginx/apache1-time.log">
        Instance backend1
        <Match>
            Regex ".*"
            DSType "DeriveInc"
            Type "requests"
            Instance "requests"
        </Match>
        <Match>
            Regex "^\\S+ \"([0-9.]+)\""
            DSType "Latency"
            Type "response_time"
            #LatencyPercentileType "response_type"
            LatencyPercentile 5
            LatencyPercentile 10
            LatencyPercentile 30
            LatencyPercentile 50
            LatencyPercentile 70
            LatencyPercentile 90
            LatencyPercentile 95
            LatencyRateType "http_rate"
            LatencyRate 0.001 0.100
            LatencyRate 0.101 0.200
            LatencyRate 0.201 0.500
            LatencyRate 0.501 1.000
            LatencyRate 1.001 2.000
            LatencyRate 2.001 5.000
            LatencyRate 5.001 0
        </Match>
    </File>
</Plugin>
```


mfournier:
I like the idea a lot ! Let me test this and come back with some feedback.

rpv-tomsk:
Hi Marc!

I would continue this work applying appropriate changes to 'statsd' plugin, if what will be required. Please inform me if this work is needed.

src/utils_latency.c
```diff
@@ -272,4 +272,79 @@ cdtime_t latency_counter_get_percentile (latency_counter_t *lc, /* {{{ */
...
+  if (lower_bin >= HISTOGRAM_NUM_BINS)
+    lower_bin = HISTOGRAM_NUM_BINS - 1;
+
+  if (upper_bin >= HISTOGRAM_NUM_BINS) {
+    upper_bin = HISTOGRAM_NUM_BINS - 1;
+    upper = 0;
+  }
+
+  sum = 0;
+  for (i = lower_bin; i <= upper_bin; i++)
```
octo: Please inline the declaration of `i`:

```c
for (size_t i = 0; …)
```

src/utils_latency.c
```diff
@@ -272,4 +272,79 @@ cdtime_t latency_counter_get_percentile (latency_counter_t *lc, /* {{{ */
...
+  {
+    sum += lc->histogram[i];
+  }
+
+  p = ((double)lower - (double)(lower_bin + 0) * (double)lc->bin_width - (double)DOUBLE_TO_CDTIME_T(0.001)) / (double)lc->bin_width;
```
octo: This needs a comment, and potentially some temporary variables, to become readable.

I assume that what you're doing is correcting for `lower` not falling neatly onto a bin boundary. I propose to calculate this as follows:

```c
/* Approximate ratio of requests below "lower" */
double lower_bin_boundary = (double) (lower_bin * lc->bin_width);
double ratio = (((double) lower) - lower_bin_boundary) / ((double) lc->bin_width);
/* Correct "sum" accordingly. */
sum -= ratio * lc->histogram[lower_bin];
```

If the `DOUBLE_TO_CDTIME_T(0.001)` is really needed for a reason I'm not seeing right now, please document that too. I think that maybe using `((double) (lower-1))` might actually be the correct way to calculate this.

src/utils_latency.c
```diff
@@ -272,4 +272,79 @@ cdtime_t latency_counter_get_percentile (latency_counter_t *lc, /* {{{ */
...
+  sum -= p * lc->histogram[lower_bin];
+
+  if (upper && upper < (upper_bin + 1) * lc->bin_width)
+  {
+    // p = ((upper_bin + 1) * bin_width - upper ) / bin_width;
+    p = ((double)(upper_bin + 1) * (double)lc->bin_width - (double)upper) / (double)lc->bin_width;
```
octo: Dito. Suggested code:

```c
/* Approximate ratio of requests above "upper" */
double upper_bin_boundary = (double) ((upper_bin+1) * lc->bin_width);
double ratio = (upper_bin_width - ((double) upper)) / ((double) lc->bin_width);
/* Correct "sum" accordingly. */
sum -= ratio * lc->histogram[upper_bin];
```

src/utils_latency.c
```diff
@@ -272,4 +272,79 @@ cdtime_t latency_counter_get_percentile (latency_counter_t *lc, /* {{{ */
...
+    sum -= p * lc->histogram[upper_bin];
+  }
+  return sum / (CDTIME_T_TO_DOUBLE (now - lc->start_time));
+
```
octo: Please move empty line above the `return` statement.

src/utils_latency.c
```diff
@@ -272,4 +272,79 @@ cdtime_t latency_counter_get_percentile (latency_counter_t *lc, /* {{{ */
...
+  if (lower < 1) {
+    //sum += lc->zero;
+    //lower = 1;
+    return (0);
+  }
+
+  if (upper && (upper < lower))
+    return (0);
+
+  /* A latency of _exactly_ 1.0 ms should be stored in the buffer 0 */
```
octo: s/should be/is/

I know this is a copy'n'paste line, sorry. Please feel free to correct this elsewhere in the file, too.

src/utils_latency.c
```diff
@@ -272,4 +272,79 @@ cdtime_t latency_counter_get_percentile (latency_counter_t *lc, /* {{{ */
...
+  sum = 0;
+  for (i = lower_bin; i <= upper_bin; i++)
+  {
+    sum += lc->histogram[i];
+  }
+
+  p = ((double)lower - (double)(lower_bin + 0) * (double)lc->bin_width - (double)DOUBLE_TO_CDTIME_T(0.001)) / (double)lc->bin_width;
+  sum -= p * lc->histogram[lower_bin];
+
+  if (upper && upper < (upper_bin + 1) * lc->bin_width)
```
octo: I don't think the `upper < (upper_bin + 1) * lc->bin_width` can ever match. Turn is into an assertion instead?

src/utils_latency_config.c
```diff
@@ -0,0 +1,163 @@
...
+ */
+
+#include "collectd.h"
+#include "common.h"
+#include "utils_latency_config.h"
+
+int latency_config_add_percentile (const char *plugin, latency_config_t *cl,
+    oconfig_item_t *ci)
+{
+  if ((ci->values_num != 1) || (ci->values[0].type != OCONFIG_TYPE_NUMBER))
```
octo: Please use `cf_util_get_double()` instead of doing this check yourself.

src/utils_latency_config.c
```diff
@@ -0,0 +1,163 @@
...
+      ci->values[1].value.number <= ci->values[0].value.number)
+  {
+    ERROR ("%s plugin: MIN must be less than MAX in \"%s\".",
+             plugin, ci->key);
+    return (-1);
+  }
+
+  if (ci->values[0].value.number < 0.001)
+  {
+    ERROR ("%s plugin: MIN must be greater or equal to 0.001 in \"%s\".",
```
octo: Why? I don't that we need this limit, but if we do, it needs to be documented in the manpage.

src/utils_latency_config.c
```diff
@@ -0,0 +1,163 @@
...
+
+  return (0);
+} /* int latency_config_add_rate */
+
+
+int latency_config_copy (latency_config_t *dst, const latency_config_t src)
+{
+  /* Copy percentiles configuration */
+  dst->percentile_num = src.percentile_num;
+  dst->percentile = malloc(sizeof (*dst->percentile) * (src.percentile_num));
```
octo: Please use `calloc()` instead. Here and blow.

src/utils_latency_config.c
```diff
@@ -0,0 +1,163 @@
...
+  }
+  cl->rates = tmp;
+  cl->rates[cl->rates_num * 2] = lower;
+  cl->rates[cl->rates_num * 2 + 1] = upper;
+  cl->rates_num++;
+
+  return (0);
+} /* int latency_config_add_rate */
+
+
```
octo: Nit: duplicate blank line.

src/utils_latency_config.c
```diff
@@ -0,0 +1,163 @@
...
+int latency_config_copy (latency_config_t *dst, const latency_config_t src)
+{
+  /* Copy percentiles configuration */
+  dst->percentile_num = src.percentile_num;
+  dst->percentile = malloc(sizeof (*dst->percentile) * (src.percentile_num));
+  if (dst->percentile == NULL)
+    goto nomem;
```
octo: Please avoid using `goto`. Instead, use `latency_config_free()` for easy freeing of all fields and return:

```c
*dst = (latency_config_t) {
  .rates = NULL,
  .rates_num = src.rates_num,
  .rates_type = NULL,
  .percentile = NULL,
  .percentile_num = src.percentile_num,
  .percentile_type = NULL,
};
/* … */
if (dst->percentile == NULL)
{
  latency_config_free (*dst);
  return (ENOMEM);
}
```

Here and below.

octo: Please avoid using `goto`. Instead, use `latency_config_free()` for easy freeing of all fields and return:

```c
*dst = (latency_config_t) {
  .rates = NULL,
  .rates_num = src.rates_num,
  .rates_type = NULL,
  .percentile = NULL,
  .percentile_num = src.percentile_num,
  .percentile_type = NULL,
};
/* … */
if (dst->percentile == NULL)
{
  latency_config_free (*dst);
  return (ENOMEM);
}
```

Here and below.

src/utils_latency_config.h
```diff
@@ -0,0 +1,61 @@
+/**
```
octo: Please format new files with `clang-format -style=LLVM`

src/utils_latency_config.c
```diff
@@ -0,0 +1,163 @@
+/**
```
octo: Please format new files with `clang-format -style=LLVM`

src/daemon/Makefile.am
```diff
@@ -69,7 +69,9 @@ collectd_SOURCES = collectd.c collectd.h \
 		   utils_tail.c utils_tail.h \
 		   utils_time.c utils_time.h \
 		   types_list.c types_list.h \
-		   utils_threshold.c utils_threshold.h
+		   utils_threshold.c utils_threshold.h \
+		   ../utils_latency_config.h ../utils_latency_config.c \
```
rpv-tomsk: I think this should be done in separate patch/PR.
Is anyone working on this?

src/utils_latency_config.c
rpv-tomsk: That is because latency_counter_add() has 

``
if (latency == 0) 
    return;
``
so these values are not taken into account.

src/collectd.conf.pod
```diff
@@ -6625,6 +6633,29 @@ their configuration can be found in L<types.db(5)>.
...
 
+=item B<LatencyPercentile> I<Percent>
+
+Calculate and dispatch the configured percentile, i.e. compute the latency, so
+that I<Percent> of all matched latency values are smaller than or equal to the
+computed latency.
+
+Different percentiles can be calculated by setting this option several times.
+
+=item B<LatencyPercentileType> I<Type>
```
rpv-tomsk: Fixed type for reported values is not appropriate, it should be configurable.

The `tail` plugin `<Match>` block has `Type` option for reported values.
But with `DSType "Latency"` set I can report percentiles and rates both in one match block. 
That is typical task for latency - to report percentiles and rates both.

When both percentiles and rates are reported from one `Match` block,  rates must be sent as different type than percentiles.

So, two options was added: `LatencyPercentileType` and `LatencyRateType`.

When no value set by these options, then `Type` option is used in reported values. 
When both values are set by these options, then `Type` option is not used at all.

What do you think about these options?

src/utils_latency.c
rpv-tomsk: No, that can match. 

Conditions example:

1) `latency_counter_get_rate(lc, 0.100, 0.150, now)`
2)  bin width 0.100

So, we should take only half from bin value and condition matches.

Another example:

1) `latency_counter_get_rate(lc, 0.100, 0.150, now)`
2)  max latency 0.110

So, we need to take full value from last bin, condition does not match.

src/utils_latency_config.c
rpv-tomsk: Like other plugins, this was done to show plugin name in error messages:
``
ERROR ("%s plugin: \"%s\" requires exactly one numeric argument."
``
No other reasons for this.

My proposal is to add plugin name into `context` and set that name in cf_register_complex().
So, cf_util_get_* functions can use that to build nice error messages.

What do you think?

CC: @rubenk , @mfournier

src/utils_latency_config.c
octo: But this is configuring the limits for the "Rate" option, not adding a measurement via `latency_counter_add()`, so this check is not actually triggered, right? Since a latency measurement must be larger than zero, configuring the lower bound with zero effectively means "no lower bound". So `min=0` behaves exactly like `max=0` in that it does not filter out any values because they are too small (too large in the max case). I find this symmetry in the config semantics very appealing.

src/daemon/Makefile.am
octo: Noone yet, but I'm happy to take this on.

src/collectd.conf.pod
octo: I think that we know quite precisely that we're reporting a latency and as such should use the `latency` type. Allowing users to use a different *type* has next to no benefit and will result in more users seeking help to set this up.

With `LatencyPercentileType` removed, there is no need to have `LatencyRateType` – the rates should be reported using the general `Type` option.

I hope that this simplifies the config and reduces the number of support requests by users significantly.

src/utils_latency.c
octo: I formulated that poorly, sorry. What I meant was: `upper < (upper_bin + 1) * lc->bin_width` is always true. You're right, though, and I was mistaken: if `upper` falls directly on a bin boundary, then this evaluates to false:

Assume `upper = 100ms` => `upper_bin = 99` => `(upper_bin + 1) * bin_width = 100ms == upper`

Still, I propose to do the following (note the "greater than or equal"):

```c
if (upper)
{
  assert (upper <= (upper_bin + 1) * lc->bin_width);
  /* … */
}
```

The suggested code below will still work: the extreme case outlined above will result in `ratio` being zero and `sum` will not be modified.

src/utils_latency_config.c
octo: That's a great idea! `ERROR()` aka. `plugin_log()` could use this also and we could remove *a lot* of "foo plugin: " prefixes from all over the codebase. Could you open a feature request issue for this please?

src/collectd.conf.pod
rpv-tomsk: > should use the latency type. 
> Allowing users to use a different type has next to no benefit 

One module (generic module!) - one 'plugin name', one type? Oh, no. It is not flexible enough.
In `tail`, I can use more than one <Match> per file with `DSType latency`. That will require configurable `type`.

src/utils_latency_config.c
rpv-tomsk: The #1621 also already opened for `0.000 value` issue.

src/utils_latency.c
rpv-tomsk: That piece is covered by tests and without DOUBLE_TO_CDTIME_T(0.001) it fails.
I will try to recover and document explanation of this nuance.

src/utils_latency.c
rpv-tomsk: Ok, after working on this piece of code I see what it will be better for code readability to put proposed assert().

With proposed improvements, code is read highly better.

Thanks, @octo!

src/utils_latency_config.c
rpv-tomsk:
```
*dst = (latency_config_t) {
  .rates = NULL,
  .rates_num = src.rates_num,
  .rates_type = NULL,
  .percentile = NULL,
  .percentile_num = src.percentile_num,
  .percentile_type = NULL,
};
```

Why this change is required? I think, all fields are already initialized to 0 / NULL by compiler.
Or not?

src/utils_latency_config.h
rpv-tomsk: PR updated.

rpv-tomsk:
PR updated.

rpv-tomsk:
Hi !

Let's continue work on this PR?

octo:
Hi Pavel,

the details of the rate calculation were still itching me, so I started playing with the code and unit tests and eventually came up with the changes in [the pr/1700 branch](https://github.com/octo/collectd/commits/pr/1700). Unfortunately I can't push those to the PR myself.

Best regards,
—octo

rpv-tomsk:
Hi!

I just checked "Allow edits from maintainers." checkbox.
Can that help you?

octo:
Thanks @rpv-tomsk, yes, that should allow me to push to your branch. I'll retry that …

octo:
This should be good now, with one exception: I'm still not convinced we should keep the `LatencyPercentileType` option. @rpv-tomsk, what do you intend to use it for? Can you give me a bit of background for this option?

rpv-tomsk:
The tail plugin <Match> block has Type option for reported values.
But with DSType "Latency" set I can report percentiles and rates both in one match block.
That is typical task for latency - to report percentiles and rates both.

When both percentiles and rates are reported from one Match block, rates must be sent as different type than percentiles. So, at least one option must be added.

We can add only one option `LatencyRateType` and use `Type` option value as `LatencyPercentileType`, but, as for me, that does configuration unclear.  

Why we prefer one option against other? Why do not opposite: add `LatencyPercentileType`  and remove `LatencyRateType`?

I propose to add both options for completeness: In current state there is `LatencyRate` + `LatencyRateType` and `LatencyPercentile` + `LatencyPercentileType`.

rpv-tomsk:
Hmm, it looks I have just broke some things....
Can you please recover?

rpv-tomsk:
fb1dfac src/utils_latency.[ch]: Improve accuracy, update unit test.
	3260a23 src/utils_latency.[ch]: Remove latency_counter_get_start_time().
	a81285d src/utils_latency.[ch]: get_rate(): Make lower bound exclusive.
	57047e6 collectd.conf(5): Update the "LatencyRate" option.
	526e0e3 src/utils_latency{,_config}.[ch]: Reformat new code with clang-format.

rpv-tomsk:
Thanks! I have updated my local tail-dstype-latency-master branch )

octo:
So, the issue I have with `LatencyPercentileType` is: what else should it be? The *type* can be thought of as a unit, such as "bytes", "requests", "celsius", … And it makes perfect sense to let the user specify the *type* of the values being matches in other situations: the user might want to count requests or calculate the average temperature.

With the `LatencyPercentile` setting, though, we know exactly what the unit of that metric is: it's a latency in seconds. Which other *type* would you set this to?

One thing that *just* occurred to me: it might be interesting to rename the options and features from "latency" to "histogram", allowing users to calculate the 95th percentile request size or the median temperature. In that case, a `HistogramPercentileType` option would make sense.

rpv-tomsk:
> With the LatencyPercentile setting, though, we know exactly what the unit of that metric is: it's a latency in seconds. Which other type would you set this to?

Example: `request_time`, `upstream_response_time`, `upstream_connect_time`, etc:

http://nginx.org/en/docs/http/ngx_http_log_module.html#var_request_time
http://nginx.org/en/docs/http/ngx_http_upstream_module.html#var_upstream_response_time
....

All these are in seconds. All these are from one line of nginx log. All these are one 'service'.

rpv-tomsk:
The next thing to discuss: `get_rate(): Make lower bound exclusive.`

1) In my version, `LatencyRate 0.001 0.100` was reported with `rate-0.001-0.100` instance.
When lower bound is exclusive, then option will be `LatencyRate 0.000 0.100` and it will be reported with `rate-0.000-0.100` type instance. But `0.000` values are not counted even if they are in the file.

That also was a reason to make lower bound inclusive. Some days we can allow to report `0.000` rate.

PS> I don't like these values, but....

2) In my version, when configured
```
LatencyRate 0.001 0.100
LatencyRate 0.101 0.200
``` 
type instances will be  `rate-0.001-0.100` and `rate-0.101-0.200`. So, we can see what value is where.

When lower bound is exclusive, instances will be `rate-0.000-0.100` and `rate-0.100-0.200`.

May be it will be good to add `+0.001` to lower bound in reported type instance?

rpv-tomsk:
> One thing that just occurred to me: it might be interesting to rename the options and features from "latency" to "histogram", **allowing users to calculate the 95th percentile request size** or the median temperature. In that case, a HistogramPercentileType option would make sense.

Great! Amazing.

Indeed, we are too fixated on 'latency'.

rpv-tomsk:
> In that case, a HistogramPercentileType option would make sense.

As that case will be different one, and in this case all things are  tied to `latency`, I propose to left this untouched: 

`DSType "Latency" -> LatencyRate ->  LatencyRateType`
`DSType "Latency" -> LatencyPercentile ->  LatencyPercentileType`

octo:
Hi @rpv-tomsk,

I've changed the patch to provide a generic "Distribution" type, which also allows to set the "type" to something that is not "latency", e.g. "request_size". I'm going to merge this before the feature freeze on Monday. We can still tweak little issues afterwards, but the overall set of _features_ will be set for this release. Things like min, max, average, … will have to wait for a later release.

Best regards,
—octo

rpv-tomsk:
>I've changed the patch to provide a generic "Distribution" type

That is "**Latency**" type, not "_Distribution_". 

It cannot be generic as all the data pass through DOUBLE_TO_CDTIME_T(). 

```
+  else if (strcasecmp ("Distribution", ds_type) == 0)
+  {
+    cm->flags = UTILS_MATCH_DS_TYPE_GAUGE | UTILS_MATCH_CF_GAUGE_LATENCY;
```

```
+    if (data->ds_type & UTILS_MATCH_CF_GAUGE_LATENCY)
+    {
+      latency_counter_add(data->latency, DOUBLE_TO_CDTIME_T(value));
```

octo:
Good catch, I'll rename `UTILS_MATCH_CF_GAUGE_LATENCY` in a bit.

Don't worry about `DOUBLE_TO_CDTIME_T()` too much: it is a way to convert a double to a uint64 representation (and back). The conversion is not ideal and at some point we probably should rewrite `utils_latency` as `utils_distribution` using doubles, but that's for another day.

> That is "**Latency**" type, not "_Distribution_". 

Sorry, I don't understand what you mean here.

Best regards,
—octo

octo:
Rebased the dev branch on *master*. Will merge once the builds succeed.

rubenk:
This broke make distcheck.

rubenk:
```
$ git bisect good
1d888f3f3a131f8eb2d9e0884665b6951ecee550 is the first bad commit
commit 1d888f3f3a131f8eb2d9e0884665b6951ecee550
Author: Pavel Rochnyack <pavel2000@ngs.ru>
Date:   Sat May 7 23:26:15 2016 +0600

    + liblatency: Added utils_latency_config code
    + tail plugin: DSType Latency added
```

rubenk:
It's probably this line: https://github.com/collectd/collectd/blob/master/src/daemon/Makefile.am#L71

Ah, I see that you mentioned it already @octo (https://github.com/collectd/collectd/pull/1700#discussion_r78966929)

Issue #1607, opened 2016-03-06
apcups plugin: Submitted values are flapping (bug in master branch)


rpv-tomsk:
When apcups plugin is configured  'PersistentConnecton true' (which is default value), flapping between real data and default values occurs.

Due to https://github.com/collectd/collectd/commit/fb075e15e3400afdb691b88f1e1a4a38578aaba9 , 
read loop aborted too early. The socket has unread data, and that data is 'status_close' packet with  zero-sized data.

The first cycle gets correct data, but read loop is aborted by 'END APC' received. The next net_recv() call returns 0 due to packet_size == 0. This leads to apc_query_server() also return 0 (success), and default values are sent by apcups_read(). As result,dispatched values are flapping between real data and default values at each cycle. 

Plugin is unuseable in default configuration, unless 'PersistentConnection' is disabled.

You can see debug output in https://github.com/collectd/collectd/issues/617.

The fb075e15e3400afdb691b88f1e1a4a38578aaba9 needs to be reverted. Reverting that patch solves the problem with flapping when persistent connection is configured.

As a solution of #617, I will offer a patch that will change the default value of 'PersistentConnection', depending on the configured read interval.

rpv-tomsk:
Created #1701 as a solution for this issue.

Issue #617, opened 2014-05-19
apcups plugin: socket errors when poll time is 'too large'


hhoffstaette:
I'm seeing a weird error with the apcups plugin on otherwise perfectly working Gentoo. When I activate the plugin - with or without explicit configuration - it complains:

```
[2014-05-19 10:31:22] [error] apcups plugin: Reading from socket failed: Bad file descriptor
[2014-05-19 10:31:22] [notice] read-function of plugin `apcups' failed. Will suspend it for 120.000 seconds.
```

The apcupsd daemon is of course running, accessible and working fine; apcaccess and the CGIs work correctly, as root or regular user.

In order to debug this I rigged up a simple in-tree standalone executable that just links against the apcups plugin's .o (and required friends) and runs the socket code. The result: everything works!

```
// un-static-fied in apcups.c & call to apc_submit() disabled
extern int apcups_config(const char* key, const char* value);
extern int apcups_read(void);

int main(int argc, char** argv)
{
	apcups_config("host", "localhost");
	apcups_config("port", "3551");
	return apcups_read();
}
```

The result, with the APCMAIN macro enabled and run against my live UPS, shows:

```
holger>./testapc                                                                                                      
net_recv = `APC      : 001,034,0864
';
  Found property: name = APC; value = 1.000000;
net_recv = `DATE     : 2014-05-19 15:35:10 +0200  
';
  Found property: name = DATE; value = 2014.000000;
  Found property: name = 15; value = 35.000000;
net_recv = `HOSTNAME : tux
';
  Found property: name = HOSTNAME; value = 0.000000;
net_recv = `VERSION  : 3.14.10 (13 September 2011) gentoo
';
  Found property: name = VERSION; value = 3.140000;
net_recv = `UPSNAME  : tux
';
  Found property: name = UPSNAME; value = 0.000000;
net_recv = `CABLE    : USB Cable
';
  Found property: name = CABLE; value = 0.000000;
net_recv = `DRIVER   : USB UPS Driver
';
  Found property: name = DRIVER; value = 0.000000;
net_recv = `UPSMODE  : Stand Alone
';
  Found property: name = UPSMODE; value = 0.000000;
net_recv = `STARTTIME: 2014-05-18 21:46:08 +0200  
';
  Found property: name = STARTTIME; value = 2014.000000;
  Found property: name = 21; value = 46.000000;
net_recv = `MODEL    : Back-UPS ES 700G 
';
  Found property: name = MODEL; value = 0.000000;
[...]

PR #1701, opened 2016-05-08
apcups plugin: Change "PersistentConnection" option default value depending on poll Interval.

rpv-tomsk:
This PR fixes core problem of #617 and reverts fb075e15e3400afdb691b88f1e1a4a38578aaba9 which causes problem, described in #1607.


mfournier:
Thanks so much for digging into this @rpv-tomsk !

Issue #1703, opened 2016-05-09
Collectd statsd plugin deadlock


Pratik-Shah:
I have a process lifecycle manager which handles collectd. While triggering the collectd restart it was found that there was a problem with the restart of the collectd service. After tracing it was found that statsd plugin had a lock which resulted in the deadlock and because of which collectd was unable to restart.
The lifecycle manger keeps on attempting the to restart the collectd process.
Below  is the snippet for the same .
[collectd] Restart explicitly requested - going to kill process & restart it...
[collectd] Restart - process terminated - process will be restarted automatically
[on/configuration/restart] config file: ../collectd/master/collectd.conf, exists: true
[collectd] Restart explicitly requested - going to kill process & restart it...
[collectd] Restart - process terminated - process will be restarted automatically

I Installed gdb https://www.gnu.org/software/gdb/ to trace the process. Please find the log trace and help me resolving this problem. ( Thread 7 contains the lock. )

udo gdb attach 18578
GNU gdb (GDB) Amazon Linux (7.6.1-64.33.amzn1)
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law. Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-amazon-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
attach: No such file or directory.
Attaching to process 18578
Reading symbols from /usr/sbin/collectd...done.
Reading symbols from /lib64/libm.so.6...(no debugging symbols found)...done.
Loaded symbols for /lib64/libm.so.6
Reading symbols from /lib64/libpthread.so.0...(no debugging symbols found)...done.
[New LWP 18596]
[New LWP 18595]
[New LWP 18594]
[New LWP 18593]
[New LWP 18592]
[New LWP 18586]
[New LWP 18585]
[New LWP 18584]
[New LWP 18583]
[New LWP 18582]
[New LWP 18581]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Loaded symbols for /lib64/libpthread.so.0
Reading symbols from /usr/lib64/libltdl.so.7...(no debugging symbols found)...done.
Loaded symbols for /usr/lib64/libltdl.so.7
[...]

PR #1704, opened 2016-05-09
statsd plugin: Fix deadlock on plugin shutdown (Issue #1703)

rpv-tomsk:
This is a patch for #1703. This is a bugfix and should be applied to all supported Collectd versions.

statsd_shutdown() holds metrics_lock before setting 'network_thread_shutdown' to 1.
So, statsd_network_thread() unable to finish work as it is awaiting same lock in
statsd_network_thread() -> statsd_network_read() -> statsd_parse_buffer() -> statsd_parse_line() -> statsd_handle_() -> statsd_metric_().

network_thread_shutdown must be set to 1 before aquiring metrics_lock in statsd_shutdown()
After that, pthread_join() will wait for statsd_network_thread() to finish. While this many 'lines' from may 'fd' buffers can be parsed by statsd_parse_line() and this can aquire 'metrics_lock' many times without lockup. After statsd_network_thread() finished statsd_shutdown() can aquire 'metrics_lock' and free 'metrics_tree'.

### DIFF
--- a/src/statsd.c
+++ b/src/statsd.c
@@ -955,8 +955,6 @@ static int statsd_shutdown (void) /* {{{ */
   void *key;
   void *value;
 
-  pthread_mutex_lock (&metrics_lock);
-
   if (network_thread_running)
   {
     network_thread_shutdown = 1;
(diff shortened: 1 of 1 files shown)


mfournier:
Merged as 961ead7, thanks a lot @rpv-tomsk, nice catch !

rpv-tomsk:
@mfournier, @rubenk 

Because it is a bug, the patch should be included in all supported branches. Can we fix this?

rpv-tomsk:
That is a Git magic... 
I don't see that patch in `git log` output for `origin/collectd-5.4` branch in my local repo copy, but see it in `git log src/statsd.c`. 
Patch was merged correctly, sorry for the noise  : - )

PR #1705, opened 2016-05-11
dbi/postgresql/oracle plugins: Fixes and improvements.

rpv-tomsk:
Hi!

Please review and approve.

P.S.: (oracle plugin has no memory leak in current code, please ignore branch name).


src/utils_db_query.c
```diff
@@ -1011,7 +1012,13 @@ int udb_query_prepare_result (udb_query_t const *q, /* {{{ */
   if ((q == NULL) || (prep_area == NULL))
     return (-EINVAL);
 
-  udb_query_finish_result (q, prep_area);
```
tokkee: Why did you change this? I didn't look at all details of the data flow here but `udb_query_finish_result` does more than deallocate fields of `prep_area`. It's generally a good idea to make sure you clean up after yourself, so keeping this call here even if it's a no-op (at the moment) is the better option imho. Otherwise, future, unrelated changes in other parts of this file may require changes here as well which should be avoided.

tokkee:
I suppose these fixes apply to 5.4 as well, so we'll want to merge it into that branch instead.

src/dbi.c
```diff
@@ -595,10 +595,16 @@ static int cdbi_read_database_query (cdbi_database_t *db, /* {{{ */
     sstrncpy (column_names[i], column_name, DATA_MAX_NAME_LEN);
   } /* }}} for (i = 0; i < column_num; i++) */
 
-  udb_query_prepare_result (q, prep_area, (db->host ? db->host : hostname_g),
+  status = udb_query_prepare_result (q, prep_area, (db->host ? db->host : hostname_g),
       /* plugin = */ "dbi", db->name,
       column_names, column_num, /* interval = */ (db->interval > 0) ? db->interval : 0);
 
+  if (0 != status)
```
rubenk: status != 0 please.

src/utils_db_query.c
rpv-tomsk: >  It's generally a good idea to make sure you clean up after yourself

That is right, but in this case (that is udb_query_prepare_result() function code) udb_query_finish_result() is called _before_, not _after_.
(

tokkee: Uh, I missed that, thanks. Anyway, in that case I actually disagree strongly with the proposed change. udb_query_prepare_result() overwrites the content in the preparation area and, hence, it's right to cleanup the old content before. It's totally fine for the caller not to care about this and neither a memory leak (if compiled without debugging) or a program crash would be acceptable.

rpv-tomsk: Me strongly disagree too.

>  It's totally fine for the caller not to care about this.

No, that is abnormal.

Before proposed PR, the 'oracle' plugin has no udb_query_finish_result() in o_read_database_query() while udb_query_prepare_result() was there. If no o_read_database_query() called more, preparation area will never be 'clean' and memory will not be freed.

After this patch, rules become more strong, and that is good. Currently, udb_query_prepare_result() and udb_query_finish_result() are done on the same level in 'oracle' plugin, so 'preparation area' freed right after use.

Freeing memory not after use but right before next use is ugly.

Currently, if dig deeper, as 'oracle.c' does not call udb_query_finish_result() and udb_query_delete_preparation_area() also does not call udb_result_finish_result(),   sfree (result_prep_area->metadata_pos) and sfree (result_prep_area->metadata_buffer) are lost on plugin shutdown.

It's totally abnormal for the caller to call udb_query_prepare_result() and do not call udb_query_finish_result().

Rules must be strong and other plugins has correct implementation before this PR:

dbi.c            has udb_query_finish_result() in cdbi_read_database_query()
postgresql.c has udb_query_finish_result() in c_psql_exec_query().

I think what all plugins should behave in the same way and must have udb_query_finish_result() in their code.

rpv-tomsk:
Fixed remark found by review. Should I open new PR against 5.4?

rpv-tomsk:
Hi!

Please fix that buildbot issue.

Thanks.

rpv-tomsk:
Hi! 

Can we merge this?

rpv-tomsk:
Hi! 

Should I rebase this to current `master` so checks will pass?

src/utils_db_query.c
rpv-tomsk: That is assert() is intended for.

rpv-tomsk:
ping

rpv-tomsk:
Rebased to current master. Let's merge this, please?

rpv-tomsk:
ping.

rpv-tomsk:
Are there any objections against merge?

rubenk:
no, looks good, go ahead.

TysonAndre:
bq. P.S.: (oracle plugin has no memory leak in current code, please ignore branch name).

For anyone looking at this PR, that comment seems to be outdated - there is a leak that's fixed in this PR?

A leak seems to have been discovered and the commit `utils_db_query: Added lost free().` was added after the comment in the PR description was added.

This PR is included in https://collectd.org/wiki/index.php/Version_5.9

PR #1707, opened 2016-05-12
Allowed custom plugin name in 'dbi', 'oracle' and 'postgresql' plugins.

rpv-tomsk:
Like added in  #1558,  #1563 and #1681, added option 'PluginName' to dbi-related plugins: 'dbi', 'oracle' and 'postgresql'.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -491,7 +491,7 @@
 #		</Result>
 #	</Query>
 #	<Database "customers_db">
-#		#PluginName "mycompany"
+#		#Plugin "mycompany"
 #		Driver "mysql"
 #		DriverOption "host" "localhost"
 #		DriverOption "username" "collectd"
@@ -1015,7 +1015,7 @@
 #    </Result>
 #  </Query>
 #  <Database "product_information">
-#    #PluginName "warehouse"
+#    #Plugin "warehouse"
 #    ConnectID "db01"
 #    Username "oracle"
 #    Password "secret"
@@ -1102,7 +1102,7 @@
 #		StoreRates true
 #	</Writer>
 #	<Database foo>
-#		#PluginName "kingdom"
+#		#Plugin "kingdom"
 #		Host "hostname"
 #		Port "5432"
 #		User "username"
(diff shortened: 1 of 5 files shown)


tokkee:
Please see my comment in https://github.com/collectd/collectd/pull/1681#issuecomment-213991806: I think it should be possible to fetch the plugin name from the query result.

rpv-tomsk:
> That said, instead of (or in addition to) a static name, I wonder if it would be better to fetch the name from a query result. In SQL plugins (e.g., #1707), this would be easy, because a static name could also be provided by the query. For the plugins in this PR, this may not always be feasible, though.

I think that feature will be excessive. I think, plugin name is toо static (unlike the plugin instance) to query it from database. In my practice, it is enough to get plugin instance from SQL/curl/so on and no need to get plugin name same way.

Can you provide an use-case of such feature? 

Also I think what all 'generic' plugins should behave in similar manner, so 'static plugin name' option is required.

tokkee:
I was thinking along the lines of adding `PluginNameFrom` to the `Query` block instead of the `Database` block. A query may provide information about multiple databases and similar to how you'd want to configure different plugin names per database, you'd then want the same feature for these queries. In fact, I even wonder if `PluginName` should be a query or database option (or both?) since the plugin name is supposed to describe the data (which is provided by the query) instead of where it comes from (the database which issues many queries).

I'll have to think about this a little more because I think we should make sure to get the semantics right. Again, a central place to discuss this might be a good idea.

octo:
As noted in #1681, please use `Plugin` (and potentially `PluginFrom`) as the option name. Also as noted in #1681 and by @tokkee above, please open a feature-request issue so we have a non-plugin-specific place to talk about this.

rpv-tomsk:
Renamed 'PluginName' option to unified 'Plugin' and rebased.

rpv-tomsk:
ping

rpv-tomsk:

rpv-tomsk:

octo:
@tokkee A `PluginFrom` option SGTM, but I'm happy to postpone implementation until users request it. Does that work for you?

Best regards,
—octo

Issue #1706, opened 2016-05-12
[COREDUMP] Perl plugin randomly crashing collectd


TomGudman:
Hi, 

We developed a custom perl plugin which seems to crash collectd from time to time, sometimes with a coredump... sometimes not.... Interestingly enough, we would say it __seems__ to only happen on Centos 6.7 and did not happen on Centos 6.6 but this could be a coincidence. We tried with collectd 5.5.0 and the latest 5.5.1 without success.
As a workaround we are now using collectdmon to respawn collectd but it obviously creates gap in the measurement when collectd crashes. 

As a consequence of the crashes, we can also find gibberish filenames in /tmp. We are running out of ideas to find why it randomly dies. We can send our code to a collectd developer privately if required.
 
```
[..] /tmp]# ls -l
total 172608
-rw-r--r--  1 root root         0 Apr 30 19:00 ?Ä
-rw-r--r--  1 root root         0 May  9 13:59 Â_?Á?9¥^á?ï{½ë.må?râ$??
-rw-------  1 root root  16691200 Apr 27 03:37 core-collectd-11-0-0-11300-1461692221
-rw-------  1 root root  16691200 Apr 29 03:23 core-collectd-11-0-0-13487-1461864181
[..]
-rw-------  1 root root  16691200 May  7 03:34 core-collectd-11-0-0-9909-1462556042
-rw-------  1 root root 164052992 Apr 28 03:08 core-collectd-6-0-0-11305-1461776881
```

## Context
  * Centos 6.7
  * Collectd 5.5.1
  * Perl plugin enabled
```
LoadPlugin perl
TypesDB "/etc/collectd.d/l2tpns.typesdb"
<Plugin perl>
  IncludeDir "/etc/collectd.plugins/perl"
  BaseName "Collectd::Plugins"
  LoadPlugin "l2tpns"
  <Plugin "l2tpns">
  </Plugin>
</Plugin>
```

## How to reproduce
It cannot be reproduced at this point in time. It happens occasionally  

## Core dump 
When it does write a core dump it seems to happen around 3 and 4pm which coincidate with logrotate. Our logrotate config will kill collectd in purpose to restart collectd so what you may see below may just be standard behaviour.  

```
# gdb /usr/sbin/collectd /tmp/core-collectd-11-0-0-5304-1462819621
GNU gdb (GDB) Red Hat Enterprise Linux (7.2-83.el6)
Copyright (C) 2010 Free Software Foundation, Inc.
[...]

PR #1710, opened 2016-05-13
Perl plugin fixes for #1706

rpv-tomsk:
* lock base thread interpreter in perl_init() too, to avoid race conditions with c_ithread_create() called from threads of already-initialized plugins. (with deadlock when perl_log() called from perl_init() detection)
* Fix coredump due to destroying interpreter on thread running perl.

See #1706 for comments.

### DIFF
--- a/src/perl.c
+++ b/src/perl.c
@@ -1,4 +1,4 @@
-/**
+/*
  * collectd - src/perl.c
  * Copyright (C) 2007-2009  Sebastian Harl
  *
@@ -1911,6 +1911,7 @@ static XS (Collectd_call_by_name)
 
 static int perl_init (void)
 {
+	int status;
 	dTHX;
 
 	if (NULL == perl_threads)
@@ -1928,7 +1929,19 @@ static int perl_init (void)
 
 	log_debug ("perl_init: c_ithread: interp = %p (active threads: %i)",
 			aTHX, perl_threads->number_of_threads);
-	return pplugin_call_all (aTHX_ PLUGIN_INIT);
+
+	/* Lock the base thread to avoid race conditions with c_ithread_create().
+	 * See https://github.com/collectd/collectd/issues/9 and
+	 *     https://github.com/collectd/collectd/issues/1706 for details. */
+
+	assert (aTHX == perl_threads->head->interp);
+	pthread_mutex_lock (&perl_threads->mutex);
+
+	status = pplugin_call_all (aTHX_ PLUGIN_INIT);
+
+	pthread_mutex_unlock (&perl_threads->mutex);
+
+	return status;
 } /* static int perl_init (void) */
 
 static int perl_read (void)
--- a/src/perl.c
+++ b/src/perl.c
@@ -2168,11 +2168,6 @@ static int perl_shutdown (void)
 			nanosleep (&ts_wait, NULL);
 		}
 		if (thr->running) {
-			/* This will crash collectd process later due to PERL_SYS_TERM() */
-			//ERROR ("perl shutdown: thread hangs inside perl. "
-			//       "Skipped perl interpreter destroy.");
-			//continue;
-			
 			ERROR ("perl shutdown: thread hangs inside perl. Thread killed.");
 			pthread_kill (thr->pthread, SIGTERM);
 		}
--- a/src/perl.c
+++ b/src/perl.c
@@ -1,4 +1,4 @@
-/*
+/**
  * collectd - src/perl.c
  * Copyright (C) 2007-2009  Sebastian Harl
  *
@@ -117,6 +117,9 @@ static XS (Collectd_call_by_name);
 typedef struct c_ithread_s {
 	/* the thread's Perl interpreter */
 	PerlInterpreter *interp;
+	_Bool running;  /* thread is inside pi */
+	_Bool shutdown;
+	pthread_t pthread;
 
 	/* double linked list of threads */
 	struct c_ithread_s *prev;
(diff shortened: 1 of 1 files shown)


rubenk:
@rpv-tomsk since these are bugfixes, can you create a PR against the oldest branch they apply too?

rpv-tomsk:
There is no much changes in `src/perl.c` since 2012, so this patch can be applied even to collectd-4.10 without manual merges.  
Rebased against collectd-5.4. That should be enough old branch?
That should be enough or other PR is required ?

rubenk:
No this should be fine. I'm not familiar with the perl API, @tokkee, would you mind reviewing this?

src/perl.c
```diff
@@ -1008,11 +1011,24 @@ static int pplugin_call_all (pTHX_ int type, ...)
...
+
+	old_running = t->running;
+	t->running = 1;
+	
+	if (t->shutdown) {
+		t->running = old_running;
+		return 0;
+	}
+
 	if ((type < 0) || (type >= PLUGIN_TYPES))
```
tokkee: t->running needs to be reset before returning here. The same is true for fc_call() below. I think instead of your current approach, we should have a wrapper around call_pv(). That's where the actual switch from C to Perl happens and I don't think the Perl stack modifications that happen here before call_pv() would contribute to the problem you're fixing.

src/perl.c
```diff
@@ -2026,13 +2062,19 @@ static void perl_log (int level, const char *msg,
 
 	/* Lock the base thread if this is not called from one of the read threads
 	 * to avoid race conditions with c_ithread_create(). See
-	 * https://github.com/collectd/collectd/issues/9 for details. */
-	if (aTHX == perl_threads->head->interp)
+	 * https://github.com/collectd/collectd/issues/9 for details.
+	 * Additionally check, if we are called from perl interpreter.
+	 * Maybe PTHREAD_MUTEX_RECURSIVE mutex type will be more appropriate?
```
tokkee: Yeah, I think a recursive mutex would be a better option.

src/perl.c
```diff
@@ -2117,12 +2159,30 @@ static int perl_shutdown (void)
...
+			WARNING ("perl shutdown: thread is running inside perl. Waiting.");
+			ts_wait.tv_sec = 0;
+			ts_wait.tv_nsec = 500000;
+			nanosleep (&ts_wait, NULL);
+		}
+		if (thr->running) {
+			/* This will crash collectd process later due to PERL_SYS_TERM() */
+			//ERROR ("perl shutdown: thread hangs inside perl. "
+			//       "Skipped perl interpreter destroy.");
+			//continue;
```
tokkee: This comment block doesn't seem to be needed?

src/perl.c
```diff
@@ -2117,12 +2159,30 @@ static int perl_shutdown (void)
...
 		 * the thread as this will free the memory */
 		t = t->prev;
 
+		thr->shutdown = 1;
+		if (thr->running) {
+			/* Give some time to thread to exit from pi */
+			WARNING ("perl shutdown: thread is running inside perl. Waiting.");
+			ts_wait.tv_sec = 0;
+			ts_wait.tv_nsec = 500000;
+			nanosleep (&ts_wait, NULL);
```
tokkee: I don't feel like this is the right approach. Instead of sleeping an arbitrary time here and then blindly killing stuff below, I tend to think that we should just wait for the plugin to finish. I believe we do the same for "regular" read plugins. Alternatively, did you check if there's a way to cleanly shut down running perl code?

src/perl.c
```diff
@@ -1008,11 +1011,24 @@ static int pplugin_call_all (pTHX_ int type, ...)
...
 	int retvals = 0;
 
+	_Bool old_running;
 	va_list ap;
 	int ret = 0;
 
 	dSP;
 
+	c_ithread_t *t = (c_ithread_t *)pthread_getspecific(perl_thr_key);
+	if (t == NULL) /* thread destroyed ( c_ithread_destroy*() -> log_debug() ) */
```
tokkee: Nit: no spaces after ( and before )

src/perl.c
rpv-tomsk: No. That is just a note for reviewers.

src/perl.c
rpv-tomsk: Agreed.

src/perl.c
rpv-tomsk: I can add additional variable and pthread_mutexattr_init(), pthread_mutexattr_settype() and pthread_mutexattr_destroy() required calls.  But I have no much desire to repeat full testing which was already done. Maybe leave this untouched at this moment? What do you think?

src/perl.c
rpv-tomsk: I have no idea how to stop program execution more correctly than proposed solution. 
This code handles perl interpreter on threads which was started by other Collectd plugins also. 
Perl interpreter can be attached to such threads too, like to other threads.

Plugins should shutdown their threads by himself and perl plugin can't (and should not) affect this. 
The perl plugin can only finish his work (if running) and deny future calls to perl interpreter.

If perl plugin code works correctly, it will do his work fast and program (thread) execution will return to collectd/plugin.  No new enter to perl interpreter will happen due to 'shutdown' flag, attached to thread.

If perl plugin code does not finish his work, I found no way to terminate it gracefully.
I even have no idea how that can be implemented. 
SIGTERM is simply an attempt to terminate a separate thread to avoid process crash. 
That is only attempt without any certainty of success.

From my point of view, we give all required to perl plugin developers to make correct implementations.
If perl plugin implementation will be correct (and even if not quite correct), all things will happen successfully.

src/perl.c
rpv-tomsk: I do not understand. Please explain more detailed.

rpv-tomsk:
Updated based on reviews. @tokkee , please look.

src/perl.c
tokkee: Fair enough. Let's go with your current implementation and we can iterate later if needed.

src/perl.c
tokkee: Please write "(c_ithread_destroy...)" instead of "( c_ithread_destroy… )" (that is, drop the space after "(" and before ")".

tokkee:
:+1: LGTM

Thanks!

I'll pull this into master and then cherry-pick it into 4.10.

tokkee:
I pulled this into 4.10 and 5.4. 5.5 will happen when doing the next merge.

PR #1713, opened 2016-05-17
add PersistFrequency option to the threshold plugin

bs-github:
Hello,
I've implemented a feature to limit the amount of notifications that the threshold plugin generates when the persist(ok) option is set.
The feature is documented and implemented similar to the hits feature.
From my point of view the feature has the flaw of being bursty, but that's documented and still better than not having the feature at all.

The flaw described: I set the PersistFrequency option to whatever value and get notifications every "frequency" in bursts for all the OK threshold checks after collectd start.

I'd like to spread out those bursts via starting with a random(frequency) initial, but I would need some help to find the right place in the code to do that and maybe even the right way to access the configured frequency there.
I'm happy to implement that as well if someone is up to point me into the right direction.

By the way, this is part of adding a direct link from collectd into Flapjack for alerting.

Cheers,
Birger

### DIFF
--- a/src/collectd-threshold.pod
+++ b/src/collectd-threshold.pod
@@ -76,6 +76,9 @@ information.
      <Type "cpu">
        Instance "idle"
        FailureMin 10
+       Persist true
+       PersistOK true
+       PersistFrequency 30
      </Type>
    
      <Plugin "memory">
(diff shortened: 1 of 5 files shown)


rubenk:
Thanks for your contribution!

I've cherry-picked c3750e3 into the collectd-5.4 and collectd-5.5 branches since this is a bugfix.

src/collectd-threshold.pod
```diff
@@ -163,6 +166,20 @@ generated for each value that is in the acceptable range. If set to B<false>
...
 the previous value was not.
 
+=item B<PersistFrequency> I<Value>
+
+Sets how often notifications are generated if B<Persist> or B<PersistOK> is set
+to B<true>. This defaults to B<0>, which means every time a value comes in, a 
+notification is dispatched.
+If set to a positive integer, persistent notifications are suppressed that number
+of B<Interval>s before dispatching the next notification.
+This can be seen as a way to rate limmit dispached persistent notifications while
```
octo: s/limmit/limit/

src/daemon/utils_cache.h
```diff
@@ -48,6 +48,9 @@ int uc_set_state (const data_set_t *ds, const value_list_t *vl, int state);
 int uc_get_hits (const data_set_t *ds, const value_list_t *vl);
 int uc_set_hits (const data_set_t *ds, const value_list_t *vl, int hits);
 int uc_inc_hits (const data_set_t *ds, const value_list_t *vl, int step);
+int uc_get_persist_frequency (const data_set_t *ds, const value_list_t *vl);
```
octo: Could we use `uc_meta_data_get_signed_int()` to avoid this highly specific interface?

src/collectd-threshold.pod
```diff
@@ -163,6 +166,20 @@ generated for each value that is in the acceptable range. If set to B<false>
 (the default) then a notification is only generated if a value is in range but
 the previous value was not.
 
+=item B<PersistFrequency> I<Value>
+
+Sets how often notifications are generated if B<Persist> or B<PersistOK> is set
+to B<true>. This defaults to B<0>, which means every time a value comes in, a 
+notification is dispatched.
+If set to a positive integer, persistent notifications are suppressed that number
```
octo: Can *Value* be a time in seconds? The *interval* setting may be different for each metric, so this setting has very different meaning for them …

PR #1565, opened 2016-02-25
 * disk plugin: call ignorelist_match() once per device, not for each send metric

rpv-tomsk:
There is no need to call ignorelist_match() for each device many times (for each collected metric). One call per cycle is enough.

PR #1723, opened 2016-05-25
Disk plugin: UdevNameAttr feature improvements

rpv-tomsk:
disk plugin: Initialize udev context only once, not per each read
disk plugin: Call disk_udev_attr_name() only when UdevNameAttr configured

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -230,7 +230,16 @@ static int disk_init (void)
 /* #endif HAVE_IOKIT_IOKITLIB_H */
 
 #elif KERNEL_LINUX
-	/* do nothing */
+#if HAVE_LIBUDEV
+	if (conf_udev_name_attr != NULL)
+	{
+		handle_udev = udev_new();
+		if (handle_udev == NULL) {
+			ERROR ("disk plugin: udev_new() failed!");
+			return (-1);
+		}
+	}
+#endif /* HAVE_LIBUDEV */
 /* #endif KERNEL_LINUX */
 
 #elif KERNEL_FREEBSD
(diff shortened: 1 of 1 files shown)


rubenk:
Thanks @rpv-tomsk!

src/disk.c
```diff
@@ -867,9 +867,13 @@ static int disk_read (void)
...
 #if HAVE_LIBUDEV
-		char *alt_name = disk_udev_attr_name (handle_udev, disk_name, conf_udev_name_attr);
-		if (alt_name != NULL)
-			output_name = alt_name;
+		char *alt_name = NULL;
+		if (conf_udev_name_attr != NULL)
+		{
+			alt_name = disk_udev_attr_name (handle_udev, disk_name, conf_udev_name_attr);
+			if (alt_name != NULL)
+				output_name = alt_name;
```
rubenk: This leaks alt_name if ignorelist_match is true and we continue the loop.
disk_udev_attr_name allocates, so the result should be freed.

tokkee: Fixed in #1741

rpv-tomsk: Hmm, I need to upgrade some my machines now...

Thank you for catching this.

rubenk: Thank Coverity ;) I should have caught it during review, sorry.

rpv-tomsk: The problem caused by #1565.  First version of #1565 was even worse, some of problems was fixed in new revision, but this one was overlooked.

rubenk: Ah well. At the end it is an improvement, and this is why we have a master branch.
Let me know if you'd like access to Coverity btw.

PR #1725, opened 2016-05-25
processes: MAXCOMLEN is not exposed on Solaris

rubenk:
MAXCOMLEN in <sys/user.h> is only exposed to kernel
code. I think it's safe to assume that it isn't going
to change, so just hardcode it.

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -130,6 +130,10 @@
 
 # include <sys/user.h>
 # include <dirent.h>
+#ifndef MAXCOMLEN
+#define MAXCOMLEN 16
+#endif
+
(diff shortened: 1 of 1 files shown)


mfournier:
Just pushed as 9ee4ca8 (amended your patch to also remove `<sys/user.h>`)

Thanks for investigating all this @rubenk !

rubenk:
I left `<sys/user.h>` in just in case Oracle would ever expose MAXCOMLEN in there. Probably never going to happen though...

mfournier:
doh, you're right, sorry ! Fixed with aee1cc1.

Issue #624, opened 2014-05-24
cpufreq plugin: no values with intel_pstate driver


hhoffstaette:
On Sandy Bridge and newer CPUs newer kernels can (and should) use the intel_pstate frequency driver instead of the ACPI driver . This driver has new/changed sysfs entries, so the currently read values at

```
/sys/devices/system/cpu/cpuX/cpufreq/scaling_cur_freq
```

no longer exist. The new values seems to be:

```
/sys/devices/system/cpu/cpuX/cpufreq/cpuinfo_cur_freq
```

Which values to read could be detected by the existence of the entry called

```
/sys/devices/system/cpu/intel_pstate/
```

rubenk:
Is this still the case? Afaict this is fixed in later kernels by https://github.com/torvalds/linux/commit/c034b02e213d271b98c45c4a7b54af8f69aaac1e

Documentation/cpu-freq/user-guide.txt also hints that scaling_cur_freq and cpuinfo_cur_freq are not the same. The latter takes turbo boost into account.

hhoffstaette:
I no longer use collectd but just verified that with 4.1 and intel_pstate `/sys/devices/system/cpu/cpuX/cpufreq/scaling_cur_freq` exists again, so apparently this has changed. If someone can confirm that it works correctly with the intel_pstate driver (which should be the automatically selected default for newer Intel CPUs) I'd be more than happy to close this.

igsol:
Please, pay attention to https://github.com/collectd/collectd/pull/1730 .

I had to create those changes, cause original cpufreq plugin doesn't collect frequency on such Linux version:
```
$ uname -a
Linux cees-dtn0 3.10.0-327.el7.x86_64 #1 SMP Thu Nov 19 22:10:57 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
$ lsb_release -a
LSB Version:	:core-4.1-amd64:core-4.1-noarch
Distributor ID:	CentOS
Description:	CentOS Linux release 7.2.1511 (Core) 
Release:	7.2.1511
Codename:	Core
```

rubenk:
This seems fixed in CentOS 7.3:
```
ruben@test1: ~$ uname -r
3.10.0-514.el7.x86_64
ruben@test1: ~$ cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq
2401000
```

Poil:
Hi,

[...]

PR #1730, opened 2016-05-27
Now cpufreq is able to get CPU frequency on CentOS 7.2

igsol:
On the latest CentOS 7.2 there is no "/sys/devices/system/cpu/cpu%d/cpufreq/scaling_cur_freq" file anymore on certain OS configurations.

The proposed changes make cpufreq plugin functional on CentOS 7.2.

### DIFF
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -31,7 +31,7 @@
 static int num_cpu = 0;
 
 static char const * freq_fname_def = "/sys/devices/system/cpu/cpu%d/cpufreq/"
-                                     "scaling_cur_freq";
+									 "scaling_cur_freq";
 /**
  * Linux kernel exposes current CPU frequency via
  * "/sys/devices/system/cpu/cpuX/cpufreq/scaling_cur_freq" file.
@@ -45,10 +45,10 @@ static char const * freq_fname_def = "/sys/devices/system/cpu/cpu%d/cpufreq/"
  * This file exposes not exactly the same thing as "scaling_cur_freq" does, but close by nature.
  *
  * Thus it's better to have a workaround to monitor some CPU frequency on even "broken" kernels.
- * So, the optional paramer "Path" was introduced to specify alternative CPU frequency path.
+ * So, the optional parameter "Path" was introduced to specify alternative CPU frequency path.
  * Example:
- * <Plugin cpufreq>/sys/devices/system/cpu/cpu%d/cpufreq/cpuinfo_cur_freq
- *     Path ""
+ * <Plugin cpufreq>
+ *     Path "/sys/devices/system/cpu/cpu%d/cpufreq/cpuinfo_cur_freq"
  * </Plugin>
  *
  * Note, "%d" has to be used only once as CPU index placeholder.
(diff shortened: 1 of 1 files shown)


src/cpufreq.c
```diff
@@ -28,18 +31,52 @@
...
+
 static int cpufreq_init (void)
 {
-        int status;
+	int status;
 	char filename[256];
 
+	if ( !setup_freq_fname() ) {
+		plugin_unregister_read ("cpufreq");
+		return 0;
```
tokkee: Don't call `plugin_unregister_read` from here. The daemon will handle that for you by returning a negative value instead of zero.

src/cpufreq.c
```diff
@@ -28,18 +31,52 @@
 
 static int num_cpu = 0;
 
+static char const * freq_fname = "/sys/devices/system/cpu/cpu%d/cpufreq/"
+								 "scaling_cur_freq";
+static char const * freq_fname_rhel = "/sys/devices/system/cpu/cpu%d/cpufreq/"
```
tokkee: The `_rhel` suffix sounds wrong. I don't think this is RHEL specific but rather due to a change or config option in the kernel. Can you dig up what triggered this or, else, use a more generic suffix?

src/cpufreq.c
```diff
@@ -28,18 +31,52 @@
...
+	if (!access (filename, R_OK))
+		return 1;
+
+	status = ssnprintf (filename, sizeof (filename),
+						freq_fname_rhel, 0);
+	if ((status < 1) || ((unsigned int)status >= sizeof (filename)))
+		return 0;
+
+	if (!access (filename, R_OK)) {
+		freq_fname = freq_fname_rhel;
```
tokkee: Instead of overwriting `freq_fname`, I think it's cleaner to use three variables: the two you have now (the templates) and then another one to store the actual filename. That way, this function could safely be called multiple times. I'd rename the existing `freq_fname` variable to something else and then use the `freq_fname` name for the new variable.

src/cpufreq.c
```diff
@@ -18,6 +18,9 @@
  *
  * Authors:
  *   Peter Holik <peter at holik.at>
+ *
+ * Changes:
+ *   2016-05-27: freq_fname_rhel was added by Igor Solovyov <igor dot solovyov at gmail dot com>.
```
tokkee: Nit: You can just add yourself to Authors. The details of what you did can be extracted from the Git history.

src/cpufreq.c
```diff
@@ -28,18 +31,52 @@
...
+	return 0;
+}
+
 static int cpufreq_init (void)
 {
-        int status;
+	int status;
 	char filename[256];
 
+	if ( !setup_freq_fname() ) {
```
tokkee: Nit: Please don't use spaces after ( and before ) (i.e. no "padding" inside the parenthesis).

igsol:
Sebastian, thanks for nice review!
I'll incorporate what you suggested above and update pull request next week.

src/cpufreq.c
igsol: Thanks for such detail, appreciated.

src/cpufreq.c
igsol: I saw came across with the problem first time on CentOS 7.1, that's why used rhel suffix.
Agree with you, but don't have enough time to dig when and why the path was changed.
So, what about "_alt" (i.e. alternative) suffix? Or any other suggestions?

src/cpufreq.c
igsol: I'll think deeply next week...

src/cpufreq.c
tokkee: Ack. "_alt" or "_old" and "_new" are fine by me if we don't have more details. I'd add a comment like:

    /* The path was renamed at some point.
     * The new path was first seen in CentOS 7.1. */

That'll explain what's going on and will still allow people to dig into it later. Sounds good?

rubenk:
I'm still not entirely sure about this change. As mentioned in https://github.com/collectd/collectd/issues/624#issuecomment-118536407, this is fixed in later kernel versions, and I'm not sure if we should be the one to workaround broken kernels.

Also, cpuinfo_cur_freq and scaling_cur_freq are not entirely the same, since the first also takes turboboost into account.

So far we've seen this only on RHEL kernels, presumably since they didn't backport https://github.com/torvalds/linux/commit/c034b02e213d271b98c45c4a7b54af8f69aaac1e, and only with the intel p-state driver. I have RHEL boxes using the acpi cpufreq driver where collectd works just fine.

tokkee:
Generally, I don't mind if we work-around broken external dependencies (kernel, database systems, whatever) *if* the work-around is simple and many people are affected. That said, in this case, if there's a (relevant) difference, it wouldn't be a good idea imho. Can we somehow take the turboboost effects into account?

rubenk:
As long as we document in the commit log / changelog what systems are affected (so RHEL7 kernels with intel_idle drivers) I'm ok with this.

I'd argue that it would even be nice to see the effect of turboboost, but it is a difference in behaviour so it might surprise some people. The only way I see to take the effect into account is to submit cpuinfo_cur_freq as a separate metric.

tokkee:
Ack, I was thinking about a separate metric as well. That'll avoid the inconsistent behavior and add value for those who see both metrics.

igsol:
Please, consider new pull request https://github.com/collectd/collectd/pull/1769 instead of this one.

PR #1708, opened 2016-05-12
Postgresql plugin: set plugin instance from query results.

rpv-tomsk:
Hi!

Proposed patches adds ability to set plugin instance from query.
This feature is useful for generic plugins.

Example:

````
<Query instance_summary>
        Statement "SELECT 'summary' as instance,\
        sum(numbackends) as numbackends, \
        sum(xact_commit) as xact_commit, \
        sum(xact_rollback) as xact_rollback, \
        sum(blks_read) as blks_read, \
        sum(blks_hit) as blks_hit, \
        sum(tup_returned) as tup_returned, \
        sum(tup_fetched) as tup_fetched, \
        sum(tup_inserted) as tup_inserted, \
        sum(tup_updated) as tup_updated, \
        sum(tup_deleted) as tup_deleted \
        FROM pg_stat_database"

        PluginInstanceFrom "instance"

        #TODO: types should have '_summary' suffix
        <Result>
                Type "pg_numbackends"
                ValuesFrom "numbackends"
        </Result>
        #
        <Result>
                Type "pg_xact"
                InstancePrefix "commit"
                ValuesFrom "xact_commit"
        </Result>
        <Result>
                Type "pg_xact"
                InstancePrefix "rollback"
                ValuesFrom "xact_rollback"
        </Result>
        #
        <Result>
                Type "pg_blks"
                InstancePrefix "blks_read"
                ValuesFrom "blks_read"
        </Result>
        <Result>
                Type "pg_blks"
                InstancePrefix "blks_hit"
                ValuesFrom "blks_hit"
        </Result>
        #
        <Result>
                Type "pg_scan"
                InstancePrefix "tup_returned"
                ValuesFrom "tup_returned"
        </Result>
        <Result>
                Type "pg_scan"
                InstancePrefix "tup_fetched"
                ValuesFrom "tup_fetched"
        </Result>
        <Result>
                Type "pg_scan"
                InstancePrefix "tup_inserted"
                ValuesFrom "tup_inserted"
        </Result>
        <Result>
                Type "pg_scan"
                InstancePrefix "tup_updated"
                ValuesFrom "tup_updated"
        </Result>
        <Result>
                Type "pg_scan"
                InstancePrefix "tup_deleted"
                ValuesFrom "tup_deleted"
        </Result>
</Query>
<Query instance_bydb>
        Statement "SELECT datname, \
        numbackends, \
        xact_commit, xact_rollback, \
        blks_read, blks_hit, \
[...]

PR #1731, opened 2016-05-29
perl plugin: new implementation with plugin_register_complex_read() and so.

rpv-tomsk:
This PR moves perl plugins registration from perl plugin level to collectd level.
So, perl plugin read/write callbacks can be called multithreaded (in parallel), not in single thread, as it was before. Also no callbacks are called by collectd unless some perl plugin registers that callback.

These changes are mostly backward-compatible for perl plugins, with only difference for `flush` operation.
Before that, all perl plugins was 'invisible' by Collectd core, all callbacks was from 'perl' plugin.
So, to make explicit flush, one should flush 'perl' plugin. After this patch, each perl plugin is registered at Collectd level with its own name, so calls to flush `perl` plugin needs to be changed to flush that plugin name as it was registered by perl plugin code. I see no problem in this, though.

Also added `G_EVAL` to call_pv() so 'die' crash in perl code will not crash Collectd completely.

This PR (branch) also has bugfix patches introduced in #1710. 
I will rebase this PR after them will be merged to master branch.

### DIFF
--- a/bindings/perl/lib/Collectd.pm
+++ b/bindings/perl/lib/Collectd.pm
@@ -296,6 +296,9 @@ sub plugin_register {
 			return plugin_register_notification($name, $data);
 		}
 		if (TYPE_FLUSH == $type) {
+			#For collectd-5.6 only
+			lock %{$plugins[$type]};
+			$plugins[$type]->{$name} = $data;
 			return plugin_register_flush($name, $data);
 		}
 		lock %{$plugins[$type]};
--- a/bindings/perl/lib/Collectd.pm
+++ b/bindings/perl/lib/Collectd.pm
@@ -180,13 +180,13 @@ sub plugin_call_all {
 	}
 
 	if (TYPE_LOG != $type) {
-		DEBUG ("Collectd::plugin_call: type = \"$type\" ("
+		DEBUG ("Collectd::plugin_call_all: type = \"$type\" ("
 			. $types{$type} . "), args=\""
 			. join(', ', map { defined($_) ? $_ : '<undef>' } @_) . "\"");
 	}
 
 	if (! defined $plugins[$type]) {
-		ERROR ("Collectd::plugin_call: unknown type \"$type\"");
+		ERROR ("Collectd::plugin_call_all: unknown type \"$type\"");
 		return;
 	}
 
--- a/bindings/perl/lib/Collectd.pm
+++ b/bindings/perl/lib/Collectd.pm
@@ -172,7 +172,6 @@ sub plugin_call_all {
 	my $type = shift;
 
 	my %plugins;
-	my $interval;
 
 	our $cb_name = undef;
 
@@ -196,21 +195,9 @@ sub plugin_call_all {
 		%plugins = %{$plugins[$type]};
 	}
 
-	$interval = plugin_get_interval ();
-
 	foreach my $plugin (keys %plugins) {
-		my $p = $plugins{$plugin};
-
-		my $status = 0;
-
-		if ($p->{'wait_left'} > 0) {
-			$p->{'wait_left'} -= $interval;
-		}
-
-		next if ($p->{'wait_left'} > 0);
-
-		$cb_name = $p->{'cb_name'};
-		$status = call_by_name (@_);
+		$cb_name = $plugins{$plugin};
+		my $status = call_by_name (@_);
 
 		if (! $status) {
 			my $err = undef;
@@ -230,23 +217,7 @@ sub plugin_call_all {
 		}
 
 		if ($status) {
-			$p->{'wait_left'} = 0;
-			$p->{'wait_time'} = $interval;
-		}
-		elsif (TYPE_READ == $type) {
-			if ($p->{'wait_time'} < $interval) {
-				$p->{'wait_time'} = $interval;
-			}
-
-			$p->{'wait_left'} = $p->{'wait_time'};
-			$p->{'wait_time'} *= 2;
-
-			if ($p->{'wait_time'} > 86400) {
-				$p->{'wait_time'} = 86400;
-			}
-
-			WARNING ("${plugin}->read() failed with status $status. "
-				. "Will suspend it for $p->{'wait_left'} seconds.");
+			#NOOP
 		}
 		elsif (TYPE_INIT == $type) {
 			ERROR ("${plugin}->init() failed with status $status. "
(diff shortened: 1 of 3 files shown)


tokkee:
I didn't look at the changes yet but the backward incompatible change to FLUSH could (and imho should) be fixed by registering a separate callback that flushes all perl plugins. That'll require to keep track of all registered FLUSH callbacks in the perl plugin.

rpv-tomsk:
> but the backward incompatible change to FLUSH could (and imho should) be fixed by registering a separate callback that flushes all perl plugins. 

Cons:
- Require to implement additional code;
- Double flushing will occur as result;
- Highly possible not used by anyone; 

Are you think this is really required by people?

tokkee:
TBH, I have no idea if this is in use by anyone. Anyway, it's got the potential to break people in a hard to identify fashion and people may not even realize it until it's too late. I don't think the extra code is overly bad but double flushing is a valid argument. I don't have an immediate good idea of how to avoid that :-/

rpv-tomsk:
Rebased against updated master branch.

rpv-tomsk:
`Anyway, it's got the potential to break people in a hard to identify fashion and people may not even realize it until it's too late.`

This change is not required to be backward-compatible as we will release new version of Collectd.
We just need to note this change in Changelog. So people can do change like

-plugin_register (TYPE_WRITE, "FooBar", "foobar_write");
+plugin_register (TYPE_WRITE, "perl", "foobar_write");

And after such simple change, things will continue to work as before upgrade.

That is the main reason why I especially noticed this fact in my comment to this PR.

tokkee:
Backward-compatibility is a concern and it should be maintained across minor versions of collectd (that is, we try not to break backward-compatibility in 5.x). Anyway, usually we solve this by adding a configuration option which decides about the behavior. Thinking about this again, this seems like the right approach here as well. That is, by default, we'd keep the current behavior and maintain compatibility but the option would allow anyone to change it on their own (and they'll then have to make sure to update their plugins as well).

rpv-tomsk:
Please explain me the policy about versions: What is the difference between 5.5.0 -> 5.5.1 version release and 5.5 -> 5.6 version release ?

tokkee:
A collectd version number is <major>.<minor>.<patch-level>. We increment the patch-level for backward and forward compatible bug fix changes, the minor version for backward compatible new features, and the major version for non-backward compatible changes. Non-backward compatible changes will usually need to be avoided. We don't bump the major version lightly and, hence, breaking changes will need to wait a while before being accepted (hence the work-around via configuration flags). For example, version 5.4.1 means major version 5, minor version 4, and patch-level 1. Bug fixes could go into 5.4.2, new features (e.g. new plugins or other, optional features) go into 5.5.0.

What potential problems do you see from your change? Note, that according to the documentation, Perl plugins are already expected to be thread-safe.

rpv-tomsk:
> Bug fixes could go into 5.4.2, new features (e.g. new plugins or other, optional features) go into 5.5.0.

Why when we do not put these into 5.5 (at least, for example):

#1557 Collectd::Unixsock: Split messages  ... (bug)
#1608 xencpu plugin (new plugin)
#1708 Postgresql plugin: set plugin instance from query results (new feature, fully backward compatible).

rpv-tomsk:
> What potential problems do you see from your change? Note, that according to the documentation, Perl plugins are already expected to be thread-safe.

Ok, then we will not worry about people with possible broken plugins.

I will try to propose workaround for flush/write callbacks in perl plugins.

octo:
→ [FAQ: What do the version numbers mean?](https://www.collectd.org/faq.shtml#faq:version_numbers)

All new features go into *master* and will eventually end up in a *feature release*. The next one is going to be 5.6.

Bugfixes should go in the appropriate branch, so if #1557 applies to collectd 5.4, the patch should have gone into the *collectd-5.4* branch. See also: [Repository → Organization](https://collectd.org/wiki/index.php/Repository#Organization)

@rpv-tomsk are you still working on this? Can @tokkee or I help you with this?

Best regards,
—octo

rpv-tomsk:
> Bugfixes should go in the appropriate branch, so if #1557 applies to collectd 5.4, the patch should have gone into the collectd-5.4 branch. See also: Repository → Organization

It is not so clear which branches are supported at the moment.  For example, #1557 may be applied to 4.x branch too, if I remember things correctly.

rpv-tomsk:
>@rpv-tomsk are you still working on this? Can @tokkee or I help you with this?

Currently I have no work in progress, and only some minor amount of time is available at the moment. I think I will have no much time to work on project until September, sorry.  But I'm online from time to time.

octo:
Is the flush issue the only backwards compatibility issue or are there others I've missed?

rpv-tomsk:
The behaviour is changed - plugins implemented in perl are registered in Collectd core directly, not in 'perl' plugin internals. 

- plugin_flush() in plugin.c allows to flush data of explicit plugin, the same as 
plugin_write()  allows to write data to explicit plugin. All perl-implemented plugins was 'accessible' by these functions as single 'perl' plugin.
Now they must be referenced by names, which was passed to Collectd::plugin_register () while registering (like C plugins).

- Also, as result, callbacks are run by multiple threads, which was the goal of the work.  Them was run by one thread before, but according to the documentation, Perl plugins are already expected to be thread-safe.

I see no other issues.

octo:
I'd argue to pursue this then. Install a flush handler for the Perl plugin and document the "double flush" behavior with a promise of this going away at a later version.
—octo

tokkee:
+1 to what @octo said.

rpv-tomsk:
Hi!

The patch for 'flush' handler fallback is ready. Please review.

The 'DisableOldFlush' option should be renamed, please propose names too.

Also the issue was found and fixed by #6a5614c.

rpv-tomsk:
Rebased to top of 'master'.

octo:
Thanks @rpv-tomsk! I'd suggest to call the config option `FlushAll`.

It looks good to me, but before we can merge this, you need to add the new option to the example config `src/collectd.conf.in` and the manpage `src/collectd.conf.pod`. I'd also prefer if @tokkee could take another look as he's more experienced with the *perl plugin* than I am.

rpv-tomsk:
> It looks good to me, but before we can merge this, you need to add the new option to the example config src/collectd.conf.in and the manpage src/collectd.conf.pod.

Could you please provide the text for manpage? Thanks.

octo:
Hi @rpv-tomsk, sorry for not coming back to this earlier.

Looking through the code, please consider renaming the option to *RegisterLegacyFlush*: we discourage negative option names ("Disable…") because double negatives are hard to comprehend (for example "DisableCallback false").

Regarding documentation, how about:

> **RegisterLegacyFlush** **true**|**false**
> The *Perl plugin* used to register one flush callback (called `"perl"`) and call all Perl-based flush handlers when this callback was called. Newer versions of the plugin wrap the Perl flush handlers and register them directly with the daemon *in addition* to the legacy `"perl"` callback. This allows to call specific Perl flush handlers, but has the downside that flushing *all* plugins now calls the Perl flush handlers twice (once directly and once via the legacy callback). Unfortunately, removing the `"perl"` callback would break backwards compatibility.
>
> This option allows you to disable the legacy `"perl"` flush callback if you care about the double call and don't call the `"perl"` callback in your setup.

Best regards,
—octo

rpv-tomsk:
Hi!

Updated.

octo:
The build is currently failing with:

```
utils_format_json_test.c:46:3: error: conflicting types for 'label_t'
```

I believe this was fixed for the 5.6 release, so rebasing on *master* should take care of that.

rpv-tomsk:
Rebased.

octo:
Thanks @rpv-tomsk!

Issue #1732, opened 2016-05-29
Reloading TypesDB by HUP signal.


rpv-tomsk:
I'm going to implement subj.

The way of implementing - merge configuration from files with already loaded configuration.
Datasources number and their names change will not be allowed, but datasource type, min, max values - will be.
If dataset will be removed from files - it will be left unchanged in running configuration.

That is the plan. 
Any suggestions/feedback ?

Found #1241 - @rubenk, would you like to see patch for that ?

rpv-tomsk:
Closed in favour of #1733.

PR #1733, opened 2016-05-29
Reloading TypesDB by HUP signal

rpv-tomsk:
Implemented reloading of TypesDB database by HUP signal.

The way of reload - merge new configuration from TypesDB files with already loaded configuration.

Datasources number change will not be allowed while updating dataset, but datasources name,type, min, max values - will be. If dataset was removed from files - it will be left unchanged in running configuration.

Memory free() calls are not used during reload as that can crash collectd. 
Many plugins call plugin_get_ds() and use that value in the same time as `ds` can be freed, so avoid that completely. The plugin_dispatch_values_internal() and write() functions also uses datasets widely without copying and locks.

This feature is very useful on 'hubs' which collect data from 'nodes'. 
It allows to add new metrics on 'nodes' without 'hub' restart.

Superseedes #1732. 

PS. Also it would be great to remove plugin_register_data_set() / plugin_unregister_data_set() completely. I see no reasons why them should be left alive. What do you think about this?


rpv-tomsk:
Closed in favour of #1733.

tokkee:
I'm worried about not removing datasets if they don't appear in the new configuration. I think that's unexpected behavior and will cause problems down the road. If we want reloading of datasets, we'll have to address the underlying issues imho.

rpv-tomsk:
The problem is: many plugins call  plugin_get_ds() which returns the same pointer to data_set_t as it registered in `data_sets`.

So, the possible ways are:

- rewrite all plugins  - implement plugin_get_ds() which will copy data_sets and require to free() that copied value in plugins code. This solution will break compatibility for people with their own plugins, introduce additional data_set  copy for each plugin_get_ds() call and requires to touch all plugins and review plugin_dispatch_values_internal() flow;
- implement reference counter and garbage collector;
- prohibit to unregister data_sets and avoid free() on registered data_sets;

What do you think about this? 

I prefer last solution.

This feature is highly required on 'hubs'. There is no types deletion in most cases, but even if some types will be deleted in file but left inside running config - I see no problems. If that type was deleted then metrics of that type should not arrive to 'hub' any more.

I will (try to) use this patch in my environment and I'll be happy not to restart Collectd when adding new types.

tokkee:
I do realize that this is a real problem and I agree that introducing a lot of copy operations isn't great :-/ Another option is to move unregistered data-sets out of the AVL tree. Then further lookups will fail but the memory location would still be valid and can be cleaned up on shutdown. That's rather hacky, though.

Maybe @octo, @mfournier, or @rubenk have another idea.

octo:
There are only a handful of plugins actually calling `plugin_get_ds()`, the vast majority relies on the call that's happening in `src/daemon/plugin.c`. To get the `SIGHUP` stuff working, we should return a copy (which needs to be freed by the caller) to plugins and use a non-copying version in `plugin_dispatch_values()` – I guess R/W locks could help with updating types after startup.

rpv-tomsk:
>  To get the SIGHUP stuff working, we should return a copy (which needs to be freed by the caller) to plugins and use a non-copying version in plugin_dispatch_values() – I guess R/W locks could help with updating types after startup.

@octo, What do you think about performance of this solution? I think adding locks and copy operations are not good and too expensive to reach typesdb reloading goals.

>Another option is to move unregistered data-sets out of the AVL tree. Then further lookups will fail but the memory location would still be valid and can be cleaned up on shutdown.

I would prefer this solution in this case.

PR #1735, opened 2016-05-31
network plugin: Improved error logging on decrypt error (wrong/missing username/password)

rpv-tomsk:
Hi!

Yesterday I found the following repeating in my collectd hub syslog:

```
May 30 19:59:16 SERVER collectd[31844]: network plugin: Decryption failed: Checksum mismatch.
May 30 19:59:16 SERVER collectd[31844]: network plugin: Decrypting AES256 part failed with status -1.
```

These messages are not informative enough. So, introduce to you the PR about logging improve.

Before:

When 'No such user':

`May 30 19:59:16 SERVER collectd[31844]: network plugin: Decrypting AES256 part failed with status -1.`

When 'Wrong password':

```
May 30 19:59:16 SERVER collectd[31844]: network plugin: Decryption failed: Checksum mismatch.
May 30 19:59:16 SERVER collectd[31844]: network plugin: Decrypting AES256 part failed with status -1.
```

After:

No such user:

`[2016-05-31 01:24:31] network plugin: Decryption failed: Failed to get cypher. Username: webserver03`

Wrong password:

`[2016-05-31 01:25:32] network plugin: Decryption failed: Checksum mismatch. Username: webserver03`

The message 'Decrypting AES256 part failed with status -1' is hidden in favour of new error messages in parse_part_encr_aes256().

This patch helps me to find which server has wrong username configured (using tcpdump + grep by found username).

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -1142,7 +1142,7 @@ static int parse_part_sign_sha256 (sockent_t *se, /* {{{ */
   if (memcmp (pss.hash, hash, sizeof (pss.hash)) != 0)
   {
     WARNING ("network plugin: Verifying HMAC-SHA-256 signature failed: "
-        "Hash mismatch.");
+        "Hash mismatch. Username: %s", pss.username);
   }
   else
   {
@@ -1226,8 +1226,8 @@ static int parse_part_encr_aes256 (sockent_t *se, /* {{{ */
   /* Make sure at least the header if available. */
   if (buffer_len <= PART_ENCRYPTION_AES256_SIZE)
   {
-    ERROR ("network plugin: Decryption failed: "
-            "Discarding short packet.");
+    NOTICE ("network plugin: parse_part_encr_aes256: "
+        "Discarding short packet.");
     return (-1);
   }
 
@@ -1242,8 +1242,8 @@ static int parse_part_encr_aes256 (sockent_t *se, /* {{{ */
   if ((part_size <= PART_ENCRYPTION_AES256_SIZE)
       || (part_size > buffer_len))
   {
-    ERROR ("network plugin: Decryption failed: "
-            "Discarding part with invalid size.");
+    NOTICE ("network plugin: parse_part_encr_aes256: "
+        "Discarding part with invalid size.");
     return (-1);
   }
 
@@ -1254,19 +1254,15 @@ static int parse_part_encr_aes256 (sockent_t *se, /* {{{ */
   if ((username_len <= 0)
       || (username_len > (part_size - (PART_ENCRYPTION_AES256_SIZE + 1))))
   {
-    ERROR ("network plugin: Decryption failed: "
-            "Discarding part with invalid username length.");
+    NOTICE ("network plugin: parse_part_encr_aes256: "
+        "Discarding part with invalid username length.");
     return (-1);
   }
 
   assert (username_len > 0);
   pea.username = malloc (username_len + 1);
   if (pea.username == NULL)
-  {
-    ERROR ("network plugin: Decryption failed: "
-            "malloc() failed.");
     return (-ENOMEM);
-  }
   BUFFER_READ (pea.username, username_len);
   pea.username[username_len] = 0;
 
@@ -1281,8 +1277,7 @@ static int parse_part_encr_aes256 (sockent_t *se, /* {{{ */
       pea.username);
   if (cypher == NULL)
   {
-    ERROR ("network plugin: Decryption failed: "
-            "Failed to get cypher. Username: %s", pea.username);
+    ERROR ("network plugin: Failed to get cypher. Username: %s", pea.username);
     sfree (pea.username);
     return (-1);
   }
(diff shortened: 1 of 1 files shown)


src/network.c
```diff
@@ -1226,8 +1226,8 @@ static int parse_part_encr_aes256 (sockent_t *se, /* {{{ */
   /* Make sure at least the header if available. */
   if (buffer_len <= PART_ENCRYPTION_AES256_SIZE)
   {
-    NOTICE ("network plugin: parse_part_encr_aes256: "
-        "Discarding short packet.");
+    ERROR ("network plugin: Decryption failed: "
+           "Discarding short packet.");
```
tokkee: Please keep the previous indentation (two extra tabs in this case). Applies to all of your change.

src/network.c
```diff
@@ -1426,12 +1433,7 @@ static int parse_packet (sockent_t *se, /* {{{ */
 			status = parse_part_encr_aes256 (se,
 					&buffer, &buffer_size, flags);
 			if (status != 0)
-			{
-				ERROR ("network plugin: Decrypting AES256 "
-						"part failed "
-						"with status %i.", status);
```
tokkee: Please keep this message around. It'll mean some duplicate log entries but it ensures that we'll get *some* error message even if someone forgets to log the actual error in `parse_part_encr_aes256` when adding further checks in the future.

tokkee:
Looks mostly good. See my inline comments.

Something that occurred to me (but it's not strictly related to your change): These log messages can be triggered by *unauthenticated remote* users. Hence, it allows *anyone* (on the same network) to DoS the server by generating lots of logs.

src/network.c
rpv-tomsk: I disagree. We just need not forget to add the actual error log in parse_part_encr_aes256 and the (-1) message is useless for people.

src/network.c
rpv-tomsk: Added some extra spaces (indent is by spaces there, not by tabs).

src/network.c
tokkee: My point is that errors happen and it'll be annoying if that means that we're missing error messages. Here, this is easy to avoid, hence my suggestion. I agree that the (-1) message is not overly useful but it will still tell us where something is going wrong. To avoid some duplication, I usually use very specific messages in the helper function and a more generic message in the caller to put the error into context. Here, this could be, for example:

    ERROR: network plugin: Package too short (%d bytes); discarding.
    ERROR: network plugin: Decrypting AES256 part failed (status=%d).

That way, you don't need to repeat "Decryption failed" in all error messages. You'll still get the same amount of information plus the safe-guard against a missing log message. If status would always be -1, we can skip it, or we can make it more useful by actually using different error codes.

What do you think?

src/network.c
tokkee: Yeah, tab here means two spaces.

src/network.c
rpv-tomsk: Of course, I still think what developers can be careful enough not to lose ERROR() at parse_part_encr_aes256() but in common cases your approach is much better.

As this type of errors are enough rare (will not count DoS attacks) and users/sysadm can fix that 'easily enough', I agree what your proposal is a good compromise. But there is more complicated case, at least one, see below.

What do you think about logging-level of these messages? 
Some messages (first three at patch) was switched from NOTICE to ERROR level as '-1' ERROR() call proposed to be removed at  upper level in parse_packet().
Should we leave these messages at ERROR level or lower them back to NOTICE?
And what is about log level of rest messages in parse_part_encr_aes256()?

P.S. More complicated case:

`Jun  5 11:37:00 SERVER collectd[24819]: uc_update: Value too old: name = dev/gate-all/gate_octets-all; value time = 1465123010.377; last cache update = 1465123020.376;
Jun  5 11:37:00 SERVER collectd[24819]: Filter subsystem: Built-in target `write': Dispatching value to all write plugins failed with status -1.
Jun  5 11:37:00 SERVER collectd[24819]: Filter subsystem: Built-in target `write': Some write plugin is back to normal operation. `write' succeeded.
`
- network packets come out of order.
- uc_update() generates first line;
- 'rrdtool' plugin has similar check while updating its cache, but silently returns (-1) in that case;
- core treat this as 'plugin failed' and logs second line;
- later metrics are written successfully, so third line appear in logs.;

tokkee: I was wondering about NOTICE vs ERROR as well. I don't have a strong opinion but I think using ERROR for all messages makes sense, since they are related to the ERROR message at the higher level (the status=-1 message).

For the more complicated case you mention, I think that's a separate issue. To some extent, it's working as expected but there are two cases we might want to avoid: (1) network packages arrive out of order; (2) there's a transient problem and collectd spams logs for that period of time. Anyway, we should discuss that separately.

rpv-tomsk:
Minimized number of proposed changes: Only add username to log output:

* Restored ERROR() in parse_packet()
* Replaced ERROR by NOTICE back.
- Removed ERROR() on pea.username malloc() as it has its own status code.

I can squash changes if necessary.

octo:
LGTM

octo:
Rebased on *master* and committed as 3e2272cea52a2dace364c82fbcfeecf9dd9fbd34 and 83d2ff540ffaa5ca8cf3f37ea428d4b4b8c8d875.

PR #1736, opened 2016-05-31
Port to Sailfish OS

rinigus:
This is a pull request introducing Sailfish OS port of collectd. The port is described in included README.md. The implementation details and possible problems are given below as well (same text as in the corresponding section in readme). 

Looking forward to your feedback.

Implementation details:

The changes in the code were done in src/daemon/collectd.c . Instead
of calling "do\_loop", glib main loop was created and used. By using
keepalive library ( https://git.merproject.org/mer-core/nemo-keepalive
), background activity interval was specified and, through glib, its
calling the callback function "do\_shot". In the end of "do\_shot", a
continuation of the loop is requested by calling
"background\_activity\_wait". Note that after calling
"background\_activity\_wait", device could suspend again.

The initialization, signal callbacks, and cleanup on shutdown were
rewritten as well. To avoid compilation warnings, "do\_loop" is
preprocessed out if HAVE\_KEEPALIVE\_GLIB is defined.

As far as I understand, the main problem with the implementation is
caused by multi-threaded approach to reading the data by
collectd. From reading the code, I am not sure whether after exiting
plugin\_read\_all all plugins have finished reading. For fully correct
implementation, that should be the case. If it is not the case, it
would be great if someone with knowledge of collectd would step in and
help to implement this call as needed.

In practice, on Nexus 4 SFOS port, every wakeup takes at least 6
seconds and within that time all plugins have been updated, as far as
I could see from the graphs.


rinigus:
@octo: I guess, this is too big to be realistically included into 5.7 . Unless you guys looked into it already. 

As far as I can see, there are maybe up to 500-600 installations of it and I have not heard any complains regarding the collectd process. Anyway, collectd is used by quite a few people now to check the device performance and debug the problems with it.

PR #1737, opened 2016-06-03
Pthread cleanup

rubenk:


configure.ac
```diff
@@ -1603,45 +1603,21 @@ if test "x$with_libhal" = "xyes"; then
...
-		with_libpthread="yes"
-	else
-		if test "x$withval" = "xno"
-		then
-			with_libpthread="no (disabled)"
-		fi
-	fi
-], [with_libpthread="yes"])
-
 AC_SUBST([PTHREAD_LIBS])
```
tokkee: Nit: While it works either way, I think it's cleaner to have AC_SUBST() after defining and populating the underlying variable.

configure.ac
```diff
@@ -1603,45 +1603,21 @@ if test "x$with_libhal" = "xyes"; then
...
-else
-	collect_pthread=0
-fi
-AC_DEFINE_UNQUOTED(HAVE_LIBPTHREAD, [$collect_pthread],
-	[Wether or not to use pthread (POSIX threads) library])
-AM_CONDITIONAL(BUILD_WITH_LIBPTHREAD, test "x$with_libpthread" = "xyes")
-# }}}
+AC_CHECK_HEADERS([pthread.h],
+  [],
+  [AC_MSG_ERROR([pthread.h not found])]
```
tokkee: We've got some more '#if HAVE_PTHREAD_H' which could now go away (plugin.h, common.h).

tokkee:
I wonder if the plugins should actually still be linked against pthread. It's not needed on Linux but I seem to recall some weird systems which might need that.

Otherwise LG.

configure.ac
rubenk: Ok, will fix.

configure.ac
rubenk: indeed, the include guard makes no sense now since you can't build without pthread.h. Will fix.

rubenk:
I tested Solaris 10 and FreeBSD, it's all good there.

If this breaks on other systems, their runtime loader is broken. Not that that would surprise me ;)
I have no plans to backport this so we'll find out soon enough.

Thanks for the review.

PR #1738, opened 2016-06-04
src/Makefile: Support silent build rules for pod2man.

tokkee:


rubenk:
LGTM. I'll merge when Jenkins is done with the current builds.

PR #1739, opened 2016-06-04
unit tests: Run all tests through Valgrind's memcheck, if available.

tokkee:
Also, fix all problems identified that way.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -58,7 +58,6 @@ AC_PROG_LEX
 AC_PROG_YACC
 
 AC_PATH_PROG([VALGRIND], [valgrind])
-AC_SUBST([VALGRIND])
 
 # Warn when pkg.m4 is missing
 m4_pattern_forbid([^_?PKG_[A-Z_]+$],[*** pkg.m4 missing, please install pkg-config])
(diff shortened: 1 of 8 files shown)


configure.ac
```diff
@@ -57,6 +57,9 @@ AC_PROG_LIBTOOL
 AC_PROG_LEX
 AC_PROG_YACC
 
+AC_PATH_PROG([VALGRIND], [valgrind])
+AC_SUBST([VALGRIND])
```
rubenk: Unless I'm mistaken AC_PATH_PROG already calls AC_SUBST on the variable.

rubenk:
One minor nit, otherwise nice!

configure.ac
tokkee: It does, indeed. Fixed here and in two other cases.

tokkee:
Ftr, we already install Valgrind in the test VMs used by Jenkins.

tokkee:
That test failure is addressed in 9d0f2b090ebdbe151360ba69e05d9c60c0930ac7.

rubenk:
Thanks!

PR #1740, opened 2016-06-04
Append -Werror to the right flags

rubenk:


rubenk:
@tokkee, I assume this was your intention initially?

tokkee:
Oops, yes. Copy&paste error, I suppose.

PR #1744, opened 2016-06-05
swrite should never have to handle a negative fd

rubenk:


src/daemon/common.c
```diff
@@ -270,6 +270,8 @@ ssize_t swrite (int fd, const void *buf, size_t count)
 	ssize_t     status;
 	struct      pollfd pfd;
 
+	assert (fd >= 0);
```
tokkee: Imho, an assertion is not right. It's meant to catch programming errors in local code but not to validate arguments of exported functions.

rubenk: I was thinking calling swrite with a negative fd in the callback struct is a programming error.
Meanwhile it helped me catch the bug I was looking for.

I'll turn this into a negative return instead.

PR #1745, opened 2016-06-05
write_graphite: only flush data if socket is open

rubenk:
wg_callback_free can be called with a negative fd,
in which case we try to flush data on a closed socket.

CID 116310

### DIFF
--- a/src/write_graphite.c
+++ b/src/write_graphite.c
@@ -297,10 +297,9 @@ static void wg_callback_free (void *data)
 
     pthread_mutex_lock (&cb->send_lock);
 
-    wg_flush_nolock (/* timeout = */ 0, cb);
-
     if (cb->sock_fd >= 0)
     {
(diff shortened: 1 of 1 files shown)


src/write_graphite.c
```diff
@@ -297,10 +297,9 @@ static void wg_callback_free (void *data)
 
     pthread_mutex_lock (&cb->send_lock);
 
-    wg_flush_nolock (/* timeout = */ 0, cb);
-
     if (cb->sock_fd >= 0)
     {
+        wg_flush_nolock (/* timeout = */ 0, cb);
```
tokkee: I think a check in `wg_send_buffer` makes even more sense. That is, check cb->sock_fd where it's actually used instead of letting callers worry about it.

rubenk: Even better, I'll do that, thanks.

rubenk: https://github.com/collectd/collectd/pull/1747

Issue #1748, opened 2016-06-05
libcollectdclient and the core define conflicting types


tokkee:
For example, both places define `value_t` with slightly different sub-types. This makes it impossible to use the client library and anything depending on `plugin.h` at the same time. For example, we couldn't have a plugin using the client lib and it's very hard to have common code between clients and the core without duplicating some of the core types yet another time.

I see these two solutions:
  * Have a common header, say `types.h` which provides all core types in one place, then use it from wherever it's needed (e.g. `plugin.h` and `client.h`). This header should not depend on anything else and any changes to it may affect the library API/ABI (hence, extra care will be required going forward).
  * Use different names for the types in the client lib. For example, use the `lcc_` prefix consistently.

I'd prefer the former.

Both of these changes would mean a library API change but I think that's justified (we may then want to clean up some other stuff as well, e.g. I noticed that `lcc_flush()` only accepts an integer timeout while the daemon support sub-precision resolution or we may want to use `char *` instead of `char[LCC_NAME_LEN]`

rubenk:
I'd prefer a common header as well. I'd say go for it 👍

PR #1749, opened 2016-06-05
Add a generic parser for the text protocol

tokkee:
The idea is to reuse the parser not only in the backend but also in client tools.

Using it for collectdctl is currently blocked by #1748.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -56,7 +56,8 @@ libcmds_la_SOURCES = utils_cmds.c utils_cmds.h \
 		utils_cmd_listval.c utils_cmd_listval.h \
 		utils_cmd_putval.c utils_cmd_putval.h \
 		utils_parse_option.c
-libcmds_la_LIBADD = daemon/libcommon.la daemon/libplugin_mock.la daemon/libmetadata.la
+libcmds_la_LIBADD = daemon/libcommon.la daemon/libmetadata.la \
+		daemon/libplugin_mock.la -lm
 check_PROGRAMS += test_utils_cmds
 TESTS += test_utils_cmds
 test_utils_cmds_SOURCES = utils_cmds_test.c testing.h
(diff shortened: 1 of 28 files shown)


tokkee:
As a side-effect, behavior of the existing commands changes a little:

  * `PUTVAL` no longer submits any values if parsing of the command fails. Previously, it would submit everything up to the first parse error.
  * `"foo=a string with spaces or \\ special characters"` is now valid syntax for specifying options in addition to foo=`"value with spaces, etc."`.
  * `foo=` is now a valid option even without quotes around the empty value.

I'm a little undecided about the first item. I consider it a bug fix but strictly-speaking it's backward-incompatible. We could preserve backward compatibility by adding a "partial failure" mode to the parser where a partial result would be returned even upon failure. That's kinda ugly, though.

rubenk:
As for the first side-effect, strictly speaking this indeed breaks compatibility but I'm not too worried about fixing it. I don't consider PUTVAL submitting values for invalid commands part of the API, it's more like undefined behaviour, and people shouldn't depend on that working.

As for the test failure, I think you'll need to link the tests with libm on FreeBSD:

/jenkins/workspace/pull-requests-build-on-freebsd10-amd64-with-default-toolchain/collectd-5.5.1.1111.g9a779b0/src/utils_cmd_flush.c:104: undefined reference to `__isfinite'
collect2: error: ld returned 1 exit status
*** Error code 1

I've been meaning to look into adding an autoconf check for libm, but since isfinite can be a macro this is a bit harder than I thought.

tokkee:
Ack, that's basically been my thought. Thanks.

octo:
> I don't consider PUTVAL submitting values for invalid commands part of the API, it's more like undefined behaviour, and people shouldn't depend on that working.

+1

octo:
@tokkee could you rebase this branch on *master* please so it can be merged?

tokkee:
Done :-)

tokkee:
This should be good to go now.

octo:
Thanks @tokkee 😃

PR #1750, opened 2016-06-05
daemon: Let all command line options override global options from the config.

### LINKED ISSUE #366: command line options should override configuration file
Example:

```
# /usr/sbin/collectd -C /etc/collectd.conf -P /tmp/collectd.pid
[...]
fopen (/var/run/collectd.pid): Permission denied
```

tokkee:
This replaces 09fb81e130ad70bf0547cec3d3c349ce8cb092cf which provided a specific solution for the `PIDFile` option. The new solution is generic and more nicely abstracted without the need for global variables across multiple modules.

Fixes #366

### DIFF
--- a/src/daemon/collectd.c
+++ b/src/daemon/collectd.c
@@ -54,7 +54,6 @@
  */
 char hostname_g[DATA_MAX_NAME_LEN];
 cdtime_t interval_g;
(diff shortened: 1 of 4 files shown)


rubenk:
LGTM, thanks!

Issue #1059, opened 2015-06-03
ETHSTAT: space in rrd names for some hosts only


TomGudman:
Hi,

The following behaviour is observed with collectd 5.4.2 and 5.5.0 (client and/or server). For an unknown reason yet it seems to only affect our database servers.

As you will see below, the path to the RRD file contains spaces.

```
[2015-06-03 16:43:42] uc_update: Value too old: name = xxx-mysqlxx.sub.domain.com.au/ethstat-eth0/derive-  TSO pkts tx; value time = 1433313813.000; last cache update = 1433313813.000;
[2015-06-03 16:43:43] uc_update: Value too old: name = db1.test.foo.bar.domain.com.au/ethstat-eth0/derive-  bcast bytes rx; value time = 1433313822.996; last cache update = 1433313822.996;
[2015-06-03 16:43:43] uc_update: Value too old: name = db1.test.foo.bar.domain.com.au/ethstat-eth0/derive-  pkts rx OOB; value time = 1433313822.996; last cache update = 1433313822.996;
[2015-06-03 16:43:43] uc_update: Value too old: name = db1.test.foo.bar.domain.com.au/ethstat-eth0/derive-  pkts rx err; value time = 1433313822.996; last cache update = 1433313822.996;
[2015-06-03 16:43:43] uc_update: Value too old: name = xxx-mysqlxx.sub.domain.com.au/ethstat-eth0/derive-  ucast bytes tx; value time = 1433313823.000; last cache update = 1433313823.000;
[2015-06-03 16:43:43] uc_update: Value too old: name = xxx-mysqlxx.sub.domain.com.au/ethstat-eth0/derive-  pkts linearized; value time = 1433313823.000; last cache update = 1433313823.000;
[2015-06-03 16:43:43] uc_update: Value too old: name = xxx-mysqlxx.sub.domain.com.au/ethstat-eth0/derive-  LRO byte rx; value time = 1433313823.000; last cache update = 1433313823.000;
[2015-06-03 16:43:43] uc_update: Value too old: name = db2.test.foo.bar.domain.com.au/ethstat-eth0/derive-  giant hdr; value time = 1433313813.095; last cache update = 1433313813.095;
[2015-06-03 16:43:43] uc_update: Value too old: name = db2.foo.bar.domain.com.au/ethstat-eth0/derive-  pkts linearized; value time = 1433313813.392; last cache update = 1433313813.392;
[...]

PR #1756, opened 2016-06-08
Bug#1059 : Remove any white spaces in ethstats stat names 

TomGudman:
Please review, I am a github rookie ;-)

### DIFF
--- a/src/ethstat.c
+++ b/src/ethstat.c
@@ -311,9 +311,22 @@ static int ethstat_read_interface (char *device)
 
   for (i = 0; i < n_stats; i++)
   {
-    const char *stat_name;
+    char *stat_name;
 
     stat_name = (void *) &strings->data[i * ETH_GSTRING_LEN];
+    /* Remove leading spaces in key name */
+    while (isspace ((int) *stat_name))
+        stat_name++;
+    /* Replace any extra spaces in the name by underscores */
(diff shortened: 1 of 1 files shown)


TomGudman:
See #1059

rubenk:
Thanks for the PR Tom!

While I agree that we should strip leading spaces, which essentially fixes your issue, I don't think we should replace all spaces with underscores. That is a change in behaviour and might break existing setups. Otherwise it looks good.

TomGudman:
I decided to also replace spaces with underscores because all other servers I had with "real" NICs heavily used underscores. I agree that it may break something somewhere but using spaces in filename is bad practice IMO. Also, maybe we could make this 'space-to-underscore' option subject to a boolean config flag in <Plugin ethstats> but that may complexify the code and it's above my limited C skills ;-) FWIW, the above code is running fine on 400+ hosts. 

Decision is up to you ;-) I don't mind either way since we package collectd locally so we can always include our own patches if required. 

Cheers,
Thomas

rubenk:
Hey Thomas,

I'd rather have underscores than spaces too, but I'd rather not break backwards compatibility now.
I've merged only the first part into collectd-5.4, collectd-5.5 and master as 21602de.

Thanks for your contribution!

PR #1765, opened 2016-06-15
write_kafka: ensure key is properly hashed to select partition

vincentbernat:
Otherwise, partition is selected solely on the first letter of the
key. If all your hosts are starting with the same letter, everything
ends up in a single partition.

cc @pyr


rubenk:
Thanks Vincent!

PR #1770, opened 2016-06-21
smart plugin: add IgnoreSleepMode option to ignore sleeping disks

swt2c:
This option enables the smart plugin to use disks that libatasmart mistakenly
reports as asleep.  This happens because libatasmart has not been updated to
incorporate support for newer idle states in the ATA spec.


rubenk:
Thanks for your contribution!

The code looks good, can you add this setting to the manpage? https://github.com/collectd/collectd/blob/master/src/collectd.conf.pod

swt2c:
Done.

rubenk:
Thank you.

PR #1774, opened 2016-06-29
rdkafka is called librdkafka

mmckinst:
The [spec file for
librdkafka](https://github.com/edenhill/librdkafka/blob/master/packaging/rpm/librdkafka.spec)
calls its package 'librdkafka-devel', not 'rdkafka-devel'. Searching the
web for 'rdkafka-devel RPM' turns up nothing, all the RPMs are named
'librdkafka-devel'.


rubenk:
Thanks for your contribution @mmckinst!

PR #1783, opened 2016-07-07
ntpd: Check STA_NANO in the server status

matwey:
When STA_NANO is set in ik->status, then results are
expressed in nanoseconds (instead of microseconds by default).

Signed-off-by: Matwey V. Kornilov <matwey.kornilov@gmail.com>

### DIFF
--- a/src/ntpd.c
+++ b/src/ntpd.c
@@ -51,6 +51,10 @@
 # include <poll.h>
 #endif
 
+#ifndef STA_NANO
+# define STA_NANO 0x2000
+#endif
+
 static const char *config_keys[] =
 {
 	"Host",
@@ -910,8 +914,7 @@ static int ntpd_read (void)
 	int i;
 
 	/* On Linux, if the STA_NANO bit is set in ik->status, then ik->offset
-	 * is is nanoseconds, otherwise it's microseconds.
-	 * TODO(octo): STA_NANO is defined in the Linux specific <sys/timex.h> header. */
+	 * is is nanoseconds, otherwise it's microseconds. */
 	double scale_loop  = 1e-6;
 	double scale_error = 1e-6;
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -755,6 +755,9 @@ AC_CHECK_HEADERS(net/pfvar.h,
 have_termios_h="no"
 AC_CHECK_HEADERS(termios.h, [have_termios_h="yes"])
 
+# For the ntpd plugin
+AC_CHECK_HEADERS(sys/timex.h)
```
rubenk: Could use AC_CHECK_HEADERS_ONCE here?

src/ntpd.c
```diff
@@ -936,6 +942,11 @@ static int ntpd_read (void)
 		return (-1);
 	}
 
+	if (ntohs(ik->status) & STA_NANO) {
```
rubenk: Is this bit guaranteed to be unset on other platforms than Linux?

rubenk:
Thanks for your contribution @matwey!
I left some minor comments inline.

src/ntpd.c
matwey: No, it isn't. ntpd just directly sends the struct obtained from ntp_adjtime syscall.

rubenk: Ok, in that case could you wrap it in a #if KERNEL_LINUX?

matwey: I can, however I think this has a little sense. In general, ```ntpd``` and ```collectd``` may run on different distributed systems. One may monitor ```ntpd``` running on Linux, using his ```collectd``` running on FreeBSD host.

rubenk: You're right.

But this leads me to another issue. What if you're monitoring ntpd running on another platform from collectd running on Linux? If there's no guarantee that 0x2000 stands for the resolution, then this will do the wrong thing.

rubenk: I did some digging, and ntpdc itself always assumes STA_NANO sets resolution (https://github.com/ntp-project/ntp/blob/stable/ntpdc/ntpdc_ops.c#L2946). It's been in ntp since 1999. It's probably safe to drop the check for timex.h and always define STA_NANO as 0x2000.

matwey: Ok, I've rebased this request.

PR #1785, opened 2016-07-08
disk plugin: release udev-based name. Fixes #1781

### LINKED ISSUE #1781: disk plugin dumps core if disk is ignored and conf_udev_name_attr == NULL
In disk plugin, under Linux, a core dump is triggered due to sfree(output_name) at certain conditions. Namely, if HAVE_LIBUDEV is defined, but the alternative name is not received, output_name is still freed. The corresponding snippet of the code from disk.c with additional DEBUG statements is shown below:
```
#if HAVE_LIBUDEV
		char *alt_name = NULL;
		if (conf_udev_name_attr != NULL)
		{
			alt_name = disk_udev_attr_name (handle_udev, disk_name, conf_udev_name_attr);
			if (alt_name != NULL)
			  output_name = alt_name;
			else
			  DEBUG ("disk plugin: udev alt_name was NULL");
		}
		else
			  DEBUG ("disk plugin: udev conf_udev_name_attr was NULL");
#endif

		if (ignorelist_match (ignorelist, output_name) != 0)
		{
			DEBUG ("disk plugin: ignoring %s %p",
			       output_name, output_name);
#if HAVE_LIBUDEV
			/* release udev-based alternate name, if allocated */
			sfree (output_name);
#endif
			continue;
		}
		else	DEBUG ("disk plugin: writing %s",
					output_name);
```
Notice that the core dump does point to the call of free() with output_name address as an argument. The address matches the one printed out by DEBUG statement

rinigus:
Fixes a core dump induced by freeing a wrong pointer when UDEV alt_name was not allocated (for some reason) earlier

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -896,7 +896,7 @@ static int disk_read (void)
 		{
 #if HAVE_LIBUDEV
 			/* release udev-based alternate name, if allocated */
-			sfree (output_name);
+			sfree (alt_name);
 #endif
 			continue;
 		}


rubenk:
@rpv-tomsk please have a look.

rubenk:
This is obviously correct, and my mistake. Thanks for the patch @rinigus.

PR #1787, opened 2016-07-09
This single line would echo the entire perl file into the syslog as it ran.

frankhereford:
Tiny little change. I was working with the version of collectd being distributed with Ubuntu Xenial, and I wrote a perl plugin, (am I the last perl developer left?!?!), and nothing I could would prevent the daemon from echoing every line of the program as it executed into the syslog. This turns into gigs of data quick.  A little investigation, and I came up with this line. So, I went on and got the latest and greatest source, download the bajillion libraries required to build this thing up correctly, since I knew I would be using this home built version going forward, and commented out this line. Low and behold, no more abhorrently full syslog. This is just an old debug statement that hung around from the 2009 development of the perl bindings I believe.


rubenk:
You might indeed be the last perl developer left ;)
I haven't actually used the perl plugin, but looking at the code log_debug is only supposed to log when collectd is compiled with --enable-debug, and then only if LogLevel is set to debug.

Is this not the case?

frankhereford:
There's always going to be a holdout for ol' perl. I have a pretty alright ZFS plugin cooked up that might make it worth buffing up the bindings after all these years, which is something that I'll take on. But first things first -- I will compile up the code again and confirm very carefully assertions to your questions and get back to you. 

Thank you for getting back to me, and I will let you know what I find.

octo:
Hi @frankhereford, thank you very much for your patch!

I think @rubenk is correct: this message is only compiled in when configured with `--enable-debug` and only logged when the log level is set to `debug`.

That said, I'm a bit surprised this logs the Perl script. I was expecting this to only log the command line flags passed to the interpreter …

frankhereford:
Hi, I have not forgotten to come back to this at all; on the contrary, I've written two more quickie perl collectd plugins so I'm totally vested in this issue. Anyway, the problem I was having is with the version that ships with 16.04 ubuntu currently; however, I have been *slammed* with work and a tough class. The second I get my head above water (think Thursday), I'm going create a vanilla 16.04 VM to isolate and test this out. Have not forgotten! Thanks for the comment.

octo:
Hi @frankhereford, do you have any update on this issue?

rubenk:
Closing stale issue.

PR #1792, opened 2016-07-13
Truncate strjoin arguments if > DATA_MAX_NAME_LEN

jconroy77:
If strjoin() runs out of buffer space when joining strings, use the
remainder of available space rather than skipping additional strings.

### DIFF
--- a/src/daemon/common.c
+++ b/src/daemon/common.c
@@ -366,10 +366,12 @@ int strjoin (char *buffer, size_t buffer_size,
 			ptr += sep_len;
 			avail -= sep_len;
 		}
+		if (avail == 0)
+			return (-1);
 
 		field_len = strlen (fields[i]);
 		if (avail < field_len)
-			return (-1);
+			field_len = avail;
 
 		memcpy (ptr, fields[i], field_len);
 		ptr += field_len;
@@ -377,7 +379,7 @@ int strjoin (char *buffer, size_t buffer_size,
 	}
 
 	assert (buffer[buffer_size - 1] == 0);
-	return ((int) strlen (buffer));
+	return ((int) ((buffer_size - 1) - avail));
 }
 
 int escape_string (char *buffer, size_t buffer_size)


src/daemon/common.c
```diff
@@ -361,13 +361,14 @@ int strjoin (char *buffer, size_t buffer_size,
...
-		field_len = strlen (fields[i]);
-		if (avail < field_len)
+		if (avail == 0)
 			return (-1);
 
-		memcpy (ptr, fields[i], field_len);
+		field_len = strlen (fields[i]);
+
+		memcpy (ptr, fields[i], ((avail < field_len) ? avail : field_len));
 		ptr += field_len;
```
portante: Should ptr be incremented by the length actually copied?

So perhaps we could have:

```
field_len = strlen (fields[i]);
copy_len = ((avail < field_len) ? avail : field_len);
memcpy (ptr, fields[i], copy_len);
ptr += copy_len;
avail -= copy_len;
```

jconroy77: Figured if avail was used instead of fieldlen, we've run out of space
anyway (we're either done concating or avail will be 0 and loop will
return(-1) when attempting the next string just as before).  Doesn't hurt
to improve the completeness and readability, though...

--John

portante:
Additionally, in that function, instead of returning `strlen(buffer)`, since `avail` is accurate, would could avoid a `strlen()` and just return `buffer_size - avail`.

jconroy77:
@portante,

I'm revising the patch slightly to make things more clear.

I think the return length should be `(buffer_size - avail - 1)` to account for the NULL terminator (avail is initialized to `buffer_size - 1`), no?

src/daemon/common.c
```diff
@@ -360,18 +360,20 @@ int strjoin (char *buffer, size_t buffer_size,
...
+			field_len = avail;
 
 		memcpy (ptr, fields[i], field_len);
 		ptr += field_len;
 		avail -= field_len;
 	}
 
 	assert (buffer[buffer_size - 1] == 0);
-	return ((int) strlen (buffer));
+	return ((int) (buffer_size - avail - 1));
```
portante: I think we might want, `((buffer_size - 1) - avail)` so it is clear we are subtracting the ending value of avail from its beginning value.

jconroy77:
If that seems acceptable, I'll squash the commits down and force push...

portante:
Nice.

octo:
Hi @jconroy77,

thank you very much for your patch! I'm not quite sure which problem you're trying to fix with this patch though. #4?

In principle I think this is a reasonable change: the *strjoin()* is primarily used to fill the *type_instance* field, which is filled via *sstrncpy()* in many other places. So truncating behavior is not new for these fields. I'd like to ensure that all callers can handle this truncation gracefully.

Best regards,
—octo

jconroy77:
Yes, had that same problem...

The issue I had run into was while monitoring databases using the exec plugin and generating type_instances using the table names.  The resulting type/type_instance string was occasionally exceeding the DATA_MAX_NAME_LEN and the strjoin would fail resulting in values from 2 different DB tables being stored in the same storage on the backend.

Truncation could also result in duplicate type/type_instance combinations, but it would be far less likely than simply failing to strjoin the type_instance on at all.

I have applied this patch and run it on our production systems that were affected with no problems to report, but I'm obviously not running all of the available plugins...

octo:
Thanks @jconroy77!

PR #1794, opened 2016-07-15
smart plugin: add option to store data using serial number

swt2c:
Since a device name (e.g., /dev/sda) can change, allow using the serial number
to store SMART data as this will ensure the data for a given drive is always
kept together.


rubenk:
Thanks @swt2c!

PR #1795, opened 2016-07-16
battery plugin: Add StateFS backend.

rinigus:
This is a new plugin that takes battery statistics from StateFS filesystem. This filesystem is used in Sailfish OS to report battery state including charge, current, voltage, time to full (on charging), time to low (on discharge), power consumption, and energy. Its used in Sailfish port for reporting battery stats instead of battery plugin. Since its useful only in the systems with StateFS, the plugin is disabled on default and has to be enabled on configuration by user.


src/statefs_battery.c
```diff
@@ -0,0 +1,144 @@
...
+#include "plugin.h"
+
+#include <stdio.h>
+
+#define STATEFS_ROOT "/run/state/namespaces/Battery/"
+#define BFSZ 512
+
+static int submitted_this_run = 0;
+
+static void battery_submit (char const *type, gauge_t value)
```
rubenk: Please use const char * here.

src/statefs_battery.c
```diff
@@ -0,0 +1,144 @@
...
+  sstrncpy (vl.plugin, "statefs_battery", sizeof (vl.plugin));
+  sstrncpy (vl.type, type, sizeof (vl.type));
+  
+  plugin_dispatch_values (&vl);
+
+  submitted_this_run++;
+}
+
+
+static _Bool getvalue(char const *fname, gauge_t *value, char *buffer, size_t buffer_size)
```
rubenk: Please use const char * here.

src/statefs_battery.c
```diff
@@ -0,0 +1,144 @@
...
+{
+  FILE *fh;
+  if ((fh = fopen (fname, "r")) == NULL)
+    return 0;
+
+  if ( fgets( buffer, buffer_size, fh ) == NULL )
```
rubenk: Please drop the extra whitespace within the brackets.

src/statefs_battery.c
```diff
@@ -0,0 +1,144 @@
...
+    {
+      fclose( fh );
```
rubenk: Same here.

src/statefs_battery.c
```diff
@@ -0,0 +1,144 @@
...
+
+  submitted_this_run++;
+}
+
+
+static _Bool getvalue(char const *fname, gauge_t *value, char *buffer, size_t buffer_size)
+{
+  FILE *fh;
+  if ((fh = fopen (fname, "r")) == NULL)
+    return 0;
```
rubenk: Don't ask, but we actually prefer return (0).

src/statefs_battery.c
```diff
@@ -0,0 +1,144 @@
...
+
+static int battery_read (void)
+{
+  char buffer[BFSZ];
+  gauge_t value = NAN;
+  
+  submitted_this_run = 0;
+  
+
+  if ( getvalue( STATEFS_ROOT "ChargePercentage", &value, buffer, BFSZ ) )
```
rubenk: Please drop the extra whitespace within the brackets within this conditional and the others below.

rinigus:
Thank you very much for a thorough review. I'll fix it and later apply the same fix to CPU sleep plugin.

octo:
Hi @rinigus, thank you very much for your patch!

I think it would make sense to expose this functionality via the *battery plugin*. The code could stay in its own file (though I'd vote for renaming it to `battery_statefs.c`). The code could try to *stat* `STATEFS_ROOT` and report the metrics if StatFS is available. Alternatively, we could add an option *ReportStatFS* for users to explicitly configure this source.

Best regards,
—octo

rinigus:
The problem with this solution is that I have 3 more statefs_* plugins that I would like to submit as well (waiting till the first one goes through). In this sense, it would be good to group them together. Namely, SailfishOS is probably the only one that has this system and that has collectd working on it. (this filesystem might be on Tizen and maybe some others, but I don't know about it). So, it would be used (and is used) only by Sailfish users. In this sense, there is no point in having an overhead for all others and that's why I also set the default in configure as disabled. From configuration point of view and maintenance its easier to have them separately, I think. At least its easier to have a small statefs_battery than hoop through all options in battery :)

Other statefs plugins are statefs_cellular (shows which network you are connected to gsm, 4g, ... and the corresponding network signal strengt), statefs_bluetooth (switched on, visible, connected), statefs_internet (signal strength for used internet connection, wifi or cellular).

Let me know what you think.

cheers,

rinigus

octo:
I think having a *StateFS* option with a default value of *false* would not be any overhead for users on other systems. The code being executed on every iteration would end up being something like this:

```c
if (query_statefs) {
  status = read_statefs ();
  /* potentially report error */
}
```

Where the majority of users will skip over the if-block.

As an alternative, if these are all specific to *SailfishOS*, I could picture a *sailfish plugin* with options like *ReportBattery <yes|no>*, *ReportNetwork <yes|no>*, …

I would like to avoid having the method by which metrics are read in the plugin name. Users don't care about that, they want to "have battery metrics" or "get the Sailfish metrics", and the plugin name should reflect that.

Best regards,
—octo

rinigus:
I see the point regarding the metrics. I don't think it would be good to have it as a Sailfish plugin since its possible that statefs may be used in future by others. 

Also, having Sailfish plugin does not seem to be consistent with collectd approach so far (there are no Linux/FreeBDS/... plugins).

With the battery, it is indeed possible to integrate it with the present plugin, agreed. Now others, to my knowledge, are not covered so far. So, while finding solution for this PR, let's keep in mind the others:

https://github.com/rinigus/collectd/blob/sailfish/src/statefs_bluetooth.c
Reported stats are 0 or 1: enabled, visible, connected 

https://github.com/rinigus/collectd/blob/sailfish/src/statefs_cellular.c
Reported: signal_quality %
Type instance is used to indicate the used network technology (2G, 3G, ...)

https://github.com/rinigus/collectd/blob/sailfish/src/statefs_internet.c
Reported: signal_quality %
Type instance is used to indicate the used network technology (WiFi, GPRS)

I have also submitted radio plugin as PR that reports on switched state of radios (wifi, bluetooth, and any other radios as reported by rfkill switch). This one should be of general interest, for Linux laptop users, for example. 

Cheers

rinigus

octo:
Agreed, creating a *sailfish plugin* is only the second best option. I take it we're in agreement with regard to the *battery* code?

The "cellular" and "internet" metrics we could add to the [Wireless plugin](https://collectd.org/wiki/index.php/Plugin:Wireless).

The "radio" and "bluetooth" plugins I'm not so sure about: booleans are not really performance metrics, but they can be interesting nonetheless when averaged over a time span. I'm not quite sure if they warrant their own plugins just for that, though. What's the use case you have in mind with those plugins?

—octo

rinigus:
I think that solution is, indeed, probably the best. I'll try to incorporate the code into _battery_ plugin

_wireless_ plugin I missed, since it didn't work on phone. sorry, I'll look how to add the code there.

_radio_ and _bluetooth_: these metrics are related to battery consumption (radios could be easily one of the main consumers on phone) and having the log allows you to see if the battery consumption was used up by forgotten radio. It also has a security aspect, if you end up having bluetooth connection that you were not aware of. These plugins correspond to what you see on Android phones as lines below Battery charge graph. With collectd + rrdtool, I can also then see averages over longer timespans based on originally boolean stats.

rinigus:
One question: You mentioned that we can keep the file for the battery_statefs separate. How would that work in this case?

octo:
You add the second `.c` file to the `SOURCES` list in `src/Makefile.am`. The *dns plugin* is an example where this is done:

```
dns_la_SOURCES = dns.c utils_dns.c utils_dns.h
```

For the implementation, we need prototypes of all functions that need to be called from the "main" *battery plugin*. The "right" way is to create a separate `battery_statefs.h` header file with all it's bells and whistles. The lazy way is to add the prototypes near the top of the `battery.c` file. In this particular case, I think the lazy way would be acceptable, i.e. I don't expect any re-use of the code in `batter_statefs.c`.

Hope that helps, let me know if I can help more!
—octo

rinigus:
Excellent! I'll look into dns and see. I don't mind doing it correctly, since I plan to submit multiple other codes and would like them to become a part of collectd

octo:
Certainly :) Thanks for bearing with me!

rinigus:
I'll deal with this moving of the code few days later. But the direction is clear now

rinigus:
@octo: sorry for delay. I'll incorporate these changes into battery plugin, as agreed. I'll be off to a trip till September and then deal with this plugin asap.

octo:
@rinigus Thanks and also thanks for the ETA. I'll be back from vacation by September.

rinigus:
@octo: 

I have moved the code and know battery plugin has statefs backend, as you suggested. I run the code through clang as well, so there should be less issues with formatting. Please review, when you have time.

cheers,

rinigus

src/battery_statefs.c
```diff
@@ -0,0 +1,136 @@
...
+  /* statefs supports 1 battery at present */
+  sstrncpy(vl.plugin_instance, "0", sizeof(vl.plugin_instance));
+  sstrncpy(vl.type, type, sizeof(vl.type));
+
+  plugin_dispatch_values(&vl);
+
+  submitted_this_run++;
+}
+
+static _Bool getvalue(const char *fname, gauge_t *value, char *buffer,
```
octo: `buffer` should be a local variable here, not an argument: the calling code is not using its content.

src/battery_statefs.c
```diff
@@ -0,0 +1,136 @@
...
+                      size_t buffer_size) {
+  FILE *fh;
+  if ((fh = fopen(fname, "r")) == NULL)
+    return (0);
```
octo: I think a `WARNING()` would be in order here.

src/types.db
```diff
@@ -219,6 +221,8 @@ swap_io                 value:DERIVE:0:U
 tcp_connections         value:GAUGE:0:4294967295
 temperature             value:GAUGE:U:U
 threads                 value:GAUGE:0:U
+timefull                value:GAUGE:0:U
```
octo: Please use "duration-full" and "duration-low" instead of declaring new types.

src/types.db
```diff
@@ -175,6 +176,7 @@ ping_droprate           value:GAUGE:0:100
 ping_stddev             value:GAUGE:0:65535
 players                 value:GAUGE:0:1000000
 power                   value:GAUGE:0:U
+power_battery           value:GAUGE:U:U
```
octo: Please use the "power" type instead.

src/battery_statefs.c
```diff
@@ -0,0 +1,136 @@
...
+    battery_submit("charge", value);
+  // Use capacity as a charge estimate if ChargePercentage is not available
+  else if (getvalue(STATEFS_ROOT "Capacity", &value, buffer, BFSZ))
+    battery_submit("charge", value);
+
+  if (getvalue(STATEFS_ROOT "Current", &value, buffer, BFSZ))
+    battery_submit("current", value * 1e-6); // from uA to A
+
+  if (getvalue(STATEFS_ROOT "Energy", &value, buffer, BFSZ))
+    battery_submit("energy", value * 1e-6); // from uWh to Wh
```
octo: I'm a bit on the fence with this one, because normally energy is measured in Joule. How about naming this one "energy_wh"?

src/types.db
rinigus: I cannot use power, since the value goes to negative during charging.

octo:
Thanks @rinigus, I left some (easily fixable) comments inline :) This is getting close :-D

src/battery_statefs.c
rinigus: Sure. its just that with batteries they use Wh. Let's use energy_wh and keep energy for the values in Joule.

rinigus:
@octo: Thank you very much! One of them [power] would be have to be fixed differently, but others seem simple. I'll deal with it late tonight.

rinigus:
@octo: I inserted all the proposed changes with the exception of power. At present, power is non-negative in types.db . I need it to be with sign (its negative on charging)

src/types.db
octo: Uhm, we might have to change this then ;)

The existing code already (tries to) submit negative values:

```
[2016-09-07 09:22:43] uc_update: example.com/battery-0/power: ds[0] = -18.249000
```

Can you simply replace the minimum value of "power" with "U"? Thanks!

rinigus: sure! I'll make this change a bit later today

octo: Actually, scratch that, I'm going to make this change in a bugfix branch and merge it to master from there, so older versions get the fix, too.

octo: Change is here: a3444db7419687386919fccf7d14b945f2e9da56

rinigus: Excellent! then I'll cherry-pick that one in my tree. As for this PR, I'll keep power untouched.

rinigus:
@octo: done! now I have to wait till you merge the power bugfix. then everything will be fine :)

octo:
Thanks @rinigus, I'll merge this PR once the fix hits *master*, which shouldn't be long.

rinigus:
Thank you!

octo:
Since the history of this contained so much back and forth and at least one merge, I decided to squash the commits. The changes have been committed as 806091e395886da83acf2435e90648bdf2cf2ed5. Thanks again for all your work! :-)

rinigus:
Thank you very much for helping with it!

PR #1796, opened 2016-07-17
configure: Use AC_MSG_RESULT to print the config summary.

tokkee:
This way, it will show up in config.log as well.


rubenk:
Good idea, thanks :)

PR #1799, opened 2016-07-18
hugepages plugin: a stats plugin for hugepages

maryamtahhan:
Plugin that allows stats for hugepages to be collected.

Change-Id: Ide5ff31c54206cf5ee3d6155899bd53e8edae7bd
Signed-off-by: Jaroslav Safka <jaroslavx.safka@intel.com>
Signed-off-by: Kim Jones <kim-marie.jones@intel.com>
Signed-off-by: Maryam Tahhan <maryam.tahhan@intel.com>

### DIFF
--- a/README
+++ b/README
@@ -128,6 +128,12 @@ Features
     - hddtemp
       Hard disk temperatures using hddtempd.
 
+    - hugepages
+      Report the total number of hugpages, the number of free/reserved
+      hugepages, and the size of the available hugepages. More info on
+      hugepages can be found here:
+      https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt.
+
     - interface
       Interface traffic: Number of octets, packets and errors for each
       interface.
--- a/README
+++ b/README
@@ -129,8 +129,7 @@ Features
       Hard disk temperatures using hddtempd.
 
     - hugepages
-      Report the total number of hugpages, the number of free/reserved
-      hugepages, and the size of the available hugepages. More info on
+      Report the number of used and free hugepages. More info on
       hugepages can be found here:
       https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt.
 
--- a/configure.ac
+++ b/configure.ac
@@ -482,9 +482,6 @@ fi
 # For hddtemp module
 AC_CHECK_HEADERS(linux/major.h)
 
-# For hugepages module
-AC_CHECK_HEADERS(dirent.h)
-
 # For md module (Linux only)
 if test "x$ac_system" = "xLinux"
 then
@@ -5549,6 +5546,7 @@ plugin_entropy="no"
 plugin_ethstat="no"
 plugin_fhcount="no"
 plugin_fscache="no"
+plugin_hugepages="no"
 plugin_interface="no"
 plugin_ipmi="no"
 plugin_ipvs="no"
@@ -5595,6 +5593,7 @@ then
 	plugin_entropy="yes"
 	plugin_fhcount="yes"
 	plugin_fscache="yes"
+	plugin_hugepages="yes"
 	plugin_interface="yes"
 	plugin_ipc="yes"
 	plugin_irq="yes"
@@ -5974,7 +5973,7 @@ AC_PLUGIN([fscache],             [$plugin_fscache],         [fscache statistics]
 AC_PLUGIN([gmond],               [$with_libganglia],        [Ganglia plugin])
 AC_PLUGIN([grpc],                [$with_grpc],              [gRPC plugin])
 AC_PLUGIN([hddtemp],             [yes],                     [Query hddtempd])
-AC_PLUGIN([hugepages],           [yes],                     [Hugepages statistics])
+AC_PLUGIN([hugepages],           [$plugin_hugepages],       [Hugepages statistics])
 AC_PLUGIN([interface],           [$plugin_interface],       [Interface traffic statistics])
 AC_PLUGIN([ipc],                 [$plugin_ipc],             [IPC statistics])
 AC_PLUGIN([ipmi],                [$plugin_ipmi],            [IPMI sensor statistics])
--- a/configure.ac
+++ b/configure.ac
@@ -482,6 +482,9 @@ fi
 # For hddtemp module
 AC_CHECK_HEADERS(linux/major.h)
 
+# For hugepages module
+AC_CHECK_HEADERS(dirent.h)
+
 # For md module (Linux only)
 if test "x$ac_system" = "xLinux"
 then
@@ -5971,6 +5974,7 @@ AC_PLUGIN([fscache],             [$plugin_fscache],         [fscache statistics]
 AC_PLUGIN([gmond],               [$with_libganglia],        [Ganglia plugin])
 AC_PLUGIN([grpc],                [$with_grpc],              [gRPC plugin])
 AC_PLUGIN([hddtemp],             [yes],                     [Query hddtempd])
+AC_PLUGIN([hugepages],           [yes],                     [Hugepages statistics])
 AC_PLUGIN([interface],           [$plugin_interface],       [Interface traffic statistics])
 AC_PLUGIN([ipc],                 [$plugin_ipc],             [IPC statistics])
 AC_PLUGIN([ipmi],                [$plugin_ipmi],            [IPMI sensor statistics])
@@ -6382,6 +6386,7 @@ AC_MSG_RESULT([    fscache . . . . . . . $enable_fscache])
 AC_MSG_RESULT([    gmond . . . . . . . . $enable_gmond])
 AC_MSG_RESULT([    grpc  . . . . . . . . $enable_grpc])
 AC_MSG_RESULT([    hddtemp . . . . . . . $enable_hddtemp])
+AC_MSG_RESULT([    hugepages . . . . . . $enable_hugepages])
 AC_MSG_RESULT([    interface . . . . . . $enable_interface])
 AC_MSG_RESULT([    ipc . . . . . . . . . $enable_ipc])
 AC_MSG_RESULT([    ipmi  . . . . . . . . $enable_ipmi])
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -446,7 +446,6 @@ if BUILD_PLUGIN_HUGEPAGES
 pkglib_LTLIBRARIES += hugepages.la
 hugepages_la_SOURCES = hugepages.c
 hugepages_la_LDFLAGS = $(PLUGIN_LDFLAGS)
-hugepages_la_LIBADD =
 endif
 
 if BUILD_PLUGIN_INTERFACE
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -442,6 +442,13 @@ hddtemp_la_LIBADD += -lsocket
 endif
 endif
 
+if BUILD_PLUGIN_HUGEPAGES
+pkglib_LTLIBRARIES += hugepages.la
+hugepages_la_SOURCES = hugepages.c
+hugepages_la_LDFLAGS = $(PLUGIN_LDFLAGS)
+hugepages_la_LIBADD =
+endif
+
 if BUILD_PLUGIN_INTERFACE
 pkglib_LTLIBRARIES += interface.la
 interface_la_SOURCES = interface.c
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -2704,8 +2704,8 @@ TCP-Port to connect to. Defaults to B<7634>.
 To collect B<hugepages> information, collectd reads directories
 "/sys/devices/system/node/*/hugepages" and
 "/sys/kernel/mm/hugepages".
-Reading of these directories can be disabled by following
-options. Default is enabled.
+Reading of these directories can be disabled by the following
+options (default is enabled).
 
 =over 4
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -2701,21 +2701,27 @@ TCP-Port to connect to. Defaults to B<7634>.
 
 =head2 Plugin C<hugepages>
 
-To get values from B<hugepages> collectd read directories
+To collect B<hugepages> information, collectd reads directories
 "/sys/devices/system/node/*/hugepages" and
-"/sys/kernel/mm/hugepages"
+"/sys/kernel/mm/hugepages".
 Reading of these directories can be disabled by following
 options. Default is enabled.
 
 =over 4
 
-=item B<EnableNuma> I<true>|I<false>
+=item B<ReportPerNodeHP> I<true>|I<false>
 
-Enable/Disable reading stats in "/sys/devices/system/node/*/hugepages"
+If enabled, information will be collected from the hugepage
+counters in "/sys/devices/system/node/*/hugepages".
+This is used to check the per-node hugepage statistics on
+a NUMA system.
 
-=item B<EnableMM> I<true>|I<false>
+=item B<ReportRootHP> I<true>|I<false>
 
-Enable/Disable reading stats in "/sys/kernel/mm/hugepages"
+If enabled, information will be collected from the hugepage
+counters in "/sys/kernel/mm/hugepages".
+This can be used on both NUMA and non-NUMA systems to check
+the overall hugepage statistics.
 
 =back
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -2699,6 +2699,26 @@ TCP-Port to connect to. Defaults to B<7634>.
 
 =back
 
+=head2 Plugin C<hugepages>
+
+To get values from B<hugepages> collectd read directories
+"/sys/devices/system/node/*/hugepages" and
+"/sys/kernel/mm/hugepages"
+Reading of these directories can be disabled by following
+options. Default is enabled.
+
+=over 4
+
+=item B<EnableNuma> I<true>|I<false>
+
+Enable/Disable reading stats in "/sys/devices/system/node/*/hugepages"
+
+=item B<EnableMM> I<true>|I<false>
+
+Enable/Disable reading stats in "/sys/kernel/mm/hugepages"
+
+=back
+
 =head2 Plugin C<interface>
 
 =over 4
--- a/src/hugepages.c
+++ b/src/hugepages.c
@@ -24,65 +24,59 @@
  *
  * Authors:
  *   Jaroslav Safka <jaroslavx.safka@intel.com>
+ *   Kim-Marie Jones <kim-marie.jones@intel.com>
  */
 
 #include "collectd.h"
 #include "common.h" /* auxiliary functions */
 #include "plugin.h" /* plugin_register_*, plugin_dispatch_values */
 
-#include <stdio.h>
-#include <string.h>
-#include <dirent.h>
+static const char g_plugin_name[] = "hugepages";
+static const char g_cfg_rpt_numa[] = "ReportPerNodeHP";
+static const char g_cfg_rpt_mm[] = "ReportRootHP";
 
-static int huge_read (void);
-static int huge_config_callback (const char *key, const char *val);
-
-static const char PLUGIN_NAME[] = "hugepages";
-static const char SYS_NODE[] = "/sys/devices/system/node";
-static const char NODE[] = "node";
-static const char HUGEPAGES_DIR[] = "hugepages";
-static const char SYS_NODE_HUGEPAGES[] = "/sys/devices/system/node/%s/hugepages";
-static const char SYS_MM_HUGEPAGES[] = "/sys/kernel/mm/hugepages";
-static const char CONFIG_NAME[] = "hugepages";
-static const char CFG_ENA_NUMA[] = "EnableNuma";
-static const char CFG_ENA_MM[] = "EnableMM";
-
-static const char *CONFIG_KEYS[] = {
-  CFG_ENA_NUMA,
-  CFG_ENA_MM,
+static const char *g_config_keys[] = {
+  g_cfg_rpt_numa,
+  g_cfg_rpt_mm,
 };
-static const size_t CONFIG_KEYS_NUM = sizeof(CONFIG_KEYS)/sizeof(*CONFIG_NAME);
-static int g_config_ena_numa = 1;
-static int g_config_ena_mm = 1;
+static size_t g_config_keys_num = STATIC_ARRAY_SIZE(g_config_keys);
+static int g_flag_rpt_numa = 1;
+static int g_flag_rpt_mm = 1;
 
-static int huge_config_callback (const char *key, const char *val)
+struct entry_info {
+  char *d_name;
+  char *node;
+};
+
+static int huge_config_callback(const char *key, const char *val)
 {
-  INFO("HugePages config key='%s', val='%s'", key, val);
+  INFO("%s: HugePages config key='%s', val='%s'", g_plugin_name, key, val);
 
-  if (0 == strcasecmp(key, CFG_ENA_NUMA)) {
-    g_config_ena_numa = IS_TRUE(val);
+  if (strcasecmp(key, g_cfg_rpt_numa) == 0) {
+    g_flag_rpt_numa = IS_TRUE(val);
     return 0;
   }
-  if (0 == strcasecmp(key, CFG_ENA_MM)) {
-    g_config_ena_mm = IS_TRUE(val);
+  if (strcasecmp(key, g_cfg_rpt_mm) == 0) {
+    g_flag_rpt_mm = IS_TRUE(val);
     return 0;
   }
 
   return -1;
 }
 
-static void submit_one (const char *plug_inst, const char *type,
-  const char *type_instance, derive_t value)
+static void submit_hp(const char *plug_inst, const char *type,
+  const char *type_instance, gauge_t free_value, gauge_t used_value)
 {
-  value_t values[1];
+  value_t values[2];
   value_list_t vl = VALUE_LIST_INIT;
 
-  values[0].derive = value;
+  values[0].gauge = free_value;
+  values[1].gauge = used_value;
 
   vl.values = values;
-  vl.values_len = 1;
+  vl.values_len = 2;
   sstrncpy (vl.host, hostname_g, sizeof (vl.host));
-  sstrncpy (vl.plugin, PLUGIN_NAME, sizeof (vl.plugin));
+  sstrncpy (vl.plugin, g_plugin_name, sizeof (vl.plugin));
   sstrncpy (vl.plugin_instance, plug_inst, sizeof (vl.plugin_instance));
   sstrncpy (vl.type, type, sizeof (vl.type));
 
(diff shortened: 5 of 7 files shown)


octo:
Hi @maryamtahhan, thank you very much for your patch! I'll leave some review comments inline.

Best regards,
—octo

configure.ac
```diff
@@ -5971,6 +5974,7 @@ AC_PLUGIN([fscache],             [$plugin_fscache],         [fscache statistics]
 AC_PLUGIN([gmond],               [$with_libganglia],        [Ganglia plugin])
 AC_PLUGIN([grpc],                [$with_grpc],              [gRPC plugin])
 AC_PLUGIN([hddtemp],             [yes],                     [Query hddtempd])
+AC_PLUGIN([hugepages],           [yes],                     [Hugepages statistics])
```
octo: Since this is a Linux-only plugin, it should only be enabled when being compiled on Linux. Search `configure.ac` for `plugin_cgroups` which is a good example of how to do that.

src/collectd.conf.pod
```diff
@@ -2699,6 +2699,26 @@ TCP-Port to connect to. Defaults to B<7634>.
...
+Reading of these directories can be disabled by following
+options. Default is enabled.
+
+=over 4
+
+=item B<EnableNuma> I<true>|I<false>
+
+Enable/Disable reading stats in "/sys/devices/system/node/*/hugepages"
+
+=item B<EnableMM> I<true>|I<false>
```
octo: `EnableMM` is quite cryptic. What does this option do? What kind of metrics will I get when enabling it?

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+ *
+ * Authors:
+ *   Jaroslav Safka <jaroslavx.safka@intel.com>
+ */
+
+#include "collectd.h"
+#include "common.h" /* auxiliary functions */
+#include "plugin.h" /* plugin_register_*, plugin_dispatch_values */
+
+#include <stdio.h>
```
octo: All three, `stdio.h`, `string.h` and `dirent.h` are already included from `src/daemon/collectd.h`.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+#include <string.h>
+#include <dirent.h>
+
+static int huge_read (void);
```
octo: Please remove these prototypes. I don't think they're needed; if I'm wrong, please reorder the functions accordingly.

src/collectd.conf.pod
```diff
@@ -2699,6 +2699,26 @@ TCP-Port to connect to. Defaults to B<7634>.
...
+
+To get values from B<hugepages> collectd read directories
+"/sys/devices/system/node/*/hugepages" and
+"/sys/kernel/mm/hugepages"
+Reading of these directories can be disabled by following
+options. Default is enabled.
+
+=over 4
+
+=item B<EnableNuma> I<true>|I<false>
```
octo: These kind of options are usually called "Report…".

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+  DEBUG("submit_one pl_inst:%s, inst_type %s, type %s, val=%"PRIu64"",
+      plug_inst, type_instance, type, value);
+
+  plugin_dispatch_values (&vl);
+}
+
+static int read_hugepage_entry(const char *path, const char* entry,
+    const char* plinst, const char* hpsize)
+{
+  char path2[512];
```
octo: Please consider using `PATH_MAX` instead of hard-coding 512.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+  name_max = pathconf(path, _PC_NAME_MAX);
+  if (name_max == -1) {    /* Limit not defined, or error */
+    name_max = 255;     /* Take a guess */
+  }
+
+  len = offsetof(struct dirent, d_name) + name_max + 1;
+  entry = malloc(len);
+  if (entry == NULL) {
+    ERROR("Malloc returned NULL");
+    return -1;
```
octo: This leaks `dir`.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  submit_one (plinst, entry, hpsize, value);
+
+  fclose(fh);
+  return 0;
+}
+
+static int read_syshugepage_dir(const char* path, const char* dirhpsize,
```
octo: Please consider using `walk_directory()` from `src/daemon/common.h` instead of reimplementing the logic.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+  if (g_config_ena_numa) {
+    read_nodes();
+  }
+
+  return 0;
+}
+
+void module_register (void)
+{
+	plugin_register_config(CONFIG_NAME, huge_config_callback, CONFIG_KEYS,
```
octo: Nit: indentation is off here.

src/types.db
```diff
@@ -82,6 +82,7 @@ file_size               value:GAUGE:0:U
 files                   value:GAUGE:0:U
 flow                    value:GAUGE:0:U
 fork_rate               value:DERIVE:0:U
+free_hugepages          value:DERIVE:0:U
```
octo: If I understand [hugetlbpage.txt](https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt) correctly, then the correct [data source type](https://collectd.org/wiki/index.php/Data_source) for these metrics would actually be `GAUGE`, i.e. the reported numbers are *not* monotonically increasing. The metrics reported by this plugin should use one "type" and use "type instance" to differentiate between the metrics. In that case, I would recommend to use the existing `vmpage_number` to report these metrics.

*   vmpage_number-free
*   vmpage_number-reserved
*   vmpage_number-surplus
*   …

Since the size of hugepages appears to be a setting, please consider to provide the user with the option to convert the metrics to bytes. In that case, the `bytes` "type" should be used.

src/types.db
```diff
@@ -148,6 +149,9 @@ node_octets             rx:DERIVE:0:U, tx:DERIVE:0:U
 node_rssi               value:GAUGE:0:255
 node_stat               value:DERIVE:0:U
 node_tx_rate            value:GAUGE:0:127
+nr_overcommit_hugepages value:DERIVE:0:U
+nr_hugepages            value:DERIVE:0:U
```
octo: Our policy is to report metrics like this in a way that they can be summed up. This is one of these cases where "total" and "free" / "reserved" are reported, which don't yield a useful sum. Could you please do something along the lines of `used = total - (free + reserved)` and then report the number of used hugepages?

maryamtahhan:
@octo Thank you for the review. I will address your comments and post an update

src/types.db
maryamtahhan: Hi @octo are you asking to report the used pages alongside the total, or only the used pages? the plugin is reporting the free pages as well as the total already and the used value could be deduced. the use case we're looking it is leveraging this information as is for intelligent workload placement so it would be nice to know what's available on the system not only what is used... what do you think?

octo: Hi @maryamtahhan, I'm suggesting to report the used pages *only*. This allows to create a "stacked" graph which sums up nicely to a (usually constant) total. This is what the *memory* and *df plugins* are doing, for example.

When reporting the "total" too, that doesn't work because some bytes are accounted for twice or more times, e.g. free bytes would be included in "free" as well as "total".

Best regards,
—octo

README
```diff
@@ -128,6 +128,12 @@ Features
     - hddtemp
       Hard disk temperatures using hddtempd.
 
+    - hugepages
+      Report the total number of hugpages, the number of free/reserved
```
rubenk: hugepages

configure.ac
```diff
@@ -482,6 +482,9 @@ fi
 # For hddtemp module
 AC_CHECK_HEADERS(linux/major.h)
 
+# For hugepages module
+AC_CHECK_HEADERS(dirent.h)
```
rubenk: You're not doing anything with the result of this check.
Best to disable the hugepages plugin if this header is not found.

src/Makefile.am
```diff
@@ -442,6 +442,13 @@ hddtemp_la_LIBADD += -lsocket
 endif
 endif
 
+if BUILD_PLUGIN_HUGEPAGES
+pkglib_LTLIBRARIES += hugepages.la
+hugepages_la_SOURCES = hugepages.c
+hugepages_la_LDFLAGS = $(PLUGIN_LDFLAGS)
+hugepages_la_LIBADD =
```
rubenk: No need to set this.

README
octo: [#hugops](https://twitter.com/hashtag/hugops) \o/

configure.ac
rubenk: Scratch that, we already include dirent.h in collectd.h and ./configure already looks for it, so you can safely assume it exists.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+static const char SYS_MM_HUGEPAGES[] = "/sys/kernel/mm/hugepages";
+static const char CONFIG_NAME[] = "hugepages";
+static const char CFG_ENA_NUMA[] = "EnableNuma";
+static const char CFG_ENA_MM[] = "EnableMM";
+
+static const char *CONFIG_KEYS[] = {
+  CFG_ENA_NUMA,
+  CFG_ENA_MM,
+};
+static const size_t CONFIG_KEYS_NUM = sizeof(CONFIG_KEYS)/sizeof(*CONFIG_NAME);
```
rubenk: This doesn't look right. sizeof(*CONFIG_NAME) equals sizeof(char) equals 1, so this will end up as 16 or 8 depending on the size of a char pointer.

It should probably be `sizeof(CONFIG_KEYS)/sizeof(*CONFIG_KEYS);`, for which we have a macro, STATIC_ARRAY_SIZE.

Minor nit, please drop the `const`.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+static int g_config_ena_numa = 1;
+static int g_config_ena_mm = 1;
+
+static int huge_config_callback (const char *key, const char *val)
+{
+  INFO("HugePages config key='%s', val='%s'", key, val);
+
+  if (0 == strcasecmp(key, CFG_ENA_NUMA)) {
```
rubenk: Minor nit: no Yoda-conditionals please.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+
+static int read_hugepage_entry(const char *path, const char* entry,
+    const char* plinst, const char* hpsize)
+{
+  char path2[512];
+  long value = 0;
+  snprintf(path2, sizeof(path2), "%s/%s", path, entry);
+
+  FILE *fh = fopen(path2, "rt");
+  if (NULL == fh) {
```
rubenk: fh == NULL please.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+      plug_inst, type_instance, type, value);
+
+  plugin_dispatch_values (&vl);
+}
+
+static int read_hugepage_entry(const char *path, const char* entry,
+    const char* plinst, const char* hpsize)
+{
+  char path2[512];
+  long value = 0;
```
rubenk: No need to initialize value.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+    const char* node)
+{
+  DIR       *dir = NULL;
+  struct dirent *entry = NULL;
+  struct dirent *result = NULL;
+  size_t name_max = 0;
+  size_t len = 0;
+
+  dir = opendir(path);
+  if (NULL == dir) {
```
rubenk: dir == NULL please.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+    ERROR("Cannot open directory %s", path);
+    return -1;
+  }
+
+  name_max = pathconf(path, _PC_NAME_MAX);
+  if (name_max == -1) {    /* Limit not defined, or error */
+    name_max = 255;     /* Take a guess */
```
rubenk: If this fails there's something seriously wrong. I think it's best to return the error.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  len = offsetof(struct dirent, d_name) + name_max + 1;
+  entry = malloc(len);
+  if (entry == NULL) {
+    ERROR("Malloc returned NULL");
```
rubenk: Can you prepend the plugin name?

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+    return -1;
+  }
+
+  while (0 == readdir_r(dir, entry, &result)) {
```
rubenk: readdir_r has become deprecated in glibc 2.24. Since we report warnings as errors by default, this will fail on newer platforms. See http://man7.org/linux/man-pages/man3/readdir_r.3.html for the details.

src/hugepages.c
```diff
@@ -0,0 +1,286 @@
...
+      break;
+    }
+
+    if (strncmp(result->d_name, HUGEPAGES_DIR, sizeof(HUGEPAGES_DIR)-1)) {
+      /* not node dir */
+      continue;
+    }
+
+    /* /sys/devices/system/node/node?/hugepages/ */
+    snprintf(path2, sizeof(path2), "%s/%s", path, result->d_name);
```
rubenk: Using ssnprintf will guarantee path2 is 0-terminated.

configure.ac
kimmariejones: Fixed

src/collectd.conf.pod
kimmariejones: Fixed

src/hugepages.c
kimmariejones: removed these

src/hugepages.c
kimmariejones: removed these

src/collectd.conf.pod
kimmariejones: New names start with Report...

src/hugepages.c
kimmariejones: Done

src/hugepages.c
kimmariejones: Fixed

src/hugepages.c
kimmariejones: Done.

src/hugepages.c
kimmariejones: Fixed

src/types.db
kimmariejones: Made changes to types: see next comment.

src/types.db
kimmariejones: So, we have done the following, I hope it is to your satisfaction:

- Added 1 new type, "hugepages"
- Gauge value, same limits as vmpage_number
- However, we have added 2 expected values: "free" and "used"
- We calculate "used" from the other metrics but we do not report the other metrics

I hope this is in line with what you requested in this and other comments.

README
kimmariejones: Fixed :)

configure.ac
kimmariejones: Deleted

src/Makefile.am
kimmariejones: Removed that line

src/hugepages.c
kimmariejones: Done

src/hugepages.c
kimmariejones: Fixed these

src/hugepages.c
kimmariejones: Fixed

src/hugepages.c
kimmariejones: Fixed

src/hugepages.c
kimmariejones: Fixed

src/hugepages.c
kimmariejones: Done

src/hugepages.c
kimmariejones: Done

src/hugepages.c
kimmariejones: Done

src/hugepages.c
kimmariejones: Done

src/collectd.conf.pod
```diff
@@ -2699,6 +2699,32 @@ TCP-Port to connect to. Defaults to B<7634>.
 
 =back
 
+=head2 Plugin C<hugepages>
+
+To collect B<hugepages> information, collectd reads directories
+"/sys/devices/system/node/*/hugepages" and
+"/sys/kernel/mm/hugepages".
+Reading of these directories can be disabled by following
```
rubenk: "the" following options?

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+static int g_flag_rpt_mm = 1;
+
+struct entry_info {
+  char *d_name;
+  char *node;
+};
+
+static int huge_config_callback(const char *key, const char *val)
+{
+  INFO("%s: HugePages config key='%s', val='%s'", g_plugin_name, key, val);
```
rubenk: should this be logged at severity INFO? It looks more like a debug statement to me.

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+
+  plugin_dispatch_values (&vl);
+}
+
+static int read_hugepage_entry(const char *path, const char *entry,
+    void *e_info)
+{
+  char path2[PATH_MAX];
+  char *type = "hugepages";
+  char *partial_type_inst = "free_used";
```
rubenk: can you make the above two variables static const arrays like you do elsewhere? They don't seem to be modified.

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  DEBUG("submit_hp pl_inst:%s, inst_type %s, type %s, free=%lf, used=%lf",
+      plug_inst, type_instance, type, free_value, used_value);
+
+  plugin_dispatch_values (&vl);
+}
+
+static int read_hugepage_entry(const char *path, const char *entry,
+    void *e_info)
```
rubenk: why the void * and not struct entry_info *?

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+{
+  char path2[PATH_MAX];
+  char *type = "hugepages";
+  char *partial_type_inst = "free_used";
+  char type_instance[PATH_MAX];
+  char *strin = NULL;
+  struct entry_info *hpsize_plinst = (struct entry_info *) e_info;
```
rubenk: unneccesary cast

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+}
+
+static int read_hugepage_entry(const char *path, const char *entry,
+    void *e_info)
+{
+  char path2[PATH_MAX];
+  char *type = "hugepages";
+  char *partial_type_inst = "free_used";
+  char type_instance[PATH_MAX];
+  char *strin = NULL;
```
rubenk: no need to initialize this to NULL.

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+    free_hp = 0;
+  }
+
+  fclose(fh);
+  return 0;
+}
+
+static int read_syshugepages(const char* path, const char* node)
+{
+  const char hugepages_dir[] = "hugepages";
```
rubenk: make this static?

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+  DIR *dir = NULL;
```
rubenk: no need to initialize this.

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+  struct dirent *result = NULL;
```
rubenk: same here

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+      /* not node dir */
+      continue;
+    }
+
+    /* /sys/devices/system/node/node?/hugepages/ */
+    ssnprintf(path2, sizeof(path2), "%s/%s", path, result->d_name);
+
+    e_info.d_name = result->d_name;
+    e_info.node = (char *) node;
+    walk_directory(path2, read_hugepage_entry, (void *) &e_info, 0);
```
rubenk: unneccesary cast

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+    if (strncmp(result->d_name, hugepages_dir, sizeof(hugepages_dir)-1)) {
+      /* not node dir */
+      continue;
+    }
+
+    /* /sys/devices/system/node/node?/hugepages/ */
+    ssnprintf(path2, sizeof(path2), "%s/%s", path, result->d_name);
+
+    e_info.d_name = result->d_name;
+    e_info.node = (char *) node;
```
rubenk: perhaps make e_info.node const so you can drop the cast?

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+  closedir(dir);
+
+  return 0;
+}
+
+static int read_nodes(void)
+{
+  const char sys_node[] = "/sys/devices/system/node";
+  const char node_string[] = "node";
+  const char sys_node_hugepages[] = "/sys/devices/system/node/%s/hugepages";
```
rubenk: make these 3 static?

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+  DIR *dir = NULL;
+  struct dirent *result = NULL;
```
rubenk: no need to initialize these 2 to NULL.

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+  return 0;
+}
+
+
+static int huge_read(void)
+{
+  const char sys_mm_hugepages[] = "/sys/kernel/mm/hugepages";
+
+  if (g_flag_rpt_mm) {
+    read_syshugepages(sys_mm_hugepages, "mm");
```
rubenk: What if this returns -1?

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+  }
+  if (g_flag_rpt_numa) {
+    read_nodes();
```
rubenk: Same here.

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+
+  closedir(dir);
+
+  return 0;
+}
+
+
+static int huge_read(void)
+{
+  const char sys_mm_hugepages[] = "/sys/kernel/mm/hugepages";
```
rubenk: static?

src/hugepages.c
```diff
@@ -0,0 +1,258 @@
...
+  char path2[PATH_MAX];
+  struct entry_info e_info;
+
+  dir = opendir(path);
+  if (dir == NULL) {
+    ERROR("%s: cannot open directory %s", g_plugin_name, path);
+    return -1;
+  }
+
+  if (pathconf(path, _PC_NAME_MAX) == -1) {
```
rubenk: Can you explain this code? You don't seem to limit anything to _PC_NAME_MAX.
Also, according to the manpage, -1 doesn't necessarily mean there's an error, it might also be that there is no limit.

src/hugepages.c
kimmariejones: Hi Ruben,

I'm using this function as a parameter to walk_directory(). The void pointer parameter is expected and is also done in the filecount.c file (see below). Do you have a preferred method?

459 static int fc_read_dir_callback (const char *dirname, const char *filename,
    460     void *user_data)
    461 {
    462   fc_directory_conf_t *dir = user_data;
    463   char abs_path[PATH_MAX];
    464   struct stat statbuf;
    465   int status;
    466
    467   if (dir == NULL)
    468     return (-1);
    469
    470   ssnprintf (abs_path, sizeof (abs_path), "%s/%s", dirname, filename);
    471
    472   status = lstat (abs_path, &statbuf);
    473   if (status != 0)
    474   {
    475     ERROR ("filecount plugin: stat (%s) failed.", abs_path);
    476     return (-1);
    477   }
    478
    479   if (S_ISDIR (statbuf.st_mode) && (dir->options & FC_RECURSIVE))
    480   {
    481     status = walk_directory (abs_path, fc_read_dir_callback, dir,
    482         /* include hidden = */ (dir->options & FC_HIDDEN) ? 1 : 0);
    483     return (status);
    484   }

rubenk: Sorry, I totally missed that walk_directory was a pre-existing function.

src/collectd.conf.pod
kimmariejones: fixed

src/hugepages.c
kimmariejones: done

src/hugepages.c
kimmariejones: Done

src/hugepages.c
kimmariejones: changed to truct entry_info *hpsize_plinst =  e_info;

src/hugepages.c
kimmariejones: fixed

src/hugepages.c
kimmariejones: done

src/hugepages.c
kimmariejones: done

src/hugepages.c
kimmariejones: done

src/hugepages.c
kimmariejones: cast removed

src/hugepages.c
kimmariejones: yup, done

src/hugepages.c
kimmariejones: done

src/hugepages.c
kimmariejones: fixed

src/hugepages.c
kimmariejones: Added return value check

src/hugepages.c
kimmariejones: Added return value check

src/hugepages.c
kimmariejones: done

src/hugepages.c
kimmariejones: Sorry, made a bit of a mess of that when reworking it. The value returned by pathconf is used now and error checking is done correctly.

src/hugepages.c
```diff
@@ -0,0 +1,274 @@
...
+    if (strncmp(result->d_name, node_string, sizeof(node_string)-1)) {
+      /* not node dir */
+      continue;
+    }
+
+    ssnprintf(path, (size_t) lim, sys_node_hugepages, result->d_name);
+    read_syshugepages(path, result->d_name);
+  }
+
+  closedir(dir);
```
rubenk: What if readdir returns an error? In that case closedir is called with a NULL argument.

kimmariejones: I agree that I need to check errno here and return -1 if it is set; will add that. 

However, dir is opened at line 216. closedir(dir) will still be required even if readdir fails with an error.

rubenk: Agreed. I guess in practice calling closedir with an invalid DIR * is not an issue. POSIX is silent about this, but their example uses the same code as you do: http://pubs.opengroup.org/onlinepubs/009695399/functions/closedir.html

rubenk:
Thanks @kimmariejones! I have nothing more to complain about, so I think we're good to go.
@octo Since we've already branched 5.6 at this point, do you want to keep it into master for one release cycle, or can this still go into 5.6?

rubenk:
Thanks for your work @maryamtahhan and @kimmariejones. I've just pulled this into master and it will be part of the next collectd release due around christmas.

maryamtahhan:
Thanks @rubenk

PR #1527, opened 2016-01-25
Track process syscalls for AIX, solaris, and linux. Add Fork rate to AIX

jquinn47:
This change set enhances the processes plugin on AIX, solaris, and linux

* A 'syscalls' metric is added to all three platforms. 
* AIX now tracks the fork_rate, similar to other platforms.

PR #1800, opened 2016-07-18
Adding metric for number of syscalls

jquinn47:
Re-opening #1525

This change lets us track the number of syscalls being invoked on AIX, solaris, and linux, using perfstat, kstat, and proc/stat respectively.

In addition, I've added 'fork_rate' to AIX. That metric is already being tracked on linux and solaris.

As per the comment in #1527, since this PR adds new functionality, I've submitted the PR to merge onto the master branch.


src/Makefile.am
```diff
@@ -849,6 +849,9 @@ processes_la_LIBADD =
 if BUILD_WITH_LIBKVM_GETPROCS
 processes_la_LIBADD += -lkvm
 endif
+if BUILD_WITH_PERFSTAT
+contextswitch_la_LIBADD += -lperfstat
+endif
```
rubenk: This seems wrong. It is already in Makefile.am

src/types.db
```diff
@@ -74,6 +74,7 @@ file_size		value:GAUGE:0:U
 files			value:GAUGE:0:U
 flow			value:GAUGE:0:U
 fork_rate		value:DERIVE:0:U
+syscalls        value:DERIVE:0:U
```
rubenk: please keep this list sorted.

src/Makefile.am
rubenk: It's also the reason for the failing tests. Did you test this locally?

shalstea:
Do you know when this will be merged (roughly) ?

src/processes.c
```diff
@@ -1295,25 +1320,35 @@ static int read_fork_rate (void)
...
 	}
 	fclose(proc_stat);
 
-	if (!value_valid)
+	if (!fr_value_valid || !intr_value_valid)
 		return (-1);
 
-	ps_submit_fork_rate (value.derive);
+	ps_submit_fork_rate (fr_value.derive);
+	ps_submit_syscalls (fr_value.derive + intr_value.derive);
```
octo: Can you please add a comment why these two seemingly unrelated metrics need to be added together here?

src/processes.c
```diff
@@ -807,6 +811,25 @@ static void ps_submit_fork_rate (derive_t value)
 
 	plugin_dispatch_values(&vl);
 }
+
+static void ps_submit_syscalls (derive_t value)
```
octo: It would be simpler if you would add a `char const *type` argument to the existing `ps_submit_fork_date()` function (and find an appropriate name, maybe `ps_submit_type()`).

octo:
LGTM, but please note the comments inline.

octo:
@shalstea this PR is pending some minor changes by @jquinn47, then it should be good to go.

jquinn47:
I'm cleaning this up a bit and will resubmit once I test the changes on my end.

PR #1801, opened 2016-07-18
Adding AIX used memory

jquinn47:
This is re-opened from #1524

The change lets us track the memory in use on AIX boxes, via perfstat. I've added it to the list of submitted metrics, keeping naming consistent with the other platforms.

In the last PR, @mfournier asked if this would break or fix the fact that all reported metrics sum up to 100% of available memory.

To answer that, I believe this would break the all metrics sum to 100% behavior, although using the `free + used` metrics is convenient. The previous metrics sum to the total amount of memory in use by an AIX platform, however graphing tools we're using that just want to know the memory `in use` would prefer a simpler metric, hence the addition of used.

The total could be summed as follows, assuming we're using PR's metric naming conventions.

`TOTAL = free + used = free + cached + system + user`


rubenk:
Please rebase on current master.

jquinn47:
No problem

rubenk:
Did you rebase? I still see 4 commits.

jquinn47:
ugh. git fumble. will fix

PR #1808, opened 2016-07-26
metadata: add comment about metadata functions use and threads safety

beorn-:
After discussing with octo we agreed that this comment should be in the code


octo:
Thanks @beorn-!

Cherry-picked into *collectd-5.4* as 92915087fd253244d323ff88cd97c54367689030

Issue #1665, opened 2016-04-09
Security Bug: Improper usage of gcry_control


yujokang:
Dear Collectd developers,

We are security researchers at Columbia University and the University of Virginia. As part of a research project, we have built a tool for automatically finding error handling bugs and are testing it on various cryptographic libraries and applications that use them.

We discovered that GCrypt's gcry_control is sometimes called without checking its return value for an error, which may cause the program to be initialized without the desired, secure settings.

Please let us know how you intend to address these issues.

When gcry_control is called with GCRYCTL_SET_THREAD_CBS or
GCRYCTL_INIT_SECMEM, it can return a non-zero error value due to reasons beyond the control of the caller, so it needs to be checked; otherwise the program may not be properly initialized. In particular, we found such uses in:

src/network.c, line 501:
```C
static void network_init_gcrypt (void)
{
...
 gcry_control (GCRYCTL_SET_THREAD_CBS, &gcry_threads_pthread);
...
 gcry_control (GCRYCTL_INIT_SECMEM, 32768);
...
}
```

and

src/libcollectdclient/network_buffer.c, line 126:
```C
static _Bool have_gcrypt (void) /* {{{ */
{
...

 gcry_control (GCRYCTL_SET_THREAD_CBS, &gcry_threads_pthread);
...
 gcry_control (GCRYCTL_INIT_SECMEM, 32768, 0);
...
}
```

Thank you,
Yuan Kang

octo:
FYI, I've sent a patch to the gcrypt development list to improve their example code in the documentation: http://lists.gnupg.org/pipermail/gcrypt-devel/2016-July/003934.html

wtps0n:
After upgrade to 5.5.2 version which resolves above issues network plugin dies with core dump:
Aug  9 08:59:25 fw collectd[18731]: network plugin: gcry_control (GCRYCTL_SET_THREAD_CBS) failed: General error
Aug  9 08:59:25 fw collectdmon[18718]: Warning: collectd was terminated by signal 6 (core dumped)
Aug  9 08:59:25 fw kernel: pid 18731 (collectd), uid 0: exited on signal 6 (core dumped)

I'm using it on FreeBSD (10.3). I've found via Google that on other systems (Debian) users reports similiar problems - http://osdir.com/ml/general/2016-07/msg40034.html 

Any ideas? Wrong configuration? libgcrypt should be upgraded?

rubenk:
[...]

PR #1810, opened 2016-07-27
network plugin: Don't abort() if gcrypt initialization failed.

tokkee:
Instead, report an error and let plugin initialization fail.


octo:
LGTM, will merge this is a minute

tokkee:
Ftr, this is a follow-up to the fix for #1665

PR #1811, opened 2016-07-27
network plugin: Fix heap overflow in parse_packet().

rubenk:
Emilien Gaspar has identified a heap overflow in parse_packet(), the
function used by the network plugin to parse incoming network packets.

This is a vulnerability in collectd, though the scope is not clear at
this point. At the very least specially crafted network packets can be
used to crash the daemon. We can't rule out a potential remote code
execution though.

Fixes: CVE-2016-6254
(cherry picked from commit b589096f907052b3a4da2b9ccc9b0e2e888dfc18)


octo:
LGTM

Issue #1790, opened 2016-07-12
Varnish plugin ignoring configuration options under libvarnishapi1 4.1.3


dbeckham:
When collectd is compiled against libvarnishapi1 version 4.1.3, available from varnish-cache.org's Debian/Ubuntu repository: https://repo.varnish-cache.org/ubuntu/, it refuses to recognize the following configuration options:

```
CollectDirectorDNS
CollectPurge
CollectSMA
CollectSM
```
All of which are described here: https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_varnish

There seems to be a discrepancy between the man pages and the wiki (https://collectd.org/wiki/index.php/Plugin:Varnish), so I removed all but `CollectSMA` option, but still saw the same start up warnings about `CollectSMA`

Example syslog output:
```
collectd[437]: Varnish plugin: Ignoring unknown configuration option: "CollectDirectorDNS". Did you forget to add an <Instance /> block around the configuration?
collectd[437]: Varnish plugin: Ignoring unknown configuration option: "CollectPurge". Did you forget to add an <Instance /> block around the configuration?
collectd[437]: Varnish plugin: Ignoring unknown configuration option: "CollectSMA". Did you forget to add an <Instance /> block around the configuration?
collectd[437]: Varnish plugin: Ignoring unknown configuration option: "CollectSM". Did you forget to add an <Instance /> block around the configuration?
```

My varnish plugin configuration looks like:

```
LoadPlugin varnish              
<Plugin varnish>                
    <Instance>
        CollectBackend     true 
        CollectBan         true 
        CollectCache       true 
        CollectConnections true 
        CollectDirectorDNS false
        CollectESI         false
        CollectFetch       true 
        CollectHCB         false
        CollectObjects     true 
        CollectPurge       true 
        CollectSession     true 
        CollectSHM         true 
        CollectSMA         true 
        CollectSMS         true 
        CollectSM          true 
        CollectStruct      true 
        CollectTotals      true 
        CollectUptime      true 
        CollectVCL         true 
        CollectVSM         true 
        CollectWorkers     true 
    </Instance>                 
</Plugin>                       
```

Start up output looks like:
```
option = FQDNLookup; value = true;
option = Interval; value = 60.000000;
option = Timeout; value = 2.000000;
Created new plugin context.
Varnish plugin: Ignoring unknown configuration option: "CollectDirectorDNS". Did you forget to add an <Instance /> block around the configuration?
[...]

PR #1812, opened 2016-07-28
varnish plugin: Add informative warnings to the config handling.

octo:
Not all config options are valid for all versions of Varnish. Rather
than removing the if-clause altogether, which results in a "Ignoring
unknown configuration option" warning, print a user-friendly warning
instead.

Fixes: #1790


rubenk:
LGTM

PR #1816, opened 2016-07-28
grpc plugin: Free the cache iterator when returning due to an error.

octo:


tokkee:
LGTM. Thanks for catching this!

PR #1821, opened 2016-07-30
treewide: replace memset to 0 with initializers

rubenk:


rubenk:
The build failures is us building with -Werror and a bug in GCC:

ascent.c:368:3: error: missing braces around initializer [-Werror=missing-braces]
   player_stats_t ps = { 0 };
   ^

Probably this issue: 
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119

rubenk:
I tried to disable the warning, but you need a recent compiler for that.
Let's retry this in 10 years, hopefully by then compilers have caught op with C99 ;)

octo:
Meh, that sucks; I really like that change. I assume that we're getting the missing braces error because the first element in the struct is a `char[]`, so GCC thinks that a zero literal is no good. :\

Maybe we can work around that by doing any one of those:

```c
player_stats_t ps = { // my favorite
    .level_sum = 0
};
player_stats_t ps = { "" }; // assumes member order
player_stats_t ps = { { 0 } }; // confusing
```

Another C99 feature I really like (but never used due to lack of support from compilers, back in the day) was declaring loop variables inside the loop condition:

```c
for (int i = 0; i < max; i++) {
  // ^^^
}
```

rubenk:
> I assume that we're getting the missing braces error because the first element in the struct is a char[], so GCC thinks that a zero literal is no good.

Oh, right! Let me check that. Clang also gets it wrong, hopefully for the same reason.
If your first example works, I think that's the nicest.

`player_stats_t ps = { "" };` can work, unless we really need to zero the first [] I guess.

While doing this I spotted some places where we do a memset followed by a ssnprintf. In that case we can get rid of the memset altogether.

> Another C99 feature I really like (but never used due to lack of support from compilers, back in the day) was declaring loop variables inside the loop condition

Yes please! It reduces scope, shadowing and makes the code much more clear IMHO.
Declaring variables just before use in general is nice.

rubenk:
Meh, GCC fails for nested structs too, see https://travis-ci.org/rubenk/collectd/jobs/148833108#L2953

octo:
I managed to reproduce this issue with GCC:

```
network.c: In function 'lcc_server_set_interface':
network.c:415:14: error: missing braces around initializer [-Werror=missing-braces]
       struct ip_mreqn mreq = { 0 };
```

This works in GCC, but if I read the C11 spec correctly, is not valid C:

```c
struct ip_mreqn mreq = (struct ip_mreqn) {};
```

This also works (and is valid C), but needs to be adapted for each location:

```c
struct ip_mreqn mreq = {
  .imr_ifindex = (int) if_index,
};
```

For this particular case, this is the nicest option though:

```c
struct ip_mreqn mreq = {
  .imr_ifindex          = (int) if_index,
  .imr_multiaddr.s_addr = addr->sin_addr.s_addr,
  .imr_address.s_addr   = ntohl (INADDR_ANY),
};
```

rubenk:
`struct ip_mreqn mreq = (struct ip_mreqn) {};` works for C++ but it is indeed invalid C. GCC has supported it as an extension, not sure how how clang handles it.

octo:
Looks good :) If it compiles, ship it ;)

rubenk:
Almost there, still having an issue with ipmi.c on RHEL5/RHEL6.

Apparently the OpenIPMI headers use anonymous unions which afaict are in C11, and before that were a GCC extension.

This doesn't compile
```
#define IPMI_OPEN_OPTION_ALL 1

typedef struct ipmi_open_option_s
{
    int option;
    union {
        long ival;
        void *pval;
    };
} ipmi_open_option_t;

int main(void)
{
    ipmi_open_option_t open_option[] = {
        [0] = {
            .option = IPMI_OPEN_OPTION_ALL,
            .ival = 1
        }
    };
}
```
```
$ gcc -std=c99 test.c 
test.c:11: warning: declaration does not declare anything
test.c: In function ‘main’:
test.c:19: error: unknown field ‘ival’ specified in initializer
test.c:20: warning: excess elements in struct initializer
test.c:20: warning: (near initialization for ‘open_option[0]’)
```

Interesting :)

rubenk:
This works, but only with -std=gnu99:

```
#define IPMI_OPEN_OPTION_ALL            1

typedef struct ipmi_open_option_s
{
    int option;
    union {
        long ival;
        void *pval;
    };
} ipmi_open_option_t;

int main(void)
{
    ipmi_open_option_t open_option[1] = {
        [0] = {
            .option = IPMI_OPEN_OPTION_ALL,
            {.ival = 1}
        }
    };
}
```

PR #1822, opened 2016-07-30
write_riemann plugin: reindent

rubenk:
The code has become a mix of tabs and spaces due to various
merges over the years. I tossed a coint and went with 4 spaces here.

The diff is easier to review if you add &w=1 to the url on github.

### DIFF
--- a/src/write_riemann.c
+++ b/src/write_riemann.c
@@ -30,10 +30,9 @@
 
 #include "collectd.h"
 
-
-#include "plugin.h"
 #include "common.h"
 #include "configfile.h"
+#include "plugin.h"
 #include "utils_cache.h"
 #include "utils_complain.h"
 #include "write_riemann_threshold.h"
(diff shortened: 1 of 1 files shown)


octo:
As discussed in #1791, I'd like to suggest to go with `clang-format -style=LLVM` (the default style). It's not too different from collectd's overall style and I'd like to avoid creating our own style definition.

rubenk:
Good idea, thanks. I'll try it on this file, let's see how it handles it :)

rubenk:
I must say it does a pretty good job! The only thing is that it seems to reorder includes so we need to be careful with that. Then again, depending on include order is often a bug in hiding. WDYT?

octo:
Looks good :)

It appears that clang-format orders `"includes"` before `<includes>`, which suits us. We might want to add a configuration along the lines of:

```yaml
IncludeCategories:
  - Regex:    '"collectd.h"'
  - Priority: -1
```

This will make sure that `"collectd.h"` is always included first.

rubenk:
Ok, I'll do that. While at it, do you mind if I set IndentWidth: 4? I like it so much more than 2 <ducks> ;)

PR #1823, opened 2016-07-30
cpufreq plugin: support for cpufreq/stats and hot-pluggable CPUs 

rinigus:
This PR adds cpufreq support for 

* hot-pluggable CPUs
* following CPU used frequencies distribution, as provided by cpufreq/stats

On mobile platforms, CPUs are frequently switched off to save battery. As a result, at the moment when CPU is not available, the corresponding cpufreq and cpufreq/stats is unavailable. The current cpufreq plugin considers such situation as an error. This PR adds support for such configurations and flags an error only if there are no CPUs providing cpufreq data.

In Linux, an accurate cpufreq statistics is given through cpufreq/stats directory. This allows to account for usage of CPU frequencies between the statistics collection by _collectd_. On mobile platforms, CPU frequency is changed frequently, and this PR adds support for this.

An example graph from the collected data on CPU frequency distribution is shown below in the system with hot-pluggable CPUs. 
   
cpufreq_stats

### DIFF
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -359,7 +359,7 @@ static int cpufreq_read (void)
 				}
 
 				hertz = atol( fields[0] );
-				time = atof( fields[1] );
+				time = atof( fields[1] ) * 10e-3;
 
 				if ( hertz <= 0 )
 				{
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -18,6 +18,23 @@
  *
  * Authors:
  *   Peter Holik <peter at holik.at>
+ *   rinigus <http://github.com/rinigus>
+ *
+ * CPU frequency is recorded either as a current value or as changes
+ * in used CPU frequency distribution. For current value,
+ * "/sys/devices/system/cpu/cpuX/cpufreq/scaling_cur_freq" are
+ * used. For reporting distribution changes, amount of used time on
+ * each supported CPU frequency is reported using DERIVE
+ * type. Distributions are collected from
+ * "/sys/devices/system/cpu/cpuX/cpufreq/stats/time_in_state", format
+ * description at https://www.kernel.org/doc/Documentation/cpu-freq/cpufreq-stats.txt.
+ *
+ * Units of reported values:
+ *
+ *  Frequencies in Hz when reported as a current value
+ *  Frequencies in kHz when used as a type instance for distributions
+ *  Distribution utilization in seconds / second
+ *
  **/
 
 #include "collectd.h"
@@ -26,39 +43,181 @@
 
 #define MODULE_NAME "cpufreq"
 
-static int num_cpu = 0;
+// Configuration options
+static const char *config_keys[] =
+{
+	"ReportDistribution",
+	"ReportByCpu"
+};
+static int config_keys_num = STATIC_ARRAY_SIZE (config_keys);
+
+// Global variables
+
+static _Bool report_distribution = 0; 	// option
+static _Bool report_by_cpu = 0; 		// option
+
+static size_t num_cpu = 0; 	// number of cpus 
+
+static long int *hertz_all_cpus = NULL;
+static double *time_all_cpus = NULL;
+static size_t hertz_size = 0;
+static size_t reported_last_run = 0;
+
+// Functions
+
+static int cpufreq_config (const char *key, const char *value)
+{
+	if (strcasecmp (key, "ReportDistribution") == 0)
+		report_distribution = IS_TRUE(value);
+	else if (strcasecmp (key, "ReportByCpu") == 0)
+		report_by_cpu = IS_TRUE(value);
+	else return (-1);
+
+	return (0);
+}
+
 
 static int cpufreq_init (void)
 {
+	char filename[128];
 	int status;
-	char filename[256];
+	_Bool done;
+	
+	// determine number of cpus. here, just check whether cpus are
+	// there. since cpus can be hot-plugged, cpufreq could be absent
+	// at this particualr time moment
 
-	num_cpu = 0;
-
-	while (1)
+	done = 0;
+	for (num_cpu = 0; !done; ++num_cpu)
 	{
 		status = ssnprintf (filename, sizeof (filename),
-				"/sys/devices/system/cpu/cpu%d/cpufreq/"
-				"scaling_cur_freq", num_cpu);
+							"/sys/devices/system/cpu/cpu%zu",
+							num_cpu);
+		
 		if ((status < 1) || ((unsigned int)status >= sizeof (filename)))
+		{
+			INFO("cpufreq plugin: error in ssnprintf");
 			break;
-
+		}
+		
 		if (access (filename, R_OK))
 			break;
-
-		num_cpu++;
 	}
 
-	INFO ("cpufreq plugin: Found %d CPU%s", num_cpu,
-			(num_cpu == 1) ? "" : "s");
-
+	INFO ("cpufreq plugin: Found %zu CPU%s", num_cpu,
+		  (num_cpu == 1) ? "" : "s");
+	
 	if (num_cpu == 0)
 		plugin_unregister_read ("cpufreq");
+	
+	if ( num_cpu > 0 && report_distribution && !report_by_cpu )
+	{
+		FILE *fp;
+		char *fields[2];
+		char buffer[256];
+		int numfields;
+		long int hertz;
+		int found;
+		size_t i, j;
+			
+		// prefill all hertz
+		hertz_all_cpus = NULL;
+		hertz_size = 0;
+
+		for (i=0; i < num_cpu; ++i)
+		{
+			status = ssnprintf (filename, sizeof (filename),
+								"/sys/devices/system/cpu/cpu%zu/cpufreq/stats/time_in_state",
+								i);
+		
+			if ((status < 1) || ((unsigned int)status >= sizeof (filename)))
+			{
+				INFO("cpufreq plugin: error in ssnprintf");
+				break;
+			}
+			
+			if ((fp = fopen (filename, "r")) == NULL)
+				// cpu could be switched off, just take the next one
+				continue;
+
+			while ( fgets(buffer, sizeof(buffer), fp) != NULL )
+			{
+				numfields = strsplit (buffer, fields, STATIC_ARRAY_SIZE(fields));
+				
+				if ( numfields != 2 ) // supported format contains 2 fields
+				{
+					// if its an empty line, just ignore this line
+					if ( numfields == 0 ) continue;
+					
+					WARNING ("cpufreq plugin: wrong number of items on a line in %s", filename);
+					fclose (fp);
+					plugin_unregister_read ("cpufreq");
+					return (0);
+				}
+					
+				hertz = atol( fields[0] );
+
+				// do we have this value already?
+				found = 0;
+				for ( j=0; j < hertz_size && !found; ++j )
+					if ( hertz == hertz_all_cpus[j] )
+						found = 1;
+
+				if ( !found )
+				{
+					hertz_size += 1;
+					hertz_all_cpus = (long int*)realloc( hertz_all_cpus, hertz_size * sizeof(long int) );
+					if ( hertz_all_cpus == NULL )
+					{
+						INFO("cpufreq plugin: realloc failed for hertz_all_cpus size %zu", hertz_size);
+						fclose (fp);
+						plugin_unregister_read ("cpufreq");
+						return (0);
+					}
+					hertz_all_cpus[ hertz_size-1 ] = hertz;
+				}
+			}
+			
+			fclose (fp);
+		}
+
+		if (hertz_size == 0)
+		{
+			INFO("cpufreq plugin: cannot find any frequencies in the stats");
+			plugin_unregister_read ("cpufreq");
+			return (0);
+		}
+
+		time_all_cpus = (double*)malloc( hertz_size * sizeof(double) );
+		if ( time_all_cpus == NULL )
+		{
+			INFO("cpufreq plugin: malloc failed for time_all_cpus size %zu", hertz_size);
+			plugin_unregister_read ("cpufreq");
+			return (0);
+		}
+			
+		INFO("cpufreq plugin: found %zu different frequencies", hertz_size);
+	}
 
 	return (0);
 } /* int cpufreq_init */
 
-static void cpufreq_submit (int cpu_num, double value)
+static int cpufreq_shutdown (void)
+{
+	if ( report_distribution )
+	{
+		sfree( hertz_all_cpus );
+		hertz_all_cpus = NULL;
+		
+		sfree( time_all_cpus );
+		time_all_cpus = NULL;
+	}
+	
+	return (0);
+} /* int cpufreq_shutdown */
+
+// used when single value is reported
+static void cpufreq_submit_current_value (size_t cpu_num, double value)
 {
 	value_t values[1];
 	value_list_t vl = VALUE_LIST_INIT;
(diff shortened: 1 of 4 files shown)


src/cpufreq.c
```diff
@@ -26,39 +43,181 @@
...
-	num_cpu = 0;
+static int cpufreq_init (void)
+{
+	char filename[128];
+	int status;
+	_Bool done;
+	
+	// determine number of cpus. here, just check whether cpus are
+	// there. since cpus can be hot-plugged, cpufreq could be absent
+	// at this particualr time moment
```
octo: s/particualr/particular/

src/cpufreq.c
```diff
@@ -71,67 +230,222 @@ static void cpufreq_submit (int cpu_num, double value)
...
+					cpufreq_submit_distribution_value( statind, hertz, time );
+				}
+				else
+				{
+					int found = 0;
+
+					if ( hertz_size==1 )
+					{
+						if ( hertz == hertz_all_cpus[0] )
+							time_all_cpus[0] += time;
```
octo: You can't add counts / `derive_t` like this. If you want to aggregate all CPUs, you need to convert each individual count to a *rate* first, using `value_to_rate()` from *common.h*.

octo:
As mentioned inline, it may be preferable to convert all the counter values to a rate, using `value_to_rate()` from `"common.h"`: with a rate you can compute the percentage spent in each frequency, which may be more useful than some arbitrary "ticks per second" that you'll get otherwise. We just spend quite some time changing the *cpu plugin* in this direction. With percentages, you can also calculate average frequency, 90th percentile frequency, etc from these rates, which may be interesting config options.

Apart from this it would be nice if you could spend some time cleaning up the coding style to the code blends into the rest of the file. Spaces inside parenthesis (`if ( foo )`) caught my attention but there is other unusual whitespace (`i=0`, `function_call( args )`, …)

The *init()* function is nested very deep, maybe you could break some functionality out into a utility function to make that easier to read.

Best regards,
—octo

rinigus:
@octo,

I'll be happy to clean up the code - I'm just used to put spaces in few places that you guys don't seem to do. 

but lets get the calculus right first. the values that are calculated are in seconds, not ticks. At least kernel documentation is rather specific about it. So, maybe I should convert it to ms as in cpusleep. I will read about the function that you propose, but so far its not clear why I cannot sum up times together and take derivative later. It seems like rather simple derivative rules that should be OK. But I'll get back after I've read about that function .

thank you for the review,

rinigus

rinigus:
@octo:

Thank you for your suggestions and review! I'd like to agree first on the calculations/algorithm and, after that, deal with the cleanup of the code.

For background: I somehow mistakenly took derive_t as a double and that's why in this and some other plugins I used casting to double asap. My mistake and I'll deal with fixing it in this code as well.
 
I looked through the proposed value_to_rate function and I don't think it would make the code easier. As I understand, value_to_rate takes the derivative that can be later used in plugin. In the proposed implementation, I report just the fraction of time spent by either one or all CPUs at given frequency. Now the units are not ticks, but fraction of time. So, its easy for user to see immediately how much the corresponding frequency was used. As you could see on the graph, it also takes into account that CPU was inactive in rather large fraction of time (sleep on mobile device) leading to the total less than 100%.

Now, to avoid rounding errors, I suggest to change the units and move to  total_time_in_ms, as we did in cpusleep. Its a bit of a pain to change units, since I have users on Sailfish who run it on their devices, but its better to do it now, before merging. I'll just have to account for it in GUI.

When reporting total_time_in_ms, it should be fine to sum up all values and feed it to the writers as derive_t. The difference between this approach and taking derivatives separately maybe in some corner cases, like rollover. However, to have an impact, single rollover of one of the summed values should not have any impact. How multiple derive_t roll overs during a single measurement would influence the result, I cannot say at the moment.

Best wishes,

rinigus

octo:
Hi @rinigus,

if the CPUs are sleeping and not actually hot-plugged, then summing up `derive_t`s should be good enough in this case. It's important, though, that counters are not reset or disappear from view when a CPU is sleeping. Overflows are not a concern: assuming that the kernel is using 64bit wide counters, it will take ~292 million years to overflow.

Changing the type to `total_time_in_ms` sounds good.

I still think that there is potential in calculating the average frequency from the distribution: right now the *cpufreq plugin* only reports the *current* frequency of the CPU, which is read only during collectd's update cycle, which may skew the measurement quite a bit. By using the distribution I think we can do much better. I'm happy to leave file a feature request issue and handle this in a future PR though.

Best regards,
—octo

rinigus:
Morning @octo!

If total_time_in_ms is fine, then I'll rewrite it to use that and as much integer math as I can.

Now few remarks regarding hot-plugged, Intel CPUs, and stats availibility.

For hot-plugged CPUs, I have only one platform to test it - Nexus 4 phone running Sailfish OS, Linux kernel 3.4. On this platform, CPUs are not hot-plugged by kernel directly, but by external qcom utility. Either because of that or some implementation detail in the kernel, counters seem to reset on hot plug/unplug cycle (and corresponding disappear in /sys folder). This is in contrast to cpuidle statistics which does not disappear and stays on plugging/unplugging CPUs on Nexus 4 [the cpuplugin is ready as well and I'll submit it a bit later]. So, I consider it a kernel limitation and would have to warn the users regarding it. Whether its limited to Nexus 4 and other qcom phones, I don't know.

One other limitation is related to the new P-State drivers by Intel CPUs. On these CPUs, frequency is not really driven by kernel and the stats on them is unavailable. So, only reported value is the current one which can be significantly skew the measurements, as you mention (collectd work driving cpu frequency up, for example). For these CPUs, following idle states makes more sense and I am doing it through other plugin, cpuidle, mentioned above.
 
Finally, to have stats available, user has to enable corresponding kernel module, which maybe not on some distributions. On Gentoo, it depends on the user.

These are the limitations that I am aware of. I'd suggest to keep implementation still simple and warn users on possible skews in the calculated distributions. In the end, on phones, you want not exact values but get an idea on whats going on. For that, the current implementation should be reasonable, in my eyes. But let me know your opinion and let's agree on overall implementation. After agreeing, I'll work on implementing it.

Best wishes,

rinigus

rinigus:
@octo : ping! Just a reminder - waiting for your reply regarding cpufreq calculations. 

In addition, do you have already some C++ map-like functionality in collectd? In C++, I could have used std::map to calculate easily the time spent by all cpus on particular frequency. Do you have similar functionality implemented already in some common files?

Best wishes,

rinigus

octo:
@rinigus Sorry, didn't realize you were waiting for me.

There are two options:

*   Convert the derives to a rate and sum up the rates. These rates are a percentage, so I'd use the type `percentage`, which is easy to grasp by users. (the time derivative of a 10ms counter is "10ms / s" = 1/100 = 1%. The rates will not necessarily sum up to 100; the remainder is the time the CPU was suspended.)
 *   Postpone this, sum up the derives directly and report them as `total_time_in_ms`. Add code to convert to *rates* when it turns out that summing the derives results in too many broken datapoints. There is going to be an extra step converting the rates back to a derive (so we can continue to report `total_time_in_ms`), but it's not the end of the world.

Either option is okay for me, though I prefer the first one: it sounds like the plugin would benefit from calculating a rate, since CPUs disappear and re-appear from our view.

Does this unblock you?
Best regards,
—octo

octo:
`src/daemon/utils_avltree.h` is the closest we have to a `std::map`.

rinigus:
@octo:

yes, it does unblock me, thank you :). I'll think a bit and would do as consistent plugin as I can. 

thank you,

rinigus

rinigus:
Dear @octo:

here is a cleaned up code and I think it is ready for the new review.

1. I modified the stats to be reported in total_time_in_ms. As a result, precision has visibly improved.

2. I added comments that should help the reader to get through the code easier. I looked into partitioning _init function, but couldn't identify a good partitioning that would not make the code harder to grasp.

3. As for hot-plugged CPUs with the stats disappearing between readouts. It seems to me that this is a limitation in kernel implementation on the phone that I have and using the kernel that I am using now. Whether it is a limited case for this particular hardware/kernel/userspace combination, I don't know. In my case, I can still use the reported data, but have to treat it with caution. Unfortunately, in my case, since the data is reset between readouts, there is no data to report either. For me, the current implementation would still help me to catch cases when multiple CPUs are suddenly switched on and start working on high frequencies. I have added corresponding information to the documentation of the plugin to warn users. I'll keep my eyes open with respect of disappearing stats and maybe it should be reported as an issue to the kernel developers.

Best wishes,

rinigus

rinigus:
@octo: Please let me know if I am supposed to resolve the merge conflict. It looks that there have been large changes in the code over the whole tree, which makes older PRs problematic. So, it would be good to know what to do with this PR.

Best,

rinigus

rubenk:
Hey @rinigus. Yes, please resolve the conflict. If you merge master instead of rebasing that will keep the history intact. As far as I can see there's just one small conflict in cpufreq.c around https://github.com/collectd/collectd/pull/1823/files#diff-614f18f527573074470e543a663b90edL81 that should be easy to resolve. You changed the ordering of the variables, so that will make it a tiny bit harder.

src/cpufreq.c
```diff
@@ -26,39 +44,213 @@
...
+	// at this particualr time moment
 
-	while (1)
+	done = 0;
+	for (num_cpu = 0; !done; ++num_cpu)
 	{
 		status = ssnprintf (filename, sizeof (filename),
-				"/sys/devices/system/cpu/cpu%d/cpufreq/"
-				"scaling_cur_freq", num_cpu);
+							"/sys/devices/system/cpu/cpu%zu",
```
rubenk: Please keep the existing indentation intact. This only lines out if you use 4 space tabs. It also makes it hard to see what this actual change does.

src/cpufreq.c
```diff
@@ -26,39 +44,213 @@
...
+							num_cpu);
+
 		if ((status < 1) || ((unsigned int)status >= sizeof (filename)))
+		{
+			INFO ("cpufreq plugin: error in ssnprintf");
 			break;
+		}
 
-		if (access (filename, R_OK))
+		if (access(filename, R_OK))
```
rubenk: Don't change the existing style.

src/cpufreq.c
```diff
@@ -26,39 +44,213 @@
...
+			// that is used later as well. Format of the
+			// file is:
+			//
+			// hertz1 time_in_10ms
+			// hertz2 time_in_10ms
+			// ...
+			//
+			// At this stage, only hertz is used.
+			status = ssnprintf (filename, sizeof (filename),
+								"/sys/devices/system/cpu/cpu%zu/cpufreq/stats/time_in_state",
```
rubenk: 4-space tabs again?

src/cpufreq.c
```diff
@@ -71,67 +263,242 @@ static void cpufreq_submit (int cpu_num, double value)
...
+			{
+				// cpu could be just switched off
+				// let's check the next one
+				continue;
+			}
+
+			if (fgets (buffer, sizeof(buffer), fp) == NULL)
+			{
+				char errbuf[1024];
+				WARNING ("cpufreq: fgets: %s",
```
rubenk: Same here

src/cpufreq.c
```diff
@@ -71,67 +263,242 @@ static void cpufreq_submit (int cpu_num, double value)
...
+						 sstrerror (errno, errbuf,
+									sizeof (errbuf)));
+				fclose (fp);
+				return (-1);
+			}
+
+			if (fclose (fp))
+			{
+				char errbuf[1024];
+				WARNING ("cpufreq: fclose: %s",
```
rubenk: And here.

src/cpufreq.c
```diff
@@ -71,67 +263,242 @@ static void cpufreq_submit (int cpu_num, double value)
...
+				// CPU could be just switched off
+				// let's check the next one
+				continue;
+			}
+
+			while (fgets(buffer, sizeof(buffer), fp) != NULL)
+			{
+				numfields = strsplit (buffer, fields, STATIC_ARRAY_SIZE(fields));
+
+				if ( numfields != 2 ) // supported format contains 2 fields
```
rubenk: No spaces inside the braces.

src/cpufreq.c
```diff
@@ -71,67 +263,242 @@ static void cpufreq_submit (int cpu_num, double value)
...
+							if (hertz == hertz_all_cpus[j])
+							{
+								found = 1;
+								time_all_cpus[j] += time;
+							}
+							else
+								++j;
+						}
+
+						if (found) last_index = j;
```
rubenk: Insert a linebreak please.

src/cpufreq.c
```diff
@@ -26,39 +44,213 @@
...
+		if (hertz_size == 0)
+		{
+			INFO ("cpufreq plugin: cannot find any frequencies in the stats");
+			plugin_unregister_read ("cpufreq");
+			return (0);
+		}
+
+		// Allocate memory for time array that will be used in
+		// read function.
+		time_all_cpus = (derive_t*)malloc (hertz_size * sizeof(derive_t));
```
rubenk: - The cast is unneccesary.
- sizeof(time_all_cpus)

rinigus:
@rubenk : thanks for a tip. I'll do the merge with the upstream and let's see how it will go

(and your comments are flying in as I type). thank you for your work!

src/cpufreq.c
octo: `sizeof (*time_all_cpus)`

Otherwise you get the size of the pointer.

rinigus: I was just looking into it due to the same doubt. I though I missed something in C [most of my work is in c++, so some adjustment is needed]. I'll use *time_all_cpus, that makes it better tahn using a type.

rinigus:
@rubenk: I think that all changes are in and the conflict is resolved. Should be ready for re-review.

Best wishes,

rinigus

src/cpufreq.c
```diff
@@ -18,48 +18,238 @@
...
+ *
+ *  Frequencies in Hz when reported as a current value
+ *  Frequencies in kHz when used as a type instance for distributions
+ *  Distribution utilization for each CPU frequency in milliseconds that this frequency was
+ *   used per second of wall time
+ *
  **/
 
 #include "collectd.h"
-
```
rubenk: Please leave this line here. It's there so that clang-format keeps this header on top, see a5377cf

rinigus: sorry, didn't notice that. my mistake, corrected below.

src/cpufreq.c
```diff
@@ -27,39 +45,212 @@
...
+		if (hertz_size == 0)
+		{
+			INFO ("cpufreq plugin: cannot find any frequencies in the stats");
+			plugin_unregister_read ("cpufreq");
+			return (0);
+		}
+
+		// Allocate memory for time array that will be used in
+		// read function.
+		time_all_cpus = malloc (hertz_size * sizeof(*time_all_cpus));
```
rubenk: Are you sure dereferencing time_all_cpus is right? It is not a pointer.

src/cpufreq.c
```diff
@@ -27,39 +45,212 @@
...
 	return (0);
 } /* int cpufreq_init */
 
-static void cpufreq_submit (int cpu_num, double value)
+static int cpufreq_shutdown (void)
+{
+	if (report_distribution)
+	{
+		sfree (hertz_all_cpus);
+		hertz_all_cpus = NULL;
```
rubenk: sfree already sets its argument to NULL.

src/cpufreq.c
```diff
@@ -27,39 +45,212 @@
...
+
+		sfree (time_all_cpus);
+		time_all_cpus = NULL;
```
rubenk: same here.

src/cpufreq.c
rinigus: time_all_cpus is a pointer (see line 66). initially its value is null, but since sizeof is evaluated by compiler it returns the size of the var type its pointing to.

src/cpufreq.c
rinigus: Right, thank you!

src/cpufreq.c
rubenk: Ah yes, missed that little '*' after derive_t.

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+	int status;
+	_Bool done;
 
-	num_cpu = 0;
+	// Determine number of cpus. Here, just check whether cpus are
+	// there. since cpus can be hot-plugged, cpufreq could be absent
+	// at this particualr time moment
 
-	while (1)
+	done = 0;
```
rubenk: I preferred the while loop to be honest.

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
 	{
 		status = ssnprintf (filename, sizeof (filename),
-				"/sys/devices/system/cpu/cpu%d/cpufreq/"
-				"scaling_cur_freq", num_cpu);
+				    "/sys/devices/system/cpu/cpu%zu",
+				    num_cpu);
+
 		if ((status < 1) || ((unsigned int)status >= sizeof (filename)))
+		{
+			INFO ("cpufreq plugin: error in ssnprintf");
```
rubenk: If this happens there is something wrong, so ERROR is appropriate.

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+	done = 0;
+	for (num_cpu = 0; !done; ++num_cpu)
 	{
 		status = ssnprintf (filename, sizeof (filename),
-				"/sys/devices/system/cpu/cpu%d/cpufreq/"
-				"scaling_cur_freq", num_cpu);
+				    "/sys/devices/system/cpu/cpu%zu",
+				    num_cpu);
+
 		if ((status < 1) || ((unsigned int)status >= sizeof (filename)))
```
rubenk: (size_t)

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+		{
+			INFO ("cpufreq plugin: error in ssnprintf");
 			break;
+		}
 
 		if (access (filename, R_OK))
```
rubenk: comparing with 0 explicitly would be more clear.

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+	// available, the corresponding frequency would be missing
+	// from the list.
+	if (num_cpu > 0 && report_distribution && !report_by_cpu)
+	{
+		FILE *fp;
+		char *fields[2];
+		char buffer[256];
+		int numfields;
+		long int hertz;
+		int found;
```
rubenk: Could you move these inside the for loop?

src/cpufreq.c
```diff
@@ -72,66 +260,241 @@ static void cpufreq_submit (int cpu_num, double value)
...
-		/* You're seeing correctly: The file is reporting kHz values.. */
-		val = atoll (buffer) * 1000;
+				time = atoll (fields[1]);
+				time *= 10; // cpufreq-stat module reports in 10ms
 
-		cpufreq_submit (i, val);
+				if (hertz <= 0)
+				{
+					WARNING ("cpufreq: something is wrong in %s: hertz = %ld", filename, hertz);
+					WARNING ("cpufreq: line in question %s %s", fields[0], fields[1]);
```
rubenk: I'd prefer if you keep log messages to a single line.

src/cpufreq.c
```diff
@@ -72,66 +260,241 @@ static void cpufreq_submit (int cpu_num, double value)
...
+					fclose (fp);
+					return (-1);
+				}
+
+				if (report_by_cpu)
+				{
+					status = ssnprintf (statind, sizeof (statind),
+										"%zu", i);
```
rubenk: indentation

src/cpufreq.c
```diff
@@ -72,66 +260,241 @@ static void cpufreq_submit (int cpu_num, double value)
...
+						WARNING("cpufreq plugin: error in ssnprintf");
+						fclose (fp);
+						return (-1);
+					}
+
+					cpufreq_submit_distribution_value (statind, hertz, time);
+				}
+				else
+				{
+					int found = 0;
```
rubenk: - Make this a _Bool?
- Move it into the else clause since it's only used there.

src/cpufreq.c
```diff
@@ -72,66 +260,241 @@ static void cpufreq_submit (int cpu_num, double value)
...
+						// and increment it by one. In practice, when
+						// all CPUs have the same frequencies it would
+						// work perfectly. If CPUs have different
+						// frequencies, the further iteration maybe
+						// required. For the first CPU and its first
+						// frequency, last_index is set to
+						// hertz_size-1, which, after increment and %
+						// operator would lead to zero.
+						size_t j = last_index + 1;
+						while (!found && j!=last_index)
```
rubenk: spaces around operators.

src/cpufreq.c
```diff
@@ -72,66 +260,241 @@ static void cpufreq_submit (int cpu_num, double value)
...
+				}
+			}
+
+			fclose (fp);
+		}
+
+		// When distributions are reported as an average among all
+		// CPUs, the times are divided by number of CPUs.
+		if (!report_by_cpu)
+			for (size_t i=0; i < hertz_size; ++i)
```
rubenk: Spaces around operators.

src/cpufreq.c
```diff
@@ -72,66 +260,241 @@ static void cpufreq_submit (int cpu_num, double value)
...
+				cpufreq_submit_distribution_value( NULL, hertz_all_cpus[i],
```
rubenk: No extra spacing between the brackets.

rubenk:
@rinigus all done, it's yours :)

rinigus:
thank you! getting on with it :)

rinigus:
and here we go again :). Your suggestions should be all in, including the while loop.

rinigus:
Hi! Recently realized that you have 'Waiting' tag attached to this PR. I hope that you don't wait for me - I think that I incorporated all suggestions. 

cheers,

rinigus

octo:
Hi @rinigus, I'm currently on vacation and will take another look when I'm back next week. Cheers, octo

rinigus:
Enjoy the vacation! I am  :)

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+				// this is a new frequency. add this
+				// to the list at the end. when
+				// scanning for stats later,
+				// frequencies should come in the same
+				// order, so the sorting is not needed
+				// as such (see read function below).
+				if (!found)
+				{
+					hertz_size += 1;
+					hertz_all_cpus = (long int*)realloc (hertz_all_cpus, hertz_size * sizeof(long int));
```
octo:
```c
long int *temp;

temp = realloc (hertz_all_cpus, (hertz_size +1) * sizeof(*hertz_all_cpus));
if (temp)
{
  sfree (hertz_all_cpus);
  // other error handling
}
hertz_all_cpus = temp;
hertz_all_cpus[hertz_size] = hertz;
hertz_size++;
```

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+					if (hertz_all_cpus == NULL)
+					{
+						INFO ("cpufreq plugin: realloc failed for hertz_all_cpus size %zu", hertz_size);
```
octo: Something went wrong and we *cannot* continue → use `ERROR()`.

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+					hertz_all_cpus[ hertz_size-1 ] = hertz;
+				}
+			}
+
+			fclose (fp);
+		}
+
+		if (hertz_size == 0)
+		{
+			INFO ("cpufreq plugin: cannot find any frequencies in the stats");
```
octo: Dito: `ERROR()`

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+			plugin_unregister_read ("cpufreq");
+			return (0);
+		}
+
+		// Allocate memory for time array that will be used in
+		// read function.
+		time_all_cpus = malloc (hertz_size * sizeof(*time_all_cpus));
+		if (time_all_cpus == NULL)
+		{
+			INFO ("cpufreq plugin: malloc failed for time_all_cpus size %zu", hertz_size);
```
octo: Dito: `ERROR()`

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+		INFO ("cpufreq plugin: found %zu different frequencies", hertz_size);
+	}
+
 	return (0);
 } /* int cpufreq_init */
 
-static void cpufreq_submit (int cpu_num, double value)
+static int cpufreq_shutdown (void)
+{
+	if (report_distribution)
```
octo: Remove the `if`, just call `sfree()` – it's shorter and shorter is easier to read.

src/cpufreq.c
```diff
@@ -72,66 +260,240 @@ static void cpufreq_submit (int cpu_num, double value)
...
 }
 
 static int cpufreq_read (void)
 {
-        int status;
-	unsigned long long val;
 	FILE *fp;
-	char filename[256];
-	char buffer[16];
+	char filename[128];
```
octo: Please use `PATH_MAX` rather than choosing an arbitrary number.

src/cpufreq.c
```diff
@@ -72,66 +260,240 @@ static void cpufreq_submit (int cpu_num, double value)
...
+
+			/* You're seeing correctly: The file is reporting kHz values.. */
+			val = atoll (buffer) * 1000;
+
+			cpufreq_submit_current_value (i, val);
 		}
+	} /* end of if ( !report_distribution ) */
+
+	else // stats on CPU frequency distribution is used
+	{
```
octo: This block is quite long and has many levels of indentation. Please move it to its own function.

src/cpufreq.c
```diff
@@ -27,39 +45,209 @@
...
+	// different frequencies, all CPUs are checked for available
+	// frequencies.
+	//
+	// Limitation of this approach: if at the time of scan a CPU
+	// that has some frequencies not covered by other CPUs is
+	// switched off and the cpufreq stat for that CPU is not
+	// available, the corresponding frequency would be missing
+	// from the list.
+	if (num_cpu > 0 && report_distribution && !report_by_cpu)
+	{
```
octo: This block is also very long and should go into its own function.

octo:
Hi @rinigus, sorry that this took a couple days longer than I though. I've left some review comments inline. Thanks for all your work!

rinigus:
Hi @octo: thank you for reviewing it! I'll make the changes, but it may take couple of days. I'll let you know when changes are in.

best,

rinigus

PS: congratulations on new release!

octo:
Thank you :-)

rinigus:
Dear @octo:

I think that I addressed all your comments in the last revision. Please review when you get time for it. Took a bit longer than I expected since there were significant changes in upstream :)

Best 

rinigus

rinigus:
Hi @octo:

it would be great to get some ETA on when do you plan to look into this and radio plugins. If warned in advance, I could try to plan my time as well and be ready to address the questions. I think the questions are addressed though and the both plugins should be quite ready for merge.

Sorry to bug you regarding these plugins - just don't want them to get stale... Radio plugin already requires some additional patch, but it requires your decision on whether you want it all first.

Best wishes,

rinigus

rinigus:
@octo: ping, as instructed on 5.7 release schedule email :)

rinigus:
Just to let you know that the PR has been reformatted in accordance with the new formatting policy. I have great hopes that this new policy will help you out!

PR #1824, opened 2016-07-30
Protobuf autoconf adjustments

### LINKED ISSUE #1817: Build system: Use pkg-config to configure the grpc plugin.
The *gRPC plugin* is currently a bit annoying to build: it requires a very recent version of *protoc*, so that you will likely need to compile it manually. At the same time, it only searches `$PATH` for the binaries and does not allow to override the path. And to make it even worse, the check for the headers and libraries does not take a prefix, so you're stuck with manually editing `CPPFLAGS` and `LDFLAGS`.

At the same time, both protobuf and grpc provide a `.pc` file for *pkg-config*. Ideally we'd take the prefix information and other required flags from that config.

rubenk:
Fixes #1817

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -76,6 +76,7 @@ then
 	AC_MSG_ERROR([bison is missing and you do not have ${srcdir}/src/liboconfig/parser.c. Please install bison])
 fi
 
+AC_ARG_VAR([PROTOC], [path to the protoc binary])
 AC_PATH_PROG([PROTOC], [protoc])
 have_protoc3="no"
 if test "x$PROTOC" != "x"; then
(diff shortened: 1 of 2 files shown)


octo:
Thank you so much @rubenk! LGTM

For the record, having only one of `--with-library=/prefix` and *pkg-config* would have been good enough for me, but obviously this is fine, too.

rubenk:
Thanks for the review!

Just using pkg-config would be a godsend, and surely would make configure.ac a whole lot shorter. I've seen a few issue with .pc files though (missing, plain wrong etc) that having an option to override it is nice. For example:

```
~$ PKG_CONFIG_PATH=/usr/local/lib/pkgconfig/ pkg-config --cflags protobuf
-pthread -I/usr/local/include
```
-pthread is a GCC-specific option, and it doesn't need to be there.
```
~$ PKG_CONFIG_PATH=/usr/local/lib/pkgconfig/ pkg-config --libs protobuf
-L/usr/local/lib -lprotobuf -pthread -lpthread
```
libprotobuf is already linked to libpthread, so it shouldn't need to specify it.

rubenk:
Another example is that libgrpc++ doesn't link to libprotobuf (it probably should, unless it's optional?) but doesn't add -lprotobuf in its .pc file.

I can forgive them, since these are pretty young projects, but look at mysql:
```
$ pkg-config --libs mariadb
-L/usr/lib64/mysql -lmysqlclient -lpthread -lz -lm -ldl -lssl -lcrypto
```
They should have gotten this right by now ;)

Ok, rant over, I'll squash and merge.

octo:
ACK, many projects just dump the CFLAGS / LDFLAGS that they themselves were built with in there, which is not the best option. Thanks for squashing! :)

rubenk:
One thing I'd like to do is write a generic macro for this, since we basically do the same for each library we check. something like `COLLECTD_CHECK_LIB([$lib], [$list of headers to check])`

PR #1825, opened 2016-07-30
collectd.spec: el7 fails to find python without which

ripienaar:
On very minimal build centos machines the yum-builddep should install
all it needs to succesfully build on centos, but this misses which and
so python cant be found.

### DIFF
--- a/contrib/redhat/collectd.spec
+++ b/contrib/redhat/collectd.spec
@@ -228,7 +228,7 @@ Source:		https://collectd.org/files/%{name}-%{version}.tar.bz2
 License:	GPLv2
 Group:		System Environment/Daemons
 BuildRoot:	%{_tmppath}/%{name}-%{version}-root
-BuildRequires:	libgcrypt-devel, kernel-headers, libtool-ltdl-devel, libcap-devel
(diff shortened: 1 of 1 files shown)


ripienaar:
Not sure how you'd even build a centos machine without which, but the official centos docker images dont have it:

```
% docker run -ti --rm centos:7 rpm -q which
package which is not installed
```

this fix enables you to build the rpm on these via just using `yum-builddep`

rubenk:
Thanks for your PR!

Interestingly I never hit this on Fedora or CentOS before in mock builds.
I don't see the which rpm being pulled in into the build roots either.

Just a thought, perhaps configure uses the build in which in the shell, depending on which shell it uses?

In any case, I'll merge this. Will also look into why were' using which in the first place and not AC_PATH_PROG.

ripienaar:
Yeah first for me too, the official centos docker images must have become a bit more minimal recently, but it builds now :) thnx

rubenk:
Ok, good to hear that, thanks :)

rubenk:
I just pushed 6431e24 to master that removes `which` altogether. I'll leave the change in the spec file for a while since it's also used to build older versions.

ripienaar:
👍

PR #1826, opened 2016-07-31
new plugin: cpusleep 

rinigus:
/* By mistake, I closed a wrong branch leaving to removal of PR. Here, the old PR is resubmitted */
/* earlier, it had flags Linux Waiting and probably one more, but don't remember which */

This plugin measures time spent by CPU in deep sleep. It uses a difference in time of two Linux clocks and calculates the time spent in sleep. On mobile platforms with autosuspend, it can be used to track time while device was in lowest power consumption mode.


configure.ac
```diff
@@ -755,6 +755,23 @@ AC_CHECK_HEADERS(net/pfvar.h,
 have_termios_h="no"
 AC_CHECK_HEADERS(termios.h, [have_termios_h="yes"])
 
+# For cpusleep plugin
+AC_CACHE_CHECK([whether clock_boottime and clock_monotonic are supported],
+		       [c_cv_have_clock_boottime_monotonic],
+		       AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
+[[[
```
rubenk: minor nit, double quotes are sufficient.

src/collectd.conf.pod
```diff
@@ -1457,6 +1457,15 @@ installed) to get the current CPU frequency. If this file does not exist make
 sure B<cpufreqd> (L<http://cpufreqd.sourceforge.net/>) or a similar tool is
 installed and an "cpu governor" (that's a kernel module) is loaded.
 
+=head2 Plugin C<cpusleep>
+
+This plugin doesn't have any options. It reads CLOCK_BOOTTIME and
+CLOCK_MONOTONIC and reports the differences between these
```
rubenk: difference (singular)?

src/collectd.conf.pod
```diff
@@ -1457,6 +1457,15 @@ installed) to get the current CPU frequency. If this file does not exist make
...
+clocks. Since BOOTTIME clock increments while device is suspended and
+MONOTONIC clock does not, the derivative of the difference between
+these clocks gives the relative amount of time the device has spent in
+suspend state. The recorded value is in milliseconds / seconds.
```
rubenk: I'm confused, is it milliseconds, seconds or both? ;)

src/cpusleep.c
```diff
@@ -0,0 +1,87 @@
...
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+ * Authors:
+ *	 rinigus <http://github.com/rinigus>
+
+ CPU sleep is reported in milliseconds / s. For that, derive type was
+ selected and the time difference between BOOT and MONOTONIC clocks
+ fed to RRD
```
rubenk: I'm sorry, but I don't understand this last sentence, and how RRD comes into play?

src/cpusleep.c
```diff
@@ -0,0 +1,87 @@
...
+		return (-1);
+	}
+
+	if ( clock_gettime(CLOCK_MONOTONIC, &m) < 0 )
+	{
+		ERROR("cpusleep plugin: clock_monotonic failed");
+		return (-1);
+	}
+
+	double db = b.tv_sec + 1e-9 * b.tv_nsec;
```
octo: I'm a bit concerned about this cast to `double`: the current time should fit into a `derive_t` (aka. `int64_t`) and would even have a bit to spare), so we would be able to calculate this without loss of precision.

src/types.db
```diff
@@ -37,6 +37,7 @@ count                   value:GAUGE:0:U
 counter                 value:COUNTER:U:U
 cpu                     value:DERIVE:0:U
 cpufreq                 value:GAUGE:0:U
+cpusleep                value:DERIVE:0:U
```
octo: I'd prefer to use the existing type `total_time_in_ms` which reports metrics in "milliseconds per second". We would need to divide the metric by 10^6 which is probably good enough.

rinigus:
@rubenk and @octo:

thank you for the feedback! I'll look into it a bit later, in 1-2 days. sorry for delay.

rinigus:
@rubenk and @octo:

I've inserted the changes: units in description are spelled out fully now; used total_time_in_ms for reporting; rounding errors reduced by calculating seconds and ns separately. I hope its OK now.

Best wishes,

rinigus

PS: (start of  ping). Since you plan to release a new version very soon, I guess that Sailfish OS pull request  https://github.com/collectd/collectd/pull/1736 will not make it this time. It may require some work still to get all done correctly. While it works in practice, we may wish to change some things before merging. (end of ping).

src/cpusleep.c
```diff
@@ -69,12 +69,9 @@ static int cpusleep_read(void)
...
 	}
 
-	double db = b.tv_sec + 1e-9 * b.tv_nsec;
-	double dm = m.tv_sec + 1e-9 * m.tv_nsec;
-
 	// to avoid false positives in counter overflow due to reboot,
-	// derive is used
-	derive_t sleep = (derive_t) ((db-dm) * 1000);	 
+	// derive is used. Sleep is calculated in milliseconds
+	derive_t sleep = (derive_t) ((b.tv_sec - m.tv_sec)*1e3 + (b.tv_nsec - m.tv_nsec)*1e-6);
```
octo: It's non-obvious, but `1e3` and `1e-6` are of type `double`, so floating point conversion is still happening.

I'd suggest to use integer arithmetic instead:

```c
derive_t sleep = ((derive_t) (b.tv_sec - m.tv_sec)) * 1000
    + ((derive_t) (b.tv_nsec - m.tv_nsec)) / 1000000;
```

octo:
Thanks @rinigus! I've left a note regarding implicit double conversion. Once that is fixed I think this is good to go.

rinigus:
@octo: I calculated using derive_t just in seconds and nanoseconds first. Otherwise, on 32-bit architectures, the int (if used as time_t) may overflow in ~24 days if we multiply it by 1000.

best wishes,

rinigus

octo:
@rinigus That's why I generally push to cast to `derive_t` early: it's guaranteed to be 64bit wide. LGTM

PR #1828, opened 2016-08-01
bind plugin: Add support for multiple servers

mirchr:
Add support for multiple servers in the bind plugin. The logic changes were modeled after the apache plugin. With this change the config format has changed slightly. Each configured server requires to be wrapped around Server tags. All original configuration options remain the same. I have tested the patch using two name servers and it appears to be working as expected. There is no built-in limitation for the number of servers that can be monitored.

Current format
```
<Plugin "bind">
  URL "http://localhost:8053/"
  OpCodes         true
  QTypes          true
  
  ServerStats     true
  ZoneMaintStats  true
  ResolverStats   false
  MemoryStats     true
  
  <View "_default">
    QTypes        true
    ResolverStats true
    CacheRRSets   true
    
    Zone "127.in-addr.arpa/IN"
  </View>

</Plugin>
```

Updated format
```
<Plugin "bind">
  <Server "ns1">
    URL "http://localhost:8053/"
    OpCodes         true
    QTypes          true
  
    ServerStats     true
    ZoneMaintStats  true
    ResolverStats   false
    MemoryStats     true

    <View "_default">
      QTypes        true
      ResolverStats true
      CacheRRSets   true
    
      Zone "127.in-addr.arpa/IN"
    </View>
  </Server>

  <Server "ns2">
    URL "http://ns2.example.com:8053/"
    OpCodes         true
    QTypes          true
  
    ServerStats     true
    ZoneMaintStats  true
    ResolverStats   false
    MemoryStats     true

    <View "_default">
      QTypes        true
      ResolverStats true
      CacheRRSets   true
    
      Zone "127.in-addr.arpa/IN"
    </View>
  </Server>
</Plugin>
```


octo:
Hi @mirchr,

thank you very much for this patch! collectd is promising its users that "major" versions stay backwards compatible, and this compatibility is mostly defined by:

*   The new version will run with the old config.
*   The new version with the old config will _at least_ generate the metrics of the previous version.

So in order to merge this, we need backwards compatibility code. Maybe we could loop over the children of the root `oconfig_item_t *ci` and if one of the children is *not* a `Server` block pass `ci` itself to `bind_config_add()` …

Also, the manpage (`src/collectd.conf.pod`) and example config (`src/collectd.conf.in`) need to be updated to document this change.

Best regards,
—octo

mirchr:
@octo Thanks for the feedback! I will address the backwards compatibility issue and also update the documentation.

octo:
Hi @mirchr, are you still interested in working on this?

mirchr:
@octo Yes - sorry. It fell of my radar. I will start working on it this week.

octo:
No worries, just doing some housekeeping ;-)

src/bind.c
```diff
@@ -1663,20 +1713,55 @@ static int bind_config_add_view (oconfig_item_t *ci) /* {{{ */
...
+  bind_t *st;
+  int status;
+
+  st = malloc (sizeof (*st));
+  if (st == NULL)
+  {
+    ERROR ("bind plugin: malloc failed.");
+    return (-1);
+  }
+  memset (st, 0, sizeof (*st));
```
octo: Can you please use `calloc` instead of `malloc` + `memset`:

```c
st = calloc (1, sizeof (*st));
```

octo:
Hi @mirchr, did you just rebase the PR? That commit doesn't appear to address the review comments …

src/bind.c
mirchr: Yes. Will do.

mirchr:
@octo I rebased it last night and fixed a merge conflict caused by 2761915bed8c6caea41018be3e675aa712cc0b0a . Now that it's clean I will update it to use calloc and work on the remaining tasks.

mirchr:
@octo After reviewing at the changes required to make the new <Server></Server> format work I was thinking about removing the new format and just enhance the current format by allowing multiple URLs. When a URL is read this will be treated as a new server and all following options will be applied to that server until another URL is found. I think this will be cleaner and more supportable long term. The data can be tracked by parsing the server(and maybe the port also) out of the URL. Let me know what you think


Example new configuration.
```
<Plugin "bind">
  URL "http://localhost:8053/"
  OpCodes         true
  QTypes          true

  ServerStats     true
  ZoneMaintStats  true
  ResolverStats   false
  MemoryStats     true

  <View "_default">
    QTypes        true
    ResolverStats true
    CacheRRSets   true

    Zone "127.in-addr.arpa/IN"
  </View>

  URL "http://ns2.example.com:8053/"
  OpCodes         true
  QTypes          true

  ServerStats     true
  ZoneMaintStats  true
  ResolverStats   false
  MemoryStats     true

  <View "_default">
    QTypes        true
    ResolverStats true
    CacheRRSets   true

    Zone "127.in-addr.arpa/IN"
  </View>

</Plugin>
```

yellowpattern:
Your original proposal was better than the latter one. Distinct blocks +1. If no blocks => 1 server only +1

PR #1829, opened 2016-08-01
Add a config file for clang-format

rubenk:
To use it, use `clang-format --style=file`

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,5 @@
+---
+BasedOnStyle: LLVM
+IncludeCategories:
+  - Regex: '"collectd.h"'
+  - Priority: -1


rubenk:
This is a good start, but I can't get the IncludeCategories option to work, it's still reordering collectd.h. I.e.
```
diff --git a/src/match_empty_counter.c b/src/match_empty_counter.c
index bff9448..c7ab01f 100644
--- a/src/match_empty_counter.c
+++ b/src/match_empty_counter.c
@@ -24,48 +24,45 @@
  *   Florian Forster <octo at collectd.org>
  **/
 
-#include "collectd.h"
 #include "common.h"
 #include "filter_chain.h"
+#include "collectd.h"
```

@octo, any ideas?

rubenk:
Ah, it does leave it alone when there's an empty line beneath it.

octo:
Hm, according to the documentation, that should be it. I only find the [docs for version 4](http://clang.llvm.org/docs/ClangFormatStyleOptions.html), maybe older versions silently ignore options they don't know?

Do we *need* to set *IndentWidth*? I thought that was the LLVM style anyway …

Regarding your question in the other PR: parts of collectd use tabs, other parts two spaces. Four spaces is not currently used anywhere and I would like to prevent introducing yet another indentation width.

rubenk:
Clang-format from clang 3.8 seems to handle all the options in the docs for version 4.
You're right, IndentWidth 2 is the default for LLVM style, I'll remove it.

Regarding my question in the other PR, I was joking, 2 spaces is fine by me :)
There are some files with 4 spaces, but most of them have a mix of spaces and tabs.

Just for kicks I did some tweaking to see if I could match the current Collectd style better and reduce the 180000 line diff clang-format generates a bit. There are some weird edge cases but in general I think it handles it pretty well.

Here's what I have now:
```
---
BasedOnStyle: LLVM
ColumnLimit: 80
AlignOperands: 'true'
AllowShortFunctionsOnASingleLine: Empty
IncludeCategories:
  - Regex: '"collectd.h"'
  - Priority: -1
BreakBeforeBinaryOperators: NonAssignment
BreakBeforeBraces: Custom
BraceWrapping: {
  AfterControlStatement: 'true'
  AfterFunction: 'true'
  AfterStruct: 'true'
  BeforeElse: 'true'
  AfterEnum: 'true'
}
SpaceBeforeParens: Always
```

PR #1831, opened 2016-08-02
Ai hints cleanup

rubenk:


src/chrony.c
```diff
@@ -306,14 +306,14 @@ static int
 connect_client(const char *p_hostname,
                const char *p_service, int p_family, int p_socktype)
 {
-  struct addrinfo hints = { 0 };
-  struct addrinfo *res = NULL, *ressave = NULL;
+  struct addrinfo ai_hints = { 0 };
```
octo: Why is this struct initialized differently? Looks like the other schema would work here, too.

src/write_sensu.c
```diff
@@ -115,9 +115,7 @@ static int sensu_connect(struct sensu_host *host) /* {{{ */
 		host->res = NULL;
 		hints.ai_family = AF_INET;
 		hints.ai_socktype = SOCK_STREAM;
-#ifdef AI_ADDRCONFIG
-		hints.ai_flags |= AI_ADDRCONFIG;
-#endif
+		hints.ai_flags = AI_ADDRCONFIG;
```
octo: You rename the variable to `ai_hints` above. Could you unify this, i.e. change this here, too?

octo: Looking at the context, it looks like `hints` is only used within the `if (!READY)` block. Can we move it into the if block?

octo:
LGTM, but please note inline comments.

src/chrony.c
rubenk: Somehow I missed that, will fix.

src/write_sensu.c
rubenk: Nice catch!

PR #1832, opened 2016-08-03
Require a compiler that understands C99

rubenk:
Upcoming changes will requires C99 support.

For newer compilers this is the default, but older
compilers needs a flag like -std=gnu99 or -std=c99.


octo:
LGTM

PR #1837, opened 2016-08-03
src/plugin.c: Stop write threads before calling shutdown callbacks.

### LINKED ISSUE #1110: Error in `collectd': free(): corrupted unsorted chunks:
When issuing "collectd -T" i often get the following output ([full version here](https://gist.githubusercontent.com/corebug/3c3cebb4397a260b0f90/raw/387079f0d90566d04a5b4c19215d3f0ae7034be5/collectd_fail.txt)): 
```
*** Error in `collectd': free(): corrupted unsorted chunks: 0x00007fb14bf38ea0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x7d1fd)[0x7fb149d3d1fd]
collectd(+0x1051a)[0x7fb14a9d751a]
/lib64/libpthread.so.0(+0x7df5)[0x7fb14a28cdf5]
/lib64/libc.so.6(clone+0x6d)[0x7fb149db61ad]
```

System details:

Component | Version
------------ | -------------
OS | CentOS Linux release 7.1.1503 (Core)
Collectd | 5.4.2 (5.4.2-1.el7, installed from EPEL).
Kernel | 3.10.0-123.el7.x86_64
glibc | 2.17-78.el7
Collectd configuration | [deployed via this template](https://github.com/CiscoCloud/microservices-infrastructure/blob/master/roles/collectd/templates/collectd.conf.j2)

octo:
I have hopes that this fixes #1110, but I'd welcome some testers because I'm not sure how reliably this can be reproduced.

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -2893,6 +2893,11 @@ static int network_write (const data_set_t *ds, const value_list_t *vl,
 {
 	int status;
 
+	/* listen_loop is set to non-zero in the shutdown callback, which is
+	 * guaranteed to be called *after* all the write threads have been shut
+	 * down. */
+	assert (listen_loop == 0);
+
 	if (!check_send_okay (vl))
 	{
 #if COLLECT_DEBUG
--- a/src/plugin.c
+++ b/src/plugin.c
@@ -1731,9 +1731,9 @@ void plugin_shutdown_all (void)
 {
 	llentry_t *le;
 
-	stop_read_threads ();
-
 	destroy_all_callbacks (&list_init);
+
+	stop_read_threads ();
 
 	pthread_mutex_lock (&read_lock);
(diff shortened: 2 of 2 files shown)


rubenk:
I could trigger this after about 5 runs of this magic loop:
`while :; do ~/opt/sbin/collectd -T; if [ $? -eq 139 ]; then break; fi; sleep 1; done`, but it seems this has fixed it, at least on OSX. Will do a build with ASAN on Linux tomorrow to confirm.

rubenk:
As promised I did a build with ASAN on Linux. Before this PR I could trigger the crash in seconds, now, not anymore.

octo:
Sounds good. I expect this is the coverage this patch will likely get, so I'm going to go ahead and merge.

PR #1840, opened 2016-08-04
write_sensu: Re-add substitute for asprintf

### LINKED ISSUE #1798: Substitute for asprintf should be re-added in write_sensu.c
In fact the older Solaris implementations are missing `asprintf`, so the vendor substitute would be good to have. This was already in pull request #912 but probably didn't make it into the release.

rubenk:
asprint is only available from Solaris 10 update 11.

Fixes #1798

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -786,7 +786,7 @@ AC_HEADER_TIME
 #
 # Checks for library functions.
 #
-AC_CHECK_FUNCS(gettimeofday select strdup strtol getaddrinfo getnameinfo strchr memcpy strstr strcmp strncmp strncpy strlen strncasecmp strcasecmp openlog closelog sysconf setenv if_indextoname setlocale)
+AC_CHECK_FUNCS(gettimeofday select strdup strtol getaddrinfo getnameinfo strchr memcpy strstr strcmp strncmp strncpy strlen strncasecmp strcasecmp openlog closelog sysconf setenv if_indextoname setlocale asprintf)
 
 AC_FUNC_STRERROR_R
 
--- a/src/write_sensu.c
+++ b/src/write_sensu.c
@@ -42,6 +42,55 @@
 #define SENSU_HOST		"localhost"
 #define SENSU_PORT		"3030"
 
+#ifdef HAVE_ASPRINTF
+#define my_asprintf asprintf
+#define my_vasprintf vasprintf
+#else
+/*
+ * asprintf() is available from Solaris 10 update 11.
+ * For older versions, use asprintf() portable implementation from
+ * https://github.com/littlstar/asprintf.c/blob/master/
+ * copyright (c) 2014 joseph werle <joseph.werle@gmail.com> under MIT license.
+ */
+
+static int my_vasprintf(char **str, const char *fmt, va_list args) {
+       int size = 0;
+       va_list tmpa;
+       // copy
+       va_copy(tmpa, args);
+       // apply variadic arguments to
+       // sprintf with format to get size
+       size = vsnprintf(NULL, size, fmt, tmpa);
+       // toss args
+       va_end(tmpa);
+       // return -1 to be compliant if
+       // size is less than 0
+       if (size < 0) { return -1; }
+       // alloc with size plus 1 for `\0'
+       *str = (char *) malloc(size + 1);
+       // return -1 to be compliant
+       // if pointer is `NULL'
+       if (NULL == *str) { return -1; }
+       // format string with original
+       // variadic arguments and set new size
+       size = vsprintf(*str, fmt, args);
+       return size;
+}
+
+static int my_asprintf(char **str, const char *fmt, ...) {
+       int size = 0;
+       va_list args;
+       // init variadic argumens
+       va_start(args, fmt);
+       // format and get size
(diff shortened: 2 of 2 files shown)


src/write_sensu.c
```diff
@@ -39,6 +39,53 @@
 #include <stddef.h>
 
 #include <stdlib.h>
+#ifdef HAVE_KERNEL_SOLARIS
```
dago: Shouldn't this be `KERNEL_SOLARIS`? Otherwise the substitute is not used. When I change the line accordingly then I get a compilation error of the following form.
The problem is that the machine where I compile the package has `vasprintf` but I disallow the linker to bind against the libc contract that provides it.
```
gmake[3]: Entering directory '/home/dam/mgar/pkg/collectd/trunk/work/solaris10-sparc/build-isa-sparcv8plus/collectd-5.5.2/src'
  CC       write_sensu.lo
write_sensu.c:52:12: error: static declaration of 'vasprintf' follows non-static declaration
 static int vasprintf(char **str, const char *fmt, va_list args) {
            ^
In file included from ./daemon/collectd.h:34:0,
                 from write_sensu.c:29:
/usr/include/stdio.h:263:12: note: previous declaration of 'vasprintf' was here
 extern int vasprintf(char **, const char *, __va_list);
            ^
write_sensu.c:76:12: error: static declaration of 'asprintf' follows non-static declaration
 static int asprintf(char **str, const char *fmt, ...) {
            ^
In file included from ./daemon/collectd.h:34:0,
                 from write_sensu.c:29:
/usr/include/stdio.h:261:12: note: previous declaration of 'asprintf' was here
 extern int asprintf(char **, const char *, ...);
            ^
Makefile:3937: recipe for target 'write_sensu.lo' failed
gmake[3]: *** [write_sensu.lo] Error 1
```

rubenk: Yeah, it should, good point.

> The problem is that the machine where I compile the package has vasprintf but I disallow the linker to bind against the libc contract that provides it.

Why do you do this? Is this machine running Solaris 10 update 11 and are you trying to emulate update 8?
Did configure determine that asprintf was available?

dago: The problem is that I need a critical ZFS update for the SSD Cache Configuration I run on the buildfarm which is only available on this Kernel Update Release. However, most users run older Solaris Kernel Releases, so I restrict the libc linker contract. `configure` correctly detects `asprintf` to be non-available, but the header files obviously include it.

If you include
```
#ifdef KERNEL_SOLARIS
#ifndef HAVE_ASPRINTF
#define asprintf _asprintf
#define vasprintf _vasprintf
#endif
#endif
```
before all other `#include`
and then
```
#undef vasprintf
#undef asprintf
```
in the Solaris asprintf substitute section it works.

But I could also understand that this is a OpenCSW special case and I would be willing to maintain this patch downstream as it is probably not helpful for everyone.

rubenk: Ah, thanks. Now that I understand it, it should be easy to fix :)
I always thought that AC_CHECK_FUNCS only checked for function declarations, but now it's obvious that is also tries to link with it (which is a good thing).

As for the fix, isn't it enough to just add
```
#undef vasprintf
#undef asprintf
```
in the substitute section?

dago: Unfortunately not as there is a function declaration in the header file. Undefining applies only to the preprocessor and a C declaration cannot be easily "undeclared". So my solution change the declaration by a macro definition to another name and then clears that renaming so the substitute goes in with the right name.

Another solution would be to use something like my_asprintf in the code and then make definitions to that to either the substitute or the builtin.

rubenk: Alright, sounds good. Please try	068bf0e

dago: This looks identical to my code, thanks!

octo: > Another solution would be to use something like my_asprintf in the code and then make definitions to that to either the substitute or the builtin.

That sounds like the better approach to me.

rubenk: Something like 3200ba0?

octo: @rubenk Yep, that's perfect, thanks! Looks good to me, but it's probably better to listen to @dago with regards to the actual issue at hand ;)

rubenk: @dago if you could test 14b8154 I'll go ahead and merge it :)

configure.ac
```diff
@@ -786,7 +786,7 @@ AC_HEADER_TIME
 #
 # Checks for library functions.
 #
-AC_CHECK_FUNCS(gettimeofday select strdup strtol getaddrinfo getnameinfo strchr memcpy strstr strcmp strncmp strncpy strlen strncasecmp strcasecmp openlog closelog sysconf setenv if_indextoname setlocale)
+AC_CHECK_FUNCS(gettimeofday select strdup strtol getaddrinfo getnameinfo strchr memcpy strstr strcmp strncmp strncpy strlen strncasecmp strcasecmp openlog closelog sysconf setenv if_indextoname setlocale, asprintf)
```
dago: The "," must go

rubenk: Fixed and rebased, thanks.

PR #1842, opened 2016-08-04
treewide: declare loop variable in loop expression

rubenk:


octo:
LGTM, build successful, go for it :) Thanks!

rubenk:
Alright, fingers crossed!

rubenk:
So far, so good. I must say it cleans up the code nicely. Now all we need is for contributors to start using this style more...

If you agree I also like to start moving variable declarations to just before usage. Not sure if I should do that before or after we fix indentation.

PR #1843, opened 2016-08-04
mysql plugin: add support for Galera statistics

xinity:
Hoping this one is for good now !

thanks to @XANi for helping me fixing

### DIFF
--- a/src/mysql.c
+++ b/src/mysql.c
@@ -54,6 +54,7 @@ struct mysql_database_s /* {{{ */
 	_Bool master_stats;
 	_Bool slave_stats;
 	_Bool innodb_stats;
+	_Bool wsrep_stats;
 
 	_Bool slave_notif;
 	_Bool slave_io_running;
@@ -179,6 +180,8 @@ static int mysql_config_database (oconfig_item_t *ci) /* {{{ */
 			status = cf_util_get_boolean (child, &db->slave_notif);
 		else if (strcasecmp ("InnodbStats", child->key) == 0)
 			status = cf_util_get_boolean (child, &db->innodb_stats);
+		else if (strcasecmp ("WsrepStats", child->key) == 0)
+			status = cf_util_get_boolean (child, &db->wsrep_stats);
 		else
 		{
 			WARNING ("mysql plugin: Option `%s' not allowed here.", child->key);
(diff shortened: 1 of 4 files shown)


rubenk:
@xinity Thanks! Did you test this?

src/mysql.c
```diff
@@ -28,7 +28,6 @@
  **/
 
 #include "collectd.h"
-
```
octo: Please don't remove this line, it is actually serving a purpose.

src/mysql.c
```diff
@@ -685,6 +687,101 @@ static int mysql_read_innodb_stats (mysql_database_t *db, MYSQL *con)
...
+		{ "wsrep_replicated_bytes",          "total_bytes",  DS_TYPE_DERIVE },
+
+		{ "wsrep_apply_window",              "window_size",  DS_TYPE_GAUGE },
+		{ "wsrep_commit_window",             "window_size",  DS_TYPE_GAUGE },
+
+		{ "wsrep_cluster_size",              "gauge",        DS_TYPE_GAUGE },
+		{ "wsrep_cert_deps_distance",        "gauge",        DS_TYPE_GAUGE },
+
+		{ "wsrep_local_recv_queue",          "queue_length", DS_TYPE_GAUGE },
+		{ "wsrep_local_recv_queue_avg",      "queue_length", DS_TYPE_GAUGE },
```
octo: Are the "*_avg" metrics actually useful? If in doubt, comment them out and let the graphing / dashboarding solution deal with averages.

src/mysql.c
xinity: ok i'll fix it right away

src/mysql.c
xinity: indeed we can comment them out @octo

xinity:
@rubenk yup i'm using it on another repo :)

xinity:
@octo : should be fine by now :)

src/mysql.c
```diff
@@ -685,6 +687,101 @@ static int mysql_read_innodb_stats (mysql_database_t *db, MYSQL *con)
...
+	{
+		ERROR ("mysql plugin: Failed to get wsrep statistics: "
+				"`%s' did not return any rows.", query);
+		mysql_free_result (res);
+		return (-1);
+	}
+
+
+		while ((row = mysql_fetch_row (res)))
+	{
```
octo: Indentation is off here, presumably a tabs vs. spaces issue.

src/mysql.c
```diff
@@ -685,6 +687,101 @@ static int mysql_read_innodb_stats (mysql_database_t *db, MYSQL *con)
...
+		val = atoll (row[1]);
+
+		for (i = 0; metrics[i].key != NULL && strcmp(metrics[i].key, key) != 0; i++)
+			;
+
+		if (metrics[i].key == NULL)
+			continue;
+
+		switch (metrics[i].ds_type) {
+			case DS_TYPE_COUNTER:
```
octo: "you ain't gonna need it": there is no COUNTER type used in `metrics` above, so remove this case here.

src/mysql.c
```diff
@@ -685,6 +687,101 @@ static int mysql_read_innodb_stats (mysql_database_t *db, MYSQL *con)
...
+				derive_submit(metrics[i].type, key, (derive_t)val, db);
+				break;
+		}
+	}
+
+	mysql_free_result(res);
+	return (0);
+} /* mysql_read_wsrep_stats */
+
+
```
octo: nit: please reduce to exactly one empty line

src/mysql.c
```diff
@@ -685,6 +687,101 @@ static int mysql_read_innodb_stats (mysql_database_t *db, MYSQL *con)
...
+		{ "wsrep_local_cert_failures",       "operations",   DS_TYPE_DERIVE },
+		{ "wsrep_local_commits",             "operations",   DS_TYPE_DERIVE },
+		{ "wsrep_local_replays",             "operations",   DS_TYPE_DERIVE },
+		{ "wsrep_received",                  "operations",   DS_TYPE_DERIVE },
+		{ "wsrep_replicated",                "operations",   DS_TYPE_DERIVE },
+
+		{ "wsrep_received_bytes",            "total_bytes",  DS_TYPE_DERIVE },
+		{ "wsrep_replicated_bytes",          "total_bytes",  DS_TYPE_DERIVE },
+
+		{ "wsrep_apply_window",              "window_size",  DS_TYPE_GAUGE },
```
octo: At your option: Is this in bytes? How about using the `bytes` data set / type?

src/mysql.c
```diff
@@ -685,6 +687,101 @@ static int mysql_read_innodb_stats (mysql_database_t *db, MYSQL *con)
...
+		while ((row = mysql_fetch_row (res)))
+	{
+		int i;
+		char *key;
+		unsigned long long val;
+
+		key = row[0];
+		val = atoll (row[1]);
+
+		for (i = 0; metrics[i].key != NULL && strcmp(metrics[i].key, key) != 0; i++)
```
octo: Please declare the loop variable `i` like this:

```c
for (int i = 0; …; …)
```

octo:
LGTM, good to go once inline commits are resolved / acknowledged. Thanks @xinity!

src/mysql.c
xinity: it's bytes indeed, but i'm pretty fine about using total_bytes easier to use when graphing , just to be sure that's a sum not a number :)

hoping to be clear

xinity:
all comments should be handle correctly (pfiou ! )

src/mysql.c
octo: I think we got our signals crossed. I was referring to `window_size`, a GAUGE.

src/mysql.c
```diff
@@ -685,6 +688,96 @@ static int mysql_read_innodb_stats (mysql_database_t *db, MYSQL *con)
...
+	while ((row = mysql_fetch_row (res)))
+	{
+		int i;
+		char *key;
+		unsigned long long val;
+
+		key = row[0];
+		val = atoll (row[1]);
+
+		for (int i = 0; metrics[i].key != NULL && strcmp(metrics[i].key, key) != 0; i++)
```
octo: D'oh! I didn't see that `i` is used again below, in that case we can't declare it in the loop condition.

xinity: should i revert the loop then ?

rubenk: Please do, because it doesn't compile otherwise :)

xinity: omw :)

strange it does compile on my laptop ...... ???????

xinity:
reverted !

let's my next PR's will be easier to merge :)

src/mysql.c
rubenk: Ah, you did declare i twice, once in the loop and once outside. Depending on your compiler this will give a shadow warning. If you're not building with --disable-werror this should break the build.

rubenk:
Great, now if you could resolve / acknowledge @octo's comment about using bytes instead of window_size, we should be good to go :)

src/mysql.c
xinity: looking at galera documentation:
apply_windows and commit_windows are somehow distances, so i think we can switch from window_size to gauge yes.

XANi:
commit_window is not in bytes:

"Average distance between highest and lowest concurrently committed sequence number. "

same with apply_window:

"Average distance between highest and lowest concurrently applied seqno"

that is why I added another data type, none seemed fitting.

as for `wsrep_local_recv_queue_avg` it is average since last status queries (which will be almost always last query made by collectd), while ones without are "current" values.

So in theory using average would be useful for longer intervals as it would take into consideration any spikes that happen between probes, but if anything else (for example nagios) would do that query it would render it pretty useless, so it is probably too fragile metric to be there, or at least be on by default. I originally added it by going with "if it moves, monitor it" mantra :)

xinity:
@XANi _avg* can be computed by grafana and friend , that's why they have been commented out
commit_window as apply_window can be handled as guauge as wsrep_cert_deps_distance which is also a distance.

are we all good this time ?:)

thanks for your help guys!

XANi:
@xinity i was talking about something else. Imagine you probe every minute:

* 0s - collectd probes
* 10s - you get big request spike and queues grow
* 40s - queue empties
* 60s - collectd probes again

If you use "current" value, you will just get 0, if you use average, you will at least notice that the queue was there.But that is honestly just misdesign of wsrep stats (it should be something like last minute average, not "from last query") so I agree, _avg is not needed.

src/mysql.c
```diff
@@ -685,6 +688,96 @@ static int mysql_read_innodb_stats (mysql_database_t *db, MYSQL *con)
...
+		{ "wsrep_apply_window",              "gauge",        DS_TYPE_GAUGE },
+		{ "wsrep_commit_window",             "gauge",        DS_TYPE_GAUGE },
+
+		{ "wsrep_cluster_size",              "gauge",        DS_TYPE_GAUGE },
+		{ "wsrep_cert_deps_distance",        "gauge",        DS_TYPE_GAUGE },
+
+		{ "wsrep_local_recv_queue",          "queue_length", DS_TYPE_GAUGE },
+		{ "wsrep_local_send_queue",          "queue_length", DS_TYPE_GAUGE },
+
+/*		{ "wsrep_local_recv_queue_avg",      "queue_length", DS_TYPE_GAUGE },
```
rubenk: Could you either add a comment here as to why this code is commented out, or (my preference) remove the code altogether?

Issue #1574, opened 2016-02-28
Build failure with latest libsigrok (0.4)


rubenk:
```
  CC       sigrok_la-sigrok.lo
sigrok.c: In function ‘sigrok_value_type’:
sigrok.c:143:12: error: ‘const struct sr_datafeed_analog’ has no member named ‘mq’
  if (analog->mq == SR_MQ_VOLTAGE)
            ^~
sigrok.c:145:17: error: ‘const struct sr_datafeed_analog’ has no member named ‘mq’
  else if (analog->mq == SR_MQ_CURRENT)
                 ^~
sigrok.c:147:17: error: ‘const struct sr_datafeed_analog’ has no member named ‘mq’
  else if (analog->mq == SR_MQ_FREQUENCY)
                 ^~
sigrok.c:149:17: error: ‘const struct sr_datafeed_analog’ has no member named ‘mq’
  else if (analog->mq == SR_MQ_POWER)
                 ^~
sigrok.c:151:17: error: ‘const struct sr_datafeed_analog’ has no member named ‘mq’
  else if (analog->mq == SR_MQ_TEMPERATURE)
                 ^~
sigrok.c:153:17: error: ‘const struct sr_datafeed_analog’ has no member named ‘mq’
  else if (analog->mq == SR_MQ_RELATIVE_HUMIDITY)
                 ^~
sigrok.c:155:17: error: ‘const struct sr_datafeed_analog’ has no member named ‘mq’
  else if (analog->mq == SR_MQ_SOUND_PRESSURE_LEVEL)
                 ^~
sigrok.c: In function ‘sigrok_feed_callback’:
sigrok.c:184:131: error: dereferencing pointer to incomplete type ‘const struct sr_dev_inst’
   ERROR("sigrok plugin: Received data from driver \"%s\" but "
                                                                                                                                   ^ 
sigrok.c:207:28: error: dereferencing ‘void *’ pointer [-Werror]
  value.gauge = analog->data[0];
                            ^
sigrok.c:207:14: error: void value not ignored as it ought to be
  value.gauge = analog->data[0];
              ^
sigrok.c: In function ‘sigrok_init_driver’:
sigrok.c:270:14: error: dereferencing pointer to incomplete type ‘struct sr_dev_inst’
    cfdev->sdi->vendor ? cfdev->sdi->vendor : "",
              ^~
[...]

PR #1844, opened 2016-08-04
sigrok: use pkg-config to find library

rubenk:
While we're doing that, look for a version less than 0.4, since we don't support that yet (#1574)


octo:
LGTM

PR #1845, opened 2016-08-04
Mysql plugin: remove double reported metrics

rubenk:
This is a rebased version of #1111


octo:
LGTM

octo:
Please check if these patches apply to *collectd-5.5* before merging into *master*.

rubenk:
While technically this is a bugfix, if we apply this to 5.5.x it might surprise some people who don't have InnodbStats set, but expect these metrics to be there.

octo:
Fair enough, *master* it is then.

PR #1846, opened 2016-08-04
python plugin: Fix conversion of value-lists from Python to C.

tokkee:
This affects data-sets with more than one data-source of different types.
Previously, the type of the first data-source would have been used to convert
all values.


tokkee:
Note that this is untested ;-)

trenkel:
Now I'm wondering why I wrote that code twice instead of dumping it into some kind of helper function ...

PR #1847, opened 2016-08-05
collectd-python(5): remove NOTES section

rubenk:
This man page just describes the python plugin so it shouldn't contain
generic information for developers.


rubenk:
@trenkel as the author of the manpage, are you ok with this?

rubenk:
Thanks!

PR #1841, opened 2016-08-04
mysql plugin: updated global status metrics

xinity:
this PR extends global status metrics

it follows PR: https://github.com/collectd/collectd/pull/1797

Issue #1722, opened 2016-05-24
Allow submit any MySQL status variable as Collectd type


rpv-tomsk:
What is about idea adding option to allow submit _any_  MySQL status variable as Collectd type?
(Possible with addition of predefined/preconfigured 'groups').

Proposed new options for mysql plugin:

`StatusVariable VARIABLE TYPE [TYPE_INSTANCE]`
`StatusVariablePrefix PREFIX TYPE`

Configuration examples, how that can be used:

1)
`StatusVariable Ssl_sessions_reused mysql_ssl_sessions_reused`

This will submit MySQL status variable `Ssl_sessions_reused` as Collectd type `mysql_ssl_sessions_reused`.

2)
`StatusVariablePrefix Table_open_cache_ mysql_table_open_cache`

This will take following variables:
```
mysql> show global status like 'Table_open_cache%';
+----------------------------+------------+
| Variable_name              | Value      |
+----------------------------+------------+
| Table_open_cache_hits      | 1818456442 |
| Table_open_cache_misses    | 989758     |
| Table_open_cache_overflows | 934412     |
+----------------------------+------------+
3 rows in set (0.00 sec)
```
and submit them as type `mysql_table_open_cache` with appropriate type instances `hits`, `misses`, `overflows`.

But, for example, `Ssl_session_cache_*` variables can't be submitted in this way (via StatusVariablePrefix), because of different variables (types) will be mixed together:

```
+-----------------------------+-------+
| Variable_name               | Value |
+-----------------------------+-------+
| Ssl_session_cache_hits      | 0     |
| Ssl_session_cache_misses    | 0     |
| Ssl_session_cache_mode      | NONE  |
| Ssl_session_cache_overflows | 0     |
| Ssl_session_cache_size      | 0     |
| Ssl_session_cache_timeouts  | 0     |
+-----------------------------+-------+
```

Them needs to be configured via explicit `StatusVariable` options with type and type instance set, or we can provide special predefined 'group' for this later.

**** PART 2 **** (unfinished idea)

Syntax: 

`StatusGroup PREDEFINED_GROUP`  (for metrics from 'SHOW GLOBAL STATUS')
`InnoDBStats PREDEFINED_GROUP`  (for InnoDB metrics from information_schema.innodb_metrics)

Benefits of `StatusGroup` over `StatusVariablePrefix` are:

1) Predefined configuration. 
1.1) Different metrics/types can be enabled/submitted in one group.
[...]

PR #1849, opened 2016-08-05
mysql plugin: extend global status metrics

xinity:


octo:
This is a follow-up to #1841

xinity:
@octo exactly

src/mysql.c
```diff
@@ -973,6 +977,31 @@ static int mysql_read (user_data_t *ud)
 				counter_submit ("mysql_innodb_rows", "read", val, db);
 			else if (strcmp (key, "Innodb_rows_updated") == 0)
 				counter_submit ("mysql_innodb_rows", "updated", val, db);
+
+ 			/* checkpoints */
+ 			else if (strcmp (key, "Innodb_checkpoint_age") == 0)
+ 				gauge_submit ("mysql_innodb_checkpoint", "age", val, db);
```
octo: The type *mysql_innodb_checkpoint* does not exist. What's the unit of this metric? Seconds?

src/mysql.c
```diff
@@ -973,6 +977,31 @@ static int mysql_read (user_data_t *ud)
...
+ 			else if (strcmp (key, "Innodb_checkpoint_max_age") == 0)
+ 				gauge_submit ("mysql_innodb_checkpoint", "max_age", val, db);
+ 
+ 			/* max id */
+ 			else if (strcmp (key, "Innodb_max_trx_id") == 0)
+ 				gauge_submit ("mysql_innodb_max", "trx_id", val, db);
```
octo: The type *mysql_innodb_max* odes not exist. Assuming that the transaction ID is an integer that is increased by one for each new transaction, I would propose to create a new type like this:

```
total_transactions value:DERIVE:0:U
```

src/mysql.c
```diff
@@ -973,6 +977,31 @@ static int mysql_read (user_data_t *ud)
...
+ 
+ 			/* page_size */
+ 			else if (strcmp (key, "Innodb_page_size") == 0)
+ 				gauge_submit ("mysql_innodb_page", "size", val, db);
```
octo: The type *mysql_innodb_page* does not exist. I don't think page size is a metric we should be collecting – it never changes.

src/mysql.c
```diff
@@ -973,6 +977,31 @@ static int mysql_read (user_data_t *ud)
...
+ 
+ 			/* open tables and files */
+ 			else if (strcmp (key, "Innodb_open_files") == 0)
+ 				gauge_submit ("mysql_innodb_open", "files", val, db);
```
octo: The type *mysql_innodb_open* does not exist. Please use the *files* type. Consider setting the *type_instance* to "open".

src/mysql.c
```diff
@@ -973,6 +977,31 @@ static int mysql_read (user_data_t *ud)
...
+ 			else if (strcmp (key, "Innodb_open_tables") == 0)
+ 				gauge_submit ("mysql_innodb_open", "tables", val, db);
+ 			else if (strcmp (key, "Innodb_opened_files") == 0)
+ 				gauge_submit ("mysql_innodb_opened", "files", val, db);
+ 			else if (strcmp (key, "Innodb_opened_tables") == 0)
+ 				gauge_submit ("mysql_innodb_opened", "tables", val, db);
```
octo: Dito.

src/mysql.c
```diff
@@ -973,6 +977,31 @@ static int mysql_read (user_data_t *ud)
...
+ 
+ 			/* page_size */
+ 			else if (strcmp (key, "Innodb_page_size") == 0)
+ 				gauge_submit ("mysql_innodb_page", "size", val, db);
+ 
+ 			/* open tables and files */
+ 			else if (strcmp (key, "Innodb_open_files") == 0)
+ 				gauge_submit ("mysql_innodb_open", "files", val, db);
+ 			else if (strcmp (key, "Innodb_open_tables") == 0)
+ 				gauge_submit ("mysql_innodb_open", "tables", val, db);
```
octo: The type *mysql_innodb_open* does not exist. I propose to use the *objects* type and set *type_instance* to "tables-open".

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
 		{
 			counter_submit ("mysql_slow_queries", NULL , val, db);
 		}
+ 		else if (strncmp (key, "Queries", strlen ("Queries")) == 0) 
+ 		{
+ 			counter_submit ("mysql_queries", NULL , val, db);
```
octo: The type *mysql_queries* does not exist. I propose to use the *total_operations* type and set the *type_instance* to "query".

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
...
+ 		}
+ 		else if (strncmp (key, "Connections", strlen ("Connections")) == 0) 
+ 		{
+ 			counter_submit ("mysql_connections", NULL , val, db);
```
octo: The type *mysql_connectios* does not exist. Please use the existing type *connections*.

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
...
+ 		}
+ 		else if (strncmp (key, "Uptime", strlen ("Uptime")) == 0) 
+ 		{
+ 			counter_submit ("mysql_uptime", NULL , val, db);
```
octo: Please remove the uptime metric.

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
...
+ 		}
+ 		else if (strncmp (key, "Questions", strlen ("Questions")) == 0) 
+ 		{
+ 			counter_submit ("mysql_questions", NULL , val, db);
```
octo: The type *mysql_queries* does not exist. I propose to use the *total_operations* type and set the *type_instance* to "question".

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
...
+ 		}
+ 		else if (strncmp (key, "Created_", strlen ("Created_")) == 0)
```
octo: Please use the *total_objects* for these and set the *type_instance* to something like "files-created".

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
...
+                 {
```
octo: Indentation of these curlies is wrong.

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
...
+                                counter_submit ("mysql_created", "tmp_tables", val, db);
+                        else if (strcmp (key, "Created_tmp_disk_tables") == 0)
+                                counter_submit ("mysql_created", "tmp_disk_tables", val, db);
+                        else if (strcmp (key, "Created_tmp_files") == 0)
+                                counter_submit ("mysql_created", "tmp_files", val, db);
+                 }
+ 		else if (strncmp (key, "Aborted_", strlen ("Aborted_")) == 0)
+ 		{
+ 			if (strcmp (key, "Aborted_clients") == 0)
+ 				counter_submit ("mysql_aborted", "clients", val, db);
```
octo: Please use *total_connections* with type instance "aborted".

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
...
+ 			else if (strcmp (key, "Aborted_connects") == 0)
+ 				counter_submit ("mysql_aborted", "connects", val, db);
```
octo: Please use *total_connections* with type instance "aborted-connects".

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
...
+ 		}
+ 		else if (strncmp (key, "Key_", strlen ("Key_")) == 0)
```
octo: Please use the type *total_operations* for these.

src/mysql.c
```diff
@@ -1014,6 +1091,8 @@ static int mysql_read (user_data_t *ud)
 
 		gauge_submit ("cache_size", "qcache",
 				qcache_queries_in_cache, db);
+ 		gauge_submit ("cache_queries", "qcache-free-memory",
```
octo: Please use the *memory* type for this and set the type instance to "free".

octo:
Hi @xinity,

thank you very much for your patch! The exact types and names used when reporting metrics are part of collectd's backwards compatibility, so getting them right from the start is very important to us. If you have questions how the *types* work, please don't hesitate to ask, I'm happy to help!

Best regards,
—octo

src/mysql.c
xinity: why that ?

xinity:
@octo i'll take care of it ASAP , thanks a lot for the review

xinity:
my patch is missing a file (really messing with github :( ) 
let me update it all, should be good by the end of the day

octo:
Thanks @xinity :) If you need a hand with git, feel free to jump into the `#collectd` channel on freenode and poke me there.

xinity:
types.db has been patched @octo let me know if i need to fix more things

src/mysql.c
rpv-tomsk: What is the sense of this metric? When I want to check uptime I refer to nagios, there I can found all required history of status changes.  As for me, I see nothing useful in triangles on similar charts.

src/mysql.c
rpv-tomsk: Why you mix 
"mysql_created", "tmp_tables"
and 
"mysql_created", "tmp_files" ?

The first is tables, the second is files. This should not be mixed in one type 'mysql_created'.

src/mysql.c
rpv-tomsk: @octo, I think total_conections should be reviewed in complex.

We put 'total_connections -> aborted-connects" here and "total_connections -> aborted-clients" upper. Full list of  type instances of this tupe should be compiled and checked.

And mix of 'connects' and 'clients' does not look nice.

src/mysql.c
rpv-tomsk: @octo, What will happen when we, for example, will collect 'Innodb_open_files' (fictional variable) ? 
By this logic, that should be 'objects' type and 'files-open' type_instance. But what will be wrong.

I propose to use something like this: 'mysql_innodb_tables' type (if it does not exists, please check!) and 'open' type instance. Then there no conflict appear in future. 

What do you think?

rubenk:
thanks @xinity. Yes, please address @octo's other comments.

src/mysql.c
```diff
@@ -973,6 +977,31 @@ static int mysql_read (user_data_t *ud)
...
+ 			else if (strcmp (key, "Innodb_page_size") == 0)
+ 				gauge_submit ("mysql_innodb_page", "size", val, db);
+ 
+ 			/* open tables and files */
+ 			else if (strcmp (key, "Innodb_open_files") == 0)
+ 				gauge_submit ("mysql_innodb_open", "files", val, db);
+ 			else if (strcmp (key, "Innodb_open_tables") == 0)
+ 				gauge_submit ("mysql_innodb_open", "tables", val, db);
+ 			else if (strcmp (key, "Innodb_opened_files") == 0)
+ 				gauge_submit ("mysql_innodb_opened", "files", val, db);
```
rpv-tomsk: There is two kind of metrics in MySQL:
- 'open': number of currently opened files, current state.
- 'opened': number of 'open' operations, i.e. 'rate'.

src/mysql.c
rpv-tomsk: I don't think what strong recommendation of using existing types should apply.
What other type instances could be added to this type? 
That type instance _MUST_ be related to cache, so we will unable to use this _type_ to other things. 
The 'memory' type is too general for this. 

In my view, this metric may be sent as 'cache_size', but this already used for 'queries number', so new type should appear. In this case, I recommend to send this as 'cache_memory' with type instance 'free'. 

So, when new metric will appear in 'show global status' output we will add it with no problems.

octo: The *memory* type is for bytes held in memory, which is *exactly* what this metric is afaIk. If you're afraid that this might clash with memory metrics that are not related to the query cache, let's use the *type instance* "qcache-free".

Adding a new type "cache_memory" is not justified.

src/mysql.c
```diff
@@ -995,6 +1024,54 @@ static int mysql_read (user_data_t *ud)
...
+ 				counter_submit ("mysql_key", "read_requests", val, db);
+ 			if(strcmp (key, "Key_writes") == 0)
+ 				counter_submit ("mysql_key", "writes", val, db);
+ 			else if (strcmp (key, "Key_write_requests") == 0)
+ 				counter_submit ("mysql_key", "write_requests", val, db);
+ 			if(strcmp (key, "Key_blocks_unused") == 0)
+ 				counter_submit ("mysql_key", "blocks_unused", val, db);
+ 			else if (strcmp (key, "Key_blocks_not_flushed") == 0)
+ 				counter_submit ("mysql_key", "not_flushed", val, db);								
+ 		}
```
rpv-tomsk: At the first, this should not be one 'mysql_key' type. 

The 

Key_reads
Key_read_requests
Key_writes
Key_write_requests

and

Key_blocks_unused
Key_blocks_not_flushed

are strongly different.

First group is about block IO, the second is about MyISAM key cache state. 
I can't understand why do you mix this in one type. Did you tried to use your patch at practice?

octo: +1

The *total_requests* and *total_operations* types come to mind for the first set of metrics.

Are the "Key_blocks_*" metrics really counters? It might make sense to add a *total_blocks* type for them. Alternatively, if the block size can be determined, we can convert and dispatch them as *total_bytes*.

rpv-tomsk:
Hi guys.

Patches like this show us one general problem with 'mysql' plugin: 
different people have different requirements about metrics to be collected.

With last patches of mysql.c, plugin start to send more and more metrics, and them not so much useful in generic setup. But plugin does not flexible enough to allow unneded metrics disable. 

For example, this PR proposes to add 4 metrics, based on following 'SHOW GLOBAL STATUS' variables:

`Key_reads
Key_read_requests
Key_writes
Key_write_requests`

As for me, it looks like what one 'cache miss rate'  metric may supersede these four metrics.
By the MySQL docs, it can be calculated as Key_reads/Key_read_requests. 

Currently I don't know, if 'Key_*' metrics will be useful for me as for sysadm. I will be unable to tune the system based on them directly. But if I will see 'cache miss rate', I will have the basis for 'key_buffer_size'  change and directly see response on my actions.

My proposal is: 
- split existing metrics by groups and allow one to tune plugin 'verbosity' by yourself.
- allow to create new metrics by plugin configuration, not only by code patching.
- Stop accepting patches to 'mysql.c' without chart examples and without description of their usefulness for people. Such description also may be included in docs and wiki which also be very good. This also will reject solutions, which was not worked through enough.

I have respect of xinity work, it is great what people tries to make project better. 
But review must be stronger.

src/mysql.c
rpv-tomsk: I'm using http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Key_blocks_unused for variables explanations.

src/mysql.c
rpv-tomsk: Let's look  into another example:

`Innodb_buffer_pool_bytes_dirty / Innodb_buffer_pool_bytes_data`

http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Innodb_buffer_pool_bytes_data 

By such logic, whem also should be 'memory' type and two another type instances.

But I see no sence in mixing all (three) of them in one heap.

rpv-tomsk:
Please look into #431, #1694 and #1722. 

@xinity, what do you think about #1722?

xinity:
@rpv-tomsk my work starts from what i can achive with prometheus metrics and what i can't with collectd, my goal is to keep collectd and extended it's metrics to allow it to keep up with growing needs in monitoring metrics (on this point the mysql plugin needs i think a bit more of a work i agree).
metrics i proposed should be easier a sysadmin or a dba/dev to monitor/debug it's mysql system :)

i had a patch for mysql variables as well but @rubenk seems to be not fan of it :)

globally yes i use my patches on my dev laptop (working on integrating Percona Monitoring Management dashboard to be used with monitoring system collectd based).

i'm far away of a great contributor :) but i'm open to any review , rework needed

i did my patch the way i though it should be, yet i'm not that sure:) i'll be happy to grow understanding more about collectd and the way patches should be contributed :)

let me know what should be done and i'll apply As soon as i can :)

thank you guys for your time, review and comment

xinity:
@rpv-tomsk @octo @rubenk , i'm back from 2 weeks away, how should we handle things regarding mysql plugin evolution then ? :)

octo:
Hello @xinity, welcome back :) I'd propose to get the MySQL metric key → collectd type and type instance mapping right first. There are numerous comments above asking for changes, addressing those should be the first step. A [Google Docs Spreadsheet](https://docs.google.com/spreadsheets/) might be best suited for this, but Github's Markdown tables might do the trick, too.

When that is settled and implemented, we can move on to code review comments. From what I've seen so far, that part should be fairly straight forward. Once comments are resolved, we can merge.

Best regards,
—octo

octo:
@rpv-tomsk I agree, the *MySQL plugins* has many metrics and not all of them are useful for every user. Grouping them and allowing users to enable / disable groups as needed is a very good suggestion; though slightly off-topic here. Could you open a feature-request issue for that? Thanks!

src/mysql.c
octo: Fair enough, but please use *tables* instead – "mysql_innodb_tables" is too specific.

src/mysql.c
octo: If that is so, this needs to be a `derive_t` instead.

xinity:
@octo i'll be working on this :)
@rpv-tomsk i'll be glad to help if i can

octo:
@xinity Awesome, thank you! :)

rpv-tomsk:
> Could you open a feature-request issue for that?

There is #1722 for this. We need to develop configuration syntax/options, do that in backward-compatible manner. As for me, that is harder part of work whan coding.

bogdanr:
+1

PR #1850, opened 2016-08-05
python plugin: return error on config failure

### LINKED ISSUE #926: Test config should return 1 if python plugin has errors
Hi,

When I test collectd's configuration, I would like that collectd returns 1 even if an error occurs inside a plugin.

```
# /opt/collectd/sbin/collectd -t -C /opt/collectd/etc/collectd.conf
python plugin: Error importing module "xxxxxx".
Unhandled python exception in importing module: IndentationError: expected an indented block (xxxxxx.py, line 117)
python plugin: Found a configuration for the "xxxxxx" plugin, but the plugin isn't loaded or didn't register a configuration callback.
python plugin: Found a configuration for the "xxxxxx" plugin, but the plugin isn't loaded or didn't register a configuration callback.
# echo $?
0
```

Centos 6.5
Collectd 5.4, commit ecbcaadde7dff838f09dfb04c7eb90f126e35abf

Thanks.

rubenk:
Return an error when something goes wrong during config.
We still loop through all specified modules so all errors are logged.

Fixes #926

### DIFF
--- a/src/python.c
+++ b/src/python.c
@@ -219,7 +219,7 @@ static char reg_shutdown_doc[] = "register_shutdown(callback[, data][, name]) ->
 		"    data if it was supplied.";
 
 
-static int do_interactive = 0;
+static _Bool do_interactive = 0;
 
 /* This is our global thread state. Python saves some stuff in thread-local
  * storage. So if we allow the interpreter to run in the background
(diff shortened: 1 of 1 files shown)


rubenk:
@trenkel I'd love to know what you think.

trenkel:
Github won't let me add comments to lines you didn't change, so I'll just dump it all here.
I think there are quite a few places where an error status should be set:
1122: Use cf_util_get_boolean and return an error on failure.
1126: Use cf_util_get_string and return an error on failure.
1129: Perhaps error-worthy, too?
1133: Return an error on failure.
1136: Use cf_util_get_boolean and return an error on failure.
1162: Return an error on failure.
1184: Return an error on failure.
1199: Return an error on failure.
1226: Perhaps error-worthy, too?

rubenk:
Thanks for your comments :) I'll add checks for the other failures too.

rubenk:
@trenkel is e5b10d0 what you had in mind?

trenkel:
Great, thanks.

Issue #27, opened 2012-02-23
python plugin fails on Mac OS X 10.7.2


blysik:
From running configure:

checking for Python CPPFLAGS... /System/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7
checking Python.h usability... yes
checking Python.h presence... yes
checking for Python.h... yes
checking for Python LDFLAGS... /System/Library/Frameworks/Python.framework/Versions/2.7/lib
checking for Python LIBS... failed with status 0 (output: )

rubenk:
Fixed by f0735d24a2195c8da2657c46a7cb9ceb790b0836

PR #1851, opened 2016-08-05
Switch to python-config for libpython detection

rubenk:
The old way of detecting libpython was unreliable and did not work on
OS X.

Instead use python-config which was introduced in python 2.6.

For older versions of python you can still set PYTHON_CPPFLAGS and
PYTHON_LDFLAGS on the ./configure command line.

By default we look for python3-config, python2-config and last
python-config. The path to python-config can be overriden by setting
PYTHON_CONFIG.

Fixes #888 and #27


rubenk:
@trenkel what do you think?

CPS86:
Hello

I'm trying to compile Collectd 5.8.0 with a very old version of python (2.4 or 2.5). Unfortunately without success.

I'm receiving the following: 
```
checking for python3-config  ../python-config
checking Python.h usability ... yes
checking Python.h presence ... yes
checking for Python.h ... yes
checking for libpython ... no 
```

What is libpython? 
I thought that all necessary information is retrieved from python-config? 
Can I somehow provide manually libpython location? Is it e.g. libpython2.4.so file ?

PR #1856, opened 2016-08-06
snmp : don't send values for empty type_instance

### LINKED ISSUE #176: using IF-MIB::ifAlias as Instance
We're using collectd 5.1.0 with the snmp plugin to monitor various Juniper equipment.

Since interface names like ge-0/0/0 don't say as much as say 'trunk to dc2', we use the ifAlias instead of the ifDescr as Instance to index into our tables.

Only the ports we use have an ifDescr. The issue we have is that for all ports that return an empty string, collectd tries to write to the same rrd files (say if_errors.rrd instead of if_errors-trunk to dc2.rrd). This results in flooding of our logs with the following errors:

rrdc_update (/collectd/rrd/vc3/snmp/if_errors.rrd, [1352632188:0:0], 1) failed with status -1

Let me know if you need more information.

rubenk:
If you have multiple Instance values that are empty string,
they all end up in the same rrd file, resulting in:
rrdc_update (/collectd/rrd/vc3/snmp/if_errors.rrd, [1352632188:0:0], 1) failed with status -1

Fixes #176

### DIFF
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -1360,8 +1360,12 @@ static int csnmp_dispatch_table (host_definition_t *host, data_definition_t *dat
     for (i = 0; i < data->values_len; i++)
       vl.values[i] = value_table_ptr[i]->value;
 
-    /* If we get here `vl.type_instance' and all `vl.values' have been set */
-    plugin_dispatch_values (&vl);
+    /* If we get here `vl.type_instance' and all `vl.values' have been set
+     * vl.type_instance can be empty, i.e. a blank port description on a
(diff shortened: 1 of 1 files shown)


octo:
Can you maybe add a short comment explaining why the `type_instance` may be empty?

Apart from that, LGTM.

rubenk:
Sure, I'll do that, thanks.

While this fixes this particular issue, I noticed a few similar bug reports for other plugins. Do you think we could add something to collectd that checks for duplicate type instances before we send the values out?

PR #1857, opened 2016-08-06
Utils dns add new rrtypes

rubenk:


src/utils_dns.c
```diff
@@ -749,22 +749,41 @@ const char *qtype_str(int t)
...
+	    case ns_t_nsec3param: return ("NSEC3PARAM");
+	    case ns_t_hip:      return ("HIP");
+	    case ns_t_spf:      return ("SPF");
 	    case ns_t_ixfr:     return ("IXFR");
+# endif
 	    case ns_t_axfr:     return ("AXFR");
 	    case ns_t_mailb:    return ("MAILB");
 	    case ns_t_maila:    return ("MAILA");
 	    case ns_t_any:      return ("ANY");
+# if __NAMESER >= 19991006
```
octo: Can we unify this block with the 19991006 block above?

src/utils_dns.c
```diff
@@ -749,22 +749,41 @@ const char *qtype_str(int t)
...
 	    case ns_t_zxfr:     return ("ZXFR");
-/* #endif __NAMESER >= 19991006 */
+# endif
+# if __NAMESER >= 20090302
```
octo: Dito?

src/utils_dns.c
rubenk: I thought about that, but I kept the same ordering as nameser.h (ordered by type id). This will make it easier to compare on feature updates.

octo: Fair enough, works for me.

PR #1861, opened 2016-08-06
RFC: Lua plugin

rubenk:
This is a rebase of @schmurfy and @octo's work on a Lua plugin back in 2010.

Having support for Lua in Collectd would be great. I tested the write and read callbacks and they work. I haven't checked the filter support yet. One thing I noticed is that the BaseDir is not appended to Lua's package.path so you'll need to do a little trick currently to import the collectd.lua module. There's a Gist at https://gist.github.com/rubenk/4946cf369f33411b54c12593010e6238 which does that.
You'll also need Lua 5.1 currently.


octo:
[I have no memory of this patch](https://www.youtube.com/watch?v=bSrf8EgSyDg)

Thanks for reviving this :) It's a bit too much code to review tonight, but in general I'm happy to have this.

octo:
@schmurfy would you be willing to relicense the code unter the MIT license? That's collectd's standard license for new code …

schmurfy:
Even I forgot I worked on this, glad to see it revived :)
No problem for the license, MIT is my favorite.

rubenk:
Superseded by #1885

Issue #1169, opened 2015-07-28
proposal: make the /proc location configurable


iverberk:
Problem statement
==============

Most Linux systems expose system configuration and statistics through a virtual filesystem that is usually mounted on /proc. When Linux is run within a container environment it gets its own version of /proc which maps specifically to the container environment and not to the host environment. This effectively renders most of the standard collectd metrics useless (unless you want to monitor the container metrics!).

It is possible to bind-mount the host /proc into the container but not at the default location of /proc since this is now occupied by the new container-specific /proc. One *can* mount the host /proc to a different folder in the container. This would allow the unmount of the container /proc and mount of the host /proc at the correct location. Unfortunately, in order to do this the container would need a lot of elevated privileges thereby sort of defeating the purpose of using a container in the first place.

Solution
======
Obviously the solution would be to make the /proc location, where collectd looks for the metrics, configurable trough the configuration file or command-line parameter. The default location will remain /proc as this covers most use-cases and ensures backwards compatibility but for container specific usage this would make it possible to run completely unprivileged (also as non-root user) collectd containers that can be run anywhere.

Benefits
======
I think it would be a great addition to collectd if this feature were to be implemented. It would mean easy deployment to any host that is capable (or only capable, like CoreOS!) of running containers and I am sure it would mean a positive uptake on collectd usage as the metrics daemon of choice for container environments.

I am wondering if the collectd community would be amenable to this change? If there are no major concerns I will try to make this change and send a pull-request.

sherzberg:
Are there any workarounds for this? Or any thoughts on how to get this to work? I would also like to run collectd inside a container to get the hosts metrics.

sherzberg:
I found this and started using it: https://github.com/fr3nd/docker-collectd. Works really well.

PR #1867, opened 2016-08-08
Making /proc and /sys path configurable

mfournier:
This is a work in progress I wanted to submit for review before finishing it up. The use-case is detailed in #1169.

Basically, set `PseudoFSPrefix "/foo"` in the configuration file, and the plugins will read from `/foo/proc/...` and `/foo/sys/...`.

As this touches a ton of plugins, I tried to make the patch as un-invasive as possible. My main concern now is that the string concatenation gets executed at each iteration, rather than once at plugin initialization time. But I'm not sure how to do this better without global variables or altering a lot of function prototypes. This also adds a lot of mostly duplicate code, but I'm not sure how to isolate that in a function (still not very confortable with passing strings around in C...)

There are a few more plugins that need to be converted, and the doc bits are missing, so this is definetely not ready to be merged.

### DIFF
--- a/src/contextswitch.c
+++ b/src/contextswitch.c
@@ -91,9 +91,15 @@ static int cs_read (void)
 	derive_t result = 0;
 	int status = -2;
 
-	fh = fopen ("/proc/stat", "r");
+	const char *prefix = global_option_get("PseudoFSPrefix");
+	const char *path   = "/proc/stat";
+	char statfile[strlen(prefix) + strlen(path) + 1];
+
+	ssnprintf(statfile, sizeof(statfile), "%s%s", prefix, path);
+	fh = fopen (statfile, "r");
 	if (fh == NULL) {
-		ERROR ("contextswitch plugin: unable to open /proc/stat: %s",
+		ERROR ("contextswitch plugin: unable to open %s: %s",
+				statfile,
 				sstrerror (errno, buffer, sizeof (buffer)));
 		return (-1);
 	}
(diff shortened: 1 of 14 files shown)


mfournier:
```
./.libs/libmount.a(utils_mount.o): In function `uuidcache_init':
/jenkins/workspace/pull-requests-build-on-jessie-amd64-with-default-toolchain/collectd-5.5.2.1209.g80294d9/src/utils_mount.c:221: undefined reference to `global_option_get'
./.libs/libmount.a(utils_mount.o): In function `cu_mount_getmntent':
/jenkins/workspace/pull-requests-build-on-jessie-amd64-with-default-toolchain/collectd-5.5.2.1209.g80294d9/src/utils_mount.c:585: undefined reference to `global_option_get'
collect2: error: ld returned 1 exit status
Makefile:3918: recipe for target 'test_utils_mount' failed
make[4]: *** [test_utils_mount] Error 1
```

So `make check` is failing because of a missing `#include "configfile.h"` statement. Not really sure where to add it though.

rubenk:
Hmm, plugin.h includes it, as well as a lot of individual plugins (which should probably be cleaned up).

I'd say stick it at the top of utils_mount.c but then it's not a nice self-contained library anymore..

mfournier:
Yes, I tried this already, but curiously it didn't help.

rubenk:
My guess is that the daemon isn't linked into the test program that tests utils_mount.

mfournier:
Dropping this in favor of in #2041.

This approach is a bit messy and is bound to be a maintenance burden.

PR #1868, opened 2016-08-08
varnish: drop support for version 2.x

mfournier:
Varnish 5 is due in a bit more than one month from now. Before adding
even more code & complexity to this plugin, remove the duplicate code
created in d010d9e and b208ddc.

For the record, the last commit to the 2.1 branch in varnish's git repo
was a bit more than 3 years ago, and the last release (2.1.5) was on
Jan 27, 2011.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5164,9 +5164,7 @@ then
 		[AC_DEFINE([HAVE_VARNISH_V4], [1], [Varnish 4 API support])],
 		[AC_CHECK_HEADERS(vsc.h,
 			[AC_DEFINE([HAVE_VARNISH_V3], [1], [Varnish 3 API support])],
-			[AC_CHECK_HEADERS(varnishapi.h,
-				[AC_DEFINE([HAVE_VARNISH_V2], [1], [Varnish 2 API support])],
-				[with_libvarnish="no (found none of the varnish header files)"])])])
+			[with_libvarnish="no (found none of the varnish header files)"])])
 
 	CPPFLAGS="$SAVE_CPPFLAGS"
 fi
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1340,12 +1340,9 @@
 #      CollectFetch false
 #      CollectHCB false
 #      CollectObjects false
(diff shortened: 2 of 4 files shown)


rubenk:
At least RHEL5 and RHEL6 still have Varnish 2 (https://apps.fedoraproject.org/packages/varnish)
That doesn't mean we as upstream have to keep supporting it, but is something to consider.

Will Varnish 5 be ready before we cut 5.6? Otherwise I'd say make 5.6 the last release that supports Varnish 2, that will give users a bit more time. But I leave it up to you if you find it worth the effort.

mfournier:
Yes, agreed. We can leave 2.x support for longer. I was assuming the 3.x & 4.x part of the code would need to evolve (ie: diverge even more from the 2.x part) to add 5.x support. I also wanted to add support for non-MAIN metrics, which would not be supported on 2.x.

Varnish 5.0 should be released on September 15. I'll take a moment to evaluate what changes are required for 5.x support, based on varnish's current master. And take a decision based on that.

rubenk:
Sounds good!

octo:
> Otherwise I'd say make 5.6 the last release that supports Varnish 2
+1

mfournier:
Good news: current varnish git HEAD doesn't break the stats API. So unless they change something before the release (~1 month from now), our plugin seems to be varnish-5 ready.

So let's leave the 2.x support in for the moment, and reconsider post 5.6.

rubenk:
Hey @mfournier, do we still need to reconsider this for 5.7? I'd say just leave it for now, since nothing is broken and there's no harm keeping support for just a little bit longer.

octo:
+1

mfournier:
Yes, we can leave the support for older versions for a bit longer, as far as I'm concerned.

The reason I prepared this PR is that adding support for 4.1 led to a lot of code duplication, which dropping 2.1 support would remove.

Also, I was worried varnish 5.x would make things even worse, but this is finally not the case: the API stayed the same as the 4.x line, regarding statistics collection.

So I'll close this PR for the moment, as it needs rebasing anyways.

PR #1869, opened 2016-08-08
Add sample-based rates

BrandonArp:
This commit adds functionality for computing "rates," but without normalizing to per-second.  This allows downstream aggregators to consume more "raw" data from collectd, but still have collectd interpret the values as gauges, counters, etc.

For example, let's say that I want to graph the number of messages sent on a message queue.  This is a counter value.  Collectd knows how to interpret values, but requires the StoreRates flag to be set to do the interpretation.  However, if StoreRates is set, all values are nomalized to per-second values.  This leads to a bit of a dilemma for downstream aggregators: to get the number of bytes sent during the sample period, you would need to turn off StoreRates.  That would require the aggregator to then know about the different collectd types and infer the difference appropriately.  On the other hand, if StoreRates = true, then the aggregator will receive the per-second rates.  This makes it difficult to compute the actual number of messages consumed.


octo:
Hi @BrandonArp,

thank you very much for your patch! I'm afraid this is not a good fit for *collectd* though. Graphing front-ends and other consumers of the metrics use the *type* to determine the unit of metrics. This patch changes the values in most write plugins without changing the type / unit of the metrics, which is very confusing.

Best regards,
—octo

BrandonArp:
Hi @octo,

I agree with the confusing part.  However, I'm not really sure what a good alternative is.  Let me explain my situation in a bit more detail.  

Take, for example, If I use StoreRates = false,  I get "raw", always increasing values for if_octets rx and tx.  As expected, the dstypes are derive.  However, if I use StoreRates = true, I get the "processed" per-second measurements (as expected), but the dstypes remain as derive.  I don't see a way to know if I'm looking at "raw" or "processed" measurements.

Also, I'm using the write_http plugin and am shipping data to another machine.  The destination machine does not necessarily have the types database.  This makes it difficult to perform the derive/counter calculations because I won't necessarily know the minimums and maximums without the types db.

Overall, I'd like a way to consume the collectd data without needing the same types.db as the producers, and without duplicating the counter/derive logic that collectd already has.  What would be a better way to approach this?  Is there a better place for this discussion?

Thanks,
Brandon

octo:
Hi @BrandonArp, thanks for following up! We can have this discussion here for the moment, eventually it might make sense to move to an RFC issue or the mailing list …

You're touching on three different issues:

1.  The JSON data produced by the *Write HTTP plugin* (and others) does not specify whether a number is a `DERIVE` or a `GAUGE`, i.e. a counter or a floating point number.
1.  The JSON data produced by the *Write HTTP plugin* (and others) does not specify whether a `DERIVE` number is the raw integer or a floating point rate calculated from a `DERIVE`.
1.  Shared state in general and the `types.db` in particular sucks.

At some point, we added the *StoreRates* option to address 1) and a bit of 3): when enabled, you get a float, don't have to calculate a derivative and done. We did not think much about 2), because the entire point was to put the consumer into this "don't care" situation.

Our longer-term ideas to fix 3) are going in the direction of pushing the *StoreRates* approach further: convert counters to a rate as soon as possible, probably as part of the `plugin_dispatch_values()` function, i.e. before they are passed to the write callbacks. If everything is a float, the primary information in the `types.db` no longer needs to be configured everywhere.

I have a feeling that these thoughts about the overall architecture of the daemon's infrastructure are a bit too theoretical for the actual problem you're trying to solve, though. Am I right in assuming that a metric such as "0.0238 orders per second" are a bit too abstract for you to put on a management-compatible dashboard and you want to display information such as "4855 orders received today"?

BrandonArp:
Hi @octo.  Sorry it has taken so very long for me to get back to this.  You are correct in assuming that the derived rate is a bit too abstract to put on a dashboard directly.  But problem extends further and has repercussions in the recombinability of the data in general. 

In the system deployed at work, we take samples frequently (1 per sec or 1 per 3sec) to get a real-time view of system behavior.  Due to the data rate, we don't persist these values directly, but rather have an intermediate aggregation step which computes aggregates over a 1 minute interval.  In order to store things consistently in the back-end, we try to avoid using rates until query/view time.  For example, you can't usually meaningfully sum rates. And having 2 machines with a differing number of samples in a time period makes averaged rates not directly comparable. We would need to compute totals and divide by the overall time to get an overall rate.  Therefore, we opt to use "raw diff" values directly. IE, how many bytes were transferred on the wire since the last sample, how many orders were created since the last sample, etc.  I'm happy to go into more depth on the use case either here or in another forum, but suffice it to say, having "raw diff" values is immensely useful for us.

That said, this PR was intended as a way to get that "raw diff" value from collectd.  I'm not tied to this implementation at all.

Should we move this discussion to the mailing list?

PR #1874, opened 2016-08-09
utils_dns.c: fix compiler warning

rubenk:
utils_dns.c: In function ‘handle_ip’:
utils_dns.c:531:29: warning: pointer of type ‘void *’ used in arithmetic
[-Wpointer-arith]
     memcpy(buf, (void *) ip + offset, len - offset);
                             ^


rubenk:
I could use a second pair of eyes, since all these casts and braces make my head hurt ;)

Issue #1819, opened 2016-07-29
grpc plugin: API changes (discussion)


octo:
I've been playing with the *grpc plugin* and started work on a Go implementation. I stumbled upon some things in the API that I would like to discuss before the interface is included in a release (and therefore set in stone).

*   **Split Services:** Currently all RPC methods live in a single "Collectd" service. I think it would be perfectly reasonable for a server to only implement the *DispatchValues* call, e.g. the stateless Prometheus proxy. To achieve this, we would need to either split this into multiple services or implementations need to return a "not implemented" error. Personally, I think the former is the cleaner solution.
*   **Add data source names:** The JSON representation of value lists, the Go implementation and potentially some more include the name of "data sources" in the payload. This removes (much of) the implied shared state of the *types.db(5)*, which IMHO is a good thing. I'd like to add this to the *ValueList* type.
*   **Use streaming RPCs:** I think that the *DispatchValues* call would benefit from being turned into a "client side" streaming call, i.e. let clients submit many value lists in a single call. Eventually, a "send me all values until I say 'stop'" call would be nice to have.
*   **Unify packages:** The protocol buffers are currently broken up into two packages, *collectd* and *collectd.types*. This makes it a wee bit cumbersome to use in Go, since you have to import two generated packages which need to live in different directories.
*   **Rename fields:** I'd like to rename `repeated Value value` in the *ValueList* message to `values`, `ValueList values` in the *DispatchValuesRequest* message to `value_list` and `repeated ValueList values` in the *QueryValuesReply* message to `value_lists`.

octo:
FYI, I have some code tackling some of these issues in [my issue/1819 branch](https://github.com/octo/collectd/commits/issue/1819).

In the mean time, I had some additional thoughts:

*   **Split Services:** It appears that the C++ API fully expects gRPC servers to only implement a subset of the services. Maybe this is overkill?
*   **Use streaming RPCs:** When writing value lists to a server, we don't normally ever stop. That means that with a "client side" streaming RPC we never get a response from the server. Maybe we should actually not do streaming here, since we don't have any intention of ever closing the stream, and do normal RPCs instead, so that we get an ACK for each value list? Alternatively, we could use bidirectional streaming and (asynchronously) report back any errors for the client's consideration?

tokkee:
Here are my thoughts:

[...]

PR #1876, opened 2016-08-10
grpc plugin: Refactor plugin.

octo:
This PR implements (most of) the changes discussed in #1819, please see that issue for details. Also, comments regarding the *design* should go in that issue while comments regarding the *implementation* should go here.

### DIFF
--- a/proto/collectd.proto
+++ b/proto/collectd.proto
@@ -32,9 +32,7 @@ import "types.proto";
 service Collectd {
 	// Query a list of values available from collectd's value cache.
 	rpc QueryValues(QueryValuesRequest) returns (stream QueryValuesResponse);
-}
 
-service Dispatch {
   // DispatchValues sends a stream of ValueLists to the server.
   rpc DispatchValues(stream DispatchValuesRequest) returns (DispatchValuesResponse);
 }
--- a/proto/collectd.proto
+++ b/proto/collectd.proto
@@ -29,11 +29,13 @@ package collectd;
 import "types.proto";
 
 service Collectd {
-	// Dispatch collected values to collectd.
-	rpc DispatchValues(DispatchValuesRequest) returns (DispatchValuesReply);
-
 	// Query a list of values available from collectd's value cache.
 	rpc QueryValues(QueryValuesRequest) returns (QueryValuesReply);
+}
+
+service Dispatch {
+  // DispatchValues sends a stream of ValueLists to the server.
+  rpc DispatchValues(stream DispatchValuesRequest) returns (DispatchValuesReply);
 }
 
 // The arguments to DispatchValues.
--- a/proto/collectd.proto
+++ b/proto/collectd.proto
@@ -30,31 +30,27 @@ option go_package = "collectd.org/rpc/proto";
 import "types.proto";
 
 service Collectd {
-	// Query a list of values available from collectd's value cache.
-	rpc QueryValues(QueryValuesRequest) returns (stream QueryValuesResponse);
+  // Query a list of values available from collectd's value cache.
+  rpc QueryValues(QueryValuesRequest) returns (stream QueryValuesResponse);
 
   // DispatchValues sends a stream of ValueLists to the server.
-  rpc DispatchValues(stream DispatchValuesRequest) returns (DispatchValuesResponse);
+  rpc DispatchValues(stream DispatchValuesRequest)
+      returns (DispatchValuesResponse);
 }
 
 // The arguments to DispatchValues.
-message DispatchValuesRequest {
-	collectd.types.ValueList value_list = 1;
-}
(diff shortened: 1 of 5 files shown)


src/grpc.cc
```diff
@@ -254,168 +258,124 @@ static grpc::Status unmarshal_value_list(const collectd::types::ValueList &msg,
...
-		value_list_t res;
-		if (parse_identifier_vl(name, &res) != 0) {
-			status = grpc::Status(grpc::StatusCode::INTERNAL,
-					grpc::string("failed to parse identifier"));
-			break;
+class CollectdImpl : public collectd::Collectd::Service {
+public:
+	grpc::Status QueryValues(grpc::ServerContext *ctx, QueryValuesRequest const *req, grpc::ServerWriter<QueryValuesResponse> *writer) override {
+		value_list_t match;
+		auto err = unmarshal_ident(req->identifier(), &match, false);
```
tokkee: Nit: `status` and `err` are now used inconsistently in this file. `status` makes more sense to me b/c the value is not an error indicator (unlike, e.g., error variables in Go which are non-nil on error).

src/grpc.cc
```diff
@@ -254,168 +258,124 @@ static grpc::Status unmarshal_value_list(const collectd::types::ValueList &msg,
...
-			continue;
-
-		if (uc_iterator_get_time(iter, &res.time) < 0) {
-			status = grpc::Status(grpc::StatusCode::INTERNAL,
-					grpc::string("failed to retrieve value timestamp"));
-			break;
+		std::queue<value_list_t> value_lists;
+		err = this->queryValuesRead(&match, &value_lists);
+		if (err.ok()) {
+			err = this->queryValuesWrite(ctx, writer, &value_lists);
```
tokkee: It would be nice if queryValuesRead could stream back to the client without caching all values first. However, with the current implementation, this would mean network I/O while holding the cache lock which would be pretty bad :-/

proto/collectd.proto
```diff
@@ -25,35 +25,32 @@
...
 
 service Collectd {
-	// Dispatch collected values to collectd.
-	rpc DispatchValues(DispatchValuesRequest) returns (DispatchValuesReply);
+  // Query a list of values available from collectd's value cache.
+  rpc QueryValues(QueryValuesRequest) returns (stream QueryValuesResponse);
 
-	// Query a list of values available from collectd's value cache.
-	rpc QueryValues(QueryValuesRequest) returns (QueryValuesReply);
+  // DispatchValues sends a stream of ValueLists to the server.
```
tokkee: This should imho document the (expected) error behavior.

tokkee:
:+1: I've added a few comments. LGTM otherwise.

The synchronous interface is much nicer (not unexpectedly). Did you check whether it actually behaves as expected, that is, handling multiple requests asynchronously?

src/grpc.cc
octo: Yeah, also the implementation of the AVL tree doesn't allow for releasing the lock while holding an iterator, so while this would be nice, it's far out :\

proto/collectd.proto
octo: I've reworded this to be phrased from the POV of the server to match QueryValues' description.

src/grpc.cc
octo: Done.

octo:
> Did you check whether [the synchronous API] actually behaves as expected, that is, handling multiple requests asynchronously?

I verified that it can handle multiple connections concurrently. I haven't done any stress testing to see the level of concurrency that we can expect, but the asynchronous implementation is there in the commit history and we can pull it out again if performance turns out to be too poor.

tokkee:
> > Did you check whether [the synchronous API] actually behaves as expected, that is, handling multiple requests asynchronously?
>
> I verified that it can handle multiple connections concurrently. I haven't done any stress testing to see the level of concurrency that we can expect, but the asynchronous implementation is there in the commit history and we can pull it out again if performance turns out to be too poor.

Ack. Thanks for checking.

proto/collectd.proto
```diff
@@ -25,35 +25,38 @@
...
+  // specified shell wildcard patterns (see fnmatch(3)). Use '*' to match
+  // any value.
+  collectd.types.Identifier identifier = 1;
 }
 
 // The response from QueryValues.
-message QueryValuesReply {
-	repeated collectd.types.ValueList values = 1;
-}
+message QueryValuesResponse { collectd.types.ValueList value_list = 1; }
```
tokkee: This inconsistent (auto-)formatting makes me sad :-/ gofmt generally does such a good job at things like that, I wonder why other tools don't manage to do that.

tokkee:
Thanks!

PR #1879, opened 2016-08-11
support older versions of riemann-c-client

rubenk:
riemann_client_set_timeout() was added in 1.8.0, but this
version hasn't landed in all distro's yet.

Fixes #986

### DIFF
--- a/src/write_riemann.c
+++ b/src/write_riemann.c
@@ -101,6 +101,7 @@ static int wrr_connect(struct riemann_host *host) /* {{{ */
                node, port);
     return -1;
   }
+#if RCC_VERSION_NUMBER >= 0x010800
   if (host->timeout.tv_sec != 0) {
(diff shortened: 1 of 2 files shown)


src/write_riemann.c
```diff
@@ -681,9 +683,11 @@ static int wrr_config_node(oconfig_item_t *ci) /* {{{ */
       if (status != 0)
         break;
     } else if (strcasecmp("Timeout", child->key) == 0) {
+#if RCC_VERSION_NUMBER >= 0x010800
       status = cf_util_get_int(child, (int *)&host->timeout.tv_sec);
       if (status != 0)
         break;
+#endif
```
octo: Please add:

```c
#else
  WARNING("write_riemann plugin: The Timeout option is not supported with your version of the Riemann client.");
  // optional: include actual version in warning message.
```

octo:
LGTM, but see inline comment.

Thanks for tackling this Ruben!

PR #1880, opened 2016-08-11
Set _GNU_SOURCE globally

rubenk:
One minor issue is that Python.h unconditionally defines
_GNU_SOURCE (https://bugs.python.org/issue1045893) so we
need to use AH_VERBATIM instead of AC_DEFINE.


octo:
I'm a bit conflicted with this one: on the one hand I don't want contributors to dive into unnecessary GNUisms, on the other hand the *continuous integration* hooks will let us know when something is truly not portable.

rubenk:
Indeed. I did this primarily for Solaris, but now that we have Solaris builders, this is not a big issue anymore.

PR #1882, opened 2016-08-12
gRPC plugin: Implement sending of metrics.

octo:
The config mimicks the *network plugin*, i.e. it looks something like this:

```
<Plugin grpc>
  <Server "example.com" "50051">
    # SSL options
  </Server>
</Plugin>
```

The SSL option names have been copied from the Apache webserver, the option names of the `Listen` block have been changed to match.


rubenk:
LGTM!

PR #1885, opened 2016-08-12
Add a Lua plugin for Collectd

rubenk:
This is a squashed version of #1861, which should make the diff a bit easier to review.

I changed a few things compared to the original work from 2010:
- It now works with Lua 5.2 and 5.3
- The BasePath is prepended to Lua's package.path
- The exported functions are in a Collectd module (aka table).
- Added a few logging functions, so collectd.log_info, collectd.log_error etc.
- Return as many errors as possible to Lua, so they can be caught.

I removed the filtering callbacks for now since I haven't had time to test them. We can always add them back later. I hope you don't mind @schmurfy. Also kept you as the author of the commit and added you to AUTHORS.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -163,6 +163,9 @@ Jérôme Renard <jerome.renard at gmail.com>
 Jiri Tyr <jiri.tyr at gmail.com>
  - fhcount plugin.
 
+Julien Ammous <j.ammous at gmail.com>
+ - Lua plugin.
+
 Kevin Bowling <kbowling at llnw.com>
  - write_tsdb plugin for http://opentsdb.net/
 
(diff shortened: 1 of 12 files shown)


rubenk:
@octo I think this is in a reasonable state to include for 5.6. What do you think?

schmurfy:
no problem for the filters :)

octo:
@rubenk, @schmurfy: This is missing a manpage. I'd suggest to write *collectd-lua(5)* in a fashion similar to *collectd-perl(5)* and *collectd-python(5)*.

Apart from that looks good to go, though I've only glanced at the code due to time-restrictions.

rubenk:
@octo thanks for looking at this! I'll get started on a manpage.

octo:
LGTM, thanks Ruben!

rubenk:
Fixed some more small issues while testing with various versions of Lua, but I think we're good to go. I'll merge when Jenkins is done.

PR #1894, opened 2016-08-16
python pllugin: Some fixes for interactive sessions

trenkel:
I have no idea why the shutdown fix is in here, I though that was already merged?

### DIFF
--- a/src/python.c
+++ b/src/python.c
@@ -925,7 +925,17 @@ static PyMethodDef cpy_methods[] = {
 static int cpy_shutdown(void) {
 	PyObject *ret;
 
-	PyEval_RestoreThread(state);
+	if (!state) {
+		printf("================================================================\n");
+		printf("collectd shutdown while running an iteractive session. This will\n");
+		printf("probably leave your terminal in a mess.\n");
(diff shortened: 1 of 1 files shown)


rubenk:
@trenkel yeah it is.

Is the only commit we need 8b6cead? If so I can fix that up and put it in 5.6

trenkel:
Yes, that's the one. Also, that's some incredibly silly compiler warning.

rubenk:
> Also, that's some incredibly silly compiler warning.

What is?

trenkel:
The thing some of the tests are complaining about:
python.c:1028:7: error: ignoring return value of 'read', declared with attribute warn_unused_result [-Werror=unused-result]
Interestingly it's not complaining about the ignored return value of close, does it have a list of functions with above average important return values?

rubenk:
Ah, that's glibc's doing. It has deemed read() more important than close() indeed ;)
It does this by adding `__attribute__((warn_unused_result))`.

rubenk:
Ok, how does #1899 look?

I fixup a typo and casted the result of read and close to void.

rubenk:
Should be fixed by 89be8ccf23c003d22d112de0fe5417d747fa20a4

PR #1897, opened 2016-08-19
Drop support for libtool version 1

rubenk:
The only distro that we still support that uses libtool version 1
is RHEL5, but that will be EOL in a few months.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ Makefile.in
 /aclocal.m4
 /autom4te.cache
 /autom4te.cache
+/build-aux/
 /compile
 /config.guess
 /config.sub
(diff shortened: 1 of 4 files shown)


rubenk:
Out of curiosity, who's feedback is this pending on? ;)

octo:
Hi @rubenk,

I was a bit concerned that the "Waiting" label was essentially a death sentence for an issue / PR because users can't remove the tag. I've therefore written a little script that tries to automate that.

In this case, it detects that the last comment was made by the creator or the issue / PR and tags it with "Pending feedback", i.e. someone from the team should take a look.

When I've cleaned that up a bit, I'll push it somewhere so we can have a discussion about the rules in place.

rubenk:
Perfect, thanks for the explanation!

rubenk:
Obsoleted by #1898

PR #1898, opened 2016-08-19
Replace libltdl with plain dlopen.

rubenk:


rubenk:
@octo what do you think?

rubenk:
Rebased and reformatted.

PR #1899, opened 2016-08-22
python plugin: Fixing possible problems with the GIL.

rubenk:
If AfterFork is called after threads have been initialized.
Also handle SIGTERM while reading from an interactive session slightly more gracefully.

### DIFF
--- a/src/python.c
+++ b/src/python.c
@@ -925,7 +925,17 @@ static PyMethodDef cpy_methods[] = {
 static int cpy_shutdown(void) {
 	PyObject *ret;
 
-	PyEval_RestoreThread(state);
+	if (!state) {
+		printf("================================================================\n");
+		printf("collectd shutdown while running an interactive session. This will\n");
+		printf("probably leave your terminal in a mess.\n");
+		printf("Run the command \"reset\" to get it back into a usable state.\n");
+		printf("You can press Ctrl+D in the interactive session to\n");
+		printf("close collectd and avoid this problem in the future.\n");
+		printf("================================================================\n");
+	}
+
+	CPY_LOCK_THREADS
 
 	for (cpy_callback_t *c = cpy_shutdown_callbacks; c; c = c->next) {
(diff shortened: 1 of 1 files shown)


rubenk:
Let's wait and see what Jenkins thinks about it before merging.

trenkel:
Jenkins is not impressed by your voidcasting.

trenkel:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=25509 looks related.

rubenk:
That bug was supposed to be fixed in gcc 4.1.0.

rubenk:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425

trenkel:
It was reported in gcc 4.1.0 and judging by the comments never really fixed. The second one seems to be a revival of the same discussion. At this point it seems like the gcc folks have promoted this bug to a feature.

rubenk:
Fun.. Do you have any ideas on how to work around it?

trenkel:
if (read()) ;
perhaps?

trenkel:
Alternatively, disable the clearly broken -W flag.

rubenk:
-Wunused-result is part of -Wall unfortunately, and it is sometimes useful, so I rather not disable it.

trenkel:
Are you sure it's part of -Wall? My man page lists -Wunknown-pragmas -Wunused-function -Wunused-label -Wunused-value -Wunused-variable but notably not -Wunused-result. Then again, gcc on my machine didn't complain about the unused result in the first place, so perhaps wrong version.
Anyway, the empty if statement should do the trick I believe.

rubenk:
In gcc 6 it's part of some default...:
``` 
-Wno-unused-result
Do not warn if a caller of a function marked with attribute "warn_unused_result" does not use its return value. The default is -Wunused-result.
```

Not that my GCC 6 warns about it either. What a mess..

trenkel:
gcc successfully silenced.

rubenk:
\o/ thanks!

PR #1906, opened 2016-09-02
pgsql plugin: avoid parsing error from query_plans

### LINKED ISSUE #1905: postgresql: db query utils: udb_result_submit: Parsing `' as derive failed.
*   Version of collectd:
5.5.1, 5.4.2, current master

*   Operating system / distribution:
Debian, not sure which version.
Postgresql versions: at least from 9.1 to 9.5

## Expected behavior

No error in logs, all values properly handled.

## Actual behavior

```
Initialization complete, entering read-loop.
Successfully connected to database postgres (user perf_user) at server
redacted:5432 (server version: 9.5.2, protocol version: 3, pid: 25691)
parse_value: Failed to parse string as derive: .
db query utils: udb_result_submit: Parsing `' as derive failed.
parse_value: Failed to parse string as derive: .
db query utils: udb_result_submit: Parsing `' as derive failed.
parse_value: Failed to parse string as derive: .
db query utils: udb_result_submit: Parsing `' as derive failed.
db query utils: udb_query_handle_result (flurish_etl, row_stats): All
results failed.
postgresql: udb_query_handle_result failed with status -1.
```

(cf https://mailman.verplant.org/pipermail/collectd/2016-June/006803.html)

## Steps to reproduce

Not been able to reproduce in isolation yet.

What @bzed, @mgedmin and myself have figured out so far, on IRC:
* The error message is misleading due to insuffiscient quoting: `parse_value: Failed to parse string as derive: .` The trailing dot is not the problem, it's the fact that an empty string is returned by the SQL query.
* We suspect the queries run against `pg_stat_user_tables` and `pg_statio_user_tables` are those triggering the error message. To be confirmed.
* When running one of these queries against a non-existing database (eg: `psql -U postgres -d template1 -c "SELECT sum(n_tup_ins) AS ins, sum(n_tup_upd) AS upd, sum(n_tup_del) AS del, sum(n_tup_hot_upd) AS hot_upd FROM pg_stat_user_tables;"`), postgresql returns one row of empty values. As opposed to zeros or other numerical values, when run against a real DB.

bzed:
idx_scan/idx_tup_fetch can be NULL, avoid trying to parse that into a number.

This should fix #1905.

### DIFF
--- a/src/postgresql_default.conf
+++ b/src/postgresql_default.conf
@@ -159,10 +159,10 @@
 </Query>
 
 <Query query_plans>
-	Statement "SELECT sum(seq_scan) AS seq, \
-			sum(seq_tup_read) AS seq_tup_read, \
-			sum(idx_scan) AS idx, \
-			sum(idx_tup_fetch) AS idx_tup_fetch \
+	Statement "SELECT coalesce(sum(seq_scan), 0) AS seq, \
+			  coalesce(sum(seq_tup_read), 0) AS seq_tup_read, \
+			  coalesce(sum(idx_scan), 0) AS idx, \
+			  coalesce(sum(idx_tup_fetch), 0) AS idx_tup_fetch \
 		FROM pg_stat_user_tables;"
 
 	<Result>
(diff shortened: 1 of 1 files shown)


mfournier:
Cherry-picked to the collectd-5.4 branch as e73ae9a and 0b97cf2. So it will be part of every upcoming bugfix and feature release.

Thanks a lot @bzed !

PR #1911, opened 2016-09-05
onewire plugin - improve error messages

zdzichu:
Actually print error messages, return value from OW API is either 0 or -1 - not very informative.


octo:
Hi @zdzichu, thank you very much for your patch!

Can you please remote that merge commit from the PR? It doesn't change the code and only messes up the repository history.

`strerror()` is not thread-safe. Please use `sstrerror()` instead.

Best regards,
—octo

PR #1913, opened 2016-09-05
target_set.c: rename 'MetaDataSet' option

### LINKED ISSUE #1910: [RFC] Target:Set: Rename "MetaDataSet" to "MetaData".
All options to the *Set target* are simply the field they set, e.g.:

```
<Target set>
  Host "foo"
  Plugin "bar"
</Target>
```

The option name `MetaDataSet` seems to break that schema, because it is not clear what the "Set" suffix is referring to. Only one meta data key is being set, not a "set of keys" and if the "set" operation is intended: that one is already in the plugin's name.

I therefore suggest to drop "Set" from the name – it's cleaner ;-)

```
<Target set>
  MetaData "key" "value"
</Target>
```

mfournier:
Fixes #1910

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -9384,7 +9384,7 @@ Available options:
 
 =item B<TypeInstance> I<String>
 
-=item B<MetaDataSet> I<String> I<String>
+=item B<MetaData> I<String> I<String>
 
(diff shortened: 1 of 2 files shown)


octo:
LGTM, thanks Marc!

PR #1915, opened 2016-09-08
write_graphite plugin: add option DropDuplicateFields

mleinart:
The Collectd naming hierarchy often contains multiple levels with duplicate naming. For example, the CPU plugin uses the string "cpu" for both plugin and type (when `ValuesPercentage false`), leading to Graphite metrics with redundant components:
`example_host.cpu.0.cpu.idle`

Other examples include both the Load plugin and Memory plugin which use the same plugin and type name:
`example_host.load.load.shortterm`
`example_host.memory.memory.cached`

This new option drops the latter of these fields when it duplicates a previous metric component.

### DIFF
--- a/src/utils_format_graphite.c
+++ b/src/utils_format_graphite.c
@@ -150,33 +150,28 @@ static int gr_format_name (char *ret, int ret_len,
         sstrncpy (tmp_plugin, n_plugin, sizeof (tmp_plugin));
 
     if (n_type_instance[0] != '\0')
+    {
         if ((flags & GRAPHITE_DROP_DUPE_FIELDS) && strcmp(n_plugin, n_type) == 0)
-        {
             sstrncpy (tmp_type, n_type_instance, sizeof (tmp_type));
-        }
         else
-        {
             ssnprintf (tmp_type, sizeof (tmp_type), "%s%c%s",
                 n_type,
                 (flags & GRAPHITE_SEPARATE_INSTANCES) ? '.' : '-',
                 n_type_instance);
-        }
+    }
     else
         sstrncpy (tmp_type, n_type, sizeof (tmp_type));
-
     /* Assert always_append_ds -> ds_name */
     assert (!(flags & GRAPHITE_ALWAYS_APPEND_DS) || (ds_name != NULL));
     if (ds_name != NULL)
+    {
         if ((flags & GRAPHITE_DROP_DUPE_FIELDS) && strcmp(tmp_plugin, tmp_type) == 0)
-        {
             ssnprintf (ret, ret_len, "%s%s%s.%s.%s",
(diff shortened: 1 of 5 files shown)


src/utils_format_graphite.c
```diff
@@ -150,18 +150,25 @@ static int gr_format_name (char *ret, int ret_len,
         sstrncpy (tmp_plugin, n_plugin, sizeof (tmp_plugin));
 
     if (n_type_instance[0] != '\0')
-        ssnprintf (tmp_type, sizeof (tmp_type), "%s%c%s",
-            n_type,
-            (flags & GRAPHITE_SEPARATE_INSTANCES) ? '.' : '-',
-            n_type_instance);
+        if ((flags & GRAPHITE_DROP_DUPE_FIELDS) && strcmp(n_plugin, n_type) == 0)
```
octo: Please add curly braces around the inner if-else to show which "else" belongs to which "if".

src/utils_format_graphite.c
```diff
@@ -150,18 +150,25 @@ static int gr_format_name (char *ret, int ret_len,
...
+                n_type_instance);
     else
         sstrncpy (tmp_type, n_type, sizeof (tmp_type));
 
     /* Assert always_append_ds -> ds_name */
     assert (!(flags & GRAPHITE_ALWAYS_APPEND_DS) || (ds_name != NULL));
     if (ds_name != NULL)
-        ssnprintf (ret, ret_len, "%s%s%s.%s.%s.%s",
-            prefix, n_host, postfix, tmp_plugin, tmp_type, ds_name);
+        if ((flags & GRAPHITE_DROP_DUPE_FIELDS) && strcmp(tmp_plugin, tmp_type) == 0)
```
octo: Dito, please add curlies.

octo:
Hi @mleinart,

thank you very much for your patch! Overall it looks great! The only thing missing is an update to the *collectd.conf(5)* manual page, which is generated from `src/collectd.conf.pod`.

Best regards,
—octo

mleinart:
Thanks for looking! I've added braces to the inner If statements (looks like it should be BSD style indent, right?). LMK if it looks like the outer Ifs could use them as well.

I've also composed an addition to the man page I think should be sufficient. Sorry to forget that part!

octo:
Sorry, I formulated this poorly. I wanted to have curlies like this:

```c
if (cond)
{ /* <-- */
  if (cond2)
    foo();
  else
    bar();
} /* <-- */
else
  qux();
```

The curlies you added in the mean time can stay, of course, but personally I'd probably remove them. Sorry for the confusion!

mleinart:
No, no problem - I'm all about getting style right :)

octo:
Thanks @mleinart!

PR #1916, opened 2016-09-08
Add missing option to write_graphite example config

mleinart:
The example config for write_graphite included in the shipped collectd.conf is missing one option: SeparateInstances

This option is present [in the wiki](https://collectd.org/wiki/index.php/Plugin:Write_Graphite) and is valid [in the current version](https://github.com/collectd/collectd/blob/collectd-5.5.2/src/write_graphite.c#L541-L543)

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1415,6 +1415,7 @@
 #    StoreRates true
 #    AlwaysAppendDS false
 #    EscapeCharacter "_"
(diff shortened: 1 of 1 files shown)


octo:
Thanks @mleinart!
Cherry-picked into *collectd-5.5* as 8d78a2b27ddcc2e0dbca30be9c888479a4757c1f

PR #1918, opened 2016-09-09
Add RFC3339 Zulu time functions

sparkprime:
"Zulu" time is always in UTC and the timezone part is just a 'Z'.  This is a pretty common form, and it is a requirement of the Google Stackdriver metric ingestion API.

This PR adds a couple of functions alongside the existing RFC3339 ones (the latter use local time instead of UTC and give the timezone in the +hh:mm form).

thanks

### DIFF
--- a/src/daemon/utils_time.c
+++ b/src/daemon/utils_time.c
@@ -54,10 +54,10 @@ cdtime_t cdtime (void) /* {{{ */
     char errbuf[1024];
     ERROR ("cdtime: clock_gettime failed: %s",
         sstrerror (errno, errbuf, sizeof (errbuf)));
-    return (0);
+    return 0;
   }
 
-  return (TIMESPEC_TO_CDTIME_T (&ts));
+  return TIMESPEC_TO_CDTIME_T (&ts);
 } /* }}} cdtime_t cdtime */
 # else /* !HAVE_CLOCK_GETTIME */
 /* Work around for Mac OS X which doesn't have clock_gettime(2). *sigh* */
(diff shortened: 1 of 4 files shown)


sparkprime:
FYI @igorpeshansky

octo:
Hi @sparkprime, thank you very much for your patch!

Could you please replace your use of "zulu" with "UTC"? For example, the argument to `format_rfc3339()` should be `_Bool use_utc`. I've done a fair bit of time handling in code in my time and haven't come across "zulu" yet, while "UTC" is very widely known. In fact, you use "UTC" to explain what "zulu" is in the PR description ;)

Thanks and best regards,
—octo

src/daemon/utils_time.h
```diff
@@ -85,12 +85,23 @@ cdtime_t cdtime (void);
...
+#define RFC3339NANO_ZULU_SIZE 31
+
 /* rfc3339 formats a cdtime_t time in RFC 3339 format with second precision. */
 int rfc3339 (char *buffer, size_t buffer_size, cdtime_t t);
 
 /* rfc3339nano formats a cdtime_t time in RFC 3339 format with nanosecond
  * precision. */
 int rfc3339nano (char *buffer, size_t buffer_size, cdtime_t t);
 
+/* rfc3339 formats a cdtime_t time in RFC 3339 zulu format with second
```
octo: Please update the functions name here.

src/daemon/utils_time.h
```diff
@@ -85,12 +85,23 @@ cdtime_t cdtime (void);
...
+ * precision. */
+int rfc3339_zulu (char *buffer, size_t buffer_size, cdtime_t t);
+
+/* rfc3339nano formats a cdtime_t time in RFC 3339 zulu format with nanosecond
```
octo: Dito: function name is out of date.

sparkprime:
I suggest keeping it because UTC could mean "2014-10-02T15:01:23.045123456+00:00" whereas Zulu (as described in the RFC) specifically refers to having a Z on the end.

octo:
How about we turn this around: let `rfc3339()` return Zulu / UTC time and add new `rfc3339_local()` functions? There are only two existing callers: the JSON export arguably should use UTC anyway and the *PostgreSQL plugin* probably doesn't care, though I haven't checked. 

Keeping it as-is is also fine (i.e. with the \_zulu suffix), I don't care too deeply, just make sure that there are comments that make it obvious that the difference between the existing functions and the new ones is the time zone, i.e. the comment should mention "UTC" and, ideally, "time zone".

igorpeshansky:

The `rfc3339()`/`rfc3339nano()` functions are now used for UTC, and I've added `rfc3339_local()`/`rfc3339nano_local()` to return local time. I've kept `rfc3339_zulu()`/`rfc3339nano_zulu()`, because the format is different, but I did mention UTC in the comments. Also factored out as much as I could into shareable helper functions.

It made sense to switch JSON export to UTC, but for backward compatibility the _PostgreSQL plugin_ needed to keep the local time (I've also updated the documentation to match reality).

Would you like me to squash the commits at some point, or should we just merge as-is?

src/daemon/utils_time.c
```diff
@@ -112,60 +164,106 @@ static int format_zone (char *buffer, size_t buffer_size, struct tm const *tm) /
...
+  return format_rfc3339 (buffer, buffer_size, &t_tm, nsec, print_nano, zone);
+} /* }}} int format_rfc3339_utc */
 
-  len = strftime (base, sizeof (base), "%Y-%m-%dT%H:%M:%S", &t_tm);
-  if (len == 0)
-    return ENOMEM;
+int format_rfc3339_local (char *buffer, size_t buffer_size, cdtime_t t, _Bool print_nano) /* {{{ */
+{
+  struct tm t_tm;
+  long nsec;
```
octo: Our continuous integration shows that some versions of GCC think that `nsec` might be used uninitialized. I don't think that's true, but I guess initializing variables is probably always a good thing.

igorpeshansky:
Thank you for the review, @octo! I've addressed your latest comments and pushed a new commit.

PR #1920, opened 2016-09-09
onewire plugin - improve error messages

zdzichu:
Actually print error messages, return value from OW API is either 0 or -1 - not very informative.

(This is redoing of pull request #1911 with comments addressed; sorry for creatine new PR)


octo:
Thanks @zdzichu!

PR #1922, opened 2016-09-09
Allow Target:Set to delete metadata and interpolate field values.

igorpeshansky:
This uses the syntax from Target:Notification to allow interpolating field and metadata values via Target:Set. It also adds a new DeleteMetaData option to allow deleting metadata keys.

### DIFF
--- a/src/daemon/meta_data.c
+++ b/src/daemon/meta_data.c
@@ -739,6 +739,7 @@ int meta_data_as_string (meta_data_t *md, /* {{{ */
   char *actual;
   char buffer[MD_MAX_NONSTRING_CHARS];  /* For non-string types. */
   char *temp;
+  int type;
 
   if ((md == NULL) || (key == NULL) || (value == NULL))
     return (-EINVAL);
@@ -752,7 +753,9 @@ int meta_data_as_string (meta_data_t *md, /* {{{ */
     return (-ENOENT);
   }
 
-  switch (e->type)
+  type = e->type;
+
+  switch (type)
   {
     case MD_TYPE_STRING:
       actual = e->value.mv_string;
@@ -774,20 +777,19 @@ int meta_data_as_string (meta_data_t *md, /* {{{ */
       break;
     default:
       pthread_mutex_unlock (&md->lock);
-      ERROR ("meta_data_as_string: unknown type %d for key `%s'",
-          e->type, e->key);
+      ERROR ("meta_data_as_string: unknown type %d for key `%s'", type, key);
       return (-ENOENT);
   }
+
+  pthread_mutex_unlock (&md->lock);
 
   temp = md_strdup (actual);
   if (temp == NULL)
   {
     pthread_mutex_unlock (&md->lock);
-    ERROR ("meta_data_as_string: md_strdup failed for key `%s'.", e->key);
+    ERROR ("meta_data_as_string: md_strdup failed for key `%s'.", key);
     return (-ENOMEM);
   }
-
-  pthread_mutex_unlock (&md->lock);
 
   *value = temp;
 
--- a/src/daemon/meta_data.c
+++ b/src/daemon/meta_data.c
@@ -26,8 +26,11 @@
 
 #include "collectd.h"
 
+#include "common.h"
 #include "plugin.h"
 #include "meta_data.h"
+
+#define MD_MAX_NONSTRING_CHARS 128
 
 /*
  * Data types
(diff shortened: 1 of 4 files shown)


src/target_set.c
```diff
@@ -92,31 +115,168 @@ static int ts_config_add_meta (meta_data_t **dest, /* {{{ */
...
 
   return (meta_data_add_string (*dest, key, string));
 } /* }}} int ts_config_add_meta */
 
+static int ts_config_add_meta_delete (ts_key_list_t **dest, /* {{{ */
+    const oconfig_item_t *ci)
+{
+  ts_key_list_t *entry = NULL;
+
+  if ((ci->values_num != 1)
```
octo: You can replace this check and the *strdup()* below with a convenience function:

```c
if (cf_util_get_string (ci, &entry->key) != 0)
  return -1; // error has already been reported
```

src/target_set.c
```diff
@@ -238,17 +421,75 @@ static int ts_invoke (const data_set_t *ds, value_list_t *vl, /* {{{ */
...
+    }
+
+    meta_entries = meta_data_toc (data->meta, &meta_toc);
+    for (int i = 0; i < meta_entries; i++)
+    {
+      const char *key = meta_toc[i];
+      char *string;
+      int status;
+
+      status = meta_data_get_string (data->meta, key, &string);
```
octo: This will result in two errors (one in `meta_data_get_string()` and one here) for each non-string meta data entry.

You have a long-ish `switch (meta_type)` above. How about we move this into `src/daemon/meta_data.c` as `meta_data_as_string()` which returns a meta data entry as string, regardless of it's actual type. We could then re-use that here.

src/target_set.c
```diff
@@ -92,31 +115,168 @@ static int ts_config_add_meta (meta_data_t **dest, /* {{{ */
...
+#define REPLACE_FIELD(t, v) \
+  if (subst_string (temp, sizeof (temp), dest, t, v) != NULL) \
+    sstrncpy (dest, temp, size);
+  REPLACE_FIELD ("%{host}", vl->host);
+  REPLACE_FIELD ("%{plugin}", vl->plugin);
+  REPLACE_FIELD ("%{plugin_instance}", vl->plugin_instance);
+  REPLACE_FIELD ("%{type}", vl->type);
+  REPLACE_FIELD ("%{type_instance}", vl->type_instance);
+
+  meta_entries = meta_data_toc (vl->meta, &meta_toc);
```
octo: I think `meta_toc` leaks. You free with:

```c
strarray_free (meta_toc, (size_t) meta_entries);
```

src/target_set.c
```diff
@@ -238,17 +421,75 @@ static int ts_invoke (const data_set_t *ds, value_list_t *vl, /* {{{ */
...
+    int meta_entries;
+    char **meta_toc;
+
+    if ((new_meta = meta_data_create()) == NULL)
+    {
+      ERROR ("Target `set': failed to create replacement metadata.");
+      return (-ENOMEM);
+    }
+
+    meta_entries = meta_data_toc (data->meta, &meta_toc);
```
octo: Dito: `meta_toc` leaks.

src/target_set.c
```diff
@@ -92,31 +115,168 @@ static int ts_config_add_meta (meta_data_t **dest, /* {{{ */
...
+static void ts_subst (char *dest, size_t size, const char *string, /* {{{ */
+    const value_list_t *vl)
+{
+  char temp[DATA_MAX_NAME_LEN];
+  int meta_entries;
+  char **meta_toc;
+
+  /* Initialize the field with the template. */
+  sstrncpy (dest, string, size);
+
```
octo: This may be premature optimization, but this function feels like it could be a hotspot. How about adding a shortcut here, along the lines of:

```c
if (strchr (dest, '%') == NULL)
  return;
```

src/target_set.c
```diff
@@ -92,31 +115,168 @@ static int ts_config_add_meta (meta_data_t **dest, /* {{{ */
...
+      case MD_TYPE_STRING:
+        {
+          char *meta_value;
+          if (meta_data_get_string (vl->meta, key, &meta_value))
+          {
+            ERROR ("Target `set': Unable to get string metadata value `%s'.",
+                key);
+            continue;
+          }
+          sstrncpy (value_str, meta_value, sizeof (value_str));
```
octo: `meta_value` leaks:

```c
sfree (meta_value);
```

src/target_set.c
```diff
@@ -28,6 +28,28 @@
 
 #include "common.h"
 #include "filter_chain.h"
+#include "meta_data.h"
+#include "utils_subst.h"
+
+struct ts_key_list_s
```
octo: Please consider using a `char **` instead of this linked list. You can use the convenience functions from `common.h` which should make this quite easy:

```c
char **keys = NULL;
size_t keys_num = 0;

strarray_add (&keys, &keys_num, "foobar");
strarray_free (keys, keys_num);
```

octo:
Thank you very much for your patch @igorpeshansky! I've left some review comments inline. Nothing drastic, as we've talked about the design beforehand, but there are some leaks that need to be fixed and I've pointed out some convenience functions that should simplify the code somewhat.

Best regards,
—octo

src/target_set.c
igorpeshansky: In this case, `data->meta` is filled in by `ts_config_add_meta()`, which uses `meta_data_add_string()`, so it's guaranteed to be a string. Plus, the error printed by `meta_data_get_string()` will not show that this was a failure in `target_set`.

I really like the idea of `meta_data_as_string()`, and added it for use in `ts_subst()`.

src/target_set.c
igorpeshansky: I've tried it out, but the cost of using the `strarray...` functions is having two data members that have to be passed around, which is somewhat ugly. If you don't mind, I'll stick with the current approach (why don't we have convenience functions for linked lists again?). If you still think I should be using `strarray...` here, please let me know and I'll change it.

igorpeshansky:

src/daemon/meta_data.c
```diff
@@ -729,4 +732,66 @@ int meta_data_get_boolean (meta_data_t *md, /* {{{ */
...
+      break;
+    case MD_TYPE_DOUBLE:
+      ssnprintf (buffer, sizeof (buffer), GAUGE_FORMAT, e->value.mv_double);
+      actual = buffer;
+      break;
+    case MD_TYPE_BOOLEAN:
+      actual = e->value.mv_boolean ? "true" : "false";
+      break;
+    default:
+      pthread_mutex_unlock (&md->lock);
```
octo: `e` is protected by `md->lock` and accessed after the lock is given up, potentially creating a race here. I'd suggest to store the type in a temporary variable:

```c
int md_type = e->type;
unlock();
ERROR("… type %d, key %s", md_type, key);
```

src/target_set.c
octo: That's fine, it was just a suggestion.

Utility functions for general linked lists are in `src/daemon/utils_llist.h`, if you want to take a look at that. Since you only use *append* and iterate over the list, the benefits of switching to that are probably few.

src/target_set.c
```diff
@@ -92,31 +115,102 @@ static int ts_config_add_meta (meta_data_t **dest, /* {{{ */
...
+
+  if (cf_util_get_string (ci, &entry->key) != 0)
+    return (-1);  /* An error has already been reported. */
+
+  if (strlen (entry->key) == 0)
+  {
+    ERROR ("Target `set': The `%s' option does not accept empty string as "
+        "first argument.", ci->key);
+    sfree (entry->key);
+    return (-1);
```
octo:
```c
sfree (entry);
```

src/target_set.c
octo: Good point, thanks.

octo:
Hi again, there is one use-after-unlock and one leak-on-error, after that it's good to go. Thanks @igorpeshansky!

igorpeshansky:
Thank you for the review, @octo! Sorry about more leaks. I've addressed your latest comments and pushed a new commit. Would you like me to squash them at some point?

octo:
Thanks @igorpeshansky!

PR #1923, opened 2016-09-09
Allow Target:Replace to manipulate metadata.

igorpeshansky:
This allows using Target:Replace to transform metadata values. It also adds a new DeleteMetaData option to allow conditionally deleting metadata keys whose value matches the regex.

### DIFF
--- a/src/target_replace.c
+++ b/src/target_replace.c
@@ -38,7 +38,7 @@ struct tr_action_s
 {
   regex_t re;
   char *replacement;
-  int may_be_empty;
+  _Bool may_be_empty;
 
   tr_action_t *next;
 };
@@ -113,7 +113,7 @@ static void tr_meta_data_action_destroy (tr_meta_data_action_t *act) /* {{{ */
 } /* }}} void tr_meta_data_action_destroy */
 
 static int tr_config_add_action (tr_action_t **dest, /* {{{ */
-    const oconfig_item_t *ci, int may_be_empty)
+    const oconfig_item_t *ci, _Bool may_be_empty)
 {
   tr_action_t *act;
   int status;
@@ -158,8 +158,7 @@ static int tr_config_add_action (tr_action_t **dest, /* {{{ */
   if (act->replacement == NULL)
   {
     ERROR ("tr_config_add_action: tr_strdup failed.");
-    regfree (&act->re);
-    sfree (act);
+    tr_action_destroy (act);
     return (-ENOMEM);
   }
 
@@ -181,7 +180,7 @@ static int tr_config_add_action (tr_action_t **dest, /* {{{ */
 } /* }}} int tr_config_add_action */
 
 static int tr_config_add_meta_action (tr_meta_data_action_t **dest, /* {{{ */
-    const oconfig_item_t *ci, int should_delete)
+    const oconfig_item_t *ci, _Bool should_delete)
 {
   tr_meta_data_action_t *act;
   int status;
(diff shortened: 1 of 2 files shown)


src/target_replace.c
```diff
@@ -215,6 +344,112 @@ static int tr_action_invoke (tr_action_t *act_head, /* {{{ */
...
+      }
+
+      meta_data_status = meta_data_add_string (result, act->key, temp);
+
+      if (meta_data_status != 0)
+      {
+        ERROR ("Target `replace': Unable to set metadata value for `%s'.",
+            act->key);
+        return (meta_data_status);
+      }
```
octo: leaks "result"

src/target_replace.c
```diff
@@ -215,6 +344,112 @@ static int tr_action_invoke (tr_action_t *act_head, /* {{{ */
...
+      continue;
+    if (value_type != MD_TYPE_STRING)
+    {
+      ERROR ("Target `replace': Attempting replace on metadata key `%s', "
+          "which isn't a string.",
+          act->key);
+      continue;
+    }
+
+    meta_data_status = meta_data_get_string (*dest, act->key, &value);
```
octo: I think "value" needs to be freed.

src/target_replace.c
```diff
@@ -153,6 +180,108 @@ static int tr_config_add_action (tr_action_t **dest, /* {{{ */
...
+
+  act->key = tr_strdup (ci->values[0].value.string);
+  if (act->key == NULL)
+  {
+    ERROR ("tr_config_add_meta_action: tr_strdup failed.");
+    sfree (act);
+    return (-ENOMEM);
+  }
+
+  status = regcomp (&act->re, ci->values[1].value.string, REG_EXTENDED);
```
octo: If you initialize `act->re` first, you can simply call `tr_meta_data_action_destroy()` if any of the *strdup()* fail later on.

src/target_replace.c
```diff
@@ -215,6 +344,112 @@ static int tr_action_invoke (tr_action_t *act_head, /* {{{ */
...
+    int value_type;
+    int meta_data_status;
+    char *value;
+
+    value_type = meta_data_type (*dest, act->key);
+    if (value_type == 0)  /* not found */
+      continue;
+    if (value_type != MD_TYPE_STRING)
+    {
+      ERROR ("Target `replace': Attempting replace on metadata key `%s', "
```
octo: This should be a `WARNING()`, because we can continue gracefully after this issue.

src/target_replace.c
```diff
@@ -215,6 +344,112 @@ static int tr_action_invoke (tr_action_t *act_head, /* {{{ */
...
+    {
+      char errbuf[1024] = "";
+
+      regerror (status, &act->re, errbuf, sizeof (errbuf));
+      ERROR ("Target `replace': Executing a regular expression failed: %s.",
+          errbuf);
+      continue;
+    }
+
+    if (act->replacement != NULL)
```
octo: Can you please invert the logic here and *continue* early:

```c
if (act->replacement == NULL)
{
  meta_data_delete (*dest, act->key);
  continue;
}
// replacement logic with one fewer level of indentation.
```

src/target_replace.c
```diff
@@ -153,6 +180,108 @@ static int tr_config_add_action (tr_action_t **dest, /* {{{ */
   return (0);
 } /* }}} int tr_config_add_action */
 
+static int tr_config_add_meta_action (tr_meta_data_action_t **dest, /* {{{ */
+    const oconfig_item_t *ci, int should_delete)
```
octo: Nit: `_Bool should_delete`

octo:
Hi @igorpeshansky, thank you very much for the patch! I've left some comments inline – once they're resolved this is good to go.

src/target_replace.c
igorpeshansky: Hmm, by that token, the failure to execute the regular expression should also be a `WARNING()`, right?

igorpeshansky:
Thank you for the review, @octo! Sorry about the leaks. I've addressed your comments and pushed a new commit.

octo:
Thanks @igorpeshansky!

igorpeshansky:
Thanks! If you don't mind, two questions on GitHub protocol:

1. Should I have responded to each comment with "Done"?

2. There was a remaining question about the regex execution failure also being a `WARNING()` - was there something I should have said in my PR response to call your attention to it?

PR #1924, opened 2016-09-09
Change Plugin:WriteLog to also write JSON; add optional configuration.

igorpeshansky:
@tokkee FYI.

### DIFF
--- a/src/write_log.c
+++ b/src/write_log.c
@@ -40,6 +40,9 @@
 #define WL_FORMAT_GRAPHITE 1
 #define WL_FORMAT_JSON 2
 
+/* Plugin:WriteLog has to also operate without a config, so use a global. */
+int wl_format = WL_FORMAT_GRAPHITE;
+
 static int wl_write_graphite (const data_set_t *ds, const value_list_t *vl)
 {
     char buffer[WL_BUF_SIZE] = { 0 };
@@ -84,16 +87,15 @@ static int wl_write_json (const data_set_t *ds, const value_list_t *vl)
 } /* int wl_write_json */
 
 static int wl_write (const data_set_t *ds, const value_list_t *vl,
-        user_data_t *user_data)
+        __attribute__ ((unused)) user_data_t *user_data)
 {
     int status = 0;
-    int mode = (int) (size_t) user_data->data;
 
-    if (mode == WL_FORMAT_GRAPHITE)
+    if (wl_format == WL_FORMAT_GRAPHITE)
     {
         status = wl_write_graphite (ds, vl);
     }
-    else if (mode == WL_FORMAT_JSON)
+    else if (wl_format == WL_FORMAT_JSON)
     {
         status = wl_write_json (ds, vl);
     }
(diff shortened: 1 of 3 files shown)


src/write_log.c
```diff
@@ -56,21 +59,111 @@ static int wl_write_messages (const data_set_t *ds, const value_list_t *vl)
...
+static int wl_config (oconfig_item_t *ci) /* {{{ */
+{
+    int mode = 0;
+    for (int i = 0; i < ci->children_num; i++)
+    {
+        oconfig_item_t *child = ci->children + i;
+
+        if (strcasecmp ("Format", child->key) == 0)
+        {
+            char *mode_str = NULL;
```
octo: I think this block would benefit from using `cf_util_get_string_buffer()`

src/write_log.c
```diff
@@ -56,21 +59,111 @@ static int wl_write_messages (const data_set_t *ds, const value_list_t *vl)
...
+        {
+            ERROR ("write_log plugin: Invalid configuration option: `%s'.",
+                child->key);
+        }
+    }
+    if (mode == 0)
+        mode = WL_FORMAT_GRAPHITE;
+
+    user_data_t ud = {
+        .data = (void *) (size_t) mode,
```
octo: I'd prefer to actually allocate the memory, because this *will* get flagged by Coverity or scan-build, the static analysis tools we use.

src/write_log.c
```diff
@@ -56,21 +59,111 @@ static int wl_write_messages (const data_set_t *ds, const value_list_t *vl)
...
 {
-    plugin_register_write ("write_log", wl_write, NULL);
+    plugin_register_complex_config ("write_log", wl_config);
+
+    user_data_t ud = {
+        .data = (void *) (size_t) WL_FORMAT_GRAPHITE,
+        .free_func = NULL
+    };
+
+    plugin_register_write ("write_log", wl_write, &ud);
```
octo: `plugin_register_write()` is called from the config callback, so this call should go.

octo: Hi @igorpeshansky, thank you very much for your patch! I don't think that setting the *mode* in the `user_data_t` struct is backwards compatible; I think we need to use a global variable instead.

Best regards,
—octo

igorpeshansky: Thanks for the review! I'd like to understand a bit more about the backwards incompatibility here. The reason for the two `plugin_register_write()` calls is exactly to allow the plugin to run with or without a config block. The code here actually works, but does log an error when the callback is overridden (so previous behavior is completely unchanged). Is there an implied invariant that `plugin_register_write()` will always be called exactly once? Instead of a global variable, would it make sense to add some sort of a callback update function that effectively modifies the `user_data_t` of an existing callback?

octo: @igorpeshansky You're right, I hadn't considered that duplicate registration with the same name would overwrite a previous registration – I was under the impression that the plugin would be called twice when a config was present.

There are a couple of plugins that register write callbacks from within the config callback, but they all *need* there to be a configuration in order to function (and usually allow to configure more than one instance of the plugin). By *moving* the registration to the config handler (and removing it from `module_register()`) we'd *require* a configuration, breaking backwards compatibility.

I'd prefer to call `plugin_register_write()` once from `module_register()` and let the *config* callback set a global variable: no other plugin is calling `plugin_register_*()` with the same name twice *on purpose*, so this behavior is not intuitive. The warning that is logged in that case exists because this is unexpected behavior. Also, passing configuration via a `user_data_t` even though there is only one instance of the plugin is unusual and therefore confusing. Other plugins with only one registered callback simply ignore that user data argument.

Best regards,
—octo

igorpeshansky: Makes sense, thanks.

igorpeshansky:

octo:
Thanks @igorpeshansky!

PR #1927, opened 2016-09-11
Tree wide: Use user_data_t compound literal where appropriate.

octo:
`user_data_t` is one of those structs where we used to create a variable just for a single function call. This patch inlines these cases into the function call using a compound literal.


rubenk:
LGTM!

PR #1930, opened 2016-09-12
Allow Match:Regex to match metadata.

igorpeshansky:


src/match_regex.c
```diff
@@ -295,6 +371,24 @@ static int mr_match (const data_set_t __attribute__((unused)) *ds, /* {{{ */
...
 				vl->type_instance) == FC_MATCH_NO_MATCH)
 		return (nomatch_value);
+	if (vl->meta != NULL)
+	{
+		for (llentry_t *e = llist_head(m->meta); e != NULL; e = e->next)
+		{
+			mr_regex_t *meta_re = (mr_regex_t *) e->value;
+			char *value;
+			int status = meta_data_get_string (vl->meta, e->key, &value);
+			if (status == 0)  /* key is present */
```
octo: Please do:

```c
if (status != 0) /* key is not present */
  continue;
/* remainder of code */
```

src/match_regex.c
```diff
@@ -93,6 +96,12 @@ static void mr_free_match (mr_match_t *m) /* {{{ */
 	mr_free_regex (m->plugin_instance);
 	mr_free_regex (m->type);
 	mr_free_regex (m->type_instance);
+	for (llentry_t *e = llist_head(m->meta); e != NULL; e = e->next)
+	{
+		free (e->key);
```
octo: Please search and replace all `free()` with `sfree()`. Ruben just did a tree-wide unification and I'd prefer not to muddy the waters again.

src/match_regex.c
```diff
@@ -184,8 +187,78 @@ static int mr_config_add_regex (mr_regex_t **re_head, /* {{{ */
...
+		{
+			log_err ("mr_config_add_meta_regex: llentry_create failed.");
+			free (key);
+			return (-1);
+		}
+		/* key and entry will now be freed by mr_free_match(). */
+		llist_append (*meta, entry);
+	}
+
+	snprintf (buffer, sizeof (buffer), "%s `%s'", ci->key, key);
```
octo: At your option: I find `ci->key` and `key` a bit hard to read. How about renaming `key` to `meta_name`?

src/match_regex.c
octo: Is this actually the correct behavior? If I set `MetaData "foo" "^bar"` and a metric does *not* have the "foo" meta data at all, I'd expect the that to be a no-match.

octo:
Thanks @igorpeshansky, great work as usual :)

src/match_regex.c
igorpeshansky: Good catch, this felt wrong to me as well. Fixed.

src/match_regex.c
igorpeshansky: I renamed to meta_key instead, to be consistent with meta_data naming.

igorpeshansky:
Thank you for the review, @octo! I've addressed your comments and pushed a new commit. Would you like me to squash them at some point?

octo:
Thanks @igorpeshansky! I spotted a use of `snprintf()` I missed on the previous iteration and fixed that to avoid having another round-trip. I'll merge as soon as the build status is available and successful.

PR #1931, opened 2016-09-13
Tree wide: Use compound literals when dealing with value_t.

octo:
There are a lot of `value_t` and `value_t[]` in "submit" functions that are effectively used to cast a function argument from `gauge_t` or `derive_t` to `value_t`. This is more elegently done using compound literals.


rubenk:
Nice cleanup, LGTM!

Issue #1921, opened 2016-09-09
collectdctl listval truncates at 64


mfournier:
I just noticed that when building collectd with larger `DATA_MAX_NAME_LEN` values than 64 and submitting values with VL members over 64 chars, they get truncated at 64 in the output of `collectdctl listval`.

I tried fiddling with https://github.com/collectd/collectd/blob/master/src/collectdctl.c#L398-L400 and https://github.com/collectd/collectd/blob/master/src/libcollectdclient/collectd/client.h#L51 but the problem doesn't seem as trivial as that.

I didn't take much time to dig into this yet, so I might have missed something.

PR #1934, opened 2016-09-14
libcollectdclient: Increase LCC_NAME_LEN to 128.

octo:
This matches the new default in collectd itself. Unfortunately, this changes the size of public structures and therefore breaks ABI compatibility, hence the update to "version-info".

Fixes: #1921


octo:
@tokkee had a good point about this: we should start using `char *` in the `identifier_t` and avoid this issue altogether. He also has additional API breaking suggestions that we should do in the same (breaking) version. See also: #1748

octo:
With collectd 6 this is no longer going to be relevant. Closing as "won't fix".

PR #1942, opened 2016-09-17
use fgets in parse_value_file, fixes #1941

### LINKED ISSUE #1941: use of parse_value_file leads to occasional garbage in entropy plugin
*   Version of collectd: master
*   Operating system / distribution: Sailfish 2.0.2.51

I merged the current master to sailfish build and started getting messages from parse_value regarding garbage values. Turned out that it was induced by entropy plugin reading entropy_avail file which occasionally [once in a few reads] resulted in garbage string after the correct readout. This is a regression from earlier behaviour: never seen something like it before.

As I could see from the diff, old way was to use fgets in entropy plugin to read in data. Possibly, kernel on my device has bugs in showing something through /proc interface and the file is not correctly terminated sometimes [although cat shows correct content, every time I tried].

Since parse_value_file is used for the files which just have one value, I suggest to reimplement it to read in using fgets. That would also allow to skip strstripnewline in that function.

Finally, in this implementation there is at least one unnecessary malloc/free in parse_value. We are going to dump the buffer anyway when calling it from parse_value_file and there is no need to make a copy and alter it later. 

As these functions (parse_value and parse_value_file) are called from all over, I think its OK to make them as efficient as possible even if it leads to duplication of some code. As a result we would get a code that is easy to use, but which is efficient as well.

rinigus:
By using fgets in parse_value_file, this PR fixes #1941.

In addition, parse_value_ext is introduced that simplifies the debugging of the issues raised in this function. It allows to add comment string that is printed as a part of INFO or other messages. It allowed me to find which file was responsible for garbage errors when used together with parse_value_file

### DIFF
--- a/src/daemon/common.c
+++ b/src/daemon/common.c
@@ -1091,6 +1091,11 @@ int parse_identifier_vl (const char *str, value_list_t *vl) /* {{{ */
 
 int parse_value (const char *value_orig, value_t *ret_value, int ds_type)
 {
+  return parse_value_ext(value_orig, ret_value, ds_type, "");
+}
+
+int parse_value_ext (const char *value_orig, value_t *ret_value, int ds_type, const char *error_identifier)
+{
   char *value;
   char *endptr = NULL;
   size_t value_len;
@@ -1129,19 +1134,19 @@ int parse_value (const char *value_orig, value_t *ret_value, int ds_type)
 
     default:
       sfree (value);
-      ERROR ("parse_value: Invalid data source type: %i.", ds_type);
+      ERROR ("parse_value %s: Invalid data source type: %i.", error_identifier, ds_type);
       return -1;
   }
 
   if (value == endptr) {
-    ERROR ("parse_value: Failed to parse string as %s: %s.",
+    ERROR ("parse_value %s: Failed to parse string as %s: %s.", error_identifier,
         DS_TYPE_TO_STRING (ds_type), value);
     sfree (value);
     return -1;
   }
   else if ((NULL != endptr) && ('\0' != *endptr))
-    INFO ("parse_value: Ignoring trailing garbage \"%s\" after %s value. "
-        "Input string was \"%s\".",
+    INFO ("parse_value %s: Ignoring trailing garbage \"%s\" after %s value. "
+        "Input string was \"%s\".", error_identifier,
         endptr, DS_TYPE_TO_STRING (ds_type), value_orig);
 
   sfree (value);
@@ -1211,14 +1216,22 @@ int parse_values (char *buffer, value_list_t *vl, const data_set_t *ds)
 
 int parse_value_file (char const *path, value_t *ret_value, int ds_type)
 {
+	FILE *fh;
 	char buffer[256];
 
-	if (read_file_contents (path, buffer, sizeof (buffer)) < 0)
-		return errno;
+	fh = fopen (path, "r");
+	if (fh == NULL)
+		return (-1);
 
-	strstripnewline (buffer);
+	if (fgets (buffer, sizeof (buffer), fh) == NULL)
+	{
+		fclose (fh);
+		return (-1);
+	}
 
-	return parse_value (buffer, ret_value, ds_type);
+	fclose (fh);
+
+	return parse_value_ext (buffer, ret_value, ds_type, path);
 } /* int parse_value_file */
 
 #if !HAVE_GETPWNAM_R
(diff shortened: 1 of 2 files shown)


rinigus:
@octo, I am puzzled with the failure on the checks and I couldn't find an error so far. As I could see, the error is in valgrind (or something similar) check at 

https://ci.collectd.org/job/pull-requests-build-on-freebsd10-amd64-with-default-toolchain/885/artifact/collectd-5.6.0.106.gfcbad61/src/daemon/test_common.log/*view*/

The line in question is 

```
value_len = strlen (value);
```

in `parse_value_ext` which should be the same as `parse_value` earlier. Would you mind to take a look and tell me where I made an error? To my knowledge, I haven't changed that part...

Best wishes,

rinigus

octo:
Hi @rinigus, thanks for your patch!

FTR, the error is:

```
==31755== Invalid read of size 4
==31755==    at 0x404DCE: parse_value_ext (common.c:1109)
==31755==    by 0x405172: parse_values (common.c:1206)
==31755==    by 0x403558: test_parse_values (common_test.c:338)
==31755==    by 0x402310: main (common_test.c:396)
==31755==  Address 0x58086d0 is 0 bytes inside a block of size 3 alloc'd
==31755==    at 0x4C226C0: malloc (in /usr/local/lib/valgrind/vgpreload_memcheck-amd64-freebsd.so)
==31755==    by 0x50DE94F: strdup (in /lib/libc.so.7)
==31755==    by 0x404DB9: parse_value_ext (common.c:1106)
==31755==    by 0x405172: parse_values (common.c:1206)
==31755==    by 0x403558: test_parse_values (common_test.c:338)
==31755==    by 0x402310: main (common_test.c:396)
==31755== 
```

The code in `parse_value_ext()` is:

```c
/*1106*/  value = strdup (value_orig); /* value is allocated and initialized */
/*1107*/  if (value == NULL)
/*1108*/    return (ENOMEM);
/*1109*/  value_len = strlen (value); /* "invalid read" */
```

I also don't see why `strdup()` should read four bytes and why this would surface now.

src/daemon/common.c
```diff
@@ -1091,6 +1091,11 @@ int parse_identifier_vl (const char *str, value_list_t *vl) /* {{{ */
 
 int parse_value (const char *value_orig, value_t *ret_value, int ds_type)
 {
+  return parse_value_ext(value_orig, ret_value, ds_type, "");
+}
+
+int parse_value_ext (const char *value_orig, value_t *ret_value, int ds_type, const char *error_identifier)
```
octo: Not suite sure about `error_identifier`: it introduces a space in front of the colon in the common `parse_value()` case and is orthogonal to the `fgets()` change. To address the first complaint, you could turn this into `func_name` and, in `parse_value_file()`, do something along the lines of:

```c
char func_name[PATH_MAX];
ssnprintf (func_name, sizeof (func_name), "parse_value_file(\"%s\")", path);
/* … */
parse_value_ext (func_name, …);
```

rinigus:
I did try to run it with valgrind on linux yesterday and it did so without any errors . Mystery...

src/daemon/common.c
octo: Please make this a `static` function. Also, other parts of the code use the `_internal` suffix for such functions.

src/daemon/common.h
```diff
@@ -323,6 +323,7 @@ int parse_identifier (char *str, char **ret_host,
 		char **ret_type, char **ret_type_instance);
 int parse_identifier_vl (const char *str, value_list_t *vl);
 int parse_value (const char *value, value_t *ret_value, int ds_type);
+int parse_value_ext (const char *value, value_t *ret_value, int ds_type, const char *error_identifier);
```
octo: Remove this internal function from the header file please.

rinigus:
can we actually trigger it to test again? I can replace empty string (last argument) with NULL and add its treatment with if, for example.

rinigus:
ceossposting, sorry. OK, maybe I should clean it up and only make changes in parse_value_file. Then I'll drop ext function and leave it as it was (although it did help for debugging). If ext function would be needed in future, we can add it later. Does that sound OK?

octo:
@rinigus let's focus on the review comments and hope that Valgrind's confusion is temporary ;)

octo:
> Then I'll drop ext function and leave it as it was (although it did help for debugging). If ext function would be needed in future, we can add it later. Does that sound OK?

Sounds great, thank you!

rinigus:
@octo: please don't consider it yet - I'd like to let the device run with these changes during a night for final testing. It should be all fine, but just to be on a safe side. I'll let you know when its ready for review again.

rinigus:
@octo: all went fine - no messages in logs and entropy recorded. Please consider this PR for review.

Best wishes,

rinigus

octo:
LGTM, thanks @rinigus!

rinigus:
Thank you! 

PS: Thank you for this brilliant implementation of battery_statefs backend!

PR #1958, opened 2016-09-23
commom: Fixed AIX doesn't have MSG_DONTWAIT

cyang16:
AIX doesn't have MSG_DONTWAIT define as MSG_NONBLOCK in common.c
This fix is similar to MSG_NONBLOCK in src/pinba.c and src/statsd.c


octo:
Thanks @cyang16!

PR #1959, opened 2016-09-24
RRDCacheD plugin: Try to reconnect upon failed operations.

tokkee:


tokkee:
I don't believe the test failures are related to this PR.

tokkee:
Ftr, oetiker/rrdtool-1.x#742 will further improve this. The rrdcached plugin should imho be rewritten to use the new API to allow multiple connections. The code introduce here would still be required to support older versions of RRDtool, though.

rubenk:
@tokkee did you ever got around to improve the collectd side of things? I bumped into this in https://github.com/collectd/collectd/issues/3067.

PR #1969, opened 2016-09-30
Nginx multihost support plugin

chindyasov:
Add a new nginx_multihost plugin using an original nginx plugin codebase. The new nginx_multihost plugin is able to collect statistics from multiple nginx hosts.


octo:
Hi @chindyasov,

thank you very much for your changes! I don't think adding an additional plugin is the best way to go though. Could you please add your changes to the existing *nginx plugin*? You'll need a little extra code to add a backwards compatibility so old configurations don't break. You can take a look at the *memcached plugin* which does this already.

Thanks and best regards,
—octo

rpv-tomsk:
Hi!

How do you use this in your setup? 
Please describe your network schema - where is collectd server located and where is your nginx instances?

Have you considered the use case when Collectd hub will poll multiple Nginx on multiple remote machines?
Your plugin does not support this currently - an config option to set the host field of reported identifiers is required for this. 
(Please look into #905) 

The value from `<Host SOMENAME>` is used as `plugin_instance` in your code. 
Maybe an `<Instance SOMENAME>` option name would be more appropriate? Currently this name is confusing as it does not affects `vl.host`.

And #1609 is also related to `nginx` plugin, this might be interested for you.

chindyasov:
Hi! I'm sorry for a delay.

@octo,

thank you for your feedback. I am working on this task.

@rpv-tomsk,

Both nginx servers are configured to return status information using `ngx_http_stub_status_module`. I was testing plugin in this network configuration:

- collectd machine;
- nginx server 1 (IP is 192.168.0.85);
- nginx server 2 (IP is 192.168.0.86).

A collectd configuration for my case looks like:

```
<Plugin nginx_multihost>
    <Host server1>
        URL "http://192.168.0.85/status"
    </Host>
    <Host server2>
        URL "http://192.168.0.86/status"
    </Host>
</Plugin>
```

rpv-tomsk:
>     collectd machine;
>     nginx server 1 (IP is 192.168.0.85);
>     nginx server 2 (IP is 192.168.0.86).

In that case nginx metrics should be reported with their own hostnames for each server, not with `hostname_g` of Collectd host.

chindyasov:
@rpv-tomsk, it is good point. I will fix it too.

jsainzvela:
Hi, is this PR still alive? It would be very useful in our use case too.

Thanks!

danwerner:
I agree; in our company, we also need to collect stats from different nginx instances (running on different ports on the same host).

gcalderhead:
Hey, I need this functionality so I've

a. brought it upto date
b. merged the changes back to the main nginx plugin
c. made the config backward compatible - supports old style (for backward compatibility) and multiple host style.
d. updated the doccos.

It's all based on changes from @chindyasov 

Changes are here:  https://github.com/gcalderhead/collectd/commit/7c084ad538c30f8d5937cc305048d767f4d6efb3

Can we take it forward - either
a. new pull request
b. Maxim updates his PR
c. open to suggestions?

chindyasov:
Hello! I'm sorry, I did not have the opportunity to work on a project up to this day. It is good if @gcalderhead finalize this PR with his changes.

gcalderhead:
Hi @octo and @rpv-tomsk 

I think I've updated chindyasov's plugin changes to incorporate all the feedback above.
+ changes to the nginx plugin, rather than a new plugin
+ supports original config or using multiple <Instance> 
+ Instance name / hostname are distinct things.
+ Hostname attribute can be set per instance (defaults to hostname_g if not set)
+ Documentation updated.

You'll find the changes here:
https://github.com/collectd/collectd/compare/master...gcalderhead:update_pr1969_multiple_nginx_instances

Feedback?

dago:
@gcalderhead I created PR #3389 to track your changes and bring this forward.

PR #1970, opened 2016-09-30
RDT monitoring plugin

serhiyx:
The **rdtmon** plugin collects information provided by monitoring features of Intel Resource Director Technology (Intel(R) RDT) like Cache Monitoring Technology (CMT), Memory Bandwidth Monitoring (MBM). CMT monitors last level cache occupancy (LLC). MBM supports two types of events reporting local and remote memory bandwidth. Local memory bandwidth (MBL) reports the bandwidth of accessing memory associated with the local socket. Remote memory bandwidth (MBR) reports the bandwidth of accessing the remote socket. Also this technology allows to monitor instructions per clock (IPC).
All events are reported on a per core basis. Monitoring of the events can be configured for group of cores.
Plugin uses PQoS library for Intel(R) RDT. <https://github.com/01org/intel-cmt-cat>

### DIFF
--- a/src/rdtmon.c
+++ b/src/rdtmon.c
@@ -1,0 +1,694 @@
+/**
+ * collectd - src/rdtmon.c
+ *
+ * Copyright(c) 2016 Intel Corporation. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+ * of the Software, and to permit persons to whom the Software is furnished to do
+ * so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ * Authors:
+ *   Serhiy Pshyk <serhiyx.pshyk@intel.com>
+ **/
+
+#include <pqos.h>
+
+#include "common.h"
+
+#define RDTMON_PLUGIN "rdtmon"
+
+#define RDTMON_MAX_SOCKETS 8
+#define RDTMON_MAX_SOCKET_CORES 64
+#define RDTMON_MAX_CORES (RDTMON_MAX_SOCKET_CORES * RDTMON_MAX_SOCKETS)
+
+struct rdtmon_core_group_s {
+  char *desc;
+  int num_cores;
+  unsigned *cores;
+  enum pqos_mon_event events;
+};
+typedef struct rdtmon_core_group_s rdtmon_core_group_t;
+
+struct rdtmon_ctx_s {
+  rdtmon_core_group_t cgroups[RDTMON_MAX_CORES];
+  struct pqos_mon_data *pgroups[RDTMON_MAX_CORES];
+  int num_groups;
+  const struct pqos_cpuinfo *pqos_cpu;
+  const struct pqos_cap *pqos_cap;
+  const struct pqos_capability *cap_mon;
+};
+typedef struct rdtmon_ctx_s rdtmon_ctx_t;
+
+static rdtmon_ctx_t *g_rdtmon = NULL;
+
+static int isdup(const uint64_t *nums, unsigned size, uint64_t val) {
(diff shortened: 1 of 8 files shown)


src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+ *   `groups'      Table of core groups to set values in.
+ *   `max'         Maximum number of core groups allowed.
+ *
+ * RETURN VALUE
+ *   On success, the number of core groups set up. On error, appropriate
+ *   negative error value.
+ */
+static int oconfig_to_cgroups(oconfig_item_t *item, rdtmon_core_group_t *groups,
+                              unsigned max) {
+  int ret;
```
octo: Please declare all variables that are not accessed outside of the `for` loop, i.e. all but `index`, inside the for loop.

src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+  assert(item != NULL);
+
+  for (int j = 0; j < item->values_num; j++) {
+    if (item->values[j].value.string != NULL &&
+        strlen(item->values[j].value.string)) {
+      char *desc = NULL;
+
+      sstrncpy(value, item->values[j].value.string, sizeof(value));
+
+      memset(cores, 0, sizeof(cores));
```
octo: Please declare and initialize `cores` inside the `for` loop:

```c
uint64_t cores[RDTMON_MAX_CORES] = {0};
```

src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+
+      n = strlisttonums(value, cores, RDTMON_MAX_CORES);
```
octo: Please use `STATIC_ARRAY_SIZE(cores)` instead of `RDTMON_MAX_CORES`.

src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+}
+
+static int cgroup_set(rdtmon_core_group_t *cg, char *desc, uint64_t *cores,
+                      int num_cores) {
+  assert(cg != NULL);
+  assert(desc != NULL);
+  assert(cores != NULL);
+  assert(num_cores > 0);
+
+  cg->cores = malloc(sizeof(unsigned) * num_cores);
```
octo: Please use `calloc(num_cores, sizeof (*cg->cores))`

src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+  if (cg->cores == NULL) {
+    ERROR(RDTMON_PLUGIN ": Error allocating core group table");
+    return (-ENOMEM);
+  }
+  cg->num_cores = num_cores;
+  cg->desc = desc;
```
octo: Please `strdup()` desc here and not in the calling functions.

src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+
+  for (int i = 0; i < num_cores; i++)
+    cg->cores[i] = (unsigned)cores[i];
```
octo: `(rdtmon_core_group_t*)->cores` is of type `unsigned`, all other places appear to use `uint64_t`. Please change either the callers of this function or the struct to unify this.

src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+  unsigned n, index = 0;
+  uint64_t cores[RDTMON_MAX_CORES];
+  char value[DATA_MAX_NAME_LEN];
+
+  assert(groups != NULL);
+  assert(max > 0);
+  assert(item != NULL);
+
+  for (int j = 0; j < item->values_num; j++) {
+    if (item->values[j].value.string != NULL &&
```
octo: Please invert this logic to make the loop more readable:

```c
if ((item->values[j].value.string == NULL) || (strlen(item->values[j].value.string) == 0))
  continue;
/* rest of for-loop */
```

src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+  int num_groups;
+  const struct pqos_cpuinfo *pqos_cpu;
+  const struct pqos_cap *pqos_cap;
+  const struct pqos_capability *cap_mon;
+};
+typedef struct rdtmon_ctx_s rdtmon_ctx_t;
+
+static rdtmon_ctx_t *g_rdtmon = NULL;
+
+static int isdup(const uint64_t *nums, unsigned size, uint64_t val) {
```
octo: Please use `size_t` for array sizes and indexes.

src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+ *   `s'         String representing list of unsigned numbers.
+ *   `nums'      Array to put converted numeric values into.
+ *   `max'       Maximum number of elements that nums can accommodate.
+ *
+ * RETURN VALUE
+ *    Number of elements placed into nums.
+ */
+static unsigned strlisttonums(char *s, uint64_t *nums, unsigned max) {
+  int ret;
+  unsigned index = 0;
```
octo: Please use `size_t` for array sizes and indexes.

src/rdtmon.c
```diff
@@ -0,0 +1,694 @@
...
+    if (p != NULL) {
+      uint64_t n, start, end;
+      *p = '\0';
+      ret = strtouint64(token, &start);
+      if (ret < 0)
+        return (0);
+      ret = strtouint64(p + 1, &end);
+      if (ret < 0)
+        return (0);
+      if (start > end) {
```
octo: Is `8-4` really a syntax you want to support? Please consider returning an error instead.

serhiyx:
hi @octo, thanks for your feedback. I've updated PR with fixes for all review comments.

octo:
Hi @serhiyx, thank you very much for your update! The code looks good to me.

One suggestion: the "mon" in "rdtmon" doesn't add much information. Is this named after something that users of this technology are already familiar with, such as a command line tool exposing the same metrics? If now, what do you think about renaming this plugin to "intel_rdt"?

Best regards,
—octo

serhiyx:
Hi @octo ,

"intel_rdt" is good name for this plugin. Will implement the changes and update the review.

Regards,
  Serhiy

octo:
Thanks @serhiyx!

PR #1971, opened 2016-09-30
OVS events plugin

vmytnykx:
The OVS events plugin monitors the link status of OVS connected interfaces, dispatches the values to collectd and send the notification whenever the link state change occurs in OVS DB.

The plugin consists of two parts:
1. OVS events plugin itself (ovs_events.c, uses OVS utils);
2. OVS utils (utils_ovs.c, utils_ovs.h). It provides API for the OVS events plugin (and future OVS plugin(s)) to subscribe to specific OVS DB table update event(s) and API to poll table data directly from OVS DB. The module implements part of the OVS DB communication routine specified by RFC7047 (https://tools.ietf.org/html/rfc7047). It includes:
- Connecting/disconnecting to/from OVS DB (via TCP/UNIX);
- Mechanism to subscribe to OVS DB table events like init/insert/modify/delete table rows;
- API to send custom request (e.g. poll table data) and receive result;
- Recovery connection mechanism with OVS DB;
- Handling of ECHO request to verify the liveness of a database connection;
- The OVS DB connection lost/establish notification;
- Helpers functions.


vmytnykx:
Hi  @octo ,

The new OVS plugin depends on YAJL library which is already integrated into collectd. The plugin uses yajl_tree.h API to easily parse JSON string. Unfortunately, some of CI builds are failed because of the following error:

`utils_ovs.h:69:28: fatal error: yajl/yajl_tree.h: No such file or directory`

Looks like the problem occurs because of old YAJL lib version pre-installed on some of CI builds. The "yajl_tree.h" API are missing in YAJL v1.x but have been added in YAJL 2.x version (~5 years ago).

Failing CI builds:

* pull-requests-build-on-precise-amd64-with-default-toolchain
* pull-requests-build-on-precise-i386-with-default-toolchain
* pull-requests-build-on-epel6-i386-with-default-toolchain
* pull-requests-build-on-epel6-x86_64-with-default-toolchain

Is the YAJL lib version 1.x expected for builds above? Should it be changed to use more recent version of YAJL?

Thanks and Regards,
Volodymyr

octo:
Hi @vmytnykx,

thank you very much for your patch! Precise and Wheezy often come with quite old library versions and should still work. In this case, the plugin should be disabled by the *configure* script. You can add something along these lines to `configure.ac`:

```
with_libyajl2="no"
# …
AC_CHECK_HEADERS(yajl/yajl_tree.h,
                 [with_libyajl2="yes"],
                 [with_libyajl2="no (yajl/yajl_tree.h not found)"])
# …
AC_CHECK_LIB(yajl, yajl_tree_parse,
             [with_libyajl2="yes"],
             [with_libyajl2="no (Symbol 'yajl_tree_parse' not found)"])
```

Then use the `with_libyajl2` variable to control whether to build your new plugin.

Best regards,
—octo

src/utils_ovs.c
```diff
@@ -0,0 +1,1294 @@
...
+    goto failure;
+
+  /* parse url depending on domain */
+  if ((nexttok = strtok_r(NULL, ":", &saveptr)) != NULL) {
+    if (tmp_conn.domain == AF_UNIX) {
+      /* <UNIX-NAME> */
+      tmp_conn.addr.s_inet.sin_family = AF_UNIX;
+      sstrncpy(tmp_conn.addr.s_unix.sun_path, nexttok, strlen(nexttok) + 1);
+    } else {
+      /* <IP:PORT> */
```
octo: This doesn't handle IPv6.

I'd recommend to implement a config option that takes either one or two arguments. The first argument is the address, the second argument an optional port, for example:

```
Address "127.0.0.1" "6640"
Address "2001:DB8::c011:ec7d" "6640"
Address "2001:DB8::c011:ec7d" "service-name"
Address "unix:/path/to/socket"
```

You can then easily detect UNIX sockets: `strncmp (arg1, "unix:", strlen("unix:")) == 0`

Otherwise, call `getaddrinfo()` to get a `struct sockaddr`: `getaddrinfo(arg1, arg2, hints, &res)`

src/collectd.conf.pod
```diff
@@ -5360,6 +5360,69 @@ refer to them from.
...
+   OvsDbServerUrl "tcp:127.0.0.1:6640"
+   Interfaces "br0" "veth0"
+   SendNotification false
+ </Plugin>
+
+The plugin provides the following configuration options:
+
+=over 4
+
+=item B<OvsDbServerUrl> I<server>
```
octo: Please consider turning this into an option with two values, "node" and "service" (aka. "host" and "port"). That makes it much easier to support IPv6. See also my comments below.

vmytnykx:
Hi @octo ,

Thank you for reviewing the plugin implementation and your comments.

I've fixed all your comments and re-implement plugin configuration based on your suggestion. Also, IPv6 support has been added (automatically) and verified.

Thanks and Regards,
Volodymyr

src/ovs_events.c
```diff
@@ -0,0 +1,629 @@
...
+    }
+    if (ovs_events_config_iface_exists(ifinfo.name) != 0)
+      /* dispatch notification */
+      ovs_events_dispatch_notification(&ifinfo);
+  }
+}
+
+/* OVD DB reply callback. It parses reply, receives
+ * interface information and dispatches the info to
+ * collecd
```
maryamtahhan: typo

src/ovs_events.c
```diff
@@ -0,0 +1,629 @@
...
+      memset(&ifinfo, 0, sizeof(ifinfo));
+      if (ovs_events_get_iface_info(YAJL_GET_ARRAY(jvalue)->values[j],
+                                    &ifinfo) < 0) {
+        ERROR(OVS_EVENTS_PLUGIN
+              "unexpected interface information data received");
+        return;
+      }
+      DEBUG("name=%s, uuid=%s, ext_iface_id=%s, ext_vm_uuid=%s", ifinfo.name,
+            ifinfo.uuid, ifinfo.ext_iface_id, ifinfo.ext_vm_uuid);
+      ovs_events_link_status_submit(&ifinfo);
```
maryamtahhan: NVM - this is for values from tables and not events...

src/utils_ovs.c
```diff
@@ -0,0 +1,1287 @@
...
+    sstrncpy(pdb->service, service, sizeof(pdb->service));
+
+  /* setup OVS DB callbacks */
+  if (cb)
+    pdb->cb = *cb;
+
+  /* prepare event thread */
+  pthread_cond_init(&pdb->event_thread.cond, NULL);
+  pthread_mutex_init(&pdb->event_thread.mutex, NULL);
+  pthread_mutex_lock(&pdb->event_thread.mutex);
```
octo: Is this mutex held on purpose when returning from this function? If so, please add this to this function's documentation.

src/utils_ovs.c
```diff
@@ -0,0 +1,1287 @@
...
+  if (ovs_db_poll_thread_stop(pdb) < 0) {
+    OVS_ERROR("stop event thread failed");
+    ovs_db_ret = (-1);
+  }
+
+  /* unsubscribe callbacks */
+  ovs_db_callback_remove_all(pdb);
+
+  /* close connection */
+  if (pdb->sock)
```
octo: Please check for `pdb->sock >= 0` – zero is a valid file descriptor (and wouldn't match) while -1 is invalid (and would match).

src/utils_ovs.h
```diff
@@ -0,0 +1,234 @@
...
+
+/* OVS DB defines */
+#define OVS_DB_ADDR_NODE_SIZE 256
+#define OVS_DB_ADDR_SERVICE_SIZE 128
+
+/* OVS DB prototypes */
+
+/*
+ * NAME
+ *   ovs_db_init
```
octo: Thank you for the great inline documentation! :-D

src/utils_ovs.h
```diff
@@ -0,0 +1,234 @@
...
+ *   YAJL value upon success or NULL if key not found.
+ */
+yajl_val ovs_utils_get_value_by_key(yajl_val jval, const char *key);
+
+/*
+ * NAME
+ *   ovs_utils_get_map_value
+ *
+ * DESCRIPTION
+ *   Get OVS DB map value by given map key (rfc7047, "Notation" section).
```
octo: Some example input might help explain what this does. It appears the input should have this schema:

```json
[
  "map",
  [
    [ "<key 1>", "<value 1>" ],
    …
    [ "<key n>", "<value n>" ]
  ]
]
```

src/ovs_events.c
```diff
@@ -0,0 +1,629 @@
...
+ * -1 is returned
+ */
+static int ovs_events_config_iface_exists(const char *ifname) {
+  if (ovs_events_ctx.config.ifaces == NULL)
+    return -1;
+
+  /* check if given interface exists */
+  for (ovs_events_iface_list_t *iface = ovs_events_ctx.config.ifaces; iface;
+       iface = iface->next)
+    return (strcmp(ifname, iface->name) == 0);
```
octo: This will only every check the first interface in the list, which I don't think is the intention here. This should probably be:

```c
if (strcmp(ifname, iface->name) == 0)
  return 1;
```

src/ovs_events.c
```diff
@@ -0,0 +1,629 @@
...
+  /* setup OVS DB interface condition */
+  for (ovs_events_iface_list_t *iface = ovs_events_ctx.config.ifaces; iface;
+       iface = iface->next, offset += ret) {
+    /* allocate new buffer (format size + ifname len is good enough) */
+    buff_size += (sizeof(option_fmt) + strlen(iface->name));
+    new_buff = realloc(buff, buff_size);
+    if (new_buff == NULL)
+      goto failure;
+    buff = new_buff;
+    ret = ssnprintf(buff + offset, buff_size, option_fmt, iface->name);
```
octo: The available buffer size is only `buff_size - offset`.

src/ovs_events.c
```diff
@@ -0,0 +1,629 @@
...
+      goto failure;
+  }
+  /* if no interfaces are configured, use default params */
+  if (buff == NULL) {
+    buff = strdup(default_opt);
+    offset = strlen(default_opt);
+  }
+
+  /* allocate memory for OVS DB select params */
+  buff_size = offset + sizeof(params_fmt);
```
octo: It would be more consistent with the above code to do:

```c
buff_size = sizeof(params_fmt) + strlen(buff);
```

octo: I'm not a huge fan of re-using variables like this. Please consider using `char *params` and `size_t params_size` instead.

src/ovs_events.c
```diff
@@ -0,0 +1,629 @@
...
+#include "common.h" /* auxiliary functions */
+
+#include "utils_ovs.h" /* OVS helpers */
+
+#define OVS_EVENTS_IFACE_NAME_SIZE 128
+#define OVS_EVENTS_IFACE_UUID_SIZE 64
+#define OVS_EVENTS_EXT_IFACE_ID_SIZE 64
+#define OVS_EVENTS_EXT_VM_UUID_SIZE 64
+#define OVS_EVENTS_PLUGIN "ovs_events"
+#define OVS_EVENTS_CTX_LOCK                                                    \
```
octo: Clever, I love it :)

src/ovs_events.c
```diff
@@ -0,0 +1,629 @@
...
+  ovs_events_iface_list_t *new_iface;
+
+  for (int i = 0; i < ci->children_num; i++) {
+    oconfig_item_t *child = ci->children + i;
+    if (strcasecmp("SendNotification", child->key) == 0) {
+      if (cf_util_get_boolean(child, &ovs_events_ctx.config.send_notification) <
+          0)
+        OVS_EVENTS_CONFIG_ERROR(child->key);
+    } else if (strcasecmp("OvsDbAddress", child->key) == 0) {
+      if (child->values_num < 1) {
```
octo: Please use `cf_util_get_string_buffer()` here, which will do all necessary checks and error reporting:

```c
cf_util_get_string_buffer(child, ovs_events_ctx.config.ovs_db_node,
    sizeof(ovs_events_ctx.config.ovs_db_node));
```

src/utils_ovs.c
```diff
@@ -0,0 +1,1287 @@
...
+
+  /* make a request to subscribe to given table */
+  OVS_YAJL_CALL(yajl_gen_get_buf, jgen, (const unsigned char **)&params,
+                &params_len);
+  if (ovs_db_send_request(pdb, "monitor", params, result_cb) < 0) {
+    OVS_ERROR("Failed to subscribe to \"%s\" table", tb_name);
+    ovs_db_ret = (-1);
+  }
+
+yajl_gen_failure:
```
vmytnykx: this label is used by OVS_YAJL_CALL() call

src/ovs_events.c
vmytnykx: Thank for that :)

vmytnykx:
Hi @octo,

Thank you for the addition review comments. I've tried to address all of them, including
- configuration update (UNIX socket and IP address configuration are separated now, which looks event better ) )
- memory leaks, critical sections, init/destroy API improvement
- others comments.

Also, I've replied to some of your comments which seems don't require any changes...

Best Regards,
Volodymyr

src/collectd.conf.pod
```diff
@@ -5360,6 +5360,68 @@ refer to them from.
 
 =back
 
+=head2 Plugin C<ovs_events>
+
+The I<ovs_events> plugin monitors the link status of OVS connected interfaces,
+dispatches the values to collectd and send the notification whenever the link
```
rubenk: s/send/sends/

src/collectd.conf.pod
```diff
@@ -5360,6 +5360,68 @@ refer to them from.
...
+   SendNotification false
+ </Plugin>
+
+The plugin provides the following configuration options:
+
+=over 4
+
+=item B<Address> I<node>
+
+The address of OVS DB server JSON-RPC interface used by the plugin. To enable
```
rubenk: The address of the OVS DB server JSON-RPC interface

src/collectd.conf.pod
```diff
@@ -5360,6 +5360,68 @@ refer to them from.
...
+=item B<Socket> I<path>
+
+The UNIX domain socket path of OVS DB server JSON-RPC interface used by the
+plugin. To enable the interface, the OVS DB daemon should be running with
+'--remote=punix:' option. See L<ovsdb-server(1)> for more details. If this
+option is set, B<Address> and B<Port> options are ignored.
+
+=item B<Interfaces> [I<ifname> ...]
+
+List of interface names to be monitored by this plugin. If this option is missed
```
rubenk: is not specified or is empty

src/ovs_events.c
```diff
@@ -0,0 +1,631 @@
...
+ * Private variables
+ */
+static ovs_events_ctx_t ovs_events_ctx = {
+    .mutex = PTHREAD_MUTEX_INITIALIZER,
+    .config = {.send_notification = 0,     /* do not send notification */
+               .ovs_db_node = "localhost", /* use default OVS DB node */
+               .ovs_db_serv = "6640",      /* use default OVS DB service */
+               .ovs_db_unix = "",          /* UNIX path empty by default */
+               .ifaces = NULL},
+    .ovs_db_select_params = NULL,
```
rubenk: there's no need to initialize static variables to their zero value.

src/ovs_events.c
```diff
@@ -0,0 +1,631 @@
...
+  int ret = 0;
+  size_t buff_size = 0;
+  size_t buff_off = 0;
+  char *opt_buff = NULL;
+  const char params_fmt[] = "[\"Open_vSwitch\"%s]";
+  const char option_fmt[] = ",{\"op\":\"select\",\"table\":\"Interface\","
+                            "\"where\":[[\"name\",\"==\",\"%s\"]],"
+                            "\"columns\":[\"link_state\",\"external_ids\","
+                            "\"name\",\"_uuid\"]}";
+  const char default_opt[] = ",{\"op\":\"select\",\"table\":\"Interface\","
```
rubenk: can you make these static?

src/ovs_events.c
```diff
@@ -0,0 +1,631 @@
...
+    buff_size += (sizeof(option_fmt) + strlen(iface->name));
+    char *new_buff = realloc(opt_buff, buff_size);
+    if (new_buff == NULL) {
+      sfree(opt_buff);
+      return NULL;
+    }
+    opt_buff = new_buff;
+    ret = ssnprintf(opt_buff + buff_off, buff_size - buff_off, option_fmt,
+                    iface->name);
+    if (ret < 0) {
```
rubenk: int ret

src/ovs_events.c
```diff
@@ -0,0 +1,631 @@
...
+      sfree(opt_buff);
+      return NULL;
+    }
+  }
+  /* if no interfaces are configured, use default params */
+  if (opt_buff == NULL)
+    opt_buff = strdup(default_opt);
```
rubenk: please check for failure.

src/ovs_events.c
```diff
@@ -0,0 +1,631 @@
...
+  sstrncpy(n.host, hostname_g, sizeof(n.host));
+  sstrncpy(n.plugin_instance, ifinfo->name, sizeof(n.plugin_instance));
+  sstrncpy(n.type, "gauge", sizeof(n.type));
+  sstrncpy(n.type_instance, "link_status", sizeof(n.type_instance));
+  plugin_dispatch_notification(&n);
+}
+
+/* Dispatch OVS interface link status value to collectd */
+static void
+ovs_events_link_status_submit(const ovs_events_iface_info_t *ifinfo) {
```
rubenk: please combine these two lines.

src/ovs_events.c
```diff
@@ -0,0 +1,631 @@
...
+ * to receive link status event(s).
+ */
+static void ovs_events_conn_initialize(ovs_db_t *pdb) {
+  int ret = 0;
+  const char tb_name[] = "Interface";
+  const char *columns[] = {"_uuid", "external_ids", "name", "link_state", NULL};
+
+  /* register update link status event if needed */
+  if (ovs_events_ctx.config.send_notification) {
+    ret = ovs_db_table_cb_register(pdb, tb_name, columns,
```
rubenk: int ret.

src/ovs_events.c
```diff
@@ -0,0 +1,631 @@
...
+  return (0);
+
+failure:
+  ovs_events_config_free();
+  return (-1);
+}
+
+/* Dispatch OVS interface link status event to collectd */
+static void
+ovs_events_dispatch_notification(const ovs_events_iface_info_t *ifinfo) {
```
vmytnykx: This is done by clang-format. I've addressed it anyway but I suppose it will be changed back during next clan-format running :)

vmytnykx:
Hi @rubenk, last commit addresses your comments. Thank you.

README
```diff
@@ -278,6 +278,15 @@ Features
...
       Query data from an Oracle database.
 
+    - ovs_events
+      The plugin monitors the link status of OVS connected interfaces,
+      dispatches the values to collectd and send the notification whenever
+      the link state change occurs in OVS DB. It requires YAJL library to be
+      installed.
+      Detailed instructions for installing and setting up Open vSwitch, see
+      OVS documentation.
+      <http://openvswitch.org/support/dist-docs/INSTALL.md.html>
```
rubenk: Heh, that link gives a 404. Should this point to https://github.com/openvswitch/ovs/blob/master/INSTALL.rst instead?

vmytnykx: Yeah, you are right! OvS has changed it (*.md -> *.rst, ) after the PR has been created. Thank you for discovering this :) I will fix it to the right one: http://openvswitch.org/support/dist-docs/INSTALL.rst.html.

rubenk:
Thanks @vmytnykx. Personally I think this is good to be merged. @octo did you have any further comments?

vmytnykx:
Hi @octo,

Do you have any farther comments regarding the changes? Sorry to ping you again but we have already implemented new plugin that rely on OVS utils API (utils_ovs.[ch]). Even, if we create new PR with the new changes based on that branch, it will also include #1971 changes, which will confuse the review process (also, rebasing will be required, etc., once current changes are merged). If you have some idea how to share the new code in this case, that would be appreciated too.

FYI, new changes: https://github.com/maryamtahhan/collectd/compare/feat_ovs_link...maryamtahhan:feat_ovs_stats?expand=1

Thanks and Regards,
Volodymyr

vmytnykx:
Thank you @octo for your feedbacks.

@rubenk, I've addressed the published comments from octo in my last commit. Thank you.

Regards,
Volodymyr

vmytnykx:
Hi @rubenk 

Looks like the branch has conflicts which must be resolved. Can I rebase the branch with latest master?

Regards,
 Volodymyr

rubenk:
Yes please.

vmytnykx:
Rebase done

maryamtahhan:
Hi @octo @rubenk 
If there are no more review comments, can this be merged into master? we would like to enable this plugin in collectd in OPNFV and need it on the master branch
Thanks in advance
BR
Maryam

vmytnykx:
Hi @rubenk ,

I've done re-base second time because of the recent changes into build system (src/Makefile.am -> Makefile.am). So, now conflicts are resolved.

Since, there are no more comments, can this be merged to avoid further re-basing ?

Thanks & Regards,
 Volodymyr

vmytnykx:
Hi, @rubenk , the `epel5` builds are failed because of mcelog error:
```
src/mcelog.c: In function 'socket_reinit':
src/mcelog.c:163: error: 'SOCK_CLOEXEC' undeclared (first use in this function)
src/mcelog.c:163: error: (Each undeclared identifier is reported only once
src/mcelog.c:163: error: for each function it appears in.)
src/mcelog.c:163: error: 'SOCK_NONBLOCK' undeclared (first use in this function)
make[1]: *** [src/mcelog.lo] Error 1
```
This failure isn't related to OVS changes. Based on this [comment](https://github.com/collectd/collectd/pull/2003#issuecomment-269134735) looks like it is expected, right?

Thanks,
 Volodymyr

rubenk:
Dear Volodymyr, yes, we've dropped support for EPEL5 but still need to adjust the CI.

Now merged, thanks for all your work!

PR #1975, opened 2016-10-04
memcached: Added Address option

rpv-tomsk:
Added new option "Address".

We use separate Memcached instances per project on single server.
Each project has its own IP in 127.0.0.0/8 and each Apache or Memcached instance binds to appropriate IP on standard 80 and 11211 ports. Check for "localhost" or "127.0.0.1", as it now implemented in `memcached` plugin, is not suitable for us. Our internal patch was reworked to be applied to current `master` branch and proposed to review and merge.

Superseedes: #905

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -705,7 +705,8 @@
 
 #<Plugin memcached>
 #	<Instance "local">
-#		Host "127.0.0.1"
+#		#Host "memcache.example.com"
+#		Address "127.0.0.1"
 #		Port "11211"
 #	</Instance>
 #</Plugin>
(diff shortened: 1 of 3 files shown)


octo:
Thank you @rpv-tomsk!

Calygos:
I just found this change after trying to determine why my Memcached metrics (six instances per server) had changed to use 127.0.0.1 after a recent OS upgrade. The documentation needs to be updated to include the Address option (FWIW, we're now running Ubuntu 16.04 with Collectd 5.5.1).

One small complaint I have about this change is that our memcached.conf files used to be identical across the servers but now has to be different on every server since they now need the hostname included (multiple times). This is easily handled via templating but it's still a forced change to a file that had previously been working fine.

PR #1976, opened 2016-10-04
write_graphite: remove linking against libyajl

mfournier:
write_graphite doesn't have anything to do with json. This seems to have
been accidentally added in 30c1111.

Fixes https://bugs.debian.org/839771

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1251,7 +1251,6 @@ pkglib_LTLIBRARIES += write_graphite.la
 write_graphite_la_SOURCES = write_graphite.c \
                         utils_format_graphite.c utils_format_graphite.h
 write_graphite_la_LDFLAGS = $(PLUGIN_LDFLAGS)
-write_graphite_la_LIBADD = libformat_json.la
 endif
 
 if BUILD_PLUGIN_WRITE_HTTP


mfournier:
FWIW, with this patch we get:
```
$ /opt/collectd/sbin/collectd  -T
option = ReadThreads; value = -1;
Created new plugin context.
plugin_load: plugin "cpu" successfully loaded.
plugin_load: plugin "write_graphite" successfully loaded.
Done parsing `/opt/collectd/share/collectd/types.db'
interval_g = 10.000;
timeout_g = 2;
hostname_g = lonquimay.wrk.lsn.camptocamp.com;
Exiting normally.
collectd: Stopping 5 write threads.
plugin: stop_write_threads: Signalling `write_cond'
write_graphite plugin: Connecting to localhost:2003 via tcp failed. The last error was: failed to connect to remote host: Connection refused
write_graphite plugin: wg_flush_nolock: timeout = 0.000; send_buf_fill = 0;
$ ldd /opt/collectd/lib/collectd/write_graphite.so 
        linux-vdso.so.1 (0x00007ffd3252e000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f216c469000)
        /lib64/ld-linux-x86-64.so.2 (0x00005572c0416000)
```

octo:
LGTM

The *write_graphite plugin* was building with `utils_format_json.c` for a long time, though. It appears that was added in 2f6cab9349c3ff8331590a7ca563cc6168f5dbf9 back in 2012 …

PR #1980, opened 2016-10-10
processes plugin: Do not gather processes IO and context switch data when unneeded

rpv-tomsk:
Processes IO and context switch values are not used when only summary is submitted.

Added new function `ps_fill_details()` as part of prepare for #836 / #1041 implementation.

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -185,9 +185,11 @@ typedef struct procstat_entry_s
 	derive_t io_wchar;
 	derive_t io_syscr;
 	derive_t io_syscw;
+	_Bool    has_io;
 
 	derive_t cswitch_vol;
 	derive_t cswitch_invol;
+	_Bool    has_cswitch;
 
 	struct procstat_entry_s *next;
 } procstat_entry_t;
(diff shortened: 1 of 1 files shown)


octo:
Hi @rpv-tomsk, thanks for the patch! I don't think the title is quite fitting, though: `ps_fill_details()` is called unconditionally – there is no change from the previous behavior as far as I can see … What did I miss?

Best regards,
—octo

rpv-tomsk:
Hi!

The `ps_fill_details()`  called from `ps_list_add()` after `ps_list_match()` check.

Before this patch processes IO and context switch values was gathered by `ps_read_process()`.

rpv-tomsk:
Ouch, I'm overoptimized my patch.

In current state, `ps_fill_details` does not check if it was called before for this entry.
There was flag for this check at prev stage, but I removed it right before make PR.

rpv-tomsk:
Patch updated.

octo:
> The `ps_fill_details()` is  called *[…]* after `ps_list_match()` check.

That makes sense, thanks for the explanation. Under which circumstances could `ps_fill_details()` be called more than once?

rpv-tomsk:
> Under which circumstances could ps_fill_details() be called more than once?

When single `entry` / `cmdline` will match some (more than one)  `procstat_t` in `list_head_g` chain.

octo:
Thanks @rpv-tomsk!

PR #1981, opened 2016-10-10
Processes plugin: improvements / fixes

rpv-tomsk:
* Remove unused fields from structures

Fields cpu_user, cpu_system, vmem_minflt, vmem_majflt are always equal to zero for new entries (which are passed to ps_list_add()).
Values of these fields are not used in `procstat_t` entries too. So, that can be safely removed.

* Fix counters initialization / spikes

processes plugin generates spikes on (re)start. That is caused by wrong counter logic: When collectd (re)started, monitored processes, which was started before collectd, already have non-zero values in cpu_user/cpu_system/vmem_minflt/vmem_majflt, so `want_init` is false. What causes spike. 
At other hand, processes which are started between read cycles, should be fully accounted without initialization. So, we must skip only first cycle for these counters.

I think this is a bug, so, If you want, I can create PR against stable branch (please specify version).

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -170,13 +170,9 @@ typedef struct procstat_entry_s
 	unsigned long vmem_code;
 	unsigned long stack_size;
 
-	unsigned long vmem_minflt;
-	unsigned long vmem_majflt;
 	derive_t      vmem_minflt_counter;
 	derive_t      vmem_majflt_counter;
 
-	unsigned long cpu_user;
-	unsigned long cpu_system;
 	derive_t      cpu_user_counter;
 	derive_t      cpu_system_counter;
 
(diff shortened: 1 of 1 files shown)


octo:
Hi @rpv-tomsk,

thanks for simplifying that code, it surely can use it. I have to admit your description made me a bit hesitant at first: the fields you remove *are* actually written to, so they are not always zero. I think you change is still correct, because these fields were never actually *read*: The only place that references these fields is in `ps_update_counter()`, where the previous value is overwritten unconditionally before being read in the same function – a local variable serves this purpose just as well.

Best regards,
—octo

tbendiksen:
In this cleanup an extra pse.cpu_system statement was left in the AIX / elif HAVE_PROCINFO_H section. This causes the compilation to fail on AIX. Removing this line will allow the compilation to succeed. 


$ cat collectd_5.7.1_aix.patch 
--- src/processes.c	2017-01-23 08:53:57.000000000 +0100
+++ src/processes_new.c	2017-03-01 15:32:46.000000000 +0100
@@ -2127,7 +2127,6 @@
        pse.cpu_user_counter = procentry[i].pi_ru.ru_utime.tv_sec * 1000000 +
                               procentry[i].pi_ru.ru_utime.tv_usec / 1000;
  
-       pse.cpu_system = 0;
        /* tv_usec is nanosec ??? */
        pse.cpu_system_counter = procentry[i].pi_ru.ru_stime.tv_sec * 1000000 +
                                 procentry[i].pi_ru.ru_stime.tv_usec / 1000;

rpv-tomsk:
Hi @tbendiksen!

Thanks for reporting this!

PR #1985, opened 2016-10-11
Add node support to network plugin

mateusz-gozdek-sociomantic:
This PR introduce basic support for registering multiple targets of network write plugin. It includes:
- support for `<node>` element in configuration file. Old syntax is still valid and just registers the target as `network`. `<node>` element registers as `network/<node name>`
- send buffer has been moved to struct and now each of target has it's own buffer. This is required, because now it's possible to send different data to different targets.
- as plugin is using `plugin_register_complex_config`, it should use `plugin_unregister_complex_config`

I prepared it to be used together with filter plugin, to send different data to different servers using one instance of collectd. It's running without a problems in my environment, but as I am NOT a C programmer, some things might be missing/broken.


mateusz-gozdek-sociomantic:
Updated, as there was compilation problem after rebasing.

octo:
Hi @mateusz-gozdek-sociomantic,

thank you very much for your patch! The description sounds sane, but I don't have the time to do the code review right now. I hope to come back to you within a week though.

Best regards,
—octo

octo:
@mateusz-gozdek-sociomantic quick update though: could you please document the new config syntax in `src/collectd.conf.pod` and `src/collectd.conf.in`? Thanks!

mateusz-gozdek-sociomantic:
Sure, I'll try.

octo:
FYI, I pushed a small fix for the .pod file. This should make the build bots happy again …

mateusz-gozdek-sociomantic:
Thank you for your review and kind words @octo.

The `Node` option is a block, because you can put multiple servers inside. For example if you want to send specific metrics to specific group of servers.

I'll fix what you mentioned soon. Do you prefer it as one more comment or should I fix and squash (also with your commit) it?

octo:
> The `Node` option is a block, because you can put multiple servers inside.

Thanks for the clarification, I managed to totally miss that fact … In that case, how about adding a `Group` argument to the `Server` block?:

```
<Server "a.example.com">
  Group "example_servers"
</Server>
<Server "b.example.com">
  Group "example_servers"
</Server>
# The flush and write callbacks called "network/example_servers" can be used to flush / write to a.example.com _and_ b.example.com
```

It would need some refactoring of the code, but the config syntax would be much easier to grasp IMHO.

mateusz-gozdek-sociomantic:
Well, using `Group` argument will make config a bit harder to parse, as we will need to first read and store all elements somewhere and then iterate over them to select, initialise and register declared groups.
My solution is simpler, it's following the `Node` style like in `write_*` plugins and also it's extendable, because in the future, maybe someone will need to separate even more things like stats (which I find useful), TTL and rest of the settings.

src/network.c
```diff
@@ -2318,8 +2324,8 @@ static int sockent_server_listen (sockent_t *se) /* {{{ */
 	return (0);
 } /* }}} int sockent_server_listen */
 
-/* Add a sockent to the global list of sockets */
-static int sockent_add (sockent_t *se) /* {{{ */
+/* Add a sockent to the passed list of sockets */
+static int sockent_add (sockent_t *se, sockent_t **sockets) /* {{{ */
```
mateusz-gozdek-sociomantic: Fixed in 8c46ec7

src/network.c
```diff
@@ -3107,7 +3122,7 @@ static int network_config_add_listen (const oconfig_item_t *ci) /* {{{ */
     return (-1);
   }
 
-  status = sockent_add (se);
+  status = sockent_add (se, NULL);
```
mateusz-gozdek-sociomantic: Fixed in 8c46ec7

src/network.c
```diff
@@ -3309,11 +3316,179 @@ static int network_notification (const notification_t *n,
...
+static int network_config_node (oconfig_item_t *ci)
+{
+	network_callback_t *cb;
+	int i;
+
+	/* allocate memory for new callback */
+	cb = calloc(1, sizeof(network_callback_t));
+	if (cb == NULL)
+	{
+		ERROR("write_network plugin: calloc failed.");
```
mateusz-gozdek-sociomantic: Fixed in 09258d9

src/network.c
```diff
@@ -3309,11 +3316,179 @@ static int network_notification (const notification_t *n,
...
+		return(-1);
+	}
+
+	cb->sending_sockets = NULL;
+
+	/* Get callback name */
+	cf_util_get_string (ci, &cb->node);
+	if (strcasecmp ("network", cb->node) == 0) {
```
mateusz-gozdek-sociomantic: Fixed in 5b2eff8

src/network.c
```diff
@@ -3309,11 +3316,179 @@ static int network_notification (const notification_t *n,
...
+		sstrncpy(cb->name, cb->node, DATA_MAX_NAME_LEN);
```
mateusz-gozdek-sociomantic: Fixed in 5b2eff8

src/network.c
```diff
@@ -3309,11 +3316,179 @@ static int network_notification (const notification_t *n,
...
+	} else {
+		ssnprintf(cb->name, DATA_MAX_NAME_LEN, "network/%s", cb->node);
```
mateusz-gozdek-sociomantic: Fixed in 5b2eff8

rpv-tomsk:
```
<Server "a.example.com">
  Group "example_servers1"
  Group "example_servers2"
</Server>
```

Is multiple groups should/may be allowed per one Server?

octo:
I'd say "no": it would be very easy to use wrong (sending metrics twice) and you can achieve the same effect by calling the "write" target multiple times.

rpv-tomsk:
>it would be very easy to use wrong (sending metrics twice) 

Implementation should protect against this.

>and you can achieve the same effect by calling the "write" target multiple times.

Then this patch is unneded. Any can achieve `Send different data to different servers` by `calling the "write" target multiple times`

I'm missing something?

mateusz-gozdek-sociomantic:
Here is configuration example:
```
LoadPlugin network
LoadPlugin match_regex

<Plugin network>
	<Node "alerts">
		Server "y.y.y.y"
	</Node>
	<Node "monitoring">
		Server "x.x.x.x"
	</Node>
</Plugin>

# Send data from cpu and df plugins only to alerts server
<Rule "alerts data">
	<Match "regex">
		Plugin "^cpu|df$"
	</Match>
	<Target "write">
		Plugin "network/alerts"
	</Target>
	Target "stop"
</Rule>

# Ignore data from test instances
<Rule "ignore test instances"
	<Match "regex">
		PluginInstance "^test$"
	</Match>
	Target "stop"
</Rule>
```

mateusz-gozdek-sociomantic:
Ping @octo

src/collectd.conf.pod
```diff
@@ -4836,6 +4839,12 @@ traffic gets only accepted, if it arrives on the given interface.
 
 =back
 
+=item B<Node> I<Nodename>
+
+The B<Node> block identifies a new servers group. The name for node is
+a canonical identifier which is used as I<plugin instance>. It is limited to
+64E<nbsp>characters in length. You can put B<Server> entries here.
```
rubenk: This is DATA_MAX_NAME_LEN, which is 128 now.

rubenk:
I like this PR in general, but I'm not too sure about the node syntax.
If I understand correctly, you could do
```
<Node "alerts">
        Server "x.x.x.x"
        Server "y.y.y.y"
</Node>
```

But in other plugins there's always a one to one mapping between Nodes and Servers.
How about:
```
<Group "alerts">
        Server "x.x.x.x"
        Server "y.y.y.y"
</Group>
```

I can even see us extending this into the feature with load balancing features, like
```
<Group "alerts">
        Server "x.x.x.x"
        Server "y.y.y.y"
        Balance "roundrobin/consistent hashing/random/whatever"
</Group>
```

PR #1987, opened 2016-10-11
utils_format_graphite: terminate buffer with \0

maage:
ssnprintf returns message_len without \0 and then that many bytes are copied with memcpy and \0 is lost. This adds it again.


octo:
Thank you @maage!

Cherry-picked into *collectd-5.5* as dcd0b4f3121d903070282ee013b2de939bc0fa49

PR #1988, opened 2016-10-12
Added 'used' to aix memory

jquinn47:
Same changes made in #1801, but I've submitted the PR for the master branch.


src/memory.c
```diff
@@ -502,7 +502,7 @@ static int memory_read_internal (value_list_t *vl)
 	}
 
 	/* Unfortunately, the AIX documentation is not very clear on how these
-	 * numbers relate to one another. The only thing is states explcitly
+	 * numbers relate to one another. The only thing is states explicitly
```
rubenk: While you're here fixing this typo, you might as well fix the other one (s/is/it/)

jquinn47:
Just pinging this one. I've got a few more in the pipe, but I'm trying to get at one a time so I don't get lost this time around :)

octo:
Hi @jquinn47, thank you very much for your patch! One of our best practices with these kinds of "categorized" metrics is to ensure that each byte is only counted once. I'm a bit skeptical that this might be the case here: is `used` really only counting bytes that are not yet represented by `user`, `system` or one of the other metrics?

Best regards,
—octo

jquinn47:
I doubt it. The comment near the code I was modifying summarizes it nicely, but I can provide a bit more detail here.

Some brief snippets from [perfstats](http://www.ibm.com/support/knowledgecenter/ssw_aix_72/com.ibm.aix.files/libperfstat.h.htm) documentation:

Field | Description
-----|-----
`real_free`  | Free real memory (in 4 KB pages).
`numperm`  | Number of frames used for files (in 4 KB pages).
`real_system`  | Real memory used by system segments (in 4 KB pages). This value is the sum of all the used pages in segment marked for system usage.  (only an approximation)
`real_process` | Real memory used by process segments (in 4 KB pages). This is real_total - real_free - numperm - real_system.
*`real_inuse`  | Real memory which is in use (in 4 KB pages).

The `real_process` description is how the `real_total = real_free + numperm + real_system + real_process` calculation was derived.

pigeon hole principle suggests we're overlapping here :) (thanks math)

`real_inuse` is just another metric , so I'd agree with you that it overlaps with the metrics being captured currently, however, it is still a useful metric.  

According to the [svmon](http://www.ibm.com/support/knowledgecenter/ssw_aix_61/com.ibm.aix.performance/amount_mem_use.htm) utility's description, inuse is "The number of pages in RAM that are in use by a process pluss the number of persistent pages that belonged to a terminated process that are still resident in RAM... This value is the total size of memory minus the free list."

Still a useful metric, but might not be what you're willing to merge in?

jquinn47:
Any updates on this one?

jquinn47:
Just pinging for an update :) @octo

jquinn47:
@octo Any feedback available?

octo:
I think we have several options:

*   Figure out how the metrics relate to each other and create a separate metric. If I understand the description correctly, then the difference between `inuse` and `real_free + numperm + real_system + real_process` is that `inuse` also contains "pages that belonged to a terminated process that are still resident in RAM". FreeBSD has a very similar metric called *inactive*, so we could segment AIX's "free" into truly free and "free but not yet reclaimed / zeroed" metric.
*   Use a separate *plugin instance* for this metric, so it's not grouped with the other metrics.
*   Add a config option so users can chose between the "inuse" metric and what is currently reported (similar to the "PerState" / "PerCore" options of the CPU plugin).

Best regards,
—octo

jquinn47:
@octo  I'd prefer option 3, but I'm okay with implementing whichever option is most likely to be accepted back. Before I get started, which option would you prefer the most?

octo:
Hi @jquinn47,

each outlined solution would be acceptable. Personally, I'd prefer option 1, but it is only feasible if such an easy relationship as I outlined exists. Also, I'm by no means an expert for AIX, so feel free to chose / propose what you think is best.

Best regards,
—octo

jquinn47:
I went with option three after talking with @yogeswaran a bit.  I had to make a few other fixes to get this up to date with the current master:

* Process plugin had a rogue reference to cpu_system. That value does not exist in the struct
* I had to guard against `-Wall` and `-Werror` in the configure.ac.  These cause `xlc` to choke . Some other locations do the same.
* Updated the comments a bit more to elaborate on what we've discussed in this PR. Would you like me to reference the PR in the source as well?

dago:
@jquinn47 I know it has been a long time since you initially contributed this PR, but would you be willing to rebase it and resolve the conflicts? It would be nice if this could be merged into the upcoming 5.11.0.

PR #1989, opened 2016-10-12
processes plugin: Code cleanup + add per-process open files count

rpv-tomsk:
The `procstat_entry_s` used in improper way.
Added new type and simplified the code - removed unneeded assignments and values copying.

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -159,9 +159,10 @@
 #endif
 #endif
 
-typedef struct procstat_entry_s {
+#define PROCSTAT_NAME_LEN 256
+typedef struct process_entry_s {
   unsigned long id;
-  unsigned long age;
+  char name[PROCSTAT_NAME_LEN];
 
   unsigned long num_proc;
   unsigned long num_lwp;
@@ -187,11 +188,21 @@ typedef struct procstat_entry_s {
   derive_t cswitch_vol;
   derive_t cswitch_invol;
   _Bool has_cswitch;
+} process_entry_t;
+
+typedef struct procstat_entry_s {
+  unsigned long id;
+  unsigned long age;
+
+  derive_t vmem_minflt_counter;
+  derive_t vmem_majflt_counter;
+
+  derive_t cpu_user_counter;
+  derive_t cpu_system_counter;
 
   struct procstat_entry_s *next;
 } procstat_entry_t;
 
-#define PROCSTAT_NAME_LEN 256
 typedef struct procstat {
   char name[PROCSTAT_NAME_LEN];
 #if HAVE_REGEX_H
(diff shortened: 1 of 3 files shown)


rpv-tomsk:
That is too hard to make changes in this plugin. There is too much unnecessary actions.

What do you think, @octo?

rpv-tomsk:
The second patch tries to fix gaps, caused by terminated processes.
When process terminates, huge value is decreased from counter, causing negative derived value.

rpv-tomsk:
Added commit here for 'Add per-process open files count #836'.

rpv-tomsk:
After I had put this on my production, I found a drop on one of charts (Updated at 07:33):



Avg value before update was around ~65 and around ~40 after upgrade.
I have checked this with `# while true; do cat /proc/2317/stat && sleep 10; done` and "spreadsheet" and found what current value is right. Before upgrade there was collectd-5.5.2 code running.
That change is only at one chart. I have two mysql instances monitored same way and same collectd process, and there is not such drop or rise. Quite strange.

rpv-tomsk:
Changed `files` to `file_handles`.

Work finished, please review/merge it.

rpv-tomsk:
@rubenk, Can you please review this?

rpv-tomsk:
ping

Mogztter:
This feature would be really useful to us.
I can't review the code because I don't know C but let me know if you need help in order to get it merged.

rpv-tomsk:
Hi!

I'm just using this (like other my PR) in my production environment, so I don't experience any problems with code quality. 

Thanks for your attention to this patch request. 
I will update it to fix merge conflict, then you can try to get it merged.

rpv-tomsk:
Rebased 

Taken the src/processes.c file from cb8683ce442a0e7c14b910a7fff30a71b1acef35, 36cfb411f16ed11dadf70619f9779bf47c363340, 7c33e5d08e5567251d0a16a18bf4fd99b36f1ad9 commits, passed throught clang-format and done new commits.

rubenk:
This is Linux-only right? If so, please add that fact to the docs and commit message.

rpv-tomsk:
Hi Ruben!

That change would be enough?

rubenk:
Enough for what?

rpv-tomsk:
For this: 

> This is Linux-only right? If so, please add that fact to the docs and commit message.

src/processes.c
```diff
@@ -512,6 +562,8 @@ static int ps_config(oconfig_item_t *ci) {
   const size_t max_procname_len = MAXCOMLEN - 1;
 #endif
 
+  procstat_t *ps;
```
rubenk: perhaps move this to line 587, so it becomes `procstat_t *ps = ps_list_register(c->values[0].value.string, NULL);`?

rubenk:
Ah, I didn't see your change. Github lag I guess. Yes, it is fine.
Thanks Pavel!

Mogztter:
Awesome, thanks @rpv-tomsk and @rubenk

PR #1992, opened 2016-10-17
plugins: add fast path CPU usage monitoring plugin

kaminek:
This commit introduces a plugin to monitor CPU usage
for cores running a fast path process.
It works using fp-cpu-usage command, that dumps in json
format, statistics about each core running packet
processing.
Since the json output and the fp-cpu-usage command are
independent for any Packet Processing CPUs, this plugin
can be reused for any Packet Processing CPUs (P8,
ARM, Octeon, x86, etc.).

Signed-off-by: Amine Kherbouche amine.kherbouche@6wind.com

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5830,6 +5830,7 @@ plugin_drbd="no"
 plugin_dpdk="no"
 plugin_entropy="no"
 plugin_ethstat="no"
+plugin_fastpath="yes"
 plugin_fhcount="no"
 plugin_fscache="no"
 plugin_gps="no"
@@ -6285,6 +6286,7 @@ AC_PLUGIN([email],               [yes],                     [EMail statistics])
 AC_PLUGIN([entropy],             [$plugin_entropy],         [Entropy statistics])
 AC_PLUGIN([ethstat],             [$plugin_ethstat],         [Stats from NIC driver])
 AC_PLUGIN([exec],                [yes],                     [Execution of external programs])
+AC_PLUGIN([fastpath],            [yes],                     [Dataplane fast path statistics])
 AC_PLUGIN([fhcount],             [$plugin_fhcount],         [File handles statistics])
 AC_PLUGIN([filecount],           [yes],                     [Count files in directories])
 AC_PLUGIN([fscache],             [$plugin_fscache],         [fscache statistics])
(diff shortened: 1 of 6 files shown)


kaminek:
Hello,
There is a fail in jenkins build on-jessie-amd64-with-default-toolchain
Error:
`
fastpath.c:32:21: fatal error: jansson.h: No such file or directory
 #include <jansson.h>
                     ^
compilation terminated.
Makefile:4317: recipe for target 'fastpath.lo' failed
make[3]: *** [fastpath.lo] Error 1
`
Solution:
`apt-get install  libjansson-dev libjansson4`

configure.ac
```diff
@@ -5830,6 +5830,7 @@ plugin_drbd="no"
 plugin_dpdk="no"
 plugin_entropy="no"
 plugin_ethstat="no"
+plugin_fastpath="yes"
```
octo: Since the fastpath plugin depends on `jansson.h`, this variable needs to be enabled only when the library / header are available.

src/fastpath.c
```diff
@@ -0,0 +1,144 @@
...
+ * Authors:
+ *   Amine Kherbouche <amine.kherbouche at 6wind.com>
+ **/
+
+#include "collectd.h"
+#include "common.h" /* auxiliary functions */
+#include "plugin.h" /* plugin_register_*, plugin_dispatch_values */
+
+#include <stdio.h>
+#include <jansson.h>
```
octo: I'd much prefer if you could use **libyajl* instead: this is the JSON library used by numerous other plugins. Having one JSON library that's used by all plugins makes maintenance much easier for us.

src/fastpath.c
```diff
@@ -0,0 +1,144 @@
...
+
+static int fp_read (void)
+{
+	FILE * cpu_usage_json = NULL;
+	int i;
+	json_t *root, *cpus_info,
+	       *data, *core_id, *busy;
+	json_error_t error;
+
+	cpu_usage_json = popen("/usr/local/bin/fp-cpu-usage --json", "r");
```
octo: Forking off another process is a no-go. Please either find out how to get the data without forking or turn this plugin into a program / script that can be called by the [Exec plugin](https://collectd.org/wiki/index.php/Plugin:Exec).

src/fastpath.c
```diff
@@ -0,0 +1,144 @@
...
+
+#include "collectd.h"
+#include "common.h" /* auxiliary functions */
+#include "plugin.h" /* plugin_register_*, plugin_dispatch_values */
+
+#include <stdio.h>
+#include <jansson.h>
+
+#define PLUGIN_NAME "fastpath"
+#define PLUGIN_VALUE_TYPE_CPU "fastpath_cpu_busy"
```
octo: From the range of this *type* it looks like a percentage. If so, please use the `percent` type instead.

kaminek:
Hi @octo 
Thanks for the quick review and the comments, I'll do a rework.

octo:
Hi @kaminek, are you still interested in working on this or should I close the PR?

Thanks and best regards,
—octo

gfrey6wind:
Hi @octo 

You can close this PR, we found a new way to do it.

Thanks
Gabriel

octo:
Thanks for the update @gfrey6wind!

PR #1994, opened 2016-10-18
Set throttle metadata keys to DEBUG

dhrupadb:
Demoted logging level of throttle metadata keys from INFO to DEBUG to prevent logspam. Metadata key values already reported as Agent metric.


dhrupadb:
Sorry, accidentally compared against upstream.

PR #2003, opened 2016-10-26
mcelog plugin 

krzysztofx:
Plugin for collecting machine check exceptions  reported by [mcelog ](http://www.mcelog.org/)
The plugin at current stage does the following:
• Checks mcelog server liveliness, reports a failure if it’s not running of if it fails.
• Retrieve Memory Corrected and Uncorrected Errors from mcelog Unix socket (client protocol)
• Publishes collected data as notifications and statistics(values)

### DIFF
--- a/README
+++ b/README
@@ -198,13 +198,18 @@ Features
       Queries very detailed usage statistics from wireless LAN adapters and
       interfaces that use the Atheros chipset and the MadWifi driver.
 
-    - md
-      Linux software-RAID device information (number of active, failed, spare
-      and missing disks).
-
     - mbmon
       Motherboard sensors: temperature, fan speed and voltage information,
       using mbmon(1).
+
+    - mcelog
+      Monitor machine check exceptions (hardware errors detected by hardware
+      and reported to software) reported by mcelog and generate appropriate
+      notifications when machine check exceptions are detected.
+
+    - md
+      Linux software-RAID device information (number of active, failed, spare
+      and missing disks).
 
     - memcachec
       Query and parse data from a memcache daemon (memcached).
--- a/configure.ac
+++ b/configure.ac
@@ -5877,6 +5877,7 @@ plugin_ipvs="no"
 plugin_irq="no"
 plugin_load="no"
 plugin_log_logstash="no"
+plugin_mcelog="no"
 plugin_memory="no"
 plugin_multimeter="no"
 plugin_nfs="no"
@@ -5925,6 +5926,7 @@ if test "x$ac_system" = "xLinux"; then
   plugin_irq="yes"
   plugin_load="yes"
   plugin_lvm="yes"
+  plugin_mcelog="yes"
   plugin_memory="yes"
   plugin_nfs="yes"
   plugin_numa="yes"
@@ -6286,6 +6288,7 @@ AC_PLUGIN([match_regex],         [yes],                     [The regex match])
 AC_PLUGIN([match_timediff],      [yes],                     [The timediff match])
 AC_PLUGIN([match_value],         [yes],                     [The value match])
 AC_PLUGIN([mbmon],               [yes],                     [Query mbmond])
+AC_PLUGIN([mcelog],              [$plugin_mcelog],          [Machine Check Exceptions notifications])
 AC_PLUGIN([md],                  [$have_linux_raid_md_u_h], [md (Linux software RAID) devices])
 AC_PLUGIN([memcachec],           [$with_libmemcached],      [memcachec statistics])
 AC_PLUGIN([memcached],           [yes],                     [memcached statistics])
@@ -6709,6 +6712,7 @@ AC_MSG_RESULT([    match_regex . . . . . $enable_match_regex])
 AC_MSG_RESULT([    match_timediff  . . . $enable_match_timediff])
 AC_MSG_RESULT([    match_value . . . . . $enable_match_value])
 AC_MSG_RESULT([    mbmon . . . . . . . . $enable_mbmon])
+AC_MSG_RESULT([    mcelog  . . . . . . . $enable_mcelog])
 AC_MSG_RESULT([    md  . . . . . . . . . $enable_md])
 AC_MSG_RESULT([    memcachec . . . . . . $enable_memcachec])
 AC_MSG_RESULT([    memcached . . . . . . $enable_memcached])
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3324,23 +3324,23 @@ TCP-Port to connect to. Defaults to B<411>.
 
 =head2 Plugin C<mcelog>
 
-The C<mcelog plugin> uses mcelog to retrieve machine check exceptions, etc.
+The C<mcelog plugin> uses mcelog to retrieve machine check exceptions.
 
 By default the plugin connects to B<"/var/run/mcelog-client"> to check if the
 mcelog server is running. When the server is running, the plugin will tail the
 specified logfile to retrieve machine check exception information and send a
 notification with the details from the logfile. The plugin will use the mcelog
-client protocol to retrieve memory errors.
+client protocol to retrieve memory related machine check exceptions.
 
 =over 4
 
 =item B<McelogClientSocket> I<Path>
-Connect to mcelog client socket using the UNIX domain socket at I<Path>.
+Connect to the mcelog client socket using the UNIX domain socket at I<Path>.
 Defaults to B<"/var/run/mcelog-client">.
 
 =item B<McelogLogfile> I<Path>
 
-mcelog file to parse. Defaults to B<"/var/log/mcelog">.
+The mcelog file to parse. Defaults to B<"/var/log/mcelog">.
 
 =back
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3322,6 +3322,28 @@ TCP-Port to connect to. Defaults to B<411>.
 
 =back
 
+=head2 Plugin C<mcelog>
+
+The C<mcelog plugin> uses mcelog to retrieve machine check exceptions, etc.
+
+By default the plugin connects to B<"/var/run/mcelog-client"> to check if the
+mcelog server is running. When the server is running, the plugin will tail the
+specified logfile to retrieve machine check exception information and send a
+notification with the details from the logfile. The plugin will use the mcelog
+client protocol to retrieve memory errors.
+
+=over 4
+
+=item B<McelogClientSocket> I<Path>
+Connect to mcelog client socket using the UNIX domain socket at I<Path>.
+Defaults to B<"/var/run/mcelog-client">.
+
+=item B<McelogLogfile> I<Path>
+
+mcelog file to parse. Defaults to B<"/var/log/mcelog">.
+
+=back
+
 =head2 Plugin C<md>
 
 The C<md plugin> collects information from Linux Software-RAID devices (md).
--- a/src/mcelog.c
+++ b/src/mcelog.c
@@ -1,0 +1,518 @@
+/*-
+ * collectd - src/mcelog.c
+ * MIT License
+ *
+ * Copyright(c) 2016 Intel Corporation. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+
+ * Authors:
+ *   Maryam Tahhan <maryam.tahhan@intel.com>
+ *   Volodymyr Mytnyk <volodymyrx.mytnyk@intel.com>
+ *   Taras Chornyi <tarasx.chornyi@intel.com>
+ *   Krzysztof Matczak <krzysztofx.matczak@intel.com>
+ */
+
+#include "common.h"
+#include "collectd.h"
+
+#include <poll.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <unistd.h>
+
+#define MCELOG_PLUGIN "mcelog"
+#define MCELOG_BUFF_SIZE 1024
+#define MCELOG_POLL_TIMEOUT 1000 /* ms */
+#define MCELOG_SOCKET_STR "SOCKET"
+#define MCELOG_DIMM_NAME "DMI_NAME"
+#define MCELOG_CORRECTED_ERR "corrected memory errors:"
+#define MCELOG_UNCORRECTED_ERR "uncorrected memory errors:"
+
+typedef struct mcelog_config_s {
+  char logfile[PATH_MAX]; /* mcelog logfile */
+  pthread_t tid;          /* poll thread id */
+} mcelog_config_t;
+
+typedef struct socket_adapter_s socket_adapter_t;
+
+struct socket_adapter_s {
+  int sock_fd;                  /* mcelog server socket fd */
+  struct sockaddr_un unix_sock; /* mcelog client socket */
+  pthread_rwlock_t lock;
+  /* function pointers for socket operations */
+  int (*write)(socket_adapter_t *self, const char *msg, const size_t len);
+  int (*reinit)(socket_adapter_t *self);
+  int (*receive)(socket_adapter_t *self, FILE **p_file);
+  int (*close)(socket_adapter_t *self);
+};
+
+typedef struct mcelog_memory_rec_s {
+  char location[DATA_MAX_NAME_LEN];  /* SOCKET x CHANNEL x DIMM x*/
+  char dimm_name[DATA_MAX_NAME_LEN]; /* DMI_NAME "DIMM_F1" */
+  int corrected_err_total;           /* x total*/
+  int corrected_err_timed;           /* x in 24h*/
+  char corrected_err_timed_period[DATA_MAX_NAME_LEN];
+  int uncorrected_err_total; /* x total*/
+  int uncorrected_err_timed; /* x in 24h*/
+  char uncorrected_err_timed_period[DATA_MAX_NAME_LEN];
+} mcelog_memory_rec_t;
+
+static int socket_close(socket_adapter_t *self);
+static int socket_write(socket_adapter_t *self, const char *msg,
+                        const size_t len);
+static int socket_reinit(socket_adapter_t *self);
+static int socket_receive(socket_adapter_t *self, FILE **p_file);
+
+static mcelog_config_t g_mcelog_config = {
+    .logfile = "/var/log/mcelog", .tid = 0,
+};
+
+static socket_adapter_t socket_adapter = {
+    .sock_fd = -1,
+    .unix_sock =
+        {
+            .sun_family = AF_UNIX, .sun_path = "/var/run/mcelog-client",
+        },
+    .lock = PTHREAD_RWLOCK_INITIALIZER,
+    .close = socket_close,
+    .write = socket_write,
+    .reinit = socket_reinit,
+    .receive = socket_receive,
+};
+
+static _Bool mcelog_thread_running = 0;
+
+static int mcelog_config(oconfig_item_t *ci) {
+  for (int i = 0; i < ci->children_num; i++) {
+    oconfig_item_t *child = ci->children + i;
+    if (strcasecmp("McelogClientSocket", child->key) == 0) {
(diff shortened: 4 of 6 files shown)


rubenk:
Thanks for your contribution @krzysztofx!
I'll leave some comments inline.

README
```diff
@@ -926,6 +931,11 @@ Prerequisites
      `varnish' plugin.
      <http://varnish-cache.org>
 
+  * mcelog
```
rubenk: I don't think you need to add this block in this section, since it is mcelog is not a library and needed for building collectd.

README
```diff
@@ -198,14 +198,19 @@ Features
...
 
-    - md
-      Linux software-RAID device information (number of active, failed, spare
-      and missing disks).
-
     - mbmon
       Motherboard sensors: temperature, fan speed and voltage information,
       using mbmon(1).
 
+    - mcelogmon
```
rubenk: The plugin is called mcelog

configure.ac
```diff
@@ -6314,6 +6314,7 @@ AC_PLUGIN([match_regex],         [yes],                     [The regex match])
 AC_PLUGIN([match_timediff],      [yes],                     [The timediff match])
 AC_PLUGIN([match_value],         [yes],                     [The value match])
 AC_PLUGIN([mbmon],               [yes],                     [Query mbmond])
+AC_PLUGIN([mcelog],              [yes],                     [Machine Check Exceptions notifications])
```
rubenk: This always builds and installs mcelog, but I think it is Linux-specific?

src/Makefile.am
```diff
@@ -669,6 +669,16 @@ mbmon_la_LIBADD += -lsocket
 endif
 endif
 
+if BUILD_PLUGIN_MCELOG
+pkglib_LTLIBRARIES += mcelog.la
+mcelog_la_SOURCES = mcelog.c
+mcelog_la_LDFLAGS = $(PLUGIN_LDFLAGS)
+mcelog_la_LIBADD =
+if BUILD_WITH_LIBSOCKET
```
rubenk: I don't think you need this, it is mainly for Solaris iirc, and mcelog seems Linux-only.

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+        notification_t n = {
+            NOTIF_FAILURE, cdtime(), "", "", MCELOG_PLUGIN, "", "", "", NULL};
+        ssnprintf(n.message, sizeof(n.message),
+                  "Connection to mcelog socket is broken.");
+        sstrncpy(n.type_instance, "mcelog_status", sizeof(n.type_instance));
+        mcelog_dispatch_notification(n);
+        break;
+      }
+      if ((poll_fd.revents & POLLIN) || (poll_fd.revents & POLLPRI)) {
+        if ((p_file = fdopen(dup(g_mcelog_config.sock_fd), "r")) != NULL) {
```
octo: More very long blocks, deeply nested. Please handle the error, not the positive case and remove levels of indentation. The parsing should probably go into its own function to make it more readable.

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+          memset(buf, 0, sizeof(buf));
+          while (fgets(buf, sizeof(buf), p_file)) {
+            /* Got empty line or "done" */
+            if ((strlen(buf) == 1) ||
+                ((strlen(buf) >= 5) && (!strcmp(buf, "done\n")))) {
```
octo: The `strlen(buf) >= 5` comparison is unnecessary.

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+        sstrncpy(n.type_instance, "mcelog_status", sizeof(n.type_instance));
+        mcelog_dispatch_notification(n);
+        break;
+      }
+      if ((poll_fd.revents & POLLIN) || (poll_fd.revents & POLLPRI)) {
+        if ((p_file = fdopen(dup(g_mcelog_config.sock_fd), "r")) != NULL) {
+          memset(buf, 0, sizeof(buf));
+          while (fgets(buf, sizeof(buf), p_file)) {
+            /* Got empty line or "done" */
+            if ((strlen(buf) == 1) ||
```
octo: I think `strcmp("\n", buf) == 0` is easier to read (and more strict) than `strlen(buf) == 1`

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+                         sizeof(n.type_instance));
+
+                if (mcelog_prepare_notification(&n, memory_record) == 0)
+                  mcelog_dispatch_notification(n);
+                if (mcelog_submit(memory_record) != 0)
+                  ERROR("%s: Failed to submit memory errors", MCELOG_PLUGIN);
+                memset(&memory_record, 0, sizeof(memory_record));
+              }
+            }
+            if (strlen(buf) >= 5) {
```
octo:
```c
if (strlen(buf) < 5)
  continue;
```

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+              if (!strcmp(buf, "done\n")) {
+                DEBUG("%s: done", MCELOG_PLUGIN);
+                break;
+              }
+              if (!memcmp(buf, socket_str, strlen(socket_str))) {
```
octo: This has a high risk of reading uninitialized memory from `buf`. Please use `strncmp()` instead.

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+
+  while (1) {
+
+    pthread_mutex_lock(&mcelog_thread_lock);
+    if (mcelog_thread_shutdown == 1) {
+      pthread_mutex_unlock(&mcelog_thread_lock);
+      pthread_exit(NULL);
+    }
+    pthread_mutex_unlock(&mcelog_thread_lock);
+    int poll_ret = poll(&poll_fd, 1, POLL_TIMEOUT);
```
octo: It might be worth sending this thread a signal (using `pthread_kill()` from the `shutdown()` callback) instead of busy looping here.

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+
+  return 0;
+}
+
+static int mcelog_shutdown(void) {
+  pthread_mutex_lock(&mcelog_thread_lock);
+  mcelog_thread_shutdown = 1;
+  pthread_mutex_unlock(&mcelog_thread_lock);
+  pthread_join(g_mcelog_config.tid, NULL);
+  pthread_mutex_destroy(&mcelog_thread_lock);
```
octo: Please don't destroy mutexes that you initialized with the static initializer.

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+  shutdown(g_mcelog_config.sock_fd, 0);
```
octo: Please check that `sock_fd` is in fact a valid file descriptor before calling `shutdown()`.

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+typedef struct mcelog_config_s mcelog_config_t;
+typedef struct mcelog_memory_rec_s mcelog_memory_rec_t;
+
+mcelog_config_t g_mcelog_config;
+
+static int g_configured;
+static _Bool mcelog_thread_shutdown = 0;
+static pthread_mutex_t mcelog_thread_lock = PTHREAD_MUTEX_INITIALIZER;
+
+static const char socket_str[] = "SOCKET";
```
octo: Having these `const char[]` up here doesn't really improve the parsing code below. Please simply use string literals in the worker thread.

krzysztofx:
@octo @rubenk   thanks for your  reviews.  I'll respond  your comments soon.

README
krzysztofx: I'll fix that

README
krzysztofx: right, it's not compilation prerequisite

configure.ac
krzysztofx: right, I'll fix that

src/Makefile.am
krzysztofx: libsocket linkage removed

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+  char corrected_err_timed_period[DATA_MAX_NAME_LEN];
+  int uncorrected_err_total; /* x total*/
+  int uncorrected_err_timed; /* x in 24h*/
+  char uncorrected_err_timed_period[DATA_MAX_NAME_LEN];
+};
+
+typedef struct mcelog_config_s mcelog_config_t;
+typedef struct mcelog_memory_rec_s mcelog_memory_rec_t;
+
+mcelog_config_t g_mcelog_config;
```
krzysztofx: I'd like to provide way to reinitalize   g_mcelog_config struct, so  how do you see usage of compound literals there ? I mean:
```c
static mcelog_config_t g_mcelog_config ;$
// skippded lines
static void mcelog_config_init_default(void) {$
  g_mcelog_config = (mcelog_config_t){$
    .logfile = "/var/log/mcelog",$
    .unix_sock = {$
      .sun_family = AF_UNIX,$
      .sun_path = "/var/run/mcelog-client",$
    },$
    .tid = 0,$
    .sock_fd = -1,$
  };$
```

src/mcelog.c
krzysztofx: You mean adding signal handler to thread function  poll_worker ? Or just terminating thread asynchronously sending SIGTERM ?  I have noticed that some plugins use that approach but  I think this is not applicable here since we have to do some cleanup like calling fclose and so on.

src/mcelog.c
krzysztofx: ok

src/mcelog.c
krzysztofx: ok

src/mcelog.c
krzysztofx: ok

src/mcelog.c
krzysztofx: Ok understood. First handle negative scenario flow

src/mcelog.c
krzysztofx: I will fix it

src/mcelog.c
krzysztofx: ok, I will change it

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+
+    pthread_mutex_lock(&mcelog_thread_lock);
+    if (mcelog_thread_shutdown == 1) {
+      pthread_mutex_unlock(&mcelog_thread_lock);
+      pthread_exit(NULL);
+    }
+    pthread_mutex_unlock(&mcelog_thread_lock);
+    int poll_ret = poll(&poll_fd, 1, POLL_TIMEOUT);
+
+    if (poll_ret > 0) {
```
krzysztofx: ok

src/mcelog.c
krzysztofx: I will terminate thread with thread with pthread_cancel and put cleanup code into callback function  set by pthread_cancel_push. Is it ok from your POV ?

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+    sstrncpy(vl.plugin_instance, plugin_instance, sizeof(vl.plugin_instance));
+  } else
+    sstrncpy(vl.plugin_instance, mr.location, sizeof(vl.plugin_instance));
+
+  sstrncpy(vl.type_instance, "corrected_memory_errors",
+           sizeof(vl.type_instance));
+  vl.values = &(value_t){.derive = (derive_t)mr.corrected_err_total};
+  plugin_dispatch_values(&vl);
+
+  snprintf(type_instance, DATA_MAX_NAME_LEN, "corrected_memory_errors_in_%s",
```
krzysztofx: I'll fix it

src/mcelog.c
```diff
@@ -0,0 +1,431 @@
...
+  /*Set socket timeout option*/
+  if (setsockopt(g_mcelog_config.sock_fd, SOL_SOCKET, SO_SNDTIMEO,
+                 (char *)&socket_timeout, sizeof(socket_timeout)) < 0)
+    ERROR("%s: Failed to set the socket timeout option.", MCELOG_PLUGIN);
+
+  if (connect(g_mcelog_config.sock_fd,
+              (struct sockaddr *)&(g_mcelog_config.unix_sock),
+              sizeof(struct sockaddr_un)) < 0) {
+    sstrerror(errno, errbuff, sizeof(errbuff));
+    ERROR("%s: Failed to connect to mcelog server. %s", MCELOG_PLUGIN, errbuff);
```
krzysztofx: ok

src/mcelog.c
krzysztofx: I will change it

krzysztofx:
Hi @octo @rubenk 
I've pushed patch addressing your comments and also fix for build failure on  epel5 setup.
Please let me know if  you will find out  anything else  that could be improved.

README
```diff
@@ -202,10 +202,10 @@ Features
       Motherboard sensors: temperature, fan speed and voltage information,
       using mbmon(1).
 
-    - mcelogmon
-      Monitor machine check errors (hardware error detected by hardware
-      and reported to software) reported by mcelog and generate appropriate
-      notifications when machine check errors are detected.
+    - mcelog
+      Logs and accounts machine check exceptions (in particular memory, IO, and
```
rubenk: I liked the original text better. The plugin doesn't actually log MCE's right?

krzysztofx: It logs only the  amount of corrected/uncorrected  memory related errors  per  memory SOCKET and CHANNEL.  No MCE error codes are logged if that's what you mean. Data source is equivalent to output of  'mcelog --client', for example:
```sh
[root@silpixa00392165 kmatczax]# mcelog --client
Memory errors
SOCKET 0 CHANNEL any DIMM any
corrected memory errors:
        0 total
        0 in 24h
uncorrected memory errors:
        8 total
        8 in 24h

SOCKET 0 CHANNEL 0 DIMM any
corrected memory errors:
        0 total
        0 in 24h
uncorrected memory errors:
        8 total
        8 in 24h
```

krzysztofx:
Hi

I've committed  fix for typedef redefinition. All check passed but now I see merge conflict with no details given.  Please let me know if I could do anything to fix that, or if any other changes in plugin  code are still required.

krzysztofx:
@rubenk @octo 
Would you mind If I rebase that PR against master branch to fix the conflict ?

krzysztofx:
Rebase done

krzysztofx:
Hi 
@rubenk @octo 
Sorry for pinging , but is there anything that could be done to get that plugin merged? If yes, please let me know. 
I'm developing extension for that plugin that collects  all MCE exceptions, so it would be easier to base on already merged version.

rubenk:
Could you re-order or squash some of your commits? You now have multiple commits with the same commit messages.

src/Makefile.am
```diff
@@ -677,6 +677,13 @@ mbmon_la_LIBADD += -lsocket
 endif
 endif
 
+if BUILD_PLUGIN_MCELOG
+pkglib_LTLIBRARIES += mcelog.la
+mcelog_la_SOURCES = mcelog.c
+mcelog_la_LDFLAGS = $(PLUGIN_LDFLAGS)
+mcelog_la_LIBADD =
```
rubenk: this is not needed.

src/collectd.conf.pod
```diff
@@ -3322,6 +3322,28 @@ TCP-Port to connect to. Defaults to B<411>.
 
 =back
 
+=head2 Plugin C<mcelog>
+
+The C<mcelog plugin> uses mcelog to retrieve machine check exceptions, etc.
```
rubenk: What does the 'etc.' stand for? It doesn't add much information, I'd either leave it out or extend the sentence.

src/collectd.conf.pod
```diff
@@ -3322,6 +3322,28 @@ TCP-Port to connect to. Defaults to B<411>.
...
+
+By default the plugin connects to B<"/var/run/mcelog-client"> to check if the
+mcelog server is running. When the server is running, the plugin will tail the
+specified logfile to retrieve machine check exception information and send a
+notification with the details from the logfile. The plugin will use the mcelog
+client protocol to retrieve memory errors.
```
rubenk: I'd prefer 'retrieve machine check exceptions' instead of 'memory errors'.

src/collectd.conf.pod
```diff
@@ -3322,6 +3322,28 @@ TCP-Port to connect to. Defaults to B<411>.
...
+
+=over 4
+
+=item B<McelogClientSocket> I<Path>
+Connect to mcelog client socket using the UNIX domain socket at I<Path>.
```
rubenk: Connect to the mcelog client socket

src/collectd.conf.pod
```diff
@@ -3322,6 +3322,28 @@ TCP-Port to connect to. Defaults to B<411>.
...
+Defaults to B<"/var/run/mcelog-client">.
+
+=item B<McelogLogfile> I<Path>
+
+mcelog file to parse. Defaults to B<"/var/log/mcelog">.
```
rubenk: The mcelog logfile to parse.

krzysztofx:
@rubenk 
Thanks for your review. I'll squash redundant commits  and address your comments soon

Thanks
Krzysztof

src/mcelog.c
```diff
@@ -0,0 +1,518 @@
...
+static int socket_reinit(socket_adapter_t *self) {
+  char errbuff[MCELOG_BUFF_SIZE];
+  int flags;
+  int ret = -1;
+  cdtime_t interval = plugin_get_interval();
+  struct timeval socket_timeout = CDTIME_T_TO_TIMEVAL(interval);
+
+  /* synchronization via write lock since sock_fd may be changed here */
+  pthread_rwlock_wrlock(&self->lock);
+  self->sock_fd = socket(PF_UNIX, SOCK_STREAM, 0);
```
rubenk: Could you use SOCK_CLOEXEC here?

rubenk: Oh, and while you're at it, add SOCK_NONBLOCK to the flags, so you don't have to do the fcntl later.

src/collectd.conf.pod
krzysztofx: At the moment mcelog's client/server protocol  returns only memory related MCE's, see http://www.mcelog.org/protocol.html. That's why I used 'memory errors' term. Would you like to extend it to  all MCE,  or maybe  use term 'memory related machine check exceptions' ?

src/mcelog.c
```diff
@@ -0,0 +1,518 @@
...
+    return ret;
+  }
+
+  flags = fcntl(self->sock_fd, F_GETFL, 0);
+  flags |= O_NONBLOCK;
+  fcntl(self->sock_fd, F_SETFL, flags);
+
+  /* Set socket timeout option */
+  if (setsockopt(self->sock_fd, SOL_SOCKET, SO_SNDTIMEO,
+                 (char *)&socket_timeout, sizeof(socket_timeout)) < 0)
```
rubenk: I think the (char *) cast is unneeded.

src/mcelog.c
krzysztofx: regarding SOCK_NONBLOCK please see [d212c7f](https://github.com/collectd/collectd/pull/2003/commits/d212c7f2fb8af6d3ade42aaf1d019c5a2d7e8d3b). It failed on epel5

src/mcelog.c
```diff
@@ -0,0 +1,518 @@
...
+    res = -1;
+
+  pthread_rwlock_unlock(&self->lock);
+  return res;
+}
+
+static void *poll_worker(__attribute__((unused)) void *arg) {
+  char errbuf[MCELOG_BUFF_SIZE];
+  mcelog_thread_running = 1;
+  FILE **pp_file = calloc(1, sizeof(FILE *));
```
rubenk: Please use sizeof on the var, not the type, so sizeof(*pp_file)

src/mcelog.c
```diff
@@ -0,0 +1,518 @@
...
+    }
+    /* timeout or no data to read */
+    else if (res == 0)
+      continue;
+
+    if (*pp_file == NULL)
+      continue;
+
+    mcelog_memory_rec_t memory_record;
+    memset(&memory_record, 0, sizeof(memory_record));
```
rubenk: Please use designated initialiser here, so `mcelog_memory_rec_t memory_record = {0}`

src/mcelog.c
rubenk: Ah right, sorry for the confusion. The situation has changed now though, after we released 5.7 we dropped support for RHEL5 from master. Jenkins is still doing build on RHEL5 but those will be cleaned up shortly and you can ignore them.

krzysztofx:
My apologies for squashing by mistake all previous  commits into [c8f08229](https://github.com/collectd/collectd/pull/2003/commits/c8f08229b48085338394a99847b7c16de252b0da). I hope it doesn't make it much harder to review. If you think I could somehow revert that rebase and squash only specific commits please let me know. 
All recent comments have been addressed in [6016e1d](https://github.com/collectd/collectd/pull/2003/commits/6016e1d7ac73b9ac0e0eb58350f9ee710ccda4a5l)

Now I see merge conflict with the master branch. Do you agree for rebase ? 

Thanks
Krzysztof

krzysztofx:
[pull-requests-github_trigger-aggregation](https://ci.collectd.org/job/pull-requests-build-on-jessie-amd64-with-default-toolchain/1379/) failed due to:
 ```
mcelog.c:437:5: error: missing braces around initializer [-Werror=missing-braces]
     mcelog_memory_rec_t memory_record = {0};
```

Looks like this job has more restrictive compiler settings than others . Should I put additional braces like {{0}} or ignore that fail ?

**Comment by rubenk** [comment]
No, that's not it, the compiler settings are the same. This job is the first that runs and the other ones are only run if this one succeeds.

This looks like https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119.

Perhaps you could try changing the order of the members of the mcelog_memory_rec_s struct around to work around it? My guess is that this won't happen if the first member is an int instead of char[]

**Comment by krzysztofx** [comment]
@rubenk 
Changing order of struct members helped for missing braces warning. Now as expected  I got failures on epel5 builds due to unknown socket flags:
 
```
mcelog.c: In function 'socket_reinit':
mcelog.c:163: error: 'SOCK_CLOEXEC' undeclared (first use in this function)
mcelog.c:163: error: (Each undeclared identifier is reported only once
mcelog.c:163: error: for each function it appears in.)
mcelog.c:163: error: 'SOCK_NONBLOCK' undeclared (first use in this function)
make[3]: *** [mcelog.lo] Error 1
```

So according to previous comments this is acceptable, right ? 
Please let me know if I should rebase to master. 

thanks
Krzysztof

rubenk:
Thanks @krzysztofx, yes, these errors on EPEL5 are perfectly alright.
Please rebase and then I'll merge.

krzysztofx:
@rubenk 
Rebase done

Thanks
Krzysztof

PR #2006, opened 2016-10-26
utils_time.h: Use more composite literals.

octo:


rubenk:
Nice cleanup! LGTM.

octo:
Oh noes!

```
curl.c: In function 'cc_submit_response_time':
curl.c:644:26: internal compiler error: Segmentation fault
```

octo:
That line seems to have a composite literal within a composite literal. Maybe these old versions of GCC didn't drink the kool-aid yet …

octo:
Thanks for reviewing this @rubenk! I'll submit when I find out how to make *precise* happy.

rubenk:
The funny thing is it does build on RHEL5 :) Speaking of which, we should really drop support for those old distros ;)

Issue #2009, opened 2016-10-28
5.6.1.219.g538f8ff crashes on Wheezy


octo:
*   Version of collectd: 5.6.1.219.g538f8ff
*   Operating system / distribution: Debian Wheezy, i386

## Actual behavior

After upgrading the Debian package from `pkg.ci.collectd.org`, the daemon does not start up anymore. It looks like glibc is catching an invalid *free()* and then just freezes:

```
*** glibc detected *** /usr/sbin/collectd: free(): invalid next size (fast): 0x080b6de8 ***
```

A stacktrace shows that this might be in `src/daemon/common.c:check_capability()`:

```
#19 0xb7e658ed in *__GI___libc_free (mem=0x80b6de8) at malloc.c:3738
        ar_ptr = 0xb7f563c0
        p = 0x8097120
#20 0x08061327 in check_capability (capability=capability@entry=7) at common.c:1764
        cap_header = 0x0
        cap_data = 0x80b6de8
#21 0xb79a53b2 in exec_init () at exec.c:814
        sa = {__sigaction_handler = {sa_handler = 0xb79a3b60 <sigchld_handler>,
            sa_sigaction = 0xb79a3b60 <sigchld_handler>}, sa_mask = {__val = {0 <repeats 32 times>}},
          sa_flags = 0, sa_restorer = 0}
```

Unfortunately the line numbers don't match with Git.

octo:
I'm happy to report that #2010 fixes this.

PR #2010, opened 2016-10-28
src/daemon/common.c: Rewrite check_capability() using cap_get_bound().

octo:
capget(2) is Linux specific and the use of the raw syscalls is
discouraged. Also, there have been interesting crashes on some systems.

Issue: #2009

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -808,7 +808,20 @@ fi
 have_cpuid_h="no"
 AC_CHECK_HEADERS(cpuid.h, [have_cpuid_h="yes"])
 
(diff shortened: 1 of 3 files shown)


rubenk:
LGTM

mfournier:
Cool, thanks @octo ! Sorry for the mess I made...

One note though: the purpose of not using `cap_get_bound()` and friends was to avoid linking an additionnal runtime library to plugins just to emit a warning message at startup.

We will at least need to properly mention this in the changelog. I can foresee users upgrading to 5.7 and complaining various plugins can't load anymore.

Issue #1932, opened 2016-09-13
OpenVPN plugin: Add support for version 2.3.2.


Giu989:
*   Version of collected: 5.6
*   Operating system / distribution: Ubuntu 14

Sorry if this is easily fixed or if this is not a bug but I am very new to collectd. I have enabled the OpenVPN plugin on my server, but in /var/log/syslog from time to time this error appears:

Sep 13 19:40:57 RekServerV2 collectd[1396]: read-function of plugin `openvpn' failed. Will suspend it for 80.000 seconds.

And then 160 seconds, 320 seconds etc...

octo:
Hi @Giu989,

that error alone only tells us that the *OpenVPN plugin* failed for some reason, but nothing about the reason itself. Are there any collectd log entries preceding that line?

Best regards,
—octo

Giu989:
Hi, thanks for the quick reply! Just after the error there are these log entries. I apologise, I should have included these in the original message:

openvpn plugin: /etc/openvpn/openvpn1.log: Unknown file format, please report this as bug. Make sure to include your status file, so the plugin can be adapted.

openvpn plugin: unable to detect status version, #011#011#011#011#011discarding status file "/etc/openvpn/openvpn1.log".

I have these 2 relevant entries in the OpenVPN server.conf located in /etc/openvpn:

status-version 2
log-append openvpn1.log 10

collectd.conf has this for openvpn:

<Plugin openvpn>
        StatusFile "/etc/openvpn/openvpn1.log"
        ImprovedNamingSchema false
        CollectCompression true
        CollectIndividualUsers false
        CollectUserCount true 
</Plugin>

Thank you very much!

Giu

Giu989:
Edit: The plugin seems to be failing as extra lines have been added at the top of the status file, so the plugin doesn't recognise the file. Removing these lines from the start of the file made the plugin no longer fail and an output was generated, but at this point 'openvpn-status.log' would no longer update with new information.

octo:
Hi @Giu989,

thanks for providing an example `openvpn-status.log`. This increases your chances of someone picking this up and implementing support considerably.

Best regards,
—octo

rpv-tomsk:
I'm unable to reproduce this issue. The code of collectd-5.6 detects version 4 properly.

I think the problem was caused by empty openvpn-status.log file.

octo:
It's likely that #2012 fixed this issue. Please re-open if you're still experiencing problems.

PR #2012, opened 2016-10-29
openvpn: Fixed openvpn_read() on empty configuration.

rpv-tomsk:
When plugin loaded without configuration, it starts to spam logs with message
'read-function of plugin `openvpn' failed'.

Related issue: #1932


octo:
Thanks @rpv-tomsk!

PR #2015, opened 2016-10-30
tail plugin: Implement the "GaugePersist" type.

octo:


src/daemon/utils_match.c
```diff
@@ -295,14 +295,13 @@ cu_match_t *match_create_simple (const char *regex,
 
 void match_value_reset (cu_match_value_t *mv)
 {
-  if (mv == NULL)
+  if ((mv == NULL)
+      || ((mv->ds_type & UTILS_MATCH_DS_TYPE_GAUGE) == 0)
+      || ((mv->ds_type & UTILS_MATCH_CF_GAUGE_PERSIST) != 0))
```
rpv-tomsk: I think what this piece of logic is hard to read.
It is unclean what this really is:

```
if (mv == NULL)
  return;

if (mv->ds_type & UTILS_MATCH_DS_TYPE_GAUGE)
 {
  if (mv->ds_type & UTILS_MATCH_CF_GAUGE_PERSIST)
    return;

   mv->value.gauge = NAN;
   mv->values_num = 0;
 }
```

src/collectd.conf.pod
```diff
@@ -7144,6 +7144,13 @@ Use the greatest number only.
 
 Use the last number found.
 
+=item B<GaugePersist>
+
+Use the last number found. The number if not reset at the end of an interval.
```
rpv-tomsk: ...**is** not reset...

octo:
@rpv-tomsk how's this?

src/daemon/utils_match.c
octo: Fixed, thanks!

rpv-tomsk:
Hi! I think it much better currently :+1:

Issue #1523, opened 2016-01-25
collectd doesn't build reproducibly on armhf


mfournier:
From https://tests.reproducible-builds.org/rb-pkg/unstable/armhf/collectd.html:

```
-checking for jni.h... found in /usr/lib/jvm/default-java/include
-checking for jni_md.h... found in /usr/lib/jvm/default-java/include
-checking for libjvm.so... found in /usr/lib/jvm/default-java/jre/lib/arm/server
-configure: Building with JAVA_CPPFLAGS set to: -I/usr/lib/jvm/default-java/include -I/usr/lib/jvm/default-java/include -I/usr/lib/jvm/default-java/include
-configure: Building with JAVA_LDFLAGS set to: -L/usr/lib/jvm/default-java/jre/lib/arm/server -Wl,-rpath -Wl,/usr/lib/jvm/default-java/jre/lib/arm/server -L/usr/lib/jvm/default-java/jre/lib/arm/server -Wl,-rpath -Wl,/usr/lib/jvm/default-java/jre/lib/arm/server
+checking for jni.h... found in /usr/lib/jvm/java-1.7.0-openjdk-armhf/include
+checking for jni_md.h... found in /usr/lib/jvm/java-1.7.0-openjdk-armhf/include
+checking for libjvm.so... found in /usr/lib/jvm/java-1.7.0-openjdk-armhf/jre/lib/arm/server
+configure: Building with JAVA_CPPFLAGS set to: -I/usr/lib/jvm/default-java/include -I/usr/lib/jvm/java-1.7.0-openjdk-armhf/include -I/usr/lib/jvm/java-1.7.0-openjdk-armhf/include
+configure: Building with JAVA_LDFLAGS set to: -L/usr/lib/jvm/default-java/jre/lib/arm/server -Wl,-rpath -Wl,/usr/lib/jvm/default-java/jre/lib/arm/server -L/usr/lib/jvm/java-1.7.0-openjdk-armhf/jre/lib/arm/server -Wl,-rpath -Wl,/usr/lib/jvm/java-1.7.0-openjdk-armhf/jre/lib/arm/server
```

A difference between both builds is that the environments used are configured with a different locale. So my assumption is that `find` behaves differently depending on that, and a fix could be as simple as slapping `LC_ALL=C` at the top of `configure.ac`.

What puzzles me though is why only on this platform (amd64 for example builds reproducibly).

[...]

PR #2017, opened 2016-11-01
configure.ac: make finding java deterministic

mfournier:
The sort order of the `find` command proves to be dependent on external
factors, which makes `./configure` pick different java versions on
different systems, making the whole build non-reproducible (see
https://reproducible-builds.org/ for more details).

Adding a call to `sort`, with the locale forced, fixes this issue.

Fixes: #1523

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2259,7 +2259,7 @@ then
 	if test -d "$with_java_home"
 	then
 		AC_MSG_CHECKING([for jni.h])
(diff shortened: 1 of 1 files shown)


mfournier:
FWIW, this patch has been part of the packages in debian unstable for a couple of weeks, achieving build reproducibility on a bunch of platforms/architectures: https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/collectd.html

PR #2024, opened 2016-11-05
ipmi plugin: Implemented IPMI LAN connection and plugin improved 

rpv-tomsk:
Changes: 

* Added instances support and IPMI LAN connection option
* Added a notification on connection state change (notify about connection issues)
* Removed  the sensor removal in case of reading errors
* Added the lock to do not request new readings until previous reading is complete
* Disabled SDRs local cache
* Added the check for sensor 'is readable' flag
* Added the check for sensor event reading type (if sensor is discrete or not)
* Added checks for sensor states 'scanning disabled' and 'initial update in progress'
* Changes in error reporting

Please review and merge.

Superseedes #2023

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3351,7 +3351,35 @@ This option is only available on Solaris.
 
 =head2 Plugin C<ipmi>
 
+The B<ipmi plugin> allows to monitor server platform status using the Intelligent
+Platform Management Interface (IPMI). Local and remote interfaces are supported.
+
+The plugin configuration consists of one or more B<Instance> blocks which
+specify one I<ipmi> connection each. Within the B<Instance> blocks, the
+following options are allowed:
+
 =over 4
+
+=item B<Address> I<Address>
+
+Hostname or IP to connect to. If not specified, plugin will try to connect to
+local management controller (BMC).
+
+=item B<Username> I<Username>
+
+=item B<Password> I<Password>
+
+The username and the password to use for the connection to remote BMC.
+
+=item B<AuthType> I<MD5>|I<rmcp+>
+
+Forces the authentication type to use for the connection to remote BMC.
+By default most secure type is seleted.
+
+=item B<Host> I<Hostname>
+
+Sets the B<host> field of dispatched values. Defaults to the global hostname
+setting.
 
 =item B<Sensor> I<Sensor>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3409,6 +3409,11 @@ If a sensor disappears a notification is sent.
 If you have for example dual power supply and one of them is (un)plugged then
 a notification is sent.
 
+=item B<NotifyIPMIConnectionState> I<true>|I<false>
+
+If a IPMI connection state changes after initialization time of a minute
+a notification is sent. Defaults to B<false>.
+
 =item B<SELEnabled> I<true>|I<false>
 
 If system event log (SEL) is enabled, plugin will listen for sensor threshold
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3355,8 +3355,15 @@ The B<ipmi plugin> allows to monitor server platform status using the Intelligen
 Platform Management Interface (IPMI). Local and remote interfaces are supported.
 
 The plugin configuration consists of one or more B<Instance> blocks which
-specify one I<ipmi> connection each. Within the B<Instance> blocks, the
-following options are allowed:
+specify one I<ipmi> connection each. Each block requires one unique string
+argument as the instance name. If instances are not configured, an instance with
+the default option values will be created.
+
+For backwards compatibility, any option other than B<Instance> block will trigger
+legacy config handling and it will be treated as an option within B<Instance>
+block. This support will go away in the next major version of Collectd.
+
+Within the B<Instance> blocks, the following options are allowed:
 
 =over 4
 
--- a/src/ipmi.c
+++ b/src/ipmi.c
@@ -902,10 +902,9 @@ static int c_ipmi_thread_init(c_ipmi_instance_t *st) {
 } /* int c_ipmi_thread_init */
 
 static void *c_ipmi_thread_main(void *user_data) {
-  int status;
   c_ipmi_instance_t *st = (c_ipmi_instance_t *)user_data;
 
-  status = c_ipmi_thread_init(st);
+  int status = c_ipmi_thread_init(st);
   if (status != 0) {
     ERROR("ipmi plugin: c_ipmi_thread_init failed.");
     st->active = 0;
@@ -974,7 +973,7 @@ static void c_ipmi_free_instance(c_ipmi_instance_t *st) {
   sfree(st);
 } /* void c_ipmi_free_instance */
 
-void c_ipmi_add_instance(c_ipmi_instance_t *instance) {
+static void c_ipmi_add_instance(c_ipmi_instance_t *instance) {
   if (instances == NULL) {
     instances = instance;
     return;
@@ -986,8 +985,6 @@ void c_ipmi_add_instance(c_ipmi_instance_t *instance) {
     last = last->next;
 
   last->next = instance;
-
-  return;
 } /* void c_ipmi_add_instance */
 
 static int c_ipmi_config_add_instance(oconfig_item_t *ci) {
@@ -1096,8 +1093,7 @@ static int c_ipmi_config(oconfig_item_t *ci) {
 } /* int c_ipmi_config */
 
 static int c_ipmi_read(user_data_t *user_data) {
-  c_ipmi_instance_t *st;
-  st = user_data->data;
+  c_ipmi_instance_t *st = user_data->data;
 
   if ((st->active == 0) || (st->thread_id == (pthread_t)0)) {
     INFO("ipmi plugin: c_ipmi_read: I'm not active, returning false.");
--- a/src/ipmi.c
+++ b/src/ipmi.c
@@ -38,6 +38,8 @@
 #include <OpenIPMI/ipmi_smi.h>
 #include <OpenIPMI/ipmiif.h>
 
+#define ERR_BUF_SIZE 1024
+
 /*
  * Private data types
  */
@@ -93,7 +95,7 @@ static c_ipmi_instance_t *instances = NULL;
  * Misc private functions
  */
 static void c_ipmi_error(c_ipmi_instance_t *st, const char *func, int status) {
-  char errbuf[4096] = {0};
+  char errbuf[ERR_BUF_SIZE] = {0};
 
   if (IPMI_IS_OS_ERR(status) || IPMI_IS_RMCPP_ERR(status) ||
       IPMI_IS_IPMI_ERR(status)) {
@@ -110,7 +112,7 @@ static void c_ipmi_error(c_ipmi_instance_t *st, const char *func, int status) {
 
 static void c_ipmi_log(os_handler_t *handler, const char *format,
                        enum ipmi_log_type_e log_type, va_list ap) {
-  char msg[1024];
+  char msg[ERR_BUF_SIZE];
 
   vsnprintf(msg, sizeof(msg), format, ap);
 
@@ -202,7 +204,7 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
       INFO("ipmi plugin: sensor_read_handler: Sensor `%s` of `%s` timed out.",
            list_item->sensor_name, st->name);
     } else {
-      char errbuf[128] = {0};
+      char errbuf[ERR_BUF_SIZE] = {0};
       ipmi_get_error_string(err, errbuf, sizeof(errbuf) - 1);
 
       if (IPMI_IS_IPMI_ERR(err))
--- a/src/ipmi.c
+++ b/src/ipmi.c
@@ -59,6 +59,7 @@ struct c_ipmi_instance_s {
   char *password;
   unsigned int authtype;
 
+  _Bool connected;
   ipmi_con_t *connection;
   pthread_mutex_t sensor_list_lock;
   c_ipmi_sensor_list_t *sensor_list;
@@ -90,7 +91,7 @@ static c_ipmi_instance_t *instances = NULL;
 /*
  * Misc private functions
  */
-static void c_ipmi_error(const char *func, int status) {
+static void c_ipmi_error(c_ipmi_instance_t *st, const char *func, int status) {
   char errbuf[4096] = {0};
 
   if (IPMI_IS_OS_ERR(status) || IPMI_IS_RMCPP_ERR(status) ||
@@ -103,7 +104,7 @@ static void c_ipmi_error(const char *func, int status) {
   }
   errbuf[sizeof(errbuf) - 1] = 0;
 
-  ERROR("ipmi plugin: %s failed: %s", func, errbuf);
+  ERROR("ipmi plugin: %s failed for `%s`: %s", func, st->name, errbuf);
 } /* void c_ipmi_error */
 
 static void c_ipmi_log(os_handler_t *handler, const char *format,
@@ -782,13 +783,13 @@ domain_entity_update_handler(enum ipmi_update_e op,
     status = ipmi_entity_add_sensor_update_handler(
         entity, entity_sensor_update_handler, /* user data = */ (void *)st);
     if (status != 0) {
-      c_ipmi_error("ipmi_entity_add_sensor_update_handler", status);
+      c_ipmi_error(st, "ipmi_entity_add_sensor_update_handler", status);
     }
   } else if (op == IPMI_DELETED) {
     status = ipmi_entity_remove_sensor_update_handler(
         entity, entity_sensor_update_handler, /* user data = */ (void *)st);
     if (status != 0) {
-      c_ipmi_error("ipmi_entity_remove_sensor_update_handler", status);
+      c_ipmi_error(st, "ipmi_entity_remove_sensor_update_handler", status);
     }
   }
 } /* void domain_entity_update_handler */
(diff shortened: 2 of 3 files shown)


rpv-tomsk:
ping

3add3287:
Is this ever going to get merged into master?

rpv-tomsk:
I will rebase this when needed by somebody, just ping me.

3add3287:
Ping ! ;-)

rpv-tomsk:
Rebased.

Please, put your feedbacks here. That will increase chanses this to be merged.

3add3287:
Wow, thanks for rebasing it so quickly!
Now the question is: what does it take the powers that be to merge it into collectd master?

rpv-tomsk:
Patch rebased to current master, updated with respect of changes recently added to ipmi.c
Unfortunately, I can't do nothing with CI build failure. That is a 'timestamp is in future' problem.

src/ipmi.c
```diff
@@ -138,43 +193,50 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
...
-      INFO("ipmi plugin: sensor_read_handler: Sensor %s not ready",
-           list_item->sensor_name);
+      INFO("ipmi plugin: sensor_read_handler: Sensor `%s` of `%s` not ready.",
+           list_item->sensor_name, st->name);
+    } else if (IPMI_IS_IPMI_ERR(err) &&
+               IPMI_GET_IPMI_ERR(err) == IPMI_TIMEOUT_CC) {
+      INFO("ipmi plugin: sensor_read_handler: Sensor `%s` of `%s` timed out.",
+           list_item->sensor_name, st->name);
     } else {
+      char errbuf[128] = {0};
```
maryamtahhan: are you sure 128 is big enough? most other errbufs use 1024... I would also suggest replacing the magic # with a #define 

#define ERR_BUF_SIZE 1024

src/ipmi.c
```diff
@@ -186,19 +248,35 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
...
+          "it is in 'initial update in progress' state.",
+          list_item->sensor_name, st->name);
     return;
   }
 
   vl.values = &(value_t){.gauge = value};
   vl.values_len = 1;
 
+  if (st->host != NULL)
+    sstrncpy(vl.host, st->host, sizeof(vl.host));
```
maryamtahhan: I think vl.host gets filled at a higher level usually? not sure if it's explicitly needed

src/ipmi.c
```diff
@@ -682,166 +815,410 @@ static void domain_connection_change_handler(ipmi_domain_t *domain, int err,
...
+        "user_data = %p);",
         (void *)domain, err, conn_num, port_num, still_connected, user_data);
 
-  status = ipmi_domain_add_entity_update_handler(
-      domain, domain_entity_update_handler, /* user data = */ NULL);
+  if (err != 0)
+    c_ipmi_error("domain_connection_change_handler", err);
+
+  if (!still_connected)
+    return;
```
maryamtahhan: would a simple info/debug message not be useful here?

src/ipmi.c
```diff
@@ -682,166 +815,410 @@ static void domain_connection_change_handler(ipmi_domain_t *domain, int err,
...
+  sfree(st->connaddr);
+  sfree(st->username);
+  sfree(st->password);
+
+  ignorelist_free(st->ignorelist);
+  pthread_mutex_destroy(&st->sensor_list_lock);
+  sfree(st);
+} /* void c_ipmi_free_instance */
+
+void c_ipmi_add_instance(c_ipmi_instance_t *instance) {
```
maryamtahhan: what about using a llist_t for instances?

src/ipmi.c
```diff
@@ -96,36 +106,81 @@ static void c_ipmi_error(const char *func, int status) {
...
+  char msg[1024];
+
+  vsnprintf(msg, sizeof(msg), format, ap);
+
+  switch (log_type) {
+  case IPMI_LOG_INFO:
+    INFO("ipmi plugin: %s", msg);
+    break;
+  case IPMI_LOG_WARNING:
+    NOTICE("ipmi plugin: %s", msg);
```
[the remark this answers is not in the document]
rpv-tomsk: Hi, Maryam! 

Thanks for your feedback.

The `OpenIPMI/ipmi_log.h` has the following explanation of error reporting levels:

```
/* General information about something that happens. */
    IPMI_LOG_INFO,

    /* An internal error (not reported directly to the user) occurred,
       but it's not a big deal, the system can still operate. */
    IPMI_LOG_WARNING,

    /* An internal error (not reported directly to the user) occured,
       and it will affect the operation of the system.  This system
       will still operation, but probably in a degraded mode. */
    IPMI_LOG_SEVERE,

    /* An internal error occured, and it is dangerous for the system
       to keep operating. */
    IPMI_LOG_FATAL,

    /* When returning errors to the user, this will often be used to
       report the general cause of the error if the cause would be
       difficult to determine from just the error return code. */
    IPMI_LOG_ERR_INFO,
```
By this explanation, and by typical use of Collectd `WARNING()` (), I think `NOTICE()` suits better.
Of course, I can be wrong...

It would be nice if Collectd will have the similar table of error logging level explanation. (I didn't find it.)

src/ipmi.c
rpv-tomsk: That was inspired by OpenIPMI examlpes code:

https://github.com/wrouesnel/openipmi/blob/master/sample/waiter_sample.c#L285
https://github.com/wrouesnel/openipmi/blob/master/cmdlang/ipmish.c#L595
https://github.com/wrouesnel/openipmi/blob/master/sample/eventd.c#L53
https://github.com/wrouesnel/openipmi/blob/master/include/OpenIPMI/ipmi_err.h#L161

I agree with your suggestion, that would be better to remove the magic.

src/ipmi.c
rpv-tomsk: Yes, you are right - vl.host can be filled by `hostname_g` at higher level.
But in this case we need to set `vl.host` of the metric to the value set in configuration, not to the `hostname_g` value.

src/ipmi.c
rpv-tomsk: At practice, that will be reported by a `c_ipmi_error()` call at several lines above.
Also, that will be reported by `c_ipmi_log()` handler too.

An example:
```
[2017-10-08 19:09:49] ipmi plugin: one 0 ipmi_lan.c(lost_connection): Connection 0 to the BMC is down
[2017-10-08 19:09:49] ipmi plugin: one 0 ipmi_lan.c(lost_connection): All connections to the BMC are down
[2017-10-08 19:09:49] ipmi plugin: domain_connection_change_handler failed for `one`: OS: Connection timed out
```

I have added a code to send Collectd notification in case of connection lost/restored and will publish it in updated PR.

src/ipmi.c
rpv-tomsk: Is there is a strong opinion/rules about this?
Why would you prefer a list instead of chain?

octo: There's no clear-cut advise/guideline for or against the generic linked list. I'd say that if you need to look up by (string) key or you'd need a double-linked list, then the generic implementation can abstract away some complexity.

HTH, —octo

src/ipmi.c
rpv-tomsk: The sources for `ipmi_get_error_string` return values are:

1) strerror() + 4 bytes of prefix ("OS: ")
2) ipmi_get_cc_string() (ipmi_ccodes[] array) + 6 bytes prefix - fits 128 bytes
3) rmcpp_error_codes[] + 7 bytes prefix  - fits 128 bytes
4) sol_error_codes[] + 5 bytes prefix    - fits 128 bytes
5) "Success (No error)" string           - fits 128 bytes
6) "Unknown: Unknown" string             - fits 128 bytes

The longest is strerror(). Will 1020 (1024 - 4) be enough for `strerror()`?

rpv-tomsk:
Updated.

src/ipmi.c
rpv-tomsk: >I'd say that if you need to look up by (string) key or you'd need a double-linked list
There is no such need here.

rpv-tomsk:
Also I'm going to add an option to disable all analog sensors polling.
So, the Collectd will handle only SEL and will send notifications only, without metrics.
That will be an good use-case of Collectd as IPMI events translator.

I have an quite old Intel SR2500 platform wich IPMI MC is stuck due to network polling. Now it does not respond to any requests - remote or local and server needs to be completely powered off (physically unplugged from outlet) to reboot MC.

src/ipmi.c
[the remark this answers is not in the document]
maryamtahhan: Fair enough

src/ipmi.c
maryamtahhan: cool

src/ipmi.c
```diff
@@ -292,55 +412,55 @@ static int sensor_list_add(ipmi_sensor_t *sensor) {
...
 
-  pthread_mutex_unlock(&sensor_list_lock);
+  pthread_mutex_unlock(&st->sensor_list_lock);
 
-  if (c_ipmi_notify_add && (c_ipmi_init_in_progress == 0)) {
+  if (st->notify_add && (st->init_in_progress == 0)) {
     notification_t n = {NOTIF_OKAY, cdtime(), "", "", "ipmi", "", "", "", NULL};
 
-    sstrncpy(n.host, hostname_g, sizeof(n.host));
+    sstrncpy(n.host, (st->host != NULL) ? st->host : hostname_g,
```
octo: This repeats at least two more times. Would it make sense to have a function that initializes a `notification_t` based on a `c_ipmi_instance_t *`? I.e.:

```c
static notification_t notification_init(c_ipmi_instance_t const *st);
/* … */
notification_t n = notification_init(st);
```

src/collectd.conf.pod
```diff
@@ -3381,6 +3409,11 @@ If a sensor disappears a notification is sent.
 If you have for example dual power supply and one of them is (un)plugged then
 a notification is sent.
 
+=item B<NotifyIPMIConnectionState> I<true>|I<false>
+
+If a IPMI connection state changes after initialization time of a minute
+a notification is sent. Defaults to B<false>.
+
```
octo: Please add a note about legacy support, i.e. that any option other than `Instance` will trigger legacy config handling which skipts the `<Instance>` block and that this support will go away in the next major version of collectd.

src/ipmi.c
```diff
@@ -138,43 +197,50 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
...
+           list_item->sensor_name, st->name);
+    } else if (IPMI_IS_IPMI_ERR(err) &&
+               IPMI_GET_IPMI_ERR(err) == IPMI_TIMEOUT_CC) {
+      INFO("ipmi plugin: sensor_read_handler: Sensor `%s` of `%s` timed out.",
+           list_item->sensor_name, st->name);
     } else {
+      char errbuf[ERR_BUF_SIZE] = {0};
+      ipmi_get_error_string(err, errbuf, sizeof(errbuf) - 1);
+
       if (IPMI_IS_IPMI_ERR(err))
```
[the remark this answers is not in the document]
rpv-tomsk: These messages are different in the second line of output format:

`"%s."` vs `"%s (%#x)."` and in a source of error code (`IPMI_GET_OS_ERR(err)` / `IPMI_GET_SOL_ERR(err)`).

The `ipmi_get_error_string()` does not always put error code into its output, so that appended locally for unified and more detailed message.

src/ipmi.c
```diff
@@ -682,166 +819,439 @@ static void domain_connection_change_handler(ipmi_domain_t *domain, int err,
...
   ipmi_domain_id_t domain_id;
   int status;
 
-  os_handler = ipmi_posix_thread_setup_os_handler(SIGIO);
-  if (os_handler == NULL) {
-    ERROR("ipmi plugin: ipmi_posix_thread_setup_os_handler failed.");
-    return -1;
-  }
+  if (st->connaddr != NULL) {
+    char *ip_addrs[1] = {NULL}, *ports[1] = {NULL};
```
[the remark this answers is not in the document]
rpv-tomsk: Thanks a lot for this note!

I think there will be a tiny memleak with these strdup(). 
Also, their return value is not checked and ... they are unneded at all.

src/ipmi.c
```diff
@@ -682,166 +819,439 @@ static void domain_connection_change_handler(ipmi_domain_t *domain, int err,
...
+  } else {
+    status = ipmi_smi_setup_con(/* if_num = */ 0, os_handler,
+                                /* user data = */ NULL, &st->connection);
+    if (status != 0) {
+      c_ipmi_error(st, "ipmi_smi_setup_con", status);
+      return -1;
+    }
   }
 
-  ipmi_open_option_t open_option[1] = {[0] = {.option = IPMI_OPEN_OPTION_ALL,
```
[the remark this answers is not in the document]
rpv-tomsk: Looks excellent!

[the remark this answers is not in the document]
rpv-tomsk: That looks excellent, but I can't compile this...
Code:
```
  ipmi_open_option_t opts[] = {
    {.option = IPMI_OPEN_OPTION_ALL, .ival = 1},
#ifdef IPMI_OPEN_OPTION_USE_CACHE
    /* OpenIPMI-2.0.17 and later: Disable SDR cache in local file */
    {.option = IPMI_OPEN_OPTION_USE_CACHE, .ival = 0},
#endif
  };
```
provides an error message:

```
src/ipmi.c:899: error: unknown field 'ival' specified in initializer
src/ipmi.c:899: error: missing braces around initializer
src/ipmi.c:899: error: (near initialization for 'opts[0].<anonymous>')
src/ipmi.c:902: error: unknown field 'ival' specified in initializer
```

I'm completely unfamiliar with a such syntax, sorry...

src/ipmi.c
```diff
@@ -682,166 +819,439 @@ static void domain_connection_change_handler(ipmi_domain_t *domain, int err,
...
 
   return 0;
 } /* int c_ipmi_read */
 
+static int c_ipmi_init(void) {
+  int status;
+  c_ipmi_instance_t *st;
+  char callback_name[3 * DATA_MAX_NAME_LEN];
+
+  os_handler = ipmi_posix_thread_setup_os_handler(SIGIO);
```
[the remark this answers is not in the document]
rpv-tomsk: Wow... That is unexpected.. 

We need to put a such nuances into 'Developers Guide'. I think we can easily found some plugins which needs this check to be added, if rules are strict. The `turbostat` is an example... Or a `tail_csv` has a broken check implementation.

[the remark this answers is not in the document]
rpv-tomsk: Can you please clarify, what are these cases?

I found `plugin_init_all` gets called twice, and call from `update_kstat()`  looks like an excessive call.

src/ipmi.c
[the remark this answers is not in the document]
octo: Oh, the struct has an unnamed union. This should work:

```c
ipmi_open_option_t opts[] = {
  {.option = IPMI_OPEN_OPTION_ALL, {.ival = 1}},
#ifdef IPMI_OPEN_OPTION_USE_CACHE
  /* OpenIPMI-2.0.17 and later */
  {.option = IPMI_OPEN_OPTION_USE_CACHE, {.ival = 1}},
#endif
};
ipmi_open_domain(/* … */, opts, STATIC_ARRAY_SIZE(opts), /* … */);
```

src/ipmi.c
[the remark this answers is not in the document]
octo: Systems that have the "kstat" service. "kstat" originates from Solaris but I think OpenBSD has it, too. Not sure about other BSDs. This is a legacy that I would really like to clean up, along with the `kc` global variable.

src/ipmi.c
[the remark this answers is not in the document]
octo: I was wondering about the `strdup()` ;) If they're not needed, all the better, you can just pass `&st->connaddr` to `ipmi_open_domain()`.

src/ipmi.c
```diff
@@ -682,166 +819,439 @@ static void domain_connection_change_handler(ipmi_domain_t *domain, int err,
...
+
+  st = instances;
+  while (NULL != st) {
+    /* The `st->name` is used as "domain name" for ipmi_open_domain().
+     * That value should be unique, so we do plugin_register_complex_read()
+     * at first as it checks the uniqueness. */
+    snprintf(callback_name, sizeof(callback_name), "ipmi/%s", st->name);
+
+    user_data_t ud = {
+        .data = st,
```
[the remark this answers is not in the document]
rpv-tomsk: We need to put such approach into style guide too..

The `free` callback does not _looks_ as expected place to stop threads ; -) 

I'm not against of a such change, but not this time (have no free time).

src/ipmi.c
```diff
@@ -682,166 +819,439 @@ static void domain_connection_change_handler(ipmi_domain_t *domain, int err,
...
+
+  while (st != NULL) {
+    c_ipmi_instance_t *next = st->next;
+
+    st->next = NULL;
+    st->active = 0;
+
+    if (st->thread_id != (pthread_t)0) {
+      pthread_join(st->thread_id, NULL);
+      st->thread_id = (pthread_t)0;
```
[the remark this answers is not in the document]
rpv-tomsk: Yes, I know that assigning `thread_id = 0` is not good enough.
Maybe replace this to `st->active == 1` check?

rpv-tomsk:
Hi, Florian!

Thanks for a review! I made some changes and push it here.
I will add `Port` option support later, most other suggestions was accepted and code improved.

src/ipmi.c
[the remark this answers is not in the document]
rpv-tomsk: >Systems that have the "kstat" service. 

How that works? Why two `init()` calls are needed in this case?

src/ipmi.c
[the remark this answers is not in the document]
rpv-tomsk: Thanks for update!

src/ipmi.c
rpv-tomsk: What do you think about `notification_init()` from `daemon/common.c`?
It used only once, in `virt` plugin.

src/ipmi.c
```diff
@@ -682,166 +814,427 @@ static void domain_connection_change_handler(ipmi_domain_t *domain, int err,
...
-  ipmi_init(os_handler);
-
-  status = ipmi_smi_setup_con(/* if_num = */ 0, os_handler,
-                              /* user data = */ NULL, &smi_connection);
-  if (status != 0) {
-    c_ipmi_error("ipmi_smi_setup_con", status);
-    return -1;
+  if (st->connaddr != NULL) {
+    status = ipmi_ip_setup_con(
+        &st->connaddr, (char * [1]){IPMI_LAN_STD_PORT_STR}, 1, st->authtype,
```
octo: Nit, at your option: I find `&(char *){IPMI_LAN_STD_PORT_STR}` easier to read.

src/ipmi.c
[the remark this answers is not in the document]
octo: `init()` is called every time the "kstat" structure changes, e.g. when devices are added to or removed from the system. As I said, it's a legacy and a bit broken and we should fix it, but it's out of scope for this PR.

src/ipmi.c
```diff
@@ -29,103 +30,166 @@
...
+  case IPMI_LOG_DEBUG_CONT:
+  case IPMI_LOG_DEBUG_END:
+    break;
+#endif
+  }
+} /* void c_ipmi_log */
+
+static notification_t c_ipmi_notification_init(c_ipmi_instance_t const *st,
+                                               int severity) {
+  notification_t n = {severity, cdtime(), "", "", "ipmi", "", "", "", NULL};
```
octo: At your option: this would be a bit nicer with a *named struct initializer*:

```c
notification_t n = {
  .severity = severity,
  .time = cdtime(),
  .plugin = "ipmi",
};
```

All the fields that are not explicitly initialized are guaranteed to be set to zero / NULL.

src/ipmi.c
rpv-tomsk: I tried to find this syntax, but with no success.
Thanks, amended.

octo:
> **rpv-tomsk** merged commit **0587ef9** into `collectd:master`

First <strike>post</strike> merge :tada:

rpv-tomsk:
Yeah! Many thanks to you for all what you have done for this!
And thanks for Collectd project itself!

maryamtahhan:
@rpv-tomsk congrats on the first merge :) Keep up the good work!

rpv-tomsk:
Thanks, Maryam!

PR #2027, opened 2016-11-06
modbus plugin: Avoid sizeof(struct sockaddr).

octo:
The previous version failed on FreeBSD with:

    modbus.c: In function 'mb_read_data':
    modbus.c:476:21: error: storage size of 'sockaddr' isn't known
         struct sockaddr sockaddr;
                         ^

### DIFF
--- a/src/modbus.c
+++ b/src/modbus.c
@@ -21,13 +21,14 @@
  **/
 
 #include "collectd.h"
-#include "common.h"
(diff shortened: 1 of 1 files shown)


octo:
The *modbus plugin* now builds successfully on FreeBSD 10, but the tests still fail due to #2029.

PR #2034, opened 2016-11-12
Pr/fix uptime plugin boottime

marcin1j:
Original uptime plugin implementation is susceptible to system date changes, which is likely to happen on embedded system without RTC. 
The cause of the problem is a combination of two factors:

1. the method used to calculate uptime (subtracting boottime from current time),
1. reading boottime only once, during plugin initialization phase.

This set of commits changes the implementation in the following way:
1. On Linux and AIX, uptime is read directly using a system call. A side effect is a simpler and faster code (no need to parse /proc filesystem, no double time() system call).
2. On BSD and Solaris, uptime is calculated using original method (subtracting boottime from current time).

### DIFF
--- a/src/uptime.c
+++ b/src/uptime.c
@@ -25,8 +25,7 @@
 #include "plugin.h"
 
 #if KERNEL_LINUX
-#define UPTIME_FILE "/proc/uptime"
-/* Using /proc filesystem to retrieve the boot time, Linux only. */
+#include <sys/sysinfo.h>
 /* #endif KERNEL_LINUX */
 
 #elif HAVE_LIBKSTAT
@@ -53,8 +52,6 @@
 /*
  * Global variables
  */
-/* boottime always used, no OS distinction */
-static time_t boottime;
 
 #if HAVE_LIBKSTAT
 extern kstat_ctl_t *kc;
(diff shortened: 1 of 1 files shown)


rubenk:
This breaks on Solaris:
```
uptime.c: In function ‘uptime_get_sys’:
uptime.c:88:11: error: implicit declaration of function ‘sysinfo’ [-Werror=implicit-function-declaration]
  status = sysinfo (&info);
           ^
uptime.c:97:24: error: ‘struct sysinfo’ has no member named ‘uptime’
  result = (time_t) info.uptime;
                        ^
```

marcin1j:
rubenk,

I've changed last commit to depend on KERNEL_LINUX and avoid conflict with Solaris sys/sysinfo.h. Could you confirm this solves the problem you mentioned?

marcin1j:
Please consider merging this commit. It's been waiting for almost a year. Since then, the source code was reformatted twice forcing me to update patch (I use this in a couple of production instances).

As for now, the PR is up-to-date with master branch.

rafwes:
@octo @zerkms

This has caused issues in lede/openwrt, why is this not upstream?

zerkms:
I'm just a couple-times contributor here without write permissions 🤷

marcin1j:
@octo 
I'll update it again (and hope this is the last time) but it needs to be tested first so give me a couple of hours.

marcin1j:
I've tested it. Everything seems OK.

octo:
Thanks @marcin1j!

Issue #1945, opened 2016-09-19
virt plugin: request for more info 


bronhaim:
-   Version of collectd: collectd-virt-5.5.2-1
-   Operating system / distribution: centos 7 and f24

Virt plugin provides the values that comes from "virsh dommemstat" - which includes the following:  - memory-actual_balloon, memory-rss, memory-swap_in, memory-total, cpu_total, disk opes and octets, and if dropped errors octets and packets.

Our monitoring requerement includes additional info that comes from "virsh domstats"
cpu.time, cpu.user, cpu.system,  balloon.current,  balloon.maximum. 

In addition if guest agent is installed on the vm we need to provide also mem_total , mem_unused, mem_free, swap_in swap_out, swap_total, swap_usage, mem_buffers, mem_cached, memUsage, migration progress and disksUsages. and for disks - read\writeLatency and flushLatency.

To fetch this info in the virt.c plugin we need additional requests to libvirt api. currently we use virDomainGetInfo and virDomainGetVcpus. I'm not sure exactly if its only one additional call to gather all the info or few, such as virDomainGetCPUStats and virDomainListGetStats

octo:
Hi @bronhaim,

thanks for opening this issue. What is your intend? Did you mean to open a feature request or would you like some design feedback before you start coding yourself?

Best regards,
—octo

bronhaim:
Thanks for the quick reply. I deploy collectd as part of my environment and need to fetch those values. This intention for opening to issue is to describe the need and get comments and suggestions for the implementation. Hopefully I'll pick it up later on, but I hoped to here if someone already started some work or know if there is a reason why those values are not presented in current implementation

octo:
When the *virt plugin* was written, *libvirt* was still a very young project. It's likely that this information was simply not available at the time. 

Right now the *virt plugin* has a read callback, `lv_read()` that is too long for my taste. I'd prefer if that was split up, for example into `lv_read_cpu()`, `lv_read_memory()`, `lv_read_disk()` and `lv_read_interface()`.

The metrics you need could then be added to these functions, hopefully not blowing them up too much. The swap metrics should probably go into their own function, `lv_read_swap()`.

It is against our best practices to report both, "cpu total" and the split up "cpu user / cpu system". The *CPU plugin* controls which one is reported with a config option called `ReportByState`.

Similarly, "mem_total" and "swap_total" should not be reported. Instead, calculate something like "swap_free = swap_total - swap_usage".

[...]

PR #2037, opened 2016-11-15
virt plugin: Rewrite from scratch

mojaves:
This patch rewrites the existing virt plugin
almost entirely, borrowing only few bits from
the existing one.

The plugin exports as superset of the metrics the
old plugin exported, and accepts a superset of the
configuration settings the old plugin expected, so
it is intended to act as drop-in replacement.

The new plugin now depends on glib2, libvirt and libxml2.
The new plugin has tests, taking inspiration from the tests
of the ceph plugin (src/ceph_test.c).

The gathering of the libvirt values has been completely overhauled,
and now we use the libvirt bulk stats API:
https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainListGetStats
which is a more modern and convenient way to fetch the
data from running libvirt domains.

We add resiliency in presence of unresponsive domains.
This can happen if a domain used shared storage (NFS, ISCSI)
and it experiences a network failure during one I/O operation.
check docs/README.virt.md to learn more about the issue
and the implemented solution.

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -1,5 +1,6 @@
 /**
  * collectd - src/virt.c
+ * Copyright (C) 2016 Francesco Romani <fromani at redhat.com>
  * Copyright (C) 2006-2008  Red Hat Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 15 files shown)


mojaves:
build fails because this patch requires libvirt >= 1.2.8. Unfortunately, we have
EPEL5 - libvirt 0.8.2
Precise - libvirt 0.9.8
Wheezy - libvirt 0.9.12
Trusty - libvirt 1.2.2

EPEL6 has libvirt 1.3.0 and fails because glib2.
@octo : are the supporte distro set going to change in the near future? Do the virt plugin need to support EPEL5, Precise and Wheezy?

src/virt.c
```diff
@@ -16,998 +17,900 @@
...
+             * http://libvirt.org/html/libvirt-libvirt-domain.html#virConnectGetAllDomainStats
+             */
+            .stats = 0,
+            .flags = 0,
+        },
+};
 
-    plugin_dispatch_values (&vl);
-}
+static virt2_context_t *virt2_get_default_context() { return &default_context; }
```
pszczerx: Shouldn't it return const virt2_context_t pointer?

src/virt.c
```diff
@@ -16,998 +17,900 @@
...
-static int nr_domains = 0;
+                                   "PluginInstanceFormat",
 
-static void free_domains (void);
-static int add_domain (virDomainPtr dom);
+                                   "Instances",
+                                   "DebugPartitioning",
+                                   NULL};
+#define NR_CONFIG_KEYS                                                         \
+  ((sizeof virt2_config_keys / sizeof virt2_config_keys[0]) - 1)
```
pszczerx: I think you could leverage STATIC_ARRAY_SIZE macro here.

src/virt.c
```diff
@@ -16,998 +17,900 @@
...
-
-static int
-lv_init (void)
-{
-    if (virInitialize () != 0)
-        return -1;
-    else
-        return 0;
+static int virt2_dispatch_block(virt2_instance_t *inst, const VMInfo *vm) {
+  value_t vals[2];
```
pszczerx: Magic number. You could use an enum to specify array size and particular indexes i.e BLOCK_STATS_RD_IDX = 0, BLOCK_STATS_WR_IDX = 1,  BLOCK_STATS_NR = 2. Such approach could be used with other metrics like interface stats.

src/vminfo.c
```diff
@@ -0,0 +1,353 @@
...
+#undef TRY_TO_PARSE
+
+static void vcpuinfo_free(VCpuInfo *vcpu) { free(vcpu->xstats); }
+
+static void blockinfo_free(BlockInfo *block) {
+  size_t i;
+  const BlockStats *stats = (block->xstats) ? block->xstats : block->stats;
+
+  for (i = 0; i < block->nstats; i++)
+    free(stats[i].xname);
```
pszczerx: Using sfree is recommended.

src/vminfo.h
```diff
@@ -0,0 +1,150 @@
...
+  BlockStats *xstats;
+  BlockStats stats[BLOCK_STATS_NUM];
+};
+
+typedef struct IFaceStats IFaceStats;
+struct IFaceStats {
+  char *xname;
+  char name[STATS_NAME_LEN];
+
+  unsigned long long rx_bytes;
```
pszczerx: Seems like you are redefining libvirt structure here. Any reason not to use in this case virDomainInterfaceStatsStruct? This comment also applies to other structures like BlockStats.

pszczerx:
@mojaves thanks for opening this PR. I left a few comments, but of course a maintainer needs to take a look at this.

@octo: Could you answer @mojaves question regarding distros? We are planning on extending virt plugin with a couple more metrics and some of them require libvirt >= 2.0.0.

rubenk:
@mojaves we tend to support those distros until they go EOL.
RHEL6 still has libvirt 0.10 and is still widely used.

rubenk:
@mojaves first of all thanks for the PR. I haven't looked at the code in depth yet, but I have some initial questions about the design and the issue that this PR is trying to solve.

You write:
> All the above considered, one robust management or monitoring application must deal with the fact that the libvirt API can block for a long time, or forever. This is not an issue or a bug of one specific
> API, but it is rather a byproduct of how libvirt and QEMU interact.

Strictly speaking I think this is an issue with the libvirt API, and a client shouldn't have to work around this.

I must say I'm a bit surprised that this is needed. I've seen calls to libvirt waiting for a lock, especially on slow storage or during live migration. Any calls during that time from the libvirt plugin just time out trying to acquire libvirts state lock. Not ideal, but worst case you loose a few stats for a specific domU.

I tried to find more details about this in the libvirt api docs, but couldn't find any.
Do you know if this is at least a known issue for the libvirt developers and if so if there are any plans to improve this behaviour?

Now if I understand your proposed solution correctly, instead of using one read callback in total you partition the vms by tags and use one read callback per tag. It seems to me that this doesn't completely solve the problem, it only limits the impact to a specific read callback. Since it seems it api calls can lock for many reasons, in practice you would need a read callback per vm instance to make sure you don't miss any stats.

mojaves:
@rubenk:
> we tend to support those distros until they go EOL.
Yes, I see your point and I accept the maintainers' decision; I still can post all the improvements I'd like while supporting ancient libvirts, with some extra work.

> RHEL6 still has libvirt 0.10 and is still widely used.
Yes, I know, and it's pretty painful to have to work with so ancient libvirts; libvirt >= 1.3.0 a truckload of fixes and improvements which makes work with it much more convenient. Bulk stats is a good example. With bulk stats, you just feed libvirt with an array of domains, and you get all you need back conveniently and easily. Compare with the libvirt 0.8.2 code on which you have to scan the domain xml to keep track of active interfaces, block devices and query the individual devices (there is also room for a little race here, but it's not that important).

BTW, just for the record, I'm actually more bothered by RHEL5 (libvirt 0.8.2) and Precise (libvirt 0.9.2).

mojaves:
@rubenk 

> >  All the above considered, one robust management or monitoring application must deal with the 
> > fact that the libvirt API can block for a long time, or forever. This is not an issue or a bug of one 
> > specific  API, but it is rather a byproduct of how libvirt and QEMU interact.
>
> Strictly speaking I think this is an issue with the libvirt API, and a client shouldn't have to work 
> around this.

Correct. Me and my team filed quite some bugs in the past about this (probably the best solution could be https://bugzilla.redhat.com/show_bug.cgi?id=1113106) but this ultimately is hard to solve, so
it's taking quite a lot of time. And meantime, I believe, a well-behaving client has to deal with this unfortunate behaviour :\

> I must say I'm a bit surprised that this is needed. 
It largely depends on how the storage is configured and behaves.
Well, it's not actually needed if one can accept some holes in the monitoring data. Sometimes this could be acceptable. 

> I've seen calls to libvirt waiting for a lock, especially on slow storage or during live migration.
> Any calls during that time from the libvirt plugin just time out trying to acquire libvirts state lock.
> Not ideal, but worst case you loose a few stats for a specific domU.

Exactly, worst case we lose some data, but in the oVirt project (http://www.ovirt.org - disclosure: I work full time in this project) we try hard to minimize those events.

> I tried to find more details about this in the libvirt api docs, but couldn't find any.
> Do you know if this is at least a known issue for the libvirt developers and if so if there are any plans 
> to improve this behaviour?

Not sure why this is not documented, but it is well known. In oVirt we have plenty of bugs reported to deal with those cases of unresponsive storage, and we implemented more aggressive workarounds, which I believe are not well suited elsewhere.

You can learn a bit more about the behaviour and the backstory from here:
https://bugzilla.redhat.com/show_bug.cgi?id=1392316
Please note that I fully agree this bug (filed by me) is too generic, I filed it too hastily hoping to reopen the conversation, but it's just better to file more fine grained bugs with more precise requests (on it already).

> Now if I understand your proposed solution correctly, instead of using one read callback in total you 
> partition the vms by tags and use one read callback per tag.

Yep, correct.

> It seems to me that this doesn't completely solve the problem, it only limits the impact to a specific 
> read callback. Since it seems it api calls can lock for many reasons, in practice you would need a 
> read callback per vm instance to make sure you don't miss any stats.

Correct, but this approach has one big drawback: the number of reader threads must be adjusted dynamically. it's the *calling thread* that could be blocked, not the single instance; this dynamic adjustment would greatly complicate the plugin for something that, as you pointed out, it's ultimately a workaround.

My solution is a partial fix, no question about that, but I believe it's a good compromise between added complexity and robustness. We leverage the fact that in the scenarios where we seen the blocking happen, the domains used shared storage and groups of VMs (domains) all use the same network storage. If VMs A, B, C all connect to storage D over the network and one I/O operation for B fails, then most likely also A and C I/O will soon fail, so it's acceptable to neglect them.

rubenk:
Thanks for your elaborate reply @mojaves!

> Yes, I know, and it's pretty painful to have to work with so ancient libvirts; libvirt >= 1.3.0 a truckload of fixes and improvements which makes work with it much more convenient.

I fully agree, but unfortunately it's the reality that we have to stay reasonably backwards compatible.
We have been discussing what 'reasonably' means recently. We certainly don't want to keep the same support cycles for collectd as those of enterprise distros.

> BTW, just for the record, I'm actually more bothered by RHEL5 (libvirt 0.8.2) and Precise (libvirt 0.9.2).

I think 5.7 will be the last release supporting RHEL5, since it goes EOL next year, so I'm fine with it if you don't spend too much effort on it. We'd love to keep support for RHEL6 for a while though, especially since  a lot of people use our upstream packages on it, since I'll never upgrade collectd in EPEL6 to Collectd5.

But back to the issue at hand. As far as I see it this PR does multiple things:
- Switches to the bulk stats api. That's a nice improvement and indeed closes a small race, but well'll have to support the older api anyway, so I'm not sure what that gains us now.
- Add support for more statistics (afaik just balloon and perf), but I believe that can also be done with the old api's.
- It works around a blocking api calls on shared storage, but this doesn't help for users who run their vms on local storage, or even distributed storage like Ceph.

It does all that by rewriting the complete plugin. While I love that you added tests, this is still a risky thing and we lose all bug fixes and testing that went into the current plugin.

You obviously went through a lot of effort writing this code, and I appreciate that, but I wonder if this effort wouldn't have been better spend improving the api from the other side. That would benefit the collectd plugin, for all use cases, and every other managing and monitoring system that uses the libvirt api out there. I know from personal experience that the libvirt guys are a friendly bunch and always willing to accept patches if they solve real issues.

mojaves:
@rubenk 

> But back to the issue at hand. As far as I see it this PR does multiple things:
>  Switches to the bulk stats api. That's a nice improvement and indeed closes a small race, but
>  well'll have to support the older api anyway, so I'm not sure what that gains us now.

Yes, zero gains given the constraints. Furthermore, there is no point in having two different code paths, so bulk stats code should go :(

>  Add support for more statistics (afaik just balloon and perf), but I believe that can also be done 
> with the old api's.

True. This is actually a mistake of mine: I squashed here a patch that was supposed to come later, and I intentionally didn't push a revert to not make an already complex PR even harder to read. As you pointed out, this is easy to achieve in the old plugin, so I'll send a new PR later, to partially solve https://github.com/collectd/collectd/issues/1945. I'm saying "partially" because some requested stats IMO doesn't fit in the virt plugin -either the existing one or the rewritten one, but let's discuss this in another PR.

> It works around a blocking api calls on shared storage, but this doesn't help for users who run their
> vms on local storage, or even distributed storage like Ceph.
Correct about ceph/distributed storage. I have too little experience here to claim this approuch would help -still investigating here-, so I don't do any claim :)

I'm not sure what you mean about local storage. Local storage could be unresponsive (in the time scale of ten of seconds or minutes, at least)  only in case of catastrophic failures, like SATA/SCSI errors or bad blocks. On the other hand, network outages with shared storage, while still relatively rare, are a more frequent issue.

I'd like to point out that this solution could scale to help for local storage; we could address this in two ways: either use one reader thread per VM (wasteful, needs dynamic tuning of number of threads, but the safest), and the tagging code I introduced here will make sure that each thread always monitors one VM; or we could consider the local storage as one single storage shared among all the VMs. In this case nothing could help :)

But I think it all boils down to one question...

> It does all that by rewriting the complete plugin. While I love that you added tests, this is still a risky > thing and we lose all bug fixes and testing that went into the current plugin.

I see your motivation, so here's my aforementioned question: Let's forget this PR and the rewrite. Would you accept one patch implementing the aforementioned tagging/partitioning strategy on top of the existing virt plugin, or there is no chance?

I'd like to stress that I implemented the domain tag/partitioning inside the virt plugin; then yes, I want to use this to workaround libvirt responsiveness.
But the key change is that I want to pin one libvirt domain on one collectd reader thread. This is what we would like to add and that I'd like to implement. Are you willing to consider this addition?

> You obviously went through a lot of effort writing this code, and I appreciate that, but I wonder if this 
> effort wouldn't have been better spend improving the api from the other side. 

The two efforts aren't mutually exclusive! :) We already have a conversation going about this topic, but those aren't easy issues to tackle.

Let me provide some rationale for the existing approach.
First, this approach could be implemented on top of libvirt 0.8.2 (albeit with more work), and would add robustness to old clients. I don't claim will fix all the issues, but it will improve greatly.
Second, this is way harder than it could look, because we need to change all the stack down to the QEMU monitor protocol; for example, adding asynchronous monitoring commands, and their handling in libvirt. There were long discussion in the past, but since this change needs to touch very core areas of the QEMU emulator, it's not easy to deliver.

> That would benefit the collectd plugin, for all use cases, and every other managing and monitoring 
> system that uses the libvirt api out there. I know from personal experience that the libvirt guys are a 
> friendly bunch and always willing to accept patches if they solve real issues.

Yep, I share the same experience, but in this case the real issue is how the QEMU monitor protocol works. This, for obvious backward compatibility reasons, is hard to change.
Eventually one solution will come, but it could be months ahead; and years ahead since we (oVirt does monitoring as well, so we are on the same boat) could consume it, much like today the project supports EL5. What to do in the meantime?

rubenk:
@rubenk 

Hi Francesco!

>> But back to the issue at hand. As far as I see it this PR does multiple things:
>>  Switches to the bulk stats api. That's a nice improvement and indeed closes a small race, but
>>  well'll have to support the older api anyway, so I'm not sure what that gains us now.

>Yes, zero gains given the constraints. Furthermore, there is no point in having two different code >paths, so bulk stats code should go :(

I'm sorry.. perhaps we could revisit this in a year or so when RHEL8 should be around the corner and we can drop support for old libvirt versions.

>>  Add support for more statistics (afaik just balloon and perf), but I believe that can also be done 
>> with the old api's.

>True. This is actually a mistake of mine: I squashed here a patch that was supposed to come  later, and I intentionally didn't push a revert to not make an already complex PR even harder to read. As you pointed out, this is easy to achieve in the old plugin, so I'll send a new PR later, to  partially solve https://github.com/collectd/collectd/issues/1945. I'm saying "partially" because  some requested stats IMO doesn't fit in the virt plugin -either the existing one or the rewritten  one, but let's discuss this in another PR.

Sounds good to me :)

>> It works around a blocking api calls on shared storage, but this doesn't help for users who run their
>> vms on local storage, or even distributed storage like Ceph.

> Correct about ceph/distributed storage. I have too little experience here to claim this approuch would help -still investigating here-, so I don't do any claim :)

Luckily librbd, the library qemu uses to talk to Ceph, has safe timeouts all over the place. I don't have much experience with it either, but haven't seen any I/O hang.


> I'm not sure what you mean about local storage. Local storage could be unresponsive (in the  time scale of ten of seconds or minutes, at least)  only in case of catastrophic failures, like  SATA/SCSI errors or bad blocks. On the other hand, network outages with shared storage, while  still relatively rare, are a more frequent issue.

You're right. I've had a lot of raid cards resetting themselves as of late. But when that happens too often, some missing stats are the least of my problems.

In any case, we should assume that any call to libvirt can hang, for whatever reason, on local, distributed or shared storage, or while doing live (storage) migrations.

> I'd like to point out that this solution could scale to help for local storage; we could address this > in two ways: either use one reader thread per VM (wasteful, needs dynamic tuning of number of > threads, but the safest), and the tagging code I introduced here will make sure that each thread > always monitors one VM; or we could consider the local storage as one single storage shared  among all the VMs. In this case nothing could help :)

I wonder how wasteful your first solution would be in practice. True, we need to dynamically change the number of threads as vms come and go, but the threads aren't doing much most of the time.

> But I think it all boils down to one question...

>> It does all that by rewriting the complete plugin. While I love that you added tests, this is still a risky > thing and we lose all bug fixes and testing that went into the current plugin.

>I see your motivation, so here's my aforementioned question: Let's forget this PR and the rewrite. Would you accept one patch implementing the aforementioned tagging/partitioning strategy on top of the existing virt plugin, or there is no chance?

>I'd like to stress that I implemented the domain tag/partitioning inside the virt plugin; then yes, I want to use this to workaround libvirt responsiveness.
But the key change is that I want to pin one libvirt domain on one collectd reader thread. This is what we would like to add and that I'd like to implement. Are you willing to consider this addition?

Certainly.

>> You obviously went through a lot of effort writing this code, and I appreciate that, but I wonder if this 
>> effort wouldn't have been better spend improving the api from the other side. 

>The two efforts aren't mutually exclusive! :) We already have a conversation going about this topic, but those aren't easy issues to tackle.

>Let me provide some rationale for the existing approach.
>First, this approach could be implemented on top of libvirt 0.8.2 (albeit with more work), and would add robustness to old clients. I don't claim will fix all the issues, but it will improve greatly.
>Second, this is way harder than it could look, because we need to change all the stack down to the QEMU monitor protocol; for example, adding asynchronous monitoring commands, and their handling in libvirt. There were long discussion in the past, but since this change needs to touch very core areas of the QEMU emulator, it's not easy to deliver.

I realize what I'm asking, and that this is no easy thing. It's great to hear that you already started a conversation.

>> That would benefit the collectd plugin, for all use cases, and every other managing and monitoring 
>> system that uses the libvirt api out there. I know from personal experience that the libvirt guys are a 
>> friendly bunch and always willing to accept patches if they solve real issues.

>Yep, I share the same experience, but in this case the real issue is how the QEMU monitor protocol works. This, for obvious backward compatibility reasons, is hard to change.
>Eventually one solution will come, but it could be months ahead; and years ahead since we (oVirt does monitoring as well, so we are on the same boat) could consume it, much like today the project supports EL5. What to do in the meantime?

Please understand that I don't push back against this rewrite just to be a pain ;)
Any code that we accept in collectd has to be maintained for years to come too, and it's better to have this discussion now than after the code ends up in any distributions.

Just a last thought, and perhaps not the best of solutions, but have you considered carrying this patch as a separate plugin in the oVirt repositories?

mojaves:
@rubenk
Hi Ruben!

> In any case, we should assume that any call to libvirt can hang, for whatever reason, on local, 
> distributed or shared storage, or while doing live (storage) migrations.

Yep, that's the unfortunate reality we are dealing with :\

> I wonder how wasteful your first solution would be in practice. True, we need to dynamically change
> the number of threads as vms come and go, but the threads aren't doing much most of the time.

From a resource usage perspective it should be manageable. I'm actually much more worried about the thread pool management: who and how should do the resize?

> > I see your motivation, so here's my aforementioned question: Let's forget this PR and the rewrite.
> > Would you accept one patch implementing the aforementioned tagging/partitioning strategy on 
> > top of the existing virt plugin, or there is no chance?
> > 
> > I'd like to stress that I implemented the domain tag/partitioning inside the virt plugin;
> > then yes, I want to use this to workaround libvirt responsiveness.
> > But the key change is that I want to pin one libvirt domain on one collectd reader thread.
> > This is what we would like to add and that I'd like to implement.
> > Are you willing to consider this addition?
> 
> Certainly.

Awesome! So please check https://github.com/collectd/collectd/pull/2048 :)

> > Please understand that I don't push back against this rewrite just to be a pain ;)
> > Any code that we accept in collectd has to be maintained for years to come too,
> > and it's better to have this discussion now than after the code ends up in any distributions.

I totally understand. I posted the PR even though it wasn't 100% ready exactly because I thought that code demonstrate the purpose and the extent of the changes way better than any description in english prose.

I did the rewrite for the ease of development _and_ because I wasn't aware you need to support libvirt back to 0.8.2. My fault, I should have been done better research :)

But since we can deliver the features/fixes on top of the existing codebase, a backport is ok for me. 

> Just a last thought, and perhaps not the best of solutions, but have you considered carrying this 
> patch as a separate plugin in the oVirt repositories?

Yes. I have code almost ready for that purpose but we *strongly* prefer to patch collectd and to have the changes upstreamed;
I don't mind some extra work to achieve this goal.

The out of tree plugin is my last resort, e.g. if you or some other maintainer definitely NACK the concept of domain tag/partition - or other changes oVirt needs.

mojaves:
I'd like to contribute those changes using finer-grained pull requests.

PR #2038, opened 2016-11-15
dpdkstat: Added support for dpdk 16.07.

taraschornyi:
Fixed issue with different linker flags for dpdk 16.04
and dpdk 16.07 on Ubuntu

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2602,7 +2602,9 @@ rte_version_minor=""
 if test "x$with_libdpdk" = "xyes"
 then
   SAVE_CPPFLAGS="$CPPFLAGS"
+  SAVE_LDFLAGS="$CPPFLAGS"
   CPPFLAGS="$CPPFLAGS $with_libdpdk_cppflags"
+  LDFLAGS="$LDFLAGS $with_libdpdk_ldflags"
   AC_RUN_IFELSE([AC_LANG_PROGRAM(
     [[#include <rte_version.h>]],
     [[return (RTE_VER_YEAR & 0xFF)]])],
@@ -2612,20 +2614,18 @@ then
     [[return (RTE_VER_MONTH & 0xFF)]])],
     [rte_version_minor=0], [rte_version_minor=$?])
   CPPFLAGS="$SAVE_CPPFLAGS"
+  LDFLAGS="$SAVE_LDFLAGS"
 fi
 
 if test "x$with_libdpdk" = "xyes"
 then
     BUILD_WITH_DPDK_LDFLAGS="$with_libdpdk_ldflags"
-    if test $rte_version_major -eq 16
+    if test "$rte_version_major" -eq 16 && test "$rte_version_minor" -le 4
     then
-        if test $rte_version_minor -le 4
-        then
-            BUILD_WITH_DPDK_LDFLAGS+=" -Wl,--no-as-needed"
-        fi
+        BUILD_WITH_DPDK_LDFLAGS="$BUILD_WITH_DPDK_LDFLAGS -Wl,--no-as-needed"
     fi
-    BUILD_WITH_DPDK_CFLAGS+="-fPIC $with_libdpdk_cppflags"
-    BUILD_WITH_DPDK_LIBS="-Wl,--whole-archive,-ldpdk -Wl,--no-whole-archive "
+    BUILD_WITH_DPDK_CFLAGS="$BUILD_WITH_DPDK_CFLAGS $with_libdpdk_cppflags"
+    BUILD_WITH_DPDK_LIBS="-Wl,--whole-archive -Wl,-ldpdk -Wl,--no-whole-archive"
     AC_SUBST(BUILD_WITH_DPDK_CFLAGS)
     AC_SUBST(BUILD_WITH_DPDK_LDFLAGS)
     AC_SUBST(BUILD_WITH_DPDK_LIBS)
--- a/configure.ac
+++ b/configure.ac
@@ -2545,77 +2545,90 @@ fi
 # }}}
 
 # --with-libdpdk {{{
+with_libdpdk_cppflags=""
+with_libdpdk_ldflags=""
 AC_ARG_WITH(libdpdk, [AS_HELP_STRING([--with-libdpdk@<:@=PREFIX@:>@], [Path to the DPDK build directory.])],
 [
 	if test "x$withval" != "xno" && test "x$withval" != "xyes"
 	then
-		RTE_BUILD="$withval"
+		with_libdpdk_cppflags="-I$withval/include"
+		with_libdpdk_ldflags="-L$withval/lib"
 		with_libdpdk="yes"
 	else
-		RTE_BUILD="/usr"
-		with_libdpdk="$withval"
+		if test "x$withval" = "xno"
+		then
+			with_libdpdk="no (disabled)"
+		fi
 	fi
-	DPDK_INCLUDE="$RTE_BUILD/include"
-	DPDK_LIB_DIR="$RTE_BUILD/lib"
-	FOUND_DPDK=yes
 ], [with_libdpdk="no"])
 
 if test "x$with_libdpdk" = "xyes"
 then
-	LOCAL_DPDK_INSTALL="no"
-	AC_CHECK_HEADER([$DPDK_INCLUDE/rte_config.h], [LOCAL_DPDK_INSTALL=yes],
-		[AC_CHECK_HEADER([$DPDK_INCLUDE/dpdk/rte_config.h],
-		[],
-		[FOUND_DPDK=no], [])], [])
+	SAVE_CPPFLAGS="$CPPFLAGS"
+	CPPFLAGS="$CPPFLAGS $with_libdpdk_cppflags"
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -2545,77 +2545,90 @@ fi
...
-		DPDK_INCLUDE=$DPDK_INCLUDE/dpdk
-	fi
+	SAVE_CPPFLAGS="$CPPFLAGS"
+	CPPFLAGS="$CPPFLAGS $with_libdpdk_cppflags"
+	AC_CHECK_HEADERS(rte_config.h, [with_libdpdk="yes"],
+		[
+			AC_CHECK_HEADERS(dpdk/rte_config.h,
+				[
+					with_libdpdk="yes"
+					with_libdpdk_cppflags="$with_libdpdk_cppflags/dpdk"
```
octo: That doesn't make sense to me: if building with `$with_libdpdk_cppflags` succeeds, then modify the variable? Seems like the `/dpdk` suffix should go on line 2554.

configure.ac
```diff
@@ -2545,77 +2545,90 @@ fi
...
+    [[#include <rte_version.h>]],
+    [[return (RTE_VER_MONTH & 0xFF)]])],
+    [rte_version_minor=0], [rte_version_minor=$?])
+  CPPFLAGS="$SAVE_CPPFLAGS"
+fi
+
+if test "x$with_libdpdk" = "xyes"
+then
+    BUILD_WITH_DPDK_LDFLAGS="$with_libdpdk_ldflags"
+    if test $rte_version_major -eq 16
```
octo: Portability issue: please quote variable expansions:

```
if test "$rte_version_major" -eq 16 && test "$rte_version_minor -le 4
then
  # …
fi
```

configure.ac
```diff
@@ -2545,77 +2545,90 @@ fi
...
+    then
+        if test $rte_version_minor -le 4
+        then
+            BUILD_WITH_DPDK_LDFLAGS+=" -Wl,--no-as-needed"
```
octo: Portability issue: `+=` is not portable (here and below):

```
BUILD_WITH_DPDK_LDFLAGS="$BUILD_WITH_DPDK_LDFLAGS -Wl,--no-as-needed"
```

configure.ac
```diff
@@ -2545,77 +2545,90 @@ fi
...
         fi
-	AC_SUBST(BUILD_WITH_DPDK_CFLAGS)
-	AC_SUBST(BUILD_WITH_DPDK_LDFLAGS)
-	AC_SUBST(BUILD_WITH_DPDK_LIBS)
+    fi
+    BUILD_WITH_DPDK_CFLAGS+="-fPIC $with_libdpdk_cppflags"
+    BUILD_WITH_DPDK_LIBS="-Wl,--whole-archive,-ldpdk -Wl,--no-whole-archive "
```
octo: Is the comma between "whole-archive" and "-ldpdk" intentional?

Nit: trailing whitespace inside the string.

src/dpdkstat.c
```diff
@@ -525,6 +560,14 @@ static int dpdk_helper_run(void) {
                 i, len);
           break;
         }
+ #if RTE_VERSION >= RTE_VERSION_16_07
+          ret = rte_eth_xstats_get_names(i, g_configuration->xnames + num_xstats,
```
octo: Indentation is off here. Please re-run `clang-format -style=LLVM` on this file.

configure.ac
```diff
@@ -2545,77 +2545,90 @@ fi
...
+    then
+        if test $rte_version_minor -le 4
+        then
+            BUILD_WITH_DPDK_LDFLAGS+=" -Wl,--no-as-needed"
         fi
-	AC_SUBST(BUILD_WITH_DPDK_CFLAGS)
-	AC_SUBST(BUILD_WITH_DPDK_LDFLAGS)
-	AC_SUBST(BUILD_WITH_DPDK_LIBS)
+    fi
+    BUILD_WITH_DPDK_CFLAGS+="-fPIC $with_libdpdk_cppflags"
```
rubenk: I already mentioned this in the original review, but please drop the -fPIC.

configure.ac
```diff
@@ -2545,77 +2545,90 @@ fi
...
 then
-	BUILD_WITH_DPDK_CFLAGS+="-I$DPDK_INCLUDE"
-	if test "x$LOCAL_DPDK_INSTALL" != "xyes"
-	then
-		BUILD_WITH_DPDK_LDFLAGS="-Wl,--no-as-needed"
-	else
-		BUILD_WITH_DPDK_LDFLAGS="-L$DPDK_LIB_DIR -Wl,--no-as-needed"
+  SAVE_CPPFLAGS="$CPPFLAGS"
+  CPPFLAGS="$CPPFLAGS $with_libdpdk_cppflags"
+  AC_RUN_IFELSE([AC_LANG_PROGRAM(
```
rubenk: Either use AC_COMPILE_IFELSE, or add the correct linker flags here, otherwise you can end up testing the wrong libdpdk.

configure.ac
taraschornyi: Dpdk headers can be installed directly to include dir or to dpdk sub-folder.
If check in line 2571 is true this mean that  dpdk headers are installed in dpdk sub-folder, 
We are using $with_libdpdk_cppflags later for geting dpdk versions

taraschornyi:
Hi @octo @rubenk 
Thanks for your comments. I've just pushed patch addressing most of your comments. And answered to those which were not addressed

configure.ac
rubenk: I think it's disgusting ;)
Seriously, what problem are you trying to solve?

> If rte_xxx.h includes others rte_xxx.h (e.g.: using just #include <rte_xxx.h>) the compilation will fail.

Then either those headers are buggy, or we're including from the wrong level.

> Dpdk headers can be installed directly to include dir or to dpdk sub-folder.

I checked the dpdk-devel package on Fedora, and it has the headers in /usr/include/dpkd/
So we should add /usr/include/dpkd to the include path and check for rte_config.h instead of checking for dpdk/rte_config.h. If I a user wants to build dpdk from source, then we should allow him to specify the include path, and check for the same header there. Alternatively allow him to override CPPFLAGS and make it the users problem. Presumably if you can build dpdk from source, you can figure out what CPPFLAGS you need. We shouldn't add hacks to configure.ac to cater to this use case.

vmytnykx: Hi @rubenk,
We are trying to resolve `--with-libdpdk=yes` use case and avoid any hard-coded system path like `/usr/include` + prefix `/dpdk`. Since, using hard-coded path `/usr/include` is fine in this case, I think we can stay with this straightforward approach. Thank you.

rubenk: I don't think you can. As you said yourself, since the headers in /usr/include/dpkd depend on other headers in that same directory, you need to have /usr/include/dpkd in your include path. And how about on other operating systems, for instance on FreeBSD where includes are in /usr/local/include?

vmytnykx: We may keep the list of standard directories [/usr/, /usr/local, ...] in configure.ac and just iterate through all of them trying to find the rte_config.h header.
Also, as a second solution it can be the path provided above but it has been refused ;)
And as a 3-rd option I see here is to disable `--with-libdpdk=yes` config option at all and ask user to provide the dpdk library path all the time.
What do you think?

rubenk: There's another option, and that is to fix the library itself. We don't have this issue with all the other libraries we use that install headers in a subdirectory of /usr/include.

Applications should be able to consume the library by just including a single header that takes care of including the rest.

configure.ac
```diff
@@ -2545,77 +2545,90 @@ fi
...
+  CPPFLAGS="$SAVE_CPPFLAGS"
+  LDFLAGS="$SAVE_LDFLAGS"
+fi
+
+if test "x$with_libdpdk" = "xyes"
+then
+    BUILD_WITH_DPDK_LDFLAGS="$with_libdpdk_ldflags"
+    if test "$rte_version_major" -eq 16 && test "$rte_version_minor" -le 4
+    then
+        BUILD_WITH_DPDK_LDFLAGS="$BUILD_WITH_DPDK_LDFLAGS -Wl,--no-as-needed"
```
rubenk: BUILD.dpdkstat.md mentions:
```
The *dpdkstat plugin* has the following requirements:

 * DPDK 16.04 or later
 * GCC 4.9 or later
```

We also agreed to this during the inital review, so please remove this check. You then also don't need to do the version checks.

configure.ac
```diff
@@ -2545,77 +2545,90 @@ fi
...
+    fi
+    BUILD_WITH_DPDK_CFLAGS="$BUILD_WITH_DPDK_CFLAGS $with_libdpdk_cppflags"
+    BUILD_WITH_DPDK_LIBS="-Wl,--whole-archive -Wl,-ldpdk -Wl,--no-whole-archive"
```
rubenk: Why is this needed?

rubenk:
@taraschornyi In the name of progress merged just the C-code changes to master in 4f9418d694d31137b11100298d8e2c18f717ccb5 since these are fine and allow me to test with dpdk 16.11.

I have some ideas for the configure.ac changes coming up.

octo:
Superseded by #2046

PR #2040, opened 2016-11-16
Naming threads

mfournier:
This is a rebased & updated version of #547.

I've been using this patchset for a few months now (on Linux only though), and haven't encountered any problem with it.

Thanks a lot @manuelluis !

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1662,6 +1662,48 @@ AC_CHECK_MEMBERS([kstat_io_t.nwritten, kstat_io_t.writes, kstat_io_t.nwrites, ks
 #endif
 	])
 
+# check for pthread_setname_np
+SAVE_LDFLAGS=$LDFLAGS
+LDFLAGS="$LDFLAGS -lpthread"
+
+AC_MSG_CHECKING([if have pthread_setname_np])
+	have_pthread_setname_np="no"
+	AC_LINK_IFELSE([AC_LANG_PROGRAM(
+[[
+#if HAVE_PTHREAD_H
+#include <pthread.h>
+#endif
+]],
+[[
+        pthread_setname_np(pthread_self(), "conftest");
+]]
+	)], [
+		have_pthread_setname_np="yes"
+		AC_DEFINE(HAVE_PTHREAD_SETNAME_NP, 1, [ pthread_setname_np() is available.])
+	])
+
+AC_MSG_RESULT([$have_pthread_set_name_np])
+
+# check for pthread_set_name_np
+AC_MSG_CHECKING([if have pthread_set_name_np])
+	have_pthread_set_name_np="no"
+	AC_LINK_IFELSE([AC_LANG_PROGRAM(
+[[
+#if HAVE_PTHREAD_H
+#include <pthread.h>
+#endif
+]],
+[[
+        pthread_set_name_np(pthread_self(), "conftest");
+]]
+	)], [
+		have_pthread_set_name_np="yes"
+		AC_DEFINE(HAVE_PTHREAD_SET_NAME_NP, 1, [ pthread_set_name_np() is available.])
+	])
+AC_MSG_RESULT([$have_pthread_set_name_np])
+
+LDFLAGS=$SAVE_LDFLAGS
+
 #
 # Checks for libraries begin here
 #
--- a/configure.ac
+++ b/configure.ac
@@ -214,7 +214,54 @@ AC_HEADER_SYS_WAIT
 AC_HEADER_DIRENT
 AC_HEADER_STDBOOL
 
-AC_CHECK_HEADERS(stdio.h errno.h math.h stdarg.h syslog.h fcntl.h signal.h assert.h sys/types.h sys/socket.h sys/select.h poll.h netdb.h arpa/inet.h sys/resource.h sys/param.h kstat.h regex.h sys/ioctl.h endian.h sys/isa_defs.h fnmatch.h libgen.h pthread_np.h)
+AC_CHECK_HEADERS([ \
+  arpa/inet.h \
+  assert.h \
+  ctype.h \
+  endian.h \
+  errno.h \
+  fcntl.h \
+  fnmatch.h \
+  fs_info.h \
+  fshelp.h \
+  grp.h \
+  kstat.h \
+  kvm.h \
+  libgen.h \
+  limits.h \
+  locale.h \
+  math.h \
+  mntent.h \
+  mnttab.h \
+  netdb.h \
+  paths.h \
+  poll.h \
+  pthread_np.h \
+  pwd.h \
+  regex.h \
+  signal.h \
+  stdarg.h \
+  stdio.h \
+  sys/fs_types.h \
+  sys/fstyp.h \
+  sys/ioctl.h \
+  sys/isa_defs.h \
+  sys/mntent.h \
+  sys/mnttab.h \
+  sys/param.h \
+  sys/resource.h \
+  sys/select.h \
+  sys/socket.h \
+  sys/statfs.h \
+  sys/statvfs.h \
+  sys/types.h \
+  sys/un.h \
+  sys/vfs.h \
+  sys/vfstab.h \
+  sys/vmmeter.h \
+  syslog.h \
+  wordexp.h \
+])
 
 # For entropy plugin on newer NetBSD
 AC_CHECK_HEADERS(sys/rndio.h, [], [],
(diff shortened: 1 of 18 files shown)


octo:
pthread_setname_np(3):
> CONFORMING TO
> These functions are nonstandard GNU extensions.

octo:
Compilation has a warning:

```
conftest.c:181:9: warning: implicit declaration of function 'pthread_set_name_np' [-Wimplicit-function-declaration]
         pthread_set_name_np(pthread_self(), "conftest");
         ^
```

But linking succeeds:

```
configure:23467: $? = 0
configure:23477: result: yes
```

I think the `HAVE_PTHREAD_H` guard is not yet defined at this point in the configure script. Additionally, FreeBSD requires `<pthread_np.h>` for `pthread_set_name_np()`.

configure.ac
```diff
@@ -1662,6 +1662,45 @@ AC_CHECK_MEMBERS([kstat_io_t.nwritten, kstat_io_t.writes, kstat_io_t.nwrites, ks
 #endif
 	])
 
+# check for pthread_setname_np
+SAVE_LDFLAGS="$LDFLAGS"
+LDFLAGS="$LDFLAGS -lpthread"
+
+AC_MSG_CHECKING([if have pthread_setname_np])
```
rubenk: if we?

configure.ac
```diff
@@ -1662,6 +1662,45 @@ AC_CHECK_MEMBERS([kstat_io_t.nwritten, kstat_io_t.writes, kstat_io_t.nwrites, ks
...
+]]
+	)], [
+		have_pthread_setname_np="yes"
+		AC_DEFINE(HAVE_PTHREAD_SETNAME_NP, 1, [pthread_setname_np() is available.])
+	])
+
+AC_MSG_RESULT([$have_pthread_setname_np])
+
+# check for pthread_set_name_np(3) (FreeBSD)
+AC_MSG_CHECKING([if have pthread_set_name_np])
```
rubenk: if we

configure.ac
```diff
@@ -1662,6 +1662,45 @@ AC_CHECK_MEMBERS([kstat_io_t.nwritten, kstat_io_t.writes, kstat_io_t.nwrites, ks
...
 	])
 
+# check for pthread_setname_np
+SAVE_LDFLAGS="$LDFLAGS"
+LDFLAGS="$LDFLAGS -lpthread"
+
+AC_MSG_CHECKING([if have pthread_setname_np])
+	have_pthread_setname_np="no"
+	AC_LINK_IFELSE([AC_LANG_PROGRAM(
+[[
```
rubenk: Can you indent this block? It used to be the case that older compilers needed this indented completely to the left, but that's not the case anymore.

src/daemon/plugin.c
```diff
@@ -2789,9 +2819,10 @@ static void *plugin_thread_start (void *arg)
 } /* void *plugin_thread_start */
 
 int plugin_thread_create (pthread_t *thread, const pthread_attr_t *attr,
-		void *(*start_routine) (void *), void *arg)
+		void *(*start_routine) (void *), void *arg, char *name)
```
rubenk: can you make this const?

src/daemon/plugin.c
```diff
@@ -2789,9 +2819,10 @@ static void *plugin_thread_start (void *arg)
...
 {
 	plugin_thread_t *plugin_thread;
+	int ret;
```
rubenk: doing 'int ret = pthread_create' is fine now.

src/network.c
```diff
@@ -3477,7 +3477,7 @@ static int network_init (void)
 		status = plugin_thread_create (&dispatch_thread_id,
 				NULL /* no attributes */,
 				dispatch_thread,
-				NULL /* no argument */);
+				NULL /* no argument */, "network dispatch");
```
rubenk: This is 1 byte too long and will be truncated. Not a big issue, but perhaps we can come up with something shorter?

src/unixsock.c
```diff
@@ -459,7 +459,7 @@ static int us_init (void)
 	loop = 1;
 
 	status = plugin_thread_create (&listen_thread, NULL,
-			us_server_thread, NULL);
+			us_server_thread, NULL, "unixsock listener");
```
rubenk: this will also be truncated.

configure.ac
```diff
@@ -214,7 +214,7 @@ AC_HEADER_SYS_WAIT
 AC_HEADER_DIRENT
 AC_HEADER_STDBOOL
 
-AC_CHECK_HEADERS(stdio.h errno.h math.h stdarg.h syslog.h fcntl.h signal.h assert.h sys/types.h sys/socket.h sys/select.h poll.h netdb.h arpa/inet.h sys/resource.h sys/param.h kstat.h regex.h sys/ioctl.h endian.h sys/isa_defs.h fnmatch.h libgen.h)
+AC_CHECK_HEADERS(stdio.h errno.h math.h stdarg.h syslog.h fcntl.h signal.h assert.h sys/types.h sys/socket.h sys/select.h poll.h netdb.h arpa/inet.h sys/resource.h sys/param.h kstat.h regex.h sys/ioctl.h endian.h sys/isa_defs.h fnmatch.h libgen.h pthread_np.h)
```
rubenk: This line is getting awfully long. Do you mind breaking it up?

configure.ac
octo: I think the usual wording is "for <something>", so that the log message reads "checking for pthread_set_name_np …"

rubenk: You're right, as usual ;)

src/daemon/plugin.c
```diff
@@ -2801,8 +2832,22 @@ int plugin_thread_create (pthread_t *thread, const pthread_attr_t *attr,
...
 	plugin_thread->arg           = arg;
 
-	return pthread_create (thread, attr,
+	ret = pthread_create (thread, attr,
 			plugin_thread_start, plugin_thread);
+
+	if (ret == 0 && name != NULL) {
+#if defined(HAVE_PTHREAD_SETNAME_NP) || defined(HAVE_PTHREAD_SET_NAME_NP)
+		char thread_name[16];
+		sstrncpy (thread_name, name, sizeof(thread_name));
```
octo: Why this copy in the first place? As @rubenk mentions, this leads to truncation and it's not clear to me why it would be necessary. It can't really be memory ownership as `thread_name` goes out of scope immediately after the `setname()` call …

src/daemon/plugin.c
```diff
@@ -2801,8 +2832,22 @@ int plugin_thread_create (pthread_t *thread, const pthread_attr_t *attr,
 	plugin_thread->start_routine = start_routine;
 	plugin_thread->arg           = arg;
 
-	return pthread_create (thread, attr,
+	ret = pthread_create (thread, attr,
 			plugin_thread_start, plugin_thread);
+
+	if (ret == 0 && name != NULL) {
```
octo: I'd prefer if an error was handled here:

```c
int ret = pthread_create();
if (ret != 0)
  return ret;

if (name != NULL) {
  // …
}
```

src/daemon/plugin.c
octo: Indeed, Linux / glibc even returns `ERANGE` if the name it too long. This should be commented in the code.

configure.ac
octo: Any way we indent this it's ugly. I've left it as is because I found the other indenting even more unreadable …

rubenk: Yeah alright, it's never going to be beautiful. It does help if we indent everything the same way, as I did in #1865.

src/daemon/plugin.c
```diff
@@ -651,7 +651,31 @@ static void *plugin_read_thread (void __attribute__((unused)) *args)
...
-static void start_read_threads (int num)
+static void set_thread_name(pthread_t tid, char const *name) {
+#if defined(HAVE_PTHREAD_SETNAME_NP) || defined(HAVE_PTHREAD_SET_NAME_NP)
+#if !defined(HAVE_PTHREAD_SETNAME_NP)
+# define pthread_setname_np pthread_set_name_np
+#endif
+
+	/* glibc limits the length of the name and fails if the passed string
+	 * is too long, so we truncate it here. */
+	char n[THREAD_NAME_MAX];
```
rubenk: I might be missing something, but where is THREAD_NAME_MAX defined?

octo: Meh, I've moved the code around a bit and it was left in the wrong place. Fixed.

octo:
Thanks @manuelluis, @mfournier and @rubenk!

PR #2045, opened 2016-11-19
README: do not point users to non-existing file

rubenk:


octo:
Hi @rubenk,

the `INSTALL` file is generated by the autotools and included in the release tarballs. It is not included in the Git repository though. How about clarifying that this is a generated file rather then removing the reference entirely?

Best regards,
—octo

rubenk:
Hey @octo
> the INSTALL file is generated by the autotools and included in the release tarballs
That's strange? We build with 'foreign' in AM_INIT_AUTOMAKE, and at least on my system, make dist doesn't produce it.

octo:
I checked the release tarballs available from the homepage, and they do include the file. No idea about the details of `AM_INIT_AUTOMAKE` though ;)

rubenk:
'foreign' should prevent the creation of this file (and NEWS and other GNU-specific things)
I just did make dist in a checkout of the 5.6.1, and no INSTALL file.. Perhaps you're using an older version of automake than I am?

rubenk:
It is gone in the 5.7.0 tarball:
```
$ tar tvf collectd-5.7.0.tar.bz2 | grep INSTALL
```

PR #2048, opened 2016-11-21
Add support for virt domain tags

mojaves:
After the discussion in https://github.com/collectd/collectd/pull/2037 , here's the same code implemented on top of the existing codebase.
As explained in https://github.com/collectd/collectd/pull/2037 , we are considering to switch to collectd in oVirt for monitoring hosts and libvirt domains in future releases.

We can't just neglect to monitor all the domains if one fraction thereof is unresponsive, so we need to enhance the monitoring code to work with that.

We will implement that on top of the more generic domain tag/partition approach I implemented here.
This patch required some refactoring, but I tried to keep this minimal.

I choose to upload few patches to make the review easier, but I have zero problems squashing them.

More code could be backported from https://github.com/collectd/collectd/pull/2037 - tests and documentations, and it will be done with later patches or later PRs.

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -73,24 +73,11 @@ static ignorelist_t *il_interface_devices = NULL;
 static int ignore_device_match(ignorelist_t *, const char *domname,
                                const char *devpath);
 
-/* Actual list of domains found on last refresh. */
-static virDomainPtr *domains = NULL;
-static int nr_domains = 0;
-
-static void free_domains(void);
-static int add_domain(virDomainPtr dom);
-
 /* Actual list of block devices found on last refresh. */
 struct block_device {
   virDomainPtr dom; /* domain */
   char *path;       /* name of block device */
 };
-
-static struct block_device *block_devices = NULL;
-static int nr_block_devices = 0;
-
-static void free_block_devices(void);
-static int add_block_device(virDomainPtr dom, const char *path);
 
 /* Actual list of network interfaces found on last refresh. */
 struct interface_device {
@@ -100,12 +87,33 @@ struct interface_device {
   char *number;     /* interface device number */
 };
 
-static struct interface_device *interface_devices = NULL;
-static int nr_interface_devices = 0;
+struct lv_read_state {
+  /* Actual list of domains found on last refresh. */
+  virDomainPtr *domains;
+  int nr_domains;
 
-static void free_interface_devices(void);
-static int add_interface_device(virDomainPtr dom, const char *path,
-                                const char *address, unsigned int number);
+  struct block_device *block_devices;
+  int nr_block_devices;
+
+  struct interface_device *interface_devices;
+  int nr_interface_devices;
+};
+
+static struct lv_read_state read_state = {
+    NULL, 0, NULL, 0, NULL, 0,
+};
+
+static void free_domains(struct lv_read_state *state);
+static int add_domain(struct lv_read_state *state, virDomainPtr dom);
+
+static void free_block_devices(struct lv_read_state *state);
+static int add_block_device(struct lv_read_state *state, virDomainPtr dom,
+                            const char *path);
+
+static void free_interface_devices(struct lv_read_state *state);
+static int add_interface_device(struct lv_read_state *state, virDomainPtr dom,
+                                const char *path, const char *address,
+                                unsigned int number);
 
 /* HostnameFormat. */
 #define HF_MAX_FIELDS 3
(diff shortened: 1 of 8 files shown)


src/virt.c
```diff
@@ -135,8 +164,8 @@ static enum plginst_field plugin_instance_format[PLGINST_MAX_FIELDS] =
 
 /* BlockDeviceFormat */
 enum bd_field {
-	target,
-	source
+        target,
+        source
```
rubenk: If you're fixing the indentation, might as well make it 2 spaces like the other enums :)

src/virt.c
```diff
@@ -480,16 +500,41 @@ lv_config (const char *key, const char *value)
...
     return -1;
 }
 
 static int
-lv_read (void)
+lv_read (user_data_t *ud)
 {
     time_t t;
+    struct lv_read_instance *inst = ud->data;
+    struct lv_read_state *state = &inst->read_state;
```
rubenk: Move this down below the check on the follow line?

src/virt.c
```diff
@@ -102,11 +92,50 @@ struct interface_device {
...
+struct lv_user_data {
+  struct lv_read_instance inst;
+  user_data_t ud;
+};
 
-static void free_interface_devices (void);
-static int add_interface_device (virDomainPtr dom, const char *path, const char *address, unsigned int number);
+#define NR_INSTANCES_DEFAULT 1
+#define NR_INSTANCES_MAX 128
+static size_t nr_instances = NR_INSTANCES_DEFAULT;
```
rubenk: size_t is a bit large for something with a max of 128. Just make this an int?

rubenk:
Hi @mojaves!

Thanks, I think this looks good, I just left some minor suggestions inline.
I think at minimum this needs some documentation for the tag format.

If you don't mind I'll wait with merging this until after we branch 5.7, which should be in a week or two.

mojaves:
@rubenk sure, there is no rush. I'll address your comments, backport docs and tests (in followup patches) and do more testing myself.

src/virt.c
mojaves: done in a followup patch, could be squashed if you wish

src/virt.c
mojaves: right, fixed

src/virt.c
mojaves: why not, I'm checking boundaries anyway in lv_config.

mojaves:
@rubenk 
Let me document my last updates:

- Addressed your comments - with followup patches, hopefully this is easier to review; feel free to squash if you prefer, either way is fine for me

- Backported the tests , but disabled them (see https://github.com/collectd/collectd/pull/2048/commits/f246c0df8de83a1635bbdecd8f88a29d1763ea53 to see the reason)

- Added documentation (https://github.com/collectd/collectd/pull/2048/commits/b3eb9e778149be7cacb7e1bfd375a8f33c2790b7)

- And finally manage to squash one bug which exists even in the old code (https://github.com/collectd/collectd/pull/2048/commits/32ad52204ef1cfa28eee8501a443d312c20e327e)

I've done testing on my side - both with backward-compatible and with the new tag support enable with no issues.

mojaves:
@rubenk @octo  I guess I need to rebase after the reformatting, right?

rubenk:
@mojaves yes please, and if you could also pull your changes through clang-format that would be great.

mojaves:
@rubenk rebase done. Unfortunately I had to force push, and this messed up the order of commits.
Here's the expected order:

999efc2 virt plugin: Document the partition/tag support
0936a9e virt plugin: Properly reset the lists
6188d7d virt plugin: Disable tests momentarily
98e38ad virt plugin: Add tests for the domain tags
19828ad virt plugin: Minor fixes
a726d25 virt plugin: Use int, for the number of instances
c84646c virt plugin: Avoid NULL pointer deference
452324c virt plugin: Add support for domain tags
9f9af17 virt plugin: Support for multiple read callbacks
2846042 virt plugin: Factor the read state into a struct

Please also note that "virt plugin: Fix bad rebase of virt_test" was squashed into "virt plugin: Add tests for the domain tags" and that "virt plugin: Fix the indentation of enums" was dropped because is no longer relevant

rubenk:
Thanks @mojaves!

rubenk:
Hi @mojaves, I was just looking at the error handling in lv_domain_get_tag, specifically https://github.com/collectd/collectd/blob/master/src/virt.c#L794.
On error, lv_domain_get_tag jumps to the done label, but we always return the result of xmlXPathRegisterNs().

Also, the err = 0 assignment on https://github.com/collectd/collectd/blob/master/src/virt.c#L779 is a no-op.

 Would you mind sending a follow up PR that fixes this?

Thanks!

mojaves:
@rubenk sure thing: https://github.com/collectd/collectd/pull/2093

PR #2050, opened 2016-11-22
Adding IPV6 support to ip_vs plugin

ben-loadbalancer:
To retrieve IPV6 stats from ipvs requires the use of netlink as well as a newer ip_vs.h than what is in most systems. We check for different versions of netlink and have included a newer ip_vs header to work with netlink.
This is based on the work by Vince Busam adding IPV6 support to ipvsadm.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -629,7 +629,7 @@ fi
 
 AM_CONDITIONAL(IP_VS_H_NEEDS_KERNEL_CFLAGS, test "x$ip_vs_h_needs_kernel_cflags" = "xyes")
 PKG_CHECK_MODULES(LIBNL_GENL3, libnl-genl-3.0, [have_libnl_genl3=yes], [have_libnl_genl3=no])
-PKG_CHECK_MODULES(LIBNL3, libnl-3.0 >= 3.1, [have_libnl3=yes], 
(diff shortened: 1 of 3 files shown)


octo:
Hi @ben-loadbalancer, thank you very much for your patch! I'm a bit skeptical about including a copy of system headers in collectd. I take it that newer distributions have the required header? If so, I'd prefer not to support old distributions over copying system header files.

ben-loadbalancer:
Hi octo, I started out by copying what was done in IPVSADM which the original plugin is based on. Here they included ip_vs.h which I thought was compatibility issues across versions. However after looking further the included version does not match any of the system files across the versions. 
The version included has the necessary data structures required for IPV6 so unfortunately is required or needs to be split up which will require additional work to maintain. I've also approached one of the kernel dev's to look at updating the system header but if this happens it will not be for a while.

rubenk:
@ben-loadbalancer are you planning on following up on this?

ben-loadbalancer:
@rubenk Yes I do, Unfortunately the workload has pilled up so I hope to get back to it soon/

rubenk:
@ben-loadbalancer, no problem, please open a new PR whenever you're ready.

PR #2051, opened 2016-11-22
virt plugin: get domains using virConnectListAllDomains

dsilakov:
virConnectListAllDomains was introduced in libvirt almost 4 years ago. Rationale for it can be found in the original RFE: https://www.redhat.com/archives/libvir-list/2012-May/msg00976.html


rubenk:
I'm sorry, but we can't apply this as is. This breaks on versions where this API is not available.

dsilakov:
Ok, I'll take a look if it is possible to add this in a more graceful way.

src/virt.c
```diff
@@ -705,19 +725,24 @@ refresh_lists (void)
...
             const char *name;
             char *xml = NULL;
             xmlDocPtr xml_doc = NULL;
             xmlXPathContextPtr xpath_ctx = NULL;
             xmlXPathObjectPtr xpath_obj = NULL;
 
+#ifdef HAVE_LIST_ALL_DOMAINS
+            virDomainPtr dom = domains[i];
+#else
+            virDomainPtr dom = NULL;
```
rubenk: these assignments are no-ops and immediately overwritten below.

rubenk:
hi @dsilakov, you might want to hold off with this PR for a bit or discuss with @mojaves since work is  ongoing in the virt plugin in #2037 and #2048

dsilakov:
Yeah, I'll take a look at that PRs. At least they also add switch to modern libvirt API.

rubenk:
the other PR's are fixed now, it would be nice if you could rebase this.

rubenk:
Indeed, the test failure is unrelated.

PR #2052, opened 2016-11-22
Fix Collectd::Unixsock

mbethke:
I just noticed that at some point the unixsock module's reporting of fractional seconds broke `Collectd::Unixsock::listval_filter` as the regexp for the data lines didn't match any more so nothing was returned. This fixes this bug.
Also updated the mock daemon to provide realistic values (with fractional parts and without in turn so as to test compatibility with both) and fixed some indentation to use all tabs as the rest of the module does.

### DIFF
--- a/bindings/perl/lib/Collectd/MockDaemon.pm
+++ b/bindings/perl/lib/Collectd/MockDaemon.pm
@@ -56,8 +56,9 @@ sub daemon {
 
 sub listval {
     my $trunc_now = substr(time, 0, -1);
+    my $toggle;
     return print_nvalues(scalar @metrics) .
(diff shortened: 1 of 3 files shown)


octo:
Hi @mbethke, thank you very much for your patch. Overall it looks great, but I think using random numbers on tests is bad style because tests are not deterministic then. Could you please replace them with a static list of (random) values and, while at it, remove the time() call as well?

Thanks!
—octo

mbethke:
Sure, makes complete sense! No use trying to be super realistic when you need reproducible results :)
Will push in a sec.

octo:
Thanks @mbethke!

PR #2053, opened 2016-11-24
netlink plugin: Adding support of drops value monitoring

nuclearcat:
Monitoring drops is very handy to see drop rate of qdisc/class (to adjust length of queue and to estimate how much data are exceeding limit), and in some cases to limit interface drops that doesn't reach ethstat and other statistics.

### DIFF
--- a/src/netlink.c
+++ b/src/netlink.c
@@ -539,7 +539,7 @@ static int qos_filter_cb(const struct nlmsghdr *nlh, void *args) {
         submit_one(dev, "ipt_packets", type_instance, bs->packets);
       }
       if (qs != NULL) {
-        submit_one(dev, "if_tx_errors", type_instance, qs->drops);
+        submit_one(dev, "if_tx_dropped", type_instance, qs->drops);
       }
     }
 
--- a/src/netlink.c
+++ b/src/netlink.c
@@ -85,6 +85,11 @@ typedef struct ir_ignorelist_s {
   struct ir_ignorelist_s *next;
 } ir_ignorelist_t;
 
+struct qos_stats {
+  struct gnet_stats_basic **bs;
+  struct gnet_stats_queue **qs;
+};
+
 static int ir_ignorelist_invert = 1;
 static ir_ignorelist_t *ir_ignorelist_head = NULL;
 
@@ -387,7 +392,9 @@ static int link_filter_cb(const struct nlmsghdr *nlh,
 
 #if HAVE_TCA_STATS2
 static int qos_attr_cb(const struct nlattr *attr, void *data) {
-  struct gnet_stats_basic **bs = (struct gnet_stats_basic **)data;
+  struct qos_stats *qdisc = (struct qos_stats *)data;
+  struct gnet_stats_basic **bs = qdisc->bs;
+  struct gnet_stats_queue **qs = qdisc->qs;
 
   /* skip unsupported attribute in user-space */
   if (mnl_attr_type_valid(attr, TCA_STATS_MAX) < 0)
@@ -400,7 +407,17 @@ static int qos_attr_cb(const struct nlattr *attr, void *data) {
       return MNL_CB_ERROR;
     }
     *bs = mnl_attr_get_payload(attr);
-    return MNL_CB_STOP;
+    return MNL_CB_OK;
+  }
+
+  if (mnl_attr_get_type(attr) == TCA_STATS_QUEUE) {
+    if (mnl_attr_validate2(attr, MNL_TYPE_UNSPEC, sizeof(**qs)) < 0) {
+      ERROR("netlink plugin: qos_attr_cb: TCA_STATS_QUEUE mnl_attr_validate2 "
+            "failed.");
+      return MNL_CB_ERROR;
+    }
+    *qs = mnl_attr_get_payload(attr);
+    return MNL_CB_OK;
   }
 
   return MNL_CB_OK;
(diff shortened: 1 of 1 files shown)


src/netlink.c
```diff
@@ -559,9 +581,9 @@ static int qos_filter_cb (const struct nlmsghdr *nlh, void *args)
       return MNL_CB_ERROR;
     }
 
-    mnl_attr_parse_nested (attr, qos_attr_cb, &bs);
+    mnl_attr_parse_nested (attr, qos_attr_cb, &q_stats);
```
octo: At your option: optimize away the temporary *q_stats* variable:

```c
mnl_attr_parse_nested (attr, qos_attr_cb, &(struct qos_stats) {
      .bs = &bs,
      .qs = &qs,
    });
```

src/netlink.c
```diff
@@ -570,8 +592,13 @@ static int qos_filter_cb (const struct nlmsghdr *nlh, void *args)
...
           tc_type, tc_inst);
 
-      submit_one (dev, "ipt_bytes", type_instance, bs->bytes);
-      submit_one (dev, "ipt_packets", type_instance, bs->packets);
+      if (bs != NULL) {
+        submit_one (dev, "ipt_bytes", type_instance, bs->bytes);
+        submit_one (dev, "ipt_packets", type_instance, bs->packets);
+      }
+      if (qs != NULL) {
+        submit_one (dev, "if_tx_errors", type_instance, qs->drops);
```
octo: Wouldn't `if_tx_dropped` be a more fitting type for this metric?

nuclearcat: Yes, it is better, just i used value that i found in older types.db

src/netlink.c
octo: Alright. In that case (temporary variable), could you please remove one level of indirection? The pointer-pointer inside the struct is then unnecessary and the struct can be simplified to:

```c
struct qos_stats {
    struct gnet_stats_basic *bs;
    struct gnet_stats_queue *qs;
};
```

nuclearcat:
Done, i changed metric and removed one level of indirection

src/netlink.c
```diff
@@ -559,9 +581,9 @@ static int qos_filter_cb (const struct nlmsghdr *nlh, void *args)
       return MNL_CB_ERROR;
     }
 
-    mnl_attr_parse_nested (attr, qos_attr_cb, &bs);
+    mnl_attr_parse_nested (attr, qos_attr_cb, &q_stats);
 
-    if (bs != NULL)
+    if (bs != NULL || qs != NULL)
```
octo: Both, `bs` and `qs` will *always* be `NULL`. You need to check `q_stats.bs` and `q_stats.qs` here. The `bs` and `qs` variables are unused and can be deleted.

src/netlink.c
```diff
@@ -430,21 +435,34 @@ static int link_filter_cb (const struct nlmsghdr *nlh,
...
   {
-    if (mnl_attr_validate2 (attr, MNL_TYPE_UNSPEC, sizeof (**bs)) < 0)
+    if (mnl_attr_validate2 (attr, MNL_TYPE_UNSPEC, sizeof (*bs)) < 0)
     {
       ERROR ("netlink plugin: qos_attr_cb: TCA_STATS_BASIC mnl_attr_validate2 failed.");
       return MNL_CB_ERROR;
     }
-    *bs = mnl_attr_get_payload (attr);
-    return MNL_CB_STOP;
+    bs = mnl_attr_get_payload (attr);
```
octo: You need to assign `qdisc->bs` here. `bs` is unused and can be removed.

src/netlink.c
```diff
@@ -430,21 +435,34 @@ static int link_filter_cb (const struct nlmsghdr *nlh,
...
+  }
+
+  if (mnl_attr_get_type (attr) == TCA_STATS_QUEUE)
+  {
+    if (mnl_attr_validate2 (attr, MNL_TYPE_UNSPEC, sizeof (*qs)) < 0)
+    {
+      ERROR ("netlink plugin: qos_attr_cb: TCA_STATS_QUEUE mnl_attr_validate2 failed.");
+      return MNL_CB_ERROR;
+    }
+    qs = mnl_attr_get_payload (attr);
```
octo: You need to assign `qdisc->qs` here. `qs` is unused and can be removed.

src/netlink.c
```diff
@@ -549,6 +567,10 @@ static int qos_filter_cb (const struct nlmsghdr *nlh, void *args)
   mnl_attr_for_each (attr, nlh, sizeof (*tm))
   {
     struct gnet_stats_basic *bs = NULL;
+    struct gnet_stats_queue *qs = NULL;
+    struct qos_stats q_stats;
```
octo: This variable is called `qdisk` above and `q_stats` here. Please pick one and use the same name in both places.

nuclearcat:
Sorry for mistakes, unable to concentrate due problems at work.
Can you check if it is fine now?  Tested in pre-production server, as i see it works fine.
I guess when it is ok, probably better to resubmit pull request?

mcq8:
Seems like the same thing we want to do in https://github.com/collectd/collectd/pull/1479
Is it possible to add qlen, backlog, requeues and overlimits as well?

nuclearcat:
Sure it's possible, but your patch looks fine, it is strange it didn't went thru since 2015!
Probably we need some advice from maintainers, if they are preferring any patch.

octo:
@nuclearcat sorry for the late response, I lost track of this :-( the patch looks good note, thank you!

@mcq8 sorry your PR didn't get any attention for so long :-( does it make sense to rebase it?

mcq8:
@octo I rebased it to the current master and changed it to the new codestyle.

octo:
I've rebased your branch on *master*. Once the build checks succeed this is good to go.

mcq8:
@octo this PR or https://github.com/collectd/collectd/pull/1479? I would like to have the qlen, backlog, requeues and overlimits.

octo:
@mcq8 no, the branch of this PR. #1479 is currently pending for your replies ;-) If you want you can rebase #1479 on top of this commit, but I'm happy to handle the merge conflicts when the review is done, too.

PR #2066, opened 2016-11-29
contrib/docker: cleanup wrapper a bit

rubenk:
- No need to declare errno
- Remove one level of indentation

### DIFF
--- a/contrib/docker/rootfs_prefix/rootfs_prefix.c
+++ b/contrib/docker/rootfs_prefix/rootfs_prefix.c
@@ -11,28 +11,22 @@
 #define BUFSIZE 256
 
 const char *add_prefix(const char *orig, char *prefixed) {
-  int status;
-  int errno;
-
-  if ((strncmp(orig, "/proc", 5) == 0) || (strncmp(orig, "/sys", 4) == 0)) {
-
-    status = snprintf(prefixed, BUFSIZE, "%s%s", PREFIX, orig);
-    if ((unsigned int)status >= BUFSIZE) {
-      error(status, ENAMETOOLONG,
-            "'%s' got truncated when adding '%s' prefix: '%s'", orig, PREFIX,
-            prefixed);
-      return orig;
-    } else if (status < 1) {
-      error(status, errno,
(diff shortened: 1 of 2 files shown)


rubenk:
@mfournier what do you think?

contrib/docker/rootfs_prefix/rootfs_prefix.c
```diff
@@ -11,28 +11,22 @@
...
-      return orig;
-    } else {
-      return (const char *)prefixed;
-    }
+  if ((strncmp(orig, "/proc", 5) != 0) && (strncmp(orig, "/sys", 4) != 0))
+    return orig;
 
-  } else {
+  int status = snprintf(prefixed, BUFSIZE, "%s%s", PREFIX, orig);
+  if ((unsigned int)status >= BUFSIZE) {
```
octo: You should do the `status < 1` check first: status may be negative due to error. When cast to an *unsigned int*, the resulting value is likely larger than `BUFSIZE`, resulting in an incorrect error message.

contrib/docker/rootfs_prefix/rootfs_prefix.c
```diff
@@ -11,28 +11,22 @@
...
-      return orig;
-    } else if (status < 1) {
-      error(status, errno,
-            "adding '%s' prefix to file path failed: '%s' -> '%s'", PREFIX,
-            orig, prefixed);
-      return orig;
-    } else {
-      return (const char *)prefixed;
-    }
+  if ((strncmp(orig, "/proc", 5) != 0) && (strncmp(orig, "/sys", 4) != 0))
```
octo: I find `strlen("/proc")` easier to read than `5`.

contrib/docker/rootfs_prefix/rootfs_prefix.c
rubenk: Ah duh, thanks for spotting this. Wil fix.

cwolfinger:
Doesnt this leak memory on each time the file is open ?

Issue #2067, opened 2016-11-30
Missing dependencies for exec and tail.


OmegaVVeapon:
*   Version of collectd: collectd-5.6.1.339.g797ed5e-2.el7
*   Operating system / distribution: CentOS 7.1.1503

## Expected behavior

Collectd should be able to find the dependencies for the plugins

## Actual behavior

Trying to start the collectd service after installing the packages produces the following errors:

> Nov 30 17:17:05 collectd: ERROR: dlopen ("/usr/lib64/collectd/exec.so") failed: /usr/lib64/collectd/exec.so: undefined symbol: cmd_parse_flush. The most common cause for this problem is missing dependencies. Use ldd(1) to check the dependencies of the plugin / shared object.
> Nov 30 17:17:05 collectd: ERROR: dlopen ("/usr/lib64/collectd/tail.so") failed: /usr/lib64/collectd/tail.so: undefined symbol: cu_tail_read. The most common cause for this problem is missing dependencies. Use ldd(1) to check the dependencies of the plugin / shared object.

I wasn't having this issue in my previous version `collectd-5.6.1.261.g41288c6-1`

## Steps to reproduce

*   step 1 - Install the latest available collectd package from http://pkg.ci.collectd.org/rpm/master/epel-7-x86_64/
*   step 2 - Make sure you are loading the tail or exec plugins. For example:
/etc/collectd.d/10-exec.conf
```
<LoadPlugin exec>
  Globals false
</LoadPlugin>
```

*   step 3
Try to start the collectd service
`systemctl start collectd`

UPDATE: Re-tested with the 5.6 branch version from https://pkg.ci.collectd.org/rpm/collectd-5.6/epel-7-x86_64/, namely collectd-5.6.2-2. Everything good there!

octo:
I can confirm that **5.6.1.339.g797ed5e-1~wheezy** has the same issue, at least with the *tail plugin*.

octo:
The *collectd-5.6* branch doesn't actually have this issue:

```
octo@octo-glaptop:/tmp/collectd$ objdump -T sbin/collectd | grep tail
00000000004134e0 g    DF .text	00000000000000b8  Base        cu_tail_read
0000000000412180 g    DF .text	00000000000000a5  Base        tail_match_destroy
00000000004122a0 g    DF .text	000000000000012e  Base        tail_match_add_match_simple
0000000000411e10 g    DF .text	0000000000000013  Base        llist_tail
[...]

PR #2069, opened 2016-12-01
src/Makefile.am: Fix dependencies for the "tail" and "match" utilities.

octo:
Since being pulled out of the core daemon, the daemon no longer provides
all the required symbols for these shared objects. The "tail", "match"
and "tail_match" utils need to be compiled in explicitly. Also, we need
to link with the liblatency.la convenience library for the
"Distribution" handling in the "match" util.

Issue: #2067


rubenk:
LGTM

PR #2071, opened 2016-12-01
ipmi plugin: add more analog sensors support

vmytnykx:
- Add more analog sensors like 'PS1 Input Power', 'System Airflow' etc. The sensor type mapping => collectd db type done using sensor base unit type (+ percentage sensor type). 
- Avoid reading value of discrete sensors by ipmi_sensor_id_get_reading() function.
 
Example of analog sensors:
```
# ipmitool sensor get 'PS1 Input Power'
Locating sensor record...
Sensor ID              : PS1 Input Power (0x54)
 Entity ID             : 10.1
 Sensor Type (Threshold)  : Other
 Sensor Reading        : 0 (+/- 0) Watts
 Status                : ok
 Lower Non-Recoverable : na
 Lower Critical        : na
 Lower Non-Critical    : na
 Upper Non-Critical    : 868.000
 Upper Critical        : 920.000
 Upper Non-Recoverable : na
 Positive Hysteresis   : Unspecified
 Negative Hysteresis   : Unspecified
 Assertion Events      :
 Assertions Enabled    : unc+ ucr+
 Deassertions Enabled  : unc+ ucr+

# ipmitool sensor get 'System Airflow'
Locating sensor record...
Sensor ID              : System Airflow (0x11)
 Entity ID             : 23.1
 Sensor Type (Threshold)  : Other
 Sensor Reading        : 54 (+/- 0) CFM
 Status                : ok
 Lower Non-Recoverable : na
 Lower Critical        : na
 Lower Non-Critical    : na
 Upper Non-Critical    : na
 Upper Critical        : na
 Upper Non-Recoverable : na
 Positive Hysteresis   : Unspecified
 Negative Hysteresis   : Unspecified
 Assertion Events      :
 Assertions Enabled    :
```
Example of discrete sensors which are added to internal ipmi list on init but falied to read using ipmi_sensor_id_get_reading() func:

```
# ipmitool sensor get 'Fan 1 Present'
Locating sensor record...
Sensor ID              : Fan 1 Present (0x40)
 Entity ID             : 29.1
 Sensor Type (Discrete): Fan
 States Asserted       : Availability State
                         [Device Present]

# ipmitool sensor get 'P1 VRD Hot'
Locating sensor record...
Sensor ID              : P1 VRD Hot (0x90)
 Entity ID             : 3.1
 Sensor Type (Discrete): Temperature
```

Regards,
Volodymyr

### DIFF
--- a/src/ipmi.c
+++ b/src/ipmi.c
@@ -449,9 +449,12 @@ static int sensor_list_add(c_ipmi_instance_t *st, ipmi_sensor_t *sensor) {
       break;
 
     INFO("ipmi plugin: sensor_list_add: Ignore sensor `%s` of `%s`, "
-         "because I don't know how to handle its type (%#x, %s). "
-         "If you need this sensor, please file a bug report.",
-         sensor_name_ptr, st->name, sensor_type,
+         "because I don't know how to handle its units (%#x, %#x, %#x). "
+         "Sensor type: (%#x, %s). If you need this sensor, please file "
+         "a bug report at http://collectd.org/.",
+         sensor_name_ptr, st->name, ipmi_sensor_get_base_unit(sensor),
+         ipmi_sensor_get_modifier_unit(sensor),
+         ipmi_sensor_get_rate_unit(sensor), sensor_type,
          ipmi_sensor_get_sensor_type_string(sensor));
     return -1;
   }
--- a/src/ipmi.c
+++ b/src/ipmi.c
@@ -79,11 +79,18 @@ struct c_ipmi_sensor_list_s {
   ipmi_sensor_id_t sensor_id;
   char sensor_name[DATA_MAX_NAME_LEN];
   char sensor_type[DATA_MAX_NAME_LEN];
+  char type_instance[DATA_MAX_NAME_LEN];
   int sensor_not_present;
   c_ipmi_sensor_list_t *next;
   c_ipmi_instance_t *instance;
   unsigned int use;
 };
+
+struct c_ipmi_db_type_map_s {
+  enum ipmi_unit_type_e type;
+  const char *type_name;
+};
+typedef struct c_ipmi_db_type_map_s c_ipmi_db_type_map_t;
 
 /*
  * Module global variables
@@ -190,7 +197,7 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
         if (st->notify_notpresent) {
           notification_t n = c_ipmi_notification_init(st, NOTIF_WARNING);
 
-          sstrncpy(n.type_instance, list_item->sensor_name,
+          sstrncpy(n.type_instance, list_item->type_instance,
                    sizeof(n.type_instance));
           sstrncpy(n.type, list_item->sensor_type, sizeof(n.type));
           snprintf(n.message, sizeof(n.message), "sensor %s not present",
@@ -243,7 +250,7 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
     if (st->notify_notpresent) {
       notification_t n = c_ipmi_notification_init(st, NOTIF_OKAY);
 
-      sstrncpy(n.type_instance, list_item->sensor_name,
+      sstrncpy(n.type_instance, list_item->type_instance,
                sizeof(n.type_instance));
       sstrncpy(n.type, list_item->sensor_type, sizeof(n.type));
       snprintf(n.message, sizeof(n.message), "sensor %s present",
@@ -285,7 +292,8 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
     sstrncpy(vl.host, st->host, sizeof(vl.host));
   sstrncpy(vl.plugin, "ipmi", sizeof(vl.plugin));
   sstrncpy(vl.type, list_item->sensor_type, sizeof(vl.type));
-  sstrncpy(vl.type_instance, list_item->sensor_name, sizeof(vl.type_instance));
+  sstrncpy(vl.type_instance, list_item->type_instance,
+           sizeof(vl.type_instance));
 
   plugin_dispatch_values(&vl);
 } /* void sensor_read_handler */
@@ -338,6 +346,24 @@ static void sensor_get_name(ipmi_sensor_t *sensor, char *buffer, int buf_len) {
   sstrncpy(buffer, sensor_name, buf_len);
 }
 
+static const char *sensor_unit_to_type(ipmi_sensor_t *sensor) {
+  static const c_ipmi_db_type_map_t ipmi_db_type_map[] = {
+      {IPMI_UNIT_TYPE_WATTS, "power"}, {IPMI_UNIT_TYPE_CFM, "flow"}};
+
+  /* check the modifier and rate of the sensor value */
+  if ((ipmi_sensor_get_modifier_unit_use(sensor) != IPMI_MODIFIER_UNIT_NONE) ||
+      (ipmi_sensor_get_rate_unit(sensor) != IPMI_RATE_UNIT_NONE))
+    return NULL;
+
+  /* find the db type by using sensor base unit type */
+  enum ipmi_unit_type_e ipmi_type = ipmi_sensor_get_base_unit(sensor);
+  for (int i = 0; i < STATIC_ARRAY_SIZE(ipmi_db_type_map); i++)
+    if (ipmi_db_type_map[i].type == ipmi_type)
+      return ipmi_db_type_map[i].type_name;
+
+  return NULL;
+} /* const char* sensor_unit_to_type */
+
 static int sensor_list_add(c_ipmi_instance_t *st, ipmi_sensor_t *sensor) {
   ipmi_sensor_id_t sensor_id;
   c_ipmi_sensor_list_t *list_item;
@@ -377,10 +403,10 @@ static int sensor_list_add(c_ipmi_instance_t *st, ipmi_sensor_t *sensor) {
    *
    * ipmi_sensor_id_get_reading() supports only 'Threshold' sensors.
    * See lib/sensor.c:4842, stand_ipmi_sensor_get_reading() for details.
-  */
+   */
   if (!ipmi_sensor_get_is_readable(sensor)) {
     INFO("ipmi plugin: sensor_list_add: Ignore sensor `%s` of `%s`, "
-         "because it don't readable! Its type: (%#x, %s). ",
+         "because it isn't readable! Its type: (%#x, %s). ",
          sensor_name_ptr, st->name, sensor_type,
          ipmi_sensor_get_sensor_type_string(sensor));
     return -1;
@@ -413,7 +439,15 @@ static int sensor_list_add(c_ipmi_instance_t *st, ipmi_sensor_t *sensor) {
     type = "fanspeed";
     break;
 
+  case IPMI_SENSOR_TYPE_MEMORY:
+    type = "memory";
+    break;
+
   default: {
+    /* try to get collectd DB type based on sensor base unit type */
+    if ((type = sensor_unit_to_type(sensor)) != NULL)
+      break;
+
     INFO("ipmi plugin: sensor_list_add: Ignore sensor `%s` of `%s`, "
          "because I don't know how to handle its type (%#x, %s). "
          "If you need this sensor, please file a bug report.",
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
https://github.com/rpv-tomsk/collectd/commit/ba30d9ba566ab3723775d3daea6d21a31859d865

maryamtahhan:
@rpv-tomsk is there any feedback on this patchset? I don't quite understand the comment

rpv-tomsk:
> I don't quite understand the comment

I'm about we are doing the same work ) and also I want to point you to my work/patch, just you to be informed about its existence.

My feedback/opinion about your changes:

```
  /* check if this is a percentage */
  if (ipmi_sensor_get_percentage(sensor))
    return "percent";
```

In IPMI, percentage is a flag, not a unit itself. I don't like idea to mix different units into one "percent" type. It will be hard to group different metrics of `percent` type on hubs / at the side of displaying the graphs.

Maybe it will be better to add '_percent' suffix to `type` in that case?

What do you think about this?

vmytnykx:
Hi @rpv-tomsk,

Thank you for your feedbacks.

I agree that mixing different units into one "percent" type is not a best idea. But, the sensors with "percenage" flag set have "unspecified" unit type. So, in that case I suggest to use sensor type + _percentage suffix. E.g.: memory_percentage, temperature_percentage etc.

Regards,
Volodymyr

vmytnykx:
Hi @rpv-tomsk,

Actually, we cannot use my suggestion above, as ipmi plugin may select sensors (using precent flag) that may be unsupported by collectd type db. See example of such sensor below:

```
# ipmitool sensor get 'MTT CPU2'
Locating sensor record...
Sensor ID              : MTT CPU2 (0x35)
 Entity ID             : 3.2
 Sensor Type (Threshold)  : Memory
 Sensor Reading        : 0 (+/- 0) percent
 Status                : ok
 Lower Non-Recoverable : 0.000
 Lower Critical        : 0.000
 Lower Non-Critical    : 0.000
 Upper Non-Critical    : 0.000
 Upper Critical        : 0.000
 Upper Non-Recoverable : 0.000
 Positive Hysteresis   : Unspecified
 Negative Hysteresis   : Unspecified
 Assertion Events      :
 Assertions Enabled    :
```

Sensor has "memory" type and provide only "percent" flag. Unit type of this sensor is "unspecified". For this reason, we cannot build and use `memory_percent` type as it isn't supported by collectd type db. So, you are correct, we should only use `type` + `_percent` suffix in that case.

Also, I think it isn't correct to select sensors by just presense of 'percent' falg. Thus, I'm going to make appropriate changes the PR to fix all described issues.

Thanks and Regards,
Volodymyr

vmytnykx:
Hi @rpv-tomsk,

I've removed logic that selects sensors by presence of of 'percent' falg. Right now, when the sensor provides percentage flag, the type of the sensor sets to `type` + `_percent` as we discussed above. Also, the support of `memory` sensor type has been added in usual way (due to the last changes).
Thanks for your feedback.

Regards,
Volodymyr

src/ipmi.c
```diff
@@ -287,6 +325,13 @@ static int sensor_list_add(ipmi_sensor_t *sensor) {
   }
   } /* switch (sensor_type) */
 
+  /* if sensor provides the percentage value, add "_percent" suffix to the
+   * sensor collectd type */
+  if (ipmi_sensor_get_percentage(sensor)) {
+    ssnprintf(percent_type, sizeof(percent_type), "%s_percent", type);
```
octo: Please use the existing `percent` type instead of introducing more specialized `foo_percent` types. The original type should go into the type instance. For example, "power-ac0" (type: "power", type instance: "ac0") would become "percent-power-ac0" (type: "percent", type instance: "power-ac0").

It's best to think of collectd's types as a "unit label", if that makes sense.

octo: Please use `snprintf()`. The `ssnprintf()` wrapper has been removed from *master*.

octo:
P.S.: Don't worry about the merge conflicts, we'll take care of them after the review concluded.

src/ipmi.c
rpv-tomsk: I dislike such a mixing of different *types* into one 'percent' type.
IMHO, personally I would not use such implementation in my systems.

rpv-tomsk:
>P.S.: Don't worry about the merge conflicts, we'll take care of them after the review concluded.

This PR and mine #2024 do the same thing by a different ways.
This implementation adds a 'is discrete' flag into 'sensors' list, mine prohibits adding discrete sensors into that list.
This PR request logic: that may be used at future implementations.
Mine logic: it makes no sensce of a such adding - these sensors are not used later _in current implementation_.

So, one of implementation completely contradicts the other.

src/ipmi.c
```diff
@@ -200,6 +211,19 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
   plugin_dispatch_values(&vl);
 } /* void sensor_read_handler */
 
+static const char *sensor_get_db_type(ipmi_sensor_t *sensor) {
```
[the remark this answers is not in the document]
rpv-tomsk: As far I remember, use of only one `ipmi_sensor_get_base_unit()` is not enough for a such task.
The sensor, besides the base unit, also has a 'modifier' and 'rate' values (ipmi_sensor_get_modifier_unit() and ipmi_sensor_get_rate_unit() respectively). May be these are rare used, but I think we need to check their values anyway - otherwise incorrect result can be obtained.

src/ipmi.c
```diff
@@ -275,10 +306,17 @@ static int sensor_list_add(ipmi_sensor_t *sensor) {
     type = "fanspeed";
     break;
 
-  default: {
-    const char *sensor_type_str;
+  case IPMI_SENSOR_TYPE_MEMORY:
```
rpv-tomsk: Can you please show 'ipmitool' output for such a sensor?
For example, by running `sensor get SENSOR_NAME` ?

Thanks.

vmytnykx: Hi @rpv-tomsk, see me comment above for this example https://github.com/collectd/collectd/pull/2071#issuecomment-296641468

maryamtahhan:
@rpv-tomsk 

> So, one of implementation completely contradicts the other.

I think the end goal is the same here... discrete sensors don't get read... so from my perspective it's fine to go with your suggestion

rpv-tomsk:
@maryamtahhan

Let's discuss what to do with 'percent' flag?

rpv-tomsk:
We have a sensor types and we have units of measurements.

We have a 'temperature' type with 'degrees C' unit.
We have a 'fanspeed' type with 'RPM' unit.

Some IPMI sensor types are mapped to Collectd types directly.

As for me, 'percent' is a unit, not a type.
Happily, I have no much sensors with percents: The only one is Intel platform which reports consumed power in sensor of `IPMI_SENSOR_TYPE_CURRENT` type and value is a percentage.
Of course, nobody measures current in percents and that sensor value means power supply load.
If that will be reported with a 'percent' type that will be not so good as it might be.

Let's solve this task starting from a practical requirements?

src/ipmi.c
vmytnykx: In nutshell, this sensor represents "Memory Thermal Throttling" and is related to memory thermal management system. Based on the DIMM thermal conditions it may restrict read and write traffic/bandwidth to main memory as a means of controlling power consumption. This metric is measured as a percentage and 0% means no memory throttling occurs. When thermal conditions are going high, the memory management system enables throttling and restricts the read or write traffic (e.g. 50%).

maryamtahhan:
@rpv-tomsk There are thermal and other sensors on Intel platforms that report values in %...

My understanding of types in collectd is that type can be thought of the unit used to measure a value (but I could be wrong) ...
If we look at some other plugins for precedent: hupepages, virt and the cpu plugins report % as a **type** for a number of different measurements... 
I would be for keeping the generic type as % as @octo suggests, but then either: 
1. using the sensor type as an extension (prefix/suffix to the type instance)...
2. use metadata for sensor type/sensor name leaving one of these as the type instance...

I would prefer option 1.

As an aside: MTT sensor is the Memory Throttling sensor on Intel platforms...

octo:
> 1. using the sensor type as an extension (prefix/suffix to the type instance)...

+1, that's exactly what I wanted to suggest.

@maryamtahhan, @rpv-tomsk How exactly does this PR relate to #2024? Should we close this in favor of #2024 or do they only overlap in the bit that deals with discrete sensors?

rpv-tomsk:
This PR intended to add new mappings of IPMI units to Collectd types.

 https://github.com/rpv-tomsk/collectd/commit/ba30d9ba566ab3723775d3daea6d21a31859d865

rpv-tomsk:
>I would prefer option 1.

What is about a such solution: map sensors with a 'percent' flag to 'percent' type, but allow to configure custom mapping by a sensor name (may be implemented later)?

vmytnykx:
Just want to summarize the remaining work on the changes:
* Don't read the discrete sensors. I'm OK to remove existing logic and use the logic implemented by @rpv-tomsk in https://github.com/rpv-tomsk/collectd/blob/0d9be8e2dd5fb3c801d805ef1d54b60963a2fccc/src/ipmi.c#L369-L395
* Map sensors with a 'percent' flag to 'percent' type and add sensor type as a prefix to `type_instance` in this case. The `type_instance` of non % sensors will remain the same (just sensor name).
* Fix https://github.com/collectd/collectd/pull/2071/files/18767f257d29e1d3ec0bdd09ecbad13ac602df05#diff-88ee790203c729461340e7b84aa31bb5 comments.

Is my understanding correct?

rpv-tomsk:
Volodymyr, I think that your summary is correct.

Does anybody will comment mine proposal of configurable mapping? Any opinions?

vmytnykx:
@rpv-tomsk, could you please clarify your proposal. Do you mean to add new bool config option into collectd conf which will enable adding prefix or not? Or you mean new option which will do the whole percent mapping or not depends on the flag?

octo:
@rpv-tomsk Yes, I'll review #2024

rpv-tomsk:
I propose to discuss adding of something similar to:
```
<Plugin ipmi>
  <Sensor "CPU2 Vcore system_board (7.2)">
      Type "voltage"
      TypeInstance "cpu2"
   </Sensor>
    <Instance>
       ....
    </Instance>
</Plugin>
```

That will allow to override all hardcode mappings, by sensor type or unit type.

maryamtahhan:
@rpv-tomsk as regards your last proposal, I think you can do this with filtering chains in collectd... so even if you are not happy with the hardcoding you can change it before the metric gets published...

rpv-tomsk:
Filtering chains...  Some while ago there was a same suggestion about use of filtering chains for setting plugin name / plugin instance in 'generic plugins'..  I prefer to configure all things in one place and in a most simplest way (by a form). Chains are closer to a hack for me, not to a solution.

octo:
How do we proceed with this? #2024 has been merged and this PR needs a rebase, ideally pulling in the existing handling of discrete sensors. Then do a final review?

vmytnykx:
@octo, working on review comments changes now. Then I can re-base these changes based on the latest master. Do you prefer to do final review before re-base or after?

octo:
Given the complexity of the rebase, I think a review after rebase makes more sense in this case.

vmytnykx:
Hi @octo, @rpv-tomsk,

I've addressed the review comments and changed the implementation as discussed in https://github.com/collectd/collectd/pull/2071#issuecomment-335177508 

The changes are re-based and squashed into one commit.

Thank you all for review :)

Regards,
Volodymyr

rpv-tomsk:
Hi, Volodymyr!

Thanks for update and your work on this.
What will you say about to make this message (https://github.com/collectd/collectd/pull/2071/commits/1f88406aadbfbd3da8f2d7f08220d2734aa286b4#diff-88ee790203c729461340e7b84aa31bb5R451) more informative? The Collectd now tries to handle sensors `units of measurements', but that message  does not show us these values. That will not give us enough data for new types mapping.

Mine proposal is to add units values to the message output, like was done here: https://github.com/rpv-tomsk/collectd/commit/ba30d9ba566ab3723775d3daea6d21a31859d865#diff-88ee790203c729461340e7b84aa31bb5R439

vmytnykx:
Hi @rpv-tomsk, the idea sounds good to me. I will fix it and update the PR. 

Regarding the build failure, seems a Jenkins issue:

```
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... configure: error: newly created file is older than distributed files!
Check your system clock
+ make -sk check
make: *** No rule to make target 'check'.
Archiving artifacts
Skipping publisher since build result is FAILURE
```
Thanks and Regards,
Volodymyr

vmytnykx:
Hi @rpv-tomsk,

rpv-tomsk:
Hi Volodymyr!

Thanks for your work on this!

maryamtahhan:
Whoop whoop! Great work all, thanks for the perseverance

PR #2072, opened 2016-12-02
Treewide: remove vim modelines from C code files

rubenk:
They often conflict with the formatting we do with clang-format.


src/aggregation.c
```diff
@@ -753,4 +753,3 @@ void module_register(void) {
   plugin_register_write("aggregation", agg_write, /* user_data = */ NULL);
 }
```
octo: Could you also delete the blank line at the end of files?

rubenk: sure, will do :)

PR #2079, opened 2016-12-05
write_prometheus plugin: Add "total" suffix to the counters only

hasso:
collectd_exporter adds "total" suffix to the counters only: 

https://github.com/prometheus/collectd_exporter/blob/master/main.go#L70-L73.

As write_prometheus plugins is meant to be a direct replacement for collectd_exporter, it makes sense not to break a lot of metrics/dashboards/etc for a current collectd_exporter users.

### DIFF
--- a/src/write_prometheus.c
+++ b/src/write_prometheus.c
@@ -665,8 +665,7 @@ static char *metric_family_name(data_set_t const *ds, value_list_t const *vl,
 
   /* Prometheus best practices:
    * cumulative metrics should have a "total" suffix. */
-  if ((ds->ds[ds_index].type == DS_TYPE_COUNTER) ||
-      (ds->ds[ds_index].type == DS_TYPE_DERIVE)) {
+  if (ds->ds[ds_index].type == DS_TYPE_COUNTER) {
     fields[fields_num] = "total";
     fields_num++;
   }


octo:
Hi @hasso, thank you for reporting this issue!

I agree, *collectd_exporter* and the *write_prometheus* plugin should behave the same way.

@brian-brazil, @beorn7: what do you think? Should the `_total` suffix for `DERIVE` metrics be added to *collectd_exporter* (I'm happy to send a PR) or should we remove the suffix in the *write_prometheus* plugin?

brian-brazil:
The best way to handle such metrics is to get the raw value (not the derived one) and expose it as a counter, in which case it should have the _total suffix.

beorn7:
...and should you still expose derivative values for whatever reason, don't give it a `_total` suffix (because that suggests it's a counter you can take a `rate` of in a meaningful way).

octo:
Thanks for your replies, @brian-brazil  and @beorn7. I think there's a misunderstanding though: both, `COUNTER` and `DERIVE` are *cumulative* metrics, i.e. `DERIVE` is a counter *to be derived*, not a derivative. I think that, from Prometheus' POV, they should be treated equally.

The difference between the two is how they expect the case to be handled, when a value is *smaller* than the previous value:

*   `DERIVE` expects that the *rate* (i.e. first order derivative) is calculated using `rate = (v_n – v_{n-1}) / (t_n – t_{n-1})`. If `v_n < v_{n-1}`, the rate will be negative. Most "types" set a minimum value to zero, so that these datapoints are simply dropped. I think this is essentially equivalent to Prometheus' `clamp_min(deriv(v), 0)` or `irate()`.
*   `COUNTER` expects the rate to be calculated in the same way, except when `v_n < v_{n-1}`. In that case, a counter overflow is assumed and the difference between the two datapoints is calculated as `2^64 + v_n - v_{n-1}`.

Our wiki has more details: https://collectd.org/wiki/index.php/Data_source

Since, with 64bit counters, counter *resets* are much more likely than counter *overflows*, `COUNTER` is mostly a legacy and all cumulative metrics collectd's standard plugins report are using `DERIVE`.

brian-brazil:
Those are both counters then, and should have the _total suffix if possible.

beorn7:
But note that for a Prometheus counter, any drop will be seen as a counter reset. Nothing would stop you from running a query like `clamp_min(deriv(v),0)`, but if I see anything ending on `_total`, I will use `rate` on it, not `deriv`.

octo:
@beorn7 that's the intended behavior for >99.9% of DERIVE metrics. I think it's best to add a special case if and when users ask for negative rates, likely in the form of a config option.

octo:
Thanks again for pointing this out, @hasso! With prometheus/collectd_exporter#41 merged, *collectd_exporter* and *write_prometheus* behave identically.

PR #2088, opened 2016-12-08
Fix processes and tcpconns plugins on OpenBSD

landryb:
All the hard work from jca@, i'm only proxying the patches already commited in our ports-tree in http://marc.info/?l=openbsd-ports-cvs&m=148087062627664&w=2

Dunno if `kvm_getfiles()` could be properly detected within configure, but as jca@ figured out, NetBSD can't use that codepath as "their `struct kinfo_file` doesnt have some needed members" (or something..)


rubenk:
Thanks @landryb!

PR #2091, opened 2016-12-08
ipmi: add support for system event log (SEL).

rkorynkx:
Listen for ipmi sensor threshold and discrete events. Send notification in case event is received. Fix typos.

### DIFF
--- a/src/ipmi.c
+++ b/src/ipmi.c
@@ -59,9 +59,12 @@ static int c_ipmi_init_in_progress = 0;
 static int c_ipmi_active = 0;
 static pthread_t thread_id = (pthread_t)0;
 
-static const char *config_keys[] = {"Sensor", "IgnoreSelected",
-                                    "NotifySensorAdd", "NotifySensorRemove",
-                                    "NotifySensorNotPresent", "SELEnabled",
+static const char *config_keys[] = {"Sensor",
+                                    "IgnoreSelected",
+                                    "NotifySensorAdd",
+                                    "NotifySensorRemove",
+                                    "NotifySensorNotPresent",
+                                    "SELEnabled",
                                     "SELClearEvent"};
 static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
 
@@ -119,7 +122,8 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
              list_item->sensor_name);
 
         if (c_ipmi_notify_notpresent) {
-          notification_t n = {NOTIF_WARNING, cdtime(), "", "", "ipmi", "", "", "", NULL};
+          notification_t n = {
+              NOTIF_WARNING, cdtime(), "", "", "ipmi", "", "", "", NULL};
 
           sstrncpy(n.host, hostname_g, sizeof(n.host));
           sstrncpy(n.type_instance, list_item->sensor_name,
@@ -167,8 +171,8 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
          list_item->sensor_name);
 
     if (c_ipmi_notify_notpresent) {
-      notification_t n = {NOTIF_OKAY, cdtime(), "", "", "ipmi", "", "", "",
-                          NULL};
+      notification_t n = {NOTIF_OKAY, cdtime(), "", "",  "ipmi",
+                          "",         "",       "", NULL};
 
       sstrncpy(n.host, hostname_g, sizeof(n.host));
       sstrncpy(n.type_instance, list_item->sensor_name,
@@ -382,8 +386,8 @@ static int sensor_list_remove(ipmi_sensor_t *sensor) {
   pthread_mutex_unlock(&sensor_list_lock);
 
   if (c_ipmi_notify_remove && c_ipmi_active) {
-    notification_t n = {NOTIF_WARNING, cdtime(), "", "",  "ipmi",
-                        "",            "",       "", NULL};
+    notification_t n = {NOTIF_WARNING, cdtime(), "", "", "ipmi", "", "", "",
+                        NULL};
 
     sstrncpy(n.host, hostname_g, sizeof(n.host));
     sstrncpy(n.type_instance, list_item->sensor_name, sizeof(n.type_instance));
@@ -526,7 +530,7 @@ static int sensor_threshold_event_handler(
     char buf[DATA_MAX_NAME_LEN] = {0};
     snprintf(buf, sizeof(buf), "0x%2.2x", raw_value);
     plugin_notification_meta_add_string(&n, "raw", buf);
-    } break;
+  } break;
   default:
     break;
   } /* switch (value_present) */
@@ -552,7 +556,7 @@ static int sensor_discrete_event_handler(ipmi_sensor_t *sensor,
   /* From the IPMI specification Chapter 2: Events.
    * If a callback handles the event, then all future callbacks called due to
    * the event will receive a NULL for the event. So be ready to handle a NULL
-   * event in all your event handlers. A NULL may also be passed to an event 
+   * event in all your event handlers. A NULL may also be passed to an event
    * handler if the callback was not due to an event. */
   if (event == NULL)
     return (IPMI_EVENT_NOT_HANDLED);
@@ -623,8 +627,7 @@ static void entity_sensor_update_handler(
       if (status) {
         char buf[DATA_MAX_NAME_LEN] = {0};
         sensor_get_name(sensor, buf, sizeof(buf));
-        ERROR("Unable to add sensor %s event handler, status: %d", buf,
-              status);
+        ERROR("Unable to add sensor %s event handler, status: %d", buf, status);
       }
     }
   } else if (op == IPMI_DELETED) {
@@ -668,8 +671,7 @@ static void domain_entity_update_handler(
 static void smi_event_handler(ipmi_con_t __attribute__((unused)) *ipmi,
                               const ipmi_addr_t __attribute__((unused)) *addr,
                               unsigned int __attribute__((unused)) addr_len,
-                              ipmi_event_t *event,
-                              void *cb_data) {
+                              ipmi_event_t *event, void *cb_data) {
   unsigned int type = ipmi_event_get_type(event);
   ipmi_domain_t *domain = cb_data;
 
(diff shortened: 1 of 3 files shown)


rpv-tomsk:
Hi!

Your patch has `ipmi_event_delete()` call. 
As I understand, that call deletes event from system event log. 
Didn't you think that it should be optional?

rkorynkx:
Hi @rpv-tomsk, I agree, I will make it optional.

maryamtahhan:
@rpv-tomsk any additional feedback here?

rpv-tomsk:
Hi Maryam!

I have no other feedbacks here.  Thanks for cooperation!

src/ipmi.c
```diff
@@ -61,14 +61,17 @@ static pthread_t thread_id = (pthread_t)0;
...
+                                    "NotifySensorNotPresent", "SELEnabled",
+                                    "SELClearEvent"};
 static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
 
 static ignorelist_t *ignorelist = NULL;
 
-static int c_ipmi_nofiy_add = 0;
-static int c_ipmi_nofiy_remove = 0;
-static int c_ipmi_nofiy_notpresent = 0;
+static int c_ipmi_notify_add = 0;
```
octo: Nit: these should be of type `_Bool`.

src/ipmi.c
```diff
@@ -200,56 +202,66 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
...
+        ssnprintf(sensor_name, sizeof(sensor_name), "%s %s", sensor_name_ptr,
+                  sensor_id_ptr);
+      }
+      /* else: don't touch sensor_name. */
     }
-    /* else: don't touch sensor_name. */
   }
-  sensor_name_ptr = sensor_name;
+
+  assert(buf != NULL);
```
octo: Please don't do argument checking with `assert()`. Instead, add something like this to the beginning of the function.

```c
if ((buf == NULL) || (buf_len == 0)) {
  return;
}
```

src/ipmi.c
```diff
@@ -200,56 +202,66 @@ static void sensor_read_handler(ipmi_sensor_t *sensor, int err,
   plugin_dispatch_values(&vl);
 } /* void sensor_read_handler */
 
-static int sensor_list_add(ipmi_sensor_t *sensor) {
-  ipmi_sensor_id_t sensor_id;
-  c_ipmi_sensor_list_t *list_item;
-  c_ipmi_sensor_list_t *list_prev;
-
+static void sensor_get_name(ipmi_sensor_t *sensor, char *buf, int buf_len) {
```
octo: I find having both, `buf` and `buffer` a bit confusing. I suggest to rename `buffer` to `temp` and `buf` to `buffer`.

src/ipmi.c
```diff
@@ -420,6 +432,168 @@ static int sensor_list_remove_all(void) {
...
+  /* From the IPMI specification Chapter 2: Events.
+   * If a callback handles the event, then all future callbacks called due to
+   * the event will receive a NULL for the event. So be ready to handle a NULL
+   * event in all your event handlers. A NULL may also be passed to an event 
+   * handler if the callback was not due to an event. */
+  if (event == NULL)
+    return (IPMI_EVENT_NOT_HANDLED);
+
+  sensor_get_name(sensor, buf, sizeof(buf));
+  sstrncpy(n.type_instance, buf, sizeof(n.type_instance));
```
octo: This seems like an unnecessary copy since you can also do:

```c
sensor_get_name(sensor, n.type_instance, sizeof(n.type_instance));
```

src/ipmi.c
```diff
@@ -420,6 +432,168 @@ static int sensor_list_remove_all(void) {
...
+  const char *event_state =
+      ipmi_get_reading_name(event_type, sensor_type, offset);
+  char buf[DATA_MAX_NAME_LEN] = {0};
+
+  /* From the IPMI specification Chapter 2: Events.
+   * If a callback handles the event, then all future callbacks called due to
+   * the event will receive a NULL for the event. So be ready to handle a NULL
+   * event in all your event handlers. A NULL may also be passed to an event 
+   * handler if the callback was not due to an event. */
+  if (event == NULL)
```
octo: Shouldn't this be the first thing we do in the function? We don't use the return value of `ipmi_get_reading_name()` if `e == NULL`.

src/ipmi.c
```diff
@@ -420,6 +432,168 @@ static int sensor_list_remove_all(void) {
...
+    ssnprintf(n.message, sizeof(n.message),
+              "sensor %s received event: %s, value not provided", buf,
+              event_state);
+
+  DEBUG("Threshold event received for sensor %s", buf);
+
+  sstrncpy(n.host, hostname_g, sizeof(n.host));
+  sstrncpy(n.type, ipmi_sensor_get_sensor_type_string(sensor), sizeof(n.type));
+  n.severity = sensor_convert_threshold_severity(threshold);
+  n.time = ipmi_event_get_timestamp(event);
```
octo: You need to use `NS_TO_CDTIME_T()` (from `"utils_time.h"`) to convert the IPMI timestamp to collectd's internal time representation.

src/ipmi.c
```diff
@@ -420,6 +432,168 @@ static int sensor_list_remove_all(void) {
...
+
+  plugin_notification_meta_add_string(&n, "severity",
+                                      ipmi_get_threshold_string(threshold));
+  plugin_notification_meta_add_string(&n, "direction",
+                                      ipmi_get_value_dir_string(high_low));
+
+  switch (value_present) {
+  case IPMI_BOTH_VALUES_PRESENT:
+    plugin_notification_meta_add_double(&n, "val", value);
```
octo: Please add a `/* fall-through */` comment to show that this is intentional.

src/ipmi.c
```diff
@@ -420,6 +432,168 @@ static int sensor_list_remove_all(void) {
...
+  const char *event_state =
+      ipmi_get_reading_name(event_type, sensor_type, offset);
+  char buf[DATA_MAX_NAME_LEN] = {0};
+
+  /* From the IPMI specification Chapter 2: Events.
+   * If a callback handles the event, then all future callbacks called due to
+   * the event will receive a NULL for the event. So be ready to handle a NULL
+   * event in all your event handlers. A NULL may also be passed to an event
+   * handler if the callback was not due to an event. */
+  if (event == NULL)
```
octo: Shouldn't this be the first thing we do in the function? We don't use the return value of `ipmi_get_reading_name()` if `e == NULL`.

Dito for `sensor_discrete_event_handler()` below.

src/ipmi.c
```diff
@@ -420,6 +432,168 @@ static int sensor_list_remove_all(void) {
...
+    return (IPMI_EVENT_NOT_HANDLED);
+
+  sensor_get_name(sensor, buf, sizeof(buf));
+  sstrncpy(n.type_instance, buf, sizeof(n.type_instance));
```
octo: This seems like an unnecessary copy since you can also do:

```c
sensor_get_name(sensor, n.type_instance, sizeof(n.type_instance));
```

src/ipmi.c
```diff
@@ -420,6 +432,168 @@ static int sensor_list_remove_all(void) {
...
+  ssnprintf(n.message, sizeof(n.message), "sensor %s received event: %s", buf,
+            event_state);
+
+  DEBUG("Discrete event received for sensor %s", buf);
+
+  sstrncpy(n.host, hostname_g, sizeof(n.host));
+  sstrncpy(n.type, ipmi_sensor_get_sensor_type_string(sensor), sizeof(n.type));
+  n.time = ipmi_event_get_timestamp(event);
```
octo: You need to use `NS_TO_CDTIME_T()` (from `"utils_time.h"`) to convert the IPMI timestamp to collectd's internal time representation.

src/ipmi.c
```diff
@@ -420,6 +432,168 @@ static int sensor_list_remove_all(void) {
...
+    plugin_notification_meta_add_string(&n, "raw", buf);
+    break;
+  default:
+    break;
+  } /* switch (value_present) */
+
+  add_event_common_data(&n, sensor, dir, event);
+
+  plugin_dispatch_notification(&n);
+
```
octo: The meta data leaks here, please call `plugin_notification_meta_free(&n)` to free the memory.

src/ipmi.c
```diff
@@ -420,6 +432,168 @@ static int sensor_list_remove_all(void) {
...
+  if (severity != -1)
+    plugin_notification_meta_add_signed_int(&n, "severity", severity);
+
+  if (prev_severity != -1)
+    plugin_notification_meta_add_signed_int(&n, "prevseverity", prev_severity);
+
+  add_event_common_data(&n, sensor, dir, event);
+
+  plugin_dispatch_notification(&n);
+
```
octo: The meta data leaks here, please call `plugin_notification_meta_free(&n)` to free the memory.

src/ipmi.c
```diff
@@ -441,7 +647,7 @@ static void entity_sensor_update_handler(
  */
 static void domain_entity_update_handler(
     enum ipmi_update_e op, ipmi_domain_t __attribute__((unused)) * domain,
-    ipmi_entity_t *entity, void __attribute__((unused)) * user_data) {
+    ipmi_entity_t * entity, void __attribute__((unused)) * user_data) {
```
octo: This looks like something *clang-format* should take care of. Please make sure to run it when preparing commits.

src/ipmi.c
```diff
@@ -551,13 +783,19 @@ static int c_ipmi_config(const char *key, const char *value) {
...
   } else if (strcasecmp("NotifySensorRemove", key) == 0) {
     if (IS_TRUE(value))
-      c_ipmi_nofiy_remove = 1;
+      c_ipmi_notify_remove = 1;
   } else if (strcasecmp("NotifySensorNotPresent", key) == 0) {
     if (IS_TRUE(value))
-      c_ipmi_nofiy_notpresent = 1;
+      c_ipmi_notify_notpresent = 1;
+  } else if (strcasecmp("SELEnabled", key) == 0) {
+    if (IS_TRUE(value))
```
octo: Please use `c_ipmi_sel_enabled = IS_TRUE(value);` here and elsewhere. If a user ends up with a config like this:

```
SELEnabled true
SELEnabled false
```

the second statement should be able to overwrite the previous statement. I'd be very grateful if you could also fix the existing options, thank you!

vmytnykx:
Hi @octo! Thanks for reviewing the changes and your feedback. I've addressed your comments in last two commits. Also, performed clang-format just to make sure all formatting issues are fixed. I haven't re-based the changes as I don't want to break the whole history. So, if you are ok with the changes please let me know, so I can re-base them.

Regards,
Volodymyr

PR #2093, opened 2016-12-09
virt plugin: fix error path in lv_domain_get_tag

mojaves:
Document the return value and fix the error path of
lv_domain_get_tag


rubenk:
Thanks!

PR #2095, opened 2016-12-09
dpdkstat plugin rework

krzysztofx:
This patch contains rework of existing dpdkstat plugin. Changes are focused on
extracting common dpdk related features into utils file that could be reused by
other DPDK plugins. Also some related structural changes are introduced.

List of changes:
- introduction of new utils_dpdk.c that handles child process spawning,
  and terminating, messaging between parent and child, DPDK configuration
  and initialization and DPDK commands execution
- DPDK EAL specific options put into separate subsection of
  dpdkstat configuration
- documentation update

Those changes are preparatory steps for upstreaming dpdkevent plugin
that is going to collect DPDK ports and logical cores status info.

### DIFF
--- a/docs/BUILD.dpdkstat.md
+++ b/docs/BUILD.dpdkstat.md
@@ -201,7 +201,11 @@ See also: http://dpdk.org/doc/guides/prog_guide/multi_proc_support.html
  *  Generate the build script as specified below. (i.e. run `build.sh`).
  *  Configure collectd with the DPDK shared library:
 
-        ./configure --with-libdpdk=/usr
+        ./configure LIBDPDK_CPPFLAGS="-I/usr/include/dpdk" LIBDPDK_LDFLAGS="-L/usr/lib"
+
+**Note:** Modify these flags according to specific environment setup.
+LIBDPDK_CPPFLAGS should contain path to dpdk headers and  LIBDPDK_LDFLAGS should
+point out to dpdk libraries location.
 
 ### Build with the static DPDK library
 
@@ -209,7 +213,7 @@ To configure collectd with the DPDK static library:
 
  *  Run *configure* with the following CFLAGS:
 
-        ./configure --with-libdpdk=/usr CFLAGS=" -lpthread -Wl,--whole-archive -Wl,-ldpdk -Wl,-lm -Wl,-lrt -Wl,-lpcap -Wl,-ldl -Wl,--no-whole-archive"
+        ./configure LIBDPDK_CPPFLAGS="-I/usr/include/dpdk" LIBDPDK_LDFLAGS="-L/usr/lib" CFLAGS=" -lpthread -Wl,--whole-archive -Wl,-ldpdk -Wl,-lm -Wl,-lrt -Wl,-lpcap -Wl,-ldl -Wl,--no-whole-archive"
 
  *  Make sure that dpdk and dpdkstat are enabled in the *configure* output.
 
@@ -218,7 +222,7 @@ To configure collectd with the DPDK static library:
         Libraries:
         ...
         libdpdk  . . . . . . . . yes
-        
+
         Modules:
         ...
         dpdkstat . . . . . . .yes
--- a/docs/BUILD.dpdkstat.md
+++ b/docs/BUILD.dpdkstat.md
@@ -148,34 +148,6 @@ instruction set manually:
 
      *  Run `ldconfig` to update the shared library cache.
 
-### Static library
-
-To build static DPDK library for use with collectd:
-
- *  To configure DPDK to build the combined static library `libdpdk.a` ensure
-    that `CONFIG_RTE_BUILD_SHARED_LIB` is set to “n” in `config/common_base` in
-    your DPDK as follows:
-
-        #
-        # Compile to share library
-        #
-        CONFIG_RTE_BUILD_SHARED_LIB=n
-
- *  Prepare the configuration for the appropriate target as specified at:
-    http://dpdk.org/doc/guides/linux_gsg/build_dpdk.html.
-
-    For example:
-
-        make config T=x86_64-native-linuxapp-gcc
-
- *  Build the target using `-fPIC`:
-
-        make EXTRA_CFLAGS=-fPIC -j
-
- *  Install DPDK to `/usr`:
-
-        sudo make install prefix=/usr
-
 ## Build collectd with DPDK
 
 **Note:** DPDK 16.04 is the minimum version and currently supported version of
@@ -199,23 +171,18 @@ implications.
 See also: http://dpdk.org/doc/guides/prog_guide/multi_proc_support.html
 
  *  Generate the build script as specified below. (i.e. run `build.sh`).
- *  Configure collectd with the DPDK shared library:
+ *  Configure collectd with the DPDK shared library. If DPDK is installed in
+    custom installation path you can specify headers include path using
+    LIBDPDK_CPPFLAGS variable and libraries path with LIBDPDK_LDFLAGS.
+    Example:
 
-        ./configure LIBDPDK_CPPFLAGS="-I/usr/include/dpdk" LIBDPDK_LDFLAGS="-L/usr/lib"
+        ./configure
 
-**Note:** Modify these flags according to specific environment setup.
-LIBDPDK_CPPFLAGS should contain path to dpdk headers and  LIBDPDK_LDFLAGS should
-point out to dpdk libraries location.
+        or for custom DPKD installation:
 
-### Build with the static DPDK library
+        ./configure LIBDPDK_CPPFLAGS="-I/home/joe/include/dpdk" LIBDPDK_LDFLAGS="-L/home/joe/usr/lib"
 
-To configure collectd with the DPDK static library:
-
- *  Run *configure* with the following CFLAGS:
-
-        ./configure LIBDPDK_CPPFLAGS="-I/usr/include/dpdk" LIBDPDK_LDFLAGS="-L/usr/lib" CFLAGS=" -lpthread -Wl,--whole-archive -Wl,-ldpdk -Wl,-lm -Wl,-lrt -Wl,-lpcap -Wl,-ldl -Wl,--no-whole-archive"
-
- *  Make sure that dpdk and dpdkstat are enabled in the *configure* output.
+ *  Make sure that libdpdk and dpdkstat are enabled in the *configure* output.
 
     Expected output:
 
@@ -229,7 +196,7 @@ To configure collectd with the DPDK static library:
 
  *  Build collectd:
 
-        make -j && make -j install.
+    make -j && make -j install
 
     **Note:** As mentioned above, if you are building on Ubuntu 14.04 with
     GCC <= 4.8.X, you need to use:
--- a/src/dpdkstat.c
+++ b/src/dpdkstat.c
@@ -1,4 +1,4 @@
-/*-
+/*
  * collectd - src/dpdkstat.c
  * MIT License
  *
@@ -26,47 +26,30 @@
  *   Maryam Tahhan <maryam.tahhan@intel.com>
  *   Harry van Haaren <harry.van.haaren@intel.com>
  *   Taras Chornyi <tarasx.chornyi@intel.com>
+ *   Serhiy Pshyk <serhiyx.pshyk@intel.com>
+ *   Krzysztof Matczak <krzysztofx.matczak@intel.com>
  */
 
 #include "collectd.h"
 
-#include "common.h" /* auxiliary functions */
-#include "plugin.h" /* plugin_register_*, plugin_dispatch_values */
-#include "utils_time.h"
+#include "common.h"
+#include "utils_dpdk.h"
 
-#include <getopt.h>
-#include <poll.h>
-#include <semaphore.h>
-#include <sys/mman.h>
-#include <sys/queue.h>
-
-#include <rte_atomic.h>
-#include <rte_branch_prediction.h>
-#include <rte_common.h>
 #include <rte_config.h>
-#include <rte_debug.h>
-#include <rte_debug.h>
-#include <rte_eal.h>
 #include <rte_ethdev.h>
-#include <rte_launch.h>
-#include <rte_lcore.h>
-#include <rte_log.h>
-#include <rte_malloc.h>
-#include <rte_memory.h>
-#include <rte_memzone.h>
-#include <rte_per_lcore.h>
-#include <rte_string_fns.h>
-#include <rte_tailq.h>
-#include <rte_version.h>
 
-#define DPDK_DEFAULT_RTE_CONFIG "/var/run/.rte_config"
-#define DPDK_MAX_ARGC 8
-#define DPDKSTAT_MAX_BUFFER_SIZE (4096 * 4)
-#define DPDK_SHM_NAME "dpdk_collectd_stats_shm"
-#define ERR_BUF_SIZE 1024
-#define REINIT_SHM 1
-#define RESET 1
-#define NO_RESET 0
+#define DPDK_STATS_PLUGIN "dpdkstat"
+#define DPDK_STATS_NAME "dpdk_collectd_stats"
+
+#define DPDK_STATS_TRACE()                                                     \
+  DEBUG("%s:%s:%d pid=%u", DPDK_STATS_PLUGIN, __FUNCTION__, __LINE__, getpid())
+
+struct dpdk_stats_config_s {
+  cdtime_t interval;
+  uint32_t enabled_port_mask;
+  char port_name[RTE_MAX_ETHPORTS][DATA_MAX_NAME_LEN];
+};
+typedef struct dpdk_stats_config_s dpdk_stats_config_t;
 
 #define RTE_VERSION_16_07 RTE_VERSION_NUM(16, 7, 0, 16)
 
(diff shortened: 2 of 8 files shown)


docs/BUILD.dpdkstat.md
```diff
@@ -201,15 +201,19 @@ See also: http://dpdk.org/doc/guides/prog_guide/multi_proc_support.html
  *  Generate the build script as specified below. (i.e. run `build.sh`).
  *  Configure collectd with the DPDK shared library:
 
-        ./configure --with-libdpdk=/usr
+        ./configure LIBDPDK_CPPFLAGS="-I/usr/include/dpdk" LIBDPDK_LDFLAGS="-L/usr/lib"
```
rubenk: Setting these flags isn't needed anymore since 8259168611932117193bcf891b1c063a619d32e3

docs/BUILD.dpdkstat.md
```diff
@@ -201,15 +201,19 @@ See also: http://dpdk.org/doc/guides/prog_guide/multi_proc_support.html
...
+point out to dpdk libraries location.
 
 ### Build with the static DPDK library
 
 To configure collectd with the DPDK static library:
 
  *  Run *configure* with the following CFLAGS:
 
-        ./configure --with-libdpdk=/usr CFLAGS=" -lpthread -Wl,--whole-archive -Wl,-ldpdk -Wl,-lm -Wl,-lrt -Wl,-lpcap -Wl,-ldl -Wl,--no-whole-archive"
+        ./configure LIBDPDK_CPPFLAGS="-I/usr/include/dpdk" LIBDPDK_LDFLAGS="-L/usr/lib" CFLAGS=" -lpthread -Wl,--whole-archive -Wl,-ldpdk -Wl,-lm -Wl,-lrt -Wl,-lpcap -Wl,-ldl -Wl,--no-whole-archive"
```
rubenk: Same here. Unrelated to this PR, but the other libraries are probably not needed too.

docs/BUILD.dpdkstat.md
```diff
@@ -201,15 +201,19 @@ See also: http://dpdk.org/doc/guides/prog_guide/multi_proc_support.html
...
 
  *  Make sure that dpdk and dpdkstat are enabled in the *configure* output.
```
rubenk: libdpdk, might be nice to fix this too.

src/dpdkstat.c
```diff
@@ -18,55 +18,37 @@
  * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
  * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
- * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
- * SOFTWARE.
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
```
rubenk: What caused this rewrapping of the lines? They were already max 80 chars and this seems like unnecessary churn.

docs/BUILD.dpdkstat.md
krzysztofx: Correct me if I'm wrong : you mean that setting these flags is not needed if we set it as above since those are default values, right ?
If so, then what about clarifying that if user needs to set custom header/lib dpdk locations, then may set it via these variables,  exactly like in [8259168](https://github.com/collectd/collectd/commit/8259168611932117193bcf891b1c063a619d32e3) ./configure LIBDPDK_CPPFLAGS="-I/path/to/your/libdpdk_headers" LIBDPDK_LDFLAGS="-L/path/to/your/libdpdk.so -Wl,--no-as-needed"

>

src/dpdkstat.c
```diff
@@ -18,55 +18,37 @@
...
  *
  * Authors:
  *   Maryam Tahhan <maryam.tahhan@intel.com>
  *   Harry van Haaren <harry.van.haaren@intel.com>
- *   Taras Chornyi <tarasx.chornyi@intel.com>
- */
```
rubenk: Please don't remove someone elses authorship.

src/dpdkstat.c
```diff
@@ -88,742 +70,462 @@
...
-    ERROR("dpdkstat: Error in shm_cleanup in %s", __func__);
-    return err;
-  }
-  err = dpdk_shm_init(shm_xstats_size);
-  if (err) {
-    WARNING("dpdkstat: Error in shm_init in %s", __func__);
-    return err;
+static int dpdk_stats_config(oconfig_item_t *ci) {
+  DPDK_STATS_TRACE();
+  int ret = 0;
```
rubenk: dead assignment.

krzysztofx:
@rubenk 
Thanks for your review. I'll addres your comments ASAP.

regards
Krzysztof

docs/BUILD.dpdkstat.md
krzysztofx: let me verify that for static linking

docs/BUILD.dpdkstat.md
krzysztofx: ok

src/dpdkstat.c
krzysztofx: This  rework was branched out long time ago and many things have been changed including license note  reformatting and also list of authors.
I can align license note  in all  files in this PR to base version  if you prefer that.

src/dpdkstat.c
krzysztofx: Ok, that wasn't intentional. see my previous comment. I'll consult that with Taras

src/dpdkstat.c
```diff
@@ -88,742 +70,462 @@
...
-   * config function initializes SHM. If there is no config, then init() will
-   * just return.
-   */
-  if (g_configuration)
-    return 0;
+static int dpdk_stats_preinit(void) {
+  DPDK_STATS_TRACE();
 
-  char errbuf[ERR_BUF_SIZE];
+  int ret = 0;
```
krzysztofx: ok

docs/BUILD.dpdkstat.md
rubenk: I'm not entirely sure what you mean. We don't support static linking of external libraries into plugins.

krzysztofx: @rubenk 
That chapter is  dedicated to  static linking of DPDK libraries into the plugin. I didn't know that it's not supported in collectd.  What does it mean exactly - you just  recommend  not doing that , or there's some technical issue that will prevent user from running collectd with statically linked plugin  ? 

If it's static linking is not supported, then maybe that chapter should be removed at all, what do you think ? 

Regards 
Krzysztof

rubenk: Collectd explicitly disables static linking for libtool (https://github.com/collectd/collectd/blob/master/configure.ac#L14)

I never tried this ./configure line, if it works for you it's fine with me.

src/dpdkstat.c
rubenk: That would be nice, thank you.

docs/BUILD.dpdkstat.md
krzysztofx: It turned out that there's no clean and universal solution for static linking of DPDK. It changes from one DPDK version to another. We decided to remove all  related content from documentation.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+
+  size_t shm_size;
+  const char *shm_name;
+
+  sem_t sema_cmd_start;
+  sem_t sema_cmd_complete;
+  cdtime_t cmd_wait_time;
+
+  pid_t pid;
+#ifdef DPDK_HELPER_USE_PIPES
```
rubenk: This is always true, so why the #ifdef?

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+  DPDK_HELPER_NOT_INITIALIZED = 0,
+  DPDK_HELPER_INITIALIZING,
+  DPDK_HELPER_WAITING_ON_PRIMARY,
+  DPDK_HELPER_INITIALIZING_EAL,
+  DPDK_HELPER_ALIVE_SENDING_EVENTS,
+  DPDK_HELPER_GRACEFUL_QUIT,
+};
+
+#define DPDK_HELPER_TRACE(_name)                                               \
+  DEBUG("%s:%s:%d pid=%lu", _name, __FUNCTION__, __LINE__, (long)getpid())
```
rubenk: %ld for the pid please.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+  }
+
+  DPDK_HELPER_TRACE(phc->shm_name);
+
+  if (ec == NULL) {
+    ERROR("Invalid argument (ec)");
+    return -EINVAL;
+  }
+
+  memcpy(&phc->eal_config, ec, sizeof(dpdk_eal_config_t));
```
rubenk: I'd prefer that you call sizeof on the variable, not the type.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+  }
+
+  DPDK_HELPER_TRACE(phc->shm_name);
+
+  if (ec == NULL) {
+    ERROR("Invalid argument (ec)");
+    return -EINVAL;
+  }
+
+  memcpy(ec, &phc->eal_config, sizeof(dpdk_eal_config_t));
```
rubenk: Same here.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+      cf_util_get_string_buffer(child, phc->eal_config.process_type,
+                                sizeof(phc->eal_config.process_type));
+      DEBUG("dpdk_common: EAL:Process type %s", phc->eal_config.process_type);
+    } else if ((strcasecmp("FilePrefix", child->key) == 0) &&
+               (child->values[0].type == OCONFIG_TYPE_STRING)) {
+      ssnprintf(phc->eal_config.file_prefix, DATA_MAX_NAME_LEN,
+                "/var/run/.%s_config", child->values[0].value.string);
+      DEBUG("dpdk_common: EAL:File prefix %s", phc->eal_config.file_prefix);
+    }
+  }
```
rubenk: Please handle errors and unknown config keys in this loop.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+    WARNING("dpdk_shm_init: Failed to open %s as SHM:%s\n", name,
+            sstrerror(errno, errbuf, sizeof(errbuf)));
+    goto fail;
+  }
+
+  int ret = ftruncate(fd, size);
+  if (ret != 0) {
+    WARNING("dpdk_shm_init: Failed to resize SHM:%s\n",
+            sstrerror(errno, errbuf, sizeof(errbuf)));
+    goto fail_close;
```
rubenk: I'm not too familiar with SHM, but don't you need to destroy / delete / close it on failure?

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+static int dpdk_shm_init(const char *name, size_t size, void **map) {
+  DPDK_HELPER_TRACE(name);
+
+  char errbuf[ERR_BUF_SIZE];
+
+  int fd = shm_open(name, O_CREAT | O_TRUNC | O_RDWR, 0666);
+  if (fd < 0) {
+    WARNING("dpdk_shm_init: Failed to open %s as SHM:%s\n", name,
+            sstrerror(errno, errbuf, sizeof(errbuf)));
+    goto fail;
```
rubenk: I really dislike goto's, especially 2 labels in the same function. Case in point, Coverity found 2 issues in this function with the error handling. I'd prefer that you do the cleanup immediately and return.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+
+  *map = mmap(0, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+  if (*map == MAP_FAILED) {
+    WARNING("dpdk_shm_init:Failed to mmap SHM:%s\n",
+            sstrerror(errno, errbuf, sizeof(errbuf)));
+    goto fail_close;
+  }
+  /*
+   * Close the file descriptor, the shared memory object still exists
+   * and can only be removed by calling shm_unlink().
```
rubenk: I don't understand this comment. Any reason you don't call shm_unlink here?

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+
+  ret = shm_unlink(name);
+  if (ret) {
+    ERROR("shm_unlink returned %d\n", ret);
+  }
+
+  return 0;
+}
+
+inline void *dpdk_helper_priv_get(dpdk_helper_ctx_t *phc) {
```
rubenk: No explicit inlines please, let the compiler figure it out.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+  if (phc)
+    return (void *)phc->priv_data;
```
rubenk: unnecessary cast.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+    DPDK_CHILD_LOG("Invalid argument(phc)\n");
+    return -EINVAL;
+  }
+
+  return (phc->shm_size - sizeof(dpdk_helper_ctx_t));
+}
+
+int dpdk_helper_init(const char *name, size_t data_size,
+                     dpdk_helper_ctx_t **pphc) {
+  int err = 0;
```
rubenk: Initialisation has no effect, please declare err when you need it.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+  }
+
+  DEBUG("%s:%s:%d pid=%lu, cmd=%d", phc->shm_name, __FUNCTION__, __LINE__,
+        (long)getpid(), cmd);
+
+  phc->cmd_wait_time = cmd_wait_time;
+
+  int ret = 0;
+
+  ret = dpdk_helper_status_check(phc);
```
rubenk: Please don't write code like this.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+    return -errno;
+  }
+
+  err = sem_init(&phc->sema_cmd_complete, 1, 0);
+  if (err != 0) {
+    ERROR("sema_cmd_complete semaphore init failed: %s\n",
+          sstrerror(errno, errbuf, sizeof(errbuf)));
+    sem_destroy(&phc->sema_cmd_start);
+    dpdk_shm_cleanup(name, shm_size, (void *)phc);
+    return -errno;
```
rubenk: same here.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+          phc->pipes[1]);
+  }
+
+  if (pipe(phc->pipes) != 0) {
+    DEBUG("dpdk_helper_spawn: Could not create helper pipe: %s\n",
+          sstrerror(errno, errbuf, sizeof(errbuf)));
+    return -1;
+  }
+
+  int pipe0_flags = fcntl(phc->pipes[0], F_GETFL, 0);
```
rubenk: please check for error.

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+  if (err != NULL)
+    *err = res;
+  return val;
+}
+
+uint128_t str_to_uint128(const char *str, int len) {
+  uint128_t lcore_mask;
+  int err = 0;
+
+  memset(&lcore_mask, 0, sizeof(uint128_t));
```
rubenk: sizeof(lcore_mask)

src/utils_dpdk.c
```diff
@@ -0,0 +1,871 @@
...
+
+    memset(high_str, 0, sizeof(high_str));
+    memset(low_str, 0, sizeof(low_str));
+
+    strncpy(high_str, str, len - 16);
+    strncpy(low_str, str + len - 16, 16);
+
+    lcore_mask.low = strtoull_safe(low_str, &err);
+    if (err)
+      goto parse_out;
```
rubenk: please just get rid of the goto and return immediately.

src/utils_dpdk.c
krzysztofx: All logic related with redirecting stdout from child to parent process was wrapped with  DPDK_HELPER_USE_PIPES ifdef just to quickly disable it in future if it's no longer needed. That redirecting was implemented because child process gets important info from  DPDK EAL into it's stdout.  If in some future DPDK version all EAL initialization info will be logged via system logger then we may disable those pipes quickly.
Do you prefer not to contaminate source code with these ifdefs?

rubenk: Yes, I'd prefer if you removed those ifdefs. If they are needed in the future, let's add them then.

src/utils_dpdk.c
krzysztofx: Yes, to be fixed. Additionally  shm_unlink should be called if  shm_open succeeded.

src/utils_dpdk.c
krzysztofx: No, I can't call shm_unlink since this is positive scenario, I mean shared memory was successfully initialized. File descriptor is no longer needed for shm operations and can be closed. I guess  that was what comment author wanted to say.
I'll  modify this comment.

src/utils_dpdk.c
krzysztofx: ok

src/utils_dpdk.c
krzysztofx: ok

src/utils_dpdk.h
```diff
@@ -0,0 +1,97 @@
...
+#define ERR_BUF_SIZE 1024
+
+#if RTE_VER_RELEASE == 16 && RTE_VER_MINOR == 0
+#if RTE_VER_MONTH == 4
+#define DPDK_VER_16_04 RTE_VERSION_NUM(16, 4, 0, 16)
+#endif
+#if RTE_VER_MONTH == 7
+#define DPDK_VER_16_07 RTE_VERSION_NUM(16, 7, 0, 16)
+#endif
+#endif
```
krzysztofx: Yes. I will remove it

krzysztofx:
@rubenk 
I have addressed your comments in the last patch. There are some merge conflicts, so should I rebase to master branch ?

Thanks
Krzysztof

rubenk:
Thank you @krzysztofx. Yes, please rebase.

krzysztofx:
@rubenk 
Rebase done.  As expected, epel5 builds are  failing due to unsupported socket flags. All other builds passed. 

Thanks
Krzysztof

PR #2097, opened 2016-12-10
zfs_arc updates for linux

YmrDtnJu:
Linux seems to have changed the layout of the statistics file. The first commit skips the first two lines of the file as they contain two headers and no statistics. The second commit removes the "other_size" gauge for Linux because it is not available (anymore).

### DIFF
--- a/src/zfs_arc.c
+++ b/src/zfs_arc.c
@@ -212,6 +212,21 @@ static int za_read(void) {
     return (-1);
   }
 
+  // Ignore the first two lines because they contain information about
+  // the rest of the file.
+  // See kstat_seq_show_headers module/spl/spl-kstat.c of the spl kernel
+  // module.
+  if (fgets(buffer, sizeof(buffer), fh) == NULL) {
+    ERROR("zfs_arc plugin: \"%s\" does not contain a single line.", ZOL_ARCSTATS_FILE);
+    fclose(fh);
+    return (-1);
+  }
+  if (fgets(buffer, sizeof(buffer), fh) == NULL) {
+    ERROR("zfs_arc plugin: \"%s\" does not contain at least two lines.", ZOL_ARCSTATS_FILE);
+    fclose(fh);
+    return (-1);
+  }
+
   while (fgets(buffer, sizeof(buffer), fh) != NULL) {
     char *fields[3];
     value_t v;
--- a/src/zfs_arc.c
+++ b/src/zfs_arc.c
@@ -75,7 +75,6 @@ static long long get_zfs_value(kstat_t *ksp, const char *key) {
 
   e = llist_search(ksp, key);
   if (e == NULL) {
-    ERROR("zfs_arc plugin: `llist_search` failed for key: '%s'.", key);
     return (-1);
   }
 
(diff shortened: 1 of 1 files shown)


YmrDtnJu:
Sorry, I cannot see the problem. The last console output from the test says: "found this new plugin: *.so". It seems, a shopt -s nullglob is missing somewhere in the script to check for new plugins? I am quite sure this has nothing to do with my code changes because they to not introduce any new plugins and are actually just minimal changes. Additionally, the code passed all 13 tests and all modules including zfs_arc have been build successfully.

Please guide me to the problem to fix. Thank you.

rubenk:
Thanks for your contribution @YmrDtnJu.

The test failure is unrelated, I'll look into it.

rubenk:
Hmm, if the statistics file will keep on changing we're going to have a problem.

The other_size gauge seems to have been replaced by 3 other stats in https://github.com/zfsonlinux/zfs/commit/25458cbef9e59ef9ee6a7e729ab2522ed308f88f, which isn't in a stable release yet. Removing other_size from collectd will break existing setups. Adding those new counters will throw a lot of errors if they don't exist.

I don't know what the old stats file looked like and how the new stats file looks, but isn't there a chance that we skip over valid stats in the first two lines with the old stats file?

YmrDtnJu:
Removing other_size is required because it will/does not exist anymore. If that breaks setups, it is because of changes in zfs on Linux and not collectd. I can change the patch to only try to get other_size and not fail if it is not there anymore. This way, those people using unstable versions of zfs don't get error messages in their logs and those who use stable versions get that metric. Sooner or later (more sooner than later) people will have to adobt anyway.

The first line is information about the stats read from the rest of the file, e.g. the timestamp the stats were taken, the type of stats (key-value-pairs, in this case). They themselves are not in key-value pairs and were not used before (but they also did not produce an error message).

The second line is just a header. It contains the names of the columns of the following lines that contain the metrics. Skipping this line is required, because the rest of the existing code assumes the third column not to contain strings.

rubenk:
> Removing other_size is required because it will/does not exist anymore. If that breaks setups, it is because of changes in zfs on Linux and not collectd. I can change the patch to only try to get other_size and not fail if it is not there anymore.

That seems like a good solution.
Another one would be to turn the errors for missing keys into debug messages, since for users it's not really something they can so anything about except for patching the source.

> This way, those people using unstable versions of zfs don't get error messages in their logs and those who use stable versions get that metric. Sooner or later (more sooner than later) people will have to adobt anyway.

For long term support distros it can take years to update to a new version of zfs.
If we can prevent any breakage for those setups we should.

> The first line is information about the stats read from the rest of the file, e.g. the timestamp the stats were taken, the type of stats (key-value-pairs, in this case). They themselves are not in key-value pairs and were not used before (but they also did not produce an error message).

> The second line is just a header. It contains the names of the columns of the following lines that contain the metrics. Skipping this line is required, because the rest of the existing code assumes the third column not to contain strings.

Ok, thanks for the explanation.

YmrDtnJu:
Sorry for the long delay.

I hope I did the right thing by commiting a new change to my pull request branch. This commit removes the logging in get_zfs_value completly and degrades the warnings in za_read_derive and za_read_gauge to debug.

rubenk:
Thanks for the update @YmrDtnJu. If you could drop the second commit from your PR (git rebase -i), then I'll merge.

YmrDtnJu:
Done. :-) Thank you.

Issue #2098, opened 2016-12-11
All plugins should be consistent about handling boolean config options


tokkee:
This is a follow-up to #2083 and something we should clean up in an 6.0 release. Different plugins support a mix of string and boolean values for boolean config options. In particular, the old (simple) config API does not allow to separate between the two. Ideally, we change all plugins ahead of a 6.0 release and introduce deprecation warnings.

octo:
Merged to *master*.

Issue #2083, opened 2016-12-06
network plugin: forwading broken since 5.6


bzed:
*   Version of collectd: 5.6.2.2.g11011dc-1~jessie
*   Operating system / distribution:
debian jessie

## Expected behavior

The network plugin receives + forewards packages.

## Actual behavior

Packets are received, written to the local output plugin (graphite in this case). Values from the local host are sent to the Server, but not the received values.

## Steps to reproduce

Install packages from pkg.ci, use the following config (same config works perfectly fine with 5.5):

```

# Generated by Puppet

#Hostname localhost
FQDNLookup true

#BaseDir "/var/lib/collectd"
#PluginDir "/usr/lib/collectd"
TypesDB "/usr/share/collectd/types.db" "/etc/collectd/conova_types.db"
WriteQueueLimitHigh 5000000
WriteQueueLimitLow 2500000
Interval 10
Timeout 2
ReadThreads 5
Include "/etc/collectd/conf.d/*.conf"
Include "/etc/collectd/conf_local.d/*.conf"
# Generated by Puppet
<LoadPlugin syslog>
  Globals false
</LoadPlugin>

<Plugin syslog>
  LogLevel info
</Plugin>

# Generated by Puppet
<LoadPlugin apache>
  Globals false
</LoadPlugin>

<Plugin "apache">
  <Instance "apache">
    URL "http://127.0.0.1/?auto"
    User "collectd-apache"
    Password "password"
  </Instance>
</Plugin>

# Generated by Puppet
<LoadPlugin contextswitch>
  Globals false
</LoadPlugin>


# Generated by Puppet
<LoadPlugin cpu>
  Globals false
</LoadPlugin>

<Plugin cpu>
  ReportByState = true
  ReportByCpu = true
  ValuesPercentage = false
</Plugin>

# Generated by Puppet
<LoadPlugin df>
  Globals false
</LoadPlugin>

<Plugin df>
  IgnoreSelected false
  ReportByDevice false
  ReportInodes true
  ReportReserved true
  ValuesAbsolute true
  ValuesPercentage true
</Plugin>

# Generated by Puppet
<LoadPlugin disk>
  Globals false
</LoadPlugin>

<Plugin disk>
  IgnoreSelected false
</Plugin>

# Generated by Puppet
<LoadPlugin entropy>
  Globals false
</LoadPlugin>


# Generated by Puppet
<LoadPlugin interface>
  Globals false
</LoadPlugin>

<Plugin interface>
[...]

PR #2099, opened 2016-12-11
configfile: Reintroduce support for boolean string config values.

tokkee:
For the network plugin, this was changed in ac73c75aed7 (which landed in 5.6)
which was a backward incompatible change breaking user configuration. Adding
support back in a central location ensures a more consistent behavior across
plugins. At the same time, we issue a warning message that this behavior is
deprecated.

GH #2083, #2098


octo:
+1: While I dislike allowing users to do stupid things like quoting booleans, I think the benefits (backwards compatibility) outweigh the downsides.

The code is spotless, thanks @tokkee!

PR #2100, opened 2016-12-12
virt plugin: handle disconnect on inst#0

mojaves:
Only one virt reader instance should take care of connection
handling (connection/disconnection) to avoid races and plugin
data corruption, potentially crashing collectd.

This bug cannot be triggered with instances=1 (default settings).

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -520,6 +520,13 @@ static int lv_connect(void) {
   return 0;
 }
 
+static void lv_disconnect(void) {
+  if (conn != NULL)
+    virConnectClose(conn);
+  conn = NULL;
+  WARNING(PLUGIN_NAME " plugin: closed connection to libvirt");
+}
+
 static int lv_read(user_data_t *ud) {
   time_t t;
   struct lv_read_instance *inst = NULL;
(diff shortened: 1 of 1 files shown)


rubenk:
This breaks CI.

src/virt.c
```diff
@@ -536,9 +543,9 @@ static int lv_read(user_data_t *ud) {
   if ((last_refresh == (time_t)0) ||
       ((interval > 0) && ((last_refresh + interval) <= t))) {
     if (refresh_lists(inst) != 0) {
-      if (conn != NULL)
-        virConnectClose(conn);
-      conn = NULL;
+      if (inst->id == 0) {
```
rubenk: Please remove the braces.

mojaves: oops, sorry. Funny that clang-format doesn't complain.

rubenk: Yeah, that's what I thought. Especially since it got it right a few lines higher.

rubenk:
```
virt.c: In function 'lv_disconnect':
virt.c:512:1: error: expected ';' before '}' token
 }
 ^
Makefile:4888: recipe for target 'virt_la-virt.lo' failed
make[3]: *** [virt_la-virt.lo] Error 1
```

mojaves:
I think the only way to unlock this is to rebase: the "resolve conflicts" button on the github UI leads me to 404 :)

rubenk:
The button worked for me, looks like GitHub is A/B testing it ;)

Thanks!

PR #2101, opened 2016-12-12
virt plugin: Open connection on init()

mojaves:
If we use more than one reader instance, we can spot errors like
this in the system logs:

Dec 12 09:59:24 $HOST collectd[19338]: reading number of
domains: invalid connection pointer in virConnectNumOfDomains
Dec 12 09:59:24 benji.rokugan.lan collectd[19338]: read-function of
plugin `virt-2' failed. Will suspend it for 20.000 seconds.

This causes unnecessary delay in the sampling of libvirt.
The reason for this is just one instance (always #0) takes care
of establishing the libvirt connection.

But this could be done safely in the plugin init callback: according
to doc, this function is called at least once before all the read
instances.

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -504,6 +504,22 @@ static int lv_config(const char *key, const char *value) {
   return -1;
 }
 
+static int lv_connect(void) {
+  if (conn == NULL) {
+    /* `conn_string == NULL' is acceptable. */
+    conn = virConnectOpenReadOnly(conn_string);
+    if (conn == NULL) {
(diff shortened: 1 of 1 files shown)


src/virt.c
```diff
@@ -732,6 +740,8 @@ static int lv_init(void) {
   if (virInitialize() != 0)
     return -1;
 
+  lv_connect();
```
rubenk: Any reason you don't check for errors here?

rubenk: Ah right. A warning could certainly do no harm.

Issue #2068, opened 2016-12-01
Types.db.d


axos88:
*   Version of collectd: 5.6.0
*   Operating system / distribution: ubuntu 16.04

## Expected behavior

Add include stnaza to typed.db, so that additions can be handled in an automated way, using say chef.

It's function should identical to the well known .d directories.

axos88:
pending feedback?...

axos88:
@octo pending feeback?...

axos88:
Current behaviour: Failed to open types database `/opt/collectd/share/collectd/types.d/*.db': No such file or directory

axos88:
Workaround: Create an entry in conf.d and add the matching TypesDB stanza there (there can be multiple)

octo:
Hi @axos88,

thank you very much for opening this feature request. Your comments mix two concepts: include statements and globbing.

*   **Include statements** would allow to reference other files from within the *types.db* file. I think changing / adding to the *types.db* syntax is not desirable.
*   **Globbing** allows you to specify `TypesDB "/path/to/types.db*"` in the *collectd.conf(5)* config file, which will expand the wildcard character `*` and load all matching files. This is something we can look into.

As a workaround, you could do something like this as part of your config management:

```sh
for f in /etc/collectd/types*
do
  echo "TypesDB \"$f\""
done >/etc/collectd/collectd.conf.d/typesdb.conf
```

Then the standard `Include "/etc/collectd/collectd.conf.d/*.conf"` should pick it up.

axos88:
Hi, @octo,

Yeah, I'm using a similar workaround actually.
I'm pretty fine with the globbing include solution, as it would solve the issue.

PR #2103, opened 2016-12-13
virt plugin: support virDomainBlockStatsFlags

mojaves:
We want to export more disk stats, like disk_time.
To do so, we must use the virDomainBlockStatsFlags API, but it
could not be available on all the platforms collectd supports.
To cope with that, we add a minimal layer around the API, and
we use virDomainBlockStatsFlags only if available.

This patch adds the code to use the new API and the fallback code
in the case the platform' libvirt is too old.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -8076,6 +8076,12 @@ How many read instances you want to use for this plugin. The default is one,
 and the sensible setting is a multiple of the B<ReadThreads> value.
 If you are not sure, just use the default setting.
 
+=item B<ExtraStats> B<string>
+
+Enable extra statistics. This allows the plugin to reported more detailed
+statistics about the behaviour of Virtual Machines. The argument is a
+space-separated list of selectors. Currently only B<disk> is supported.
+
 =back
 
 =head2 Plugin C<vmem>
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -8078,9 +8078,13 @@ If you are not sure, just use the default setting.
 
 =item B<ExtraStats> B<string>
 
-Enable extra statistics. This allows the plugin to reported more detailed
-statistics about the behaviour of Virtual Machines. The argument is a
-space-separated list of selectors. Currently only B<disk> is supported.
+Report additional extra statistics. The default is no extra statistics, preserving
+the previous behaviour of the plugin. If unsure, leave the default. If enabled,
+allows the plugin to reported more detailed statistics about the behaviour of
+Virtual Machines. The argument is a space-separated list of selectors.
+Currently supported selectors are:
+B<disk> report extra statistics like number of flush operations and total
+service time for read, write and flush operations.
 
 =back
 
--- a/src/virt.c
+++ b/src/virt.c
@@ -380,6 +380,10 @@ static void disk_submit(struct lv_block_info *binfo, virDomainPtr dom,
   if ((binfo->bi.rd_bytes != -1) && (binfo->bi.wr_bytes != -1))
     submit_derive2("disk_octets", (derive_t)binfo->bi.rd_bytes,
                    (derive_t)binfo->bi.wr_bytes, dom, type_instance);
+
+  if ((binfo->rd_total_times != -1) && (binfo->wr_total_times != -1))
+    submit_derive2("disk_time", (derive_t)binfo->rd_total_times,
+                   (derive_t)binfo->wr_total_times, dom, type_instance);
 }
 
 static int lv_config(const char *key, const char *value) {
--- a/src/virt.c
+++ b/src/virt.c
@@ -60,6 +60,7 @@ static const char *config_keys[] = {"Connection",
                                     "PluginInstanceFormat",
 
                                     "Instances",
+                                    "ExtraStats",
 
                                     NULL};
 #define NR_CONFIG_KEYS ((sizeof config_keys / sizeof config_keys[0]) - 1)
@@ -163,6 +164,12 @@ enum bd_field { target, source };
 
 /* InterfaceFormat. */
 enum if_field { if_address, if_name, if_number };
+
+/* ExtraStats */
+#define EX_STATS_MAX_FIELDS 8
+#define EX_STATS_DISK "disk"
+enum ex_stats { ex_stats_none = 0, ex_stats_disk = 1 };
+static enum ex_stats extra_stats = ex_stats_none;
 
 /* BlockDeviceFormatBasename */
 _Bool blockdevice_format_basename = 0;
@@ -386,17 +393,19 @@ static void disk_submit(struct lv_block_info *binfo, virDomainPtr dom,
     submit_derive2("disk_octets", (derive_t)binfo->bi.rd_bytes,
                    (derive_t)binfo->bi.wr_bytes, dom, type_instance);
 
-  if ((binfo->rd_total_times != -1) && (binfo->wr_total_times != -1))
-    submit_derive2("disk_time", (derive_t)binfo->rd_total_times,
-                   (derive_t)binfo->wr_total_times, dom, type_instance);
+  if (extra_stats & ex_stats_disk) {
+    if ((binfo->rd_total_times != -1) && (binfo->wr_total_times != -1))
+      submit_derive2("disk_time", (derive_t)binfo->rd_total_times,
+                     (derive_t)binfo->wr_total_times, dom, type_instance);
 
-  if (binfo->fl_req != -1)
-    submit(dom, "total_requests", flush_type_instance,
-           &(value_t){.derive = (derive_t)binfo->fl_req}, 1);
-  if (binfo->fl_total_times != -1) {
-    derive_t value = binfo->fl_total_times / 1000; // ns -> ms
-    submit(dom, "total_time_in_ms", flush_type_instance,
-           &(value_t){.derive = value}, 1);
+    if (binfo->fl_req != -1)
+      submit(dom, "total_requests", flush_type_instance,
+             &(value_t){.derive = (derive_t)binfo->fl_req}, 1);
+    if (binfo->fl_total_times != -1) {
+      derive_t value = binfo->fl_total_times / 1000; // ns -> ms
+      submit(dom, "total_time_in_ms", flush_type_instance,
+             &(value_t){.derive = value}, 1);
+    }
   }
 }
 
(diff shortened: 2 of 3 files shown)


src/virt.c
```diff
@@ -306,6 +371,17 @@ static void submit_derive2(const char *type, derive_t v0, derive_t v1,
...
+static void disk_submit(struct lv_block_info *binfo, virDomainPtr dom,
+                        const char *type_instance) {
+  if ((binfo->bi.rd_req != -1) && (binfo->bi.wr_req != -1))
+    submit_derive2("disk_ops", (derive_t)binfo->bi.rd_req,
+                   (derive_t)binfo->bi.wr_req, dom, type_instance);
+
+  if ((binfo->bi.rd_bytes != -1) && (binfo->bi.wr_bytes != -1))
+    submit_derive2("disk_octets", (derive_t)binfo->bi.rd_bytes,
+                   (derive_t)binfo->bi.wr_bytes, dom, type_instance);
+}
```
rubenk: I might be missing something, but it looks like you're not sending the new statistics you fetch from libvirt?

rubenk: Yeah that would be nice.

mojaves:
@rubenk, @octo 

Next, I'd like to add the block device "flush" stats:

- fl_req: total flush requests 

- fl_total_times: total time (ns) spent on cache flushing 

But I'm not sure how to map them. I was thinking to extend the existing types like this:

diff --git a/src/types.db b/src/types.db
index 6855187..bc2985c 100644
--- a/src/types.db
+++ b/src/types.db
@@ -52,8 +52,10 @@ disk_latency            read:GAUGE:0:U, write:GAUGE:0:U
 disk_merged             read:DERIVE:0:U, write:DERIVE:0:U
 disk_octets             read:DERIVE:0:U, write:DERIVE:0:U
 disk_ops                read:DERIVE:0:U, write:DERIVE:0:U
+disk_ops_flush          read:DERIVE:0:U, write:DERIVE:0:U, flush:DERIVE:0:U
 disk_ops_complex        value:DERIVE:0:U
 disk_time               read:DERIVE:0:U, write:DERIVE:0:U
+disk_time_flush         read:DERIVE:0:U, write:DERIVE:0:U, flush:DERIVE:0:U
 dns_answer              value:DERIVE:0:U
 dns_notify              value:DERIVE:0:U
 dns_octets              queries:DERIVE:0:U, responses:DERIVE:0:U

Or perhaps I can add a new type for disk_flush. What do you think?

src/virt.c
```diff
@@ -380,6 +380,10 @@ static void disk_submit(struct lv_block_info *binfo, virDomainPtr dom,
   if ((binfo->bi.rd_bytes != -1) && (binfo->bi.wr_bytes != -1))
     submit_derive2("disk_octets", (derive_t)binfo->bi.rd_bytes,
                    (derive_t)binfo->bi.wr_bytes, dom, type_instance);
+
+  if ((binfo->rd_total_times != -1) && (binfo->wr_total_times))
```
rubenk: What if wr_total_times is 0? Isn't that a valid reading?

mojaves: Silly mistake, fixing.

rubenk:
We tend to be somewhat hesitant to adding new types and map stats to existing types as much as possible. I'm not sure what the best type would be in this case. cc'ing @octo who might have an idea.

mojaves:
@rubenk @octo perhaps we can avoid to extend the default types.db. We can ship our own amended, or add a snippet if you accept https://github.com/collectd/collectd/issues/2068 .

Is it ok to submit some stats not declared in the default types.db? e.g. something like

+  submit_derive2("disk_flush", (derive_t)binfo->fl_req,
+                     (derive_t)binfo->fl_total_times, dom, type_instance);

octo:
@mojaves you get the best *types* when you think of them as "units". For the most part, I think you answer the question youself:

> fl_req: total flush requests

→ type `total_requests`, type instance `flush`

> fl_total_times: total time (ns) spent on cache flushing

Add a new entry to *types.db*:

```
total_time_in_ns        value:DERIVE:0:U
```

Best regards,
—octo

mojaves:
@octo thanks! will update my patch soon.

mojaves:
I see we have already total_time_in_ms in types.db. I will investigate if ms resolution is good for us, so we can avoid adding a new type.

mojaves:
Looks like the ms resolution is sufficient for use for the time being. We'll discuss in the future a finer-grained resolution (ns) should the need arise.
I had to update my last commit because we need to report per-disk flush stats, so I added the disk name in the type instance. It's not the most elegant solution out there, but should be good enough for a start. Suggestions of course welcome =)

Other than that the patches are under testing on my side. So far, so good, no issues spotted.

mojaves:
> Other than that the patches are under testing on my side. So far, so good, no issues spotted.
tests completed on my side, no issues.

mojaves:
ping?

mojaves:
ping?

mojaves:
hit wrong button, reponened

src/virt.c
```diff
@@ -640,27 +770,22 @@ static int lv_read(user_data_t *ud) {
 
   /* Get block device stats for each domain. */
   for (int i = 0; i < state->nr_block_devices; ++i) {
-    struct _virDomainBlockStats stats;
+    struct block_device *bdev = &(state->block_devices[i]);
+    int ret;
```
rubenk: can you declare this on line 778?

rubenk:
Hi @mojaves, I'm sorry for the late response. Life got in the way..

In general this looks good, I left one minor comment.
One thing I've been thinking about, is if it would be possible to make gathering the new statistics optional. My guess is not everyone is interested in flush requests for example, I know I have no use for it in my environment but they will still be gathered and stored for thousands of vms. It would be great if we could make it configurable which stats to gather. I'd love to hear your thoughts on this.

src/virt.c
mojaves: Sure, will move.

mojaves:
Hi @rubenk , no problem, I totally understand.

I'm totally fine with configurable stats, I also think it is a good idea. Let me refresh my patches.

fromanirh:
sorry about the mess, I'm transitioning from my personal account (@mojaves) to my work account (@fromanirh). Should be OK with my last push. If it isn't, I'm fine opening a new cleaner pull request.

mojaves:
This is the confirmation from the other side that I'm transitioning my collectd contributions from @mojaves (personal account) to @fromanirh (work account)

src/virt.c
```diff
@@ -500,6 +603,24 @@ static int lv_config(const char *key, const char *value) {
     return 0;
   }
 
+  if (strcasecmp(key, "ExtraStats") == 0) {
+    char *localvalue = sstrdup(value);
+    if (localvalue != NULL) {
+      char *exstats[EX_STATS_MAX_FIELDS];
+      int numexstats;
+
+      numexstats = strsplit(localvalue, exstats, STATIC_ARRAY_SIZE(exstats));
```
rubenk: can you combine line 610 and 612?

src/virt.c
```diff
@@ -527,6 +648,42 @@ static void lv_disconnect(void) {
...
 
+static int lv_domain_block_info(virDomainPtr dom, const char *path,
+                                struct lv_block_info *binfo) {
+#ifdef HAVE_BLOCK_STATS_FLAGS
+  virTypedParameterPtr params = NULL;
+  int nparams = 0;
+  int rc = -1;
+  int ret;
+
+  ret = virDomainBlockStatsFlags(dom, path, NULL, &nparams, 0);
```
rubenk: please combine this with line 657

src/virt.c
```diff
@@ -640,27 +797,20 @@ static int lv_read(user_data_t *ud) {
...
-    if (virDomainBlockStats(state->block_devices[i].dom,
-                            state->block_devices[i].path, &stats,
-                            sizeof stats) != 0)
+    if (lv_domain_block_info(bdev->dom, bdev->path, &binfo) < 0)
       continue;
 
     char *type_instance = NULL;
     if (blockdevice_format_basename && blockdevice_format == source)
-      type_instance = strdup(basename(state->block_devices[i].path));
+      type_instance = strdup(basename(bdev->path));
```
rubenk: please check for failure of strdup()

src/virt.c
```diff
@@ -640,27 +797,20 @@ static int lv_read(user_data_t *ud) {
...
     else
-      type_instance = strdup(state->block_devices[i].path);
-
-    if ((stats.rd_req != -1) && (stats.wr_req != -1))
-      submit_derive2("disk_ops", (derive_t)stats.rd_req, (derive_t)stats.wr_req,
-                     state->block_devices[i].dom, type_instance);
+      type_instance = strdup(bdev->path);
```
rubenk: same here.

src/collectd.conf.pod
```diff
@@ -8076,6 +8076,12 @@ How many read instances you want to use for this plugin. The default is one,
 and the sensible setting is a multiple of the B<ReadThreads> value.
 If you are not sure, just use the default setting.
 
+=item B<ExtraStats> B<string>
+
+Enable extra statistics. This allows the plugin to reported more detailed
+statistics about the behaviour of Virtual Machines. The argument is a
+space-separated list of selectors. Currently only B<disk> is supported.
```
rubenk: I accidentally interpreted this as "this enables disk statistics in general". Might be nice to add a note that setting "disk" enables extra dist statistics like flush requests, with the emphasis on extra.

src/collectd.conf.pod
```diff
@@ -8076,6 +8076,12 @@ How many read instances you want to use for this plugin. The default is one,
 and the sensible setting is a multiple of the B<ReadThreads> value.
 If you are not sure, just use the default setting.
 
+=item B<ExtraStats> B<string>
+
+Enable extra statistics. This allows the plugin to reported more detailed
```
rubenk: report

rubenk:
Thanks @fromanirh. I left just a few minor nits and then this is good to go in.

src/virt.c
fromanirh: Done

src/virt.c
fromanirh: Done

src/virt.c
fromanirh: Done, but rearranging the code in a slightly and hopefully better way

src/virt.c
fromanirh: ditto

src/collectd.conf.pod
fromanirh: Done

PR #2105, opened 2016-12-14
snmp_agent: Implement SNMP Agent plugin.

rkorynkx:
The SNMP Agent plugin is an AgentX subagent that receives and handles queries from SNMP master agent and returns the data collected by read plugins. The SNMP Agent plugin handles requests only for OIDs specified in configuration file. To handle SNMP queries the plugin gets data from collectd and translates requested values from collectd's internal format to SNMP format. This plugin is a generic plugin and cannot work without configuration. For more details on AgentX subagent see <http://www.net-snmp.org/tutorial/tutorial-5/toolkit/demon/>


rubenk:
Did you forget to check-in snmp.agent.c and snmp_agent.h?

rkorynkx:
Hi @rubenk, snmp_agent.c is already there. Working on compilation fix.

rkorynkx:
Hi @octo, buildbot/collectd-solaris10-i386 task is in pending state more than 2 days but in Details it shows that is has finished successfully. Could you please take a look? Should I perform any actions?

octo:
@rkorynkx the build finished successfully but the status change didn't make it to GitHub. :-(

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+
+  if (dd == NULL || *dd == NULL)
+    return;
+
+  /* unregister scalar type OID */
+  if ((*dd)->table == NULL) {
+    for (int i = 0; i < (*dd)->oids_len; i++)
+      unregister_mib((*dd)->oids[i].oid, (*dd)->oids[i].oid_len);
+  } else {
+    /* unregister all table OIDs */
```
octo: Why don't you just call `snmp_agent_free_table((*dd)->table)` here?

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+    data_definition_t *dd = (data_definition_t *)de->value;
+    snmp_agent_free_data(&dd);
+  }
+
+  llist_destroy((*td)->columns);
+
+  void *key = NULL;
+  void *value = NULL;
+
+  c_avl_destroy((*td)->index_instance);
```
octo: Please add a comment explaining that `index_instance` and `instance_index` contain the *same* pointers (assuming that that's what's happening).

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+
+  pthread_mutex_lock(&g_agent->lock);
+
+  memcpy(oid.oid, requests->requestvb->name,
+         sizeof(oid.oid[0]) * requests->requestvb->name_length);
+  oid.oid_len = requests->requestvb->name_length;
+
+  snmp_agent_oid_to_string(temp, sizeof(temp), &oid);
+
+  DEBUG(PLUGIN_NAME ": Get request received for table OID '%s'.", temp);
```
octo: `temp` is not a good name for this variable. How about `oid_str`? It should likely be in an `#if COLLECT_DEBUG` block to avoid unused variable warnings when debugging is disabled.

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+            return SNMP_ERR_NOERROR;
+          }
+
+          format_name(temp, sizeof(temp), hostname_g, dd->plugin, instance,
+                      dd->type, dd->type_instance);
+          DEBUG(PLUGIN_NAME ": Identifier '%s'", temp);
+
+          value_t *values;
+          size_t values_num;
+          const data_set_t *ds;
```
octo: Please assign on the same line:

```c
const data_set_t *ds = plugin_get_ds(dd->type);
```

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+
+            return SNMP_ERR_NOERROR;
+          }
+
+          format_name(temp, sizeof(temp), hostname_g, dd->plugin, instance,
+                      dd->type, dd->type_instance);
+          DEBUG(PLUGIN_NAME ": Identifier '%s'", temp);
+
+          value_t *values;
+          size_t values_num;
```
octo: Please initialize `values` and `values_num` to `NULL` / zero so that it's obvious that they are never used uninitialized.

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+          (snmp_oid_ncompare(oid.oid, oid.oid_len, td->index_oid.oid,
+                             td->index_oid.oid_len,
+                             MIN(oid.oid_len, td->index_oid.oid_len)) == 0)) {
+
+      DEBUG(PLUGIN_NAME ": Handle '%s' table index OID.", td->name);
+
+      int index = oid.oid[oid.oid_len - 1];
+      char *instance;
+
+      int ret = c_avl_get(td->index_instance, &index, (void **)&instance);
```
octo: Compound literals are great for when you need to pass a pointer but don't actually need about the value that's assigned. By using an anonymous variable, you're signaling to the reader that you don't care what's assigned.

```c
int ret = c_avl_get(td->index_instance, &index, &(void *){NULL});
```

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+
+  pthread_mutex_lock(&g_agent->lock);
+
+  memcpy(oid.oid, requests->requestvb->name,
+         sizeof(oid.oid[0]) * requests->requestvb->name_length);
+  oid.oid_len = requests->requestvb->name_length;
+
+  snmp_agent_oid_to_string(temp, sizeof(temp), &oid);
+
+  DEBUG(PLUGIN_NAME ": Get request received for scalar OID '%s'.", temp);
```
octo: As above: s/temp/oid_str/?

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+  }
+
+  pthread_mutex_unlock(&g_agent->lock);
+
+  return SNMP_NOSUCHINSTANCE;
+}
+
+
+static int
+snmp_agent_scalar_oid_handler(struct netsnmp_mib_handler_s *handler,
```
octo: This function has a lot of duplicate code that's also in `snmp_agent_table_oid_handler()`. Is it possible to deduplicate this a bit?

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+    snmp_agent_free_data(&dd);
+    return (-ENOMEM);
+  }
+
+  llist_append(td->columns, entry);
+
+  return (0);
+}
+
+static int snmp_agent_config_data(oconfig_item_t *ci) {
```
octo: This duplicates a lot of code from `snmp_agent_config_table_data()`. Could this be simplified?

src/snmp_agent.c
rkorynkx: snmp_agent_free_data() frees scalar and tabular data. If data is tabular it will remove all rows of table for specified column. For example, we have unicast and broadcast packets counters as columns in the table, so if to call this  function with broadcast dd as a parameter it will free data for each interface for broadcast counter only. In turn, snmp_agent_free_table() iterates through all columns (dd) and calls snmp_agent_free_data().

src/snmp_agent.c
```diff
@@ -0,0 +1,1513 @@
...
+  }
+
+  sfree((*dd)->name);
+  sfree((*dd)->plugin);
+  sfree((*dd)->plugin_instance);
+  sfree((*dd)->type);
+  sfree((*dd)->type_instance);
+  sfree((*dd)->oids);
+
+  sfree(*dd);
```
rkorynkx: Actually it's not. dd is not the owner of the table as it does not allocate it. It holds the reference to the table it belongs to. I will make it const for better understanding.

rkorynkx:
Hi @octo, I'm seeing Makefile.am has changed and currently there are conflicts. Should I rebase the branch with the latest master to solve this?

src/snmp_agent.c
```diff
@@ -0,0 +1,1478 @@
...
+      data_definition_t *dd = de->value;
+
+      for (int i = 0; i < dd->oids_len; i++) {
+        int ret = snmp_oid_ncompare(oid.oid, oid.oid_len, dd->oids[i].oid,
+                                    dd->oids[i].oid_len,
+                                    MIN(oid.oid_len, dd->oids[i].oid_len));
+        if (ret != 0)
+          continue;
+
+        if (!td->index_oid.oid_len) {
```
octo: You can do this right after initializing `td` and avoid two nested loops that will never be able to perform any useful action.

src/snmp_agent.c
rkorynkx: data_definition_t is used not only for table data type, it's also used for scalar data type. And in this case columns won't be the proper name.

rkorynkx:
Hi @octo, thanks for quick review. I'll fix the review comments. Have one question, should I do rebase and merge conflicts as a separate commit after new patch (with fixed review comments) will be uploaded? What is the procedure?

Thanks,
  Roman.

rkorynkx:
Hi @octo, I've fixed the review comments, did the rebase and merged the conflicts. Could you please take a look?

Thanks,
Roman.

rkorynkx:
Hi @octo, just a reminder that review comments have been fixed, waiting for your feedback.

maryamtahhan:
Hi folks
would appreciate any additional feedback here

octo:
Hi, sorry for the long delay, I lost sight of this :(

The code looks good now. I've done some minor cleanups (re-ran clang-format; removed parenthesis around return values, see #2277). I'll push those and merge once the CI build is complete.

Thanks again for all your work! This is a really awesome addition to collectd!
—octo

Issue #2108, opened 2016-12-14
plugin_value_list_clone() in plugin.c logs that CollectInternalStats is broken 


charliesignalfx:
*   Version of collectd: 5.7
*   Operating system / distribution: CentOS 7 and Alpine Linux

## Expected behavior

When `CollectInternalStats` is `true`, collectd cycles through and does not log error messages indicating that a plugin is broken.

## Actual behavior
`plugin_value_list_clone` logs a message that there is a broken plugin.

```
[2016-12-14 21:53:16] Initialization complete, entering read-loop.
[2016-12-14 21:53:16] plugin_value_list_clone: Unable to determine interval from context for value list "d78ff1a9a5d5/collectd-write_queue/queue_length". This indicates a broken plugin. Please report this problem to the collectd mailing list or at <http://collectd.org/bugs/>.
[2016-12-14 21:53:16] plugin_value_list_clone: Unable to determine interval from context for value list "d78ff1a9a5d5/collectd-write_queue/derive-dropped". This indicates a broken plugin. Please report this problem to the collectd mailing list or at <http://collectd.org/bugs/>.
[2016-12-14 21:53:16] plugin_value_list_clone: Unable to determine interval from context for value list "d78ff1a9a5d5/collectd-cache/cache_size". This indicates a broken plugin. Please report this problem to the collectd mailing list or at <http://collectd.org/bugs/>.
[2016-12-14 21:53:26] plugin_value_list_clone: Unable to determine interval from context for value list "d78ff1a9a5d5/collectd-write_queue/queue_length". This indicates a broken plugin. Please report this problem to the collectd mailing list or at <http://collectd.org/bugs/>.
[2016-12-14 21:53:26] plugin_value_list_clone: Unable to determine interval from context for value list "d78ff1a9a5d5/collectd-write_queue/derive-dropped". This indicates a broken plugin. Please report this problem to the collectd mailing list or at <http://collectd.org/bugs/>.
[2016-12-14 21:53:26] plugin_value_list_clone: Unable to determine interval from context for value list "d78ff1a9a5d5/collectd-cache/cache_size". This indicates a broken plugin. Please report this problem to the collectd mailing list or at <http://collectd.org/bugs/>.
```
[...]

PR #2122, opened 2016-12-23
src/daemon/plugin.c: Register plugin_update_internal_statistics() as …

octo:
…a read callback.

The previous special-snowflake-solution had issues since it didn't set a plugin
context up.

Fixes: #2108

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -153,13 +153,11 @@ static const char *plugin_get_dir(void) {
     return (plugindir);
 }
 
-static void plugin_update_internal_statistics(void) { /* {{{ */
-
+static int plugin_update_internal_statistics() { /* {{{ */
   gauge_t copy_write_queue_length = (gauge_t)write_queue_length;
 
   /* Initialize `vl' */
   value_list_t vl = VALUE_LIST_INIT;
-  sstrncpy(vl.host, hostname_g, sizeof(vl.host));
   sstrncpy(vl.plugin, "collectd", sizeof(vl.plugin));
 
   /* Write queue */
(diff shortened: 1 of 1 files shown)


octo:
@charliesignalfx could you please check if this fixes the issue for you? Thanks!

src/daemon/plugin.c
```diff
@@ -153,13 +153,11 @@ static const char *plugin_get_dir(void) {
     return (plugindir);
 }
 
-static void plugin_update_internal_statistics(void) { /* {{{ */
-
+static int plugin_update_internal_statistics() { /* {{{ */
```
rubenk: Please keep the (void) in the argument list. While compilers still support K&R style, it has been deprecated since I think C89 ;)

rubenk:
One minor comment, otherwise LGTM.

charliesignalfx:
@octo Thank you for fixing this! This has resolved the issue I was seeing. I think this is "good to go" once @rubenk 's comment is addressed.  Sorry for the delay. (I was on vacation)

charliesignalfx:
@octo, I am not sure how I messed up in validating this, but I must have pulled down the wrong branch originally.  

This proposed fix _does not_ correct the issue I reported.  

To be sure, in a fresh ubuntu container I pulled down collectd:collectd-5.7 added the fix that you've proposed, cleaned, built, configured,  `make install all`ed and ran with the config I referenced earlier.  I still see the same broken plugin statements.

mfournier:
I added 2 commits, addressing the feedbacks on @octo's patch. Merging now, as jenkins seems to have properly digested the changes.

PR #2124, opened 2016-12-28
Add SFP signal power to types.db

tg90nor:
This adds a SFP optical power data type.

If it is more appropriate to make separate Data entries in my snmp conf. for rx and tx power, please tell me.

Example of how im using it with the snmp plugin:

    <Data "juniper_optical_power">
      Type "sfp_signal_power"
      Table true
      InstancePrefix "int-snmp-"
      Values "JUNIPER-DOM-MIB::jnxDomCurrentLaneRxLaserPower" "JUNIPER-DOM-MIB::jnxDomCurrentLaneTxLaserOutputPower"
    </Data>

### DIFF
--- a/src/types.db
+++ b/src/types.db
@@ -212,6 +212,7 @@ routes                  value:GAUGE:0:U
 satellites              value:GAUGE:0:U
 segments                value:GAUGE:0:65535
 serial_octets           rx:DERIVE:0:U, tx:DERIVE:0:U
+sfp_optical_power       rx:GAUGE:-100:100, tx:GAUGE:-100:100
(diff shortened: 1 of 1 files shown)


octo:
Hi @tg90nor,

thank you very much for opening this PR. We strongly recommend to use *data sets* (aka. *types*) with a single *data source* only – see [this FAQ entry](https://collectd.org/faq.shtml#faq:split_up_data_sets) for the rationale behind this. It would therefore be better if you would use the existing `signal_power` type and use "rx" and "tx" as type instances.

Best regards,
—octo

Issue #2123, opened 2016-12-25
mqtt plugin possible resource leak


nuclearcat:
*   Version of collectd: git latest
*   Operating system / distribution: ubuntu

## Expected behavior
Stable operation

## Actual behavior
After while:
Dec 24 06:25:36 monitor-server collectd[24549]: mqtt plugin: unable to reconnect to broker
Dec 24 06:25:36 monitor-server collectd[24549]: mqtt plugin: publish failed: Connection pending.
Dec 24 06:25:36 monitor-server collectd[24549]: mqtt_connect_broker: mosquitto_connect failed: Lookup error.
Dec 24 06:25:36 monitor-server collectd[24549]: mqtt plugin: unable to reconnect to broker
Dec 24 06:25:36 monitor-server collectd[24549]: mqtt plugin: publish failed: Connection pending.
Dec 24 06:25:36 monitor-server collectd[24549]: mqtt_connect_broker: mosquitto_connect failed: Lookup error.
Dec 24 06:25:36 monitor-server collectd[24549]: mqtt plugin: unable to reconnect to broker
Dec 24 06:25:36 monitor-server collectd[24549]: mqtt plugin: publish failed: Connection pending.
Dec 24 06:25:36 monitor-server collectd[24549]: mqtt_connect_broker: mosquitto_connect failed: Lookup error.

in lsof i can see a lot of:
collectd  24549 24567        root  779u     IPv4          491031328        0t0        TCP localhost:34144->localhost:1883 (CLOSE_WAIT)
collectd  24549 24567        root  780u     IPv4          491036255        0t0        TCP localhost:35194->localhost:1883 (CLOSE_WAIT)
collectd  24549 24567        root  781u     IPv4          491036602        0t0        TCP localhost:36204->localhost:1883 (CLOSE_WAIT)
collectd  24549 24567        root  782u     IPv4          491032276        0t0        TCP localhost:35162->localhost:1883 (CLOSE_WAIT)
collectd  24549 24567        root  783u     IPv4          491036251        0t0        TCP localhost:35184->localhost:1883 (CLOSE_WAIT)

This is clearly resource leak

Also there is possible other bug which trigger this leak (i will debug why it return this error), before leak happening i noticed this error messages:
Dec 23 11:44:02 monitor-server collectd[60403]: mqtt plugin: mosquitto_publish failed: Invalid function arguments provided.
Dec 23 11:44:02 monitor-server collectd[60403]: mqtt plugin: publish failed: Connection pending.
[...]

PR #2134, opened 2017-01-06
Bugfixes for MQTT resource leak and invalid symbols in topic

nuclearcat:
#2123 
Should be safe, since topic is null terminated


octo:
Thanks @nuclearcat! Committed to the *collectd-5.6* branch as 98bfe15ca31dc5deaf29b9fbcdfb5a4241ea677c and 6969c629641e681069d17c117829779191b8896d.

PR #2135, opened 2017-01-06
python: Extend Notification class

vmytnykx:
Extended python Notification class to support metadata in the same
way as it is implemented in Values class. The Values metadata class
implementation has been used as a example/source.

### DIFF
--- a/src/collectd-python.pod
+++ b/src/collectd-python.pod
@@ -475,7 +475,7 @@ Methods defined here:
 
 =over 4
 
-=item B<dispatch>([type][, values][, plugin_instance][, type_instance][, plugin][, host][, time][, interval]) -> None.  Dispatch a value list.
+=item B<dispatch>([type][, message][, plugin_instance][, type_instance][, plugin][, host][, time][, severity][, meta]) -> None.  Dispatch a notification.
 
 Dispatch this instance to the collectd process. The object has members for each
 of the possible arguments for this method. For a detailed explanation of these
@@ -500,6 +500,16 @@ generated.
 
 The severity of this notification. Assign or compare to I<NOTIF_FAILURE>,
 I<NOTIF_WARNING> or I<NOTIF_OKAY>.
+
+=item meta
+
+These are the meta data for the Notification object.
+It has to be a dictionary of numbers, strings or bools. All keys must be
+strings. I<int> and I<long> objects will be dispatched as signed integers unless
+they are between 2**63 and 2**64-1, which will result in a unsigned integer.
+One of these storage classes can be forced by using the classes
+B<collectd.Signed> and B<collectd.Unsigned>. A meta object received by a
+notification callback will always contain B<Signed> or B<Unsigned> objects.
 
 =back
 
--- a/src/python.c
+++ b/src/python.c
@@ -446,19 +446,21 @@ static int cpy_write_callback(const data_set_t *ds,
       } else if (type == MD_TYPE_SIGNED_INT) {
         if (meta_data_get_signed_int(meta, table[i], &si))
           continue;
-        temp = PyObject_CallFunctionObjArgs((void *)&SignedType,
-                                            PyLong_FromLongLong(si),
+        PyObject *sival = PyLong_FromLongLong(si); /* New reference */
+        temp = PyObject_CallFunctionObjArgs((void *)&SignedType, sival,
                                             (void *)0); /* New reference. */
         PyDict_SetItemString(dict, table[i], temp);
         Py_XDECREF(temp);
+        Py_XDECREF(sival);
       } else if (type == MD_TYPE_UNSIGNED_INT) {
         if (meta_data_get_unsigned_int(meta, table[i], &ui))
           continue;
-        temp = PyObject_CallFunctionObjArgs((void *)&UnsignedType,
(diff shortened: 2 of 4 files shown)


maryamtahhan:
Hi Folks
Any updates on this PR?

BR
Maryam

maryamtahhan:
@octo @rubenk is there anything we can do here to expedite this review?

Thanks in advance
Maryam

rubenk:
cc @trenkel

src/collectd-python.pod
```diff
@@ -475,7 +475,7 @@ Methods defined here:
 
 =over 4
 
-=item B<dispatch>([type][, values][, plugin_instance][, type_instance][, plugin][, host][, time][, interval]) -> None.  Dispatch a value list.
+=item B<dispatch>([type][, message][, plugin_instance][, type_instance][, plugin][, host][, time][, severity][, meta]) -> None.  Dispatch a notification.
```
octo: Thanks for noticing this! I've cherry-picked this fix into *collectd-5.7* as be906f960ede42f72317eab6aa6a5f116857a484.

src/python.c
```diff
@@ -510,6 +510,35 @@ static int cpy_notification_callback(const notification_t *notification,
...
+  for (notification_meta_t *meta = notification->meta;
+       meta != NULL; meta = meta->next) {
+    PyObject *temp = NULL;
+    if (meta->type == NM_TYPE_STRING) {
+      temp = cpy_string_to_unicode_or_bytes(meta->nm_value.nm_string); /* New reference. */
+      PyDict_SetItemString(dict, meta->name, temp);
+      Py_XDECREF(temp);
+    } else if (meta->type == NM_TYPE_SIGNED_INT) {
+      PyObject *sival = PyLong_FromLongLong(meta->nm_value.nm_signed_int);
+      temp = PyObject_CallFunctionObjArgs((void *)&SignedType, sival,
```
octo: Why can't you just do this:?

```c
PyDict_SetItemString(dict, meta->name,
                     PyLong_FromLongLong(meta->nm_value.nm_signed_int));
```

Dito for `NM_TYPE_UNSIGNED_INT` below.

src/python.c
```diff
@@ -510,6 +510,35 @@ static int cpy_notification_callback(const notification_t *notification,
...
+    } else if (meta->type == NM_TYPE_UNSIGNED_INT) {
+      PyObject *uval = PyLong_FromUnsignedLongLong(meta->nm_value.nm_unsigned_int);
+      temp = PyObject_CallFunctionObjArgs((void *)&UnsignedType, uval,
+                                          (void *)0); /* New reference. */
+      PyDict_SetItemString(dict, meta->name, temp);
+      Py_XDECREF(temp);
+    } else if (meta->type == NM_TYPE_DOUBLE) {
+      temp = PyFloat_FromDouble(meta->nm_value.nm_double); /* New reference. */
+      PyDict_SetItemString(dict, meta->name, temp);
+      Py_XDECREF(temp);
```
octo: Why do you call `Py_XDECREF()` on the return value of `PyFloat_FromDouble()` but not `PyLong_From[Unsigned]LongLong()`? According to the docs, all of these return a "new reference", don't they need to be treated equally?

src/pyvalues.c
```diff
@@ -42,6 +50,26 @@
     PyMem_Free(host);                                                          \
   } while (0)
 
+#define CPY_BUILD_META_FUNC(meta_type, func, val_type)                         \
```
octo: This is some serious macro magic. Please move the macro definition and invocation closer together so it's easier to understand what's going on.

src/pyvalues.c
```diff
@@ -541,9 +568,22 @@ static meta_data_t *cpy_build_meta(PyObject *meta) {
     Py_DECREF(key);
   }
   Py_XDECREF(l);
+  return 0;
+}
+
+CPY_BUILD_META_HANDLER(meta_data, meta_data_t);
+CPY_BUILD_META_HANDLER(plugin_notification_meta, notification_t);
+
+static meta_data_t *cpy_build_value_meta(PyObject *meta) {
```
octo: I think this pretty wrapper should get the pretty `cpy_build_meta()` name. The other function could be renamed to `cpy_build_meta_generic()` to underline that it works with more than one meta type.

src/python.c
vmytnykx: It's possible but collectd python plugin doesn't use `long` type directly. It uses collectd python types (`Signed` and `Unsigned`) instead. Thus, we have to convert python long/unsigned long type into `Signed`/`Unsigned` types. Same has been done for python value.
https://github.com/collectd/collectd/blob/master/src/collectd-python.pod

octo: That makes sense, thank you! I didn't realize right away that the `Signed` / `Unsigned` types are *not* built-in literals but classes.

src/python.c
vmytnykx: This is a good question! Based o my knowledge and testing it looks like a potential memory leak. Same logic is used when metadata is created for a value, so the problem needs to be fixed in both places in this cases.

vmytnykx:
Hi @octo, I've fixed all your comments. Thanks for your code review.

maryamtahhan:
Thanks Evenryone!

PR #2140, opened 2017-01-13
mcelog: Bug fixes and code style changes

pszczerx:
This series of patches introduces following changes:

- Increase delay between socket re-initialization from 1ms to 1000ms
- Check if location (plugin instance) was correctly parsed
- Dispatch notification on successful connection to mcelog server
- Free notification metadata
- Other minor changes

### DIFF
--- a/src/mcelog.c
+++ b/src/mcelog.c
@@ -106,19 +106,19 @@ static int mcelog_config(oconfig_item_t *ci) {
           0) {
         ERROR(MCELOG_PLUGIN ": Invalid configuration option: \"%s\".",
               child->key);
-        return -1;
+        return (-1);
       }
     } else if (strcasecmp("McelogLogfile", child->key) == 0) {
       if (cf_util_get_string_buffer(child, g_mcelog_config.logfile,
                                     sizeof(g_mcelog_config.logfile)) < 0) {
         ERROR(MCELOG_PLUGIN ": Invalid configuration option: \"%s\".",
               child->key);
-        return -1;
+        return (-1);
       }
     } else {
       ERROR(MCELOG_PLUGIN ": Invalid configuration option: \"%s\".",
             child->key);
-      return -1;
+      return (-1);
     }
   }
   return (0);
@@ -141,7 +141,7 @@ static int socket_close(socket_adapter_t *self) {
     }
   }
   pthread_rwlock_unlock(&self->lock);
-  return ret;
+  return (ret);
 }
 
 static int socket_write(socket_adapter_t *self, const char *msg,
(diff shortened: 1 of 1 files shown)


rubenk:
This seems to fix several things. Could you split it up into multiple commits, one commit per change?

pszczerx:
@rubenk wouldn't it be enough to just modify the commit title to better reflect the changes? It's quite a small commit and I'm not sure if it needs to be split up. I have already squashed the changes on my local branch, so splitting the commit would require more effort on my side. If you are fine with such approach, do you have a preferred title?

rubenk:
I wouldn't call a 150 line diff a small commit. I'm sorry, but squashing everything together makes reviewing very hard.

pszczerx:
@rubenk, as requested, commit has been split.

src/mcelog.c
```diff
@@ -210,95 +213,105 @@ static int socket_reinit(socket_adapter_t *self) {
...
 
-static int mcelog_submit(mcelog_memory_rec_t mr) {
+static int mcelog_submit(const mcelog_memory_rec_t *mr) {
 
-  value_list_t vl = VALUE_LIST_INIT;
-  vl.values_len = 1;
-  vl.time = cdtime();
+  if (!mr) {
+    ERROR(MCELOG_PLUGIN ": %s: NULL pointer", __FUNCTION__);
+    return -1;
```
rubenk: return (-1)

src/mcelog.c
```diff
@@ -210,95 +213,105 @@ static int socket_reinit(socket_adapter_t *self) {
...
-    ERROR("%s: add corrected errors period meta data failed", MCELOG_PLUGIN);
+  if (plugin_notification_meta_add_signed_int(n,
+                                              "uncorrected memory timed errors",
+                                              mr->uncorrected_err_timed) < 0) {
+    ERROR(MCELOG_PLUGIN ": add corrected timed errors meta data failed");
     plugin_notification_meta_free(n->meta);
     return (-1);
   }
+  if (mr->uncorrected_err_timed_period[0] != '\0')
+    if (plugin_notification_meta_add_string(n, "uncorrected errors time period",
```
rubenk: If you reverse the condition and return early, you can get rid of the nested indentation. Alternatively, combine the two conditions and && them into a single if.

src/mcelog.c
```diff
@@ -210,95 +213,105 @@ static int socket_reinit(socket_adapter_t *self) {
...
 
   ssnprintf(vl.type_instance, sizeof(vl.type_instance),
-            "uncorrected_memory_errors_in_%s", mr.uncorrected_err_timed_period);
-  vl.values = &(value_t){.derive = (derive_t)mr.uncorrected_err_timed};
+            "uncorrected_memory_errors_in_%s",
+            mr->uncorrected_err_timed_period);
+  vl.values = &(value_t){.derive = (derive_t)mr->uncorrected_err_timed};
   plugin_dispatch_values(&vl);
 
   return 0;
```
rubenk: unrelated to this PR, but I just now noticed this. This file mixes return n and return (n). Could you pick one style?

PR #2141, opened 2017-01-14
unixsock.so: undefined symbol: handle_getthreshold -> link unixsock against libcmds

piotr1212:
unixsock_la_LIBS is not being processed by automake and thus being incorrectly linked. This results in an symbol error when loading the module. Changing unixsock_la_LIBS to unixsock_la_LIBADD fixes this on my system (Fedora 25).

Error message:
> Jan 14 01:15:18 fedora collectd[20351]: dlopen ("/home/piotr/collectd/lib/collectd/unixsock.so") failed: /home/piotr/collectd/lib/collectd/unixsock.so: undefined symbol: handle_getthreshold. The most common cause for this problem is missing dependencies. Use ldd(1) to check the dependencies of the plugin / shared object.


rubenk:
Thanks!

PR #2150, opened 2017-01-24
This fixes #2149

### LINKED ISSUE #2149: write_prometheus plugin fails when IPv6 is disabled on OS
*   Version of collectd: 5.7.0.151
*   Operating system / distribution: CentOS Linux release 7.3.1611

## Expected behavior

write_prometheus plugin should bind to the network interface and serve requests

## Actual behavior

write_prometheus plugin fails with the following error:
```
collectd[11578]: write_prometheus plugin: MHD_start_daemon() failed.
collectd[11578]: Initialization of plugin `write_prometheus' failed with status -1. Plugin will be unloaded.
```

## Steps to reproduce

*   Install on RHEL/CentOS 7 Server with IPv6 disabled
*   Configure write_prometheus to load
*   Start collectd and examine logs

## Cause of behavior

_src/write_prometheus.c:759_
In an environment where IPv6 is disabled, the use of the ```MHD_USE_DUAL_STACK``` flag causes MHD_start_daemon() to fail.

msaraf:


src/write_prometheus.c
```diff
@@ -756,9 +764,9 @@ static int prom_init() {
 
   if (httpd == NULL) {
     unsigned int flags = MHD_USE_THREAD_PER_CONNECTION;
-#if MHD_VERSION >= 0x00093300
```
rpv-tomsk:
```
#if MHD_VERSION >= 0x00093300
    if (enable_ipv6) {
      flags |= MHD_USE_DUAL_STACK;
    }
#endif
```
MHD_USE_DUAL_STACK is available only if MHD_VERSION >= 0x00093300 ....

src/write_prometheus.c
```diff
@@ -735,6 +741,8 @@ static int prom_config(oconfig_item_t *ci) {
         httpd_port = (unsigned short)status;
     } else if (strcasecmp("StalenessDelta", child->key) == 0) {
       cf_util_get_cdtime(child, &staleness_delta);
+    } else if (strcasecmp("EnableIPv6", child->key) == 0) {
```
rpv-tomsk:
```
#if MHD_VERSION >= 0x00093300
cf_util_get_boolean(child, &enable_ipv6);
#else
       WARNING("write_prometheus plugin: Сonfiguration option EnableIPv6 is not supported for XXX versions < 9.33"
#endif
```

What if rework PR in this way?

src/write_prometheus.c
msaraf: @rpv-tomsk Thanks! I couldn't figure out why the CI build was failing.  Agreed on the way to rework the PR. Thanks again!

msaraf:
After looking at feature request #2173 (write_prometheus should support listening on specific interface) I've decided to close this pull request.  The implementation of that feature is a better way of fixing the issue and provides more flexibility.

PR #2155, opened 2017-01-28
notify_email: don't explicitly link against libssl and libcrypto

mfournier:
libesmtp takes care of this.

Pointed out by https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=852924


mfournier:
Actually, I realize this even applies to collectd-5.6.

mfournier:
Merged to collectd-5.6 as 33c0e7174

PR #2156, opened 2017-01-30
dpdkstat: Fix metrics' types

pszczerx:
This patch fixes parsing of metrics' types. Some statistics, such as
rx_q0errors, rx_q0bytes or rx_q0packet, don't contain suffix _errors, _bytes or
_packets. This has caused them to use generic type 'derive', rather than
more specific one, for instance 'if_rx_errors'. Additionally, some
statistics contain string 'error', rather than 'errors'.

Signed-off-by: Przemyslaw Szczerbik <przemyslawx.szczerbik@intel.com>

### DIFF
--- a/src/dpdkstat.c
+++ b/src/dpdkstat.c
@@ -314,17 +314,17 @@ static void dpdk_stats_resolve_cnt_type(char *cnt_type, size_t cnt_type_len,
   type_end = strrchr(cnt_name, '_');
 
   if ((type_end != NULL) && (strncmp(cnt_name, "rx_", strlen("rx_")) == 0)) {
-    if (strncmp(type_end, "_errors", strlen("_errors")) == 0) {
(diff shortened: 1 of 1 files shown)


rubenk:
Would you mind reopening this against the oldest branch that contains dpkdstat? You can pick the branch on github when you open a pull request. That would save us some time cherry-picking, since we then can just forward-merge the stable branch into master.

pszczerx:
@rubenk branch _collectd-5.7_ contains dpdkstat plugin, however I believe that PR #2095 wasn't cherry-picked there. It means that this commit won't apply cleanly on branch _collectd-5.7_.

rubenk:
Hmm yes, #2095 was a bit too large for a bugfix. I can pull this into master, but if you rather have this fix into the stable branch, can you do the backport?

pszczerx:
Opened #2167, which is based on collectd release 5.7

PR #2157, opened 2017-01-30
Plugin for getting DPDK ports link status and keep alive events.

krzysztofx:
Dpdkevents plugin collects and reports following events from DPDK based
applications:
- link status of network ports bound with DPDK
- keep alive events related to DPDK logical cores

In order to get link status plugin forks child process that attaches itself to
DPDK application as secondary DPDK process and fetches link status data using
DPDK API. This is the same approach like for dpdkstat plugin, also the same
utils_dpdk.c helper is utilized.

For getting keep alive events plugin communicates with DPDK application
via shared memory area. This is possible only if keep alive feature
is implemented in monitored application.
More details can be found under link
http://dpdk.org/doc/guides/sample_app_ug/keep_alive.html

Minimal required version of DPDK is 16.07.

Change-Id: Id820407ea933b1ecc31a72b6e344a6ec186ec780

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -770,7 +770,6 @@ if BUILD_PLUGIN_DPDKEVENTS
 pkglib_LTLIBRARIES += dpdkevents.la
 dpdkevents_la_SOURCES = src/dpdkevents.c src/utils_dpdk.c src/utils_dpdk.h
 dpdkevents_la_CPPFLAGS = $(AM_CPPFLAGS) $(LIBDPDK_CPPFLAGS)
-#dpdkevents_la_CFLAGS = $(AM_CFLAGS) $(BUILD_WITH_DPDK_CFLAGS)
 dpdkevents_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(LIBDPDK_LDFLAGS)
 dpdkevents_la_LIBADD = -ldpdk
 endif
--- a/Makefile.am
+++ b/Makefile.am
@@ -766,6 +766,15 @@ dns_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(BUILD_WITH_LIBPCAP_LDFLAGS)
 dns_la_LIBADD = $(BUILD_WITH_LIBPCAP_LIBS)
 endif
 
+if BUILD_PLUGIN_DPDKEVENTS
+pkglib_LTLIBRARIES += dpdkevents.la
+dpdkevents_la_SOURCES = src/dpdkevents.c src/utils_dpdk.c src/utils_dpdk.h
+dpdkevents_la_CPPFLAGS = $(AM_CPPFLAGS) $(LIBDPDK_CPPFLAGS)
+#dpdkevents_la_CFLAGS = $(AM_CFLAGS) $(BUILD_WITH_DPDK_CFLAGS)
+dpdkevents_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(LIBDPDK_LDFLAGS)
+dpdkevents_la_LIBADD = -ldpdk
+endif
+
 if BUILD_PLUGIN_DPDKSTAT
 pkglib_LTLIBRARIES += dpdkstat.la
 dpdkstat_la_SOURCES = src/dpdkstat.c src/utils_dpdk.c src/utils_dpdk.h
--- a/configure.ac
+++ b/configure.ac
@@ -2700,6 +2700,8 @@ AC_ARG_VAR([LIBDPDK_LDFLAGS], [Linker flags for libdpdk])
 
 AC_ARG_WITH([libdpdk], [AS_HELP_STRING([--without-libdpdk], [Disable libdpdk.])])
 
+rte_version_major=""
+rte_version_minor=""
 if test "x$with_libdpdk" != "xno"
 then
 	if test "x$LIBDPDK_CPPFLAGS" = "x"
@@ -2709,7 +2711,37 @@ then
 	SAVE_CPPFLAGS="$CPPFLAGS"
 	CPPFLAGS="$LIBDPDK_CPPFLAGS $CPPFLAGS"
 	AC_CHECK_HEADERS([rte_config.h],
-		[with_libdpdk="yes"],
+		[
+      with_libdpdk="yes"
+      AC_RUN_IFELSE(
+        [
+          AC_LANG_PROGRAM(
+            [[
+              #include <rte_version.h>
+            ]],
+            [[
+              return (RTE_VER_YEAR & 0xFF)
+            ]]
+          )
+        ],
+        [rte_version_major=0],
+        [rte_version_major=$?]
+      )
+      AC_RUN_IFELSE(
+        [
+          AC_LANG_PROGRAM(
+            [[
+              #include <rte_version.h>
+            ]],
+            [[
+              return (RTE_VER_MONTH & 0xFF)
+            ]]
+          )
+        ],
+        [rte_version_minor=0],
+        [rte_version_minor=$?]
+      )
+    ],
 		[with_libdpdk="no (rte_config.h not found)"]
 	)
 	CPPFLAGS="$SAVE_CPPFLAGS"
@@ -5881,6 +5913,7 @@ plugin_curl_xml="no"
 plugin_df="no"
 plugin_disk="no"
 plugin_drbd="no"
+plugin_dpdkevents="no"
 plugin_dpdkstat="no"
 plugin_entropy="no"
 plugin_ethstat="no"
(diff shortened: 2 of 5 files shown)


krzysztofx:
@octo @rubenk 
Please let me know if you'll find some time for reviewing proposed plugin.
Any comments appreciated.

Thanks
Krzysztof

Makefile.am
```diff
@@ -766,6 +766,15 @@ dns_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(BUILD_WITH_LIBPCAP_LDFLAGS)
 dns_la_LIBADD = $(BUILD_WITH_LIBPCAP_LIBS)
 endif
 
+if BUILD_PLUGIN_DPDKEVENTS
+pkglib_LTLIBRARIES += dpdkevents.la
+dpdkevents_la_SOURCES = src/dpdkevents.c src/utils_dpdk.c src/utils_dpdk.h
+dpdkevents_la_CPPFLAGS = $(AM_CPPFLAGS) $(LIBDPDK_CPPFLAGS)
+#dpdkevents_la_CFLAGS = $(AM_CFLAGS) $(BUILD_WITH_DPDK_CFLAGS)
```
rubenk: please remove this line.

configure.ac
```diff
@@ -2709,7 +2711,37 @@ then
 	SAVE_CPPFLAGS="$CPPFLAGS"
 	CPPFLAGS="$LIBDPDK_CPPFLAGS $CPPFLAGS"
 	AC_CHECK_HEADERS([rte_config.h],
-		[with_libdpdk="yes"],
+		[
+      with_libdpdk="yes"
+      AC_RUN_IFELSE(
```
rubenk: use AC_COMPILE_IFELSE here

configure.ac
```diff
@@ -2709,7 +2711,37 @@ then
...
+            ]],
+            [[
+              return (RTE_VER_YEAR & 0xFF)
+            ]]
+          )
+        ],
+        [rte_version_major=0],
+        [rte_version_major=$?]
+      )
+      AC_RUN_IFELSE(
```
rubenk: use AC_COMPILE_IFELSE here

configure.ac
```diff
@@ -2709,7 +2711,37 @@ then
 	SAVE_CPPFLAGS="$CPPFLAGS"
 	CPPFLAGS="$LIBDPDK_CPPFLAGS $CPPFLAGS"
 	AC_CHECK_HEADERS([rte_config.h],
-		[with_libdpdk="yes"],
+		[
+      with_libdpdk="yes"
```
rubenk: what's with the indentation?

configure.ac
```diff
@@ -6231,6 +6264,14 @@ fi
 
 if test "x$with_libdpdk" = "xyes"
 then
+  plugin_dpdkevents="no (DPDK version < 16.07)"
```
rubenk: I really dislike version checks. Is there no symbol in DPDK that you can check for existence?

src/dpdkevents.c
```diff
@@ -0,0 +1,592 @@
...
+ *   Krzysztof Matczak <krzysztofx@intel.com>
+ */
+
+#include "common.h"
+#include "plugin.h"
+#include "semaphore.h"
+#include "sys/mman.h"
+#include "utils_dpdk.h"
+#include "utils_time.h"
+#include "collectd.h"
```
rubenk: this should be the first include, separated by a blank line from the others.

src/dpdkevents.c
```diff
@@ -0,0 +1,592 @@
...
+
+static int dpdk_helper_link_status_get(dpdk_helper_ctx_t *phc) {
+  dpdk_events_ctx_t *ec = DPDK_EVENTS_CTX_GET(phc);
+
+  /* get Link Status values from DPDK */
+  uint8_t nb_ports = rte_eth_dev_count();
+  if (nb_ports == 0) {
+    DPDK_CHILD_LOG("dpdkevent-helper: No DPDK ports available. "
+                   "Check bound devices to DPDK driver.\n");
+    return -ENODEV;
```
rubenk: I'd just return -1, this is the only way this function can fail.

But looking further I see you never check the return value, so please fix that or make this return void.

src/dpdkevents.c
```diff
@@ -0,0 +1,592 @@
...
+                       link.link_status ? "UP" : "DOWN");
+      }
+    }
+  }
+
+  return 0;
+}
+
+/* this function is called from helper context */
+int dpdk_helper_command_handler(dpdk_helper_ctx_t *phc, enum DPDK_CMD cmd) {
```
rubenk: I don't see this handler being registered anywhere.

src/dpdkevents.c
```diff
@@ -0,0 +1,592 @@
...
+    char core_name[DATA_MAX_NAME_LEN];
+    ssnprintf(core_name, sizeof(core_name), "lcore%u", i);
+
+    if (!ec->config.keep_alive.send_updated ||
+        (ec->core_info[i].lcore_state !=
+         ec->config.keep_alive.shm->core_state[i])) {
+      ec->core_info[i].lcore_state = ec->config.keep_alive.shm->core_state[i];
+      ec->core_info[i].read_time = cdtime();
+
+#if RTE_VERSION >= RTE_VERSION_16_07
```
rubenk: ./configure already checks that you have dpdk >= 16.07, so this is unneeded.

src/dpdkevents.c
```diff
@@ -0,0 +1,592 @@
...
+        }
+        ec->link_info[i].status_updated = 0;
+      }
+    }
+  }
+
+  return 0;
+}
+
+static int dpdk_events_keep_alive_dispatch(dpdk_helper_ctx_t *phc) {
```
rubenk: you never check the return value and this function always return 0, so please make this void.

src/dpdkevents.c
```diff
@@ -0,0 +1,592 @@
...
+
+  return 0;
+}
+
+static int dpdk_events_read(user_data_t *ud) {
+  DPDK_EVENTS_TRACE();
+
+  if (g_hc == NULL) {
+    ERROR(DPDK_EVENTS_PLUGIN ": plugin not initialized.");
+    return -EINVAL;
```
rubenk: Just return -1.

src/dpdkevents.c
```diff
@@ -0,0 +1,592 @@
...
+            __FUNCTION__, ret);
+      return ret;
+    }
+  }
+  return 0;
+}
+
+static int dpdk_events_shutdown(void) {
+  DPDK_EVENTS_TRACE();
+  int ret = 0;
```
rubenk: no need to initialize ret here.

rubenk:
I left some comments inline, I'll see if I can do a more thorough review soon.

krzysztofx:
@rubenk 

Thanks for your comments. I'll address them as soon as possible.

configure.ac
krzysztofx: AC_COMPILE_IFELSE does only compilation, it doesn't run the AC_LANG_PROGRAM snippet. I need AC_RUN_IFELSE to extract RTE_VER_YEAR  and RTE_VER_MONTH values from rte_version.h  in runtime.

configure.ac
krzysztofx: see comment below

Makefile.am
krzysztofx: done

configure.ac
krzysztofx: I'll fix it

configure.ac
krzysztofx: There's no single symbol that would determine if  current DPDK instance supports keepalive feature, if that's what you mean. We have to check if version is >= 16.07 this way or other. It would be possible to simplify a bit this logic since DPDK header rte_version.h provides also compound macro  RTE_VERSION  with major and minor version numbers included, so I could apply it and related changes if you like it.

src/dpdkevents.c
krzysztofx: ok

configure.ac
rubenk: Ah yes. The thing is, these runtime tests are slow, so perhaps you could abuse the preprocessor and do a compile time test like:
```
#if RTE_VERSION < RTE_VERSION_NUM(16,7,0,0)
#error "version too old"
#endif
```

krzysztofx: You mean applying it along with AC_COMPILE_IFELSE, right ? Seems reasonable, I'll  do it.  Thanks

rubenk: Yes exactly. Thanks!

src/dpdkevents.c
krzysztofx: This is plugin specific implementation of handler interface declared in utils_dpdk.[ch]

rubenk: You're right, I missed that, sorry.

PR #2158, opened 2017-01-30
Java bindings - Adding Atomic* support

pmauduit:
AtomicIntegers are used to address some concurrency problems in Java, and this is advised by some metrics libraries like Netflix's servo:

https://github.com/Netflix/servo/wiki#jmx

Unfortunately, the GenericJMXConfValue.java class is not able to get the values from such types, hence this PR.

I scanned the Java doc for the Java package:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/package-summary.html
I think I have not forgotten any types.

Tests: tested at runtime

### DIFF
--- a/bindings/java/org/collectd/java/GenericJMXConfValue.java
+++ b/bindings/java/org/collectd/java/GenericJMXConfValue.java
@@ -30,6 +30,8 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Collection;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.Iterator;
 import java.util.ArrayList;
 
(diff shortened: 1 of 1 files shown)


rubenk:
Thank you @pmauduit. I know next to nothing of Java, so please bear with me. Have you tested these changes with older JRE's? I can imagine that not all JRE's provide these types.

pmauduit:
> Have you tested these changes with older JRE's? I can imagine that not all JRE's provide these types.

Relying on the official documentation, the `java.util.concurrent.atomic` package exists since 1.5, major distributions are currently shipping at least JRE >= 1.7, but to be honest, I did not try onto older runtimes than Java 1.7. Which JVM comes into your mind when dealing with collectd ?

rubenk:
Yeah, 1.5 comes to mind, although that's ancient by now. Seeing that this PR passes the build on all platforms we support, this should probably work fine.

PR #2165, opened 2017-02-08
intel_rdt: bugfix configuration

maryamtahhan:
Fix the configuration to stop collectd bombing out and deal with configuration
errors gracefully. Otherwise collectd will fail at the configuration phase
with: "Error: Reading the config file failed!".

Signed-off-by: Tahhan, Maryam <maryam.tahhan@intel.com>


rubenk:
Thank you @maryamtahhan!

PR #2168, opened 2017-02-09
virt: add support for new metrics

pszczerx:
Extend virt plugin functionality to support new statistics:

- Performance monitoring events
- Domain state & reason
- Percentage CPU usage
- CPU pinning (affinity)
- Disk errors
- File system information
- Job statistics

Leverage recently added 'ExtraStats' logic to allow user to enable or disable new metrics.
Plugin decodes domain state and reason enum values to human-readable string format, which will make it easier to interpret notifications.

Known limitation:
Missing CPU utilization metric is expected in some cases. In order to compute %CPU, we require 2 samples of CPU time. However, connection to libvirt is periodically refreshed, by default every 60 seconds. This interval can be changed with RefreshInterval virt plugin configuration option. Whenever connection is refreshed, we lose previous readings of CPU time, which means that we are unable to dispatch %CPU at next dispatch. With default settings - Plugin Interval 10 sec and RefreshInterval 60 sec CPU utilization will be missing every 6th dispatch. We are already working on solution for this issue, fix coming soon.

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -17,6 +17,7 @@
  *
  * Authors:
  *   Richard W.M. Jones <rjones@redhat.com>
+ *   Przemyslaw Szczerbik <przemyslawx.szczerbik@intel.com>
  **/
 
 #include "collectd.h"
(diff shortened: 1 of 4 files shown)


src/virt.c
```diff
@@ -527,6 +707,445 @@ static void lv_disconnect(void) {
...
+  sfree(vinfo);
+  return 0;
+}
+
+#ifdef DOMAIN_REASON_AVAILABLE
+static int get_domain_state(virDomainPtr domain) {
+  int domain_state = 0;
+  int domain_reason = 0;
+
+  int status = virDomainGetState(domain, &domain_state, &domain_reason, 0);
```
maryamtahhan: I believe a valid value for the domain_reason is NULL... we should check this before submitting this value
Please see https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainGetState

rubenk:
I'll hold off on this for a bit, since there's already work going on to handle these issues in the virt plugin. See the discussions in #2037 and #2048.

src/virt.c
pszczerx: _virDomainGetState_ API can be called with NULL as a domain_state, if one doesn't wish to retrieve domain reason i.e. `virDomainGetState(domain, &domain_state, NULL, 0); `.

In this patch  _domain_state_ variable is created on the stack, and then its address is passed to libvirt API. The value received from the API is an integer, thus I don't see any valid reason to make extra check.

pszczerx:
@rubenk, #2037, #2048 and recently merged changes #2175, #2103 do not contain all metrics we are interested in. Anyway, virt plugin code has changed since I opened this PR which means that I'll have to do a rebase.

rubenk:
Yes, and there are more changes coming, so to prevent duplicate work please wait a bit, or perhaps you can bundle your efforts with @fromanirh.

fromanirh:
Hi, just to notify that I don't have any more RFEs in the pipeline, we added all the metrics we needed.

pszczerx:
@fromanirh good to know, thanks for the information.

pszczerx:
@rubenk, PR has been rebased and slightly reworked (added support for recently added ExtraStats logic, domain state changed from metric to notification). I'm looking forward to hearing your comments.

rubenk:
Thanks again @pszczerx!

PR #2169, opened 2017-02-09
virt plugin: fix skipping inactive domains

ipavlikevich:
This patch fixes collecting disk/iface metrics from paused domains and prevents failures in virDomainBlockStats()/virDomainInterfaceStats() with timeout error:
```
libvirtd: Timed out during operation: cannot acquire state change lock
collectd: libvirt: QEMU Driver error : Timed out during operation: cannot acquire state change lock
```

Signed-off-by: Igor Pavlikevich <ipavlikevich@gmail.com>

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -824,11 +824,6 @@ static int lv_read(user_data_t *ud) {
       continue;
     }
 
-    if (info.di.state != VIR_DOMAIN_RUNNING) {
-      /* only gather stats for running domains */
-      continue;
-    }
-
     pcpu_submit(state->domains[i], &info);
     cpu_submit(info.di.cpuTime, state->domains[i], "virt_cpu_total");
     memory_submit((gauge_t)info.di.memory * 1024, state->domains[i]);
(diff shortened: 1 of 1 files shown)


rubenk:
Thank you. This looks like the correct fix, but I'll hold off merging for a bit since large parts of the virt plugin are being rewritten as we speak. See #2037 for the discussion.

rubenk:
Thanks. Can you please rebase?

ipavlikevich:
Rebased

src/virt.c
```diff
@@ -1143,6 +1140,18 @@ static int refresh_lists(struct lv_read_instance *inst) {
         goto cont;
       }
 
+      status = virDomainGetInfo(dom, &info);
+      if (status != 0) {
+        ERROR(PLUGIN_NAME " plugin: virDomainGetInfo failed with status %i.",
```
rubenk: Should this really be logged as an error? I can imagine it happens quite often that a domain is stopped or disappears between the lookup above and this call. Perhaps a warning is more appropriate?

ipavlikevich: There are two calls directly above, virDomainLookupByID() and virDomainGetName(), so this exception must be rather rare situation.

rubenk: Ok, we'll see how it goes. If it shows up too much we can always downgrade it.

PR #2170, opened 2017-02-09
Fix rpm spec file for 5.7.1

javicacheiro:
Update version in redhat rpm spec file to fix error building rpm package for 5.7.1 version.


javicacheiro:
Done!

mfournier:
Thanks folks !

PR #2175, opened 2017-02-14
virt plugin: Add physical cpu reporting

fromanirh:
Add a metric to report physical cpu consumed by the hypervisor,
split per user/system time. This is provided by libvirt since version
0.9.11, and used by oVirt.

We also extend the newly-added 'ExtraStats' option to allow
users to toggle on/off the new metric. Default is off.

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -484,7 +484,7 @@ static void disk_submit(struct lv_block_info *binfo, virDomainPtr dom,
   }
 }
 
-static unsigned int parse_ex_stats_flags(const char *exstats, int numexstats) {
+static unsigned int parse_ex_stats_flags(char **exstats, int numexstats) {
   int extra_stats = ex_stats_none;
   for (int i = 0; i < numexstats; i++) {
     for (int j = 0; ex_stats_table[j].name != NULL; j++) {
--- a/src/virt.c
+++ b/src/virt.c
@@ -485,18 +485,18 @@ static void disk_submit(struct lv_block_info *binfo, virDomainPtr dom,
 }
 
 static unsigned int parse_ex_stats_flags(char **exstats, int numexstats) {
-  int extra_stats = ex_stats_none;
+  unsigned int ex_stats_flags = ex_stats_none;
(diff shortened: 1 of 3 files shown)


rubenk:
```
src/virt.c: In function 'parse_ex_stats_flags':
src/virt.c:491:22: error: passing argument 1 of 'strcasecmp' makes pointer from integer without a cast [-Werror]
       if (strcasecmp(exstats[i], ex_stats_table[j].name) == 0) {
                      ^
In file included from ./src/daemon/collectd.h:45:0,
                 from src/virt.c:22:
/usr/include/string.h:534:12: note: expected 'const char *' but argument is of type 'char'
 extern int strcasecmp (const char *__s1, const char *__s2)
            ^
src/virt.c: In function 'lv_config':
src/virt.c:702:42: error: passing argument 1 of 'parse_ex_stats_flags' from incompatible pointer type [-Werror]
       extra_stats = parse_ex_stats_flags(exstats, numexstats);
                                          ^
src/virt.c:487:21: note: expected 'const char *' but argument is of type 'char **'
 static unsigned int parse_ex_stats_flags(const char *exstats, int numexstats) {
                     ^
Makefile:6081: recipe for target 'src/virt_la-virt.lo' failed
make[1]: *** [src/virt_la-virt.lo] Error 1
```

src/virt.c
```diff
@@ -409,6 +484,21 @@ static void disk_submit(struct lv_block_info *binfo, virDomainPtr dom,
   }
 }
 
+static unsigned int parse_ex_stats_flags(char **exstats, int numexstats) {
+  int extra_stats = ex_stats_none;
```
rubenk: this shadows the global extra_stats, which is an unsigned int.

rubenk:
Thanks again!

rubenk:
@fromanirh so, what's next? ;)

fromanirh:
Hi @rubenk! Next in lines are just fixes, we believe we filled all the gaps we had, so from now on it's polishing. I'll be in touch!
Thanks for your help!

PR #2176, opened 2017-02-14
dpdkstat: fix fixed shared mem object name

maryamtahhan:
Fix the shared memory object name so that it's configurable. This way multiple
DPDK applications running in containers can be monitored at the same time
(assuming collectd is also running in that container).

Signed-off-by: Maryam Tahhan <maryam.tahhan@intel.com>

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -532,6 +532,7 @@
 #    ProcessType "secondary"
 #    FilePrefix "rte"
 #  </EAL>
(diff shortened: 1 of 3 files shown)


rubenk:
If this applies to the 5.7 branch, please reopen against 5.7.

maryamtahhan:
Hi Ruben
this applies against the update to dpdkstat on master that uses utils rather than 5.7.

BR
Maryam

rubenk:
Ok, thanks Maryam!

PR #2182, opened 2017-02-21
fix for memory leak in utils_match.c

krzysztofx:
Added  releasing of regex resources.

### DIFF
--- a/src/utils_match.c
+++ b/src/utils_match.c
@@ -29,6 +29,7 @@
 #include <regex.h>
 
 #define UTILS_MATCH_FLAGS_FREE_USER_DATA 0x01
+#define UTILS_MATCH_FLAGS_REGEX 0x04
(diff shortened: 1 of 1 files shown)


src/utils_match.c
```diff
@@ -29,6 +29,7 @@
 #include <regex.h>
 
 #define UTILS_MATCH_FLAGS_FREE_USER_DATA 0x01
+#define UTILS_MATCH_FLAGS_REGEX 0x04
```
rubenk: please swap this with the line below, so the flags sort in numerical order.

rubenk:
Thanks @krzysztofx. You couldn't have known, but we actually don't support 4.10 anymore. I'd swear we had this documented somewhere, but 5.6 is the oldest branch we support.

krzysztofx:
created PR https://github.com/collectd/collectd/pull/2192 for collectd 5.6

PR #2189, opened 2017-02-23
ntpd: Correct loop offset and error calculation

NeilW:
Stop the scaling calculation dropping the sign of the underlying value

Fixes: 2188


mfournier:
Thanks @NeilW !

I also noticed this problem, and confirm your patch fixes it. I cherry-picked it to the oldest release branch as c0760b142.

Thanks a lot !

Issue #2201, opened 2017-03-03
Build warning in ovs_stats with gcc 7


rubenk:
```
src/ovs_stats.c: In function ‘ovs_stats_plugin_read’:
src/ovs_stats.c:898:46: warning: ‘snprintf’ output may be truncated before the last format character [-Wformat-truncation=]
           snprintf(devname, sizeof(devname), "%s.%s", bridge->name, port->name);
                                              ^~~~~~~
src/ovs_stats.c:898:11: note: ‘snprintf’ output 2 or more bytes (assuming 256) into a destination of size 255
           snprintf(devname, sizeof(devname), "%s.%s", bridge->name, port->name);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

rubenk:
@maryamtahhan would you be so kind to have a look?

maryamtahhan:
Hi @rubenk ,
@taraschornyi  is going to have a look at this.

BR
Maryam

rubenk:
Fixed by 37b665223a5e8d25cf9a9411f8266106be9024e4.

PR #2207, opened 2017-03-06
ovs_stats: Increase devname buffer

taraschornyi:
Increase devname buffer to ensure that snprintf output would not be truncated before the last format character
Fix for #2201 


rubenk:
Thanks Taras!

PR #2225, opened 2017-03-22
dpdkstat: Enable all ports by default.

taraschornyi:
If "EnabledPortMask" option is not set
all DPDK ports will be monitored.


taraschornyi:
Looks like build failed because of  exceed of disc quota on build host

> Assembler messages:
Fatal error: can't create src/.libs/threshold.o: Disc quota exceeded
gmake[1]: *** [Makefile:5531: src/threshold.lo] Error 1

Issue #1990, opened 2016-10-12
 processes plugin: wrong disk I/O metric types/values


rpv-tomsk:
Collectd reports `/proc/PID/io` fields `io_rchar/io_wchar` as `ps_disk_octets` type and `io_syscr/io_syscw` as `ps_disk_ops` type.
The names of these types do not match values they represent. 

1) For `io_rchar/io_wchar` data type should be named as `ps_io_octets`. In Linux, these values represent any I/O of process, not disk I/O only: they include sockets, pipes, console I/O, etc.
The `ps_io_octets` type can be added in addition to the existing `ps_disk_octets` for which values can be obtained from `read_bytes/write_bytes` of /proc/PID/io. These values will be much more correct for reporting as `ps_disk_octets` than existing `io_rchar/io_wchar`. Please note what `read_bytes/write_bytes` is _real_ disk I/O, it does not account file I/O from cache.

2) For `io_syscr/io_syscw` correct type should be named `ps_io_ops`.
Reported values do not match any real disk ops. 

My proposal is to fix these issues.

Related documentation: http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/proc.txt?id=HEAD#n1558

rpv-tomsk:
Is these changes will be good to go to 5.7?

rpv-tomsk:
Metric `ps_disk_octets` based on `read_bytes/write_bytes` fully match disk IO metrics:

![disk-traffic](https://cloud.githubusercontent.com/assets/13328211/19716610/f2dbe7a8-9b86-11e6-9499-750b08ce20a3.png)
![ps-disk-octets](https://cloud.githubusercontent.com/assets/13328211/19716611/f2dc44c8-9b86-11e6-97a4-0a0c017129d8.png)

octo:
+1 for reading better metrics from `/proc/<PID>/io`. Changing the types is however a backwards incompatible change and therefore only possible in a new major release, which is not currently planned.

rpv-tomsk:
Hi! 

Why title was changed? Processes plugin already reads I/O metrics from /proc/<PID>/io, so, the title is wrong now: The plugin should not _start reading_ but it should report these existing readings in _correct_ types.

rpv-tomsk:
Plugin declares `ps_disk_io` metric.  
In a fact, values reported in this metric is not a disk IO and this may be considered as a Collectd bug.
[...]

Issue #1225, opened 2015-08-16
Processes plugin: Provide actual disk I/O per process


PhilippWendler:
Under Linux, the processes plugin reads the values `rchar` and `wchar` from `/proc/<pid>/io` and exposes them as `ps_disk_octets`. These values include I/O that was satisfied by the page cache and even tty I/O.
There are also the values `read_bytes` and `write_bytes` which are an estimate of how much bytes were actually read from/written to the disk. For me, these values would be more interesting, thus I would like to propose that the processes plugin also exposes them.

Documentation: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/proc.txt?id=v4.1#n1486

bradjorgensen:
I'm using 5.4 and I've made several modifications to the source I use on my servers including what you mention here.  I'm planning on submitting most of them in git sometime hopefully before the end of the year.

One thing that will need to be decided, though, is the naming and types.  As you mention, the current metrics use the type `ps_disk_octets`, but it doesn't really represent disk traffic.  Because of that, I'm not sure if I should create a new type as I did in my local code or use a type instance.

rpv-tomsk:
PR #2232 is about the same.

rpv-tomsk:
Closed in #2232.

PR #2232, opened 2017-03-29
processes: Show real disk IO in addition to process IO (Linux only)

rpv-tomsk:
Collectd reports `/proc/PID/io` fields `io_rchar/io_wchar` as `ps_disk_octets` type and `io_syscr/io_syscw` as `ps_disk_ops` type.
The names of these types do not match values they represent. 

1) For `io_rchar/io_wchar` data type should be named as `ps_io_octets`. In Linux, these values represent any I/O of process, not disk I/O only: they include sockets, pipes, console I/O, etc.
The `ps_io_octets` type can be added in addition to the existing `ps_disk_octets` for which values can be obtained from `read_bytes/write_bytes` of /proc/PID/io. These values will be much more correct for reporting as `ps_disk_octets` than existing `io_rchar/io_wchar`. Please note what `read_bytes/write_bytes` is _real_ disk I/O, it does not account file I/O from cache.

2) For `io_syscr/io_syscw` correct type should be named `ps_io_ops`.
Reported values do not match any real disk ops. 

My proposal is to fix these issues.

New, correct mapping implemented:
```
 io_rchar/io_wchar      -> ps_io_octets   (renamed from ps_disk_octets)
 io_syscr/io_syscw      -> ps_io_ops      (renamed from ps_disk_ops)
 read_bytes/write_bytes -> ps_disk_octets (new data collected to existing type)
```

Related documentation: http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/proc.txt?id=HEAD#n1558

Issues: #1990, #1225

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -185,6 +185,8 @@ typedef struct process_entry_s {
   derive_t io_wchar;
   derive_t io_syscr;
   derive_t io_syscw;
+  derive_t io_diskr;
+  derive_t io_diskw;
   _Bool has_io;
 
   derive_t cswitch_vol;
@@ -209,6 +211,8 @@ typedef struct procstat_entry_s {
   derive_t io_wchar;
   derive_t io_syscr;
   derive_t io_syscw;
+  derive_t io_diskr;
+  derive_t io_diskw;
 
   derive_t cswitch_vol;
   derive_t cswitch_invol;
@@ -242,6 +246,8 @@ typedef struct procstat {
   derive_t io_wchar;
   derive_t io_syscr;
   derive_t io_syscw;
+  derive_t io_diskr;
+  derive_t io_diskw;
 
   derive_t cswitch_vol;
   derive_t cswitch_invol;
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
@Mogztter, you might be interested.

rpv-tomsk:
Related to: #1225

Mogztter:
This looks good to me 👍 
Thanks for the link on the documentation.

The only "breaking change" is that `ps_disk_ops` will not be reported anymore but I think that's fine because the reported value was incorrect (arguably a bug fix). 
If we really want backward compatibility we could continue to report this value (even if the value don't match the label).

PhilippWendler:
If the data is different, shouldn't a new name be used anyway, to make sure that it won't be confused with old data? Then a new type can be used, too.

rpv-tomsk:
Types renamed.

```
New mapping implemented:

 io_rchar/io_wchar      -> io_octets   (renamed from ps_disk_octets)
 io_syscr/io_syscw      -> io_ops      (renamed from ps_disk_ops)
 read_bytes/write_bytes -> disk_octets (new data collected)

Obsolete datatypes, not used anymore:

 ps_disk_octets
 ps_disk_ops
```

Issue #492, opened 2013-11-26
write_mongo doesn't work with ./configure


dothak:
Using mongo-c-driver v0.9
./configure output: http://pastie.org/8509965
make output: http://pastie.org/8509950

jbergstroem:
I think it's time to move to mongo-c-driver. We should probably switch to checking through pkg-config as well.

ChrisLundquist:
From the [current Readme](https://github.com/mongodb/mongo-c-driver/tree/221708f15c4ab12a881f1a3bd597c8614534f5d5)

> There are absolutely no guarantees of API/ABI stability at this point. But generally, we won't break API/ABI unless we have good reason.

> mongo-c-driver depends on Libbson. Libbson will automatically be built if you do not have it installed on your system.

Means that even though they are > 1.0 they might still break the API (presumably they mention this because they will at a less than major release version)

This is a little worrying.

Looking at the make output @dothak posted it seems they (?re)moved `bson_create` and `bson_dispose`

I took a quick look right now, and it seems at against mongo-c-driver 1.1.7 there are a few things that need to change. Currently, we can't even build write_mongo using the latest driver.

- [ ] Fix configure.ac to look for `<mongoc.h>` instead of `<mongo.h>`
- [ ] Add libbson to configure.ac

Once we have it building again, then we can go about updating things.

jbergstroem:
@ChrisLundquist do we really have to check for libbson? It's implied if using mongo-c-driver >1.0. If we use pkg-config to check for libmongo-c-1.0 we'd get proper linking/include details as well.

ChrisLundquist:
I agree it is implied. Shouldn't we let someone specify a libbson path though( if we can't figure it out)?

Currently the only way I can get a test program to include those headers looks like this:
`gcc -c -g -O2 -I/usr/local/include/libmongoc-1.0 -I/usr/local/include/libbson-1.0 ./test.c `

`mongoc.h` has `#include <bson.h>` in it, so we have to make that work.

Note that currently `configure.ac` basically checks for `$prefix/include/mongo.h` and `$prefix/lib/libmongo` which no longer works because they are in the `libmongoc-1.0` folder.
I'm not sure if this is just the case on OS X, or if they are installed this way on RHEL / Debian as well.

@jbergstroem It sounds like you might know more about M4 macros / autotools / etc than I do. If you can fix that part up, I can take a look at the C part.

jbergstroem:
[...]

PR #2236, opened 2017-04-06
Update mongodb write code to use latest API (Fixes: #492)

saiarcot895:
Also state that pkg-config is no longer optional.

Fixes: #492


saiarcot895:
Note that there is also a change to how the data is stored in MongoDB. Previously, each entry had its own row (or document, as it's called in MongoDB). Now, there is one document for each unique combination of (hour, plugin, plugin instance, type, type instance), and data for each minute and second (as given) is stored in subdocuments. This results in less redundant data being stored and better usage of the documents. This format of storing data is outlined in [this](https://www.mongodb.com/blog/post/schema-design-for-time-series-data-in-mongodb) blog post.

octo:
Hi @saiarcot895,

thank you very much for your patch! Could you please split your PR into two parts, one fixing the build and adapt to the new API and the second PR to change the way metrics are stored? The former is a relatively straight-forward fix, the latter needs some discussion since it does not sound to be backwards compatible.

Thanks and best regards,
—octo

saiarcot895:
Will do, give me a day or two.

saiarcot895:
Done. Also, it should be noted that the automated build do not appear to be building the write_mongodb plugin; I'm guessing the library isn't installed on those systems.

configure.ac
```diff
@@ -3321,6 +3321,15 @@ AC_ARG_WITH([libmongoc],
 SAVE_CPPFLAGS="$CPPFLAGS"
 SAVE_LDFLAGS="$LDFLAGS"
 
+if test "x$with_libmongoc" = "xyes"
```
rubenk: Please use the same style as the rest of configure.ac, so `if; then` on the same line, and indent with 2 spaces.

configure.ac
```diff
@@ -3321,6 +3321,15 @@ AC_ARG_WITH([libmongoc],
...
+then
+	if test "x$PKG_CONFIG" != "x"
```
rubenk: This check is unneeded, since pkg_config is no longer optional.

configure.ac
```diff
@@ -3321,6 +3321,15 @@ AC_ARG_WITH([libmongoc],
...
+	then
+		LIBMONGOC_CPPFLAGS="$LIBMONGOC_CPPFLAGS `$PKG_CONFIG --cflags libmongoc-1.0`"
```
rubenk: Please use the PKG_CHECK_MODULES macro.

rubenk:
You're right, libmongoc isn't installed on all the builders. I'll try get it installed but we'll need to build new vms so it might take a while. I don't think that should block merging this PR however.

saiarcot895:
Updated. The flag assignments might need to be tweaked if a custom prefix is passed in, but on a standard system, it is correctly detected.

rubenk:
Thank you. I think the custom prefix handling will conflict with the pkg-config checks if you use both. The .pc file will point to the system-installed libmongoc and the system include path. If you have a version of libmongoc installed in a custom location, but none installed in the system location, the pkg-config check will fail. If you have both, there's a great chance you end up with the wrong LDFLAGS.

It might be best to remove the support for a custom prefix, people who need to link with libmongoc in a custom location can always override LIBMONGOC_CFLAGs and LIBMONGOC_LDFLAGS on the command line.

I realize that there are more checks in configure.ac that do the same, my plan is to clean those up at some point.

rubenk:
The commit that adds the mongodb libs to the builders is at https://github.com/collectd/collectd-ci/commit/12eb07e8ca09ec81c6eab5d35d37ea788bd0af0a

saiarcot895:
Done, and I also fixed a mutex unlock issue in case there was an error in initializing a connection.

configure.ac
```diff
@@ -3310,48 +3310,58 @@ AC_ARG_WITH([libmongoc],
...
 
-CPPFLAGS="$SAVE_CPPFLAGS"
-LDFLAGS="$SAVE_LDFLAGS"
+  CPPFLAGS="$SAVE_CPPFLAGS"
+  LDFLAGS="$SAVE_LDFLAGS"
+fi
 
 if test "x$with_libmongoc" = "xyes"; then
-  BUILD_WITH_LIBMONGOC_CPPFLAGS="$LIBMONGOC_CPPFLAGS"
+  BUILD_WITH_LIBMONGOC_CPPFLAGS="$LIBMONGOC_CFLAGS"
```
rubenk: One last nitpick. Now that pkg-config sets LIBMONGOC_CFLAGS, this may contain arguments that are not acceptable as CPPFLAGS but are as CFLAGS. Please change the variable here and in Makefile.am to BUILD_WITH_LIBMONGOC_CFLAGS.

saiarcot895:
Done.

Makefile.am
```diff
@@ -1758,7 +1758,7 @@ endif
 if BUILD_PLUGIN_WRITE_MONGODB
 pkglib_LTLIBRARIES += write_mongodb.la
 write_mongodb_la_SOURCES = src/write_mongodb.c
-write_mongodb_la_CPPFLAGS = $(AM_CPPFLAGS) $(BUILD_WITH_LIBMONGOC_CPPFLAGS)
+write_mongodb_la_CFLAGS = $(AM_CPPFLAGS) $(BUILD_WITH_LIBMONGOC_CFLAGS)
```
rubenk: This needs to be AM_CFLAGS.

saiarcot895:
Changed.

src/write_mongodb.c
```diff
@@ -55,179 +51,239 @@ struct wm_node_s {
   char *passwd;
 
   _Bool store_rates;
+  bool connected;
```
octo: Please use the `_Bool` type.

src/write_mongodb.c
```diff
@@ -55,179 +51,239 @@ struct wm_node_s {
...
 
-  bson_init(ret); /* matched by bson_destroy() */
-  bson_append_date(ret, "time", (bson_date_t)CDTIME_T_TO_MS(vl->time));
-  bson_append_string(ret, "host", vl->host);
-  bson_append_string(ret, "plugin", vl->plugin);
-  bson_append_string(ret, "plugin_instance", vl->plugin_instance);
-  bson_append_string(ret, "type", vl->type);
-  bson_append_string(ret, "type_instance", vl->type_instance);
+  BSON_APPEND_DATE_TIME(ret, "timestamp", CDTIME_T_TO_MS(vl->time));
+  BSON_APPEND_UTF8(ret, "host", vl->host);
```
octo: note that collectd does not ensure that these are valid UTF-8. Is there another, more generic macro? or could you add some error checking?

src/write_mongodb.c
```diff
@@ -55,179 +51,239 @@ struct wm_node_s {
...
-                    const value_list_t *vl, user_data_t *ud) {
-  wm_node_t *node = ud->data;
-  char collection_name[512];
-  bson *bson_record;
-  int status;
+static int wm_initialize(wm_node_t *node) /* {{{ */
+{
+  char *uri;
+  int uri_length;
+  char *format_string;
```
octo: some static code analyzers may freak out when a `char*` is used as a format string. can you either use a define or a `char const *`, please.

src/write_mongodb.c
```diff
@@ -55,179 +51,239 @@ struct wm_node_s {
...
+            node->user);
+      node->connected = 0;
+      free(uri);
+      uri = NULL;
+      return -1;
+    }
+  } else {
+    format_string = "mongodb://%s:%d";
+    uri_length = strlen(format_string) + strlen(node->host) + 5 + 1;
+    if ((uri = malloc(sizeof(char) * uri_length)) == NULL) {
```
octo: please use `calloc` instead

src/write_mongodb.c
```diff
@@ -55,179 +51,239 @@ struct wm_node_s {
...
+  INFO("write_mongodb plugin: Connecting to [%s]:%i",
+       (node->host != NULL) ? node->host : "localhost",
+       (node->port != 0) ? node->port : MONGOC_DEFAULT_PORT);
+
+  if ((node->db != NULL) && (node->user != NULL) && (node->passwd != NULL)) {
+    format_string = "mongodb://%s:%s@%s:%d/?authSource=%s";
+    uri_length = strlen(format_string) + strlen(node->user) +
+                 strlen(node->passwd) + strlen(node->host) + 5 +
+                 strlen(node->db) + 1;
+    if ((uri = malloc(sizeof(char) * uri_length)) == NULL) {
```
octo: please use `calloc` instead

src/write_mongodb.c
```diff
@@ -55,179 +51,239 @@ struct wm_node_s {
...
       ERROR("write_mongodb plugin: Connecting to [%s]:%i failed.",
             (node->host != NULL) ? node->host : "localhost",
-            (node->port != 0) ? node->port : MONGO_DEFAULT_PORT);
-      mongo_destroy(node->conn);
-      pthread_mutex_unlock(&node->lock);
-      return (-1);
+            (node->port != 0) ? node->port : MONGOC_DEFAULT_PORT);
+      node->connected = 0;
+      free(uri);
+      uri = NULL;
```
octo: use the `sfree()` macro, which will set the pointer to `NULL` for you

src/write_mongodb.c
saiarcot895: I added a validation check at the end of the method that makes sure all strings are valid UTF-8 strings.

src/write_mongodb.c
```diff
@@ -162,7 +171,7 @@ static int wm_initialize(wm_node_t *node) /* {{{ */
     uri_length = strlen(format_string) + strlen(node->user) +
                  strlen(node->passwd) + strlen(node->host) + 5 +
                  strlen(node->db) + 1;
-    if ((uri = malloc(sizeof(char) * uri_length)) == NULL) {
+    if ((uri = calloc(uri_length, sizeof(char))) == NULL) {
```
rubenk: here and below: sizeof(char) is always 1 as defined by the C standard. uri_length should be the 2nd argument, so `calloc(1, uri_length)`.

src/write_mongodb.c
```diff
@@ -140,14 +140,23 @@ static bson_t *wm_create_bson(const data_set_t *ds, /* {{{ */
...
+    return NULL;
+  }
+
   return ret;
 } /* }}} bson *wm_create_bson */
 
 static int wm_initialize(wm_node_t *node) /* {{{ */
 {
   char *uri;
   int uri_length;
```
rubenk: make this size_t please.

src/write_mongodb.c
saiarcot895: I can change it to 1 (I wanted to use `sizeof(char)` just to be explicit), but on my man page, the number of members/elements is the first argument, and the second argument is the size of each element.

rubenk: `calloc(1, uri_length)` reads as "allocate 1 element of size uri_length for the url", whereas `calloc(uri_length, 1)` reads as "allocate uri_length elements of size 1". It's hard to explain the difference, and in practice it doesn't matter, but we use the first form throughout the codebase.

saiarcot895:
Done.

rubenk:
Thanks for your contribution @saiarcot895!

PR #2241, opened 2017-04-11
Tree-wide: remove last remnants of sizeof(char)

rubenk:
sizeof(char) is 1 by definition.

### DIFF
--- a/src/write_sensu.c
+++ b/src/write_sensu.c
@@ -132,7 +132,7 @@ static int add_str_to_list(struct str_list *strs,
     ERROR("write_sensu plugin: Unable to alloc memory");
     return -1;
   }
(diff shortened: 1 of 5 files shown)


src/write_sensu.c
```diff
@@ -231,7 +231,7 @@ static char *build_json_str_list(const char *tag,
   char *ret_str = NULL;
   char *temp_str;
   if (list->nb_strs == 0) {
-    ret_str = malloc(sizeof(char));
+    ret_str = malloc(1);
```
octo: So many problems in that function … I'll send a separate PR for this one.

rubenk: heh yes, it has been bothering me too for a while now ;)

rubenk:
Thanks @octo!

PR #2242, opened 2017-04-11
Bugfix/issue 2239 java target match config

bvarner:
* Moves JVM initialization to the `cjni_config_callback`.
    * Appends things to the config_block, or initializes on first invocation.
    * Shuts down any previously initialized JVM (if more than one invocation occurs).
    * Starts a new JVM with the full contents of the config_block.
* Adds a few ignore patterns (IntelliJ IDEA, and some stuff from building on Mac OS X)

### DIFF
--- a/src/java.c
+++ b/src/java.c
@@ -95,6 +95,8 @@ static pthread_mutex_t       java_callbacks_lock = PTHREAD_MUTEX_INITIALIZER;
 
 static oconfig_item_t       *config_block = NULL;
 
+static int cjni_shutdown (void);
+
 /*
  * Prototypes
  *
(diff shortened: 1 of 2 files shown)


rubenk:
Can you please rebase this against the collectd-5.6 branch? 5.6 is the oldest version we currently support.

A-Garg123:
sir can you please explain briefly what is the bug so that i can go in the right direction to solve it

src/java.c
```diff
@@ -2400,32 +2404,55 @@ static int cjni_config_callback (oconfig_item_t *ci) /* {{{ */
 
   if (config_block == NULL)
```
A-Garg123, line 2405: if(config_block == NULL)

src/java.c
```diff
@@ -2400,32 +2404,55 @@ static int cjni_config_callback (oconfig_item_t *ci) /* {{{ */
...
   {
+    DEBUG("java plugin: Setting initial config block");
+
     config_block = ci_copy;
-    return (0);
```
A-Garg123, line 2404: return 0;

PR #2243, opened 2017-04-12
Switch to new storage scheme for MongoDB

saiarcot895:
Currently, the write_mongodb plugin creates a new record (or document, as it's called in MongoDB) for each write call made by a plugin. This results in a thousand or more documents being created after a few hours (depending on how many plugins are enabled, their configuration, and the logging interval). After about 4 days, with collectd running on 3 hosts and writing to a single MongoDB database, about 900 MB is used to store the data.

[This MongoDB blog post](https://www.mongodb.com/blog/post/schema-design-for-time-series-data-in-mongodb) highlights another storage schema for use cases where per-second data needs to be stored. It involves creating one record per minute, and having 60 subdocuments contain the value for each second. This results in better efficiency of each document, as each documents has padding added onto it to have the size be a power of two. Alternatively, one record can be created per hour, and 60 subdocuments can be created for each minute, and 60 subdocuments per second within each minute.

The new storage schema I'm proposing will create one record per hour, and then create (as needed) 60 subdocuments for each minute. Within each minute's document, another subdocument for the second will be created (as needed). With a logging timer of 10 seconds, this means there will be 6 subdocuments within each minute document. With this storage format, about 230 MB is needed to store 7 days's worth of data for 3 hosts (in other words, 1/4 of the storage space for almost double the data).

### DIFF
--- a/src/write_mongodb.c
+++ b/src/write_mongodb.c
@@ -62,37 +62,66 @@ typedef struct wm_node_s wm_node_t;
 /*
  * Functions
  */
-static bson_t *wm_create_bson(const data_set_t *ds, /* {{{ */
-                              const value_list_t *vl, _Bool store_rates) {
-  bson_t *ret;
+static int wm_create_bson(const data_set_t *ds, /* {{{ */
+                          const value_list_t *vl, _Bool store_rates,
+                          bson_t *filter_bson, bson_t *update_bson) {
+  bson_t set_subdocument;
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Is I'm correct that this is not backward-compatible change?

saiarcot895:
Correct, this is not backwards-compatible.

PR #2246, opened 2017-04-12
Feat mcelog notification bugfixes

maryamtahhan:
For mcelog's unix socked data source (default /var/run/mcelog-client), mcelog plugin generates collectd notifications that merge both corrected and uncorrected memory error info. This is incorrect behaviour. Separate notifications should be created for each category and also severity should be differentiated. This pull request fixes this.
This pull request also updates the mcelog plugin to only distpatch notifications for memory corrected and uncorrected errors when the value changes for each dimm and implements a persist option if you wish to continuously send notifications.

### DIFF
--- a/src/mcelog.c
+++ b/src/mcelog.c
@@ -46,6 +46,8 @@
 #define MCELOG_DIMM_NAME "DMI_NAME"
 #define MCELOG_CORRECTED_ERR "corrected memory errors"
 #define MCELOG_UNCORRECTED_ERR "uncorrected memory errors"
+#define MCELOG_CORRECTED_ERR_TIMED "corrected memory timed errors"
+#define MCELOG_UNCORRECTED_ERR_TIMED "uncorrected memory timed errors"
 #define MCELOG_CORRECTED_ERR_TYPE_INS "corrected_memory_errors"
 #define MCELOG_UNCORRECTED_ERR_TYPE_INS "uncorrected_memory_errors"
 
@@ -106,6 +108,7 @@ static socket_adapter_t socket_adapter = {
 };
 
 static _Bool mcelog_thread_running;
+static _Bool mcelog_apply_defaults;
 
 static void mcelog_free_dimms_list_records(llist_t *dimms_list) {
 
@@ -138,6 +141,7 @@ static llentry_t *mcelog_dimm(const mcelog_memory_rec_t *rec,
     char *p_name = strdup(dimm_name);
     if (p_name == NULL) {
       ERROR(MCELOG_PLUGIN ": strdup: error");
+      free(dimm_mr);
       return NULL;
     }
 
@@ -146,6 +150,7 @@ static llentry_t *mcelog_dimm(const mcelog_memory_rec_t *rec,
     if (dimm_le == NULL) {
       ERROR(MCELOG_PLUGIN ": llentry_create(): error");
       free(dimm_mr);
+      free(p_name);
       return NULL;
     }
     pthread_mutex_lock(&g_mcelog_config.dimms_lock);
@@ -221,6 +226,10 @@ static int mcelog_config(oconfig_item_t *ci) {
       return (-1);
     }
   }
+
+  if (!use_logfile && !use_memory)
+    mcelog_apply_defaults = 1;
+
   return (0);
 }
 
@@ -357,58 +366,37 @@ static int mcelog_dispatch_mem_notifications(const mcelog_memory_rec_t *mr) {
   else
     sstrncpy(n.plugin_instance, mr->location, sizeof(n.plugin_instance));
 
-  if (dispatch_corrected_notifs) {
+  if (dispatch_corrected_notifs &&
+      (mr->corrected_err_total > 0 || mr->corrected_err_timed > 0)) {
     /* Corrected Error Notifications */
-    if (mr->corrected_err_total > 0 || mr->corrected_err_timed > 0) {
-      if (plugin_notification_meta_add_signed_int(
-              &n, MCELOG_CORRECTED_ERR, mr->corrected_err_total) < 0) {
-        ERROR(MCELOG_PLUGIN ": add corrected errors meta data failed");
-        plugin_notification_meta_free(n.meta);
-        return (-1);
-      }
-      if (plugin_notification_meta_add_signed_int(
-              &n, "corrected memory timed errors", mr->corrected_err_timed) <
-          0) {
-        ERROR(MCELOG_PLUGIN ": add corrected timed errors meta data failed");
-        plugin_notification_meta_free(n.meta);
-        return (-1);
-      }
-      ssnprintf(n.message, sizeof(n.message), "Corrected Memory Errors");
-      sstrncpy(n.type_instance, MCELOG_CORRECTED_ERR_TYPE_INS,
-               sizeof(n.type_instance));
-      plugin_dispatch_notification(&n);
-
-      if (n.meta)
-        plugin_notification_meta_free(n.meta);
-        n.meta = NULL;
-    }
+    plugin_notification_meta_add_signed_int(&n, MCELOG_CORRECTED_ERR,
+                                            mr->corrected_err_total);
+    plugin_notification_meta_add_signed_int(&n, MCELOG_CORRECTED_ERR_TIMED,
+                                            mr->corrected_err_timed);
+    ssnprintf(n.message, sizeof(n.message), MCELOG_CORRECTED_ERR);
+    sstrncpy(n.type_instance, MCELOG_CORRECTED_ERR_TYPE_INS,
+             sizeof(n.type_instance));
+    plugin_dispatch_notification(&n);
+    if (n.meta)
+      plugin_notification_meta_free(n.meta);
+    n.meta = NULL;
   }
 
-  if (dispatch_uncorrected_notifs) {
+  if (dispatch_uncorrected_notifs &&
+      (mr->uncorrected_err_total > 0 || mr->uncorrected_err_timed > 0)) {
     /* Uncorrected Error Notifications */
-    if (mr->uncorrected_err_total > 0 || mr->uncorrected_err_timed > 0) {
-      if (plugin_notification_meta_add_signed_int(
-              &n, MCELOG_UNCORRECTED_ERR, mr->uncorrected_err_total) < 0) {
-        ERROR(MCELOG_PLUGIN ": add uncorrected errors meta data failed");
-        plugin_notification_meta_free(n.meta);
-        return (-1);
-      }
-      if (plugin_notification_meta_add_signed_int(
-              &n, "uncorrected memory timed errors",
-              mr->uncorrected_err_timed) < 0) {
-        ERROR(MCELOG_PLUGIN ": add uncorrected timed errors meta data failed");
-        plugin_notification_meta_free(n.meta);
-        return (-1);
-      }
-      ssnprintf(n.message, sizeof(n.message), "Uncorrected Memory Errors");
-      sstrncpy(n.type_instance, MCELOG_UNCORRECTED_ERR_TYPE_INS,
-               sizeof(n.type_instance));
-      n.severity = NOTIF_FAILURE;
-      plugin_dispatch_notification(&n);
-      if (n.meta)
-        plugin_notification_meta_free(n.meta);
-        n.meta = NULL;
-    }
+    plugin_notification_meta_add_signed_int(&n, MCELOG_UNCORRECTED_ERR,
+                                            mr->uncorrected_err_total);
+    plugin_notification_meta_add_signed_int(&n, MCELOG_UNCORRECTED_ERR_TIMED,
+                                            mr->uncorrected_err_timed);
+    ssnprintf(n.message, sizeof(n.message), MCELOG_UNCORRECTED_ERR);
+    sstrncpy(n.type_instance, MCELOG_UNCORRECTED_ERR_TYPE_INS,
+             sizeof(n.type_instance));
+    n.severity = NOTIF_FAILURE;
+    plugin_dispatch_notification(&n);
+    if (n.meta)
+      plugin_notification_meta_free(n.meta);
+    n.meta = NULL;
   }
 
   return (0);
@@ -638,8 +626,7 @@ static void *poll_worker(__attribute__((unused)) void *arg) {
 }
 
 static int mcelog_init(void) {
-  if (g_mcelog_config.logfile != NULL &&
-      socket_adapter.unix_sock.sun_path != NULL) {
+  if (mcelog_apply_defaults) {
     INFO(MCELOG_PLUGIN
          ": No configuration selected defaulting to memory errors.");
     memset(g_mcelog_config.logfile, 0, sizeof(g_mcelog_config.logfile));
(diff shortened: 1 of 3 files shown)


src/mcelog.c
```diff
@@ -97,6 +106,75 @@ static socket_adapter_t socket_adapter = {
...
+    ssnprintf(dimm_name, sizeof(dimm_name), "%s_%s", rec->location,
+              rec->dimm_name);
+  } else
+    sstrncpy(dimm_name, rec->location, sizeof(dimm_name));
+
+  llentry_t *dimm_le = mcelog_get_dimm(dimm_name, dimms_list);
+
+  if (dimm_le == NULL) {
+    mcelog_memory_rec_t *dimm_mr =
+        (mcelog_memory_rec_t *)calloc(1, sizeof(mcelog_memory_rec_t));
```
rubenk: unnecessary cast

src/mcelog.c
```diff
@@ -97,6 +106,75 @@ static socket_adapter_t socket_adapter = {
...
+    if (!dimm_mr) {
```
rubenk: compare with NULL explicitly like you do in the rest of the code please.

src/mcelog.c
```diff
@@ -97,6 +106,75 @@ static socket_adapter_t socket_adapter = {
...
+    if (p_name == NULL) {
+      ERROR(MCELOG_PLUGIN ": strdup: error");
+      return NULL;
+    }
+
+    /* add new dimm */
+    dimm_le = llentry_create(p_name, dimm_mr);
+    if (dimm_le == NULL) {
+      ERROR(MCELOG_PLUGIN ": llentry_create(): error");
+      return NULL;
```
rubenk: doesn't this leak dimm_mr?

src/mcelog.c
```diff
@@ -212,64 +290,97 @@ static int socket_reinit(socket_adapter_t *self) {
...
-       0)) {
-    ERROR(MCELOG_PLUGIN ": add corrected errors period meta data failed");
-    plugin_notification_meta_free(n->meta);
+  int dispatch_corrected_notifs = 0, dispatch_uncorrected_notifs = 0;
+
+  if (mr == NULL)
     return (-1);
+
+  llentry_t *dimm = mcelog_dimm(mr, g_mcelog_config.dimms_list);
+  if (!dimm) {
```
rubenk: compare with NULL

src/mcelog.c
```diff
@@ -212,64 +290,97 @@ static int socket_reinit(socket_adapter_t *self) {
...
+      ERROR(MCELOG_PLUGIN ": Error adding/getting dimm memory item to/from cache");
+      return -1;
   }
-  if (plugin_notification_meta_add_signed_int(n, MCELOG_UNCORRECTED_ERR,
-                                              mr->uncorrected_err_total) < 0) {
-    ERROR(MCELOG_PLUGIN ": add corrected errors meta data failed");
-    plugin_notification_meta_free(n->meta);
-    return (-1);
+
+  mcelog_memory_rec_t *mr_old = (mcelog_memory_rec_t *)dimm->value;
```
rubenk: is the cast necessary?

src/mcelog.c
```diff
@@ -282,13 +393,21 @@ static int mcelog_submit(const mcelog_memory_rec_t *mr) {
     return (-1);
   }
 
+  llentry_t *dimm = mcelog_dimm(mr, g_mcelog_config.dimms_list);
+  if (!dimm) {
```
rubenk: compare with NULL explicitly.

rubenk:
> This patch also changes the derive types that were used for values to counter
types.

Could you explain why?

src/mcelog.c
rubenk: please use the sizeof on the variable, not the type, so sizeof(dimm_mr).

maryamtahhan:
@rubenk
I will update the code with the comments above...

>> This patch also changes the derive types that were used for values to counter
>>types.
> Could you explain why?

The types from mcelog are uint64s... derive is an int64 as far as I'm aware... so this is just to bring the types inline.

rubenk:
Thanks @maryamtahhan.

I want to understand the change better since I think there's a possibility that this will break existing graphs.

Derive and counter have different behaviour (https://collectd.org/wiki/index.php/Data_source). The mcelog counters should rarely overflow, I'm vague on the details, but I remember @octo saying we're trying to get away from counter types.

@octo any thoughts?

src/mcelog.c
maryamtahhan: But I want to allocate sizeof mcelog_memory_rec_t bytes? not the size of the pointer?

rubenk: uh yes, sizeof(*dimm_mr) then.

maryamtahhan: sorry slow moment :$ ... should have the update up soon

rubenk:
Thanks @maryamtahhan. Since you say these are bugfixes, shouldn't they be applied to 5.6?
Could you split out the change of types out into a different PR, since this warrants further discussion.

maryamtahhan:
@rubenk agreed - the mcelog counters should rarely overflow... I can revert that change if you think it's necessary... but to be safe my preference would be to keep the types inline....

rubenk:
To be safe I prefer to not change types and potentially break backwards compatibility, until I fully understand the details.

Can you pull out the other changes and open a PR against the 5.6 branch?

maryamtahhan:
@rubenk  I reversed the type change... Note: there's no mcelog commits on 5.6? so I can push against it but it would essentially by adding a new file - is this OK?

src/mcelog.c
```diff
@@ -537,7 +657,12 @@ static int mcelog_shutdown(void) {
       ret = -1;
     }
   }
-
+  pthread_mutex_lock(&g_mcelog_config.dimms_lock);
+  mcelog_free_dimms_list_records(g_mcelog_config.dimms_list);
+  llist_destroy(g_mcelog_config.dimms_list);
+  pthread_mutex_unlock(&g_mcelog_config.dimms_lock);
+  pthread_mutex_destroy(&g_mcelog_config.dimms_lock);
+  g_mcelog_config.dimms_list = NULL;
```
octo: Please set `g_mcelog_config.dimms_list = NULL;` while you're still holding the lock.

src/mcelog.c
```diff
@@ -212,64 +290,97 @@ static int socket_reinit(socket_adapter_t *self) {
...
+
+  sstrncpy(n.host, hostname_g, sizeof(n.host));
+
+  if (mr->dimm_name[0] != '\0')
+    ssnprintf(n.plugin_instance, sizeof(n.plugin_instance), "%s_%s",
+              mr->location, mr->dimm_name);
+  else
+    sstrncpy(n.plugin_instance, mr->location, sizeof(n.plugin_instance));
+
+  if (dispatch_corrected_notifs) {
```
octo: I think this would be easier to read with one less level of indentation:

```c
if (dispatch_corrected_notifs &&
    (mr->corrected_err_total > 0 || mr->corrected_err_timed > 0)) {
```

src/mcelog.c
```diff
@@ -212,64 +290,97 @@ static int socket_reinit(socket_adapter_t *self) {
...
+    /* Corrected Error Notifications */
+    if (mr->corrected_err_total > 0 || mr->corrected_err_timed > 0) {
+      if (plugin_notification_meta_add_signed_int(
```
octo: Is this error sever enough that you don't want to limp on? In the unlikely case that `plugin_notification_meta_add_signed_int()` fails, it might make sense to send the notification anyway.

If we could skip the error handling, the code would be quite a bit simpler:

```c
plugin_notification_meta_add_signed_int(&n, MCELOG_UNCORRECTED_ERR, mr->uncorrected_err_total);
plugin_notification_meta_add_signed_int(&n, "uncorrected memory timed errors", mr->uncorrected_err_timed);
```

Here and below (4 instances total).

rubenk:
My apologies, I'd swear I merged this plugin in 5.6 or 5.7. So there's no need to split anything out @maryamtahhan.

src/mcelog.c
```diff
@@ -97,6 +106,75 @@ static socket_adapter_t socket_adapter = {
...
+static void mcelog_free_dimms_list_records(llist_t *dimms_list) {
+
+  for (llentry_t *e = llist_head(dimms_list); e != NULL; e = e->next) {
+    sfree(e->key);
+    sfree(e->value);
+  }
+
+}
+
+static llentry_t *mcelog_get_dimm(const char *name, llist_t *dimms_list) {
```
maryamtahhan: no stress... I will update this

src/mcelog.c
```diff
@@ -97,6 +106,75 @@ static socket_adapter_t socket_adapter = {
...
+  if (dimm_le == NULL) {
+    mcelog_memory_rec_t *dimm_mr = calloc(1, sizeof(*dimm_mr));
+    if (dimm_mr == NULL) {
+      ERROR(MCELOG_PLUGIN ": Error allocating dimm memory item");
+      return NULL;
+    }
+    char *p_name = strdup(dimm_name);
+    if (p_name == NULL) {
+      ERROR(MCELOG_PLUGIN ": strdup: error");
+      return NULL;
```
maryamtahhan: yup, will fix.

src/mcelog.c
```diff
@@ -97,6 +106,75 @@ static socket_adapter_t socket_adapter = {
...
+    }
+
+    /* add new dimm */
+    dimm_le = llentry_create(p_name, dimm_mr);
+    if (dimm_le == NULL) {
+      ERROR(MCELOG_PLUGIN ": llentry_create(): error");
+      free(dimm_mr);
+      return NULL;
```
maryamtahhan: yup, will fix.

src/mcelog.c
```diff
@@ -212,64 +290,97 @@ static int socket_reinit(socket_adapter_t *self) {
   return (ret);
 }
 
-static int mcelog_prepare_notification(notification_t *n,
-                                       const mcelog_memory_rec_t *mr) {
-  if (n == NULL || mr == NULL)
-    return (-1);
+static int mcelog_dispatch_mem_notifications(const mcelog_memory_rec_t *mr) {
+  notification_t n = {.severity = NOTIF_WARNING,
```
maryamtahhan: will do

src/mcelog.c
maryamtahhan: Ok

src/mcelog.c
```diff
@@ -212,64 +290,97 @@ static int socket_reinit(socket_adapter_t *self) {
...
+    /* Corrected Error Notifications */
+    if (mr->corrected_err_total > 0 || mr->corrected_err_timed > 0) {
+      if (plugin_notification_meta_add_signed_int(
+              &n, MCELOG_CORRECTED_ERR, mr->corrected_err_total) < 0) {
+        ERROR(MCELOG_PLUGIN ": add corrected errors meta data failed");
+        plugin_notification_meta_free(n.meta);
+        return (-1);
+      }
+      if (plugin_notification_meta_add_signed_int(
+              &n, "corrected memory timed errors", mr->corrected_err_timed) <
```
maryamtahhan: not intentional..

maryamtahhan: I will fix so that one style is used

src/mcelog.c
```diff
@@ -212,64 +290,97 @@ static int socket_reinit(socket_adapter_t *self) {
...
   }
-  if ((mr->uncorrected_err_timed_period[0] != '\0') &&
-      (plugin_notification_meta_add_string(n, "uncorrected errors time period",
-                                           mr->uncorrected_err_timed_period) <
-       0)) {
-    ERROR(MCELOG_PLUGIN ": add corrected errors period meta data failed");
-    plugin_notification_meta_free(n->meta);
-    return (-1);
+
+  if (dispatch_uncorrected_notifs) {
```
maryamtahhan: ok

src/mcelog.c
maryamtahhan: ok, will fix

src/mcelog.c
maryamtahhan: ok

src/mcelog.c
```diff
@@ -496,15 +638,30 @@ static void *poll_worker(__attribute__((unused)) void *arg) {
 }
 
 static int mcelog_init(void) {
+  if (g_mcelog_config.logfile != NULL &&
```
maryamtahhan: g_mcelog_config.logfile == NULL and the same for the socket, will fix this

src/mcelog.c
```diff
@@ -44,10 +46,18 @@
...
+#define MCELOG_CORRECTED_ERR_TIMED "corrected memory timed errors"
+#define MCELOG_UNCORRECTED_ERR_TIMED "uncorrected memory timed errors"
+#define MCELOG_CORRECTED_ERR_TYPE_INS "corrected_memory_errors"
+#define MCELOG_UNCORRECTED_ERR_TYPE_INS "uncorrected_memory_errors"
 
 typedef struct mcelog_config_s {
   char logfile[PATH_MAX]; /* mcelog logfile */
   pthread_t tid;          /* poll thread id */
+  llist_t *dimms_list;    /* DIMMs list */
+  /* lock for dimms cache */
```
octo: Nit: looks like this comment should precede `dimms_lock`.

src/mcelog.c
```diff
@@ -96,32 +108,129 @@ static socket_adapter_t socket_adapter = {
...
+
+  if (strlen(rec->dimm_name) > 0) {
+    ssnprintf(dimm_name, sizeof(dimm_name), "%s_%s", rec->location,
+              rec->dimm_name);
+  } else
+    sstrncpy(dimm_name, rec->location, sizeof(dimm_name));
+
+  llentry_t *dimm_le = llist_search(g_mcelog_config.dimms_list, dimm_name);
+
+  if (dimm_le == NULL) {
```
octo: You can remove one level of indentation by inverting the logic:

```c
if (dimm_le != NULL)
  return dimm_le;
/* allocate new linked list entry */
```

src/mcelog.c
```diff
@@ -96,32 +108,129 @@ static socket_adapter_t socket_adapter = {
...
+      if (use_logfile) {
+        ERROR(MCELOG_PLUGIN ": Invalid configuration option: \"%s\", Logfile "
+                            "option is already configured.",
+              child->key);
+        return (-1);
+      }
+      use_memory = 1;
+      oconfig_item_t *mem_child = child->children;
+      for (int j = 0; j < child->children_num; j++) {
+        mem_child += j;
```
octo: I don't think this behaves as intended: it will skip over an increasing number of config options, the first being index 3 and then might access memory past the end of the array. The (incorrect) assumption is that `mem_child` is pointing to the beginning of the array.

This should work correctly:

```c
oconfig_item_t *mem_child = child->children + j;
```

src/mcelog.c
```diff
@@ -96,32 +108,129 @@ static socket_adapter_t socket_adapter = {
...
         return -1;
       }
+      memset(socket_adapter.unix_sock.sun_path, 0,
+             sizeof(socket_adapter.unix_sock.sun_path));
+    } else if (strcasecmp("Memory", child->key) == 0) {
+      if (use_logfile) {
+        ERROR(MCELOG_PLUGIN ": Invalid configuration option: \"%s\", Logfile "
+                            "option is already configured.",
+              child->key);
+        return (-1);
```
octo: We have (recently) standardized on using `return` without parenthesis, i.e. `return v;`. Could you please update the places the three way merge didn't know to fix?

maryamtahhan:
Hi @octo Thanks for the review. No stress, @roma or I will fix these and push up an update.

octo:
Thanks @rkorynkx and @maryamtahhan, this looks good now :)

rkorynkx:
Thanks @octo for quick review.

PR #2255, opened 2017-04-18
sensors: support current values

clemensg:
Add support for current sensor data collection and add ifdefs for old lm-sensors versions.

### DIFF
--- a/src/sensors.c
+++ b/src/sensors.c
@@ -367,6 +367,9 @@ static int sensors_load_conf(void) {
       if ((feature->type != SENSORS_FEATURE_IN) &&
           (feature->type != SENSORS_FEATURE_FAN) &&
           (feature->type != SENSORS_FEATURE_TEMP) &&
+#if SENSORS_API_VERSION >= 0x402
+          (feature->type != SENSORS_FEATURE_CURR) &&
+#endif
           (feature->type != SENSORS_FEATURE_POWER)) {
         DEBUG("sensors plugin: sensors_load_conf: "
               "Ignoring feature `%s', "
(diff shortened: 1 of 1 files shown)


src/sensors.c
```diff
@@ -367,6 +367,9 @@ static int sensors_load_conf(void) {
       if ((feature->type != SENSORS_FEATURE_IN) &&
           (feature->type != SENSORS_FEATURE_FAN) &&
           (feature->type != SENSORS_FEATURE_TEMP) &&
+#if (SENSORS_API_VERSION >= 0x402)
```
rubenk: There's no need for the round brackets around the preprocessor expression.

src/sensors.c
```diff
@@ -383,6 +386,9 @@ static int sensors_load_conf(void) {
         if ((subfeature->type != SENSORS_SUBFEATURE_IN_INPUT) &&
             (subfeature->type != SENSORS_SUBFEATURE_FAN_INPUT) &&
             (subfeature->type != SENSORS_SUBFEATURE_TEMP_INPUT) &&
+#if (SENSORS_API_VERSION >= 0x402)
```
rubenk: Same here.

src/sensors.c
```diff
@@ -513,6 +519,10 @@ static int sensors_read(void) {
       type = "temperature";
     else if (fl->feature->type == SENSORS_FEATURE_POWER)
       type = "power";
+#if (SENSORS_API_VERSION >= 0x402)
```
rubenk: And here.

clemensg:
Ah right, thanks! I updated the PR.

PR #2258, opened 2017-05-02
libcollectdclient: Add network server code.

octo:
This PR adds parsing and server code to the *collectdclient* library. The long-term plan is to eventually change the *network plugin* to use the library for sending and receiving network packets.

The implementation takes a lot of inspiration from the [collectd.org/network](https://godoc.org/collectd.org/network) Go package. For example: The `lcc_listen_and_write()` function will block while it is listening and receiving network packets (it is the responsibility of the caller to create a separate thread if concurrency is needed). The `lcc_network_parse()` function takes a `lcc_value_list_writer_t` callback which is called for every value list parsed. Many test cases for the unit tests are direct copies of the Go implementation's test cases.

I spent about 500 CPU hours fuzzing the parser with AFL and it didn't crash.

Please let me know if you'd prefer the patch be split into smaller chunks, for example the network parsing and networking code could easily be in separate PRs. Please pay special attention to the *interface* (i.e. the `*.h` files) since it will be essentially immutable once this is released.


rubenk:
Amazing work @octo! I'll try to find some time to review this as soon as I can.

octo:
FreeBSD is happy! Hooray! :tada:

rpv-tomsk:
Hi!

>libcollectdclient: Add network server code.

What is the goal of this work? How that can be used / how did you use libcollectdclient with this feature?

octo:
Hi @rpv-tomsk, the goal of this work is to provide a C level library for parsing collectd network data, which allows to develop random tools much easier. For example, you could build a very simple top-style curses application. Another application would be to implement Perl and Python bindings so that the network parsing doesn't need to be implemented in each language all over again.

I haven't used this code in any project yet (and don't have any concrete plans to do that), but since *I'm* having such an easy time coming up with ideas, I'm sure the community will do something cool with this.

Best regards,
—octo

rpv-tomsk:
Thanks for the explanation!

Makefile.am
```diff
@@ -471,20 +475,34 @@ endif
...
 	$(AM_CPPFLAGS) \
-	-I$(srcdir)/src/libcollectdclient/collectd \
-	-I$(top_builddir)/src/libcollectdclient/collectd \
+	-I$(srcdir)/src/libcollectdclient \
+	-I$(top_builddir)/src/libcollectdclient \
 	-I$(srcdir)/src/daemon
-libcollectdclient_la_LDFLAGS = -version-info 1:0:0
-libcollectdclient_la_LIBADD = 
+libcollectdclient_la_LDFLAGS = -version-info 2:0:1
+libcollectdclient_la_LIBADD = -lm
```
rubenk: we should do some detection as for when we need to link with libmath, but let's do that later. What function / symbol needs this?

src/libcollectdclient/network_parse.c
```diff
@@ -0,0 +1,598 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ * Authors:
+ *   Florian octo Forster <octo at collectd.org>
+ **/
+
+#if HAVE_CONFIG_H
```
rubenk: I'd prefer if we do not guard this include. I'd rather have compilation fail because the header can't be found.

src/libcollectdclient/network_parse.c
```diff
@@ -0,0 +1,598 @@
...
+
+/* for be{16,64}toh */
+#if HAVE_ENDIAN_H
+#include <endian.h>
+#elif HAVE_SYS_ENDIAN_H
+#include <sys/endian.h>
+#endif
+
+#define GCRYPT_NO_DEPRECATED
+#include <gcrypt.h>
```
rubenk: I take it gcrypt is now a required dependency?

rubenk:
@octo I think the api is great, and have no remarks about that.

I left a few minor remarks inline, but other than that, this looks good to me :)

Makefile.am
octo: I don't remember to be honest. It compiles on my local machine without `-lm` – I may have added the link because of another platform. My guess would be `isnan()` on FreeBSD …

src/libcollectdclient/network_parse.c
octo: That's reasonable. Done.

src/libcollectdclient/network_parse.c
octo: Good point, that is in an suboptimal condition. We should either not compile the library at all if gcrypt is missing or go down the `#ifdef` rabbit hole. I'd be fine with either but lean towards the ifdeffery. What are your thoughts?

rubenk: I'm slightly in favour of making it required. It's one less code path to worry about, and most distro packages require it anyhow.

octo:
ACK, let's keep it simple.

octo:
I absolutely underestimated how much depends on *libcollectdclient*: *collectdctl*, *collectd-nagios* and *collectd-tg*. I don't think it's reasonable not to build those tools because a feature they don't use is missing from the library. I think it's more reasonable to ifdef-out the crypto features.

PR #2260, opened 2017-05-03
redis: adding support for keys count per database

bufadu:
This PR adds support for keys count to redis plugin.
Currently redis plugin collects statistics about redis server but doesn't care about databases usage.

This patch simply add parsing of following lines 
```
# Keyspace
db0:keys=4,expires=0,avg_ttl=0
db1:keys=5581,expires=0,avg_ttl=0
db12:keys=3,expires=0,avg_ttl=0
```
Note that `expires` and `avg_ttl` are not parsed.

Resulting identifiers will be like
```
<server>/redis-<node>/records-0
<server>/redis-<node>/records-1
<server>/redis-<node>/records-12
```

### DIFF
--- a/src/redis.c
+++ b/src/redis.c
@@ -365,24 +365,24 @@ static int redis_db_stats(char *node, char const *info_line) /* {{{ */
   int i;
 
   for (db = 0; db < max_db; db++) {
-      sprintf(field_name, "db%d:keys", db);
+	ssnprintf(field_name, sizeof(field_name), "db%d:keys", db);
 
-      str = strstr(info_line, field_name);
-      if (str) {
-		  str += strlen(field_name) + 1; /* also skip the '=' */
-		  for (i = 0; (*str && (isdigit((unsigned char)*str) || *str == '.')); i++, str++)
-			buf[i] = *str;
-		  buf[i] = '\0';
+	str = strstr(info_line, field_name);
+	if (str) {
+	  str += strlen(field_name) + 1; /* also skip the '=' */
+	  for (i = 0; (*str && (isdigit((unsigned char)*str) || *str == '.')); i++, str++)
+		buf[i] = *str;
+	  buf[i] = '\0';
 
-		  if (parse_value(buf, &val, DS_TYPE_GAUGE) == -1) {
-			  WARNING("redis plugin: Unable to parse field `%s'.", field_name);
-			  return (-1);
-		  }
-
-		  sprintf(db_id, "%d", db);
-		  redis_submit (node, "records", db_id, val);
+	  if (parse_value(buf, &val, DS_TYPE_GAUGE) == -1) {
+		WARNING("redis plugin: Unable to parse field `%s'.", field_name);
+		return (-1);
 	  }
-    }
+
+	  ssnprintf(db_id, sizeof(db_id), "%d", db);
+	  redis_submit (node, "records", db_id, val);
+	}
+  }
   return (0);
 
 } /* }}} int redis_db_stats */
--- a/src/redis.c
+++ b/src/redis.c
@@ -36,6 +36,7 @@
 #define REDIS_DEF_PASSWD ""
 #define REDIS_DEF_PORT 6379
 #define REDIS_DEF_TIMEOUT 2000
+#define REDIS_DEF_DB_COUNT 16
 #define MAX_REDIS_NODE_NAME 64
 #define MAX_REDIS_PASSWD_LENGTH 512
 #define MAX_REDIS_VAL_SIZE 256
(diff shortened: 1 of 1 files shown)


src/redis.c
```diff
@@ -353,6 +353,41 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
   return 0;
 } /* }}} int redis_handle_query */
 
+static int redis_db_stats(char *node, char const *info_line) /* {{{ */
+{
+  unsigned char db;
+  unsigned char max_db = 16;
+  static char field_name[10];
```
octo: Most of these variables are only used inside the `for` loop: please declare them in an inner scope, to they have the shortest possible lifespan.

src/redis.c
```diff
@@ -353,6 +353,41 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
...
+  char *str;
+  static char buf[MAX_REDIS_VAL_SIZE];
+  value_t val;
+  int i;
+
+  for (db = 0; db < max_db; db++) {
+	ssnprintf(field_name, sizeof(field_name), "db%d:keys", db);
+
+	str = strstr(info_line, field_name);
+	if (str) {
```
octo: Please invert this logic:

```c
if (!str)
  continue;
/* remaining code */
```

This way there is one less level of indentation, which is usually easier to read.
This way

src/redis.c
```diff
@@ -353,6 +353,41 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
...
+  unsigned char db;
+  unsigned char max_db = 16;
+  static char field_name[10];
+  static char db_id[3];
+  char *str;
+  static char buf[MAX_REDIS_VAL_SIZE];
+  value_t val;
+  int i;
+
+  for (db = 0; db < max_db; db++) {
```
octo: Nit, at your option: the currently recommended coding style is to declare the loop variable as part of the `for` loop, i.e.:

```c
for (int db = 0; db < max_db, db++) {
```

src/redis.c
```diff
@@ -353,6 +353,41 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
   return 0;
 } /* }}} int redis_handle_query */
 
+static int redis_db_stats(char *node, char const *info_line) /* {{{ */
+{
+  unsigned char db;
```
octo: Please use `int` instead of `unsigned char` for `db` and `max_db`.

src/redis.c
```diff
@@ -353,6 +353,41 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
...
+  unsigned char max_db = 16;
+  static char field_name[10];
+  static char db_id[3];
+  char *str;
+  static char buf[MAX_REDIS_VAL_SIZE];
+  value_t val;
+  int i;
+
+  for (db = 0; db < max_db; db++) {
+	ssnprintf(field_name, sizeof(field_name), "db%d:keys", db);
```
octo: The `%d` format expects an `int` but `db` is actually an `unsigned char`, which has the potential for an out of bounds read. As mentioned above, simply change `db` to an `int` to fix this issue.

src/redis.c
```diff
@@ -353,6 +353,41 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
...
+
+	str = strstr(info_line, field_name);
+	if (str) {
+	  str += strlen(field_name) + 1; /* also skip the '=' */
+	  for (i = 0; (*str && (isdigit((unsigned char)*str) || *str == '.')); i++, str++)
```
octo: `isdigit()` expects an `int`, so you should do:

```c
isdigit((int) *str)
/*      ^^^^^ */
```

src/redis.c
```diff
@@ -353,6 +353,41 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
...
+		buf[i] = *str;
+	  buf[i] = '\0';
+
+	  if (parse_value(buf, &val, DS_TYPE_GAUGE) == -1) {
```
octo: Parse value may return other non-zero values on error, such as `EINVAL`, so you should check for `!= 0` instead.

src/redis.c
```diff
@@ -429,6 +464,8 @@ static int redis_read(void) /* {{{ */
     redis_handle_info(rn->name, rr->str, "total_bytes", "output",
                       "total_net_output_bytes", DS_TYPE_DERIVE);
 
+	redis_db_stats (rn->name, rr->str);
```
octo: Indentation is off here. Please run `clang-format -style=file -i src/redis.c` to fix these formatting issues.

octo:
Hi @bufadu,

thank you very much for your patch! I've left some comments inline, once they're resolved this is good to go :)

Best regards,
—octo

bufadu:
Hello @octo,

Thanks for your comments, I have pushed edited code following your recommendations.
Hope it will be good for you.

src/redis.c
```diff
@@ -355,39 +356,37 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
...
+    char *str;
+    int i;
+
+    ssnprintf(field_name, sizeof(field_name), "db%d:keys", db);
+
+    str = strstr(info_line, field_name);
+    if (!str)
+      continue;
+
+    str += strlen(field_name) + 1; /* also skip the '=' */
```
octo: This comment is only helpful if you know the format being parsed. Please add a comment to the function describing the expected input format, e.g.:

```
/* redis_db_stats parses and dispatches Redis database statistics,
 * currently the number of keys for each database.
 * info_line needs to have the following format:
 *   db0:keys=4,expires=0,avg_ttl=0
 */

File: src/redis.c [hunk 10]
```diff
@@ -355,39 +356,37 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
...
+    for (i = 0; (*str && (isdigit((int)*str) || *str == '.')); i++, str++)
```
octo: Same as for the outer loop: Declare the `i` variable inside the `for` loop:

```c
for (int i = 0; …) {
```

bufadu: I don't have declared `i` inside the `for` loop since `i` is used outside the loop to set the NUL at the end of `buf`
```C
    for (i = 0; (*str && isdigit((int)*str)); i++, str++)
      buf[i] = *str;
    buf[i] = '\0';
```
By the way, thanks to your comment, I noticed that `*str == '.'` test is not required in this loop (keys count is an integer value). I will fix that in my next commit.

octo: Good point, I missed that use of `i`. Thanks!

octo:
Thanks @bufadu!

PR #2274, opened 2017-05-12
src/utils_cmd_putval.c: Fix multi-value PUTVAL.

octo:
`vl.values` was allocated outside of the loop and then copied into each `ret_putval->vl[]`. This means that later values overwrote the values stored in previous `ret_putval->vl[]`.

### DIFF
--- a/src/utils_cmd_putval.c
+++ b/src/utils_cmd_putval.c
@@ -138,14 +138,6 @@ cmd_status_t cmd_parse_putval(size_t argc, char **argv,
   type = NULL;
   type_instance = NULL;
 
-  vl.values_len = ds->ds_num;
-  vl.values = malloc(vl.values_len * sizeof(*vl.values));
-  if (vl.values == NULL) {
-    cmd_error(CMD_ERROR, err, "malloc failed.");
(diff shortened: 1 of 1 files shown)


src/utils_cmd_putval.c
```diff
@@ -177,33 +169,46 @@ cmd_status_t cmd_parse_putval(size_t argc, char **argv,
...
       result = CMD_PARSE_ERROR;
+      vl.values_len = 0;
+      sfree(vl.values);
       break;
     }
 
-    tmp = (value_list_t *)realloc(ret_putval->vl, (ret_putval->vl_num + 1) *
-                                                      sizeof(*ret_putval->vl));
+    tmp = realloc(ret_putval->vl,
+                  (ret_putval->vl_num + 1) * sizeof(*ret_putval->vl));
```
rubenk: we could really use an implementation of OpenBSD's reallocarray(3)

rubenk:
Nice catch! The code looks much better now to.

octo:
Thanks :) I did miss a semi-colon. I'll merge if the CI builders are happy now.

PR #2276, opened 2017-05-12
Performance monitoring plugin intel_pmu

serhiyx:
The intel_pmu plugin reads performance counters provided by the Linux kernel perf interface. The plugin uses jevents library to resolve named events to perf events and access perf interface.


src/collectd.conf.pod
```diff
@@ -3082,6 +3082,86 @@ Defaults to B<false>.
...
+    KernelPMUEvents true
+    SWEvents true
+    HWSpecificEvents "L2_RQSTS.CODE_RD_HIT,L2_RQSTS.CODE_RD_MISS"
+  </Plugin>
+
+B<Options:>
+
+=over 4
+
+=item B<HWCacheEvents> B<false>|B<true>
```
octo: Configuration options with this functionality are usually called `ReportFoo`, for example `ReportNumCpu` of the *CPU plugin*.

src/collectd.conf.pod
```diff
@@ -3082,6 +3082,86 @@ Defaults to B<false>.
...
+  - dTLB-stores
+  - dTLB-store-misses
+  - dTLB-prefetches
+  - dTLB-prefetch-misses
+  - iTLB-loads
+  - iTLB-load-misses
+  - branch-loads
+  - branch-load-misses
+
+=item B<KernelPMUEvents> B<false>|B<true>
```
octo: Dito: `ReportKernelPMUEvents`

src/collectd.conf.pod
```diff
@@ -3082,6 +3082,86 @@ Defaults to B<false>.
...
+Enable or disable measuring of the following events:
+  - cpu-cycles
+  - instructions
+  - cache-references
+  - cache-misses
+  - branches
+  - branch-misses
+  - bus-cycles
+
+=item B<SWEvents> B<false>|B<true>
```
octo: Dito: `ReportSWEvents`; please consider expanding "SW" to "Software" to make this option easier to read.

src/collectd.conf.pod
```diff
@@ -3082,6 +3082,86 @@ Defaults to B<false>.
...
+  - cpu-migrations
+  - page-faults
+  - minor-faults
+  - major-faults
+  - alignment-faults
+  - emulation-faults
+
+=item B<HWSpecificEvents> I<events>
+
+This field is a list of comma separated event names. To be able to monitor all
```
octo: > comma separated

Config options can take multiple arguments. Please take a list of strings instead of a single string that is then parsed inside the plugin.

src/collectd.conf.pod
```diff
@@ -3082,6 +3082,86 @@ Defaults to B<false>.
...
+  - task-clock
+  - context-switches
+  - cpu-migrations
+  - page-faults
+  - minor-faults
+  - major-faults
+  - alignment-faults
+  - emulation-faults
+
+=item B<HWSpecificEvents> I<events>
```
octo: With the boolean options all having the form "Report… true|false", I'd suggest to rename this option to **HardwareEvents**.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+ * SOFTWARE.
+ *
+ * Authors:
+ *   Serhiy Pshyk <serhiyx.pshyk@intel.com>
+ **/
+
+#include "collectd.h"
+#include "common.h"
+
+#include "jevents.h"
```
octo: Please use angle brackets for files not included with collectd: `#include <jevents.h>`. Dito for `<jsession.h>`.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+
+  DEBUG(PMU_PLUGIN ": Events:");
+
+  struct event *e;
+
+  for (e = g_ctx.event_list->eventlist; e; e = e->next) {
+    DEBUG(PMU_PLUGIN ":   event       : %s", e->event);
+    DEBUG(PMU_PLUGIN ":     group_lead: %d", e->group_leader);
+    DEBUG(PMU_PLUGIN ":     end_group : %d", e->end_group);
+    DEBUG(PMU_PLUGIN ":     type      : 0x%X", e->attr.type);
```
octo: Please use the `%#X` format instead of writing out the `0x` manually. Here and below.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+    DEBUG(PMU_PLUGIN ":     config    : 0x%X", (int)e->attr.config);
```
octo: Please cast to `(unsigned)` when using the `%X` format.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+
+static void pmu_submit_counter(int cpu, char *event, counter_t value) {
+  value_list_t vl = VALUE_LIST_INIT;
+
+  vl.values = &(value_t){.counter = value};
+  vl.values_len = 1;
+
+  sstrncpy(vl.plugin, PMU_PLUGIN, sizeof(vl.plugin));
+  if (cpu == -1) {
+    snprintf(vl.plugin_instance, sizeof(vl.plugin_instance), "all");
```
octo: Please use collectds wrapper, `ssnprintf()`. Here and below.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+
+  return (0);
+}
+
+static int pmu_add_events(struct eventlist *el, uint32_t type,
+                          event_info_t *events, int count) {
+
+  for (int i = 0; i < count; i++) {
+    struct event *e =
+        calloc(sizeof(struct event) + sizeof(struct efd) * el->num_cpus, 1);
```
octo: This way of calling `calloc` defeats its purpose. Please use this instead:

```c
struct event *e = calloc(el->num_cpus, sizeof(*e) + sizeof(struct efd);
```

Please add a comment explaining why the `sizeof(struct efd)` is needed.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+    e->event = strdup(events[i].name);
+  }
+
+  return (0);
+}
+
+static int pmu_parse_events(struct eventlist *el, char *events) {
+  char *s, *tmp;
+
+  events = strdup(events);
```
octo: Please assign this value to a new variable and change the argument to `char const *`.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+  if (!events)
+    return -1;
+
+  for (s = strtok_r(events, ",", &tmp); s; s = strtok_r(NULL, ",", &tmp)) {
+    bool group_leader = false, end_group = false;
```
octo: Please use `_Bool` and `0` / `1` instead of the `<stdbool.h>` aliases.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+    if (s[0] == '{') {
+      s++;
+      group_leader = true;
+    } else if (len = strlen(s), len > 0 && s[len - 1] == '}') {
+      s[len - 1] = 0;
+      end_group = true;
+    }
+
+    struct event *e =
+        calloc(sizeof(struct event) + sizeof(struct efd) * el->num_cpus, 1);
```
octo: As above: `calloc(el->num_cpus, sizeof(*e) + sizeof(struct efd))`

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+    if (e == NULL) {
+      ERROR(PMU_PLUGIN ": Failed to allocate event structure");
+      return (-ENOMEM);
+    }
+
+    e->attr.type = type;
+    e->attr.config = events[i].config;
+    e->attr.size = PERF_ATTR_SIZE_VER0;
+    e->group_leader = false;
```
octo: Please don't use the `false` define, instead use `0`. Since `calloc` zeroes the memory, you can just remote this and the following line.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+    if (el->eventlist_last)
+      el->eventlist_last->next = e;
+    el->eventlist_last = e;
+    e->event = strdup(events[i].name);
+  }
+
+  return (0);
+}
+
+static int pmu_parse_events(struct eventlist *el, char *events) {
```
octo: As mentioned above: the config parser can give you a list of strings – please use that instead of implementing your own (sub-)parsing.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+    if (ret != 0) {
+      ERROR(PMU_PLUGIN ": Failed to setup perf events for the event list.");
+      goto init_error;
+    }
+  } else {
+    WARNING(PMU_PLUGIN
+            ": Events list is empty. No events were setup for monitoring.");
+  }
+
+  return (0);
```
octo: Please unify this style. Both, `return foo;` and `return (foo);` are okay, just use the same everywhere in the file.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+static int pmu_parse_events(struct eventlist *el, char *events) {
+  char *s, *tmp;
+
+  events = strdup(events);
+  if (!events)
+    return -1;
+
+  for (s = strtok_r(events, ",", &tmp); s; s = strtok_r(NULL, ",", &tmp)) {
+    bool group_leader = false, end_group = false;
+    int len;
```
octo: This should be a `size_t`.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+}
+
+static int pmu_read(__attribute__((unused)) user_data_t *ud) {
+  int ret;
+
+  DEBUG(PMU_PLUGIN ": %s:%d", __FUNCTION__, __LINE__);
+
+  ret = read_all_events(g_ctx.event_list);
+  if (ret != 0) {
+    DEBUG(PMU_PLUGIN ": Failed to read values of all events.");
```
octo: This should be an `ERROR()` since you abort and return prematurely.

src/intel_pmu.c
```diff
@@ -0,0 +1,512 @@
...
+    return (0);
+  }
+
+  ret = pmu_dispatch_data();
+  if (ret != 0) {
+    DEBUG(PMU_PLUGIN ": Failed to dispatch event values.");
```
octo: Dito: Use `ERROR()` here.

src/intel_pmu.c
rpv-tomsk: @octo 
There is a WIP about parentheses removing (#2277). 
Maybe syntax with parentheses should be noted as forbidden / deprecated?

octo: Agreed, with #2277, please remove all parenthesis from this PR.

src/intel_pmu.c
serhiyx: Comment added, but size did not change because I need  **sizeof(event) + sizeof(efd)*num_cpus** but not **(sizeof(event) + sizeof(efd)) * num_cpus**

src/intel_pmu.c
serhiyx: Refactored this function to receive list of strings provided by config parser. But some subparsing routine is still needed as each entry can contain multiple event names to create group of events.

src/collectd.conf.pod
serhiyx: Refactored this functionality to use list of strings provided by config parser. But some subparsing routine is still needed as each entry can contain multiple event names to create group of events.

contrib/systemd.collectd.service
```diff
@@ -11,6 +11,20 @@ EnvironmentFile=-/etc/default/collectd
...
+#
+# Here's a (incomplete) list of the plugins known environment requirements:
+#   intel_pmu       XDG_CACHE_HOME
+#
+# Example, if you use the intel_pmu plugin and cache is located in the /opt
+# directory:
+# Environment=XDG_CACHE_HOME=/opt
+#
+# By default, the HOME directory is chosen:
+Environment=
```
rubenk: Instead of getting this from the environment, would it be possible to turn this into a config option settable in collectd.conf?

configure.ac
```diff
@@ -4364,6 +4364,77 @@ then
 fi
 # }}}
 
+# --with-libjevents {{{
+with_libjevents_cppflags=""
+with_libjevents_ldflags=""
+AC_ARG_WITH(libjevents, [AS_HELP_STRING([--with-libjevents@<:@=PREFIX@:>@], [Path to libjevents.])],
```
rubenk: Please keep the same code style as the rest of configure.ac

configure.ac
```diff
@@ -4364,6 +4364,77 @@ then
...
+  SAVE_CPPFLAGS="$CPPFLAGS"
+  CPPFLAGS="$CPPFLAGS $with_libjevents_cppflags"
+
+  AC_CHECK_HEADERS(jevents.h, [with_libjevents="yes"], [with_libjevents="no (jevents.h not found)"])
+
+  CPPFLAGS="$SAVE_CPPFLAGS"
+fi
+if test "x$with_libjevents" = "xyes"
+then
+  SAVE_CPPFLAGS="$CPPFLAGS"
```
rubenk: AC_CHECK_LIB doesn't use CPPFLAGS, so no need to save them.

configure.ac
```diff
@@ -4364,6 +4364,77 @@ then
...
+  CPPFLAGS="$SAVE_CPPFLAGS"
+  LDFLAGS="$SAVE_LDFLAGS"
+fi
+if test "x$with_libjevents" = "xyes"
+then
+  SAVE_CPPFLAGS="$CPPFLAGS"
+  SAVE_LDFLAGS="$LDFLAGS"
+  SAVE_LIBS="$LIBS"
+  CPPFLAGS="$CPPFLAGS -fPIC"
+  LDFLAGS="$LDFLAGS -shared"
```
rubenk: Setting -fPIC and -shared is not needed, you're creating an executable, not a library.

configure.ac
```diff
@@ -4364,6 +4364,77 @@ then
...
+  AC_LINK_IFELSE([AC_LANG_SOURCE(
+    [[
+      #include <stdio.h>
+      #include "jevents.h"
+      void print_cpu(void){
+        printf("%s", get_cpu_str());
+      }
+    ]]
+  )],
+  [with_libjevents="yes"], [with_libjevents="no (could not link to libjevents. Check jevents is compiled with -fPIC.)"])
```
rubenk: I don't understand this. If libjevents is not compiled with -fPIC, it is not a shared library.

configure.ac
```diff
@@ -4364,6 +4364,77 @@ then
...
+  CPPFLAGS="$SAVE_CPPFLAGS"
+  LDFLAGS="$SAVE_LDFLAGS"
+  LIBS="$SAVE_LIBS"
+fi
+if test "x$with_libjevents" = "xyes"
+then
+  BUILD_WITH_LIBJEVENTS_CPPFLAGS="$with_libjevents_cppflags"
+  BUILD_WITH_LIBJEVENTS_LDFLAGS="$with_libjevents_ldflags"
+  BUILD_WITH_LIBJEVENTS_LIBS="-ljevents"
+  AC_SUBST(BUILD_WITH_LIBJEVENTS_CPPFLAGS)
```
rubenk: You should not use AC_SUBST in a conditional statement. Please move it below the fi.

contrib/systemd.collectd.service
serhiyx: This environment variable is not used by the plugin itself. It is used inside **jevents** library that is used by the plugin.

rubenk: And there's no way to pass this to jevents except from the environment?

configure.ac
serhiyx: Actually I want to create shared object that is linked with jevents library just to check if jevents was compiled with -fPIC. I've added this logic to avoid errors during collectd installation/compilation - if collectd is compiled on the system with jevents installed we check if jevents version is suitable for us and if not just disable intel_pmu plugin during configure.

configure.ac
serhiyx: By default jevents library is compiled without -fPIC. But this flag is required when jevents library is linked with pmu plugin (intel_pmu.so) to avoid the following error:

`/usr/bin/ld: /usr/local/lib64/libjevents.a(cache.o): relocation R_X86_64_32S against `.bss' can not be used when making a shared object; recompile with -fPIC`

configure.ac
rubenk: I think it's more trouble than it's worth, you've already a note in the README that jevents should be position independent.

contrib/systemd.collectd.service
serhiyx: Yes, it is possible. Will rework the solution to use config option and get rid of using of environment variable..

configure.ac
serhiyx: Agree. Will remove this.

README
```diff
@@ -240,7 +245,7 @@ Features
 
     - netapp
       Plugin to query performance values from a NetApp storage system using the
-      “Manage ONTAP” SDK provided by NetApp.
+      “Manage ONTAP” SDK provided by NetApp.
```
rubenk: What changed on this line?

configure.ac
```diff
@@ -4364,6 +4364,53 @@ then
 fi
 # }}}
 
+# --with-libjevents {{{
+with_libjevents_cppflags=""
+with_libjevents_ldflags=""
+AC_ARG_WITH([libjevents],
+  [AS_HELP_STRING([--with-libjevents@<:@=PREFIX@:>@], [Path to libjevents.])],
+  [
+    if test "x$withval" != "xno" && test "x$withval" != "xyes"
```
rubenk: put if and then on the same line please.

configure.ac
```diff
@@ -4364,6 +4364,53 @@ then
...
+  ],
+  [with_libjevents="yes"]
+)
+
+if test "x$with_libjevents" = "xyes"
+then
+  SAVE_CPPFLAGS="$CPPFLAGS"
+  CPPFLAGS="$CPPFLAGS $with_libjevents_cppflags"
+
+  AC_CHECK_HEADERS(jevents.h, [with_libjevents="yes"], [with_libjevents="no (jevents.h not found)"])
```
rubenk: please quote all the arguments.

configure.ac
```diff
@@ -4364,6 +4364,53 @@ then
...
+
+  CPPFLAGS="$SAVE_CPPFLAGS"
+fi
+if test "x$with_libjevents" = "xyes"
+then
+  SAVE_LDFLAGS="$LDFLAGS"
+  LDFLAGS="$LDFLAGS $with_libjevents_ldflags"
+
+  AC_CHECK_LIB(jevents, json_events, [with_libjevents="yes"], [with_libjevents="no (Can't find libjevents)"])
```
rubenk: please quote all the arguments, and break this long line.

configure.ac
```diff
@@ -4364,6 +4364,53 @@ then
...
+
+  LDFLAGS="$SAVE_LDFLAGS"
+fi
+if test "x$with_libjevents" = "xyes"
+then
+  BUILD_WITH_LIBJEVENTS_CPPFLAGS="$with_libjevents_cppflags"
+  BUILD_WITH_LIBJEVENTS_LDFLAGS="$with_libjevents_ldflags"
+  BUILD_WITH_LIBJEVENTS_LIBS="-ljevents"
+fi
+AC_SUBST(BUILD_WITH_LIBJEVENTS_CPPFLAGS)
```
rubenk: please quote the argument.

README
serhiyx: must have been fixed by my editor. no-break space symbol (c2a0) was changed to space symbol (20).
Should I revert this change?

rubenk: Heh, nice one. No problem, please leave it in, I prefer spaces anyway :)

Makefile.am
```diff
@@ -876,6 +876,14 @@ hugepages_la_SOURCES = src/hugepages.c
 hugepages_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 endif
 
+if BUILD_PLUGIN_INTEL_PMU
+pkglib_LTLIBRARIES += intel_pmu.la
+intel_pmu_la_SOURCES = src/intel_pmu.c
+intel_pmu_la_CFLAGS = $(AM_CFLAGS) $(BUILD_WITH_LIBJEVENTS_CPPFLAGS)
```
rubenk: BUILD_WITH_LIBJEVENTS_CPPFLAGS should be added to intel_pmu_la_CPPFLAGS.

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+  for (e = g_ctx.event_list->eventlist; e; e = e->next) {
+    DEBUG(PMU_PLUGIN ":   event       : %s", e->event);
+    DEBUG(PMU_PLUGIN ":     group_lead: %d", e->group_leader);
+    DEBUG(PMU_PLUGIN ":     end_group : %d", e->end_group);
+    DEBUG(PMU_PLUGIN ":     type      : %#x", e->attr.type);
+    DEBUG(PMU_PLUGIN ":     config    : %#x", (unsigned)e->attr.config);
+    DEBUG(PMU_PLUGIN ":     size      : %d", e->attr.size);
+  }
+
+  return;
```
rubenk: no need to return from a void function.

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+  DEBUG(PMU_PLUGIN ": Config:");
+  DEBUG(PMU_PLUGIN ":   hw_cache_events   : %d", g_ctx.hw_cache_events);
+  DEBUG(PMU_PLUGIN ":   kernel_pmu_events : %d", g_ctx.kernel_pmu_events);
+  DEBUG(PMU_PLUGIN ":   software_events   : %d", g_ctx.sw_events);
+
+  for (size_t i = 0; i < g_ctx.hw_events_count; i++) {
+    DEBUG(PMU_PLUGIN ":   hardware_events[%zu]: %s", i, g_ctx.hw_events[i]);
+  }
+
+  return;
```
rubenk: no need to return from a void function.

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+    }
+
+    g_ctx.hw_events_count++;
+  }
+
+  return 0;
+}
+
+static int pmu_config(oconfig_item_t *ci) {
+  int ret = 0;
```
rubenk: can you reduce the scope by moving this into the for loop?

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+    } else if (strcasecmp("EventList", child->key) == 0) {
+      ret = cf_util_get_string_buffer(child, g_ctx.event_list_fn,
+                                      sizeof(g_ctx.event_list_fn));
+    } else if (strcasecmp("HardwareEvents", child->key) == 0) {
+      ret = pmu_config_hw_events(child);
+    } else if (strcasecmp("ReportSoftwareEvents", child->key) == 0) {
+      ret = cf_util_get_boolean(child, &g_ctx.sw_events);
+    } else {
+      ERROR(PMU_PLUGIN ": Unknown configuration parameter \"%s\".", child->key);
+      ret = (-1);
```
rubenk: no need for the braces.

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+  } else {
+    ssnprintf(vl.plugin_instance, sizeof(vl.plugin_instance), "%d", cpu);
+  }
+  sstrncpy(vl.type, "counter", sizeof(vl.type));
+  sstrncpy(vl.type_instance, event, sizeof(vl.type_instance));
+
+  plugin_dispatch_values(&vl);
+}
+
+static int pmu_dispatch_data(void) {
```
rubenk: this function always returns 0, make it void?

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+  DEBUG(PMU_PLUGIN ": %s:%d", __FUNCTION__, __LINE__);
+
+  ret = read_all_events(g_ctx.event_list);
+  if (ret != 0) {
+    ERROR(PMU_PLUGIN ": Failed to read values of all events.");
+    return 0;
+  }
+
+  ret = pmu_dispatch_data();
+  if (ret != 0) {
```
rubenk: this is a no-op, pmu_dispatch_data always returns 0.

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+
+static int pmu_read(__attribute__((unused)) user_data_t *ud) {
+  int ret;
+
+  DEBUG(PMU_PLUGIN ": %s:%d", __FUNCTION__, __LINE__);
+
+  ret = read_all_events(g_ctx.event_list);
+  if (ret != 0) {
+    ERROR(PMU_PLUGIN ": Failed to read values of all events.");
+    return 0;
```
rubenk: why not return an error?

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+  }
+
+  ret = pmu_dispatch_data();
+  if (ret != 0) {
+    ERROR(PMU_PLUGIN ": Failed to dispatch event values.");
+    return 0;
```
rubenk: why not return an error?

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+        calloc(sizeof(struct event) + sizeof(struct efd) * el->num_cpus, 1);
+    if (e == NULL) {
+      ERROR(PMU_PLUGIN ": Failed to allocate event structure");
+      return -ENOMEM;
+    }
+
+    e->attr.type = type;
+    e->attr.config = events[i].config;
+    e->attr.size = PERF_ATTR_SIZE_VER0;
+    e->next = NULL;
```
rubenk: this is a no-op, the struct has already been zeroed by calloc.

src/intel_pmu.c
```diff
@@ -0,0 +1,577 @@
...
+  if (ret != 0) {
+    ERROR(PMU_PLUGIN ": Failed to dispatch event values.");
+    return 0;
+  }
+
+  return 0;
+}
+
+static int pmu_add_events(struct eventlist *el, uint32_t type,
+                          event_info_t *events, int count) {
```
rubenk: count should be size_t.

rubenk:
@serhiyx I left a few more review comments. Once these are resolved I think this is good to go in. Feel free to also add yourself to AUTHORS if you want to.

rubenk:
@octo I think this is good to go in, and think all your remarks have been addressed. Do you want to merge this?

octo:
Thank you very much for your great work, @serhiyx!

PR #2277, opened 2017-05-13
Remove parentheses around return arguments

rubenk:
They don't do any harm but they are not really needed either.
Contributors are not used to following this style, and reviewers have to
keep pointing it out in reviews. This takes up valuable time for both
the contributor and the reviewer and distracts from the more important
issues.

I used the following Coccinelle patch:

```diff
@@
expression e;
@@
- return (e);
+ return e;
```

spatch is having trouble with some files so I have cleaned up the rest
with a few regexes.


rubenk:
No problem, it was a pretty big diff after all. Thanks for merging!

PR #2278, opened 2017-05-14
perl plugin: Fix potential double-free.

octo:
The av2notification_meta() function updated it's parameter by doing

    (*m) = malloc();

Afterwards, if an error occurred, it would call `free(*m);` and (potentially)
return, leaving an invalid pointer stored in `m`. The caller would then try to
free the returned pointer because it was non-NULL.

This patch fixes this by doing several code cleanups:

*   The meta argument is not updated unnecessarily. Instead, a local variable
    is allocated and used within the loop and only on success is the return
    pointer updated.
*   Introduce additional local variables instead of re-using the `tmp` variable
    multiple times.
*   Name the variable pointing to the end of the linked list appropriately
    `tail` and remove one level of indirection. Reading code that is using
    pointers to pointers is unnecessarily hard.

Thanks to clang's scan-build for pointing this out!

### DIFF
--- a/src/perl.c
+++ b/src/perl.c
@@ -462,16 +462,15 @@ static int av2data_set(pTHX_ AV *array, char *name, data_set_t *ds) {
  *   meta     => [ { name => <name>, value => <value> }, ... ]
  * }
  */
-static int av2notification_meta(pTHX_ AV *array, notification_meta_t **meta) {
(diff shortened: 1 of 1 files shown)


rubenk:
I've been staring for ages at that code before, trying to fix the scan-build report, so thanks for showing me how it's done!

PR #2279, opened 2017-05-15
virt plugin: Replace virTypedParamsFree() with virTypedParamsClear() …

octo:
…and `sfree()`.

The parameters are allocated with `calloc()`, freeing them with a library function is unintuitive at best. Also, the library functions appears to be missing the appropriate function attribute to tell clang / GCC that
it will free the pointer, resulting in a false-positive scan-build report.


rubenk:
Thanks @octo!

Issue #2266, opened 2017-05-06
Array indexes mix-up in curl_json plugin


madchrist:
*   Version of collectd: 5.6.2
*   Operating system / distribution: gentoo

## Bug description
I'm trying to fetch two values from a json array with curl_json
I want "datapoints/0/1" and "datapoints/0/2"
"datapoints/0/1" works normaly but "datapoints/0/2" will have the value of "datapoints/0/3"
Also, the RRD for "datapoints/0/1" seems to be used twice in one interval, i'm guessing "datapoint/0/2" is somehow added to this RRD but is rejected because too recent

## remote json content
`{"datapoints":[[10,11,12,13,14,15,16]]}`

## collectd.conf extract
```
LoadPlugin rrdtool
LoadPlugin curl_json
<Plugin curl_json>
    <URL "http://madx.org/collectd.json">
        Host "test"
        Instance "bug"
                  
        <Key "datapoints/0/1">
            Type "gauge" 
            Instance "eleven"
        </Key>

        <Key "datapoints/0/2">
            Type "gauge"
            Instance "twelve"
        </Key>
    </URL>
</Plugin>
```

## syslog output
```
May  6 15:33:04 baldur collectd[13470]: plugin_load: plugin "syslog" successfully loaded.
May  6 15:33:04 baldur collectd[13471]: cpufreq plugin: Found 4 CPUs
May  6 15:33:04 baldur collectd[13471]: Initialization complete, entering read-loop.
May  6 15:33:04 baldur collectd[13471]: rrdtool plugin: Adjusting "RandomTimeout" to 0.000 seconds.
May  6 15:33:04 baldur collectd[13471]: nut plugin: Connection to (localhost, 3493) established.
May  6 15:33:05 baldur collectd[13471]: cu_rrd_create_file: File "test/curl_json-bug/gauge-eleven.rrd" is already being created.
May  6 15:33:05 baldur collectd[13471]: Filter subsystem: Built-in target `write': Dispatching value to all write plugins failed with status -1.
May  6 15:33:05 baldur collectd[13471]: Filter subsystem: Built-in target `write': Some write plugin is back to normal operation. `write' succeeded.
May  6 15:33:15 baldur collectd[13471]: rrdtool plugin: rrd_update_r (test/curl_json-bug/gauge-eleven.rrd) failed: test/curl_json-bug/gauge-eleven.rrd: illegal attempt to update using time 1494077595 when last update time is 1494077595 (minimum one second step)
[...]

PR #2281, opened 2017-05-16
curl_json plugin: Fix array index and key handling.

octo:
Previously, the "key" was loaded by calling `cj_cb_map_key()` from `cj_cb_inc_array_index()`. That means that the key for the previous element was loaded as the array index was updated for the next element, resulting in an off-by-one error. Also the key was not unset in time, resulting in two metrics with the same identifier being created.

This patch fixes this with the following changes:

*   `cj_advance_array()` (nee `cj_cb_inc_array_index()`) now loads the key for the new index position instead of the previous one.
*   The initial "0" key is loaded from `cj_cb_start_array()`.
*   `cj_advance_array()` always updates the key. The "update_key" argument has been removed.
*   Refactoring: key loading has been moved out of `cj_cb_map_key()` and into its own function, `cj_load_key()`.
*   The unit tests have been expanded to cover this case.

Fixes: #2266


octo:
TODO(octo): rebase once #2283 is merged; merge when successful.

Possibly separate fix and tests so that we can (more easily) cherry-pick the fix into *collectd-5.6*.

PR #2285, opened 2017-05-16
curl_json plugin: Refactor the way trees/keys are stored.

octo:
Previously, keys had a "magic" as their first member which was used to differentiate between the two types when they were returned from the binary search tree.

This patch creates a new struct, cj_tree_entry_t, which includes an enum identifying which union member is valid.

I have to admit I'm a bit on the fence on this one. It feels more "correct" but unfortunately isn't as readable as I'd hope it to be. Thoughts?

### DIFF
--- a/src/curl_json.c
+++ b/src/curl_json.c
@@ -946,9 +946,13 @@ static int cj_read(user_data_t *ud) /* {{{ */
 
   db->depth = 0;
   memset(&db->state, 0, sizeof(db->state));
-  db->state[0].entry = &(cj_tree_entry_t){
-      .type = TREE, .tree = db->tree,
-  };
+
+  /* This is not a compound literal because EPEL6's GCC is not cool enough to
+   * handle anonymous unions within compound literals. */
+  cj_tree_entry_t root = {0};
+  root.type = TREE;
+  root.tree = db->tree;
+  db->state[0].entry = &root;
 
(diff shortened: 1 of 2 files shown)


octo:
EPEL6 fails with:

```
src/curl_json.c: In function 'cj_read':
src/curl_json.c:950: error: unknown field 'tree' specified in initializer
src/curl_json.c:950: error: missing braces around initializer
src/curl_json.c:950: error: (near initialization for '(anonymous).<anonymous>')
make[1]: *** [src/curl_json_la-curl_json.lo] Error 1
```

Old compilers suck. :\

octo:
My guess is that this is because `tree` is inside an anonymous union:

```c
typedef struct {
  enum { KEY, TREE } type;
  union {
    c_avl_tree_t *tree; /* <--- */
    cj_key_t *key;
  };
} cj_tree_entry_t;
```

octo:
So … if I understand the internet correctly, anonymous unions were a GNU extension and as such disabled with `-std=c99`. `-std=gnu99`, as used by the CI, *should* allow this. Since C11, anonymous unions are officially specified.

The anonymous union is used all over the place, though, and this is the only complaint by the compiler, so my money is (now) on anonymous union (not part of c99) and compound literal (new in c99) don't sit well with this old piece of s…oftware.

PR #2287, opened 2017-05-17
fix build on automake 1.11

BrandonArp:
This change fixes builds using automake 1.11.  

As far as I can tell, automake 1.12 changed the java timestamp files from classdist_noinst.stamp to classnoinst.stamp.  This change adds a version check to configure.ac to set a variable (JAVA_TIMESTAMP_FILE) that is used in Makefile.am to select which timestamp file is used.  In order to do the version checking, I used AX_COMPARE_VERSION documented at https://www.gnu.org/software/autoconf-archive/ax_compare_version.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -11,6 +11,8 @@ dnl we don't really need the 'u' even in older toolchains.  Then there is
 dnl older libtool, which spelled it AR_FLAGS
 m4_divert_text([DEFAULTS], [: "${ARFLAGS=cr} ${AR_FLAGS=cr}"])
 
+m4_include([m4/ax_compare_version.m4])
+
 LT_INIT([dlopen disable-static])
 
 AM_INIT_AUTOMAKE([subdir-objects tar-pax dist-bzip2 no-dist-gzip foreign])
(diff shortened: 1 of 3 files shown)


rubenk:
Hi Brandon,

What distro do you use that still uses automake 1.11?

BrandonArp:
Hi Ruben,

I'm building on CentOS 6.  Apparently the latest version in the default repos is 1.11.1
From a docker container:

[root@ca7a8d45828d /]# cat /etc/redhat-release
CentOS release 6.9 (Final)
[root@ca7a8d45828d /]# automake --version
automake (GNU automake) 1.11.1

configure.ac
```diff
@@ -10,6 +10,7 @@ dnl older automake's default of ARFLAGS=cru is noisy on newer binutils;
 dnl we don't really need the 'u' even in older toolchains.  Then there is
 dnl older libtool, which spelled it AR_FLAGS
 m4_divert_text([DEFAULTS], [: "${ARFLAGS=cr} ${AR_FLAGS=cr}"])
+m4_include(m4/ax_compare_version.m4)
```
rubenk: Please quote this, so `m4_include([m4/ax_compare_version.m4])`.
Can you also add a newline above it to make it clear that this code has nothing to do with libtool?

configure.ac
```diff
@@ -2861,6 +2862,7 @@ AC_ARG_WITH([java],
   [with_java="yes"]
 )
 
+AX_COMPARE_VERSION([$am__api_version],[lt],[1.12],[JAVA_TIMESTAMP_FILE="classdist_noinst.stamp"], [JAVA_TIMESTAMP_FILE="classnoinst.stamp"])
```
rubenk: Can you break this line up?

rubenk:
Hi Brandon,

Thanks for letting me know. We never hit this in CI because we build from tarball there.
I've left two minor nits. If you fix them, I'll merge.

BrandonArp:
I put the conditional, if-true block and if-false block on different lines.  How does that look?

rubenk:
Perfect, thank you.

PR #2290, opened 2017-05-18
allow configuration of snmp_agent in rpm builds

BrandonArp:
This plumbs the configuration for the snmp_agent plugin through the RPM spec.


octo:
LGTM, thanks @BrandonArp!

PR #2295, opened 2017-05-19
snmp plugin: Avoid allocation of temporary buffers on the heap.

octo:


src/snmp.c
```diff
@@ -1247,16 +1240,17 @@ static int csnmp_dispatch_table(host_definition_t *host,
...
       instance_list_ptr = instance_list_ptr->next;
     else
       value_table_ptr[0] = value_table_ptr[0]->next;
   } /* while (have_more) */
 
-  sfree(vl.values);
-  sfree(value_table_ptr);
-
-  return 0;
+  return (0);
```
rubenk: return 0 ;)

octo: Oh sorry, that's likely the result of merge conflict resolution.

PR #2300, opened 2017-05-26
Remove close() call from sread()

rpv-tomsk:
Function sread() behaves incorrectly and unexpectedly in case of EOF - it closes FD. 
At the same time, in case of socket errors it does not close FD.

Function sread() used in two modules: zone.c and apcups.c.
Both of them do `close(fd)` when sread() return value is not equal to zero - i.e in both cases of socket error and socket EOF. Duplicated `close()` call occurs.

I think *read* function should not close fd in any cases, and other developers expects the same, I think.

This patch is proposed to fix this.

The `swrite()` function is ok, no similar fix needed.

src/apcups.c:

```
if (sread(*sockfd, (void *)&packet_size, sizeof(packet_size)) != 0) {
    close(*sockfd);
    *sockfd = -1;
    return -1;
  }
...

  if (sread(*sockfd, (void *)buf, packet_size) != 0) {
    close(*sockfd);
    *sockfd = -1;
    return -1;
  }
```

src/zone.c

```
  if (sread(fd, buf, bufsize) != 0) {
    char errbuf[1024];
    ERROR("zone plugin: Reading \"%s\" failed: %s", procfile,
          sstrerror(errno, errbuf, sizeof(errbuf)));
    close(fd);
    return 1;
  }
```

### DIFF
--- a/src/daemon/common.c
+++ b/src/daemon/common.c
@@ -230,10 +230,7 @@ ssize_t sread(int fd, void *buf, size_t count) {
       return status;
 
     if (status == 0) {
-      DEBUG("Received EOF from fd %i. "
-            "Closing fd and returning error.",
-            fd);
-      close(fd);
+      DEBUG("Received EOF from fd %i. ", fd);
       return -1;
     }
 
--- a/src/daemon/common.c
+++ b/src/daemon/common.c
@@ -212,7 +212,7 @@ void sfree (void **ptr)
 }
 #endif
 
-ssize_t sread(int fd, void *buf, size_t count) {
+int sread(int fd, void *buf, size_t count) {
   char *ptr;
   size_t nleft;
   ssize_t status;
@@ -243,7 +243,7 @@ ssize_t sread(int fd, void *buf, size_t count) {
   return 0;
 }
 
-ssize_t swrite(int fd, const void *buf, size_t count) {
+int swrite(int fd, const void *buf, size_t count) {
   const char *ptr;
   size_t nleft;
   ssize_t status;
--- a/src/daemon/common.h
+++ b/src/daemon/common.h
@@ -79,8 +79,7 @@ char *sstrerror(int errnum, char *buf, size_t buflen);
  *
  * DESCRIPTION
  *   Reads exactly `n' bytes or fails. Syntax and other behavior is analogous
- *   to `read(2)'. If EOF is received the file descriptor is closed and an
- *   error is returned.
+ *   to `read(2)'.
  *
  * PARAMETERS
  *   `fd'          File descriptor to write to.
--- a/src/daemon/common.h
+++ b/src/daemon/common.h
@@ -90,7 +90,7 @@ char *sstrerror(int errnum, char *buf, size_t buflen);
  *   Zero upon success or non-zero if an error occurred. `errno' is set in this
  *   case.
  */
-ssize_t sread(int fd, void *buf, size_t count);
+int sread(int fd, void *buf, size_t count);
 
 /*
  * NAME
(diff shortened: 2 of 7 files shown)


rpv-tomsk:
CC @rubenk, ping.

rubenk:
I'd expect sread to behave similar to read(2), so this patch seems a good step in the right direction.

As you rightly noticed you can't distinguish between a socket error and EOF. Returning 0 on EOF would make sense. You'll have to adjust the callers then too. What do you think?

rpv-tomsk:
I.e. the `sread()` should return 0 at EOF, `count` if read succeeded and full buffer received, and value less than 0 in case of error?

Then the `swrite()` should be touched in the same way, and there are some places exists where `sread()` can be used instead of similar local implementation inside plugin.

rubenk:
Yeah sounds good 👍

rpv-tomsk:
After some digging, I propose to discuss (and cancel) this proposed change. 

The `sread()` is used as a function which intended to read _fixed amount of data_.
If that data is not available for any reason (EOF, error) - for plugin there is no difference in the cause of this.

The `swrite()` does and behaves almost the same - it used as a function which intended to write _fixed amount of data_, and there is no difference in the causes of write failure.

If plugin can handle EOF, then it needs to get the amount of data actually readed/written, which this change does not provide.

If we made amount of data readed/written available from these functions, then these will be completely another functions: functions, which intended to read/write data as much as possible. That does not required by existing plugins.


Also, there is a significant difference will appear: `swrite()` returns _positive_ value in case of error (value, obtained from `errno`), which will conflict with the number of bytes actually written.


Collectd code widely uses zero value as success, and non-zero otherwise. These functions follows this rule.

I propose do not change this - these changes are unneeded in the current plugins and I do not see any useful changes for other plugins too.

A function, which reads data into buffer until EOF may be added instead. It can be used, for example, in `src/zookeeper.c:157`, but I found only this one place where it can be used, so also no any code added/changed.

rpv-tomsk:
@rubenk, what do you think about this?

rubenk:
I agree with your reasoning, since these functions handle short reads and writes already, there's no need for the callers to handle EOF.

rpv-tomsk:
Let's merge?

octo:
The intention for having the `sread()` and `swrite()` functions is so the callers don't need to worry about partial reads and writes. In that context, hitting EOF before the requested amount of bytes are read is an *error*. I think there are many plugins that rely on this behavior.

rpv-tomsk:
>  In that context, hitting EOF before the requested amount of bytes are read is an error. I think there are many plugins that rely on this behavior.

There are only two plugins which uses `sread()`. Both of them do `close()` in case of error returned.
The `sread()` behaviour does not changed by this patch. In case of EOF `sread()` still returns an error.

The only change is `close()` was removed  from inside of `sread()`. 
There should be no `close()`  there, it is read function.

And again: There are only two plugins which uses `sread()`. Both of them do `close()` in case of error returned. There will be two `close()` calls in case of EOF in version without this patch.

rpv-tomsk:
Hi!
Is there are any other questions about this patch, or it may be merged?

rpv-tomsk:
Added two another commits, and updated the first one.

rpv-tomsk:
PR rebased.
I want to notice one important comment:

>  In that context, hitting EOF before the requested amount of bytes are read is an error. I think there are many plugins that rely on this behavior.

The `sread()` behaviour does not changed by this patch. In case of EOF `sread()` still returns an error.

The only change is `close()` was removed  from inside of `sread()`. 
There should be no `close()`  there, it is read function.

There are only two plugins which uses `sread()`. Both of them do `close()` in case of error returned. There will be two `close()` calls in case of EOF in version without this patch.

PR #2301, opened 2017-05-26
apcups: Fix reconnect implementation

rpv-tomsk:
Implementation was broken, it doesn't reconnects as was planned.


rpv-tomsk:
CC @rubenk, ping.

Issue #2291, opened 2017-05-18
snmp plugin: double free or heap corruption 


cullorblind:
*   Version of collectd: 5.7.0-3~bpo8+1 from debian jessie backports repo
*   Operating system / distribution: debian 8.8 jessie

## Expected behavior
I am polling connected clients across about 150 ubiquiti access points from the SNMP Station table.   A few of the Access points have 0 clients, and therefore return nothing.   If the instance string returns nothing it should gracefully log(optional) and move on to the next task.

## Actual behavior
After much testing I narrowed it down to a single access point which I realized had no clients attached.  With only that access point configured I get the double free or corruption error.
 collectd -T
*** Error in `collectd': double free or corruption (!prev): 0x000000000236dd10 ***
Aborted

If the instance OID returns no data it dies here.

## Configs to reproduce
**my_types.db**
```ubnt_rm5_sta            signal:GAUGE:U:U, noisefloor:GAUGE:U:U, ccq:GAUGE:U:U, amq:GAUGE:U:U, amc:GAUGE:U:U```

**snmp-rm5.conf**
```
LoadPlugin snmp
<Plugin snmp>
    <Data "stations">
      Type "ubnt_rm5_sta"
      Table true
      Instance ".1.3.6.1.4.1.41112.1.4.7.1.2"
      Values ".1.3.6.1.4.1.41112.1.4.7.1.3" ".1.3.6.1.4.1.41112.1.4.7.1.4"  ".1.3.6.1.4.1.41112.1.4.7.1.6" ".1.3.6.1.4.1.41112.1.4.7.1.8" ".1.3.6.1.4.1.41112.1.4.7.1.9"
    </Data>

   <Host "ubnt-rm5-allstate-N-120">
       Address "172.16.16.253"
       Version 1
       Community "public"
       Collect "stations"
   </Host>
</Plugin>
```

Extra info from the mib for what I'm polling.
```
snmptranslate -m /root/MIBs/UBNT-MIB.txt -Tz | grep ubntSta
"ubntStaTable"                  "1.3.6.1.4.1.41112.1.4.7"
"ubntStaEntry"                  "1.3.6.1.4.1.41112.1.4.7.1"
"ubntStaMac"                    "1.3.6.1.4.1.41112.1.4.7.1.1"
"ubntStaName"                   "1.3.6.1.4.1.41112.1.4.7.1.2"
"ubntStaSignal"                 "1.3.6.1.4.1.41112.1.4.7.1.3"
[...]

PR #2309, opened 2017-06-01
snmp plugin: Fix double free / handle missing variables

rpv-tomsk:
Issue: #2291
This PR is against 5.7.

### DIFF
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -1357,11 +1357,13 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
     if (oid_list_todo_num == 0) {
       /* The request is still empty - so we are finished */
       DEBUG("snmp plugin: all variables have left their subtree");
+      snmp_free_pdu(req);
       status = 0;
       break;
     }
 
     res = NULL;
+    /* snmp_sess_synch_response always frees our req PDU */
     status = snmp_sess_synch_response(host->sess_handle, req, &res);
     if ((status != STAT_SUCCESS) || (res == NULL)) {
       char *errstr = NULL;
@@ -1376,8 +1378,6 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
         snmp_free_pdu(res);
       res = NULL;
 
-      /* snmp_synch_response already freed our PDU */
-      req = NULL;
       sfree(errstr);
       csnmp_host_close_session(host);
 
@@ -1492,9 +1492,6 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
     snmp_free_pdu(res);
   res = NULL;
 
-  if (req != NULL)
-    snmp_free_pdu(req);
-  req = NULL;
 
   if (status == 0)
     csnmp_dispatch_table(host, data, instance_list_head, value_list_head);
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -1363,8 +1363,11 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
     }
 
     res = NULL;
-    /* snmp_sess_synch_response always frees our req PDU */
     status = snmp_sess_synch_response(host->sess_handle, req, &res);
+
+    /* snmp_sess_synch_response always frees our req PDU */
+    req = NULL;
+
     if ((status != STAT_SUCCESS) || (res == NULL)) {
       char *errstr = NULL;
 


src/snmp.c
```diff
@@ -1357,11 +1357,13 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
     if (oid_list_todo_num == 0) {
       /* The request is still empty - so we are finished */
       DEBUG("snmp plugin: all variables have left their subtree");
+      snmp_free_pdu(req);
       status = 0;
       break;
     }
 
     res = NULL;
+    /* snmp_sess_synch_response always frees our req PDU */
```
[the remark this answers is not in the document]
rpv-tomsk: >I think that req is (sometimes) freed when the return status is STAT_ERROR, otherwise we need to 
>free it ourselves.

Ok, we see some corner case when req is not freed when the return status from `snmp_sess_synch_response` is STAT_ERROR. That looks like error in net-snmp, and that practically should't happen, because session is opened.

Unpatched code of Collectd does not make a `snmp_free_pdu()` call when `status == STAT_ERROR`, so no regression added here by mine patch.

There are two calls of `snmp_sess_synch_response`

1) In `csnmp_read_table()`. 

Note what error handling code has comment "snmp_synch_response already freed our PDU".

```
  while (status == 0) {

    ...

    status = snmp_sess_synch_response(host->sess_handle, req, &res);
    if ((status != STAT_SUCCESS) || (res == NULL)) {
      ...
      if (res != NULL)
        snmp_free_pdu(res);
      res = NULL;

      /* snmp_synch_response already freed our PDU */
      req = NULL;        <<<---------------------
      ...

      status = -1;
      break;
    }
    ...
  }
....
  if (req != NULL)
    snmp_free_pdu(req);
```
No `snmp_free_pdu(req)` will be called in case of `snmp_sess_synch_response` errors.

2) In `csnmp_read_value()`. 

That implementation uses the same `snmp_pdu_create() + snmp_sess_synch_response()` chain of calls, and there is no `snmp_free_pdu(req)` at all.

rpv-tomsk:
>I'm a bit hesitant to merge this, because it will leak req during normal operation, which is clearly a regression.

The https://github.com/collectd/collectd/issues/2291#issuecomment-304523206 comment has detailed explanation of issue causes.

The `snmp_free_pdu(req);` which was in previous code after `while` loop (at line 1496), now placed in line 1360. During normal operation it does the same work as before.
There is no [new] memory leak in proposed code.

rpv-tomsk:
Hi! When you will have a time to continue this review?
I think there is nothing suspicious in this patch, and I'm ready to provide proofs,  explanations and answers to reviewer questions.

src/snmp.c
```diff
@@ -1357,11 +1357,13 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
...
     status = snmp_sess_synch_response(host->sess_handle, req, &res);
```
octo: Can you please move the above comment to here and add:

```c
/* snmp_sess_synch_response always frees our req PDU */
req = NULL;
```

src/snmp.c
```diff
@@ -1397,6 +1397,38 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
       break;
     }
 
+    if (res->errstat != SNMP_ERR_NOERROR) {
+      if (res->errindex != 0 && res->errindex < oid_list_len) {
```
octo: This if-block is unnecessarily long, please invert this logic keep things short:

```c
if ((res->errindex == 0) || (res->errindex >= oid_list_len)) {
  ERROR("snmp plugin: host %s; data %s: response error: %s (%li) ",
        host->name, data->name, snmp_errstring(res->errstat), res->errstat);
  status = -1;
  break;
}
/* previous if-block. */
```

octo: Is `errindex` one-based? If so, the second condition should be `res->errindex <= oid_list_len`. (or `res->errindex > oid_list_len` if you take my previous comment into account).

rpv-tomsk: Yes, `errindex` is one-based. 

Example:
https://github.com/hardaker/net-snmp/blob/a7bc508a8930a484c3a666cbea4ab226d2a3aa88/apps/snmpget.c#L210

rpv-tomsk:
Previously published implementation of `res->errstat` contains an error/mistake.

After 4 hours of looking into code, I decided to split these to two patches:
'bugfix' and 'feature'.  Bugfix should go to 5.7, and the 'feature' should be reviewed again.

Thanks for review ;-)

rpv-tomsk:
>Alright, you have me convinced that 365af0c is sane. 

Moreover, I needed to convince myself. :-)
Details have been lost since the time.

>Next time, I'd prefer to get these changes in two PRs to keep each PR focused on one issue.

It was impossible to work on the second part unless 'double free' gets resolved. So, all the changes was done in same branch.

octo:
Cherry-picked into *collectd-5.6* as d16c24542b2f96a194d43a73c2e5778822b9cb47
Cherry-picked into *collectd-5.6* as a71c3ebcfc6ca265c702117c89f1bb552723d1fa

PR #2310, opened 2017-06-05
intel_rdt plugin: Remove unnecessary goto.

octo:


rubenk:
LGTM.
Might be nice to remove the other goto's as well. For one thing I think the label order reads backwards, and the cleanup logic is not that complicated that it needs a goto.

PR #2314, opened 2017-06-13
Ipvs ipv6

Bencabot:
Added support for IPV6 services. Using libnl1, libnl3 and libnl-genl (http://www.infradead.org/~tgr/libnl/) to query IPV6 ipvs services through netlink. Different distributions include different version of libnl by default so this covers Debian / CentOS bases. 
A new header has been included which contain the necessary structures for IPV6 which are not included in the standard kernel headers.
This work is based on the IPV6 updates made to ipvsadm by Vince Busam and Julius Volz.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -611,6 +611,24 @@ else
 fi
 
 AM_CONDITIONAL([IP_VS_H_NEEDS_KERNEL_CFLAGS], [test "x$ip_vs_h_needs_kernel_cflags" = "xyes"])
+PKG_CHECK_MODULES(LIBNL_GENL3, libnl-genl-3.0, [have_libnl_genl3=yes], [have_libnl_genl3=no])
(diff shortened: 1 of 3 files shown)


rubenk:
Dear @Bencabot, thank you for your contribution.

Before you spend time rebasing this, I'd first like to ask why you need to include this header.
From a cursory glance it looks like it's a copy of linux/ipvs.h from the kernel, with perhaps 20% changed. Keeping this in sync with the kernel will be a nightmare. Do you have any idea why these ipv6 structures are not included in the kernel headers?

rubenk:
> Another option could be remove the ipv4 only code and force the use of netlink. Looking at http://www.infradead.org/~tgr/libnl/ it seems to show that the main distributions have at least Netlink 1.1 packaged. This will be cleaner but with the potential to stop current users updating. How do you think it is best to proceed?

I'd even go as far as to require at least libnl3. As far as I can tell all the major distributions we support have that.

Issue #2302, opened 2017-05-26
Collectd SNMP : custom snmp port


gugansankar:
Version : 
collectd-5.7.1-1.el7.x86_64

OS:
CentOS 7 64bit

Seems there is no option to add custom snmp port in the Host configuration section. Is there any workaround solution for custom snmp port ?.

as far as i am concerned, custom port field is necessary one.

Expected entry:

```
<Host "some.switch.mydomain.org">
  Address "192.168.0.2"
  Version 1
  Community "community_string"
  Collect "std_traffic"
  Interval 120
  Port 1661
</Host>
```

rubenk:
There's currently no support in the code to set a custom port. Why do you feel running snmp on a custom port is necessary?

gugansankar:
as most of the organisation trying to run the service on custom port instead of using it on default one for some security reason.

we've more than 50,000 devices running across world wide. all are running with custom snmp port.

Even all the network monitoring tools does provide custom snmp port add option. Why don't we have this option in collectd+snmp ?. Is there any specific reason ?

rubenk:
Moving SNMPv1 to a different port doesn't offer you any extra security. SNMPv1 still sends community strings in plain text.

My guess is this has never been implemented because nobody ever needed it, and it offers questionable benefits.

I'm not opposed to accepting a patch which adds support for this however.

gugansankar:
Nope, I'm not using SNMP v1 in our organisation. Using only v2 and v3. I just pasted the above configuration from Collectd snmp sample entry docs :) .

its good to have a support for custom port :+1:

rubenk:
Ah ok. Will you be able to send a PR? I'm afraid that otherwise this issue will stay open indefinitely.

gugansankar:
Sorry , I'm not a programmer. Any timeline for this feature ?.

rubenk:
If this is important to you or your organisation, you'll have to learn how to program, or find someone who can. This is open source software, and features are not generally implemented on request, I'm sorry.

toni-moreno:
Hi @gugansankar  , if you would like to test other tools, We have built a snmp collecting tool to get data to put into influxdb database, and you can configure any snmp related configuration per device. You have also a nice web user interface to configure or via REST api if you prefer.

https://github.com/toni-moreno/snmpcollector/wiki/Component:-SNMP-Devices

Feel free to test and give me feedback about your experience.

gugansankar:
Sure thanks, will try this one.

[...]

PR #2327, opened 2017-06-25
snmp plugin: Option `Address` documented in more details.

rpv-tomsk:
As described at http://net-snmp.sourceforge.net/dev/agent/structsnmp__session.html,  `peername` field of `snmp_session` structure, and, as result, `Address` option of Collectd `snmp` plugin configuration, may include transport specifier and/or port number. 

Examples of transport specifier are here: http://net-snmp.sourceforge.net/docs/man/snmpcmd.html

I haven't check how it works with TCP transport, but at least it tries to connect by TCP.

Configuration examples are updated too.

Closes: #2302


rubenk:
Thanks @rpv-tomsk!

PR #2331, opened 2017-06-28
openldap: check ld structure before passing it to ldap_unbind()

mfournier:
This prevents collectd from segfaulting when the ldap session setup
fails before opening a connection to openldap (syntax error in the URL
option for example).


rubenk:
Thanks Marc!

PR #2334, opened 2017-06-29
Fix errno handling, and getgrnam_r on OpenBSD

danieljakots:
Hi,

I read Guidelines.md, I acknowledge that it says
`Focus: Fix one thing in your PR. The smaller your change, the faster it will be reviewed and merged.`

This PR fixes two things but they're really related so hopefully you won't get (too much) mad at me :3

Anyway here's my story. I began to use the plugin exec stating:
<Plugin exec>
        Exec "_collectd:_collectd" "/usr/local/some/really/nice/thing"
</Plugin>
but it said

> collectd[32946]: exec plugin: Failed to get group information for group ``_collectd'': Undefined error:

which is not really helpful, so I asked help from a friend who came with first commit. Then it said 

> collectd[55462]: exec plugin: Failed to get group information for group ``_collectd'': Result too large

So I bumped the size and it worked fine. So I fixed the similar results grep would give me in the whole src/. It's already committed downstream (i.e. OpenBSD): https://marc.info/?l=openbsd-ports-cvs&m=149869644612265&w=2

edit: added a third commit which does the same than the first.
edit: and a fourth now, which does the same than the first, sorry :D

Cheers,
Daniel

### DIFF
--- a/src/email.c
+++ b/src/email.c
@@ -411,7 +411,7 @@ static void *open_connection(void __attribute__((unused)) * arg) {
     if (status != 0) {
       char errbuf[1024];
       log_warn("getgrnam_r (%s) failed: %s", group,
-               sstrerror(errno, errbuf, sizeof(errbuf)));
+               sstrerror(status, errbuf, sizeof(errbuf)));
     } else if (grp == NULL) {
       log_warn("No such group: `%s'", group);
     } else {
--- a/src/email.c
+++ b/src/email.c
@@ -403,7 +403,7 @@ static void *open_connection(void __attribute__((unused)) * arg) {
   {
     struct group sg;
     struct group *grp;
-    char grbuf[2048];
+    char grbuf[4096];
     int status;
 
     grp = NULL;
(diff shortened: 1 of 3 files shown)


rubenk:
First, thanks for your contribution!

While just doubling the size of the buffer works, I'd prefer if we could use `sysconf(_SC_GETGR_R_SIZE_MAX)` to determine the size of the buffer, as documented in [getgrent(3)](https://man.openbsd.org/getgrent).

rubenk:
I took your patches for error handling in 71f65087f9d5f42f8944fa7ce416cb0e2f7f043e, and for the buffer size in 24bdf52decaf93359aab0364fb11420af5cd4a21.

danieljakots:
Yeah Jérémie told me about the sysconf way but it was too complicated for me (I'm not a C programmer).
Thanks for taking care of this!

rubenk:
No problem, 4096 bytes should be enough for everyone ;)

You're welcome!

Issue #2328, opened 2017-06-26
curl plugins: should not store redirect target as new request url


bzed:
*   Version of collectd: 5.7.0-3~bpo8+1
*   Operating system / distribution: jessie

## Expected behavior

The curl_json (and I assume other curl based plugins) should not save the target of a redirect as new URL to send all coming requests to, but instead use the configured URL (and follow a redirect again, if there is one...).

## Actual behavior
With the curl_json plugin in this case:

```
Everything correct so far:

1.2.3.4 - - [19/Jun/2017:20:51:33 +0200] "GET /internal/state/health HTTP/1.1" 200 122 "-" "collectd/5.7.0.git"
1.2.3.4 - - [19/Jun/2017:20:51:43 +0200] "GET /internal/state/health HTTP/1.1" 200 121 "-" "collectd/5.7.0.git"

oops, internal application error happened, redirecting to error tracking page:
1.2.3.4 - - [19/Jun/2017:20:51:53 +0200] "GET /internal/state/health HTTP/1.1" 303 - "-" "collectd/5.7.0.git"
1.2.3.4 - - [19/Jun/2017:20:51:53 +0200] "GET /error/;jsessionid=18E435DB7F065A4EA79E427F2E97E1DF.hostname?errorId=915f6d68-4bf8-4f4f-9f42-cb57cc1096cc HTTP/1.1" 302 - "-" "collectd/5.7.0.git"

WTF? Why does collectd keep retrieving that error page instead of the configured URL???
1.2.3.4 - - [19/Jun/2017:20:51:53 +0200] "GET /error/;jsessionid=18E435DB7F065A4EA79E427F2E97E1DF.hostname?errorId=915f6d68-4bf8-4f4f-9f42-cb57cc1096cc HTTP/1.1" 200 23435 "-" "collectd/5.7.0.git"
1.2.3.4 - - [19/Jun/2017:20:52:13 +0200] "GET /error/;jsessionid=18E435DB7F065A4EA79E427F2E97E1DF.hostname?errorId=915f6d68-4bf8-4f4f-9f42-cb57cc1096cc HTTP/1.1" 200 15621 "-" "collectd/5.7.0.git"
[...]

PR #2341, opened 2017-07-04
curl plugins: Use configured URL for all poll cycles

rpv-tomsk:
Issue: #2328

After redirect received, Collectd send subsequent requests to new location.
That is wrong - Collectd should use configured URL for all poll cycles, regardless of the responses received previously.

Problem was caused by https://github.com/curl/curl/issues/1631. To avoid that when old `libcurl` version used, we set request url in each poll cycle.

### DIFF
--- a/src/apache.c
+++ b/src/apache.c
@@ -328,7 +328,6 @@ static int init_host(apache_t *st) /* {{{ */
 #endif
   }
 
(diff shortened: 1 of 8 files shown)


rubenk:
thanks @rpv-tomsk!

Issue #2330, opened 2017-06-26
Python plugin ignores read callback return value


danielkza:
*   Version of collectd:

5.7

*   Operating system / distribution:

Ubuntu 16.04

## Expected behavior

Explicitly returning a non-zero integer value from a Python read callback should indicate failure, as it would with a C callback.

My use case is a bit peculiar. I'm attempting to monitor Apache Spark. At any given time, there may be zero or more driver programs running in a worker node in different ports, but I do not have an easy way to acquire that list of ports.

Hence, I try all ports in a predefined list. Hence, read callback can fail most of the time if a driver is not running in a given port. I want to be able to signal failure, such that the interval doubling kicks in, without necessarily printing a stack trace.

## Actual behavior

Failure is only signaled if an exception is raised.

## Steps to reproduce

Run the following plugin, and observe how the callback interval never raises, as it would be expected from failures.

```python

import collectd

def read_fail():
    return 1

collectd.register_read(read_fail)
```

trenkel:
Any specific reason why the current implementation of raising an exception doesn't work for you?

danielkza:
Only because it clogs the logs with exception stack traces, and makes it much harder to distinguish between an expected failure and a real problem.

trenkel:
Makes sense. Still, I don't like the idea of a return value. How about a special exception class (collectd.Error or something like that) that you can raise which will never print a stacktrace and just a single log line instead. Would that work for you?

danielkza:
Sure, any way to signal "expected" failure will work just fine.

Em sex, 30 de jun de 2017 18:16, trenkel <notifications@github.com>
escreveu:

rpv-tomsk:
>Still, I don't like the idea of a return value.

Can you please explain to those interested in,  why you don't like the idea of a return value?
That it is not accepted code style in Python, or smth else?

Thanks.

rpv-tomsk:
>Still, I don't like the idea of a return value.

@trenkel,  Can you please explain to those interested in,  why you don't like the idea of a return value?
That it is not accepted code style in Python, or smth else?

Thanks.

trenkel:
Has there ever been a Python function that indicated an error condition by returning a certain value as opposed to throwing an exception? Probably somewhere, but you'd have to search really hard to find it.

rpv-tomsk:
Thanks for explanation, @trenkel !
I got it.

octo:
[...]

PR #2346, opened 2017-07-05
Add CollectdError exception which can be thrown without causing a sta…

trenkel:
This should address #2330.

### DIFF
--- a/src/collectd-python.pod
+++ b/src/collectd-python.pod
@@ -68,8 +68,10 @@ use multiple B<ModulePath> lines to add more than one directory.
 If a Python script throws an exception it will be logged by collectd with the
 name of the exception and the message. If you set this option to true it will
 also log the full stacktrace just like the default output of an interactive
-Python interpreter. This should probably be set to false most of the time but
-is very useful for development and debugging of new modules.
+Python interpreter. This does not apply to the CollectError exception, which
+will never log a stacktrace.
+This should probably be set to false most of the time but is very useful for
+development and debugging of new modules.
 
 =item B<Interactive> I<bool>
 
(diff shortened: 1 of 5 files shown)


rubenk:
Thanks Sven, this looks good. Could you add a short description about this to collectd-python(5) as well?

src/python.c
```diff
@@ -318,11 +325,17 @@ void cpy_log_exception(const char *context) {
     typename = "NamelessException";
   if (message == NULL)
     message = "N/A";
-  Py_BEGIN_ALLOW_THREADS ERROR("Unhandled python exception in %s: %s: %s",
```
octo: This unfortunate formatting is very likely clang-format's fault. Can we add a semi-colon after the macro to put a stop to this?

trenkel: I'll add those if it happens again.

trenkel:
Done.

src/collectd-python.pod
```diff
@@ -248,6 +250,18 @@ collectd you're done.
...
+=head2 CollectdError
+
+This is an exception. If any Python script raises this exception it will
+still be treated like an error by collectd but it will be logged as a
+warning instead of an error and it will never generate a stacktrace.
+
+ class CollectdError(Exception)
+
+Basic exception for collectd Python scripts.
+Throwing this exception will not cause a stacktrace to be logged, even if
```
octo: Is it possible to create a custom exception which is a subclass of this class?

src/python.c
rpv-tomsk: I'll add these semi-colons...

rpv-tomsk:
Merged.
Thanks for your module and its improvement, Sven!

PR #2347, opened 2017-07-06
apcups: use APCUPS_DEFAULT_SERVICE fallback

pmjdebruijn:
Previously the apcups plugin could not be used without defining
an explicit configuration.

While the apcups plugin did provide a localhost fallback, it
did not however provide a 3551 port fallback in practice,
eventhough a constant for that purpose was already present.

### DIFF
--- a/src/apcups.c
+++ b/src/apcups.c
@@ -423,10 +423,12 @@ static int apcups_read(void) {
 
   int status =
       apc_query_server(conf_node == NULL ? APCUPS_DEFAULT_NODE : conf_node,
-                       conf_service, &apcups_detail);
+                       conf_service == NULL ? APCUPS_DEFAULT_SERVICE : conf_service,
+                       &apcups_detail);
   if (status != 0) {
     DEBUG("apcups plugin: apc_query_server (\"%s\", \"%s\") = %d",
-          conf_node == NULL ? APCUPS_DEFAULT_NODE : conf_node, conf_service,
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Hi!
Thanks for your patch. That works well.

@rubenk, lets merge this?

rubenk:
Thanks for your patch.

I'd rather like to see that we set the default host and port in apcups_config once, instead of on every read.

rpv-tomsk:
Ahh...

It was ok when it was
```
apc_query_server(conf_node == NULL ? APCUPS_DEFAULT_NODE : conf_node,
                       conf_service, &apcups_detail);
```
but become 'not ok' when it become
```
apc_query_server(conf_node == NULL ? APCUPS_DEFAULT_NODE : conf_node,
                       conf_service == NULL ? APCUPS_DEFAULT_SERVICE : conf_service,
                       &apcups_detail);
```
:-)

I will publish new PR with `apcups_init()` implemented.

rubenk:
Well, I didn't like the original code either, but since we have to change it anyway, might as well do it right ;)

rpv-tomsk:
>Well, I didn't like the original code either, but since we have to change it anyway, might as well do it right ;)

Agreed with you.

One of proverb which I am guided by, being translated, looks like to:

"Do it as best you can, it will happen bad by itself."

(In the sense of the meaning "In order to get bad results, you do not need to make an effort.").


@rubenk, Can you please look into 2349?

PR #2349, opened 2017-07-06
Free `userdata` if plugin_register_complex_read() has failed.

rpv-tomsk:
Citation from Collectd wiki (https://collectd.org/wiki/index.php/User_data_t):

> If the callback is deleted or callback registration is failed, the struct is freed using the my_config_free function.

There are several functions which register callbacks and has `user_data_t` argument:

- `plugin_register_write(), plugin_register_log(), plugin_register_notification(), plugin_register_flush()` - which calls `create_register_callback()`.
- `plugin_register_complex_read()`, which implemented in other way.

Most of plugins expects what free() will happen if registration failed.

For first functions group, in case of failure, `create_register_callback()` calls `destroy_callback()` which frees userdata.
But there was no code in `plugin_register_complex_read()` for userdata free.

That code was added, and some plugins (and Collectd itself) code was adjusted to new behaviour.

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -191,16 +191,21 @@ static int plugin_update_internal_statistics(void) { /* {{{ */
   return 0;
 } /* }}} int plugin_update_internal_statistics */
 
+static void free_userdata(user_data_t const *ud) /* {{{ */
+{
+  if (ud == NULL)
+    return;
+
+  if ((ud->data != NULL) && (ud->free_func != NULL)) {
+    ud->free_func(ud->data);
+  }
+} /* }}} void free_userdata */
+
 static void destroy_callback(callback_func_t *cf) /* {{{ */
 {
   if (cf == NULL)
     return;
-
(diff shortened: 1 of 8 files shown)


rpv-tomsk:
Hi Florian,
please review and merge this.

Thanks!

Rebased, passed through clang-format.   cc @octo

src/daemon/plugin.c
```diff
@@ -345,6 +349,7 @@ static int create_register_callback(llist_t **list, /* {{{ */
 
   cf = calloc(1, sizeof(*cf));
   if (cf == NULL) {
+    free_userdata(*ud);
```
octo: This dereferences `ud` without checking for `NULL`. I'd suggest to change `free_userdata` to take a `user_data_t*` (i.e. a pointer) instead and check for `NULL` there.

rpv-tomsk: Updated, thanks!

Things become clearer after this change.

PR #2351, opened 2017-07-07
apcups: allow to use plugin without explicit configuration

rpv-tomsk:
Previously the apcups plugin could not be used without defining
an explicit configuration.

While the apcups plugin did provide a localhost fallback, it
did not however provide a 3551 port fallback in practice,
eventhough a constant for that purpose was already present.

Thanks to Pascal de Bruijn for report.

Supersedes: #2347

### DIFF
--- a/src/apcups.c
+++ b/src/apcups.c
@@ -421,13 +421,11 @@ static int apcups_read(void) {
       .linefreq = NAN,
   };
 
-  int status =
-      apc_query_server(conf_node == NULL ? APCUPS_DEFAULT_NODE : conf_node,
-                       conf_service, &apcups_detail);
+  int status = apc_query_server(conf_node, conf_service, &apcups_detail);
+
   if (status != 0) {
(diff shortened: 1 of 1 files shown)


src/apcups.c
```diff
@@ -436,8 +434,19 @@ static int apcups_read(void) {
...
+
+  if (conf_service == NULL)
+    conf_service = APCUPS_DEFAULT_SERVICE;
+
+  return 0;
+} /* apcups_init */
+
 void module_register(void) {
   plugin_register_complex_config("apcups", apcups_config);
+  plugin_register_init("apcups", apcups_init);
```
rubenk: I might miss something, but why don't you do this in apcups_config?

rpv-tomsk: The goal is to allow plugin use without without defining an explicit configuration.
Then there will be `LoadPlugin apcups` only, without `<Plugin apcups>` block  in config file.
The `apcups_config()` will not be called at all in that case.

rubenk: I see, thanks for the explanation.

Issue #2025, opened 2016-11-05
acpups plugin dispatches missing/unknown values


Self-Perfection:
*   Version of collectd: 5.6.0
*   Operating system / distribution: Arch linux

`apcupsd` does not provide frequency and temperature for APC Back-UPS RS 900G

`apcups` plugin never receives `ITEMP` and `LINEFREQ` tokens and disaptches default values of frequency and temperature variables.

Therefore collectd saves extra data which does not make sense. e.g. graph with constant line at -300.0 degrees it useless.

![graph php](https://cloud.githubusercontent.com/assets/3757038/20031246/b7350090-a38d-11e6-920f-a8594b0df4de.png)

I assume values that have not been actually measured should not be dispatched to writing plugins.

PR #2353, opened 2017-07-09
apcups plugin: fix missing rrd files

guidosarducci:
The collect 5.5.3 release included commit 0a95eb509bebdd867a65105cdcbc0325d1cccddc to fix issue https://github.com/collectd/collectd/issues/2025, but it introduced a regression that broke existing graphing as seen on OpenWRT and LEDE.

Since not all APC UPSes report the same information, the apcups plugin acts as an abstraction layer and provides default values for unreported details. Some of those values were non-physical however (e.g. -300 degree temp). The commit 0a95eb509bebdd867a65105cdcbc0325d1cccddc changed the defaults to NaN (and rightly so) but then also threw away any NaNs.  This later change meant that some `.rrd` files previously created for unreported details would now be missing, which breaks both the consistent abstraction and also some working graphing.

This PR simply preserves the NaNs and restores the previous working graphing. It has been tested on 5.5.3 but still applies to master.

### DIFF
--- a/src/apcups.c
+++ b/src/apcups.c
@@ -385,9 +385,6 @@ static int apcups_config(oconfig_item_t *ci) {
 
 static void apc_submit_generic(const char *type, const char *type_inst,
                                gauge_t value) {
-  if (isnan(value))
-    return;
-
   value_list_t vl = VALUE_LIST_INIT;
   vl.values = &(value_t){.gauge = value};
   vl.values_len = 1;


rpv-tomsk:
IMHO: Collectd should not send metrics if it has no data for them.

There are other plugins in Collected with the same behaviour as you can see in apcups plugin and we can treat that as common behaviour.

PR #2355, opened 2017-07-10
collection3: Silence CGI::param list context warning in Graph/Common.pm

cjmayo:
```
CGI::param called in list context from collection3/lib/Collectd/Graph/Common.pm line 529,
this can lead to vulnerabilities. See the warning in
"Fetching the value or values of a single named parameter"
at /usr/lib64/perl5/vendor_perl/5.24.1/CGI.pm line 404.
```

This does need checking it isn't a real problem. As I understand it the new function just hides the warning.
http://search.cpan.org/~leejo/CGI-4.36/lib/CGI.pod#Fetching_the_names_of_all_the_parameters_passed_to_your_script

Added in CGI-4.08 on 2014-10-18:
http://cpansearch.perl.org/src/LEEJO/CGI-4.36/Changes

Newer CGI module present in Ubuntu 16.04LTS and Debian Jessie.


rpv-tomsk:
Maybe `local $CGI::LIST_CONTEXT_WARN = 0;` will be better solution than breaking support of older perl / CGI.pm versions?

cjmayo:
Not sure a nice solution exists really.
Adding an extra assignment to every request ... I think I'd rather just patch my own.
But then a global assignment isn't great in itself, plus for long-term maintenance it would be some distance away from the call.
Breaking support of older versions isn't good either.

Maybe best just to wait until the older installations go away.

dago:
How about
```
my @temp = (defined &CGI::multi_param ? multi_param( $part ) : param( $part ) );
```

rpv-tomsk:
```
--- Common.pm.orig      2018-12-04 19:30:34.000000000 +0000
+++ Common.pm.new       2020-02-26 18:28:07.993486936 +0000
@@ -79,6 +79,12 @@
 $subInstance->{disk} = 1;
 $subInstance->{virt} = 1;

+BEGIN {
+  unless (defined &CGI::multi_param) {
+     *Collectd::Graph::Common::multi_param = \&CGI::param;
+  }
+}
+
 return (1);

 sub _sanitize_generic_allow_minus
@@ -558,7 +564,7 @@
   for (qw(hostname plugin plugin_instance type type_instance))
   {
     my $part = $_;
-    my @temp = param ($part);
+    my @temp = multi_param ($part);
     if (!@temp)
     {
       next;
@@ -569,13 +575,13 @@
       for my $plugin (@temp)
       {
         if ($plugin =~ /^([^-]+)-(.*)$/) {
-          push $ident->{plugin}, [
+          push @{$ident->{plugin}}, [
             _sanitize_generic_no_minus ($1),  #plugin
             _sanitize_generic_allow_minus ($2),  #plugin_instance. It can contain dash?
           ];
           next;
         };
-        push $ident->{plugin}, _sanitize_generic_no_minus ($plugin);
+        push @{$ident->{plugin}}, _sanitize_generic_no_minus ($plugin);
       };
     }
     elsif ($part eq 'type')
@@ -615,7 +621,7 @@
     $ret{$_} = 0;
   }

-  for (param ('hostname'))
+  for (multi_param ('hostname'))
   {
     my $host = _sanitize_generic_allow_minus ($_);
     if (defined ($ret{$host}))
@@ -640,12 +646,12 @@
   my @hosts = get_host_selection ();

   my $selectedPlugins = {};
-  for my $plugin (param ('plugin'))
+  for my $plugin (multi_param ('plugin'))
   {
     $selectedPlugins->{$plugin} = 1;
   }
   my $selectedInstances = {};
-  for my $instance (param ('plugin_instance'))
+  for my $instance (multi_param ('plugin_instance'))
   {
     $selectedInstances->{$instance} = 1;
   }
```

dago:
@rpv-tomsk A bit hackish to redefine the function glob as fallback, but given the limited usecase I would say this is pretty much ok 👍 Would you mind updating the PR?

rpv-tomsk:
>Would you mind updating the PR?

I think I have no permissions for that, sorry.

rpv-tomsk:
>Added in CGI-4.08 on 2014-10-18:
>http://cpansearch.perl.org/src/LEEJO/CGI-4.36/Changes
>
>Newer CGI module present in Ubuntu 16.04LTS and Debian Jessie.

@dago,
however regarding the need of this trick: I think it can be omitted, because another two years pass since PR opened. 6 years looks like enough period to start use new CGI version.

dago:
@rpv-tomsk You are probably right about the workaround, but at least the rest of your patch should be applied. I wrongly assumed this PR was from you, that is why I asked for an update. Do you want to create a new PR with this change?

octo:
Hi @cjmayo, @dago, does one of you want to port this change to a recent version of *main* so it can be merged?

cjmayo:
I've rebased on main, but I forgot to move the PR target branch until afterwards so the checks will need to be rerun.

Looks like the `$ident->{plugin}` and `param ('plugin_instance')` changes aren't relevant any more? I haven't had a chance to investigate.

dago:
There seems to be a problem with the CI. Can you please add a dummy commit (like one extra empty line) to retrigger the CI?

cjmayo:
Well, I've rebased and force pushed without any changes.

I worked around one CI failure for another commit by basing on an earlier commit on main, now there is another problem...

dago:
This is a real bug - #3532

jnahmias:
Hi @cjmayo , 
can you please see #3471. It would be good to fix collection.cgi as well. Also, making the scalar references to CGI::param() explicit.

cjmayo:
> Hi @cjmayo ,
> can you please see #3471. I think it would be good to fix collection.cgi as well. Also, making the scalar references to CGI::param() explicit.

Looks like this one is a subset of that - I really don't mind what happens, I'll leave that to someone who can merge - whether just to drop this one and merge the other.

jnahmias:
@cjmayo - i'm good either way. probably nobody uses collection.cgi or collection3 anymore, but I hate having useless warnings when we can fix them :)

Issue #2338, opened 2017-07-02
potential null dereference in snmp_agent plugin


rubenk:
```
  CC       src/snmp_agent_la-snmp_agent.lo
src/snmp_agent.c:423:8: warning: Dereference of null pointer
  if (!(*dd)->table->index_oid.oid_len) {
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

rubenk:
(*dd)->table can be NULL here, but I'm not sure of the correct fix.

@vmytnykx would you mind taking a look?

rpv-tomsk:
CC @rkorynkx

rpv-tomsk:
That is not potential but real issue.

```
Core was generated by `/opt/collectd/sbin/collectd -C /opt/collectd-snmp_agent-test.conf -f'.
Program terminated with signal 11, Segmentation fault.
#0  snmp_agent_free_data (dd=0x8443b44) at src/snmp_agent.c:423
423       if (!(*dd)->table->index_oid.oid_len) {
(gdb) bt full
#0  snmp_agent_free_data (dd=0x8443b44) at src/snmp_agent.c:423
No locals.
#1  0xb74ccb58 in snmp_agent_free_config () at src/snmp_agent.c:1495
        de = <optimized out>
#2  0xb74ce6e0 in snmp_agent_config (ci=0x8421198) at src/snmp_agent.c:1561
        ret = <optimized out>
#3  0x08050136 in dispatch_block_plugin (ci=0x8421198) at src/daemon/configfile.c:395
        old_ctx = {interval = 0, flush_interval = 0, flush_timeout = 0}
        ret_val = <optimized out>
        cb = 0x8431998
        name = 0x84254f8 "snmp_agent"
#4  0x080505c8 in dispatch_block (ci=<optimized out>) at src/daemon/configfile.c:423
No locals.
#5  cf_read (filename=0xbfe6fe54 "/opt/collectd-snmp_agent-test.conf") at src/daemon/configfile.c:1020
        i = 10
        conf = 0x8421088
        ret = 0
#6  0x0804d545 in main (argc=4, argv=0xbfe6de54) at src/daemon/collectd.c:522
```

That binary was build from bea5daaed7b5cf9160207a95292947873fdcf927 

The configuration itself is very simple:

```
Hostname "test"
FQDNLookup false

PIDFile     "/var/run/collectd-xxx.pid"

LoadPlugin syslog
[...]

PR #2362, opened 2017-07-13
snmp_agent: Fixes and improvements

rpv-tomsk:
List of changes:

* Added comments into code;
* Fixed possible race conditions between thread start and locks initialization. Lock initialization should be done before starting pthread;
* Added missing `llist_destroy()` calls;
* Added a check for `llist_create()` success;
* Fix null pointer dereference (Issue #2338);
* Changed plugin initialization:
  * Do not start working thread if plugin not configured.
   That allows to avoid useless CPU consumption by polling loop.
  * Do not register 'write' / 'missing' callbacks if there was no `<Table>` configured.
   These callbacks do not do anything in this case.
  * Registration of 'shutdown' callback delayed to 'init' phase.
* Documented `PluginInstance` option.

### DIFF
--- a/src/snmp_agent.c
+++ b/src/snmp_agent.c
@@ -98,6 +98,7 @@ static snmp_agent_ctx_t *g_agent = NULL;
       (_dd->type ? !strcmp(_dd->type, _t) : 0) &&                              \
       (_dd->type_instance ? !strcmp(_dd->type_instance, _ti) : 1)
 
+static int snmp_agent_shutdown(void);
 static void *snmp_agent_thread_run(void *arg);
 static int snmp_agent_register_oid(oid_t *oid, Netsnmp_Node_Handler *handler);
 static int snmp_agent_set_vardata(void *dst_buf, size_t *dst_buf_len,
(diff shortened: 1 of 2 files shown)


src/snmp_agent.c
```diff
@@ -109,13 +110,13 @@ static u_char snmp_agent_get_asn_type(oid *oid, size_t oid_len) {
   struct tree *node = get_tree(oid, oid_len, g_agent->tp);
 
   return (node != NULL) ? mib_to_asn_type(node->type) : 0;
-}
+} /* u_char snmp_agent_get_asn_type */
```
rubenk: I know that lot's of existing code in the code base uses these kinds of comments, but this plugin doesn't and introducing them doesn't gain much. It's easy enough to see where a function ends.

rubenk:
Hi Pavel,

Thanks for doing this. I left just one comment, if you could address that I'll merge.

rpv-tomsk:
Hi Ruben!

Commit with comments added was removed.
Thanks for your review.

PR #2363, opened 2017-07-16
rrdtool plugin: Fix flushing

rpv-tomsk:
Last days I wanted to spent some time to tune its performance and smooth his disk IO load.

I have the following chart of disk operations:

disk_ops

The configuration is:

```
WritesPerSecond 60
RandomTimeout 900     # 15 min
CacheTimeout 3600     #  1 hour
CacheFlush 600          # 10 min
```

The `CacheFlush` was set to such a small value during one of previous attempts to smooth load.
My logic was: 'I will do that check more often so less values will be flushed to disk at once, 
even at the cost of high CPU load.`.

The expected smoothness of disk IO was not achieved. 
But this does not bother anyone, so it works this way for a long time.

Later code review shows that my logic is unallowed. 

The next attempt to smooth load came. It started from IO analysis.

On chart I see IO bursts. Their period matches to `CacheTimeout` option value.
The width of the bursts corresponds to `RandomTimeout` value and increases with time, while height decreases.
That is expected behaviuor. 

But observed bursts with a period of 10h were not expected.

While Collectd code explore I found a cause of 10 hour periodical action.

```
  if (cache_timeout == 0) {
    cache_flush_timeout = 0;
  } else if (cache_flush_timeout < cache_timeout)
    cache_flush_timeout = 10 * cache_timeout;
```

This limitation was not documented.

So, my configuration treated as

```
WritesPerSecond 60
RandomTimeout 900     # 15 min
CacheTimeout 3600     #  1 hour
CacheFlush 36000      # 10 hours
```

The `cache_flush_timeout` (set from `CacheFlush` option) used in 

```
  if ((cache_timeout > 0) &&
      ((cdtime() - cache_flush_last) > cache_flush_timeout))
    rrd_cache_flush(cache_flush_timeout);
```

The `rrd_cache_flush(cdtime_t timeout)` flushes cache elements if their 'age' greater than 'timeout'.

While the value is '10h' the flush should not occur either! But it happens...

Then I found a strange code:

```
static void rrd_cache_flush(cdtime_t timeout) {
   ...
  
   timeout = TIME_T_TO_CDTIME_T(timeout);
   ...

```

That looks like mistake.

For `CacheTimeout 3600` the `cache_timeout` value is 0x0384 0000 0000.
The `cache_flush_timeout` then equal to 0x0000 2328 0000 0000 and that value passed to `rrd_cache_flush()`.

After `<< 30` operation done in `TIME_T_TO_CDTIME_T`, there will be a zero value in `timeout`, 
so flush occurs for all cache elements.

Also, the value to `cache_flush_timeout` was assigned as a simple integer seconds, while it was later used as a `cdtime_t`, so the condition

```
if (cache_flush_timeout < cache_timeout)
    cache_flush_timeout = 10 * cache_timeout;
```
is always true.

These all was fixed in this PR.

That patch was put to my production server (and RandomTimeout set to 1800). 
After update it works as expected, no full cache flush occurs anymore.

disk_ops_2

Spike after 10h of work caused by the fact what `rrd_cache_flush(cache_timeout);` call does not honour `RandomTimeout`.

All cache values with positive `random_variation` value are flushed.
In my case, that is 1/4 of cache ( RandomTimeout 1800 / 2 / CacheTimeout 3600) was flushed.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6957,7 +6957,8 @@ one (exclusive).
 
 When the C<rrdtool> plugin uses a cache (by setting B<CacheTimeout>, see below)
 it writes all values for a certain RRD-file if the oldest value is older than
-(or equal to) the number of seconds specified. If some RRD-file is not updated
+(or equal to) the number of seconds specified by B<CacheTimeout>.
+That check happens on new values arriwal. If some RRD-file is not updated
 anymore for some reason (the computer was shut down, the network is broken,
 etc.) some values may still be in the cache. If B<CacheFlush> is set, then the
 entire cache is searched for entries older than B<CacheTimeout> seconds and
@@ -6965,6 +6966,10 @@ written to disk every I<Seconds> seconds. Since this is kind of expensive and
 does nothing under normal circumstances, this value should not be too small.
 900 seconds might be a good value, though setting this to 7200 seconds doesn't
 normally do much harm either.
+
+Default value for this option is 10x of B<CacheTimeout>.
+If value of B<CacheFlush> less than value of B<CacheTimeout> then default value
+used too.
 
 =item B<CacheTimeout> I<Seconds>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6963,13 +6963,14 @@ anymore for some reason (the computer was shut down, the network is broken,
 etc.) some values may still be in the cache. If B<CacheFlush> is set, then
 every I<Seconds> seconds the entire cache is searched for entries older than
 B<CacheTimeout> + B<RandomTimeout> seconds. The entries found are written to
-disk. Since entire cache scan is kind of expensive and does nothing under normal
-circumstances, this value should not be too small. 900 seconds might be a good
-value, though setting this to 7200 seconds doesn't normally do much harm either.
+disk. Since scanning the entire cache is kind of expensive and does nothing
+under normal circumstances, this value should not be too small. 900 seconds
+might be a good value, though setting this to 7200 seconds doesn't normally
+do much harm either.
 
-Default value for this option is 10x of B<CacheTimeout>.
-If value of B<CacheFlush> less than value of B<CacheTimeout> then default value
-used too.
+Defaults to 10x B<CacheTimeout>.
+B<CacheFlush> must be larger than or equal to B<CacheTimeout>, otherwise the
+above default is used.
 
 =item B<CacheTimeout> I<Seconds>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6960,12 +6960,12 @@ it writes all values for a certain RRD-file if the oldest value is older than
 (or equal to) the number of seconds specified by B<CacheTimeout>.
 That check happens on new values arriwal. If some RRD-file is not updated
 anymore for some reason (the computer was shut down, the network is broken,
-etc.) some values may still be in the cache. If B<CacheFlush> is set, then the
-entire cache is searched for entries older than B<CacheTimeout> seconds and
-written to disk every I<Seconds> seconds. Since this is kind of expensive and
-does nothing under normal circumstances, this value should not be too small.
-900 seconds might be a good value, though setting this to 7200 seconds doesn't
-normally do much harm either.
+etc.) some values may still be in the cache. If B<CacheFlush> is set, then
+every I<Seconds> seconds the entire cache is searched for entries older than
+B<CacheTimeout> + B<RandomTimeout> seconds. The entries found are written to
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
Collectd hub with all these patches and configuration
```
        WritesPerSecond 60
        RandomTimeout 3600
        CacheTimeout 3600
        CacheFlush 3600
```
produces the following disk load 
disk_ops_3

No more cache flushes and no more spikes. It starts to write data right after Collectd start.
There is a light skew after start, while cache populates, but all works according to configuration and implementation.

src/rrdtool.c
```diff
@@ -606,23 +605,10 @@ static int rrd_cache_flush_identifier(cdtime_t timeout,
...
-    INFO("rrdtool plugin: Adjusting \"RandomTimeout\" to %.3f seconds.",
-         CDTIME_T_TO_DOUBLE(cache_timeout));
-    random_timeout = cache_timeout;
-  }
-
-  max = (long)(random_timeout / 2);
-  min = max - ((long)random_timeout);
-
-  return (int64_t)cdrand_range(min, max);
+  return (int64_t)cdrand_range(-random_timeout, random_timeout);
```
octo: Excellent solution, thank you!

src/rrdtool.c
```diff
@@ -885,7 +871,7 @@ static int rrd_config(const char *key, const char *value) {
             "be greater than 0.\n");
       return 1;
     }
-    cache_flush_timeout = tmp;
+    cache_flush_timeout = TIME_T_TO_CDTIME_T(tmp);
```
octo: I'd prefer to copy the way `CacheTimeout` is handled and use a double, `atof()` and `DOUBLE_TO_CDTIME_T()` here.

src/collectd.conf.pod
```diff
@@ -6957,14 +6957,19 @@ one (exclusive).
...
-etc.) some values may still be in the cache. If B<CacheFlush> is set, then the
-entire cache is searched for entries older than B<CacheTimeout> seconds and
-written to disk every I<Seconds> seconds. Since this is kind of expensive and
-does nothing under normal circumstances, this value should not be too small.
-900 seconds might be a good value, though setting this to 7200 seconds doesn't
-normally do much harm either.
+etc.) some values may still be in the cache. If B<CacheFlush> is set, then
+every I<Seconds> seconds the entire cache is searched for entries older than
+B<CacheTimeout> + B<RandomTimeout> seconds. The entries found are written to
+disk. Since entire cache scan is kind of expensive and does nothing under normal
```
octo: "Since scanning the entire cache is …"

src/collectd.conf.pod
```diff
@@ -6957,14 +6957,19 @@ one (exclusive).
...
+circumstances, this value should not be too small. 900 seconds might be a good
+value, though setting this to 7200 seconds doesn't normally do much harm either.
+
+Default value for this option is 10x of B<CacheTimeout>.
```
octo: "Defaults to 10x B&lt;CacheTimeout&gt;."

src/collectd.conf.pod
```diff
@@ -6957,14 +6957,19 @@ one (exclusive).
...
+If value of B<CacheFlush> less than value of B<CacheTimeout> then default value
```
octo: "B&lt;CacheFlush&gt; must be larger than or equal to B&lt;CacheTimeout&gt;, otherwise the above default is used."

rpv-tomsk:
Thanks for a review. 

The new commit added with these remarks fixed.

octo:
FYI, this regression was likely introduced with d91c8449c887c87cd776d06eb07b78cca73707b1. I'm therefore going to pull this into the *collectd-5.6* branch as a bugfix.

octo:
Rebased as:

*   c7cbb2af24de8b33186dd45b5731b87ea7152173
*   20115ea505208ac2f1e8b453a35284b871db1321
*   c1facb5733721a14454f41bab2b97f8c8ceffc8e
*   055e37f556974bca6152bda8a1144742da371905
*   f5177e4f60648ac19d785235a91235e6e401f0cb
*   2739b2c65f22e459d947bdd10cb9a9807ca729e8

rpv-tomsk:
>FYI, this regression was likely introduced with d91c844. I'm therefore going to pull this into the collectd-5.6 branch as a bugfix.

Should we fix `perl` plugin to reflect this change too?

octo:
@rpv-tomsk ACK, I think we should.

@tokkee The Perl-plugin seems to simply pass the raw `cdtime_t` to Perl code. I'd prefer if Perl code would get something compatible with [Time::HiRes](http://perldoc.perl.org/Time/HiRes.html), i.e. likely a float scalar. Does that make sense or am I missing something?

PR #2369, opened 2017-07-18
Add new NFS 4.2 counters on Linux

marekbecka:
This patch adds new NFS 4.2 procedures that are used in Linux since 3.18


octo:
Thanks @marekbecka for your work and @jumitche for the review! I've rebased on *master* and re-formatted the code.

PR #2374, opened 2017-07-20
intel_pmu: fix compilation error caused by migration to snprintf().

rkorynkx:
Fixed compilation errors. ssnprintf() was not replaced to snprintf() while merging intel_pmu plugin to master branch.


octo:
Eek! Thanks for fixing this @rkorynkx!

Issue #2358, opened 2017-07-12
`Value too old` for the `0.0.0.0`, ntp plugin


zerkms:
*   Version of collectd: 5.5.1
*   Operating system / distribution: ubuntu 16.04.2 server

I just installed a collectd and an ntpd on a new ubuntu 16.04.2 server and found that collectd emits annoying messages to the syslog:

```
Jul 12 01:59:45 server collectd[2773]: uc_update: Value too old: name = server.domain.tld/ntpd/time_dispersion-0.0.0.0; value time = 1499824785.998; last cache update = 1499824785.998;
Jul 12 01:59:45 server collectd[2773]: uc_update: Value too old: name = server.domain.tld/ntpd/time_offset-0.0.0.0; value time = 1499824785.998; last cache update = 1499824785.998;
Jul 12 01:59:45 server collectd[2773]: uc_update: Value too old: name = server.domain.tld/ntpd/delay-0.0.0.0; value time = 1499824785.998; last cache update = 1499824785.998;
```

`ntpdc` returns

```
# ntpdc -c peers localhost
     remote           local      st poll reach  delay   offset    disp
=======================================================================
=timeball1.its.w 10.50.200.3     16   64    0 0.00000  0.000000 4.00000
=timeball3.its.w 10.50.200.3     16   64    0 0.00000  0.000000 4.00000
=ns2.tdc.akl.tel 10.50.200.3      2   64    1 0.00873  0.000019 2.81735
=ntp3.ntp.net.nz 10.50.200.3     16   64    0 0.00000  0.000000 4.00000
=ns1.tdc.akl.tel 10.50.200.3     16   64    0 0.00000  0.000000 4.00000
=ns2.att.wlg.tel 10.50.200.3     16   64    0 0.00000  0.000000 4.00000
=ntp1.ntp.net.nz 10.50.200.3      1   64    1 0.01031  0.001045 2.81735
=122.252.188.99  10.50.200.3     16   64    0 0.00000  0.000000 4.00000
=0.0.0.0         0.0.0.0         16   64    0 0.00000  0.000000 4.00000
=0.0.0.0         0.0.0.0         16   64    0 0.00000  0.000000 4.00000
=0.0.0.0         0.0.0.0         16   64    0 0.00000  0.000000 4.00000
=0.0.0.0         0.0.0.0         16   64    0 0.00000  0.000000 4.00000
=0.0.0.0         0.0.0.0         16   64    0 0.00000  0.000000 4.00000
```

[...]

PR #2376, opened 2017-07-22
Skip `0.0.0.0` hosts in ntpd plugin

zerkms:
This PR addresses https://github.com/collectd/collectd/issues/2358

Here I'm simply skipping `0.0.0.0` hosts that are returned for POOL servers that should not be reported.


rubenk:
Makes sense, thanks @zerkms!

PR #2377, opened 2017-07-24
openldap: Fix crash on shutdown when openldap connection failed

rpv-tomsk:
When plugin failes to connect to LDAP server, the `ldap_unbind_ext_s(st->ld, NULL, NULL);` is called.
According to `man 3 ldap_unbind_ext_s` (https://linux.die.net/man/3/ldap_unbind_ext_s), that call is used to unbind from the directory, terminate the current association, and free the resources contained in the ld structure. 
When plugin shutdown, the `ldap_unbind_ext_s` called again, which causes coredump.

The plugin code was changed to set `st->ld=NULL` after `ldap_unbind_ext_s` call.
No more coredumps.

### DIFF
--- a/src/openldap.c
+++ b/src/openldap.c
@@ -47,7 +47,6 @@ struct cldap_s /* {{{ */
   char *password;
   char *cacert;
   char *host;
-  int state;
   _Bool starttls;
   int timeout;
   char *url;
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
This change was checked in different conditions on my LDAP server, and it works perfectly.

Let's merge it, please.

rpv-tomsk:
Ping.

octo:
Hi @rpv-tomsk, thanks, this looks much better now!

Issue #2370, opened 2017-07-18
Adding metadata support to the GRPC plugin


cramertj:
I'm interested in adding support for writing metric metadata to the existing GRPC plugin. I discussed this privately with @tokkee, and they were concerned about adding the ability to specify metadata externally because metadata was originally intended for in-proc data. Is this a design constraint that collectd's maintainers want to enforce, or would you all be interested in adding support for writing metadata through plugins?

Implementation wise, this should be relatively simple-- I believe all that is required is to modify the existing `ValueList` GRPC proto definition to include metadata, and to modify [marshal_value_list and unmarshal_value_list](https://github.com/collectd/collectd/blob/27d117edf8dd7ae0b0f85772361f3df7ac933ae6/src/grpc.cc#L200) so that they copy metadata from the `value_list_t` to the `collectd::types::ValueList` and vice versa.

octo:
Hi @cramertj,

thank you for reaching out, it's very useful to clear such questions prior to spending time on an implementation. There is precedence for exporting the meta data, e.g. the *Write Riemann* and *Write TSDB* plugins, so I'd be happy to accept such a PR.

Best regards,
—octo

cramertj:
Great! I'll be sure to take a look at those plugins when I implement this.

PR #2378, opened 2017-07-24
Add metadata support to GRPC plugin

cramertj:
As mentioned in #2370, I've added metadata support to the GRPC plugin. In order to do this, I also had to add the `uc_iterator_get_meta` function for getting metadata out of a cache iterator.

I'm still relatively inexperienced in C/C++, so let me know if there are any changes I should make to be more idiomatic.

### DIFF
--- a/src/daemon/utils_cache.c
+++ b/src/daemon/utils_cache.c
@@ -917,11 +917,6 @@ int uc_iterator_get_meta(uc_iter_t *iter, meta_data_t **ret_meta) {
   if ((iter == NULL) || (iter->entry == NULL) || (ret_meta == NULL))
     return -1;
 
-  if (iter->entry->meta == NULL) {
-    *ret_meta = NULL;
-    return 0;
-  }
-
   *ret_meta = meta_data_clone(iter->entry->meta);
 
   return 0;
--- a/src/daemon/utils_cache.c
+++ b/src/daemon/utils_cache.c
@@ -913,6 +913,20 @@ int uc_iterator_get_interval(uc_iter_t *iter, cdtime_t *ret_interval) {
   return 0;
 } /* int uc_iterator_get_name */
 
+int uc_iterator_get_meta(uc_iter_t *iter, meta_data_t **ret_meta) {
+  if ((iter == NULL) || (iter->entry == NULL) || (ret_meta == NULL))
+    return -1;
+
+  if (iter->entry->meta == NULL) {
+    *ret_meta = NULL;
+    return 0;
+  }
+
+  *ret_meta = meta_data_clone(iter->entry->meta);
+
+  return 0;
+} /* int uc_iterator_get_meta */
+
 /*
  * Meta data interface
  */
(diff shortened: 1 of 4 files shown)


src/daemon/utils_cache.c
```diff
@@ -913,6 +913,20 @@ int uc_iterator_get_interval(uc_iter_t *iter, cdtime_t *ret_interval) {
   return 0;
 } /* int uc_iterator_get_name */
 
+int uc_iterator_get_meta(uc_iter_t *iter, meta_data_t **ret_meta) {
+  if ((iter == NULL) || (iter->entry == NULL) || (ret_meta == NULL))
+    return -1;
+
+  if (iter->entry->meta == NULL) {
```
octo: `meta_data_clone()` gracefully deals with `NULL` arguments (by returning `NULL`), so this special case is not needed.

src/grpc.cc
```diff
@@ -170,9 +170,87 @@ static grpc::Status marshal_value_list(const value_list_t *vl,
   msg->set_allocated_time(new google::protobuf::Timestamp(t));
   msg->set_allocated_interval(new google::protobuf::Duration(d));
 
+  msg->clear_meta_data();
+  if (vl->meta != nullptr) {
```
octo: This `if` block is a bit too long. Could you break this out into a `marshal_meta_data()` function? Dito for unmarshal below.

src/grpc.cc
```diff
@@ -170,9 +170,87 @@ static grpc::Status marshal_value_list(const value_list_t *vl,
...
+        if (meta_data_get_boolean(vl->meta, key, &bool_value) != 0) {
+          strarray_free(meta_data_keys, meta_data_keys_len);
+          return grpc::Status(grpc::StatusCode::INTERNAL,
+                            grpc::string("missing metadata"));
+        }
+        md_value.set_bool_value(bool_value);
+        break;
+      default:
+        strarray_free(meta_data_keys, meta_data_keys_len);
+        return grpc::Status(grpc::StatusCode::INTERNAL,
```
octo: This patch adds an error message for this case when handling metric values. For symmetry I think we should have an error message here, too.

src/grpc.cc
```diff
@@ -249,8 +364,11 @@ static grpc::Status unmarshal_value_list(const collectd::types::ValueList &msg,
   if (status.ok()) {
     vl->values = values;
     vl->values_len = values_len;
-  } else if (values) {
-    free(values);
+  } else {
+    meta_data_destroy(vl->meta);
+    if (values) {
```
octo: This special case is not required since `free()` handles `NULL` gracefully.

src/grpc.cc
```diff
@@ -170,9 +170,87 @@ static grpc::Status marshal_value_list(const value_list_t *vl,
...
+        break;
+      default:
+        strarray_free(meta_data_keys, meta_data_keys_len);
+        return grpc::Status(grpc::StatusCode::INTERNAL,
+                            grpc::string("unknown metadata type"));
+      }
+
+      (*msg->mutable_meta_data())[grpc::string(key)] = md_value;
+
+      if (meta_data_keys != nullptr) {
```
octo: This case is not really needed; `strarray_free(NULL, 0)` is handled gracefully.

PR #2379, opened 2017-07-25
Push "collectd_plugin" to Lua registry to identify plugins.

alnsn:
Previously, lua_tostring() was used to identify readers/writers.
Unfortunately, it doesn't work on functions (it returns NULL).
Lua 5.2 and 5.3 come with luaL_tolstring() but it can (rather
theorethically in the current implementation) return identical
strings for two functions defined in two different Lua state.

This commit is a first step in adding unique identifiers to
readers and writers. It saves "lua/script.lua" in Lua registry
and uses it in register_read() and register_write() functions.

Unique identifier of read/write functions is still missing.

### DIFF
--- a/src/lua.c
+++ b/src/lua.c
@@ -281,8 +281,9 @@ static int lua_cb_register_read(lua_State *L) /* {{{ */
 
   luaL_checktype(L, 1, LUA_TFUNCTION);
 
-  char function_name[DATA_MAX_NAME_LEN];
-  snprintf(function_name, sizeof(function_name), "lua/%s", lua_tostring(L, 1));
+  lua_getfield(L, LUA_REGISTRYINDEX, "collectd_plugin");
+  char *function_name = strdup(lua_tostring(L, -1));
+  lua_pop(L, 1);
 
   int callback_id = clua_store_callback(L, 1);
   if (callback_id < 0)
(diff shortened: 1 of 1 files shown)


alnsn:
Both register_read() and register_write() create a new thread. An address of their thread object is unique across all Lua states and it can be used to identifies readers and writers. It's not human-readable, though. Perhaps a concatenation of collectd_plugin key and the address is the best solution. E.g. lua/myscript.lua/0xdeadbeef.

alnsn:
Commit a6df177ca4329d1b6573ce88e05b1cae314479b2 adds a thread to a unique identifier.

feckert:
Hello @alnsn I am working on some collectd lua plugins for OpenWrt and I have exactly the same problem! I am unable to registry more then one lua callback. I have seen your change for NetBSD http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/sysutils/collectd/patches/patch-src_lua.c?annotate=1.1

Wy is this PR blocked?

alnsn:
> Wy is this PR blocked?

Hi @feckert, I have no idea why it's blocked.

feckert:
Wy do we not rewrite this PR to the solution from the NetBSD repository from you?
http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/sysutils/collectd/patches/patch-src_lua.c?annotate=1.1

feckert:
@alnsn I'd suggest we start a new attempt to finally solve the problem. How do we get the attention from the collectd maintainer.

PR #2385, opened 2017-07-30
memcached: Fix hitratio reporting, add connections rate report

rpv-tomsk:
When Collectd calculates 'hitratio', it divides two continiously-grown values.
So, reported metric contains the average since 'memcached' start, which is incorrect.
Collectd should report actual hitratio, not average for weeks-months-years.

The second commit adds connections rate reporting.

hitratio

The value of `hitratio` is calculated as `get_hits / cmd_get` , unlike the rest, `incr_hitratio` and `decr_hitratio`, which are calculated like `incr_hits / (incr_hits + incr_misses)`.

The `get_hits / (get_hits + get_misses)` formula may be used to unify the code, if use only these variables.  But new `get_*` stats was added in memcached: the `get_expired` in memcached-1.4.26 , and `get_flushed` in memcached-1.4.28, so that unification cannot be applied.

### DIFF
--- a/src/memcached.c
+++ b/src/memcached.c
@@ -392,8 +392,9 @@ static void submit_gauge2(const char *type, const char *type_inst,
   plugin_dispatch_values(&vl);
 }
 
-static gauge_t calculate_rate(gauge_t part, gauge_t total, gauge_t *prev_part,
-                              gauge_t *prev_total) {
+static gauge_t calculate_ratio_percent(gauge_t part, gauge_t total,
+                                       gauge_t *prev_part,
+                                       gauge_t *prev_total) {
   if (isnan(*prev_part) || isnan(*prev_total) || (part < *prev_part) ||
       (total < *prev_total)) {
     *prev_part = part;
@@ -413,8 +414,8 @@ static gauge_t calculate_rate(gauge_t part, gauge_t total, gauge_t *prev_part,
   return 100.0 * num / denom;
 }
 
-static gauge_t calculate_rate2(gauge_t part1, gauge_t part2, gauge_t *prev1,
-                               gauge_t *prev2) {
+static gauge_t calculate_ratio_percent2(gauge_t part1, gauge_t part2,
+                                        gauge_t *prev1, gauge_t *prev2) {
   if (isnan(*prev1) || isnan(*prev2) || (part1 < *prev1) || (part2 < *prev2)) {
     *prev1 = part1;
     *prev2 = part2;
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Ping.

src/memcached.c
```diff
@@ -285,6 +297,47 @@ static void submit_gauge2(const char *type, const char *type_inst,
   plugin_dispatch_values(&vl);
 }
 
+static gauge_t calculate_rate(gauge_t part, gauge_t total, gauge_t *prev_part,
```
octo: I realize you copied this from the previous implementation, but "rate" is incorrect and confusing. What you're calculating here is a *ratio*.

src/memcached.c
```diff
@@ -285,6 +297,47 @@ static void submit_gauge2(const char *type, const char *type_inst,
...
+  gauge_t num = part - *prev_part;
+  gauge_t denom = total - *prev_total;
+
+  *prev_part = part;
+  *prev_total = total;
+
+  if (num == 0 || denom == 0)
+    return 0;
+
+  return 100.0 * num / denom;
```
octo: Either rename this function to `…_ratio` and return `num / denom` here, or use `…_ratio_percent` and return `100.0 * num / denom`.

src/memcached.c
octo: I find it a bit unintuitive that you're effectively calculating the rated of `gauge_t`. It looks like the raw metrics are integers, can you change these functions to take `derive_t` instead?

src/memcached.c
```diff
@@ -285,6 +297,47 @@ static void submit_gauge2(const char *type, const char *type_inst,
...
+  }
+
+  gauge_t num = part - *prev_part;
+  gauge_t denom = total - *prev_total;
+
+  *prev_part = part;
+  *prev_total = total;
+
+  if (num == 0 || denom == 0)
+    return 0;
```
octo: If there were no requests, the ratio should be `NAN`, not zero. Otherwise this will skew e.g. averages.

src/memcached.c
octo: Please consider using `value_to_rate()` to convert both, "part" and "total" to a rate and then simply calculating the ratio as `part_rate / total_rate`.

octo:
FYI, I've cherry-picked 7590014374870a7109e5cbaba2b03635c92ff3cf as 66ac89cce88debf8cb9d43dd7b63a865d0e1d471.

src/memcached.c
rpv-tomsk: The current code has following variables:
```
  gauge_t hits = NAN;
  gauge_t gets = NAN;
  gauge_t incr_hits = NAN;
```
https://github.com/collectd/collectd/pull/2385/commits/e8fdf93c32d752022478bed93315950f499bc271#diff-e842cfa6e2070dbfa4c72d77b5b2fc71R351

These variables are passed to `calculate_ratio()` and `calculate_ratio2()` functions, so they take `gauge_t` type as  as argument. What to do?

octo: Convert those variables to `derive_t`, too. I'm fairly certain they were only `gauge_t` so `rate = 100.0 * hits / gets;` could be calculated without a cast.

src/memcached.c
rpv-tomsk: Renamed to `..._ratio_percent`.

src/memcached.c
rpv-tomsk: Renamed.

rpv-tomsk: Updated, variables converted to `derive_t`.

rpv-tomsk: I prefer to leave this untouched. I think use of `value_to_rate()` is excessive for this task.
That adds time calculations, which will be reducted later as the numerator and denominator will have a same value.

src/memcached.c
rpv-tomsk: Updated, thanks!

src/memcached.c
rpv-tomsk: Renamed to `.._ratio_percent`.

rpv-tomsk:
Hi, Florian!

I had updated this. All nits are accepted except proposal of `value_to_rate()` usage.

Thanks for a review.

src/memcached.c
```diff
@@ -380,29 +390,68 @@ static void submit_gauge2(const char *type, const char *type_inst,
...
+
+  if (num == 0 || denom == 0)
+    return NAN;
+
+  return 100.0 * (gauge_t)num / (gauge_t)denom;
+}
+
+static gauge_t calculate_ratio_percent2(derive_t part1, derive_t part2,
+                                        derive_t *prev1, derive_t *prev2) {
+  if (isnan(*prev1) || isnan(*prev2) || (part1 < *prev1) || (part2 < *prev2)) {
```
octo:
```
src/memcached.c: In function 'calculate_ratio_percent2':
src/memcached.c:417:3: error: non-floating-point argument in call to function '__builtin_isnan'
   if (isnan(*prev1) || isnan(*prev2) || (part1 < *prev1) || (part2 < *prev2)) {
   ^
```

I think `(*prev1 == 0) || (*prev2 == 0)` should do the trick.

src/memcached.c
octo: Fair enough, SGTM

src/memcached.c
rpv-tomsk: Thanks for a notice ;-)

PR #2388, opened 2017-07-31
memcached: Persistent connections with IO timeouts

rpv-tomsk:
Implemented persistent connections. It is a must-have feature for statistics daemon.

### DIFF
--- a/src/memcached.c
+++ b/src/memcached.c
@@ -67,6 +67,7 @@ static void memcached_free(void *arg) {
   if (st->fd >= 0) {
     shutdown(st->fd, SHUT_RDWR);
     close(st->fd);
+    st->fd = -1;
   }
 
   sfree(st->name);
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
Ping.

src/memcached.c
```diff
@@ -124,77 +144,153 @@ static int memcached_connect_inet(memcached_t *st) {
...
-  if (fd < 0) {
+  memcached_connect(st);
+  if (st->fd < 0) {
     ERROR("memcached plugin: Instance \"%s\" could not connect to daemon.",
           st->name);
     return -1;
   }
 
-  status = (int)swrite(fd, "stats\r\n", strlen("stats\r\n"));
+  struct pollfd pollfd;
```
octo: As above: please use a struct initializer; consider a compound literal.

rpv-tomsk:
Patch updated.

Thanks for a review!

rpv-tomsk:
Added a small patch from #2300. That is related.

Issue #896, opened 2015-01-16
No equivalent for plugin_register_complex_read() available for Perl plugins


UnitedMarsupials:
We'd very much like to be able to use the same read-function to monitor multiple instances of a server. The most straightforward way to do this is by passing some sort of `user_data` to the function (as part of `@_`). Unfortunately, the list of types found in `Collectd.pm` does not include `TYPE_COMPLEX_READ` and the read-function is invoked without any arguments.

Perhaps, instead of (or in addition to) adding the new type, the `Collectd.pm` should simply be modified to treat all read-functions as "complex":
 1. Add (optional) argument to plugin_register -- allowing config-function, for example, to register the same read-function multiple times with references to different scalar, array, or hash.
 1. Pass that value to other functions -- and if they don't use the, fine:
```perl
     sub my_read {
            my $data = shift;
     ....
```
At the very least, the already-existing registration `name` (the second argument of the current `plugin_register`) should be available to *all* registered methods so the method can make a lookup in its own hash (indexed by name). For example:
```perl
    sub my_write
    {
	    my $type = shift;
	    my $ds   = shift;
	    my $vl   = shift;
        my $name = shift; # The name provided by the method-registration
        my $data = $instances{$name}; # The instances-hash built-up by the config-method
    ...
```
The ideal would allow passing the arbitrary "user data" to *all* registered methods (not just reads), but that's not possible even for the C plugins at the moment...

rpv-tomsk:
The patch added.

UnitedMarsupials:
Thank you. Though not the complete user-data option, this is good enough. However, shouldn't the `pluginname` and the `subname` be `_const_ char *`?

rpv-tomsk:
```
newSVpv Creates a new SV and copies a string into it.  The reference count for the SV is set to 1.  If "len" is zero, Perl will
               compute the length using strlen().  For efficiency, consider using "newSVpvn" instead.

                       SV*     newSVpv(const char *const s, const STRLEN len)
```

Yes, you are right. That сomes from the old `perl.c` code, so... .

Let's go to #2392 for future discussions?

UnitedMarsupials:
If [I have my way](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=62029) the "can be `const` but is not" will become a compiler warning some day...

rpv-tomsk:
[...]

PR #2392, opened 2017-08-02
perl plugin: Pass plugin name back to perl plugin callback function

rpv-tomsk:
As proposed in #896, the registration name (the second argument of the current plugin_register) added as last argument of perl callback calls.

The PLUGIN_READ, PLUGIN_WRITE, PLUGIN_LOG, PLUGIN_NOTIF, PLUGIN_FLUSH callback types are handled via C code.

The PLUGIN_INIT, PLUGIN_SHUTDOWN, PLUGIN_CONFIG types are handled by Perl code of Collectd.pm.


rpv-tomsk:
cc @UnitedMarsupials

PR #2397, opened 2017-08-14
Plugin for measuring thermal throttle count

notbaab:
This plugin allows you to monitor the core and package throttle count on Linux machines by looking in the /sys/devices/system/cpu/cpuN/thermal_throttle/ directory for all N cores on the machine.

### DIFF
--- a/src/thermal_throttle.c
+++ b/src/thermal_throttle.c
@@ -1,0 +1,100 @@
+/**
+ * collectd - src/thermal_throttle.c
+ * Copyright (C) 2017      notbaab
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; only version 2 of the License is applicable.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ *
+ * Authors:
+ *   notbaab <notbaab at gmail.com>
+ **/
+
+#include "common.h"
+#include "plugin.h"
+#include "collectd.h"
+
+static int num_cpu = 0;
+
+static int thermal_throttle_init() {
+  int status;
+  char filename[256];
+  num_cpu = 0;
+
+  while (1) {
+    status = snprintf(filename, sizeof(filename),
+                      "/sys/devices/system/cpu/cpu%d/thermal_throttle/"
+                      "core_throttle_count",
+                      num_cpu);
+
+    if ((status < 1) || (unsigned int)status >= sizeof(filename)) {
+      break;
+    }
+
+    if (access(filename, R_OK)) {
+      break;
+    }
+
+    num_cpu++;
+  }
+  return 0;
+}
+
+static void thermal_throttle_submit(int cpu_num, value_t v_core,
+                                    value_t v_package) {
+  value_list_t v1 = VALUE_LIST_INIT;
+  value_t values[] = {v_core, v_package};
+  v1.values = values;
+  v1.values_len = 2;
+
+  sstrncpy(v1.plugin, "thermal_throttle", sizeof(v1.plugin));
+  sstrncpy(v1.type, "thermal_throttle", sizeof(v1.type));
+  snprintf(v1.type_instance, sizeof(v1.type_instance), "%i", cpu_num);
+
+  plugin_dispatch_values(&v1);
+}
+
+static int thermal_throttle_read(void) {
+  for (int i = 0; i < num_cpu; i++) {
+    char filename[1024];
+    snprintf(
+        filename, sizeof(filename),
+        "/sys/devices/system/cpu/cpu%d/thermal_throttle/core_throttle_count",
+        i);
+
+    value_t v_core;
+    if (parse_value_file(filename, &v_core, DS_TYPE_COUNTER) != 0) {
+      WARNING("thermal_throttle plugin: Reading \"%s\" failed.", filename);
+      continue;
+    }
+
+    snprintf(
+        filename, sizeof(filename),
+        "/sys/devices/system/cpu/cpu%d/thermal_throttle/package_throttle_count",
+        i);
+
+    value_t v_package;
+    if (parse_value_file(filename, &v_package, DS_TYPE_COUNTER) != 0) {
+      WARNING("thermal_throttle plugin: Reading \"%s\" failed.", filename);
+      continue;
+    }
+
+    thermal_throttle_submit(i, v_core, v_package);
+  }
+
+  return 0;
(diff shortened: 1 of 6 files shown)


src/thermal_throttle.c
```diff
@@ -0,0 +1,100 @@
+/**
+ * collectd - src/thermal_throttle.c
+ * Copyright (C) 2017      notbaab
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
```
octo: The project as a whole is moving more and more towards the [MIT license](https://opensource.org/licenses/MIT). Please consider if you'd be willing to license your work under that license, too.

src/thermal_throttle.c
```diff
@@ -0,0 +1,100 @@
...
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ *
+ * Authors:
+ *   notbaab <notbaab at gmail.com>
+ **/
+
+#include "common.h"
+#include "plugin.h"
+#include "collectd.h"
```
octo: The `"collectd.h"` include must be the first.

src/thermal_throttle.c
```diff
@@ -0,0 +1,100 @@
...
+
+static int num_cpu = 0;
+
+static int thermal_throttle_init() {
```
octo: I'd prefer to do the `if (access() == 0) { break; }` check in the main read function, rather than counting the (expected) number of CPUs at startup. The number of active CPUs may vary over the runtime of the daemon.

src/thermal_throttle.c
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    num_cpu++;
+  }
+  return 0;
+}
+
+static void thermal_throttle_submit(int cpu_num, value_t v_core,
+                                    value_t v_package) {
+  value_list_t v1 = VALUE_LIST_INIT;
```
octo: s/v1/vl/

"vl" is short for "**v**alue **l**ist".

src/types.db
```diff
@@ -235,6 +235,7 @@ swap_io                 value:DERIVE:0:U
 tcp_connections         value:GAUGE:0:4294967295
 temperature             value:GAUGE:U:U
 threads                 value:GAUGE:0:U
+thermal_throttle        core_throttle_count:COUNTER:0:U, package_throttle_count:COUNTER:0:U
```
octo: We require new types to have a single *data source* only and that that *data source* be named "value". Also, `COUNTER` must only be used for 32bit counters that overflow frequently (more than once per hour). All other cumulative metrics must use `DERIVE`.

I recommend to create a new type like this:

```
total_throttles value:DERIVE:0:U
```

Then dispatch your metrics as

*   thermal_throttle-cpu0/total_throttles-core
*   thermal_throttle-cpu0/total_throttles-package
*   …
*   thermal_throttle-cpu*n*/total_throttles-core
*   thermal_throttle-cpu*n*/total_throttles-package

In other words, set `vl.plugin_instance` to `"cpu%d"` and `vl.type_instance` to either `"core"` or `"package"`.

src/thermal_throttle.c
```diff
@@ -0,0 +1,100 @@
...
+  sstrncpy(v1.plugin, "thermal_throttle", sizeof(v1.plugin));
+  sstrncpy(v1.type, "thermal_throttle", sizeof(v1.type));
+  snprintf(v1.type_instance, sizeof(v1.type_instance), "%i", cpu_num);
+
+  plugin_dispatch_values(&v1);
+}
+
+static int thermal_throttle_read(void) {
+  for (int i = 0; i < num_cpu; i++) {
+    char filename[1024];
```
octo: 1024 is arbitrary and too large. You can either use `PATH_MAX` (still large, but not arbitrary) or use a reasonably sized buffer. For example:

You need at most 88 bytes (`"/sys/devices/system/cpu/cpu18446744073709551615/thermal_throttle/package_throttle_count"`). Incidentally that's a size the compiler can nicely align on 64bit systems.

notbaab:
These are all good points and I will address them soon in another pull request, thanks!

src/thermal_throttle.c
octo: Yes, that's what I'm recommending.

If that's the only thing you do in the init function, then you can remove it entirely – it's not mandatory.

Best regards,
—octo

notbaab:
I believe I've address most of the issues. I look forward to your feedback.

notbaab:
Is the build error an issue with my code?  Clicking on the details it seems to be a java problem but I'm not sure if I'm reading the output correctly.

rpv-tomsk:
Several cores belongs to one package.
So, I think it is not reasonable to submit stats for package in each its cores.

rpv-tomsk:
cc @mrunge 

I have such case:
```
# uname -a
Linux host1 3.2.0-4-amd64 #1 SMP Debian 3.2.78-1 x86_64 GNU/Linux

# dmesg |grep CPU0
[    0.044605] CPU0: Thermal monitoring enabled (TM1)
[    0.089955] CPU0: Intel(R) Xeon(R) CPU E5-2430 0 @ 2.20GHz stepping 07
[332835.628811] CPU0: Core power limit notification (total events = 1)
[332835.628991] CPU0: Package power limit notification (total events = 1)
[332837.132346] CPU0: Core power limit normal
[332837.132380] CPU0: Package power limit normal
[2852154.804880] CPU0: Core power limit notification (total events = 2)
[2852154.804964] CPU0: Package power limit notification (total events = 2)
[2852154.805418] CPU0: Core power limit normal
[2852154.805569] CPU0: Package power limit normal

# ls /sys/devices/system/cpu/cpu0/thermal_throttle/
core_power_limit_count  core_throttle_count  package_power_limit_count  package_throttle_count

# cat /sys/devices/system/cpu/cpu0/thermal_throttle/core_power_limit_count
8

# cat /sys/devices/system/cpu/cpu0/thermal_throttle/core_throttle_count
0

# cat /sys/devices/system/cpu/cpu0/thermal_throttle/package_power_limit_count
8

# cat /sys/devices/system/cpu/cpu0/thermal_throttle/package_throttle_count
0
```
Is these counters incremented once per limit occurence, or they are incremented during it? At which rate?

Is there is a (known theoretical) way to get such Core/Pacakge notifications as Collectd notifications? (via Netlink? Any examples?)
I expect to get recovery notifications too.

mrunge:
I would think it *is* useful to differentiate cpu cores, i.e. read out core_throttle_count by cpu* and also to read package_throttle_count by cpu core. 

Looking through the multi-processor machines I have access to: the number of linux cpus seem to alternate between physical processors and then cores. As example, cpu0 is core0 on physical cpu A cpu1 is core 0 on physical cpu B.
Let me put the question the other way around: is there a good reason to accumulate the readout here? With process pinning to (linux) cpus, I would like to see the output per cpu.

sunkuranganath:
On few generations of CPUs, sequential core numbers alternate between sockets, while on few generation CPUs core numbers are sequential on same socket before increment'ing on to next available socket.
Given there are 2 separate counters provided it is prudent to provide out 2 individual metrics for each of core_throttle_count and package_throttle_count.

@rpv-tomsk in relation to your question on notification requirement based on these metrics, I am checking with experts to see if there are any guidelines. Alternative would be scour through 600+ page of Intel software developer's manual.

PR #2398, opened 2017-08-16
intel_pmu: add metadata with information about scaling

serhiyx:
In case reported counters were scaled metadata contains information basing on which scaled value was calculated. 
If there are more events than counters, the kernel uses time multiplexing. With multiplexing, at the end of the run, the counter is scaled basing on total time enabled vs time running: final_count = raw_count * time_enabled/time_running

### DIFF
--- a/src/intel_pmu.c
+++ b/src/intel_pmu.c
@@ -285,22 +285,20 @@ static void pmu_submit_counter(int cpu, char *event, counter_t value,
   plugin_dispatch_values(&vl);
 }
 
-meta_data_t *pmu_event_get_meta(struct event *e, int cpu) {
+meta_data_t *pmu_meta_data_create(const struct efd *efd) {
   meta_data_t *meta = NULL;
 
   /* create meta data only if value was scaled */
-  if (e->efd[cpu].val[1] != e->efd[cpu].val[2] && e->efd[cpu].val[2]) {
+  if (efd->val[1] != efd->val[2] && efd->val[2]) {
     meta = meta_data_create();
     if (meta == NULL) {
       ERROR(PMU_PLUGIN ": meta_data_create failed.");
       return NULL;
     }
 
-    meta_data_add_unsigned_int(meta, "intel_pmu:raw_count", e->efd[cpu].val[0]);
-    meta_data_add_unsigned_int(meta, "intel_pmu:time_enabled",
-                               e->efd[cpu].val[1]);
-    meta_data_add_unsigned_int(meta, "intel_pmu:time_running",
-                               e->efd[cpu].val[2]);
+    meta_data_add_unsigned_int(meta, "intel_pmu:raw_count", efd->val[0]);
+    meta_data_add_unsigned_int(meta, "intel_pmu:time_enabled", efd->val[1]);
+    meta_data_add_unsigned_int(meta, "intel_pmu:time_running", efd->val[2]);
   }
 
   return meta;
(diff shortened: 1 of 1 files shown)


src/intel_pmu.c
```diff
@@ -283,6 +285,27 @@ static void pmu_submit_counter(int cpu, char *event, counter_t value) {
   plugin_dispatch_values(&vl);
 }
 
+meta_data_t *pmu_event_get_meta(struct event *e, int cpu) {
```
maryamtahhan: it would be great to rename this function to pmu_event_set_meta rather than get... but other than that this looks good to me

src/intel_pmu.c
```diff
@@ -297,17 +320,30 @@ static void pmu_dispatch_data(void) {
...
       all_value += value;
 
+      /* get meta data with information about scaling */
+      meta_data_t *meta = pmu_event_get_meta(e, i);
+
       /* dispatch per CPU value */
-      pmu_submit_counter(i, e->event, value);
+      pmu_submit_counter(i, e->event, value, meta);
+
+      if (meta) {
```
vmytnykx: the `meta_data_destroy` handles the null ptr, so the `if` statement is not required here.

src/intel_pmu.c
vmytnykx: const struct event *e

octo: Typically we call constructors (functions that allocate, initialize and return a `foo_t`) `foo_create`. The `pmu` prefix is fine, which would make this `pmu_meta_data_create`.

octo: You only ever seem to access `e->efd[cpu]` – can you pass that instead of `e` and `cpu` individually?

src/intel_pmu.c
```diff
@@ -285,22 +285,20 @@ static void pmu_submit_counter(int cpu, char *event, counter_t value,
   plugin_dispatch_values(&vl);
 }
 
-meta_data_t *pmu_event_get_meta(struct event *e, int cpu) {
+meta_data_t *pmu_meta_data_create(const struct efd *efd) {
   meta_data_t *meta = NULL;
 
   /* create meta data only if value was scaled */
-  if (e->efd[cpu].val[1] != e->efd[cpu].val[2] && e->efd[cpu].val[2]) {
+  if (efd->val[1] != efd->val[2] && efd->val[2]) {
```
octo: Please invert the logic to remove one level of indentation:

```c
if (efd->val[1] == efd->val[2] || !efd->val[2]) {
  return NULL;
}

meta_data_t *meta = meta_data_create();
/* … */
```

src/intel_pmu.c
octo: Just checking that you saw this comment: there is no need to check if `meta` is non-NULL and imho the code is cleaner without the check, but the decision is up to you.

(A good practice is to reply something along the lines of "ACK; I prefer to keep as-is" to optional comments to show that you considered them.)

src/intel_pmu.c
serhiyx: Ok

src/intel_pmu.c
serhiyx: Ok
Wanted to keep as-is but if two reviewers say the code will be cleaner I agree :)

octo:
I manually triggered a build all is well: https://ci.collectd.org/job/pull-requests-manual_trigger-aggregation/248/

Issue #2404, opened 2017-08-17
failing on dpdk pkg-config check if no pkg-config is available


cpaelzer:
*   Version of collectd: head
*   Operating system / distribution: Ubuntu/Debian

## Expected behavior

Build without pkg-config

## Actual behavior

Fails since PR 2400

## Steps to reproduce

* build

## Background and solution

Due to a recent change (my fault in [PR 2400](https://github.com/collectd/collectd/pull/2400)) the Ci is now failing all the time.
See for example this [CI log](https://ci.collectd.org/job/pull-requests-prepare-tarball/1844/console)

```
checking for DPDK... no
configure: error: Package requirements (libdpdk) were not met:

No package 'libdpdk' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables DPDK_CFLAGS
and DPDK_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
Build step 'Execute shell' marked build as failure
Archiving artifacts
Finished: FAILURE

```

The reason is that while working correctly in the pkg-config case no matter if old or new dpdk headers are used.
It has issues if no pkg-config is provided at all.
The reason for that is that the macro for the check errors out if there is no definition for the "else" part.
The fix is easy and a PR will be here soon.

cpaelzer:
The linked PR fixes the case for me, please accept my excuse for breaking it in the first place and take a look.

rubenk:
Fixed by 9e6bf7d

PR #2405, opened 2017-08-17
configure.ac: dpdk: tolerate no pkg-config

cpaelzer:
Since pull request #2400 pkg-config works great for dpdk, but not if
there is no pkg-config around at all.
To fix that the PKG_CHECK_MODULES macro needs to define a valid fourth
argument, otherwise it will error out.

This closes 2404

Signed-off-by: Christian Ehrhardt <christian.ehrhardt@canonical.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2361,7 +2361,7 @@ AC_ARG_WITH([libdpdk],
 
 if test "x$with_libdpdk" != "xno"; then
   PKG_CHECK_MODULES([DPDK], [libdpdk], [],
-		    [AC_MSG_NOTICE([no DPDK pkg-config, using defaults])])
+                    [AC_MSG_NOTICE([no DPDK pkg-config, using defaults])])
   if test "x$LIBDPDK_CPPFLAGS" = "x"; then
     LIBDPDK_CPPFLAGS="-I/usr/include/dpdk"
   fi
--- a/configure.ac
+++ b/configure.ac
@@ -2361,7 +2361,7 @@ AC_ARG_WITH([libdpdk],
 
 if test "x$with_libdpdk" != "xno"; then
   PKG_CHECK_MODULES([DPDK], [libdpdk], [],
-                    [AC_MSG_NOTICE([no DPDK pkg-config, using defaults])])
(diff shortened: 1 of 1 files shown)


cpaelzer:
Sorry once again, on the former PR the first CI check was good so none of us took the time to evaluate the former ones. Fortunately I've seen this and can provide the fix right away to clean up my mess.

cpaelzer:
Closes #2404 

I forgot the # in the commit message, let me know if I shall fix that.

cpaelzer:
Ok, the comment linked it up correctly I think.

cpaelzer:
With the fix in place in a case without pkg-config for dpdk it will now go like:

```
checking for DPDK... no
configure: no DPDK pkg-config, using defaults
checking rte_config.h usability... yes
checking rte_config.h presence... yes
checking for rte_config.h... yes
checking for rte_eal_init in -ldpdk... yes
```

That is so much better than aborting

configure.ac
```diff
@@ -2360,7 +2360,8 @@ AC_ARG_WITH([libdpdk],
 )
 
 if test "x$with_libdpdk" != "xno"; then
-  PKG_CHECK_MODULES([DPDK], [libdpdk])
+  PKG_CHECK_MODULES([DPDK], [libdpdk], [],
+		    [AC_MSG_NOTICE([no DPDK pkg-config, using defaults])])
```
rubenk: Please use 2 spaces for indentation.

cpaelzer: Fixing to spaces, but the actual indent should be aligned with the [DPDK above right?

rubenk: No it shouldn't, the rest of configure.ac also uses 2 spaces.

cpaelzer:
Pushed an indent fix as requested.
I made the actual change in a container which has none of my usual setup which would have made me aware of the tab earlier - thanks for spotting it @rubenk

configure.ac
cpaelzer: My re-upload and this comment raced by seconds, pushing another fix to make it 2 as requested.

cpaelzer:
Commit added following 2 spaces now.

PR #2417, opened 2017-09-01
configure.ac: dpdk: use pkg-config - 5.7

bluca:
To detect cflags and libs use the sometimes provided pkg-config for
libdpdk. That avoids build errors on systems where special flags are
needed and provided by dpdk via pkg-config, but not yet considered by
the collectd build system.

Backport of https://github.com/collectd/collectd/pull/2400 and https://github.com/collectd/collectd/pull/2405 tested on Debian.


bluca:
Ping :-)

octo:
LGTM, thanks @bluca!

Issue #2421, opened 2017-09-06
Feature Request: command line flag to exit if BaseDir does not exist


nwf:
*   Version of collectd: 5.7.2
*   Operating system / distribution: LEDE

I run collectd on my router and persist the data sets (rrds) across reboot by having them written to a USB stick.  However, early in boot, collectd sometimes (often? always?) manages to come up before the mount completes, and so touches the on-router flash, creating its basedir and datadir and the rrds and so on.  I would much prefer that it not do this and instead, optionally, abort if the basedir does not exist, rather than creating it.  In my particular case, it's fine to bail out because the procd supervisor will restart the daemon periodically, so even though procd lacks a notion of dependencies, this will work out great.

Thanks!

rpv-tomsk:
IMHO.
You try to solve issue with your init system by adding code into Collectd. 
I think that is a wrong way, and race in init system should be solved instead.

nwf:
There is no mechanism in OpenWRT's procd (the init system in question) to wait for mounts to come up, so far as I know.  I am in general agreement that it would be nice if it were there, but it's not.  The mounts are managed by the general hotplug infrastructure, whereas collectd is viewed as a general daemon.  I am sure I could find some way to munge around it, but...

The existing behavior of collectd, to create its data directories, is also a little atypical, AFAIK; most programs I can think of would be happy to create individual files, or directories within the specified base, but not the base directory itself, absent a directive to do so.  I recognize that it's a bit of a hackish workaround, but IMHO it represents more typical behavior and just happens to solve the init problem as well.

If collectd doesn't wish to take it, in light of the fact that it's motivated by LEDE's init system, I'll push the patch their way instead (and maybe that will motivate them to add path/mount dependencies to procd. ;) )

PR #2422, opened 2017-09-06
Add daemon option to avoid making BaseDir

nwf:
One possible implementation of collectd/collectd#2421

### DIFF
--- a/src/daemon/collectd.c
+++ b/src/daemon/collectd.c
@@ -158,7 +158,7 @@ static int init_global_variables(void) {
   return 0;
 } /* int init_global_variables */
 
-static int change_basedir(const char *orig_dir) {
+static int change_basedir(const char *orig_dir, int nocreate) {
   char *dir;
   size_t dirlen;
   int status;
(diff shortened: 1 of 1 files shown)


src/daemon/collectd.c
```diff
@@ -158,7 +158,7 @@ static int init_global_variables(void) {
   return 0;
 } /* int init_global_variables */
 
-static int change_basedir(const char *orig_dir) {
+static int change_basedir(const char *orig_dir, int nocreate) {
```
octo: Please use a positive logic here, "not nocreate" is a double negation and hard to read and I'd much prefer a `create` boolean.

Also, please use `_Bool` instead of `int`.

src/daemon/collectd.c
```diff
@@ -191,13 +191,15 @@ static int change_basedir(const char *orig_dir) {
...
   }
 
-  status = mkdir(dir, S_IRWXU | S_IRWXG | S_IRWXO);
-  if (status != 0) {
-    char errbuf[1024];
-    ERROR("change_basedir: mkdir (%s): %s", dir,
-          sstrerror(errno, errbuf, sizeof(errbuf)));
-    free(dir);
-    return -1;
+  if (nocreate == 0) {
```
octo: This tries the `chdir()` call a second time, which is not elegant. How about we extend the above logic to:

```c
else if (!create || (errno != ENOENT)) {
  //     ^^^^^^^^^^
  /* report error, return */
}
```

nwf: That's better, I agree.

src/daemon/collectd.c
nwf: Both done.

octo:
Thanks @nwf!

PR #2425, opened 2017-09-08
Add Redis mem_fragmentation_ratio metric

vbatoufflet:
Hi,

This PR adds support for `mem_fragmentation_ratio` metric in the Redis plugin.

Let me know if this change suits to you.

Regards,
Vincent

### DIFF
--- a/src/redis.c
+++ b/src/redis.c
@@ -441,6 +441,8 @@ static int redis_read(void) /* {{{ */
                       DS_TYPE_GAUGE);
     redis_handle_info(rn->name, rr->str, "memory_lua", NULL, "used_memory_lua",
                       DS_TYPE_GAUGE);
+    redis_handle_info(rn->name, rr->str, "memory_fragmentation_ratio", NULL,
+                      "mem_fragmentation_ratio", DS_TYPE_GAUGE);
     /* changes_since_last_save: Deprecated in redis version 2.6 and above */
     redis_handle_info(rn->name, rr->str, "volatile_changes", NULL,
                       "changes_since_last_save", DS_TYPE_GAUGE);
--- a/src/types.db
+++ b/src/types.db
@@ -1,280 +1,281 @@
-absolute                value:ABSOLUTE:0:U
-apache_bytes            value:DERIVE:0:U
-apache_connections      value:GAUGE:0:65535
-apache_idle_workers     value:GAUGE:0:65535
-apache_requests         value:DERIVE:0:U
-apache_scoreboard       value:GAUGE:0:65535
-ath_nodes               value:GAUGE:0:65535
-ath_stat                value:DERIVE:0:U
-backends                value:GAUGE:0:65535
-bitrate                 value:GAUGE:0:4294967295
-blocked_clients         value:GAUGE:0:U
-bucket                  value:GAUGE:0:U
-bytes                   value:GAUGE:0:U
-cache_eviction          value:DERIVE:0:U
-cache_operation         value:DERIVE:0:U
-cache_ratio             value:GAUGE:0:100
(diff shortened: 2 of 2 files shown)


src/redis.c
```diff
@@ -441,6 +441,8 @@ static int redis_read(void) /* {{{ */
                       DS_TYPE_GAUGE);
     redis_handle_info(rn->name, rr->str, "memory_lua", NULL, "used_memory_lua",
                       DS_TYPE_GAUGE);
+    redis_handle_info(rn->name, rr->str, "memory_fragmentation_ratio", NULL,
```
octo: We avoid such specific *types* and prefer to introduce more generic types that can be reused elsewhere. I'd suggest to add a new `ratio` type and set the *type instance* to `memory_fragmentation`. If the ratio is always in the [0-1] range, the existing `percent` type would also be an option.

vbatoufflet: Hi @octo,

Sure, it makes sense. I'll apply those change to the PR and let you now as soon it's done.

octo: Thank you!

dago:
@vbatoufflet Are you still interested in getting this merged? Would you mind adjusting the types as suggested by octo?

rpv-tomsk:
@dago, 
at first we really should discuss an idea of `We avoid such specific types and prefer to introduce more generic types`.

Suppose, one system (one plugin) reports several ratio values, and these ratios are unrelated (should be grouped into two independend groups). We need specific types then - so use of specific types is not bad idea in overall.

dago:
@rpv-tomsk How I understand it is that octo would prefer to differentiate by instance name instead of type. I guess I do not understand yet why you prefer to use different types instead of multiple instances of the same type.

rpv-tomsk:
From my POV, type is a metrics set which could be put on same chart because these metrics are from same group.

Suppose, a system like redis/memcached/mysql etc has two caches.
One cache is query cache and another is thread pool cache.

We have two metrics, hit rate for query cache and hit rate to thread pool. Both are `percent`.

These metrics are unrelated and typically should not be put on same chart. 
So, they should not have same type.

Why to put these two into same type?
I don't find any issue which 'use of generic types only' tries to solve.
Instead, you can bind configuration to `non-generic` type in your preferred webinterface and that is easier to find grouped metrics. This proposal proposes to remove a kind of useful information - which is stored in types.

I use rrdtool plugin and simple web interface `collection3` which builds nice charts from a box with minimal setup (I have >100 servers of different types). 
It simply works. I even use `collection3` on server where Graphana installed for developers.

If you will put several multiple instance-based metrics for metrics from different groups, then `collection3` will put all them on same chart which will be incorrect result.  
Also, it will be harder to find which metrics are grouped, if you use another interface.

Ok, this example is not so good, as there is only two metrics.
But suppose we have a service which has incoming connections and several groups of outgoing connections, and all connections could be in different states.
Using logic, proposed here, all these metrics will have same type `connections`, while metrics are unrelated. What is incorrect.

I found no reasoning why do not use custom types.

rpv-tomsk:
A very good example of this approach disadvantages is `protocols.c` plugin which reports all its values as  `protocol_counter` type.

I found that plugin completely unuseful.

dago:
@rpv-tomsk I suggest to move the architectural discussion to the mailing list https://collectd.org/wiki/index.php/Mailing_list to increase the scope of involved people. Would you mind reposting to the list?

rpv-tomsk:
>Would you mind reposting to the list?

No, I have no much time for that, sorry.

vbatoufflet:
Closing this PR as I no longer have a need for this change. Hopefully this helps clean up the backlog a bit :)

PR #2430, opened 2017-09-15
nfs plugin: Add config options to ignore versions

usev6:


usev6:
At my day job I was looking for a way to avoid metrics for NFSv2 hitting a loaded graphite server, today. Looking at src/nfs.c, I thought, I might as well open a PR for this functionality.

This is my first PR for collectd, so please have patience and review carefully ;-)

src/collectd.conf.pod
```diff
@@ -5166,6 +5166,25 @@ statistics available. Defaults to B<false>.
...
+The I<nfs plugin> collects information about the usage of the Network File
+System (NFS). It counts the number of procedure calls for each procedure,
+grouped by version and whether the system runs as server or client.
+
+It is possibly to omit metrics for a specific NFS version by setting one or
+more of the following options to B<true> (all of them default to B<false>). 
+
+=over 4
+
+=item B<IgnoreV2> B<false>|B<true>
```
octo: We avoid using negated config options, because the resulting double-negations ("ignore: false!") are unnecessarily hard to read. The typical naming for such options is "Report…", i.e. in your case `ReportV2`, etc.

src/nfs.c
```diff
@@ -31,6 +31,12 @@
 #include <kstat.h>
 #endif
 
+static const char *config_keys[] = {"IgnoreV2", "IgnoreV3", "IgnoreV4"};
+static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
+static int ignore_v2 = 0;
```
octo: These variables should be of type `_Bool`.

src/nfs.c
```diff
@@ -295,6 +301,19 @@ static kstat_t *nfs4_ksp_client;
 static kstat_t *nfs4_ksp_server;
 #endif
 
+static int nfs_config(const char *key, const char *value) {
+  if (strcasecmp(key, "IgnoreV2") == 0 && IS_TRUE(value))
+    ignore_v2 = 1;
```
octo: Please just assign the valur of `IS_TRUE()`, i.e. `ignore_v2 = IS_TRUE(value)`. This allows (convoluted but correct) configs such as

```
<Plugin nfs>
  IgnoreV2 true
  IgnoreV2 false
</Plugin>
```

src/nfs.c
```diff
@@ -495,18 +514,18 @@ static void nfs_read_linux(FILE *fh, const char *inst) {
     if (fields_num < 3)
       continue;
 
-    if (strcmp(fields[0], "proc2") == 0) {
+    if (strcmp(fields[0], "proc2") == 0 && ignore_v2 == 0) {
```
octo: With boolean variables (`_Bool`), please use `!ignore_v2` instead. Here and below.

usev6:
Hi @octo ,

thanks a lot for your comments! I changed my PR accordingly and I fully agree that the resulting code is clearer.

Best regards,
usev6

Issue #2424, opened 2017-09-07
Incorrect uptime reported for lxc/lxd containers


zerkms:
*   Version of collectd: master
*   Operating system / distribution: ubuntu 16.04

the `uptime` plugin reads from `"/proc/stat"` on linux, while under lxs/lxd that reports the host machine statistics. How about using `/proc/uptime` (as `uptime` tool does).

zerkms:
Btw, I can contribute the PR if everybody agrees to use `/proc/uptime`, which I think is a proper source for the uptime info.

octo:
SGTM, thanks @zerkms

PR #2431, opened 2017-09-18
Reimplemented `uptime` plugin to read from `/proc/uptime`

zerkms:
It's a fix for the https://github.com/collectd/collectd/issues/2424

I changed implementation to read from `/proc/uptime` instead of `/proc/stat`.

### DIFF
--- a/src/uptime.c
+++ b/src/uptime.c
@@ -25,7 +25,7 @@
 #include "plugin.h"
 
 #if KERNEL_LINUX
-#define STAT_FILE "/proc/stat"
+#define UPTIME_FILE "/proc/uptime"
 /* Using /proc filesystem to retrieve the boot time, Linux only. */
 /* #endif KERNEL_LINUX */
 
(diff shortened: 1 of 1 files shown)


src/uptime.c
```diff
@@ -86,46 +86,38 @@ static int uptime_init(void) /* {{{ */
...
 
-  while (fgets(buffer, 1024, fh) != NULL) {
-    /* look for the btime string and read the value */
-    ret = sscanf(buffer, "btime %lu", &starttime);
-    /* avoid further loops if btime has been found and read
-     * correctly (hopefully) */
-    if (ret == 1)
-      break;
-  }
+  ret = fscanf(fh, "%lf", &uptime_seconds);
```
octo: Please use the `parse_value_file()` function from `"common.h"` (`src/daemon/common.h`).

zerkms: @octo `uptime_seconds` is an intermediate variable used to calculate the static `boottime = time(NULL) - (long)uptime_seconds;`. Do you still think it's relevant?

zerkms: The original code I replaced was

```c
  while (fgets(buffer, 1024, fh) != NULL) { 
    /* look for the btime string and read the value */ 
    ret = sscanf(buffer, "btime %lu", &starttime); 
    /* avoid further loops if btime has been found and read 
     * correctly (hopefully) */ 
    if (ret == 1) 
      break; 
  } 
```

octo: TL;DR: `fscanf()` if fine here, I was confused about the content of `/prod/uptime`.

Previously, the code scanned for a specific line which had the `btime` prefix. Your new code appeared to read a file that contained a *single value* only – which is a common use case and we have the `parse_value_file()` function for it.

Looking at `/proc/uptime` though, it appears that there are two values and the convenience function cannot be used (it complains when there are trailing characters). `fscanf()` is perfectly fine for dealing with this. I briefly pondered whether we should read the second value into an anonymous var just to make it explicit that we're ignoring it, but I think that would make the code harder to read rather than easier.

zerkms: Yep, now I see what you mean. So, do you want me to change something else then? I'm a bit confused whether my implementation is fine or not. To me - it's okay'ish :-)

octo:
Thank you very much, @zerkms!

zerkms:
@octo and as soon as you're here could you please have a look at https://github.com/collectd/collectd/pull/2376

It's literally just 2 lines of code.

Issue #901, opened 2015-01-20
Feature Request: needs for "total" memory in plugin memory


toni-moreno:
I've seen in the code you are computing memory used from Total

mem_used = mem_total - (mem_free + mem_buffered + mem_cached + mem_slab_total);

and finally  the total memory are not output written.

https://github.com/collectd/collectd/blob/master/src/memory.c#L351-L363

INHO when working with VM's and when you need %used and also the size  it is important to have  the total.

dennisjac:
This calculation is really a kludge. On modern Systems you can use MemAvailable to get a better representation of "(mem_free + mem_buffered + mem_cached + mem_slab_total)"

[root@haproxy1-lb1 ~]# cat /proc/meminfo 
MemTotal:        3883056 kB
MemFree:         2339100 kB
MemAvailable:    2648232 kB
...

Link to commit with explanation:
https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773

TLDR extract:
"User space really should not be expected to know kernel internals to come up with an estimate for the amount of free memory. It is more convenient to provide such an estimate in /proc/meminfo. If things change in the future, we only have to change it in one place."

toni-moreno:
hi @dennisjac What I'm asking for is MemTotal of /proc/meminfo  not available at the output plugin , this is "phisical / installed RAM".

I can not understand  what are you referring in your last comment

sundarams73tech:
@toni-moreno 

I'm also looking for the total memory to make the graph complete. 

Do you found any work around for this ?

TeChn4K:
Total memory info would be great

ntmggr:
+1

chubutin:
This would be nice to have it.

utkarshcmu:
+1

purplefish32:
+1

neutronsharc:
I also need this feature:  "memory" plugin reports "total amount of memory".

Apollon77:
Is there any workaround? Or manual way?

StephenPCG:
+1

rralleres:
What about using aggregation plugin? It looks like it would be possible to get the total memory from a calculated sum from the rest of the values reported by memory

StephenPCG:
@rralleres No, the values reported by memory does not sum up to total.

PR #2436, opened 2017-09-21
memory: add total for Linux

maage:
This should address: #901 

Strictly not needed as:

https://github.com/maage/collectd/blob/master/src/memory.c#L325

and total can be then summed from these

memory-{used,free,buffered,cached,slab_total}

But anyways it is kind of hard to go to right place into sourcefile and search exact formula and then implement the sum.

As plugin outputs absolute and percent, I needed to write out macro, or percent would not be okay.

Patch is tested only on modern Linux with detailed_slab_info.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -329,13 +329,31 @@ static int memory_read_internal(value_list_t *vl) {
    * They sum up to the value of Slab, which is available on older & newer
    * kernels. So SReclaimable/SUnreclaim are submitted if available, and Slab
    * if not. */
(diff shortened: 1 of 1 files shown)


PrestonR:
Bump. What's the status on this?

drNowak:
Ping, we are waiting for this. Otherwise we will have to patch this change to our own copy of collectd.

Krzysztof

dago:
@maage @PrestonR @drNowak Would any of you gentlemen mind having a look at the issue addressed from octo?

PR #2440, opened 2017-09-22
write_redis plugin: Add max_set_duration to set duration for value

s1061123:
This change introduces max_set_duration, the duration for the value
as max_set_size. max_set_size can keep the db size, however, sometime
max_set_size allows to contain old data. max_set_duration can keep
the latest data.

### DIFF
--- a/src/write_redis.c
+++ b/src/write_redis.c
@@ -45,6 +45,7 @@ struct wr_node_s {
   char *prefix;
   int database;
   int max_set_size;
+  int max_set_duration;
   _Bool store_rates;
 
   redisContext *conn;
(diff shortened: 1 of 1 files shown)


src/write_redis.c
```diff
@@ -125,6 +126,16 @@ static int wr_write(const data_set_t *ds, /* {{{ */
       freeReplyObject(rr);
   }
 
+  if (node->max_set_duration > 0) {
+    rr = redisCommand(node->conn, "ZREMRANGEBYSCORE %s -1 (%d", key,
```
octo: The not-matched open parenthesis in `(%d` looks suspicious. If this in on purpose, please add a comment explaining it.

s1061123: @octo, Thank you for your comment! '(%d' means 'less than %d' of redis score. I will update it soon.

s1061123: @octo I've added comments in latest diff. Thanks again!

s1061123:
Missing comment for documenting (src/collectd.conf.pod and src/collectd.conf.in), hence I will file another pull request for it.

Issue #2433, opened 2017-09-19
Ability to disable swap io.


zerkms:
lxc/lxd containers at the moment cannot properly report the swap io. Since the corresponding `/proc/vmstat` are not virtualised. Relevant bug report: https://github.com/lxc/lxcfs/issues/212

It would be ideal if `swap` plugin had a boolean switch to disable swap io in case if it's not needed.

As usually: if you think it's acceptable and a good idea - I might implement it.

Thanks.

octo:
SGTM, thanks @zerkms! `ReportIO` would be a good (i.e. consistent) name for such a config option.

PR #2441, opened 2017-09-23
Added `ReportIO` option to the `swap` plugin

zerkms:
Default value: true

Usage:

```
<Plugin "swap"> 
  ReportIO false 
</Plugin> 
```

This would disable reporting the Swap IO.

This is an implementation for #2433

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1344,6 +1344,7 @@
 #	ReportBytes true
 #	ValuesAbsolute true
 #	ValuesPercentage false
+#	ReportIO true
 #</Plugin>
 
 #<Plugin table>
@@ -1479,7 +1480,7 @@
 #	SystemManagementInterrupt true
 #	DigitalTemperatureSensor true
 #	PackageThermalManagement true
-#	RunningAveragePowerLimit "7"	
+#	RunningAveragePowerLimit "7"
 #</Plugin>
 
 #<Plugin unixsock>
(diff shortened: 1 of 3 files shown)


rpv-tomsk:
Documentation changes are required, but missing in this PR.
Can you please add them?

zerkms:
@rpv-tomsk I'm happy to, I couldn't find where the corresponding files are though.

rpv-tomsk:
Thanks for your updates.

IMHO:

- As for me, I don't like [to do/push] changes, not related to solved issue.
- Such minor changes as last commit can be 'amended' ('git commit --amend') and, if needed, pushed forcibly (`push -f`). 

All this PR should consist of only one commit, aggregating all changes.
In this case, there is no benefits from splitting PR to separate commits.

That is mine 'IMHO' only. 
Thanks for your work.

zerkms:
Indeed. Squashed and replaced with a single commit

rpv-tomsk:
Also you might to add `Closes: #2433` into your commit message. 
That will close the related issue when the patch is merged.

If you will change the commit message, it's worth noting the reason of your work - add a message about lxc. Also, in most of cases plugin name is put at the beginning of the message header. (I'm unsure about 'plugin' word after name).

So, the better one (IMHO) message:

```
swap plugin: Added `ReportIO` option

lxc/lxd containers at the moment cannot properly report the swap io (the corresponding /proc/vmstat are not virtualized), so added an option to disable that reporting. Enabled by default.

Closes: #2433
```

octo:
LGTM, thanks @zerkms!

Also thanks to @rpv-tomsk for doing the review!

Issue #2443, opened 2017-09-24
turbostat plugin: linux >= 4.12 no longer provides <asm/msr-index.h>


Whissi:
Since kernel commit 25dc1d6cc3082aab293e5dad47623b550f7ddd2a ("x86: stop exporting msr-index.h to userland"), <asm/msr-index.h> is no longer exported to userspace. Therefore, any toolchain built with kernel headers >= 4.12 will no longer have this header file, causing a build failure in collectd.

octo:
Hi @Whissi, thanks for reporting this! The build should obviously not fail, but it sounds like we need to disable this plugin for newer kernels. Do you know if there's an approved way to read those metrics?

Best regards,
—octo

octo:
Hi @Whissi,

the *turbostat plugin* already depends on `<asm/msrindex.h>` and is disabled if the header file is not found. Could you please provide the build failure you're observing so I can see what's going on?

Thanks and best regards,
—octo

octo:
CC: @Feandil, @jdelvare You worked on the *turbostat plugin* before, do you have any insight into what the kernel folks are doing? Thanks!

Whissi:
Well, collectd's configure stops before trying to build:
```
 * Package:    app-admin/collectd-5.7.2
 * Repository: gentoo
 * Maintainer: whissi@gentoo.org
 * Upstream:   https://github.com/collectd/collectd/issues
 * USE:        abi_x86_64 amd64 collectd_plugins_turbostat debug elibc_glibc filecaps kernel_linux python_single_target_python3_4 python_targets_python2_7 python_targets_python3_4 python_targets_python3_5 python_targets_python3_6 udev userland_GNU
 * FEATURES:   network-sandbox preserve-libs sandbox userpriv usersandbox
 * Package:    app-admin/collectd-5.7.2
 * Repository: gentoo
 * Maintainer: whissi@gentoo.org
 * Upstream:   https://github.com/collectd/collectd/issues
 * USE:        abi_x86_64 amd64 collectd_plugins_turbostat debug elibc_glibc filecaps kernel_linux python_single_target_python3_4 python_targets_python2_7 python_targets_python3_4 python_targets_python3_5 python_targets_python3_6 udev userland_GNU
 * FEATURES:   network-sandbox preserve-libs sandbox userpriv usersandbox
 * Determining the location of the kernel source code
 * Found kernel source directory:
 *     /usr/src/linux
 * Found kernel object directory:
[...]

PR #2446, opened 2017-09-25
turbostat: import msr-index header from linux

Feandil:
Since Linux 4.12, the kernel does not expose msr-index.h to the userland
anymore. This commit pulls and hardcode the required definitions.

Should fix https://github.com/collectd/collectd/issues/2443

### DIFF
--- a/src/msr-index.h
+++ b/src/msr-index.h
@@ -1,0 +1,74 @@
+/*
+ * Partial header file imported from the linux kernel
+ * (arch/x86/include/asm/msr-index.h)
+ * as it is not provided by the kernel sources anymore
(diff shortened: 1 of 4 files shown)


src/msr-index.h
```diff
@@ -0,0 +1,74 @@
+/*
```
octo: Please add a license header, presumably GPL since this is from the kernel.

Feandil: Sorry, didn't realize it was missing
I can't include a Copyright statement though, as there is none in the sources...

octo:
Thank you @Feandil!

PR #2449, opened 2017-09-26
 snmp plugin: Added res->errstat check

rpv-tomsk:
Collectd does not check for `res->errstat` value after
`snmp_sess_synch_response()` call. In case of error, there is no any data in
`res->variables` actually, but variables are tried to be processed as usual.
Suffix calculation will fail, so all subtrees will be marked as failed, not
only one subtree which caused an error.

The csnmp_instance_list_add() call will fail too, and, as result,
`csnmp_read_table` will finish it's work without any data submission.

The log message like
"snmp plugin: host HOSTNAME: csnmp_instance_list_add failed",
which is put into logs in this case, also has no enough diagnostic data.

Added code to proper check for `res->errstat` and to try to get available data.

Issue: #2291

### DIFF
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -1389,35 +1389,47 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
     }
 
     if (res->errstat != SNMP_ERR_NOERROR) {
-      if (res->errindex != 0 && res->errindex < oid_list_len) {
+      if (res->errindex != 0) {
         /* Find the OID which caused error */
         for (i = 1, vb = res->variables; vb != NULL && i != res->errindex;
              vb = vb->next_variable, i++)
           /* do nothing */;
+      }
 
-        char oid_buffer[1024] = {0};
-        snprint_objid(oid_buffer, sizeof(oid_buffer) - 1, vb->name,
-                      vb->name_length);
-        NOTICE("snmp plugin: host %s; data %s: OID `%s` failed: %s",
-               host->name, data->name, oid_buffer,
-               snmp_errstring(res->errstat));
+      if ((res->errindex == 0) || (vb == NULL)) {
+        ERROR("snmp plugin: host %s; data %s: response error: %s (%li) ",
+              host->name, data->name, snmp_errstring(res->errstat),
+              res->errstat);
+        status = -1;
+        break;
+      }
 
-        /* Skip that OID */
-        i = res->errindex - 1;
+      char oid_buffer[1024] = {0};
+      snprint_objid(oid_buffer, sizeof(oid_buffer) - 1, vb->name,
+                    vb->name_length);
+      NOTICE("snmp plugin: host %s; data %s: OID `%s` failed: %s", host->name,
+             data->name, oid_buffer, snmp_errstring(res->errstat));
+
+      /* Calculate value index from todo list and skip OID found */
+      i = 0;
+      size_t j = 1;
+      for (;;) {
         while ((i < oid_list_len) && !oid_list_todo[i])
           i++;
 
-        oid_list_todo[i] = 0;
+        if (j == res->errindex)
+          break;
 
-        snmp_free_pdu(res);
-        res = NULL;
-        continue;
+        i++;
+        j++;
       }
 
-      ERROR("snmp plugin: host %s; data %s: response error: %s (%li) ",
-            host->name, data->name, snmp_errstring(res->errstat), res->errstat);
-      status = -1;
-      break;
+      assert(i < oid_list_len);
+      oid_list_todo[i] = 0;
+
+      snmp_free_pdu(res);
+      res = NULL;
+      continue;
     }
 
     for (vb = res->variables, i = 0; (vb != NULL);
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -1388,6 +1388,38 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
       break;
     }
 
+    if (res->errstat != SNMP_ERR_NOERROR) {
+      if (res->errindex != 0 && res->errindex < oid_list_len) {
+        /* Find the OID which caused error */
+        for (i = 1, vb = res->variables; vb != NULL && i != res->errindex;
+             vb = vb->next_variable, i++)
+          /* do nothing */;
+
+        char oid_buffer[1024] = {0};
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
@octo, 
I found no more elegant solution than such cycle:
https://github.com/collectd/collectd/pull/2449/files#diff-15314dd4787306bec3b25576017b94b3R1413 

Please look and say what you think about this.

This new code was carefully debugged. 
Reproduced error conditions for previous implementation (https://github.com/collectd/collectd/pull/2449/commits/9816488b3c779b85e285956e8a920bfceac43840) to fail.
New one works correctly.

src/snmp.c
```diff
@@ -1388,6 +1388,50 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
...
+      }
+
+      char oid_buffer[1024] = {0};
+      snprint_objid(oid_buffer, sizeof(oid_buffer) - 1, vb->name,
+                    vb->name_length);
+      NOTICE("snmp plugin: host %s; data %s: OID `%s` failed: %s", host->name,
+             data->name, oid_buffer, snmp_errstring(res->errstat));
+
+      /* Calculate value index from todo list and skip OID found */
+      i = 0;
```
octo: One way to simplify this might be with a temporary array:

```c
size_t oid_active[oid_list_len] = {0};
size_t oid_active_len = 0;

for (size_t i = 0; i < oid_list_len; i++) {
  if (!oid_list_todo[i]) {
    continue;
  }
  oid_active[oid_active_len] = i;
  oid_active_len++;
}

assert(res->errindex <= oid_active_len);
size_t failed_index = oid_active[res->errindex - 1];
oid_list_todo[failed_index] = 0;
``` 

It's also quite long, but IMHO easier to read than two indexes that get incremented in two nested loops with conditionals …

rpv-tomsk: So, the same `oid_active` should be used in next cycle (https://github.com/collectd/collectd/pull/2449/files#diff-15314dd4787306bec3b25576017b94b3R1435) too? That will do things better, IMHO.

octo: Good eye! Yes, that sounds like an improvement. (There are many improvements to be made to the *snmp plugin's* code ;)

rpv-tomsk: Hmm, if use such array for 'next cycle' then 'two indexes' will appear there.
I prefer to leave 'next cycle' untouched.

Don't you think it is superfluous to form a temporary array for rare errors?
I think that this is not optimal. May be left that cycle untouched, without temporary array  but with two indexes?

src/snmp.c
```diff
@@ -1327,22 +1327,24 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
...
     if (req == NULL) {
       ERROR("snmp plugin: snmp_pdu_create failed.");
       status = -1;
       break;
     }
 
-    oid_list_todo_num = 0;
+    int oid_list_todo_num = 0;
+    size_t var_idx[oid_list_len];
+    memset(&var_idx, 0, oid_list_len * sizeof(size_t));
```
octo: `sizeof(var_idx)` should do what you need.

src/snmp.c
```diff
@@ -1327,22 +1327,24 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
...
-
     req = snmp_pdu_create(SNMP_MSG_GETNEXT);
     if (req == NULL) {
       ERROR("snmp plugin: snmp_pdu_create failed.");
       status = -1;
       break;
     }
 
-    oid_list_todo_num = 0;
+    int oid_list_todo_num = 0;
```
octo: Nit: please use `size_t` for array indexes.

rpv-tomsk:
Patch updated, amended.

src/snmp.c
```diff
@@ -1327,22 +1327,24 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
...
     if (req == NULL) {
       ERROR("snmp plugin: snmp_pdu_create failed.");
       status = -1;
       break;
     }
 
-    oid_list_todo_num = 0;
+    size_t oid_list_todo_num = 0;
+    size_t var_idx[oid_list_len];
+    memset(&var_idx, 0, oid_list_len * sizeof(var_idx));
```
octo: This writes past the end of `var_idx` and will corrupt the stack. What you need is:

```c
memset(var_idx, 0, sizeof(var_idx));
```

Sorry I wasn't clear enough about that.

rpv-tomsk: Sorry, I'm not a qualified-enough C-programmer.)

Thanks for catching this.

octo: I disagree, you're obviously plenty qualified ;) Thanks for fixing :)

rpv-tomsk: I'm glad to co-work on this project! ;-)

What will the final decision on this patch?

src/snmp.c
```diff
@@ -1327,22 +1327,24 @@ static int csnmp_read_table(host_definition_t *host, data_definition_t *data) {
...
     if (req == NULL) {
       ERROR("snmp plugin: snmp_pdu_create failed.");
       status = -1;
       break;
     }
 
-    oid_list_todo_num = 0;
+    size_t oid_list_todo_num = 0;
+    size_t var_idx[oid_list_len];
+    memset(&var_idx, 0, sizeof(var_idx));
```
octo: Sorry for yet another nit: When initializing arrays like that, we don't explicitly take the address of the array, i.e.: `memset(var_ids, 0, sizeof(var_idx));` without the `&`. (Both work, but we never use `&array` elsewhere …)

I really wish variable-length arrays could be zero-initialized with `{0}` …

src/snmp.c
rpv-tomsk: src/tcpconns.c:376

```
memset(&count_total, '\0', sizeof(count_total));
```
The declaration is:
```
static uint32_t count_total[TCP_STATE_MAX + 1];
```
)

src/snmp.c
rpv-tomsk: https://github.com/collectd/collectd/blob/master/src/tcpconns.c#L376
```
memset(&count_total, '\0', sizeof(count_total));
```
The declaration is:
```
static uint32_t count_total[TCP_STATE_MAX + 1];
```
:-)

rpv-tomsk: Updated, amended.

PR #2437, opened 2017-09-21
Varnish plugin update

dmatetelki:
Hi guys, please give me some feedback on how to reach your merge criteria.

PR #2453, opened 2017-09-27
Varnish plus plugin upgrade

dmatetelki:
extending the varnish plugin with varnish-plus counters (mainly the MSE storage engine)

i don't think we neither need separate plugin for this, extra macros (HAVE_VARNISH_V4_PLUS) nor updating the autotools

### DIFF
--- a/src/varnish.c
+++ b/src/varnish.c
@@ -88,6 +88,7 @@ struct user_config_s {
   _Bool collect_mgt;
   _Bool collect_smf;
   _Bool collect_vbe;
+  _Bool collect_mse;
 #endif
 };
 typedef struct user_config_s user_config_t; /* }}} */
@@ -227,6 +228,9 @@ static int varnish_monitor(void *priv,
     else if (strcmp(name, "esi_warnings") == 0)
       return varnish_submit_derive(conf->instance, "esi", "total_operations",
                                    "warning", val);
+    else if (strcmp(name, "esi_maxdepth") == 0)
+      return varnish_submit_derive(conf->instance, "esi", "total_operations",
+                                   "max_depth", val);
   }
 
   if (conf->collect_backend) {
@@ -588,12 +592,18 @@ static int varnish_monitor(void *priv,
     else if (strcmp(name, "s_req_bodybytes") == 0)
       return varnish_submit_derive(conf->instance, "totals", "total_bytes",
                                    "req_body", val);
+    else if (strcmp(name, "s_req_protobytes") == 0)
+      return varnish_submit_derive(conf->instance, "totals", "total_bytes",
+                                   "req_proto", val);
     else if (strcmp(name, "s_resp_hdrbytes") == 0)
       return varnish_submit_derive(conf->instance, "totals", "total_bytes",
                                    "resp_header", val);
     else if (strcmp(name, "s_resp_bodybytes") == 0)
       return varnish_submit_derive(conf->instance, "totals", "total_bytes",
                                    "resp_body", val);
+    else if (strcmp(name, "s_resp_protobytes") == 0)
+      return varnish_submit_derive(conf->instance, "totals", "total_bytes",
+                                   "resp_proto", val);
     else if (strcmp(name, "s_pipe_hdrbytes") == 0)
       return varnish_submit_derive(conf->instance, "totals", "total_bytes",
                                    "pipe_header", val);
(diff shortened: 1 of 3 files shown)


octo:
Hi @dmatetelki, does this supersede #2437?

dmatetelki:
Hi @octo, i would prefer to have them as 2 commits.

After the 1st is ok to go, I'll rebase&update 2nd PR.
The reason I've put this out here and not waiting for the merge of the 1st is in case someone would like to use it, while the 1st PR is still waiting to be chiseled.

dmatetelki:
@octo: the build system seems to have problems. How can i re-trigger the build?

src/varnish.c
```diff
@@ -227,6 +228,9 @@ static int varnish_monitor(void *priv,
     else if (strcmp(name, "esi_warnings") == 0)
       return varnish_submit_derive(conf->instance, "esi", "total_operations",
                                    "warning", val);
+    else if (strcmp(name, "esi_maxdepth") == 0)
+      return varnish_submit_derive(conf->instance, "esi", "derive",
```
octo: I'd recommend to use `"esi", "total_operations", "maxdepth"` to identify this metric.

src/varnish.c
```diff
@@ -867,6 +884,99 @@ static int varnish_monitor(void *priv,
...
+                                  "total_operations", "full_allocs", val);
+    else if (strcmp(name, "c_truncated") == 0)
+      return varnish_submit_derive(conf->instance, "mse",
+                                  "total_operations", "truncated_allocs", val);
+    else if (strcmp(name, "c_expanded") == 0)
+      return varnish_submit_derive(conf->instance, "mse",
+                                  "total_operations", "expanded_allocs", val);
+    else if (strcmp(name, "c_failed") == 0)
+      return varnish_submit_derive(conf->instance, "mse",
+                                  "", "failed_allocs", val);
```
octo: This call has an empty *type* argument, which is invalid.

src/varnish.c
```diff
@@ -867,6 +884,99 @@ static int varnish_monitor(void *priv,
...
+      return varnish_submit_gauge(conf->instance, "mse",
+                                  "total_operations", "bans_persisted", val);
+    else if (strcmp(name, "g_bans_lost") == 0)
+      return varnish_submit_gauge(conf->instance, "mse",
+                                  "total_operations", "bans_lost", val);
+
+     /* mse seg */
+    else if (strcmp(name, "g_journal_bytes") == 0)
+      return varnish_submit_gauge(conf->instance, "mse_reg",
+                                  "", "journal_bytes", val);
```
octo: Invalid empty *type*. I'd recomment to use `"bytes".

src/varnish.c
```diff
@@ -867,6 +884,99 @@ static int varnish_monitor(void *priv,
...
+    else if (strcmp(name, "g_journal_space") == 0)
+      return varnish_submit_gauge(conf->instance, "mse_reg",
```
octo: Are this and the previous metric bytes used for journaling and total journal space? If so, the recommended way of reporting these metrics is as "used" and "free", not "used" and "total", so that they can be stacked.

src/varnish.c
```diff
@@ -867,6 +884,99 @@ static int varnish_monitor(void *priv,
...
+                                  "", "journal_space", val);
```
octo: Invalid empty *type*.

src/varnish.c
```diff
@@ -867,6 +884,99 @@ static int varnish_monitor(void *priv,
...
+                                  "objects", "object_nodes_in_use", val);
+    else if (strcmp(name, "g_extnode") == 0)
+      return varnish_submit_gauge(conf->instance, "mse",
+                                  "objects", "extent_nodes_in_use", val);
+    else if (strcmp(name, "g_bigextfree") == 0)
+      return varnish_submit_gauge(conf->instance, "mse",
+                                  "objects", "free_big_extents", val);
+    else if (strcmp(name, "c_pruneloop") == 0)
+      return varnish_submit_gauge(conf->instance, "mse",
+                                  "total_operations", "prune_loops", val);
```
octo: You're calling `submit_gauge()` here, but the `total_operations` type is a derive. Please fix. (Here and below.)

octo:
I manually triggered a build which succeeded: https://ci.collectd.org/job/pull-requests-manual_trigger-aggregation/249/

PR #2456, opened 2017-09-28
write_redis_plugin: Add document for MaxSetDuration

s1061123:
This change adds description for MaxSetDuration, which is introduced at #2440.


s1061123:
This is not your fault, just my fault what I missed to add it in previous PR...
Anyway, thank you for your immediate merge, @octo !

PR #2459, opened 2017-09-29
there no such command in make make-dist-bz2

XakRu:


rubenk:
Since bzip2 is the default, can we shorten this to just `make dist`?

octo:
@rubenk Sure, go for it. Personally I don't care enough to write a PR ;-)

rubenk:
Alright, fixed in bfa038b7f879811f276377cf1737770f7d2d6536 :-)

Issue #2462, opened 2017-09-29
varnish plugin: Metrics with invalid data source type.


octo:
*   Version of collectd: *master*
*   Operating system / distribution: n/a

## Expected behavior

The `varnish_submit_gauge()` helper function should only be called with *types* (the third argument of the function) that are defined as `GAUGE` in `src/types.db`. Inversely, `varnish_submit_derive()` should only be called with `DERIVE` metric types.

## Actual behavior

The following Varnish metrics are passed to `varnish_submit_gauge()` but reference `DERIVE` metric types:

*   `bans_persisted_bytes` (references `total_bytes`)
*   `bans_persisted_fragmentation` (references `total_bytes`)
*   `g_alloc` (references `total_operations`)
*   `req` (references `http_requests`)

The following Varnish metrics are passed to `varnish_submit_derive()` but reference `GAUGE` metric types:

*   `creat` (references `objects`)
*   `destroy` (references `objects`)
*   `exp_mailed` (references `objects`)
*   `exp_received` (references `objects`)
*   `thread_queue_len` (references `queue_length`)

octo:
@dmatetelki You seem to know Varnish metrics quite well; can you help with this?

dmatetelki:
Sure, let me check them.

octo:
Fixed by #2463, thanks @dmatetelki!

PR #2463, opened 2017-10-02
varnish plugin: Metrics with invalid data source type. #2462

dmatetelki:


src/varnish.c
```diff
@@ -852,19 +852,19 @@ static int varnish_monitor(void *priv,
       return varnish_submit_derive(conf->instance, "mempool",
                                    "total_operations", "frees", val);
     else if (strcmp(name, "recycle") == 0)
-      return varnish_submit_derive(conf->instance, "mempool",
```
dmatetelki: I've also changed to "submit as gauge type" at the `mempool` section where the counters have type `objects`

src/varnish.c
```diff
@@ -852,19 +852,19 @@ static int varnish_monitor(void *priv,
...
+      return varnish_submit_gauge(conf->instance, "mempool",
```
dmatetelki: I've also changed to "submit as gauge type" at the `mempool` section where the counters have type `objects`

5 changes

dmatetelki:
Thank you @octo for your careful reviews!

Issue #2343, opened 2017-07-04
ceph plugin: JSON handler failed with status -1


f1-outsourcing:
```
collectd.x86_64                  5.7.1-2.el7                 @CentOS7_64-epel
collectd-ceph.x86_64             5.7.1-2.el7                 @CentOS7_64-epel
ceph.x86_64                      1:12.1.0-0.el7              @ceph
ceph-base.x86_64                 1:12.1.0-0.el7              @ceph
ceph-common.x86_64               1:12.1.0-0.el7              @ceph
ceph-mds.x86_64                  1:12.1.0-0.el7              @ceph
ceph-mgr.x86_64                  1:12.1.0-0.el7              @ceph
ceph-mon.x86_64                  1:12.1.0-0.el7              @ceph
ceph-osd.x86_64                  1:12.1.0-0.el7              @ceph
```

```
Jul  4 20:50:37 c01 collectd[26753]: ceph plugin: ds FinisherMonFinisher.completeLatency.avgtime was not properly initialized.
Jul  4 20:50:37 c01 collectd[26753]: last message repeated 125 times
Jul  4 20:50:37 c01 collectd[26753]: ceph plugin: ds FinisherMonFinisher.completeLatency.avgtime was not properly initialized.
Jul  4 20:50:37 c01 collectd[26753]: Available write targets: [none]
Jul  4 20:50:37 c01 collectd[26753]: ceph plugin: JSON handler failed with status -1.
Jul  4 20:50:37 c01 collectd[26753]: Available write targets: [none]
Jul  4 20:50:37 c01 collectd[26753]: Available write targets: [none]
Jul  4 20:50:37 c01 collectd[26753]: ceph plugin: JSON handler failed with status -1.
Jul  4 20:50:37 c01 collectd[26753]: Available write targets: [none]
Jul  4 20:50:37 c01 collectd[26753]: Available write targets: [none]
Jul  4 20:50:37 c01 collectd[26753]: ceph plugin: cconn_handle_event(name=mon.a,i=0,st=4): error 1
Jul  4 20:50:37 c01 collectd[26753]: Available write targets: [none]
Jul  4 20:50:37 c01 collectd[26753]: Available write targets: [none]
Jul  4 20:50:37 c01 collectd[26753]: ceph plugin: cconn_handle_event(name=mon.a,i=0,st=4): error 1
Jul  4 20:50:37 c01 collectd[26753]: Available write targets: [none]
```

## /etc/collectd.d/ceph.conf:

[...]

PR #2464, opened 2017-10-04
Add support for ceph version luminous

AlexZzz:
This patch is not backward compatible with previous ceph versions.
It adds support for [this](https://github.com/ceph/ceph/commit/767502e26ac0cfff5f060fe2ff9eca88de31258d#diff-b6cb32f55d238418c321e793d17d0da3) commit.

### DIFF
--- a/src/ceph.c
+++ b/src/ceph.c
@@ -278,7 +278,8 @@ static int ceph_cb_number(void *ctx, const char *number_val,
 
   /* Special case for latency metrics. */
   if ((strcmp("avgcount", state->key) == 0) ||
-      (strcmp("sum", state->key) == 0)) {
+      (strcmp("sum", state->key) == 0) ||
+      (strcmp("avgtime",state->key) == 0)) {
     latency_type = 1;
 
     /* depth >= 2  =>  (stack[-1] != NULL && stack[-2] != NULL) */
@@ -913,27 +914,34 @@ static int node_handler_fetch_data(void *arg, const char *val,
       // return after saving avgcount - don't dispatch value
       // until latency calculation
       return 0;
-    } else {
-      double sum, result;
-      sscanf(val, "%lf", &sum);
-
+    } else if (vtmp->avgcount_exists == 0) {
       if (vtmp->avgcount == 0) {
         vtmp->avgcount = 1;
       }
-
-      /** User wants latency values as long run avg */
+      vtmp->avgcount_exists = 1;
+      // user wants latency values as long run avg
+      // skip this step
       if (long_run_latency_avg) {
-        result = (sum / vtmp->avgcount);
-      } else {
-        result = get_last_avg(vtmp->d, ds_name, vtmp->latency_index, sum,
-                              vtmp->avgcount);
-        if (result == -ENOMEM) {
-          return -ENOMEM;
-        }
+        return 0;
       }
-
+      double sum, result;
+      sscanf(val, "%lf", &sum);
+      result = get_last_avg(vtmp->d, ds_name, vtmp->latency_index, sum,
+                            vtmp->avgcount);
+      if (result == -ENOMEM) {
+        return -ENOMEM;
+      }
       uv.gauge = result;
+      vtmp->latency_index = (vtmp->latency_index + 1);
+    } else if (vtmp->avgcount_exists == 1 ) {
       vtmp->avgcount_exists = -1;
+      // skip this step if no need in long run latency
+      if (!long_run_latency_avg) {
+        return 0;
+      }
+      double result;
+      sscanf(val, "%lf", &result);
+      uv.gauge = result;
       vtmp->latency_index = (vtmp->latency_index + 1);
     }
     break;
--- a/src/ceph.c
+++ b/src/ceph.c
@@ -532,6 +532,8 @@ static int parse_keys(char *buffer, size_t buffer_size, const char *key_str) {
       cut_suffix(tmp, tmp_size, key_str, ".sum");
     } else if (has_suffix(key_str, ".avgtime")) {
       cut_suffix(tmp, tmp_size, key_str, ".avgtime");
+    } else {
+      sstrncpy(tmp, key_str, sizeof(tmp));
     }
   } else {
     sstrncpy(tmp, key_str, sizeof(tmp));
--- a/src/ceph.c
+++ b/src/ceph.c
@@ -518,6 +518,12 @@ static int parse_keys(char *buffer, size_t buffer_size, const char *key_str) {
   size_t tmp_size = sizeof(tmp);
   const char *cut_suffixes[] = {".type", ".avgcount", ".sum", ".avgtime"};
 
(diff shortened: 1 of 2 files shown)


src/ceph.c
```diff
@@ -913,27 +914,34 @@ static int node_handler_fetch_data(void *arg, const char *val,
...
+      if (result == -ENOMEM) {
+        return -ENOMEM;
       }
-
       uv.gauge = result;
+      vtmp->latency_index = (vtmp->latency_index + 1);
+    } else if (vtmp->avgcount_exists == 1 ) {
       vtmp->avgcount_exists = -1;
+      // skip this step if no need in long run latency
+      if (!long_run_latency_avg) {
```
octo: I don't understand the logic of `long_run_latency_avg` – whether it is set to true or false, the user will always get a latency average … Why?

AlexZzz:
Hi @octo! Thank you for the answer and i'm sorry for poor description of the request.
This patch fixes issue [#2343](https://github.com/collectd/collectd/pull/2464). We've faced with the same error after upgrade ceph to luminous and made this patch to fix the problem.

src/ceph.c
```diff
@@ -913,27 +914,34 @@ static int node_handler_fetch_data(void *arg, const char *val,
       // return after saving avgcount - don't dispatch value
       // until latency calculation
       return 0;
-    } else {
-      double sum, result;
-      sscanf(val, "%lf", &sum);
-
+    } else if (vtmp->avgcount_exists == 0) {
```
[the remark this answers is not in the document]
AlexZzz: As I can see - ```ds_name``` has a long metric name, not just ```avgcount``` or ```sum```. So, I replaced ```avgcount_exists``` with readable enum.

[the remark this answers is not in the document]
octo: That's still a state machine and as a result extremely brittle. It's sensitive to new metrics being added and depends on the order of metrics.

Can't you use the existing `has_suffix()` function to determine which metric it is?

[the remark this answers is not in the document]
AlexZzz: I tried to find easy way to do it and did some tests.
Unfortunately, ```key``` and ```ds_name``` doesn't contain suffix we need. ```ceph_cb_number``` cut suffix for latency metrics. Do you think it should be fixed within this pull request?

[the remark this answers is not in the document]
octo: Yes, I think this should be fixed in this PR: the reason we have to revisit this code for a new Ceph release it because it is so brittle. A good fix for this problem should ensure that the code works even if Ceph should decide to

*   add a median, 95th percentile or distribution of the latency, or
*   re-order the fields, e.g. "sum" first, "avgcount" second.

AlexZzz:
Hi @octo !

I've changed code to make ```node_handler_fetch_data()``` stateless. Also, added a warning if unknown latency metric caught and add ```cut_suffix()``` for cutting suffix.
This code was tested on Jewel and Luminous ceph. So, it is backward compatible with previous ceph versions - not just "works on luminous".
Let me know if it needs more changes or I missed something!

AlexZzz:
Hi, @octo ,
Thanks for help with this patch.
I did some changes and it looks better now. Tests are passed.

src/ceph.c
```diff
@@ -522,18 +517,24 @@ static size_t count_parts(char const *key) {
...
+    if (has_suffix(key_str, ".type")) {
+      cut_suffix(tmp, tmp_size, key_str, ".type");
+    } else if (has_suffix(key_str, ".avgcount")) {
+      cut_suffix(tmp, tmp_size, key_str, ".avgcount");
+    } else if (has_suffix(key_str, ".sum")) {
+      cut_suffix(tmp, tmp_size, key_str, ".sum");
+    } else if (has_suffix(key_str, ".avgtime")) {
+      cut_suffix(tmp, tmp_size, key_str, ".avgtime");
+    } else {
+      sstrncpy(tmp, key_str, sizeof(tmp));
```
octo: You're using `tmp_size` above and `sizeof(tmp)` here. Please unify this. Since `tmp_size` is not modified, it's more common to just use `sizeof(tmp)` over and over again.

src/ceph.c
```diff
@@ -522,18 +517,24 @@ static size_t count_parts(char const *key) {
...
-
-  if ((count_parts(key_str) > 2) && has_suffix(key_str, ".type")) {
-    /* strip ".type" suffix iff the key has more than two parts. */
-    size_t sz = strlen(key_str) - strlen(".type") + 1;
-
-    if (sz > sizeof(tmp))
-      sz = sizeof(tmp);
-    sstrncpy(tmp, key_str, sz);
+  /* Strip suffix if it is ".type" or one of latency metric suffix. */
+  if (count_parts(key_str) > 2) {
```
octo: At your option:

```c
/* initialize tmp */
sstrncpy(tmp, key_str, sizeof(tmp));

char const *cut_suffixes[] = {".type", ".avgcount", ".sum", ".avgtime"};
for (size_t i = 0; i < STATIC_ARRAY_SIZE(cut_suffixes); i++) {
  if (has_suffix(key_str, cut_suffixes[i])) {
    cut_suffix(tmp, sizeof(tmp), key_str, cut_suffixes[i]);
    break;
  }
}
```

This has less repetition and makes it easier to see just which suffixes are being cut.

src/ceph.c
```diff
@@ -507,6 +485,23 @@ static _Bool has_suffix(char const *str, char const *suffix) {
   return 0;
 }
 
+static int cut_suffix(char new_str[], size_t new_str_len, char const *str,
```
octo: 1.  The most common name for the first two arguments are `buffer` and `buffer_size`.
2.  Please use a `char *` instead of a `char []`. (Both are correct and mean essentially the same thing, but the latter is very uncommon, at least within the C codebase).
3.  You only ever return the value zero and you never check the return value. Make this a `void` function.

```c
static void cut_suffix(char *buffer, size_t buffer_size, ...)
```

src/ceph.c
```diff
@@ -907,34 +908,41 @@ static int node_handler_fetch_data(void *arg, const char *val,
...
+      result = get_last_avg(vtmp->d, ds_name, vtmp->latency_index, sum,
+                            vtmp->avgcount);
+      if (result == -ENOMEM) {
+        return -ENOMEM;
       }
-
       uv.gauge = result;
-      vtmp->avgcount_exists = -1;
       vtmp->latency_index = (vtmp->latency_index + 1);
+    } else if (has_suffix(key, ".avgtime")) {
```
octo: Please add a comment here along the lines of:

> The "avgtime" metric reports ("sum" / "avgcount"), i.e. the average time per request since the start of the Ceph daemon. Report this only when the user has configured "long running average". Otherwise, use the *rate* of "sum" and "avgcount" to calculate the current latency.

AlexZzz:
Hi, @octo, thank you for review!
I did requested changes:)

octo:
Thanks @AlexZzz!

PR #2467, opened 2017-10-05
Library refactor

SeanCampbell:
A bit of refactoring so that reading the config file is done outside of main and global variables are separated into their own files.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -199,6 +199,8 @@ collectd_SOURCES = \
 	src/daemon/configfile.h \
 	src/daemon/filter_chain.c \
 	src/daemon/filter_chain.h \
+	src/daemon/globals.c \
+	src/daemon/globals.h \
 	src/daemon/meta_data.c \
 	src/daemon/meta_data.h \
 	src/daemon/plugin.c \
--- a/src/daemon/collectd.c
+++ b/src/daemon/collectd.c
@@ -445,23 +445,19 @@ static int notify_systemd(void) {
 }
 #endif /* KERNEL_LINUX */
 
-int main(int argc, char **argv) {
-  const char *configfile = CONFIGFILE;
-  int test_config = 0;
-  int test_readall = 0;
-  const char *basedir;
-  _Bool opt_create_basedir = 1;
-#if COLLECT_DAEMON
-  pid_t pid;
-  int daemonize = 1;
-#endif
-  int exit_status = 0;
+struct cmdline_config {
+  _Bool test_config;
+  _Bool test_readall;
+  _Bool create_basedir;
+  const char *configfile;
+  _Bool daemonize;
+};
 
+void read_cmdline(int argc, char **argv, struct cmdline_config *config) {
   /* read options */
   while (1) {
     int c;
-
-    c = getopt(argc, argv, "BhtTC:"
+    c = getopt(argc, argv, "htTC:"
 #if COLLECT_DAEMON
                            "fP:"
 #endif
@@ -472,19 +468,19 @@ int main(int argc, char **argv) {
 
     switch (c) {
     case 'B':
-      opt_create_basedir = 0;
+      config->create_basedir = 0;
       break;
     case 'C':
-      configfile = optarg;
+      config->configfile = optarg;
       break;
     case 't':
-      test_config = 1;
+      config->test_config = 1;
       break;
     case 'T':
-      test_readall = 1;
+      config->test_readall = 1;
       global_option_set("ReadThreads", "-1", 1);
 #if COLLECT_DAEMON
-      daemonize = 0;
+      config->daemonize = 0;
 #endif /* COLLECT_DAEMON */
       break;
 #if COLLECT_DAEMON
@@ -492,7 +488,7 @@ int main(int argc, char **argv) {
       global_option_set("PIDFile", optarg, 1);
       break;
     case 'f':
-      daemonize = 0;
+      config->daemonize = 0;
       break;
 #endif /* COLLECT_DAEMON */
     case 'h':
@@ -502,19 +498,17 @@ int main(int argc, char **argv) {
       exit_usage(1);
     } /* switch (c) */
   }   /* while (1) */
+}
 
-  if (optind < argc)
-    exit_usage(1);
-
-  plugin_init_ctx();
-
+int configure_collectd(struct cmdline_config *config) {
+  const char *basedir;
   /*
    * Read options from the config file, the environment and the command
    * line (in that order, with later options overwriting previous ones in
    * general).
    * Also, this will automatically load modules.
    */
-  if (cf_read(configfile)) {
+  if (cf_read(config->configfile)) {
     fprintf(stderr, "Error: Reading the config file failed!\n"
                     "Read the logs for details.\n");
     return 1;
@@ -528,22 +522,46 @@ int main(int argc, char **argv) {
     fprintf(stderr,
             "Don't have a basedir to use. This should not happen. Ever.");
     return 1;
-  } else if (change_basedir(basedir, opt_create_basedir)) {
+  } else if (change_basedir(basedir, config->create_basedir)) {
     fprintf(stderr, "Error: Unable to change to directory `%s'.\n", basedir);
     return 1;
   }
 
   /*
-   * Set global variables or, if that failes, exit. We cannot run with
+   * Set global variables or, if that fails, exit. We cannot run with
    * them being uninitialized. If nothing is configured, then defaults
    * are being used. So this means that the user has actually done
    * something wrong.
    */
   if (init_global_variables() != 0)
-    exit(EXIT_FAILURE);
+    return 1;
 
-  if (test_config)
+  return 0;
+}
+
+int main(int argc, char **argv) {
+#if COLLECT_DAEMON
+  pid_t pid;
+#endif
+  int exit_status = 0;
+
+  struct cmdline_config config = {
+      .daemonize = 1, .create_basedir = 1, .configfile = CONFIGFILE,
+  };
+
+  read_cmdline(argc, argv, &config);
+
+  if (config.test_config)
     return 0;
+
+  if (optind < argc)
+    exit_usage(1);
+
+  plugin_init_ctx();
+
+  int status;
+  if ((status = configure_collectd(&config)) != 0)
+    exit(EXIT_FAILURE);
 
 #if COLLECT_DAEMON
   /*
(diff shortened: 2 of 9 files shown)


Makefile.am
```diff
@@ -199,6 +199,8 @@ collectd_SOURCES = \
 	src/daemon/configfile.h \
 	src/daemon/filter_chain.c \
 	src/daemon/filter_chain.h \
+        src/daemon/globals.c \
```
octo: Nit: please fix whitespace. Looks like a tabs vs. spaces issue.

src/daemon/collectd.c
```diff
@@ -454,21 +444,17 @@ static int notify_systemd(void) {
...
-  int test_config = 0;
-  int test_readall = 0;
-  const char *basedir;
-#if COLLECT_DAEMON
-  pid_t pid;
-  int daemonize = 1;
-#endif
-  int exit_status = 0;
+struct cmdline_config {
+  int test_config;
```
octo: Boolean fields should be of type `_Bool`.

SeanCampbell:
Hi @octo,

Thanks for your patience. I tidied up the PR so it's just focused on the refactor.

Let me know if there's any other changes I should make :)

Thanks,
-Sean

octo:
Hi @SeanCampbell,

thanks for the update! Unfortunately you merged *master* into your PR, which makes code reviews confusing. I pushed a cleaned-up branch of your changes to [octo/collectd pr/2467](https://github.com/collectd/collectd/compare/master...octo:pr/2467), can we use that to discuss this change? I'm happy to push it into the PR branch, or you could do something along these lines:

```sh
# Make sure you're on the right branch
git checkout library-refactor
# Download my commits
git fetch git@github.com:octo/collectd.git pr/2467
# Reset the "library-refactor" branch to point to the latest commit
git reset --hard FETCH_HEAD
# Force pushing of the new head to the tracking remote
git push -f
```

Best regards,
—octo

SeanCampbell:
Ah, it said there were conflicts that prevented it from being automatically merged, so I thought that would make it easier. I ran the commands you said to pull in the updates from your branch, and happy to make any changes :)

Thank you,
Sean

src/daemon/globals.c
```diff
@@ -0,0 +1,13 @@
+#include "globals.h"
+#include "plugin.h"
+/*
+ * Global variables
+ */
+char hostname_g[DATA_MAX_NAME_LEN];
+const int hostname_g_size = sizeof (hostname_g);
+cdtime_t interval_g;
+int  pidfile_from_cli = 0;
```
octo: `pidfile_from_cli` is unused, likely a remnant of an unrelated change. Please remove.

src/daemon/globals.c
```diff
@@ -0,0 +1,13 @@
+#include "globals.h"
```
octo: This file needs a license header. In case of the core daemon, we insist on the [MIT license](https://opensource.org/licenses/MIT).

Dito for `src/daemon/globals.h`.

src/daemon/globals.c
```diff
@@ -0,0 +1,13 @@
...
+#include "plugin.h"
+/*
+ * Global variables
+ */
+char hostname_g[DATA_MAX_NAME_LEN];
+const int hostname_g_size = sizeof (hostname_g);
```
octo: I'm not quite happy about the introduction of this variable. How about we change `hostname_g` to be a `char *` instead? I think it would also be a good idea to add a simple setter function:

```c
/* hostname_set updates hostname_g */
void hostname_set(char const *hostname);
```

src/daemon/globals.c
```diff
@@ -0,0 +1,13 @@
+#include "globals.h"
+#include "plugin.h"
```
octo: I think you need `"plugin.h"` for `DATA_MAX_NAME_LEN`, which is one of those defines that might actually make sense to move to here (and out of `"plugin.h"`).

src/daemon/globals.c
```diff
@@ -0,0 +1,13 @@
...
+/*
+ * Global variables
+ */
+char hostname_g[DATA_MAX_NAME_LEN];
+const int hostname_g_size = sizeof (hostname_g);
+cdtime_t interval_g;
+int  pidfile_from_cli = 0;
+int  timeout_g;
+#if HAVE_LIBKSTAT
+kstat_ctl_t *kc;
```
octo: This type is not defined because the header is missing. I recommend to move

```c
#if HAVE_KSTAT_H
#include <kstat.h>
#endif
```

out of `src/daemon/collectd.h` and into here. Longer term, I'd like to get rid of the `kc` variable, but that's out of scope for here.

src/daemon/globals.c
```diff
@@ -1,13 +1,38 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ **/
+
+#include "common.h"
 #include "globals.h"
-#include "plugin.h"
+
+void hostname_set(char const *hostname) {
+  sstrncpy(hostname_g, hostname, sizeof(hostname_g));
```
octo: `sizeof()` only works with `char[]` (array) but not with `char*` (pointer).

I'd recommend doing something like this:

```c
void hostname_set(char const *hostname) {
  char *h = strdup(hostname);
  if (h == NULL)
    return;

  free(hostname_g);
  hostname_g = h;
}

File: src/daemon/collectd.c [hunk 9]
```diff
@@ -83,11 +83,11 @@ static int init_hostname(void) {
 
   str = global_option_get("Hostname");
   if ((str != NULL) && (str[0] != 0)) {
-    sstrncpy(hostname_g, str, hostname_g_size);
+    hostname_set(str);
     return 0;
   }
 
-  if (gethostname(hostname_g, hostname_g_size) != 0) {
+  if (gethostname(hostname_g, sizeof(hostname_g)) != 0) {
```
octo: With `hostname_g` being a `char*`, you need to use a temporary buffer here. Sizing is a bit tricky, but the correct way would be:

```c
long hostname_len = sysconf(_SC_HOST_NAME_MAX);
if (hostname_len == -1) {
  hostname_len = NI_MAXHOST;
}
char hostname[hostname_len];
```

Then use `hostname` instead of `hostname_g` for the remainder of the function.

P.S.: If the below `getaddrinfo()` call fails, you'll have to add a `hostname_set(hostname);` to keep the existing behavior of "configured-but-unresolvable is better than no hostname".

src/daemon/collectd.h
```diff
@@ -186,10 +186,6 @@
 #include <sys/param.h>
 #endif
 
-#if HAVE_KSTAT_H
```
octo: `"collectd.h"` used to be the header defining the global variables, so it should include `"globals.h"` now so that plugins continue working / compiling.

src/daemon/globals.c
```diff
@@ -1,13 +1,38 @@
+/**
+ * collectd - src/globals.c
+ * Copyright (C) 2017  Florian octo Forster
```
octo: I think the OSPO would want you to put "Google, Inc." here. Dito for `globals.h`.

SeanCampbell: Gotcha. I put "Google LLC" because of the recent name change. Is that right?

SeanCampbell:
Fixed the issue with hostname_g[] (for some reason, it was compiling fine on my machine even though it broke the tests). I'm happy to take any other comments or suggestions.

src/daemon/collectd.h
SeanCampbell: Yes, it's included at the bottom :)

SeanCampbell:
Friendly ping :)

src/daemon/collectd.c
```diff
@@ -91,13 +81,20 @@ static int init_hostname(void) {
   struct addrinfo *ai_list;
   int status;
 
+  long hostname_len = sysconf(_SC_HOST_NAME_MAX);
+  if (hostname_len == -1) {
+    hostname_len = NI_MAXHOST;
+  }
+  char hostname[hostname_len];
+  hostname_set(hostname);
```
octo: You're calling `hostname_set()` with an uninitialized buffer here. I think this needs to be removed – the code below is taking care of calling `hostname_set()` once the buffer is initialized.

src/daemon/globals.c
```diff
@@ -0,0 +1,49 @@
...
+char *hostname_g;
+cdtime_t interval_g;
+int  timeout_g;
+#if HAVE_KSTAT_H
+kstat_ctl_t *kc;
+#endif
+
+void hostname_set(char const *hostname) {
+    char *h = strdup(hostname);
+      if (h == NULL)
```
octo: The indentation here is wrong. Please run `clang-format -style=file -i $file` on each `.c` and `.h` file that was changed or added.

src/daemon/globals.c
```diff
@@ -0,0 +1,49 @@
...
+            return;
+
+        free(hostname_g);
```
octo: Nit: Please use the `sfree()` macro for consistency with other code.

src/perl.c
```diff
@@ -2390,6 +2384,11 @@ static void xs_init(pTHX) {
    * accessing any such variable (this is basically the same as using
    * tie() in Perl) */
   /* global strings */
+  struct {
```
octo: Why did this need to be moved?

SeanCampbell: It was breaking the integration tests with the error message [1]:

src/perl.c:268:1: error: initializer element is not constant
 } g_strings[] = {{"Collectd::hostname_g", hostname_g}, {"", NULL}};
 ^

Since it was a global variable, it could not be initialized with hostname_g as type char*. Changing it to a local variable fixed the issue, and I didn't see anywhere else it was used.

On that note though, I suppose the "g_" prefix would no longer be appropriate. What do you think?

[1] https://ci.collectd.org/job/pull-requests-build-on-jessie-amd64-with-default-toolchain/1766/console

SeanCampbell:
For some reason, after I ran clang-format, it's complaining about the formatting. Any ideas on why that is?

I tried running clang-format on all the files again, just to see if it makes a difference, but it didn't.

octo:
@SeanCampbell Which version of *clang-format* were you using?

octo:
LGTM, thanks @SeanCampbell!

I'd really be interested in which *clang-format* version you were using. The formatting check is brand-new and it's supposed to be helpful, not annoying.

Since we're talking about this: as a contributor, would you prefer if the bot would push the formatted files to the branch instead of reporting a test failure?

Best regards,
—octo

SeanCampbell:
Yeah, I'd like that better, especially since it would eliminate the possibility of differences in clang-format versions causing issues.

Thanks!
-Sean

octo:
FTR, you can use the same *clang-format* the bot is using via cURL:

```
curl --data-binary '@-' https://clang-format.appspot.com <input.c >output.c
```

I'll add a wrapper around this to `contrib/` in a bit.

Issue #2386, opened 2017-07-30
memcached: Wrong type is used for 'listen_disabled_num' field.


rpv-tomsk:
memcached.c:
```
if (FIELD_IS("listen_disabled_num")) {
   submit_derive("connections", "listen_disabled", atof(fields[2]), st);
}
```

types.db:
```
connections             value:DERIVE:0:U
```

Choosen data type is incorrect. That is not 'connections rate', but a 'counter how many times limit was reached'. We are unable to know how many connections was lost because of that.

Presense of that metric with type `connections` and type instance `listen_disabled` will break other metrics of `connections` type which could be added:

`rejected` - from field `rejected_connections` which contains  'Conns rejected in maxconns_fast mode'
`opened`  - from field `total_connections` which contains 'Total number of connections opened since the server started running' 

At least, we need to change it's type, but my proposal is to remove that metric. 
If needed, we can add notification instead.

octo:
ACK, `connections` is not the right type. I suggest to use `operations` instead. This is in the gray area between a bug fix and a breaking change, so I'd push this to *master* instead of a version branch.

PR #2468, opened 2017-10-06
memcached plugin: Report "listen_disabled_num" as "operations".

octo:
Fixes: #2386

### DIFF
--- a/src/memcached.c
+++ b/src/memcached.c
@@ -523,7 +523,7 @@ static int memcached_read(user_data_t *user_data) {
     else if (FIELD_IS("curr_connections")) {
       submit_gauge("memcached_connections", "current", atof(fields[2]), st);
     } else if (FIELD_IS("listen_disabled_num")) {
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
As per https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L681, the source field of this metric represents a 'Number of times server has stopped accepting new connections'.  
This is not a number of operations, but a number of events. 
Unfortunately, Collectd now has no such type as `events`, so I can't propose the type to use.

octo:
"listen [was] disabled" sounds like an operation that was performed, not an event that just happened, but I know nothing about memcached, so I might be way off here ;-)

rpv-tomsk:
What would we do this this?
Proposals: `events_rate` (new common type to be added).

rpv-tomsk:
[Auto]Merged. ;-)

Thanks for a work!

PR #2470, opened 2017-10-07
Add client interface for the jolokia.org JMX client.

dothebart:
ChangeLog: curl_jolokia plugin: integrate jolokia to fetch jmx counters via HTTP/json

Http://jolokia.org is another way to query jmx counters. It breaks the dependency chain by specifying a REST/HTTP API to access the values of http counters.

The module itself is intended to be configured by using the attached python script.
The configuration of the plugin consists of 3 parts:
- URL setup of the servlet container
- the Json Post data, which is a 'BULK-Request' in the nomenclature of jolokia
- a table of matches from the JMX names and their Attributes to the name of the metric on collectd side

The Python script uses the pyjolokia library to access your jolokia host. It queries the list interface of jolokia, which gives you all available beans. it then filters this list by patterns, so you don't query unavailable beans. 

The plugin was tested with weblogic.

Another companion product of jolokia to be mentioned is http://hawt.io

This plugin was derived from the curl_json plugin. This adapts PR #667 to the current collectd state.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -2269,18 +2269,18 @@ Jolokia JSON reply retrieved via B<libcurl> (L<http://curl.haxx.se/>)
      Password "passvoid"
      Post <JOLOKIA json post data>
 
-     <BeanName "PS_Scavenge">
-          MBean "java.lang:name=PS Scavenge,type=GarbageCollector"
-          BeanNameSpace "java_lang"
-          <AttributeName "collectiontime" >
-                 Attribute "CollectionTime"
-                 type "gauge"
-          </AttributeName>
-          <AttributeName "collectioncount" >
-                 Attribute "CollectionCount"
-                 type "gauge"
-          </AttributeName>
-     </BeanName>
+   <BeanName "PS_Scavenge">
+        MBean "java.lang:name=PS Scavenge,type=GarbageCollector"
+        BeanNameSpace "java_lang"
+        <AttributeName "collectiontime" >
+               Attribute "CollectionTime"
+               type "gauge"
+        </AttributeName>
+        <AttributeName "collectioncount" >
+               Attribute "CollectionCount"
+               type "gauge"
+        </AttributeName>
+   </BeanName>
   </Plugin>
 
 The plugin is intended to be written in a simple manner. Thus it doesn't 
@@ -2348,13 +2348,26 @@ I<cURL> plugin. Please see there for a detailed description.
 =item B<E<lt>BeanNameE<gt>>
 
 One B<BeanName> block configures the translation of the gauges of one bean
-to their respective collectd names.
+to their respective collectd names, where BeanName sets the main name.
 
-=back
+=item B<MBean> I<MBean>
 
-The following options are valid within B<Key> blocks:
+The name of the Bean on the server
+
+=item B<BeanNameSpace> I<BeanNameSpace>
+
+The name space the Bean resides under
 
 =over 4
+
+=item B<AttributeName> I<AttributeName>
+
+A bean can contain several Attributes with gauges. Each one can be matched by a 
+AttributeName section or be ignored. 
+
+=item B<Attribute> I<Attribute>
+
+How should this attribute be called under the BeanName in the collectd hierarchy?
 
 =item B<Type> I<Type>
 
@@ -2362,10 +2375,7 @@ Sets the type used to dispatch the values to the daemon. Detailed information
 about types and their configuration can be found in L<types.db(5)>. This
 option is mandatory.
 
-=item B<Instance> I<Instance>
-
-Type-instance to use. Defaults to the current map key or current string array element value.
-
+=back
 =back
 
 =head2 Plugin C<curl_xml>
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -2249,6 +2249,124 @@ option is mandatory.
 Type-instance to use. Defaults to the current map key or current string array element value.
 
 =back
+=head2 Plugin C<curl_jolokia>
+
+The B<curl_jolokia plugin> collects values from MBeanServevr - servlet engines equipped 
+with the jolokia (L<https://jolokia.org>) MBean. It sends a pre-configured
+JSON-Postbody to the servlet via HTTP commanding the jolokia Bean to reply with
+a singe JSON equipped with all JMX counters requested.
+By reducing TCP roundtrips in comparison to conventional JMX clients that
+query one value via tcp at a time, it can return hundrets of values in one roundtrip.
+Moreof - no java binding is required in collectd to do so.
+
+It uses B<libyajl> (L<https://lloyd.github.io/yajl/>) to parse the 
+Jolokia JSON reply retrieved via B<libcurl> (L<http://curl.haxx.se/>)
+
+ <Plugin curl_jolokia>
+   <URL "http://10.10.10.10:7101/jolokia-war-1.2.0/?ignoreErrors=true&canonicalNaming=false";>
+     Host "_APPPERF_JMX"
+     User "webloginname"
+     Password "passvoid"
+     Post <JOLOKIA json post data>
+
+     <BeanName "PS_Scavenge">
+          MBean "java.lang:name=PS Scavenge,type=GarbageCollector"
+          BeanNameSpace "java_lang"
+          <AttributeName "collectiontime" >
+                 Attribute "CollectionTime"
+                 type "gauge"
+          </AttributeName>
+          <AttributeName "collectioncount" >
+                 Attribute "CollectionCount"
+                 type "gauge"
+          </AttributeName>
+     </BeanName>
+  </Plugin>
+
+The plugin is intended to be written in a simple manner. Thus it doesn't 
+try to solve the task of generating the jolokia post data, or automatically 
+map the values, but rather leans on a verbose config containing the prepared
+flat JSON post data and a config section per gauge transformed (as one sample shown
+above). However, Jolokia can output all available gauges, and we have a python 
+script to filter them, and generate a configuration for you: 
+
+  jolokia_2_collectcfg.py
+
+it can gather all interesting gauges, write a simple one value per line config 
+for itself and subsequent calls.
+You can remove lines from this file manually, or create filter lists. 
+You then use the script to generate a collectd config. 
+The script can then inspect data files from some testruns, and remove 
+all gauges, that don't contain any movement. 
+
+The base config looks like this:
+
+The following options are valid within B<URL> blocks:
+
+=over 4
+
+=item B<Host> I<Name>
+
+Use I<Name> as the host name when submitting values. Defaults to the global
+host name setting.
+
+=item B<Plugin> I<Plugin>
+
+Use I<Plugin> as the plugin name when submitting values.
+Defaults to C<curl_jolokia>.
+
+=item B<Instance> I<Instance>
+
+Sets the plugin instance to I<Instance>.
+
+=item B<Interval> I<Interval>
+
+Sets the interval (in seconds) in which the values will be collected from this
+URL. By default the global B<Interval> setting will be used.
+
+=item B<User> I<Name>
+
+=item B<Password> I<Password>
+
+=item B<Digest> B<true>|B<false>
+
+=item B<VerifyPeer> B<true>|B<false>
+
+=item B<VerifyHost> B<true>|B<false>
+
+=item B<CACert> I<file>
+
+=item B<Header> I<Header>
+
+=item B<Post> I<Body>
+
+=item B<Timeout> I<Milliseconds>
+
+These options behave exactly equivalent to the appropriate options of the
+I<cURL> plugin. Please see there for a detailed description.
+
+=item B<E<lt>BeanNameE<gt>>
+
+One B<BeanName> block configures the translation of the gauges of one bean
+to their respective collectd names.
+
+=back
+
+The following options are valid within B<Key> blocks:
+
+=over 4
+
+=item B<Type> I<Type>
+
+Sets the type used to dispatch the values to the daemon. Detailed information
+about types and their configuration can be found in L<types.db(5)>. This
+option is mandatory.
+
+=item B<Instance> I<Instance>
+
+Type-instance to use. Defaults to the current map key or current string array element value.
+
+=back
 
 =head2 Plugin C<curl_xml>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -2249,6 +2249,7 @@ option is mandatory.
 Type-instance to use. Defaults to the current map key or current string array element value.
 
 =back
+
 =head2 Plugin C<curl_jolokia>
 
 The B<curl_jolokia plugin> collects values from MBeanServevr - servlet engines equipped 
@@ -2376,6 +2377,7 @@ about types and their configuration can be found in L<types.db(5)>. This
 option is mandatory.
 
 =back
+
 =back
 
 =head2 Plugin C<curl_xml>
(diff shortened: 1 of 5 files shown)


dothebart:
hm, I don't feel guilty for the failures jenkins reports here.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ *
+ * Authors:
+ *   Doug MacEachern <dougm at hyperic.com>
+ *   Florian octo Forster <octo at collectd.org>
+ *   Wilfried dothebart Goesgens <dothebart at citadel.org>
+ **/
+
+#include "common.h"
```
octo: 1.  `"collectd.h"` needs to be the first header included. You will need to add an empty line after it to keep `clang-format` from sorting includes alphabetically.
2.  Please run `clang-format -style=file -i src/curl_jolokia.c`

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+#include <sys/un.h>
+
+#include <curl/curl.h>
+
+#include <yajl/yajl_parse.h>
+#if HAVE_YAJL_YAJL_VERSION_H
+#include <yajl/yajl_version.h>
+#endif
+
+#if defined(YAJL_MAJOR) && (YAJL_MAJOR > 1)
```
octo: FYI, I'm happy to abandon support for libyajl1 – it's ancient. I just added appropriate configure conditionals to #2472 that we could pull into here if libyajl1 support would produce additional work.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+  char *bean_name;
+  char *mbean_match;
+  char *mbean_namespace;
+  size_t mbean_match_len;
+  llist_t *attributes; /* list of cjo_attribute_t */
+};
+typedef struct cjo_bean_s cjo_bean_t;
+/* }}} */
+
+typedef enum _expect_token { eNone = 0, eValue, eMBean } cjo_expect_token;
```
octo: Identifiers must not start with an underscore (according to C / POSIX). Shouldn't this work:?

```c
typedef enum { eNone = 0, eValue, eMBean } cjo_expect_token;
```

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+}
+
+/*
+ * Aligning data with Configs & Sending it
+ */
+
+static int cjo_get_type(cjo_attribute_t *attribute) {
+  const data_set_t *ds;
+
+  assert(attribute->type != NULL);
```
octo: Please don't do argument checking with `assert()`. Either

1.  add a comment along the lines of `/* ensured in cjo_somefunc */`, or
2.  change to something like `if (attr->type == NULL) { return EINVAL; };`

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+static void cjo_init_buffer(cjo_membuffer_t *buf, size_t initial_size) {
+  buf->buffer = malloc(initial_size);
+  buf->used = 0;
+  buf->size = initial_size;
+}
+
+static void cjo_append_buffer(cjo_membuffer_t *buf, const char *append,
+                              size_t appendlen) {
+  if (buf->used + appendlen + 1 > buf->size) {
+    char *newbuf;
```
octo: Please declare variables when initializing them, where possible. E.g.:

```c
size_t newsize = buf->size * 2;

char *newbuf = calloc(1, newsize);
```

Here and below.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+    size_t newsize;
+
+    newsize = buf->size * 2;
+
+    newbuf = malloc(newsize);
```
octo: Please use `calloc()` always.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+      memcpy(attribute, db->curr_attribute->json_name,
+             db->curr_attribute->json_name_len);
+      attribute[db->curr_attribute->json_name_len] = '\0';
+
+      ERROR("curl_jolokia plugin: failed to map type for [%s:%s:%s]",
+            db->match_this_bean->bean_name, attribute, curr_attribute->type);
+      continue;
+    }
+
+    memset(&ret_value, 0, sizeof(ret_value));
```
octo: Please use static initializers when possible. E.g. in this case:

```c
value_t ret_value = {0};
```

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+
+    memcpy(buffer, db->curr_attribute->json_value,
+           db->curr_attribute->json_value_len);
+
+    buffer[db->curr_attribute->json_value_len] = '\0';
+
+    status = parse_value(buffer, &ret_value, ds_type);
```
octo: Only assign the return value to a variable if you need it later, e.g. to log it. Otherwise, do:

```c
if (parse_value(/* … */) != 0 {
  /* … */
}
```

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+              db->match_this_bean->bean_name, attribute, buffer);
+      continue;
+    }
+
+    vl.values = &ret_value;
+    vl.values_len = 1;
+
+    if ((db->host == NULL) || (strcmp("", db->host) == 0) ||
+        (strcmp(CJO_DEFAULT_HOST, db->host) == 0))
+      host = hostname_g;
```
octo: Please leave `vl.host` empty in this case, e.g. `host = ""`.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1100 @@
...
+    curl_easy_setopt(db->curl, CURLOPT_TIMEOUT_MS,
+                     (long)CDTIME_T_TO_MS(plugin_get_interval()));
+#endif
+  return 0;
+} /* }}} int cjo_init_curl */
+
+static int cjo_sock_perform(cjo_t *db) /* {{{ */
+{
+  char errbuf[1024];
+  struct sockaddr_un sa_unix = {};
```
octo: Please use a struct initializer here:

```c
struct sockaddr_un sa_unix = {
  .sun_family = AF_UNIX,
};
```

dothebart:
Ok, here we go. I hope its done properly.

src/collectd.conf.pod
```diff
@@ -1879,6 +1879,124 @@ option is mandatory.
...
+I<cURL> plugin. Please see there for a detailed description.
+
+=item B<E<lt>BeanNameE<gt>>
+
+One B<BeanName> block configures the translation of the gauges of one bean
+to their respective collectd names.
+
+=back
+
+The following options are valid within B<Key> blocks:
```
octo: `Key` blocks don't show up in the above synopsis and options *inside* the `BeanName` blocks don't show up here. Please clarify.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1045 @@
...
+
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <sys/un.h>
+
+#include <curl/curl.h>
+
+#include <yajl/yajl_parse.h>
+#if HAVE_YAJL_YAJL_VERSION_H
+#include <yajl/yajl_version.h>
```
octo: It looks like you removed all the *libyajl 1* compatibility, as I suggested. With that, it looks like you're no longer using the `<yajl/yajl_version.h>` header.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1045 @@
...
+#endif
+
+#define CJO_DEFAULT_HOST "localhost"
+
+typedef size_t yajl_len_t;
```
octo: At your option: I think we only introduced the `yajl_len_t` type so it could be defined differently for *libyajl 1* and *2*. I think it's safe to `s/yajl_len_t/size_t/` and remove it.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1045 @@
...
+}
+
+static void cjo_release_buffercontent(cjo_membuffer_t *buf) {
+	sfree(buf->buffer);
+	buf->buffer = NULL;
+	buf->size = 0;
+	buf->used = 0;
+}
+
+static void cjo_remember_value(cjo_membuffer_t *ShouldBeInHere,
```
octo: Please use `snake_case` rather than `CamelCase`.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1045 @@
...
+			       db->curr_attribute->json_name_len);
+			attribute[db->curr_attribute->json_name_len] = '\0';
+
+			ERROR("curl_jolokia plugin: failed to map type for [%s:%s:%s]",
+			      db->match_this_bean->bean_name, attribute, curr_attribute->type);
+			continue;
+		}
+
+		value_t ret_value;
+		memset(&ret_value, 0, sizeof(ret_value));
```
octo:
```c
value_t ret_value = {0};
```

src/curl_jolokia.c
```diff
@@ -0,0 +1,1045 @@
...
+ */
+
+static int cjo_get_type(cjo_attribute_t *attribute) {
+	if (attribute->type != NULL) {
+		return -EINVAL;
+	}
+
+	const data_set_t *ds = plugin_get_ds(attribute->type);
+	if (ds == NULL) {
+		static char type[DATA_MAX_NAME_LEN] = "!!!invalid!!!";
```
octo: Where is this string literal coming from? I don't see it a second time; how can `attribute->type` be set to this?

Also, in theory `!!!invalid!!!` is a valid type name. If you want to ensure that the string is not a valid type name, include a slash or a hyphen (or both), for example `/--invalid--/`.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1045 @@
...
+			ERROR("curl_jolokia plugin: failed to locate attribute [%s:\"%s\"]",
+			      db->match_this_bean->bean_name, attribute);
+
+			continue; /* we don't know this! */
+		}
+		cjo_attribute_t *curr_attribute = cfgptr->value;
+
+		/* Create a null-terminated version of the string. */
+		int ds_type = cjo_get_type(curr_attribute);
+		if (ds_type == -1) {
```
octo: `cjo_get_type()` may contain `-EINVAL` too, so `(ds_type < 0)` would be a better check.

src/curl_jolokia.c
```diff
@@ -0,0 +1,1045 @@
...
+		buf->buffer = newbuf;
+		buf->size = newsize;
+	}
+	memcpy(buf->buffer + buf->used, append, appendlen);
+	buf->used += appendlen;
+}
+
+static void cjo_release_buffercontent(cjo_membuffer_t *buf) {
+	sfree(buf->buffer);
+	buf->buffer = NULL;
```
octo: The `sfree()` macro will set `buf->buffer` to `NULL` already.

src/curl_jolokia.c
octo: This function frightens and confuses me. Here is what I *think* is going on:

The HTTP response is read into a `cjo_membuffer_t` and passed to this function as `ShouldBeInHere`. The JSON string pointed to by `ptr` is likely a pointer into this buffer. If not, then the value is copied into a scratch space, the `BufferHereIfNot`. The pointer `SetThisOne` is then updated to either point into the response buffer `ShouldBeInHere` or the scratch space `BufferHereIfNot`.

I think this is very complex and the benefit (reduced memory footprint?) doesn't justify the complexity. I'd much rather always allocate memory and store it in the pointer, for example:

```c
char *cjo_json_strdup(const unsigned char *s, yajl_len_t len) {
  char *out = calloc(1, len + 1);
  if (out == NULL)
    return NULL;
  memcpy(out, s, len);
  /* out[len] == 0 */
  return out;
}
/* … */
/* in cjo_cb_string */
sfree(attr->json_name);
attr->json_name = cjo_json_strdup(val, len);
```

dothebart: hm, the jolokia replies can become huge (up to megs) and this mechanism only catches the case that you have a gauge where the value was spread across two tcp chunks. Since the value is copied afterwards anyways, I'd like to leave it that way.

octo: Okay, but … we only need to copy attribute names and values, not the entire response. And that buffer is flushed when a map closes, presumably that's when an MBean has been read completely? Last but not least, the number of attributes being cached is capped at `MaxReadAttributes`.

How many attributes would you say an MBean has at most? Put another way, what you suggest users set `MaxReadAttributes` to?

src/curl_jolokia.c
```diff
@@ -0,0 +1,1038 @@
...
+	memset(db->attributepool, 0,
+	       sizeof(*db->attributepool) * db->max_attribute_count);
+
+} /* }}} int cjo_submit */
+
+/* yajl callbacks */
+#define CJO_CB_ABORT 0
+#define CJO_CB_CONTINUE 1
+
+static int cjo_cb_string(void *ctx, const unsigned char *val, size_t len) {
```
octo: Could you please add a comment that demonstrates what the expected JSON looks like? That would help me understand this logic tremendously.

dothebart: Their documentation explains it: https://jolokia.org/reference/html/protocol.html as previously noted, this thing is basically a dumbed down curl_json, which takes a pre-configured json post document, and based on the rest of the configuration tries to match the tree structure of the jolokia reply.

dothebart: The streaming nature of yail makes it a bit more complicated. If there was a way to address stuff like you do in https://stedolan.github.io/jq/ it would be easy ;-)

dago:
@dothebart I know it has been quite some time since your contribution, but would you be willing to rebase this PR and resolve the conflict? In addition, libyajl has been deprecated in favor of libjansson for some time now, maybe a switch is advised?

dothebart:
Hi, 
if you are willing to merge it, I would do that. I don't use jolokia for myself anymore - though I find it a pretty good match for collectd.

dago:
@dothebart The PR is essentially well audited by octo and I'd be happy to merge it if it is brought to a current state.

dothebart:
ok, merged.

dago:
@dothebart CI failure on trusty is expected after a previous PR (Varnish) that still needs fixing, but redhat_default_toolchain container:collectd/ci:el6_x86_64 is strange - would you mind taking a look?

dothebart:
lets see whether thats it - cut'n'pasted from curl_json.c

dago:
@dothebart I noticed that you merged master into this branch instead of rebasing which makes following changes more difficult. Would you mind rebasing this PR to master and force-push?

dothebart:
like that?

dago:
Yes, much better 😄

dothebart:
ok, so now all that is expected to be green is green, right?

dago:
It looks like there is still a change request by octo although I resolved all items from the review conversation. @octo can you resolve this issue? I can only dismiss your review which I would rather not do.

dothebart:
I think at the last time I reviewed this PR I didn't have a jolokia enabled server at hand, and none on their irc channel wanted to give me some sample replies to create some unittest-dummies. 

but, i.e. 
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/jolokia/jolokia_test.go
has some, maybe one could steal it there.

dothebart:
Yay! so 6 Years after I quit the job that I needed this for it can finally be merged?

octo:
@dothebart Yeah, sorry for the long turn-around! :( And thanks to @dago for resurfacing this :)

kwiatrox:
@octo, It looks like this PR is merged only against master, probably it should be merged also into main and 5.12 branches.

PR #2471, opened 2017-10-07
Currently curl_json will barely ignore boolean values in a non erroneous

dothebart:
Currently curl_json will barely ignore boolean values in a non erroneous
way, so that if you have an array like this: `{ foo: [false, 123] }`
you will be able to access the 123 number like `foo[1]`

We now rather call our number parser with "0" for false, "1" for true.

This PR superseedes #1332

### DIFF
--- a/src/curl_json.c
+++ b/src/curl_json.c
@@ -226,11 +226,6 @@ static void cj_advance_array(cj_t *db) {
 #define CJ_CB_ABORT 0
 #define CJ_CB_CONTINUE 1
 
-static int cj_cb_boolean(void *ctx, int boolVal) {
-  cj_advance_array(ctx);
-  return CJ_CB_CONTINUE;
-}
-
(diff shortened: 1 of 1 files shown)


octo:
LGTM, but I see some indentation that's off – please re-run *clang-format*

Thanks, —octo

src/curl_json.c
```diff
@@ -292,6 +287,14 @@ static int cj_cb_string(void *ctx, const unsigned char *val, yajl_len_t len) {
   return cj_cb_number(ctx, (const char *)val, len);
 } /* int cj_cb_string */
 
+static int cj_cb_boolean(void *ctx, int boolVal) {
+  if (boolVal) {
+   return (cj_cb_number (ctx, "1", 1));
```
octo: Please remove the braces around the return value, i.e.:

```c
return cj_cb_number (…);
```

We recently unified the code base on this style.

PR #2472, opened 2017-10-09
New plugin: Write Stackdriver writes to Google Stackdriver Monitoring

octo:
This new plugin writes metrics to [Stackdriver Monitoring](https://cloud.google.com/monitoring/) (previously Google Cloud Monitoring).

It is similar to but not identical to Stackdriver's own *write_gcm plugin*: this implementation uses the standard `timeseries` API endpoint and *custom metrics* to write metrics. The "monitored resource" is configurable and defaults to `global`.

There is one somewhat open issue: Stackdriver expects a "start time" of "cumulative metrics" (`DERIVE`s). The first timeseries point this code is submitting has start time and end time equal, which is rejected by the API. The start time is stored in the metric's meta data and all subsequent metrics will send the same start time. However, there is a spike at the beginning of cumulative timeseries, which is suboptimal.

I realize that this PR is too big. If you want, we can treat this as the "umbrella" PR and I can split the "OAuth", "compute engine", "GCM format" and actual *write_gcm plugin* parts into separate PRs.

CC: @igorpeshansky


Makefile.am
```diff
@@ -518,6 +518,68 @@ liboconfig_la_SOURCES = \
...
+libgce_la_CPPFLAGS = \
+	$(AM_CPPFLAGS) \
+	$(BUILD_WITH_LIBCURL_CFLAGS)
+libgce_la_LIBADD = \
+	$(BUILD_WITH_LIBCURL_LIBS)
+endif
+endif
+endif
+
+if BUILD_WITH_LIBYAJL2
```
rubenk: This probably needs ```if BUILD_WITH_LIBSSL``` as well?

configure.ac
```diff
@@ -5069,6 +5071,59 @@ PKG_CHECK_MODULES([LIBSIGROK], [libsigrok < 0.4],
 )
 # }}}
 
+# --with-libssl {{{
+with_libssl_cflags=""
+with_libssl_ldflags=""
+AC_ARG_WITH(libssl, [AS_HELP_STRING([--with-libssl@<:@=PREFIX@:>@], [Path to libssl.])],
```
rubenk: From here on, can you please keep the style of the existing configure.ac? So quote all macro arguments, if; then on the same line etc.

configure.ac
```diff
@@ -6650,6 +6711,7 @@ AC_PLUGIN([vserver],             [$plugin_vserver],         [Linux VServer stati
 AC_PLUGIN([wireless],            [$plugin_wireless],        [Wireless statistics])
 AC_PLUGIN([write_graphite],      [yes],                     [Graphite / Carbon output plugin])
 AC_PLUGIN([write_http],          [$with_libcurl],           [HTTP output plugin])
+AC_PLUGIN([write_stackdriver],   [$plugin_write_stackdriver], [Google Stackdriver Monitoring output plugin])
```
rubenk: please reindent the whole AC_PLUGIN list.

src/daemon/utils_cache_mock.c
```diff
@@ -24,9 +24,10 @@
  *   Florian octo Forster <octo at collectd.org>
  */
 
-#include <errno.h>
 #include "utils_cache.h"
 
+#include <errno.h>
```
rubenk: unintended change?

src/utils_gce.c
```diff
@@ -0,0 +1,288 @@
...
+  if (!curl) {
+    ERROR("utils_gce: curl_easy_init failed.");
+    return NULL;
+  }
+
+  struct curl_slist *headers = NULL;
+  headers = curl_slist_append(headers, GCE_METADATA_HEADER);
+
+  char curl_errbuf[CURL_ERROR_SIZE];
+  blob_t blob = {NULL, 0};
```
rubenk: `blob_t blob = {NULL}` should be enough to zero-initialize it, or am I mistaken?

src/utils_gce.c
```diff
@@ -0,0 +1,288 @@
...
+  DEBUG("utils_gce: Checking whether I'm running on GCE ...");
+
+  CURL *curl = curl_easy_init();
+  if (!curl) {
+    ERROR("utils_gce: curl_easy_init failed.");
+    return 0;
+  }
+
+  struct curl_slist *headers = NULL;
+  headers = curl_slist_append(headers, GCE_METADATA_HEADER);
```
rubenk: can you combine these two lines?

src/utils_format_stackdriver.c
```diff
@@ -0,0 +1,709 @@
...
+    return (int)status;
+
+  return 0;
+} /* }}} int json_string */
+
+static int json_time(yajl_gen gen, cdtime_t t) {
+  char buffer[64];
+  size_t status;
+
+  status = rfc3339(buffer, sizeof(buffer), t);
```
rubenk: please combine this with the line above.

src/utils_format_stackdriver.c
```diff
@@ -0,0 +1,709 @@
...
+ *   }
+ * }
+ */
+static int format_gcm_resource(yajl_gen gen, sd_resource_t *res) /* {{{ */
+{
+  int status;
+
+  yajl_gen_map_open(gen);
+
+  status = json_string(gen, "type") || json_string(gen, res->type);
```
rubenk: int status

src/utils_format_stackdriver.c
```diff
@@ -0,0 +1,709 @@
...
+
+  if (res->labels_num != 0) {
+    size_t i;
+
+    status = json_string(gen, "labels");
+    if (status != 0)
+      return status;
+
+    yajl_gen_map_open(gen);
+    for (i = 0; i < res->labels_num; i++) {
```
rubenk: for (size_t i = 0...

src/utils_format_stackdriver.c
```diff
@@ -0,0 +1,709 @@
...
+
+  unsigned char const *tmp = NULL;
+  yajl_gen_get_buf(gen, &tmp, &(size_t){0});
+  sstrncpy(buffer, (void const *)tmp, buffer_size);
+
+  yajl_gen_free(gen);
+  return 0;
+} /* }}} int sd_format_metric_descriptor */
+
+/* vim: set sw=2 sts=2 et fdm=marker : */
```
rubenk: please remove any modelines from this PR, we stopped using them a while ago.

Makefile.am
octo: "format_stackdriver" doesn't use libssl itself, but it's exclusively used by code that does.

configure.ac
octo: Fixed.

configure.ac
octo: Done in 212138510373833514ed9792e9b2493bafb8ed19

src/daemon/utils_cache_mock.c
octo: More like a drive-by fix. `"utils_cache.h"` should be included first.

src/utils_gce.c
octo: Yes. Went with `{0}` as its more commonly used to zero initialize.

rubenk:
No worries @octo, glad to see you're back!

vutukuribharat:
Hi 
I Need to Monitor my On-Premises Host using Google Cloud StackDriver . Kindy Suggest me the Steps.

PR #2473, opened 2017-10-09
memcached: Fix CPU usage reporting

rpv-tomsk:
Memcached returns CPU usage as a float values, in seconds.
Example output of Memcached 'stats' command:
```
...
STAT rusage_user 2693.628341
STAT rusage_system 7624.136478
...
```

The Collectd code uses `atoll()` for these values, which leads to huge precision loss.

Before the patch:
memcached_hourly

With patch applied:
memcached_hourly_fixed

This chart match CPU load of 0.3%-0.9%-1.x% which I see at 'top' output.

### DIFF
--- a/src/memcached.c
+++ b/src/memcached.c
@@ -436,9 +436,10 @@ static int memcached_read(user_data_t *user_data) {
      * CPU time consumed by the memcached process
      */
     if (FIELD_IS("rusage_user")) {
-      rusage_user = atoll(fields[2]);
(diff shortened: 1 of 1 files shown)


octo:
@igorpeshansky FYI, this issue could benefit from the "cumulative floating point" datasource type we discussed.

octo:
Cherry-picked into *collectd-5.6* as cb966fce03a0129ecaf028d0ff61bc543b343c5f

Issue #2104, opened 2016-12-13
write_plugin SPOF


faxm0dem:
*   Version of collectd: 5.5.2
*   Operating system / distribution: CentOS7

## Expected behavior

Multiple write plugins are configured, for instance network and write_riemann.
One of the remote servers is failing, for instance riemann.
After some time riemann comes back.

It is expected that only the riemann data has missing points.
The remote collectd does not experience data loss.

## Actual behavior

Both distant servers collectd and riemann experienced data loss.
Even unixsock reports a small fraction of plugins

It seems sane to expect only the riemann server to be missing data, but both collectd and riemann do.
The reason is that collectd's write queue is shared among all write plugins, so when one fails, all fail.

In fact, it's even very hard to kill collectd when the queue is full

## Steps to reproduce

* configure 2 or more write_plugins
* cause failure on one of the destinations
* wait
* experience failure on all write plugins

ludmata:
Same here, I have two write_carbon configured nodes:
`<Plugin write_graphite>
  <Node "node1.example.com">
    Host "5.6.7.8"
    Port "2003"
    Protocol "tcp"
    LogSendErrors true
    Prefix "CL.RL."
    #Postfix "collectd"
    StoreRates true
    AlwaysAppendDS false
    EscapeCharacter "_"
  </Node>
  <Node "node2.example.com">
    Host "1.2.3.4"
    Port "2003"
    Protocol "tcp"
    LogSendErrors true
    Prefix "CL.RL."
    #Postfix "collectd"
    StoreRates true
    AlwaysAppendDS false
    EscapeCharacter "_"
  </Node>
</Plugin>
`
If one of the nodes fail, the other should continue, but unfortunately that is not the case.

faxm0dem:
This has popped up again in production: when a riemann server goes haywire, the collectd server stops getting data. @octo is this a complicated change in core?

rpv-tomsk:
Did you try to use `WriteQueueLimitHigh` and `WriteQueueLimitLow` options?

octo:
The reason for this behavior is that each metric is enqueued once. It is then dequeued by a write thread which calls all write plugins with that one metric sequentially. Presumably, the failed Riemann essentially starved the write threads.

I think the most universal approach would be to use an exponential backoff to skip failing read plugins for a while but still call working write plugins.

rpv-tomsk:
>to skip failing read plugins

You want to say 'write plugins'?

octo:
[...]

PR #2480, opened 2017-10-12
[WiP] src/daemon/plugin.c: Implement exponential backoff for write callbacks.

octo:
This is an untested proof-of-concept implementation of an exponential backoff.

The back-off is tracked in a `write_func_t`, (very) similar to the `read_func_t`. In fact, I think it would make sense to consolidate the two, e.g. by introducing a `exp_backoff_t` structure.

I'd also like to make the exponential back-off base (currently 100 ms) and the maximum back-off interval (currently 60 s) configurable, possibly like this:

```
<LoadPlugin "write_foo">
  BackoffBase 0.5 # 500 ms
  BackoffMax 120 # 2 minutes
</LoadPlugin>
```

Fixes: #2104

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -77,6 +77,19 @@ struct read_func_s {
 };
 typedef struct read_func_s read_func_t;
 
+typedef struct {
+  callback_func_t base;
+  pthread_mutex_t lock;
+
+  /* backoff_interval is the interval for which this callback is diabled. If
+   * zero, the callback is healthy, if non-zero the callback is unhealthy. */
+  cdtime_t backoff_interval;
+  /* retry_time is the wall-clock time when the callback should be retried. If
+   * backoff_interval is non-zero and retry_time is zero, then another thread is
+   * currently retrying this callback. */
+  cdtime_t retry_time;
+} write_func_t;
+
 struct write_queue_s;
 typedef struct write_queue_s write_queue_t;
 struct write_queue_s {
@@ -97,6 +110,7 @@ typedef struct flush_callback_s flush_callback_t;
 static c_avl_tree_t *plugins_loaded = NULL;
 
 static llist_t *list_init;
+/* list_write is a linked list of write_func_t*'s */
 static llist_t *list_write;
 static llist_t *list_flush;
 static llist_t *list_missing;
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Most of write plugins are not intended theis write callback to be called from multiple threads and will lock parallel calls.

These will lock and behave as single-threaded: 

network.c: will lock at global `send_buffer_lock` (only one callback can be registered)
snmp_agent.c: will lock at global `g_agent->lock` (only one callback can be registered)
write_prometheus.c: will lock at global `metrics_lock` (only one callback can be registered).


These will lock and are single-threaded per callback by a fact:

write_http.c: has a callback lock (multiple callbacks can be registered)
write_sensu.c: has a callback lock (multiple callbacks can be registered)


Partially locked (prepare data then do locks) (per-callback lock, multiple callbacks can be registered):

write_graphite.c: format_graphite() then locked send 
write_mongodb.c: wm_create_bson() then locked send 
write_redis.c: Prepare data with format_values() and then do locked send 
write_rienmann.c: Prepare data then do locked send  
write_tsdb.c: Prepare messages then do locked send (each message send protected by a cb lock) 
amqp.c: Prepare data, then do locked publish 
mqtt.c: Prepare data, then do locked publish

No locks:

csv.c: Allows calls from multiple threads to be processed without lock (single callback registered)
write_kafka.c: Allows calls from multiple threads to be processed without lock (multiple callbacks can be registered)
write_log.c: Allows calls from multiple threads to be processed without lock (single callback registered)
rrdcached.c: Allows calls from multiple threads to be processed without lock (single callback registered)

rpv-tomsk:
@octo, What do you think about to use one write thread per one registered write callback?

Many of write plugins use locks (global or per callback), what does these calls from multiple threads not too expedient. (but some plugins are free of locks!).

One thread per callback will allow us not to worry if plugin hangs for a while.
------------------

Some while ago I thought about TCP implementation for `network.c`.
TCP will be useful when hub and agents are in distant locations.
Use of TCP will solve the issue with reordered UDP packets.

The connection lose also needs to be handled by an implementation then. 
The idea was to enqueue metrics on agent until connection to hub recovers.

Such a queue can be implemented on plugin level or on core level. 
Implementation at core level can be used for all TCP-based write plugins, so it preferred over plugin-level queue implementation.

It is proposed to use existing write queue to store data pending write.
For each registered write callback we need to save a reference to a point in writequeue chain.
That is a point where metric should be got to pass it to write callback. 
After a write succeded, we will move a point to a next element of writequeue chain.  
So, no personal queues for plugins - reduce memory usage.

Writequeue elm will be dequeued when all plugins are succeded or when time to live expired.

One write thread per one registered write callback can be a part of such implementation, to make things simple. 
If use many write threads (which is still useful for `rrdcached.c`, `write_log.c`,`write_kafka.c`, `csv.c`, as they are lock-free), then write thread will need to select which plugin/callback to proceed. Exponential-backoff and locking will come again then.

rpv-tomsk:
So what do you think about my idea not only do write-backoff, but also save data in queue for possible future delivery?

rpv-tomsk:
>It would probably be a good idea to design such a change "on paper" first before jumping into an implementation.

Of course, that is in proggress "on paper". Yes, there are many corners and changes....

octo:
@rpv-tomsk, @maryamtahhan: → *Per-(write)-plugin threads* #2486

octo:
One problem that just came to mind: if there is only one callback and *n* threads and there is a problem, the remaining *n–1* threads will "handle" the queue very quickly, dequeuing each *value list*, not calling the callback (as designed) and freeing the *value list*. That is not what we want.

Issue #2475, opened 2017-10-10
write_prometheus plugin fails with "MHD_start_daemon() failed" (on version 5.7.2.44)


tcarreira:
*   Version of collectd: 5.7.2.44.gbe906f9
*   Operating system / distribution: Debian/wheezy

This last update (5.7.2.44.gbe906f9) makes the write_prometheus fail. 
On the previous version (I had 5.7.2.37.g45492ba) it worked perfectly.

The logs only show this:

```
[2017-10-10 18:35:52] plugin_load: plugin "write_prometheus" successfully loaded.
[2017-10-10 18:35:52] write_prometheus plugin: MHD_start_daemon() failed.
[2017-10-10 18:35:52] write_prometheus plugin: MHD_start_daemon() failed.
[2017-10-10 18:35:52] Initialization of plugin `write_prometheus' failed with status -1. Plugin will be unloaded.
[2017-10-10 18:35:52] plugin_unregister_read: No such read function: write_prometheus
[2017-10-10 18:35:52] Error: one or more plugin init callbacks failed.
```

## Steps to reproduce

*  activate write_prometheus plugin

I will be glad to help (just don't know what to do)

octo:
Judging from the commit range, this is a regression introduced in #2466.

octo:
Hi @tcarreira, thanks for reporting this! Could you please provide the version of *libmicrohttpd* you're using? E.g. the output of:

```
apt-cache policy libmicrohttpd-dev
```

tcarreira:
Sure!
```
# apt-cache policy libmicrohttpd-dev
libmicrohttpd-dev:
  Installed: 0.9.20-1+deb7u1
```

octo:
Thanks! Could you try running collectd in the foreground (`collectd -f`) to see if the HTTP library we use logs anything to `STDERR`?

tcarreira:
There is nothing relevant in the output:
```
# collectd -C /etc/collectd/collectd.conf -f
option = Hostname; value = xxxx;
option = FQDNLookup; value = false;
Done parsing `/usr/share/collectd/types.db'
option = Interval; value = 60.000000;
option = Timeout; value = 2.000000;
option = ReadThreads; value = 5.000000;
option = WriteThreads; value = 5.000000;
Created new plugin context.
[2017-10-10 23:07:13] plugin_load: plugin "logfile" successfully loaded.
[2017-10-10 23:07:13] type = logfile, key = LogLevel, value = info
```

[...]

PR #2481, opened 2017-10-12
Fix 2475

tcarreira:
Fixes: #2475

### DIFF
--- a/src/collectd-python.pod
+++ b/src/collectd-python.pod
@@ -475,7 +475,7 @@ Methods defined here:
 
 =over 4
 
-=item B<dispatch>([type][, values][, plugin_instance][, type_instance][, plugin][, host][, time][, interval]) -> None.  Dispatch a value list.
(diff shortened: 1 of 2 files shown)


octo:
FTR, the difference to *master* appears to be that the `http_port` argument is non-zero.

tcarreira:
@octo The only difference from my commit is indeed the `http_port` argument (which was the argument before the bug was introduced in the commit 38617fa88d22997e91176ef901f842f384ff3711  (line 768 previously to the commit))

It seems to me a really quick and simple fix.

I applied this commit to the `collectd-5.7` branch, which already had the conflict on https://github.com/collectd/collectd/pull/2481/conflicts

Would you be able to solve this conflict (as I do not have context on it) and hotfix it? 
This bug is preventing me from using collectd (and compiling it for hundreds of servers does not seem appropriate for me right now).

Thank you

octo:
Cherry-picked into *collectd-5.7* as e21b137ec98244a6dc8da6898504ec3763a2ce13

PR #2482, opened 2017-10-13
Implementation of "Prefix" option for write_http plugin

outini:
I needed to be able to override the standard "collectd" prefix applied to metrics send to KairosDB.

A new option "Prefix" is now available when using the KAIROSDB format with the write_http plugin.
If not specified, this option defaults to "collectd".

### DIFF
--- a/src/utils_format_kairosdb.c
+++ b/src/utils_format_kairosdb.c
@@ -184,7 +184,7 @@ static int value_list_to_kairosdb(char *buffer, size_t buffer_size, /* {{{ */
                                   int store_rates,
                                   char const *const *http_attrs,
                                   size_t http_attrs_num, int data_ttl,
-                                  char *metrics_prefix) {
+                                  char const *metrics_prefix) {
   char temp[512];
   size_t offset = 0;
   int status;
@@ -265,7 +265,7 @@ static int format_kairosdb_value_list_nocheck(
     size_t *ret_buffer_fill, size_t *ret_buffer_free, const data_set_t *ds,
     const value_list_t *vl, int store_rates, size_t temp_size,
     char const *const *http_attrs, size_t http_attrs_num, int data_ttl,
-    char *metrics_prefix) {
+    char const *metrics_prefix) {
   char temp[temp_size];
   int status;
 
@@ -341,7 +341,7 @@ int format_kairosdb_value_list(char *buffer, /* {{{ */
                                const data_set_t *ds, const value_list_t *vl,
                                int store_rates, char const *const *http_attrs,
                                size_t http_attrs_num, int data_ttl,
-                               char *metrics_prefix) {
+                               char const *metrics_prefix) {
   if ((buffer == NULL) || (ret_buffer_fill == NULL) ||
       (ret_buffer_free == NULL) || (ds == NULL) || (vl == NULL))
     return -EINVAL;
--- a/src/utils_format_kairosdb.c
+++ b/src/utils_format_kairosdb.c
@@ -183,7 +183,8 @@ static int value_list_to_kairosdb(char *buffer, size_t buffer_size, /* {{{ */
                                   const data_set_t *ds, const value_list_t *vl,
                                   int store_rates,
                                   char const *const *http_attrs,
-                                  size_t http_attrs_num, int data_ttl) {
+                                  size_t http_attrs_num, int data_ttl,
+                                  char *metrics_prefix) {
   char temp[512];
   size_t offset = 0;
   int status;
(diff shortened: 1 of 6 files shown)


beorn-:
that's an improvement. :+1:

jaroug:
I need this too :+1:

octo:
LGTM, thanks @outini!

outini:
You are welcome !

rpv-tomsk:
Before this will be merged, I want to ask: Maybe newly added option should be named `MetricsPrefix` ?

octo:
@rpv-tomsk I'd say `Prefix` is descriptive enough, I don't know which other prefix users would confuse this with.

rpv-tomsk:
Also this needs some `const` to be added. I will take care about this slightly later,  if Denis will not do that before.

outini:
Hi,

I wanted the code to be explicite and used _metrics_prefix_ to be sure no one would be confuse with the possible use of an internal prefix. However, I think that the use of _Prefix_ in configuration options should be pretty understandable. It is also already used in configuration of other plugins such as _write_graphite_ for example.

I also tried to use _const_ instead of _char *_ but the use of _cf_util_get_string_ then raise an error as it expects argument to be a _char *_. It is possible I'm doing something wrong, do you have any advice to use _const_ in this case ?

src/utils_format_kairosdb.h
```diff
@@ -41,7 +41,8 @@ int format_kairosdb_value_list(char *buffer, size_t *ret_buffer_fill,
                                size_t *ret_buffer_free, const data_set_t *ds,
                                const value_list_t *vl, int store_rates,
                                char const *const *http_attrs,
-                               size_t http_attrs_num, int data_ttl);
+                               size_t http_attrs_num, int data_ttl,
+                               char *metrics_prefix);
```
octo: This is the missing `const` @rpv-tomsk was talking about. This should be `char const *metrics_prefix`.

src/write_http.c
```diff
@@ -80,6 +84,7 @@ struct wh_callback_s {
   pthread_mutex_t send_lock;
 
   int data_ttl;
+  char *metrics_prefix;
```
octo: Don't worry about this one, it can stay a `char *metric_prefix`.

src/utils_format_kairosdb.h
outini: ok thanks, I'm on it, I'll amend my previous commit.

src/write_http.c
outini: Great, I'll just modify the format files and leave this one untouched then.

src/utils_format_kairosdb.c
```diff
@@ -183,7 +183,8 @@ static int value_list_to_kairosdb(char *buffer, size_t buffer_size, /* {{{ */
                                   const data_set_t *ds, const value_list_t *vl,
                                   int store_rates,
                                   char const *const *http_attrs,
-                                  size_t http_attrs_num, int data_ttl) {
+                                  size_t http_attrs_num, int data_ttl,
+                                  char *metrics_prefix) {
```
rpv-tomsk: IMHO, this one also needs to be altered.

outini:
Hi !
I have done the necessary changes. Everything's building well by my side.
Let me know if this is good for you.

outini:
My pull-request doesn't pass the ckeck, the only error I can see is:

    + make -sk check
    make: *** No rule to make target 'check'.

Do you have any advice on the cause of such an error ?

rpv-tomsk:
Of course, it's good. It can not be otherwise. 
Thanks for your work, Denis! 

Also I have added a simple check for `strdup()` result and, as a separate commit, was added the ability to disable prefix when configuration specifies empty prefix.

So, now this requires your review. 
I don't use this feature at practice so I don't know is that solution is correct. 
Please review. Thanks.

outini:
You're totally right on this feature, I missed this use case.
I agree, this possibility of having an empty prefix really makes sense.

Thank you for your help !

outini:
After build and tests on my platform, everything is going well with empty, defaut, and specific prefix.
This is a go.

rpv-tomsk:
Ok, merged!

Many thanks for your work, Denis! 
Hope you will continue to work on Collectd project! 
We have many interesting tasks for developers!

outini:
thanks!

It's been 10 years since I wrote C code :)
This is the occasion for me to revive/improve my (very basic) skills.
I'll probably come back with new plugins as some of my needs are not covered by existing ones :)

Should I close this PR or let you do ?

PR #2483, opened 2017-10-13
filecount: Controls whether or not to include only regular files in the count

alex2grad:
The new option 'IncludeOnlyRegular' added to have possibility count the files on '/proc' or '/sys'


src/filecount.c
```diff
@@ -498,37 +501,40 @@ static int fc_read_dir_callback(const char *dirname, const char *filename,
       return 0;
   }
 
-  if (dir->mtime != 0) {
-    time_t mtime = dir->now;
+  if (S_ISREG(statbuf.st_mode)) {
```
octo: This is a very long block. Please do this instead:

```c
if (!S_ISREG(statbuf.st_mode)) {
  dir->files_num++;
  return 0;
}
/* rest of the block */
```

src/collectd.conf.pod
```diff
@@ -2657,6 +2657,11 @@ this many threads will be started immediately setting this to a very high
 value will waste valuable resources. Defaults to B<5> and will be forced to be
 at most B<16384> to prevent typos and dumb mistakes.
 
+=item B<IncludeOnlyRegular> I<true>|I<false>
```
octo: I'm not a fan of the name. How about `RegularOnly` or `IncludeIrregular`?

octo: This option has been added to the wrong plugin (*email*).

octo:
Thanks @alex2grad!

PR #2494, opened 2017-10-19
ping plugin: format #1636 by nward to current style

trustchk:
This PR updates the original #1636 to be mergeable after the format changes that happened since it was made. I swapped the fields in src/collectd.conf.pod, however, to be consistent with the documentation. It would be really awesome to have this in 5.8.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -23,6 +23,7 @@ collectd.conf - Configuration for the system statistics collection daemon B<coll
   LoadPlugin ping
   <Plugin ping>
     Host "example.org"
+    Host "example.org" "example"
     Host "provider.net"
   </Plugin>
 
@@ -6113,10 +6114,11 @@ Available configuration options:
 
 =over 4
 
-=item B<Host> I<IP-address>
+=item B<Host> I<IP-address> [I<Alias>]
 
 Host to ping periodically. This option may be repeated several times to ping
-multiple hosts.
+multiple hosts. If an alias is specified, it will be used as the type
+instance, rather than the IP-Address.
 
 =item B<Interval> I<Seconds>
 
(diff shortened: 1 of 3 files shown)


trustchk:
That's odd, the jenkins build complains about timestamps being in the future, I don't think that's related to this PR.

rpv-tomsk:
Hi! Thanks for your work on this!

Can you do more deep work, for example switch the module to use of `plugin_register_complex_config`?

rpv-tomsk:
cc @nward

rpv-tomsk:
Guys, why you don't use /etc/hosts or DNS to reach the same result?

src/ping.c
```diff
@@ -462,15 +465,33 @@ static int ping_config(const char *key, const char *value) /* {{{ */
       return 1;
     }
 
+    if ((value_alias = strchr(value, ' ')) == NULL)
```
octo: For doing non-trivial configurations, the simple config callback is not the right choice anymore. Instead, this should use the "complex" config callback to get the two values individually.

src/collectd.conf.pod
```diff
@@ -6113,10 +6114,11 @@ Available configuration options:
 
 =over 4
 
-=item B<Host> I<IP-address>
+=item B<Host> I<IP-address> [I<Alias>]
```
octo: I'd prefer to have a config along the lines of:

```
<Host "app-prod-01-customer01.example.com">
  Alias "customer01.example.com"
</Host>
```

Have you checked the manpage if other plugins already use the option name `Alias`? If other plugins do something similar, we should try to use the same terminology if possible.

dago:
@trustchk Are you still interested in this enhancement? If yes, would you mind reviewing the comments from octo?

nward:
If @trustchk doesn't reply in a week-ish, I can have a look at these suggestions/comments and do another PR

Issue #2426, opened 2017-09-08
LVM plugin does not report thin pool usage


ceecko:
*   Version of collectd: 
*   Operating system / distribution: CentOS 7.3.1611 (Core). Kernel 3.10.0-514.21.1.el7.x86_64

## Expected behavior

It seems that according to https://github.com/collectd/collectd/pull/603 the LVM plugin should report thin-pool free/used metrics.

## Actual behavior

The LVM plugin does not report free/used metrics for thin-pools.

## Steps to reproduce

* Create thin-pool through lvm
* Enable LVM plugin
* Observe metrics

```
# vgs
  VG   #PV #LV #SN Attr   VSize   VFree
  data   1   2   0 wz--n- 100.00g    0
```

```
# lvs
  LV          VG   Attr       LSize  Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  data        data -wi-ao----  9.00g
  docker_thin data twi-aotz-- 90.00g             33.74  6.03
```

rpv-tomsk:
> Version of collectd:
Which version of Collectd are you using?

ceecko:
The version is: `5.7.2.sfx0`
I have managed to resolve the issue by adding more capabilities in systemd
`CapabilityBoundingSet=CAP_SYS_ADMIN`

octo:
Hi @ceecko, thank you very much for your update. `CAP_SYS_ADMIN` seems a bit excessive for the read-only access required by *collectd*. Is there any less privileged capability that will get you there? We should add a capability check to the plugin so users get an informative warning message in case capabilities are missing.

Best regards,
—octo

ceecko:
@octo I have tried only a couple. Since I'm not familiar with the capabilities and I don't mind excessive persmissions. In this case I opted for the `CAP_SYS_ADMIN`. In case less privileges are sufficient, I'll be happy to use them.

octo:
@ceecko I did some a quick search and from what I found it sounds like users with *read* permissions on `/dev/lvm` can read LVM metrics. Can you confirm this? —octo

ceecko:
@octo I'm running CentOS 7 and there's no such file as /dev/lvm

PR #2499, opened 2017-10-22
lvm update: Check for the CAP_SYS_ADMIN capability.

octo:
Fixes: #2426

### DIFF
--- a/src/lvm.c
+++ b/src/lvm.c
@@ -21,12 +21,16 @@
  *   Benjamin Gilbert <bgilbert at backtick.net>
  **/
 
-#include <lvm2app.h>
-
 #include "collectd.h"
 
 #include "common.h"
 #include "plugin.h"
+
+#include <lvm2app.h>
+
+#ifdef HAVE_SYS_CAPABILITY_H
+#include <sys/capability.h>
+#endif /* HAVE_SYS_CAPABILITY_H */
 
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
```
src/lvm.c:190:12: error: conflicting types for 'lvm_init'
 static int lvm_init(void) {
            ^
In file included from src/lvm.c:29:0:
/usr/include/lvm2app.h:260:7: note: previous declaration of 'lvm_init' was here
 lvm_t lvm_init(const char *system_dir);
       ^
Makefile:6048: recipe for target 'src/lvm_la-lvm.lo' failed
make[1]: *** [src/lvm_la-lvm.lo] Error 1
```

Please, don't merge PR untill it passes CI. ;-)

rpv-tomsk:
Failed compilaton issue was fixed.

octo:
Sorry, it failed so fast I assumed it was the clock issue again 😢

We really need to get that fixed!

rpv-tomsk:
>We really need to get that fixed!

What is the issue cause?

octo:
Me neither, that's the problem. I suspect a VM is missing a kernel module, causing it's (virtual) "hardware" clock to drift. When it's sever enough, the next VM will complain about a timestamp in the future, but every once in a while they're in sync enough for everything to work …

PR #2505, opened 2017-10-24
Correction of tracing from child process and new EAL params support

jiriproX:
Correction of tracing from child process
- There was read 16kb from pipe but only less then 1kb was traced.
  Rest of trace data was lost.
Added support for two EAL parameters:
  LogLevel - optional EAL paramenter which enable debug traces from rte libs
             If parameter is not used default trace level = 7 (INFO) is used
             LogLevel "8" - (DEBUG) can be set to collectd.conf into dpdkstat
             EAL section
  RteDriverLibPath - optional EAL parameter which enable loading of shared
                     pmd driver libs. Param value can be full path to single
                     pmd diriver lib or directory where pmd driver libs are
                     located. E.g.: "/usr/lib/dpdk-pmd/librte_pmd_i40e.so"
                                 or "/usr/lib/dpdk-pmd"

Signed-off-by: Jiri Prokes <jirix.x.prokes@intel.com>

### DIFF
--- a/src/utils_dpdk.c
+++ b/src/utils_dpdk.c
@@ -45,7 +45,9 @@
 
 #define DPDK_DEFAULT_RTE_CONFIG "/var/run/.rte_config"
 #define DPDK_EAL_ARGC 10
-// Complete trace should fit into 1024 chars
+// Complete trace should fit into 1024 chars. Trace contain some headers
+// and text together with traced data from pipe. This is the reason why
+// we need to limit DPDK_MAX_BUFFER_SIZE value.
 #define DPDK_MAX_BUFFER_SIZE 896
 #define DPDK_CDM_DEFAULT_TIMEOUT 10000
 
@@ -731,7 +733,7 @@ static void dpdk_helper_check_pipe(dpdk_helper_ctx_t *phc) {
     DEBUG("%s:dpdk_helper_check_pipe: read nbytes=%d", phc->shm_name, nbytes);
     if (nbytes <= 0)
       break;
-    buf[nbytes] = '\n';
+    buf[nbytes] = '\0';
     sstrncpy(out, buf, (nbytes + 1));
     DEBUG("%s: helper process:\n%s", phc->shm_name, out);
   }
--- a/src/utils_dpdk.c
+++ b/src/utils_dpdk.c
@@ -44,8 +44,9 @@
 #include "utils_dpdk.h"
 
 #define DPDK_DEFAULT_RTE_CONFIG "/var/run/.rte_config"
-#define DPDK_EAL_ARGC 5
-#define DPDK_MAX_BUFFER_SIZE (4096 * 4)
+#define DPDK_EAL_ARGC 10
+//Complete trace should fit into 1024 chars
+#define DPDK_MAX_BUFFER_SIZE 896
 #define DPDK_CDM_DEFAULT_TIMEOUT 10000
 
 enum DPDK_HELPER_STATUS {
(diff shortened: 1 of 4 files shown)


src/utils_dpdk.c
```diff
@@ -697,6 +717,8 @@ static void dpdk_helper_check_pipe(dpdk_helper_ctx_t *phc) {
       .fd = phc->pipes[0], .events = POLLIN,
   };
   int data_avail = poll(&fds, 1, 0);
+  DEBUG("%s:dpdk_helper_check_pipe: poll data_avail=%d", phc->shm_name,
```
octo: Ideally this should have a "dpdk_common:" prefix like the other debug messages.

src/utils_dpdk.h
```diff
@@ -51,6 +51,8 @@ struct dpdk_eal_config_s {
   char memory_channels[DATA_MAX_NAME_LEN];
   char socket_memory[DATA_MAX_NAME_LEN];
   char file_prefix[DATA_MAX_NAME_LEN];
+  char log_level[DATA_MAX_NAME_LEN];
+  char rte_driver_lib_path[DATA_MAX_NAME_LEN];
```
octo: I'd use `PATH_MAX` as the size of this buffer – the define is better suited for this purpose.

src/utils_dpdk.c
```diff
@@ -705,10 +727,12 @@ static void dpdk_helper_check_pipe(dpdk_helper_ctx_t *phc) {
     }
   }
   while (data_avail) {
-    int nbytes = read(phc->pipes[0], buf, sizeof(buf));
+    int nbytes = read(phc->pipes[0], buf, (sizeof(buf) - 1));
+    DEBUG("%s:dpdk_helper_check_pipe: read nbytes=%d", phc->shm_name, nbytes);
     if (nbytes <= 0)
       break;
-    sstrncpy(out, buf, nbytes);
+    buf[nbytes] = '\n';
```
octo: Why `\n`? Did you intend to nil-terminate the string? In that case use `buf[nbytes] = 0;` instead.

src/utils_dpdk.c
```diff
@@ -705,10 +727,12 @@ static void dpdk_helper_check_pipe(dpdk_helper_ctx_t *phc) {
...
+    sstrncpy(out, buf, (nbytes + 1));
```
octo: Use `sizeof(out)` as the third parameter (after making sure that `buf` is nil-terminated, see above).

src/utils_dpdk.c
```diff
@@ -493,6 +504,15 @@ static int dpdk_helper_eal_init(dpdk_helper_ctx_t *phc) {
   argp[argc++] = "--proc-type";
   argp[argc++] = "secondary";
 
+  if (strcasecmp(phc->eal_config.log_level, "") != 0) {
```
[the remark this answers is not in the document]
jiriproX: You are right. Nevertheless I just reused what was used for --socket-mem. I would let it there.

src/utils_dpdk.c
jiriproX: I don't think so. It is consistent with traces which were already present there.

src/utils_dpdk.c
jiriproX: It is just defined char. sstrncpy ignore char at that position anyway.  Nevertheless you are right  '\0' will be better here.

src/utils_dpdk.c
jiriproX: sstrncpy does not check nil in string, so it will copy complete buffer. In our case only real data are copied.

src/utils_dpdk.h
jiriproX: Done

src/utils_dpdk.c
```diff
@@ -44,8 +44,9 @@
 #include "utils_dpdk.h"
 
 #define DPDK_DEFAULT_RTE_CONFIG "/var/run/.rte_config"
-#define DPDK_EAL_ARGC 5
-#define DPDK_MAX_BUFFER_SIZE (4096 * 4)
+#define DPDK_EAL_ARGC 10
+// Complete trace should fit into 1024 chars
```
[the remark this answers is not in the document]
jiriproX: I've updated comment.

src/utils_dpdk.c
octo: `sstrncpy()` expects a **nil-terminated** string as second argument and copies it into the memory pointed to by the first argument. It will, however, not copy more than the third argument.

While your variant also works, it is uncommon and therefore confusing. Please use `sizeof(out)` instead of `nbytes + 1`.

jiriproX: I've changed that as you requested. Nevertheless if I understood documentation right then it is less efficient than original one. It always write complete buffer even if read nbytes  is only few chars. Rest of buffer is filled with zeros.

PR #2513, opened 2017-11-01
Append configuration boolean option 'StoreTimestamp' to MQTT plugin

ilgizar:
Default: true
Description: it control append or not timestamp before value on MQTT replicas
Explanation: instead "1509519673.14:123" with "StoreTimestamp=false" you push to MQTT clear value "123"

### DIFF
--- a/src/mqtt.c
+++ b/src/mqtt.c
@@ -75,6 +75,7 @@ struct mqtt_client_conf {
   /* For publishing */
   char *topic_prefix;
   _Bool store_rates;
+  _Bool store_timestamp;
   _Bool retain;
 
   /* For subscribing */
@@ -500,6 +501,13 @@ static int mqtt_write(const data_set_t *ds, const value_list_t *vl,
   }
 
   status = format_values(payload, sizeof(payload), ds, vl, conf->store_rates);
+  if (!conf->store_timestamp) {
+    char *p, *t;
+    strtok_r(payload, ":", &t);
+    p = strtok_r(NULL, ":", &t);
+    strcpy(payload, p);
+  }
+
   if (status != 0) {
     ERROR("mqtt plugin: format_values failed with status %d.", status);
     return status;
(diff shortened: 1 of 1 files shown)


src/mqtt.c
```diff
@@ -499,7 +505,22 @@ static int mqtt_write(const data_set_t *ds, const value_list_t *vl,
...
+  }
+
   status = format_values(payload, sizeof(payload), ds, vl, conf->store_rates);
+  if (!conf->store_timestamp) {
+    char *p = strchr(payload, ':');
+    int len = strlen(p);
+    if (len > 0) {
+      p = p + 1;
+    }
+    strcpy(payload, p);
```
octo: `payload` and `p` overlap, which will cause `strcpy` to sefault. Use `memmove()` instead (it explicitly supports overlapping regions of memory).

src/mqtt.c
```diff
@@ -608,6 +634,10 @@ static int mqtt_config_publisher(oconfig_item_t *ci) {
       ERROR("mqtt plugin: Unknown config option: %s", child->key);
   }
 
+  if (conf->pattern != NULL) {
+    regcomp(&conf->topic_pattern, conf->pattern, REG_EXTENDED);
```
octo: Please check the return value of `regcomp()`. Regular expressions are easy to get wrong, and we have to report that fact to the user.

src/mqtt.c
```diff
@@ -75,7 +76,10 @@ struct mqtt_client_conf {
   /* For publishing */
   char *topic_prefix;
   _Bool store_rates;
+  _Bool store_timestamp;
   _Bool retain;
+  char *pattern;
```
octo: There's no need to keep this string around – it's not read outside of the config function, which stores it. Instead, make this a local variable in the config callback.

dago:
@ilgizar I know has been quite some time since your contribution, but would you be willing to rebase this PR and address the suggestions on the code? I would like to include this PR in the upcoming 5.11 release.

PR #2517, opened 2017-11-02
libcollectdclient: add defines for AIX

trustchk:
This adds two defines that are missing on AIX because they don't have
the endian.h header file. AIX is always big endian, so no further check
is required for that. As I have no access to IBMs XLC compiler, this
patch limits the definitions to GNU GCC.

### DIFF
--- a/src/libcollectdclient/network_parse.c
+++ b/src/libcollectdclient/network_parse.c
@@ -43,10 +43,23 @@
 #include <endian.h>
 #elif HAVE_SYS_ENDIAN_H
 #include <sys/endian.h>
-#elif defined(_AIX) && defined(__GNUC__)
-/* AIX is always big endian */
-#define be16toh(x) (x)
-#define be64toh(x) (x)
+#else /* fallback */
+__attribute__((const)) static uint16_t be16toh(uint16_t n) {
+  uint8_t tmp[2];
+  memmove(tmp, &n, sizeof(tmp));
+
+  return ((uint16_t)tmp[0] << 8) | ((uint16_t)tmp[1] << 0);
+}
+
+__attribute__((const)) static uint64_t be64toh(uint64_t n) {
+  uint8_t tmp[8];
+  memmove(tmp, &n, sizeof(tmp));
+
+  return ((uint64_t)tmp[0] << 56) | ((uint64_t)tmp[1] << 48) |
+         ((uint64_t)tmp[2] << 40) | ((uint64_t)tmp[3] << 32) |
+         ((uint64_t)tmp[4] << 24) | ((uint64_t)tmp[5] << 16) |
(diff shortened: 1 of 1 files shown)


octo:
Hi @trustchk,

thank you very much for your patch! It looks reasonable to me, but since I don't have experience with AIX, I'd prefer if someone with AIX experience could review.

@manuelluis, @cyang16: you've done the most recent AIX changes. Would one of you like to take a look at this?

Best regards,
—octo

trustchk:
At least for GCC, there's a set of builtins that should cover most of this, see https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Other-Builtins.html although at least in the past, there were doubts that these are efficient: http://hardwarebug.org/2010/01/14/beware-the-builtins/
If using these, most of the platforms are covered (assuming CLANG knows these too), but if someone uses the platform compiler (xlC for AIX and sun studio(?) for Solaris), things might look a little different. I don't have xlC, sadly.

If only GCC and clang are to be supported, this should work for AIX and quite possibly for Solaris as well, since these two don't use GLIBC but have their own libc:
```C
#if !defined(__GNU_LIBRARY__) && !defined(__GLIBC__)
#if BYTE_ORDER == LITTLE_ENDIAN
        #define htobe64(x) __builtin_bswap64(x)
        #define htole64(x) (x)
        #define be64toh(x) __builtin_bswap64(x)
        #define le64toh(x) (x)
#elif BYTE_ORDER == BIG_ENDIAN
        #define htobe64(x) (x)
        #define htole64(x) __builtin_bswap64(x)
        #define be64toh(x) (x)
        #define le64toh(x) __builtin_bswap64(x)
#else
        #error byte order not supported
#endif
#endif /* __GLIBC__ */
```
This would also require to use ntohs etc. like you mentioned in the solaris issue.

octo:
Hi @trustchk,

I pushed something to your branch. Can you let me know if that works for you? It should fix this issue for AIX, Solaris and other systems without `<endian.h>`.

Best regards,
—octo

src/libcollectdclient/network_parse.c
```diff
@@ -43,10 +43,16 @@
...
-#define be16toh(x) (x)
-#define be64toh(x) (x)
+#else /* fallback */
+#include <arpa/inet.h>
+#define be16toh(x) ntohs(x)
+__attribute__((const)) static uint64_t be64toh(uint64_t n) {
+  uint8_t tmp[8];
+  memmove(tmp, &n, sizeof(tmp));
+
+  return (tmp[0] << 56) | (tmp[1] << 48) | (tmp[2] << 40) | (tmp[3] << 32) |
```
octo: CC: @SeanCampbell

PR #2526, opened 2017-11-08
postgresql: fix last sum() related parse errors

mfournier:
These queries return empty values when a database doesn't have any
tables yet, which results in collectd logging this error every Interval:

```
db query utils: udb_result_submit: Parsing `' as derive failed.
```

Related to #1905

### DIFF
--- a/src/postgresql_default.conf
+++ b/src/postgresql_default.conf
@@ -38,9 +38,9 @@
 </Query>
 
 <Query queries>
-	Statement "SELECT sum(n_tup_ins) AS ins, \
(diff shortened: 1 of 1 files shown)


mfournier:
Thanks for taking a look @octo !

I just cherry-picked this to collectd-5.7 now as 19b51c4ab.

I'm a bit puzzled by the CI error (unrelated to the patch itself), probably due to https://github.com/collectd/collectd-ci/commit/9442d1603. I won't bother digging further as 5.6 is retired.

PR #2528, opened 2017-11-09
nut: add readout of power

elratt0r:
ChangeLog: nut: Adding support for the "output.realpower" value from the ups code.

Calculating the value from voltage and current does not work for all USV's,
so this would be a benefit to have, when the USV provides the value.

### DIFF
--- a/src/nut.c
+++ b/src/nut.c
@@ -316,7 +316,7 @@ static int nut_read(user_data_t *user_data) {
       else if (strcmp("output.voltage", key) == 0)
         nut_submit(ups, "voltage", "output", value);
       else if (strcmp("output.realpower", key) == 0)
-        nut_submit(ups, "energy", "output", value);
+        nut_submit(ups, "realpower", "output", value);
       else if (strcmp("output.power", key) == 0)
         nut_submit(ups, "power", "output", value);
     } else if (strncmp("ups.", key, 4) == 0) {
--- a/src/nut.c
+++ b/src/nut.c
@@ -316,6 +316,8 @@ static int nut_read(user_data_t *user_data) {
       else if (strcmp("output.voltage", key) == 0)
         nut_submit(ups, "voltage", "output", value);
       else if (strcmp("output.realpower", key) == 0)
+        nut_submit(ups, "energy", "output", value);
+      else if (strcmp("output.power", key) == 0)
         nut_submit(ups, "power", "output", value);
     } else if (strncmp("ups.", key, 4) == 0) {
       if (strcmp("ups.load", key) == 0)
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
Hi! 

Thanks for your report and improvement.

As I see, Nut has both `output.realpower` and `output.power`. 
What do you think, how should we proceed if we get both values from Nut?

--
Nut variables list: http://networkupstools.org/docs/developer-guide.chunked/apas01.html

elratt0r:
Hi!
Good question. I've chosen realpower, because the readout in Watt seems to be more useful.
Actually most of the plugins use the power type with values based on Watt. (ted, mic for example).

On the contary, the nut-plugin also reads ups.power, which comes in Voltampere and
is submitted to power-ups.

I would suggest the following:
output.realpower -> energy-output
output.power -> power-output

So at least power-* shares the same unit and does not break existing graph panels, etc.

Greetings

octo:
In AC systems, there is an important difference between "apparent power" and "real power" (→ [wikipedia](https://en.wikipedia.org/wiki/AC_power)) – by elimination I guess that "power" refers to apparent power. If nut reports both, we can actually calculate another very important metric, cos(φ) = *real power* / *apparent power*.

I think it would be best to report them as such, i.e.:

*   power → `power-apparent`
*   realpower → `power-real`

elratt0r:
The reports depend on the USV, same for the correctness of the values.
My USV (Powerware 5105, 1500VA/960W) for example gives the following values:
```
output.current: 1.14
output.power: 0.26
output.realpower: 180.00
output.realpower.nominal: 960.00
output.voltage: 231.80
ups.power.nominal: 1500.00
```
As you see, `voltage * current ≠ realpower`
(The nominal values are static values that don't change.)
In my eyes, it makes sense to collect as many (useful) values as possible.
Without calculations in the source, based on the problems shown above.

I think your approach is not applicable, because the type-instance is loaded with
the matching "connector" (input, output -> power-input, power-output).

How about adding something more specific to the types.db:

- [type_instance].power → `voltampere-[type_instance]`
- [type_instance].realpower → `watt-[type_instance]`

rpv-tomsk:
The types.db mostly has types, not units. Volts, Amperes, Volt-Amperes - are units.

So, I think to add `realpower` into `types.db` will be nicer than 'voltamperes' and 'watts'.

Also, the Collectd has a 'hack' for similar case in `impi` plugin: the type instance can have a unit inside  (as 'percent-' prefix).

We can do the similar here. The `voltage * current` can be reported as `power.voltampere-output`, and real power can be reported as `power.watt-output` then. (The form is `type.type_instance`).

elratt0r:
@rpv-tomsk Sorry for the late reply. You're right about the types/units. I followed you suggestion and updated the PR. Could you please look into it?

basilfx:
I'm also interested in this PR. I have a UPS where I would be interested in `ups.realpower` instead of `ups.power`. Anything I can do to help?

PR #2536, opened 2017-11-15
Fix documentation typo for PostgreSQL plugin

feikesteenbergen:
In collectd/src/postgresql_default.conf the name for the backends
query is plural, yet in the documentation it is singular. This caused
me some lost time when running this plugin.

It is also wrongly listed on the wiki:

https://collectd.org/wiki/index.php/Plugin:PostgreSQL

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1171,7 +1171,7 @@
 #	<Database bar>
 #		Interval 60
 #		Service "service_name"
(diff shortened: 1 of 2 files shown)


octo:
Thanks for your patch @feikesteenbergen!

Cherry-picked into *collectd-5.7* as 9c6a92b8a30d63f6b476ab0bf0d166f3de41a7b1

PR #2541, opened 2017-11-15
src/daemon/meta_data.c: Fix double unlock.

octo:
CID: 179245


octo:
The build failed for an unrelated issue (collectd/collectd-ci#28)

PR #2547, opened 2017-11-17
snmp plugin: Ensure that index is within bounds.

octo:
CID: 81389, 81390


rpv-tomsk:
What does "CID: 81389, 81390" mean?

octo:
@rpv-tomsk "Coverity ID". These are bugs found by the static code analyser. I should probably invite you to that project …

PR #2551, opened 2017-11-17
disk: in linux, reset the disk when it disappear from /proc/diskstats

klnikita:
If you don't reset diskstats_t for a given diskname when it disappear and the diskname is reused later (for exemple an iSCSI disk) the "disk" plugin detect an "counter wrap around" because linux /proc/diskstats are
resetted to 0 for that diskname. Because of that the plugin send incorrect stats for the new disk.

Note1: the number of bytes written is calculated by multiplying the number of sector written by 512, which seems incorrect if those sector are 4k ones but I haven't tested that.

Note2: the patch proposed here is racy for the disk removal detection, a better way to detect it would be through udev, but I guess it will necessitate its own thread and more linux specific code.

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -662,6 +662,7 @@ static int disk_read(void) {
   char buffer[1024];
 
   char *fields[32];
+  static unsigned int local_poll_count = 0;
 
   derive_t read_sectors = 0;
   derive_t write_sectors = 0;
@@ -714,6 +715,7 @@ static int disk_read(void) {
         disklist = ds;
       else
         pre_ds->next = ds;
+      ds->poll_count = local_poll_count;
     }
 
     is_disk = 0;
@@ -875,6 +877,27 @@ static int disk_read(void) {
 #endif
   } /* while (fgets (buffer, sizeof (buffer), fh) != NULL) */
 
+  local_poll_count++;
+  for (ds = disklist, pre_ds = disklist; ds != NULL;) {
+      /* test if we have seen the disk in diskstats */
+      if (ds->poll_count != local_poll_count) {
+        diskstats_t *old_ds = ds;
+        /* free the ds */
+        if (pre_ds == disklist) {
+            /* first element */
+            disklist = ds->next;
+            ds = disklist;
+            pre_ds = ds;
+        } else {
+            pre_ds->next = ds->next;
+            ds = ds->next;
+        }
+        free(old_ds);
+      } else {
+        pre_ds = ds;
+        ds = ds->next;
+      }
+  }
   fclose(fh);
 /* #endif defined(KERNEL_LINUX) */
 
--- a/src/disk.c
+++ b/src/disk.c
@@ -820,7 +820,7 @@ static int disk_read(void) {
       ds->poll_count = poll_count;
       continue;
     }
-    ds->poll_count++;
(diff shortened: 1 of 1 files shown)


src/disk.c
```diff
@@ -815,22 +814,21 @@ static int disk_read(void) {
...
-      DEBUG("disk plugin: (ds->poll_count = %i) <= "
-            "(min_poll_count = 2); => Not writing.",
-            ds->poll_count);
+    /* Skip first cycle for newly-added disk */
+    if (ds->poll_count == 0) {
+      DEBUG("disk plugin: (ds->poll_count = 0) => Skipping.");
+      ds->poll_count = poll_count;
       continue;
     }
+    ds->poll_count++;
```
octo: I think using

```c
ds->poll_count = poll_count;
```

would be more robust.

src/disk.c
```diff
@@ -875,6 +873,30 @@ static int disk_read(void) {
...
+    /* Disk exists */
+    if (ds->poll_count == poll_count) {
+      pre_ds = ds;
+      ds = ds->next;
+      continue;
+    }
+
+    /* Disk is missing, remove it */
+    diskstats_t *missing_ds = ds;
+    if (pre_ds == disklist) {
```
octo: This should check for `ds == disklist`.

src/disk.c
```diff
@@ -875,6 +873,30 @@ static int disk_read(void) {
 #endif
   } /* while (fgets (buffer, sizeof (buffer), fh) != NULL) */
 
+  /* Remove disks that have disappeared from diskstats */
```
octo: Are newly discovered disks also added to diskstats?

octo:
Thanks @rpv-tomsk! Looks good overall, but there's a small bug when the second element in the linked list goes missing.

src/disk.c
rpv-tomsk: Yes, in L712: https://github.com/collectd/collectd/pull/2551/files#diff-352f7d19b14cf77c9be0554e611634faR712

src/disk.c
rpv-tomsk: Ouch. )

src/disk.c
octo: Thanks (I did the review from my phone which makes reading code hard ;)

src/disk.c
```diff
@@ -875,6 +873,30 @@ static int disk_read(void) {
...
+      ds = ds->next;
+      continue;
+    }
+
+    /* Disk is missing, remove it */
+    diskstats_t *missing_ds = ds;
+    if (ds == disklist) {
+      disklist = ds->next;
+      ds = disklist;
+      pre_ds = ds;
```
octo: Optional: I find this use of `ds` a bit confusing. Suggestion:

```c
disklist = pre_ds = ds->next;
ds = disklist;
```

rpv-tomsk: I stopped at the following code:
```c
    /* Disk is missing, remove it */
    diskstats_t *missing_ds = ds;
    if (ds == disklist) {
      pre_ds = disklist = ds->next;
    } else {
      pre_ds->next = ds->next;
    }
    ds = ds->next;

```

rpv-tomsk:
@klnikita , Thanks for reporting and working on this issue!

PR #2556, opened 2017-11-18
rrd plugin: Fix writing of long values.

mlichvar:
The buffer provided for value_list_to_string() was not long enough to
fit negative values using exponent and maximum precision of the gauge
format.

It seems this was a regression from commit 291879be6c66d00ff8dad09a0e44b332db31459a.

### DIFF
--- a/src/rrdtool.c
+++ b/src/rrdtool.c
@@ -794,7 +794,7 @@ static int rrd_write(const data_set_t *ds, const value_list_t *vl,
   if (value_list_to_filename(filename, sizeof(filename), vl) != 0)
     return -1;
 
-  char values[32 * ds->ds_num];
+  char values[32 * (ds->ds_num + 1)];
   if (value_list_to_string(values, sizeof(values), ds, vl) != 0)
     return -1;
 


octo:
Hi @mlichvar, thanks for your patch!

I'm a bit hesitant to replace one arbitrary limit with another equally arbitrary limit. We usually print gauges with a precision of 15 decimals, so 32 charts *should* be plenty. If they're not, I'd rather use exponential notation than increase the buffer size.

mlichvar:
In my case, 32 was not enough to fit values like this:
1511000588:-3.28999999510415e-06

That's 32 chars and one extra is needed for NUL. It took me a while to figure out why my graphs looked weird and why only some values were accepted, so I thought there should be a bit of extra room in case someone changes the precision of the gauge format, etc. Would you prefer if I changed it from 50 to 33?

octo:
Thanks for the concrete example, that's very helpful!

I think the problem is that the buffer doesn't take the timestamp into account. I suggest to fix this like so:

```c
char values[32 * (ds->ds_num + 1)];
//                           ^^^
```

mlichvar:
That makes sense, thanks for the suggestion. I've updated the patch.

octo:
Hi @mlichvar!

Thanks for the update! I rebase this on the *collectd-5.8* branch. I usually then push to the contributor's branch. I didn't do it here, because I didn't want mess up your *master* branch. Instead, I created #2566 which will get merged shortly.

Best regards,
—octo

PR #2567, opened 2017-11-21
Df metadata for plugin instance

vtols:
This pull request adds original plugin instance to value list meta of `df` plugin. This is done because after replacing `/` with `-` (and also `root`) reconstructing original plugin instance is ambiguous. With this change it would be possible to fetch the original name from metadata

### DIFF
--- a/src/df.c
+++ b/src/df.c
@@ -161,7 +161,7 @@ __attribute__((nonnull(2))) static int df_submit_one(char *plugin_instance,
   vl.meta = NULL;
 
   return 0;
-} /* void df_submit_one */
+} /* int df_submit_one */
 
 static int df_read(void) {
 #if HAVE_STATVFS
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Such a solution will not work when 'hub' topology (with Collectd-agents which send metrics to a central Collectd-hub) is used. Metadata is not sent over network.

vtols:
@rpv-tomsk, yes we know about the problem. At this moment we have these options:

1. Discard the changes, original path metadata won't be added.
2. Leave things as is, metadata is available only for local plugins, and is not sent over network
3. Make changes allowing network plugin to send metadata over network.

Have you any suggestions?

faxm0dem:
Hi,
It's unfortunate this got closed.
I believe it was a good idea, and as a user of write_riemann, I would definitely have use of this

PR #2569, opened 2017-11-22
add sql writer example for timescaledb

larsks:
Timescaledb is an extension for postgresql that adds optimizations for
the storage of time series data.  This means that the manual time
partitioning in the stock postgresql example is unnecessary.


larsks:
Well, maybe some other time.

PR #2582, opened 2017-11-28
Perl plugin: Remove the defunct c_ithread_destructor() callback.

octo:
This callback was only implicitly called by `pthread_key_delete()` from `perl_shutdown()`. Importantly, this happens *after*:

  * All threads referenced by the linked list in `perl_threads` have been destroyed.
  * The `perl_threads->mutex` has been destroyed.

The callback therefore cannot work and likely wasn't called in the past. I think it's save to remove.

### DIFF
--- a/src/perl.c
+++ b/src/perl.c
@@ -1226,31 +1226,6 @@ static void c_ithread_destroy(c_ithread_t *ithread) {
   return;
 } /* static void c_ithread_destroy (c_ithread_t *) */
 
-static void c_ithread_destructor(void *arg) {
-  c_ithread_t *ithread = (c_ithread_t *)arg;
-  c_ithread_t *t = NULL;
-
-  if (NULL == perl_threads)
-    return;
-
-  pthread_mutex_lock(&perl_threads->mutex);
-
-  for (t = perl_threads->head; NULL != t; t = t->next)
-    if (t == ithread)
-      break;
-
-  /* the ithread no longer exists */
-  if (NULL == t) {
-    pthread_mutex_unlock(&perl_threads->mutex);
-    return;
-  }
-
-  c_ithread_destroy(ithread);
-
-  pthread_mutex_unlock(&perl_threads->mutex);
-  return;
-} /* static void c_ithread_destructor (void *) */
-
 /* must be called with perl_threads->mutex locked */
 static c_ithread_t *c_ithread_create(PerlInterpreter *base) {
   c_ithread_t *t = NULL;
@@ -2420,7 +2395,7 @@ static int init_pi(int argc, char **argv) {
   }
 #endif /* COLLECT_DEBUG */
 
-  if (0 != pthread_key_create(&perl_thr_key, c_ithread_destructor)) {
+  if (0 != pthread_key_create(&perl_thr_key, NULL)) {
     log_err("init_pi: pthread_key_create failed");
 
     /* this must not happen - cowardly giving up if it does */


rpv-tomsk:
What is about threads (interpreters), attached to threads of other plugins?

octo:
`pthread_setspecific(perl_thr_key)` is only called from `c_ithread_create()`, which also adds threads to the `perl_threads` linked list that's freed in `perl_shutdown()`. I think all threads with the key are cleaned up when the key is destroyed.

rpv-tomsk:
When some plugin will create and stop threads at will (which we don't prohibit to do), that will be a kind of  memleak in perl plugin - it would save all these interpreter instances until Collectd shutdown.

I vote against this 'cleanup'. Please, don't repair things which are not broken.

rpv-tomsk:
>I think all threads with the key are cleaned up when the key is destroyed.

Maybe we have differend understanding of some things. Let's clarify:

```
PTHREAD_KEY_CREATE(P)                                                                             POSIX Programmer's Manual                                                                            PTHREAD_KEY_CREATE(P)

NAME
       pthread_key_create - thread-specific data key creation

SYNOPSIS
       #include <pthread.h>

       int pthread_key_create(pthread_key_t *key, void (*destructor)(void*));

An optional destructor function may be associated with each key value.  At thread exit, if a key value has a non-NULL destructor pointer, and the thread has a non-NULL value associated with that key, the value  of
       the  key  is set to NULL, and then the function pointed to is called with the previously associated value as its sole argument.
```

octo:
I see what you're saying: the *perl plugin* creates a lot of short-lived threads and they rely on `c_ithread_destructor` to clean up after themselves. I agree with that and the `c_ithread_destructor()` function needs to keep existing.

I still think the `perl_shutdown()` function has severe flaws and should be rewritten. Here are the problems I see by following two threads, represented by the two levels of indentation:

  * `perl_shutdown()` is called
      * An ephemeral thread is created
  * **locks** `perl_threads->mutex`
  * Sets `t->shutdown`
      * Ephemeral thread exits.
      * The *pthread* library calls `c_ithread_destructor(t)`.
      * `c_ithread_destructor`: Blocks on `perl_threads->mutex`
  * Calls `c_ithread_destroy(t)`
  * Loop continues until the entire linked list of `c_ithread_t*` are freed. Both `perl_threads->head` and `perl_threads->tail` are now `NULL`.

Now there's a race. This can go a number of ways; here are one good and two bad ways this can go:

The **good** case. Since the *perl plugin* does not routinely crash, this must be a common case:

  * **unlocks** `perl_threads->mutex`
      * **locks** `perl_threads->mutex`
      * "Iterates" over `perl_threads->head` (now `NULL`) to find `ithread` (now an invalid pointer); finding nothing
      * **unlocks** `perl_threads->mutex` and returns.
  * **destroys** `perl_threads->mutex`
  * **frees** `perl_threads`

Here is **bad** case number one:

  * **unlocks** `perl_threads->mutex`
      * **locks** `perl_threads->mutex`
  * **destroys** `perl_threads->mutex`
      * "Iterates" over `perl_threads->head` (now `NULL`) to find `ithread` (now an invalid pointer); finding nothing
      * Tries to unlock `perl_threads->mutex` → **crash!**

Here is **bad** case number two:

  * **unlocks** `perl_threads->mutex`
      * **locks** `perl_threads->mutex`
  * **destroys** `perl_threads->mutex`
  * **frees** `perl_threads`
      * Tries to dereference `perl_threads` → **crash!**

rpv-tomsk:
> I see what you're saying: the perl plugin creates a lot of short-lived threads and they rely on c_ithread_destructor to clean up after themselves. I agree with that and the c_ithread_destructor() function needs to keep existing.

Just a note: Not the perl plugin itself creates these threads. These threads can be created by a other plugin or Collectd core, and perl plugin may attach a perl interpreter to that thread if that will be needed by a callback call. 

I'm unsure how things will work if a perl-written plugin will start own thread inside of a perl code.

PR #2585, opened 2017-11-30
Initial version of notify_icinga_plugin

tnt4brain:
Hi, this is the very first implementation of "notify_icinga" plugin. Its main purpose is to push notifications to Icinga2 REST API (HTTPS is not supported yet).

Could you please review it?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1261,6 +1261,16 @@ notify_nagios_la_SOURCES = src/notify_nagios.c
 notify_nagios_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 endif
 
+if BUILD_PLUGIN_NOTIFY_ICINGA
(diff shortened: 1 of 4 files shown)


octo:
Hi @tnt4brain,

I'd be happy to review when you're ready. One piece of high level feedback: please consider if this could be added as an additional format to the *write_http plugin* instead.

Best regards,
—octo

tnt4brain:
Hi @octo,

I'm not really sure with all this, but I am afraid that fusing two plugins together could distort/hide the real purpose of this plugin - it is not a writer thing but merely a plugin to push notifications to Icinga. Their output formats differ too much anyway, and Icinga's one is fixed (so we can provide a reasonable defaults for it, except exact host/port values).

Sincerely,
tnt4brain

Issue #2583, opened 2017-11-28
make check fails without libgcrypt on CentOS 7


rubenk:
*   Version of collectd: HEAD (2b3e3f67e7c89772379a94987700e2a5346e36c4)
*   Operating system / distribution:
Centos 7.4

```
  CC       src/libcollectdclient/test_libcollectd_network_parse-network_parse_test.o
src/libcollectdclient/network_parse_test.c: In function ‘test_decrypt_aes256’:
src/libcollectdclient/network_parse_test.c:463:3: error: implicit declaration of function ‘decrypt_aes256’ [-Werror=implicit-function-declaration]
   int status = decrypt_aes256(
   ^
cc1: all warnings being treated as errors
make[2]: *** [src/libcollectdclient/test_libcollectd_network_parse-network_parse_test.o] Error 1
make[2]: Leaving directory `/home/ruben/src/collectd'
make[1]: *** [check-am] Error 2
make[1]: Leaving directory `/home/ruben/src/collectd'
make: *** [check] Error 2
```

rubenk:
Very strange that CI doesn't catch this by the way.

octo:
We're building on RHEL and Fedora, but not CentOS, presumably there's a difference ;-)

octo:
… or the test builders simply all have libgcrypt installed?

rubenk:
> … or the test builders simply all have libgcrypt installed?

ah yes, that's it!

PR #2586, opened 2017-11-30
network_parse_test.c: Only test security features if libgcrypt is available.

octo:
Fixes: #2583


rubenk:
Great, thanks for fixing this @octo!

PR #2590, opened 2017-12-01
Fix collectd ovs deadlock

### LINKED ISSUE #2574: collectd -T doesn't return with ovs_events and ovs_stats plugin enabled
*   Version of collectd: 5.8
*   Operating system / distribution: RHEL 7.4

I installed the ovs_events and ovs_stats plugin. While tweaking config, I realized, collectd -T doesn't return any more, and you have to kill -9. It happens with any of these plugins enabled.

For reference, I have a repository with collectd 5.8 builds on RHEL 7.4 and debug enabled here: http://www.matthias-runge.de/fedora/collectd/

vmytnykx:
During running collectd test commands (`collectd -T`) the deadlock issue has been discovered. This PR resolces this deadlock. During investigation, a few deadlock issue have been detected and fixed.

- **Dead-lock on destroyed mutex**
  The issue happens when EVENT thread is already destroyed (event mutex) during ovs_db_destroy() 
  function but POLL thread just connected to the socket and trying to send the 
  OVS_DB_EVENT_CONN_ESTABLISHED message to EVENT thread via ovs_db_event_post() function which 
  requires the EVENT thread mutex which has been destroyed already.

- **Internal dead-lock on pdb->mutex & pdb->event_thread.mutex**
- **OvS stats dead-lock on g_stats_lock**
- **Event can be skipped (low reproduce rate ~5%)**
  There is a race condition when ovs_db_event_post() is called at the same time (e.g. during the 
  ovs_db_destroy() call) when the EVENT thread is awake up because of the timeout. In this case the 
  EVENT thread will skip the event because of the timeout return value instead of 0 (success). 

fixes #2574

### DIFF
--- a/src/ovs_stats.c
+++ b/src/ovs_stats.c
@@ -974,9 +974,9 @@ static int ovs_stats_plugin_read(__attribute__((unused)) user_data_t *ud) {
 
 /* Shutdown OvS Stats plugin */
 static int ovs_stats_plugin_shutdown(void) {
-  pthread_mutex_lock(&g_stats_lock);
   DEBUG("OvS Statistics plugin shutting down");
   ovs_db_destroy(g_ovs_db);
+  pthread_mutex_lock(&g_stats_lock);
   ovs_stats_free_bridge_list(g_bridge_list_head);
   ovs_stats_free_bridge_list(g_monitored_bridge_list_head);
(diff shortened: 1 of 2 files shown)


octo:
Thanks @vmytnykx and @maryamtahhan!

PR #2597, opened 2017-12-05
[WiP] Factors out libcollectd library.

SeanCampbell:
This is necessary for dynamically loading in plugins on Windows.


SeanCampbell:
@igorpeshansky Could you take a look?

SeanCampbell:
@octo Would you be able to review this PR? Thank you!

SeanCampbell:
@tokkee Are you able to take a look?

Makefile.am
```diff
@@ -235,6 +233,16 @@ collectd_LDADD = \
 	$(COMMON_LIBS) \
 	$(DLOPEN_LIBS)
 
+collectd_SOURCES = \
+	src/daemon/collectd.c \
+	src/daemon/collectd.h
+collectd_CFLAGS = $(AM_CFLAGS)
+collectd_CPPFLAGS = $(AM_CPPFLAGS)
+collectd_LDFLAGS = -export-dynamic
```
tokkee: I think this isn't needed anymore with this patch.

Makefile.am
```diff
@@ -221,12 +221,10 @@ collectd_SOURCES = \
...
 	src/daemon/utils_threshold.h
-
-
-collectd_CFLAGS = $(AM_CFLAGS)
-collectd_CPPFLAGS = $(AM_CPPFLAGS)
-collectd_LDFLAGS = -export-dynamic
-collectd_LDADD = \
+libcollectd_la_CPPFLAGS = $(AM_CPPFLAGS) $(LTDLINCL)
+libcollectd_la_CFLAGS = $(AM_CFLAGS)
+libcollectd_la_LDFLAGS = $(COMMON_LDFLAGS)
```
tokkee: What is COMMON_LDFLAGS?

Makefile.am
```diff
@@ -96,7 +96,9 @@ pkginclude_HEADERS = \
 	src/libcollectdclient/collectd/server.h \
 	src/libcollectdclient/collectd/types.h
 
-lib_LTLIBRARIES = libcollectdclient.la
+lib_LTLIBRARIES = \
+	libcollectd.la \
```
tokkee: So, this adds a new library installed into /usr/lib/ (or whatever it is on the target system) which means we'll have to start doing proper versioning, etc. I'd like @octo to comment on that as well.

Makefile.am
SeanCampbell: That is only set on Windows, and I forgot to remove it when separating out these changes. Removed it now.

Makefile.am
SeanCampbell: Done.

octo:
Hi @SeanCampbell and @tokkee,

thanks for this PR. I'm a bit hesitant because putting 99% of the daemon into a library makes a lot of previously private symbols public, with everything that entails. Could we maybe tweak the build system to only create that library on Windows?

Best regards,
—octo

SeanCampbell:
I'm not too familiar with GNU make tools, but I think putting libcollectd.la in noinst_LTLIBRARIES for non-Windows platforms should do it. Is that right?

Added another commit with this change.

Thanks!
-Sean

Makefile.am
```diff
@@ -123,6 +125,9 @@ noinst_LTLIBRARIES = \
 	libmetadata.la \
 	libmount.la \
 	liboconfig.la
+if !BUILD_WIN32
```
octo: Could you please move this up so the two options are closer together:

```
if BUILD_WIN32
lib_LTLIBRARIES …
else
noinst_LTLIBRARIES …
endif
```

configure.ac
```diff
@@ -7072,3 +7077,4 @@ if test "x$dependency_warning" = "xyes"; then
 fi
 
 # vim: set fdm=marker sw=2 sts=2 ts=2 et :
+
```
octo: Do you think you could filter out this newline?

SeanCampbell: When I get rid of this, the build fails on cygwin with:
/usr/bin/m4:configure.ac: ERROR: end of file in comment

Is it ok to leave it? As far as I can tell, adding the line doesn't affect builds on Linux.

Makefile.am
SeanCampbell: Done.

SeanCampbell:
@rubenk, anything you think I should change on this? Thanks!

-Sean

rubenk:
Sorry for the late response.

I'm not familiar with how linking works on Windows, so could you expand a bit on why this is actually necessary? We already link the collectd binary with -export-dynamic which should make the necessary symbols available.

SeanCampbell:
@igorpeshansky Feel free to add anything I'm missing here.

As I understand it, it won't work on Windows because Windows imports must be resolved at library link-time, rather than at load time as in Linux. So without separating the plugin functionality into a separate DLL, building the plugins fails with "undefined reference" errors e.g. "undefined reference to `plugin_log'"

rubenk:
https://stackoverflow.com/questions/3752634/dll-get-symbols-from-its-parent-loader/3756083#3756083 seems to suggest that there's no need for a separate dll.

rubenk:
Ah, good that you got it to work! I'm a bit surprised that libtool can't figure out the correct linker incantation based on -export-dynamic (which is a libtool option, not an option for ld). After all that's what it's for.

> The related changes are adding flags to PLUGIN_LDFLAGS, adding COMMON_DEPS to plugins, and adding additional LD_FLAGS to the collectd binary. Is that preferable to breaking out the functionality into a separate DLL? In that case, I can close out this PR and follow up with a PR for the Windows port.

I looked through your diff quickly and the changes are not too intrusive, so yes, please follow up with a PR with these changes.

Thanks!

PR #2598, opened 2017-12-06
processes plugin: Add support for Linux Delay Accounting.

octo:
[Linux Delay Accounting](https://www.kernel.org/doc/Documentation/accounting/delay-accounting.txt) reports the time a task was delayed by

  * waiting for a CPU (while being runnable),
  * completion of synchronous block I/O initiated by the task,
  * swapping in pages,
  * memory reclaim.

This patch adds four metrics per configured process, one for each of the bullet points. Metrics are reported in percent rather than, for example, nanoseconds per second.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6891,8 +6891,9 @@ Disabled by default.
 If enabled, collect Linux Delay Accounding information for matching processes.
 Delay Accounting provides the time processes wait for the CPU to become
 available, for I/O operations to finish, for pages to be swapped in and for
-freed pages to be reclaimed. The metrics are reported as a percentage, e.g.
-C<percent-delay-cpu>. Disabled by default.
+freed pages to be reclaimed. The metrics are reported as "seconds per second"
+using the C<delay_rate> type, e.g. C<delay_rate-delay-cpu>.
+Disabled by default.
 
 This option is only available on Linux, requires the C<libmnl> library and
 requires the C<CAP_NET_ADMIN> capability at runtime.
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6895,7 +6895,7 @@ freed pages to be reclaimed. The metrics are reported as a percentage, e.g.
 C<percent-delay-cpu>. Disabled by default.
 
 This option is only available on Linux, requires the C<libmnl> library and
-requires root privileges at runtime.
+requires the C<CAP_NET_ADMIN> capability at runtime.
 
 =item B<CollectFileDescriptor> I<Boolean>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6842,22 +6842,25 @@ The statistics collected for matched processes are:
  - number of memory mapped files (under Linux)
  - io data (where available)
  - context switches (under Linux)
- - minor and major pagefaults.
+ - minor and major pagefaults
+ - Delay Accounting information (Linux only, requires libmnl)
 
 B<Synopsis:>
 
  <Plugin processes>
-   CollectFileDescriptor true
-   CollectContextSwitch true
+   CollectFileDescriptor  true
+   CollectContextSwitch   true
+   CollectDelayAccounting false
    Process "name"
    ProcessMatch "name" "regex"
    <Process "collectd">
-     CollectFileDescriptor false
-     CollectContextSwitch false
+     CollectFileDescriptor  false
+     CollectContextSwitch   false
+     CollectDelayAccounting true
    </Process>
    <ProcessMatch "name" "regex">
      CollectFileDescriptor false
-     CollectContextSwitch true
+     CollectContextSwitch  true
    </Process>
  </Plugin>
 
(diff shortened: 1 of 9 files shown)


dothebart:
it seems the STRERROR macro isn't used properly, will error out on centos7

octo:
Yeah, the `STRERROR` and `STRERRNO` macros are relatively new, see #2519.

src/processes.c
```diff
@@ -573,8 +649,15 @@ static void ps_tune_instance(oconfig_item_t *ci, procstat_t *ps) {
...
-      ERROR("processes plugin: Option `%s' not allowed here.", c->key);
+    else if (strcasecmp(c->key, "CollectDelayAccounting") == 0) {
+#if HAVE_LIBTASKSTATS
+      cf_util_get_boolean(c, &ps->report_delay);
+#else
+      WARNING("processes plugin: The plugin has been compiled without support "
+              "for the \"CollectDelayAccounting\" option.");
+#endif
+    } else {
+      ERROR("processes plugin: Option `%s' not allowed heeere.", c->key);
```
rpv-tomsk: Please fix a typo here ;-)

We are watch for changes! ;-)

src/collectd.conf.pod
```diff
@@ -6883,6 +6886,17 @@ I<name> must not contain slashes.
 Collect the number of context switches for matched processes.
 Disabled by default.
 
+=item B<CollectDelayAccounting> I<Boolean>
+
+If enabled, collect Linux Delay Accounding information for matching processes.
+Delay Accounting provides the time processes wait for the CPU to become
+available, for I/O operations to finish, for pages to be swapped in and for
+freed pages to be reclaimed. The metrics are reported as a percentage, e.g.
+C<percent-delay-cpu>. Disabled by default.
```
rpv-tomsk: Hi!
IMHO documented type instance does not match implementation - there will be no 'delay' word.

Can you please re-check this?

rpv-tomsk:
That is amazing feature!

That is much-much more useful than mine try to add metrics of processes/threads states (similar to 'ps_state' metrics reported for a system-wide process list, but for selected processes only and their threads).
For my changes, most of metrics reported 'sleeping' state for threads and processes, so I find my change not so useful as I expected when work was started.

rpv-tomsk:
Also, want to notice - for myself, in my systems I will report these metrics with a 'ps_delay' type, not a 'percent'.

Plugins may report different sets of metrics which units are percents, but related to different datasets ("delays", "cpu usage", other ratios). I dislike 'percent' as a type in a such cases.

octo:
> That is amazing feature!

Agreed. What blows my mind is that `<linux/taskstats.h>` has a copyright notice from 2006.

Please feel free to merge when you're happy – I don't have any pending changes on this.

Best regards,
—octo

src/collectd.conf.pod
octo: Thanks for noticing! This is a regression introduced in 17b81d4fa4eabbb7c029ccf424f43e4d89c27e35, fixed in 023790e3c1eb24fab52cafbcd90004312059787c.

octo: Thanks for noticing! This is a regression introduced in 17b81d4fa4eabbb7c029ccf424f43e4d89c27e35, fixed in 023790e3c1eb24fab52cafbcd90004312059787c.

rpv-tomsk:
cc @tokkee 

Hi!

I have a minor note for you, related to this change.
To use this update in my Debian package, I also updated `debian/rules` in the following way:

```
--- a/debian/rules
+++ b/debian/rules
@@ -363,6 +362,7 @@ binary-arch: build install-arch
                -dDepends debian/collectd-core/usr/lib/collectd/rrdtool.so
        dpkg-shlibdeps -Tdebian/collectd-core.substvars \
                -dDepends debian/collectd-core/usr/sbin/* \
+               -dDepends debian/collectd-core/usr/lib/collectd/processes.so \
                -dSuggests debian/collectd-core/usr/lib/collectd/*.so
        grep shlibs:Suggests debian/collectd-core.substvars \
                | sed -e 's/shlibs:Suggests/shlibs:Recommends/' \
```

I think `processes` plugin is used in 'all' setups, so we need to _depend_ on libmnl rather than just suggest/recommend it.

rpv-tomsk:
>Please feel free to merge when you're happy – I don't have any pending changes on this.

Ok, if we leave reported type as a `percent`, then I have no other remarks.
Thanks for this cool feature!

octo:
@rpv-tomsk Regarding the *percent* type: it has one serious shortcoming: its maximum value of (a little over) 100. A process with five threads can be blocked for five seconds every second, i.e. 500%.

There are two ways to fix this:

  * Remove the upper bound from the *percent* type.
  * Introduce a new type.

I suggest to introduce `delay_rate` and then report the metric as "seconds per second". What do you think?

rpv-tomsk:
>A process with five threads can be blocked for five seconds every second, i.e. 500%.

Of course, I like `delay_rate` proposal. But another thought about this - I think it will be better if values will be normalised. Maybe we can use `cpu_run_real_total` or `cpu_run_virtual_total` fields for this?

octo:
> Maybe we can use `cpu_run_real_total` or `cpu_run_virtual_total` fields for this?

But those fields report entirely different metrics …?

rpv-tomsk:
>But those fields report entirely different metrics …?

My thought was the following:

suppose the process awakened and it takes 50 units of time before it went to sleep again. In real world that was 100 units of time, so the process uses 50% of CPU.

During that awakened time it might to use 50 units of CPU, or it might to be delayed for a while.

For example, it can spent 5 units of time for IO waiting. So, then we report 10% as 'io delay'. 

Without a such normalisation we would report only 5% as 'io delay'.

I hoped what we can get "50 units" value from cpu_run_real_total field.
I'm missing something?

rpv-tomsk:
Yeah, I also see 140% as IO wait value . )

rpv-tomsk:
Each of these variangs has own advantages:
I think what values in CPU usage units (non-normalized) will be better to use at charts, and normalized is somewhat better for thresholds. As a final decision, I think non-normalized is enough. ;-)

octo:
Agreed, let's go with the wall clock time for now. We can figure with the metrics some more in the future.

PR #2599, opened 2017-12-07
df: fix memory leak in error case

takahashi-tsc:
Because the memory of the STATANYFS list is not released
when the read callback function error occurs,
a memory leak may occur.
With this change, the memory is always released
by not returning in the loop.

### DIFF
--- a/src/df.c
+++ b/src/df.c
@@ -152,6 +152,7 @@ static int df_read(void) {
 #elif HAVE_STATFS
   struct statfs statbuf;
 #endif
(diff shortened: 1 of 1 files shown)


octo:
Thanks @takahashi-tsc!

Cherry-picked into *collectd-5.8* as 29e3c6c5c3936b75f16811071e77904900edd86e

rpv-tomsk:
Hi!

I think this PR can be closed as it was merged, right?

PR #2603, opened 2017-12-07
Red Hat NFVPE Procevent Plugin

abays:
Plugin that monitors process starts and stops via netlink library


abays:
Cleaning and re-requesting

PR #2604, opened 2017-12-07
Red Hat NFVPE Sysevent Plugin

abays:
Plugin that monitors rsyslog messages and filters them (if desired)


abays:
Cleaning and re-requesting

PR #2606, opened 2017-12-07
Remove recv() call to make code more portable.

SeanCampbell:
Based on discussion in Stackdriver/collectd [windows-fixes branch](https://github.com/Stackdriver/collectd/commit/bee5e168175e3078b681585fa0fe3f7361d90055).

### DIFF
--- a/src/daemon/common.c
+++ b/src/daemon/common.c
@@ -262,10 +262,7 @@ int swrite(int fd, const void *buf, size_t count) {
   pfd.events = POLLIN | POLLHUP;
   pfd.revents = 0;
   if (poll(&pfd, 1, 0) > 0) {
-    char buffer[32];
-    if (recv(fd, buffer, sizeof(buffer), MSG_PEEK | MSG_DONTWAIT) == 0) {
-      /* if recv returns zero (even though poll() said there is data to be
-       * read), that means the connection has been closed */
+    if (pfd.revents & POLLHUP) {
       errno = ECONNRESET;
       return -1;
     }


SeanCampbell:
@igorpeshansky To keep you in the loop

src/daemon/common.c
```diff
@@ -262,10 +262,7 @@ int swrite(int fd, const void *buf, size_t count) {
   pfd.events = POLLIN | POLLHUP;
   pfd.revents = 0;
   if (poll(&pfd, 1, 0) > 0) {
-    char buffer[32];
-    if (recv(fd, buffer, sizeof(buffer), MSG_PEEK | MSG_DONTWAIT) == 0) {
-      /* if recv returns zero (even though poll() said there is data to be
-       * read), that means the connection has been closed */
+    if (pfd.revents & POLLHUP) {
```
rpv-tomsk: Unfortunately, my check shows this change not equivalent to existing implementation.

rpv-tomsk:
Hi!

I made a test to check if your proposed change is safe.
That was done by adding some lines into `swrite()`:
```
--- a/src/daemon/common.c
+++ b/src/daemon/common.c
@@ -262,10 +262,15 @@ int swrite(int fd, const void *buf, size_t count) {
   pfd.events = POLLIN | POLLHUP;
   pfd.revents = 0;
   if (poll(&pfd, 1, 0) > 0) {
+    DEBUG("Received Poll Event from fd %i. ", fd);
+    if (pfd.revents & POLLHUP) {
+      DEBUG("Received POLLHUP from fd %i. ", fd);
+    }
     char buffer[32];
     if (recv(fd, buffer, sizeof(buffer), MSG_PEEK | MSG_DONTWAIT) == 0) {
       /* if recv returns zero (even though poll() said there is data to be
        * read), that means the connection has been closed */
+      DEBUG("Received recv() == 0 from fd %i. ", fd);
       errno = ECONNRESET;
       return -1;
     }
```
I got the following debug output:
```
[2017-12-11 01:26:55] Received Poll Event from fd 3.
[2017-12-11 01:26:55] Received recv() == 0 from fd 3.
[2017-12-11 01:26:55] memcached plugin: Instance "local": write(2) failed: Connection reset by peer
[2017-12-11 01:26:55] read-function of plugin `memcached/local' failed. Will suspend it for 20.000 seconds.
```

So, as you can see, expected line `Received POLLHUP from fd 3.` is missing.
I catched a deadloop

SeanCampbell:
Thanks, Pavel. I'm working on getting this to build on Windows with Cygwin. Since this doesn't work, I'll see if there's a working portable way, or perhaps go with a Windows-specific version.

src/daemon/common.c
SeanCampbell: I'm having trouble recreating this issue. Running the write_graphite plugin, I actually had the opposite problem where the pfd.revents & POLLHUP block is executing and not the recv(...) == 0 block, but I think I have a fix for that.

With the memcached plugin enabled for me, neither seems to be called.

What is your configuration?

rpv-tomsk: >Running the write_graphite plugin, I actually had the opposite problem where the pfd.revents & POLLHUP block is executing and not the recv(...) == 0 block, but I think I have a fix for that.

That is caused by a fact what `write_graphite` writes mostly, and `memcached` mostly reads.
As I say before,
>It [POLLHUP] appears there only after write() attempt, so I see no way to code changes.
That may be possible what for `write_graphite`, you found `POLLHUP` set after previous write attempt has failed.

```
ReadThreads  1
WriteThreads 1
MaxReadInterval 20

LoadPlugin memcached
<Plugin memcached>
       <Instance "local">
               Address "127.0.0.1"
               Port "11211"
               #Socket "/var/run/memcached/memcached.sock"
       </Instance>
</Plugin>

+network plugin as write plugin
+logfile plugin to get console output 
+configure --enable-debug of course
```

I used a memcached instance or 'nc -l 127.0.0.1 -p 11211' for debug.
Also I use a `strace` for debugging: 
Look for main Collectd pid, then do 'cd /proc/PID/tasks', 'ls -l'. 
Last one (with max value in my case) will be a PID of read thread.

Issue #2587, opened 2017-12-01
tail plugin: wrong metric name reported to graphite for latency distribution


PhantomPhreak:
*   Version of collectd: 5.8.0.34.g3d082a9-1~jessie
*   Operating system / distribution: Debian 8.5

## Expected behavior
Metrics reported to graphite with following name (accordingly to [documentation](https://github.com/collectd/collectd/blob/master/src/collectd.conf.pod#plugin-tail))
`[<Instance>-]<Percent>`
In my case, metric reported to graphite should be like `something.instance-percent`
## Actual behavior
Metrics reported in format
`[<Instance>-]<Percent>.00`
In my case, metric reported to graphite looks like `something.instance-percent.00`

```
# ls -alh
drwxr-xr-x  2 _graphite _graphite 4.0K Nov 30 19:36 req_200-15
-rw-r--r--  1 _graphite _graphite  13M Nov 30 19:40 req_200-15.wsp

# ls -alh req_200-15
-rw-r--r--  1 _graphite _graphite  13M Dec  1 17:38 00.wsp

```

I guess this problem may be related to #2442

rpv-tomsk:
It looks like percent value is reported in form of `15.00`. 
Can you please show tail plugin configuration?

rpv-tomsk:
@rubenk , can you please fix 'snprintf' format so both GCC 7 will be happy and backward compatibility will be retained?

We need to revert change of `src/utils_tail_match.c` by f46d76e976ed76d2d8fa297918ecd308680d7ae9 commit and do the fix in a different way. Can you please do this?

rubenk:
@rpv-tomsk I'm sorry, but I don't understand what you want me to do?

rpv-tomsk:
Ruben, have you a GCC 7 to check code changes?

Commit f46d76e976ed76d2d8fa297918ecd308680d7ae9  has change:
```
-      snprintf(vl.type_instance, sizeof(vl.type_instance), "%s-%.0f",
+      snprintf(vl.type_instance, sizeof(vl.type_instance), "%.117s-%.2f",
```

That changes format of type instance to fix 'gcc 7 format-truncation errors'.
Unfortunately that changes format of a number, now it adds two digits after the radix character.
Example of  format change result:
```
# perl -e 'print sprintf("%.0f\n", 2);'
2
# perl -e 'print sprintf("%.2f\n", 2);'
2.00
```

[...]

PR #2611, opened 2017-12-10
tail plugin, latency: Fix regressions after #2535

rpv-tomsk:
As reported in #2587, the f46d76e976ed76d2d8fa297918ecd308680d7ae9 causes regression.

1) It adds two fixed decimal places into type instance of reported percentiles.

Before: 

```
# perl -e 'print sprintf("%.0f\n", 10.05);'
10
```
After:
```
# perl -e 'print sprintf("%.2f\n", 10.00);'
10.00
```

Also, the difference in format of percentile appeared after f46d76e976ed76d2d8fa297918ecd308680d7ae9 : the '%.0f' was in use when no type instance is configured, and '%.2f' - when type instance is set.

This was fixed by setting format to '%.4g' for all cases. 
I think it's enough to use '%.4g' to output percentile values which are in range from 0 to 100.

2) It reduces accuracy of bounds in reported buckets.

Before:
```
# perl -e 'print sprintf("%g\n", 10.05);'
10.05
```
After:
```
# perl -e 'print sprintf("%.2g\n", 100);'
1e+02
# perl -e 'print sprintf("%.2g\n", 10.05);'
10
```

This was fixed by reverting format back to '%g' as it was before f46d76e976ed76d2d8fa297918ecd308680d7ae9. 

Summary:
- Removed two fixed decimal places from type instance of percentiles
- Recovered accuracy of bounds in type instance of buckets
+ Allowed (compared to state before #2535) to report percentiles with fractional values
* Changed maximum length of reported type and type instance to 50 chars in all latency metrics

References: #2535, #2587

Have anybody any objections against merging this?

CC @taraschornyi, @mfournier, @PhantomPhreak

### DIFF
--- a/src/utils_tail_match.c
+++ b/src/utils_tail_match.c
@@ -118,10 +118,10 @@ static int latency_submit_match(cu_match_t *match, void *user_data) {
   sstrncpy(vl.type, data->type, sizeof(vl.type));
   for (size_t i = 0; i < data->latency_config.percentile_num; i++) {
     if (strlen(data->type_instance) != 0)
-      snprintf(vl.type_instance, sizeof(vl.type_instance), "%.117s-%.2f",
+      snprintf(vl.type_instance, sizeof(vl.type_instance), "%.50s-%.5g",
                data->type_instance, data->latency_config.percentile[i]);
     else
-      snprintf(vl.type_instance, sizeof(vl.type_instance), "%.0f",
+      snprintf(vl.type_instance, sizeof(vl.type_instance), "%.5g",
                data->latency_config.percentile[i]);
 
     vl.values = &(value_t){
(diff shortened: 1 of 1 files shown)


src/utils_tail_match.c
```diff
@@ -118,10 +118,10 @@ static int latency_submit_match(cu_match_t *match, void *user_data) {
   sstrncpy(vl.type, data->type, sizeof(vl.type));
   for (size_t i = 0; i < data->latency_config.percentile_num; i++) {
     if (strlen(data->type_instance) != 0)
-      snprintf(vl.type_instance, sizeof(vl.type_instance), "%.117s-%.2f",
+      snprintf(vl.type_instance, sizeof(vl.type_instance), "%.50s-%.4g",
```
octo: This format was originally `"%s-%.0f"`. Can we go back to using `%.0f` again?

rpv-tomsk: As stated before, that is done to allow to report percentiles with fractional values.
```
# perl -e 'use POSIX; printf("%.0f\n", 99.5);'
100
# perl -e 'use POSIX; printf("%.4g\n", 99.5);'
99.5
# perl -e 'use POSIX; printf("%.4g\n", 99);'
99
```
Configuration allows to set fractional percentile values, so we need to allow report them correctly.
If you want, we can split this change - set `%.0f` in 5.8 and `%.4g` in master.

octo: The docs don't say that fractional percentiles are forbidden, so it makes sense to support them. Put the `%g` variant into 5.8.

Maybe we can increase the precision to `%.5g` though – somebody is going to want to track five nines … ;)

PR #2618, opened 2017-12-12
Write amqp1 plugin

ajssmith:
This write plugin allows data to be sent to an AMQP 1.0 intermediary such as the Apache Qpid Dispatch Router (e.g. direct messaging) or Apache Artemis Broker (e.g. queue backed messaging).

It is similar to the existing amqp plugin which is for AMQP version 0.9.1 (e.g. rabbitmq server).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -542,6 +542,18 @@ amqp_la_LIBADD = \
 	libformat_json.la
 endif
 
+if BUILD_PLUGIN_AMQP1
+pkglib_LTLIBRARIES += amqp1.la
+amqp1_la_SOURCES = src/amqp1.c
+amqp1_la_CPPFLAGS = $(AM_CPPFLAGS) $(BUILD_WITH_LIBQPIDPROTON_CPPFLAGS)
+amqp1_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(BUILD_WITH_LIBQPIDPROTON_LDFLAGS)
+amqp1_la_LIBADD = \
+	$(BUILD_WITH_LIBQPIDPROTON_LIBS) \
+	libcmds.la \
+	libformat_graphite.la \
+	libformat_json.la
+endif
+
 if BUILD_PLUGIN_APACHE
 pkglib_LTLIBRARIES += apache.la
 apache_la_SOURCES = src/apache.c
--- a/Makefile.am
+++ b/Makefile.am
@@ -544,7 +544,9 @@ endif
 
 if BUILD_PLUGIN_AMQP1
 pkglib_LTLIBRARIES += amqp1.la
-amqp1_la_SOURCES = src/amqp1.c
+amqp1_la_SOURCES = \
+	src/amqp1.c \
+	src/utils_deq.h
 amqp1_la_CPPFLAGS = $(AM_CPPFLAGS) $(BUILD_WITH_LIBQPIDPROTON_CPPFLAGS)
 amqp1_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(BUILD_WITH_LIBQPIDPROTON_LDFLAGS)
 amqp1_la_LIBADD = \
--- a/configure.ac
+++ b/configure.ac
@@ -4713,7 +4713,16 @@ if test "x$with_libqpid_proton" = "xyes"; then
   CPPFLAGS="$SAVE_CPPFLAGS"
 fi
 
-# add library checks for function calls
+if test "x$with_libqpid_proton" = "xyes"; then
+  SAVE_LDFLAGS="$LDFLAGS"
+  LDFLAGS="$LDFLAGS $with_libqpid_proton_ldflags"
+
+  AC_CHECK_LIB([qpid-proton], [pn_connection],
+    [with_libqpid_proton="yes"],
+    [with_libqpid_proton="no (Symbol 'pn_connection' not found)"])
+
+  LDFLAGS="$SAVE_LDFLAGS"
+fi
 
 if test "x$with_libqpid_proton" = "xyes"; then
   BUILD_WITH_LIBQPIDPROTON_CPPFLAGS="$with_libqpid_proton_cppflags"
--- a/configure.ac
+++ b/configure.ac
@@ -4686,6 +4686,47 @@ if test "$with_libpython" != "xno"; then
 fi
 # }}} --with-libpython
 
+# --with-libqpid_proton {{{
+AC_ARG_WITH([libqpid_proton],
+  [AS_HELP_STRING([--with-libqpid_proton@<:@=PREFIX@:>@], [Path to libqpid_proton.])],
+  [
+    if test "x$withval" != "xno" && test "x$withval" != "xyes"; then
+      with_libqpid_proton_cppflags="-I$withval/include"
+      with_libqpid_proton_ldflags="-L$withval/lib"
+      with_libqpid_proton="yes"
+    else
+      with_libqpid_proton="$withval"
+    fi
+  ],
+  [with_libqpid_proton="yes"]
+)
+
+if test "x$with_libqpid_proton" = "xyes"; then
+  SAVE_CPPFLAGS="$CPPFLAGS"
+  CPPFLAGS="$CPPFLAGS $with_libqpid_proton_cppflags"
+
+  AC_CHECK_HEADERS([proton/proactor.h],
+    [with_libqpid_proton="yes"],
+    [with_libqpid_proton="no (proton/proactor.h not found)"]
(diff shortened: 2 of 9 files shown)


leifmadsen:
Just realized that you also need to adjust the RPM spec file so that the module can be added to the RPM.

https://github.com/collectd/collectd/blob/master/contrib/redhat/collectd.spec

Makefile.am
```diff
@@ -542,6 +542,18 @@ amqp_la_LIBADD = \
 	libformat_json.la
 endif
 
+if BUILD_PLUGIN_AMQP1
+pkglib_LTLIBRARIES += amqp1.la
+amqp1_la_SOURCES = src/amqp1.c
```
leifmadsen: You also need to add `src/utils_deq.h` or else it won't be included in the tarball which is used for building RPMs.

```
amqp1_la_SOURCES = src/amqp1.c \
        src/utils_deq.h
```

bendoerr:
+1 for this! I've got [prior art](https://github.com/bendoerr/collectd-write-qpid-json) in the form of a python plugin, but this on the surface looks better.

configure.ac
```diff
@@ -4686,6 +4686,47 @@ if test "$with_libpython" != "xno"; then
...
+
+  AC_CHECK_HEADERS([proton/proactor.h],
+    [with_libqpid_proton="yes"],
+    [with_libqpid_proton="no (proton/proactor.h not found)"]
+  )
+
+  CPPFLAGS="$SAVE_CPPFLAGS"
+fi
+
+# add library checks for function calls
```
rubenk: would be nice to indeed check if you can link against the library.

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#define BUFSIZE 8192
+#define AMQP1_FORMAT_JSON 0
+#define AMQP1_FORMAT_COMMAND 1
+#define AMQP1_FORMAT_GRAPHITE 2
+
+typedef struct amqp1_config_transport_t {
```
rubenk: we use _s for the struct tagname, _t for the typedef as a convention

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+ * Globals
+ */
+pn_connection_t *conn = NULL;
+pn_link_t *sender = NULL;
+pn_proactor_t *proactor = NULL;
+pthread_mutex_t send_lock;
+cd_message_list_t out_messages;
+uint64_t cd_tag = 1;
+uint64_t acknowledged = 0;
+amqp1_config_transport_t *transport = NULL;
```
rubenk: can these be made static?

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+static bool stopping = false;
+static int event_thread_running = 0;
+static pthread_t event_thread_id;
+
+/*
+ * Functions
+ */
+static void cd_message_free(cd_message_t *cdm) {
+  if (cdm->mbuf.start) {
+    free((void *)cdm->mbuf.start);
```
rubenk: unnecessary cast

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+
+static bool stopping = false;
+static int event_thread_running = 0;
+static pthread_t event_thread_id;
+
+/*
+ * Functions
+ */
+static void cd_message_free(cd_message_t *cdm) {
+  if (cdm->mbuf.start) {
```
rubenk: free accepts NULL so no need for the check

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+  char *user;
+  char *password;
+  char *address;
+  int retry_delay;
+} amqp1_config_transport_t;
+
+typedef struct amqp1_config_instance_t {
+  DEQ_LINKS(struct amqp1_config_instance_t);
+  char *name;
+  _Bool notify;
```
rubenk: It's fine to use bool from <stdbool.h>

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+  if ((n == NULL) || (user_data == NULL))
+    return EINVAL;
+
+  instance = user_data->data;
+
+  if (instance->notify != true) {
+    ERROR("amqp1 plugin: write notification failed");
+  }
+
+  cdm = NEW(cd_message_t);
```
rubenk: I'd prefer it if you just use malloc / calloc.

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+    if (status != 0)
+      break;
+  }
+
+  if (status != 0) {
+    amqp1_config_instance_free(instance);
+    return status;
+  } else {
+    char tpname[1024];
+    int status;
```
rubenk: this shadows status from the outer scope

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+              child->key);
+    if (status != 0)
+      break;
+  }
+
+  if (status != 0) {
+    amqp1_config_instance_free(instance);
+    return status;
+  } else {
+    char tpname[1024];
```
rubenk: why 1024? we normally use DATA_MAX_NAME_LEN I believe.

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+  int status = 0;
+
+  transport = calloc(1, sizeof(*transport));
+  if (transport == NULL) {
+    ERROR("amqp1 plugin: calloc failed.");
+    return ENOMEM;
+  }
+
+  /* Initialize transport configuration {{{ */
+  transport->name = NULL;
```
rubenk: No need to set this, calloc has already zero-initialized it.

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+static int amqp1_config(oconfig_item_t *ci) /* {{{ */
+{
+
+  for (int i = 0; i < ci->children_num; i++) {
+    oconfig_item_t *child = ci->children + i;
+
+    if (strcasecmp("Transport", child->key) == 0)
+      amqp1_config_transport(child);
+    else
+      WARNING("amqp1 plugin: Ignoring unknown config iption \%s\".",
```
rubenk: s/iption/option/

src/amqp1.c
MariuszX: in whole code please migrate from _Bool to bool

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+      status = cf_util_get_flag(child, &instance->graphite_flags,
+                                GRAPHITE_PRESERVE_SEPARATOR);
+    else if (strcasecmp("GraphitePrefix", child->key) == 0)
+      status = cf_util_get_string(child, &instance->prefix);
+    else if (strcasecmp("GraphitePostfix", child->key) == 0)
+      status = cf_util_get_string(child, &instance->postfix);
+    else if (strcasecmp("GraphiteEscapeChar", child->key) == 0) {
+      char *tmp_buff = NULL;
+      status = cf_util_get_string(child, &tmp_buff);
+      if (strlen(tmp_buff) > 1)
```
MariuszX: Check status before. tmp_buff could still be NULL here on incorrect data in config file.

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+  amqp1_config_instance_t *instance;
+
+  instance = calloc(1, sizeof(*instance));
+  if (instance == NULL) {
+    ERROR("amqp1 plugin: calloc failed.");
+    return ENOMEM;
+  }
+
+  /* Initialize instance configuration {{{ */
+  instance->name = NULL;
```
MariuszX: No need to set this, calloc has already zero-initialized it.

src/amqp1.c
```diff
@@ -0,0 +1,730 @@
...
+  char *name;
+  char *host;
+  char *port;
+  char *user;
+  char *password;
+  char *address;
+  int retry_delay;
+} amqp1_config_transport_t;
+
+typedef struct amqp1_config_instance_t {
```
MariuszX: use _s for the struct tagname (same as for amqp1_config_transport_t  above)

ajssmith:
Thanks for taking the time to review and provide comments. I have made the updates to the plugin and look forward to any additional feedback and comments.

paramite:
+1

rubenk:
Thank you for your patience @ajssmith. Merged into master.

paramite:
Thanks a lot @rubenk !

ajssmith:
@rubenk Thanks!

PR #2627, opened 2017-12-15
turbostat plugin: Increase turbostat_read() resiliency when sched_getaffinity() fails

pllopis:
The turbostat plugin uses sched_getaffinity() to save and restore collectd's cpu affinity settings, since get_counters() calls issued will modify this affinity.
On some systems, sched_getaffinity() will fail due to inconsistency of the cpu set size between userspace and kernel. In these cases, turbostat_read() will detect the  unsuccessful call to sched_getaffinity() and fail with an error, preventing data collection.
However, restoring the affinity is not essential to data collection. In fact, the kernel's turbostat tool, upon which this plugin is based, uses a different strategy. It restores the affinity by using cpu_present_set (affinity to any/all cpus).
This patch falls back to the same strategy when sched_getaffinity() fails.
This way, systems where this is supported will keep working as before, while systems where it breaks will function the same way as turbostat(8) does.


pllopis:
I just updated the code to something different and IMO more elegant: the current version adds a new config option to control how the affinity is restored. Please see the last commit message for details.

PR #2628, opened 2017-12-15
kafka: replace deprecated rd_kafka_errno2err

vincentbernat:
rd_kafka_errno2err() is depecrated in recent versions of
librdkafka. We replace it by rd_kafka_lasterror() which was introduced
in librdkafka 0.9.1 (available from Ubuntu Zesty and Debian Stretch).

Maybe too soon?


vincentbernat:
Notably, CI doesn't handle that. I'll update to push a patch to configure instead.

PR #2631, opened 2017-12-20
Graphite 1.1+ tag support

DanCech:
This PR adds support for sending metrics to Graphite using the tagging support added in the 1.1.1 release.

To enable tags, set `UseTags true` in the Carbon configuration block.

When enabled, the metric names generated will look like `test.single;host=example.com;plugin=test;plugin_instance=f.o.o;type=single;type_instance=b.a.r` and can be queried from Graphite like `seriesByTag("name=test.single", "plugin=test")`.

For more information on tag support in Graphite, see [the documentation](http://graphite.readthedocs.io/en/latest/tags.html).

If tag support is not enabled, there is no change to the metric names that are sent to Graphite.

When tag support is enabled `DropDuplicateFields`, `AlwaysAppendDS`, `Prefix` & `Postfix` configuration options all retain their meaning, though `Postfix` is applied to the end of the "name" part of the metric (before the first `;` that separates the name from the tags).  `SeparateInstances` and `PreserveSeparator` are not used when tag support is enabled.

### DIFF
--- a/src/utils_format_graphite.c
+++ b/src/utils_format_graphite.c
@@ -97,6 +97,86 @@ static void gr_copy_escape_part(char *dst, const char *src, size_t dst_len,
   }
 }
 
+static int gr_format_name_tagged(char *ret, int ret_len, value_list_t const *vl,
+                                 char const *ds_name, char const *prefix,
+                                 char const *postfix, char const escape_char,
+                                 unsigned int flags) {
+  char n_host[DATA_MAX_NAME_LEN];
+  char n_plugin[DATA_MAX_NAME_LEN];
+  char n_plugin_instance[DATA_MAX_NAME_LEN];
+  char n_type[DATA_MAX_NAME_LEN];
+  char n_type_instance[DATA_MAX_NAME_LEN];
+
+  char tmp_plugin[DATA_MAX_NAME_LEN + 8];
+  char tmp_plugin_instance[DATA_MAX_NAME_LEN + 17];
+  char tmp_type[DATA_MAX_NAME_LEN + 6];
+  char tmp_type_instance[DATA_MAX_NAME_LEN + 15];
+  char tmp_metric[3 * DATA_MAX_NAME_LEN + 2];
+  char tmp_ds_name[DATA_MAX_NAME_LEN + 9];
+
+  if (prefix == NULL)
+    prefix = "";
+
+  if (postfix == NULL)
+    postfix = "";
+
+  gr_copy_escape_part(n_host, vl->host, sizeof(n_host), escape_char, 1);
+  gr_copy_escape_part(n_plugin, vl->plugin, sizeof(n_plugin), escape_char, 1);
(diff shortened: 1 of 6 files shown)


src/utils_format_graphite.c
```diff
@@ -199,15 +279,26 @@ int format_graphite(char *buffer, size_t buffer_size, data_set_t const *ds,
...
-      return status;
+    if (flags & GRAPHITE_USE_TAGS) {
+      status = gr_format_name_tagged(key, sizeof(key), vl, ds_name, prefix,
+                                     postfix, escape_char, flags);
+      if (status != 0) {
+        ERROR("format_graphite: error with gr_format_name_tagged");
+        sfree(rates);
+        return status;
+      }
+    } else {
```
jdblack: This else is unnecessary, as  the  first half of the contitional returns.

jdblack:
This code does not appear to support writing custom tags; e.g.  adding tags for  DC, availability zone,  and such.

src/utils_format_graphite.c
DanCech: It only returns if `status != 0`

DanCech:
The existing `postfix` option can be used to add tags by specifying it like `;tag1=value1;tag2=value2`

The tag order isn't important in this scenario as they will be normalized (to alphabetical order by tag name) by carbon when the metrics are received.

DanCech:
Is there anything I can do to help move this forward?

AsenZahariev:
@jdblack is there any news around this PR ?

rpv-tomsk:
We also need documentation update (`collectd.conf.pod`, `collectd.conf.in`) for this to be merged.

> SeparateInstances and PreserveSeparator are not used when tag support is enabled.

Documentation update should reflect this too.

rpv-tomsk:
A question about this plugin, but not related to this PR update.

The goal of one possible improvement might be rewrite of `gr_format_name*` functions to fill `*ret` buffer directly, without intermediate variables, and with check of `snprintf` return value. But plugin does compare of escaped values `n_plugin`, `n_type` when `GRAPHITE_DROP_DUPE_FIELDS` flag is set. So, we still need to have these values in intermediate variables too.

What do you think, will it affect somebody's production systems if we replace compare to compare of unescaped `vl->plugin`, `vl->type` , `vl->plugin_instance`, `vl->type_instance`?

Thanks.

rpv-tomsk:
As series name has form `plugin.type`, do we really need to add these fields as tags?

rpv-tomsk:
This PR rebased against current master, documentation added and `write_kafka` plugin too.

Can anybody review last changes?

Thanks.

rpv-tomsk:
PR merged. Thanks for your work on this feature!

Hipska:
Could it be possible to update the documentation [Example data](https://collectd.org/wiki/index.php/Plugin:Write_Graphite#Example_data) with an example of data with tags?

PR #2649, opened 2018-01-11
[WIP] Knot Plugin

julianbrost:
This pull request implementes a new plugin that allows collecting metrics from the [Knot DNS server](https://www.knot-dns.cz/) from its UNIX domain socket using libknot.

It's still work in progress. The logic should be done but the handling of the individual metrics is not fully done. A few values are not handled at all (see TODO comment in the source), some like reply-size also don't really make sense to periodically collect in my opinion, so they probably won't be implemented at all (Knot keeps counters for packet sizes in 256 buckets, I think it makes more sense to just plot the total number as a histogram, but not in a collectd plugin I guess).

The second open question corresponds to types. For most of the metrics, there already exist types in types.db. Unfortunately, the dns_octets type doesn't fit the request-bytes/response-bytes as it has both rx and tx values as there is no direct 1:1 mapping between requests (query, update, other) and and reponses (reply, transfer, other). For now, I've just used some other types which kind of fit, but maybe this justifies adding a new type?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -166,6 +166,9 @@ Jérôme Renard <jerome.renard at gmail.com>
 Jiri Tyr <jiri.tyr at gmail.com>
  - fhcount plugin.
 
+Julian Brost <julian at 0x4a42.net>
+ - knot plugin.
+
 Julien Ammous <j.ammous at gmail.com>
  - Lua plugin.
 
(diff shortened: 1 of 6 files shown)


julianbrost:
I'm closing this as I probably won't put more work into this. If anyone wants to take over, feel free to use this as the base for your own PR.

PR #2652, opened 2018-01-16
iptables chain name pattern matching feature.

vijaykilledar:
in some organisation security team creates their own custom chains and jump to those chains from basic chains. currently iptables dont support chain pattern matching feature. Adding the same.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3476,7 +3476,7 @@ Select the iptables/ip6tables filter rules to count packets and bytes from.
 
 If only I<Table> and I<Chain> are given, this plugin will collect the counters
 of all rules which have a comment-match. The comment is then used as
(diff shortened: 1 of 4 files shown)


src/collectd.conf.pod
```diff
@@ -3485,6 +3485,21 @@ used as the type-instance.
 If I<Name> is supplied, it will be used as the type-instance instead of the
 comment or the number.
 
+  <Plugin iptables>
+    Chain "mangle" "/^custom_chain/"
+    Chain6 "magle" "custom_chain" "comment_name"
```
shastah: Typo? ("magle" instead of "mangle")

vijaykilledar: Yes its typo. I will raise another pull request for this change.

PR #2660, opened 2018-01-20
Adding support for CDAB endian 32-bit modbus polls

pbostley:
Modbus used big endian 16-bit registers.  With 32 bit and float it would follow that the 16bit registers that are combined to form a 32bit value would also be big endian.  Sadly some folks don't do it this way and we have a mixed endian world.

Instead of ABCD with register[n] having AB and register[n+1] having CD, you end up with CDAB where register[n] has the low order bytes (CD) and register[n+1} has the high order bytes.  

This is and issue with some Xetawave LLC and Freewave Technologies MODBUS IO devices, some versions of ABB Totalflow.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -4128,11 +4128,19 @@ Configures the base register to read from the device. If the option
 B<RegisterType> has been set to B<Uint32> or B<Float>, this and the next
 register will be read (the register number is increased by one).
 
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
Hi! 

Thanks for your work! 

Before this gets merged, we need to update plugin documentation (`src/collectd.conf.pod`) as well. 
Can you please do this?

Also, can you add some (short) info describing how was the need in this patch appeared (what environment/equipment you have)?  Your patch request description is very short )

Thanks!

Issue #2651, opened 2018-01-13
disk plugin does not use udev


connection-reset:
*   Version of collectd: 5.8.0
*   Operating system / distribution: Arch Linux

## Expected behavior
Use udev attributes in disk plugin
```
LoadPlugin disk
<Plugin disk>
    UdevNameAttr "ID_SERIAL"
</Plugin>
```

## Actual behavior
disk plugin does not use udev
`disk plugin: The "UdevNameAttr" option is only supported if collectd is built with libudev support`

## Steps to reproduce

collectd is compiled with libsystemd 236.0 which provides libudev
from configure:
```
checking libudev.h usability... yes
checking libudev.h presence... yes
checking for libudev.h... yes
checking for udev_new in -ludev... yes
```

It works fine if I remove all the `HAVE_UDEV_H` checks from disk.c

dstathis:
#2688

rubenk:
Should be fixed by ed696c1b3adcb42d55759efe6f8034337f0fd243

PR #2688, opened 2018-02-15
change HAVE_UDEV_H to HAVE_LIBUDEV_H for bug 2651

dstathis:
This fixes a bug which makes it so the disk plugin can't use UDEV support.

https://github.com/collectd/collectd/issues/2651

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -139,7 +139,7 @@ static int pnumdisk;
 #error "No applicable input method."
 #endif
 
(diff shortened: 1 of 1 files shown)


dstathis:
Could someone take a look at this? It's a simple bug fix and it would be great to get it in to an upcoming release.

rubenk:
Thanks @dstathis!

PR #2691, opened 2018-02-17
ping plugin: support specifying the address family

### LINKED ISSUE #2478: ping plugin: Can't ping hostnames with IPv6 addresses without IPv6 connectivity
*   Version of collectd: 5.7.2
*   Operating system / distribution: Arch Linux x86_64

## Expected behavior

ping aur.archlinux.org without issues like the command line ping tool does.

## Actual behavior

No coresponding data collected, log always says
```
ping plugin: host aur.archlinux.org has not answered 10 PING requests, triggering resolve
```

Because it gets an IPv6 address.

I manually add a ping_setopt call in the source code to select AF_INET and it works well.

lilydjwg:
This closes #2478.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6185,6 +6185,10 @@ long string is used so that the packet size of an ICMPv4 packet is exactly
 Sets the source address to use. I<host> may either be a numerical network
 address or a network hostname.
 
+=item B<AddressFamily> I<af>
+
+Sets the address family to use. I<af> may be "any", "ipv4" or "ipv6". This option will be ignored if you set a B<SourceAddress>.
+
 =item B<Device> I<name>
 
 Sets the outgoing network device to be used. I<name> has to specify an
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6187,7 +6187,8 @@ address or a network hostname.
 
 =item B<AddressFamily> I<af>
 
-Sets the address family to use. I<af> may be "any", "ipv4" or "ipv6". This option will be ignored if you set a B<SourceAddress>.
+Sets the address family to use. I<af> may be "any", "ipv4" or "ipv6". This
+option will be ignored if you set a B<SourceAddress>.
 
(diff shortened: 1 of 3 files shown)


lilydjwg:
Is there anyone that can take a look at this? What do you think about this way (adding an option) to solve the issue?

src/ping.c
```diff
@@ -468,6 +475,25 @@ static int ping_config(const char *key, const char *value) /* {{{ */
...
+      return status;
+
+    if (strncmp(af, "any", 3) == 0) {
+      ping_af = AF_UNSPEC;
+    } else if (strncmp(af, "ipv4", 4) == 0) {
+      ping_af = AF_INET;
+    } else if (strncmp(af, "ipv6", 4) == 0) {
+      ping_af = AF_INET6;
+    } else {
+      ERROR("ping plugin: Bad address family: %s", af);
```
rpv-tomsk: Rest of this function uses `WARNING` on 'invalid'/unsupported option values.
I think code should be consistent. What do you think about this?

lilydjwg: I overlooked that. Consistency is important, and I see that collectd is trying to continue as it can. Updating....

src/collectd.conf.pod
```diff
@@ -6185,6 +6185,10 @@ long string is used so that the packet size of an ICMPv4 packet is exactly
 Sets the source address to use. I<host> may either be a numerical network
 address or a network hostname.
 
+=item B<AddressFamily> I<af>
+
+Sets the address family to use. I<af> may be "any", "ipv4" or "ipv6". This option will be ignored if you set a B<SourceAddress>.
```
rpv-tomsk: The lines in this file must be shorter than 80 characters.

Please fix this, and this PR will be merged then.

Thanks.

lilydjwg: Oops, I was busy testing and didn't notice that Vim didn't wrap it automatically.

rpv-tomsk:
Thanks for your work!

PR #2694, opened 2018-02-22
Remove judgemental "small" from description

Maikel-Predikkta:
There is probably no agreed definition of "small daemon". From my perspective, it uses 52 MB of memory on my server which makes it the third biggest daemon there.

Removing "small" also makes the description more concise about the things collectd does.

### DIFF
--- a/README
+++ b/README
@@ -5,9 +5,8 @@ https://collectd.org/
 About
 -----
 
-  collectd is a small daemon which collects system information periodically
-  and provides mechanisms to store and monitor the values in a variety of
-  ways.
+  collectd is a daemon which collects system information periodically and
+  provides mechanisms to store and monitor the values in a variety of ways.
 
 
 Features


rpv-tomsk:
```
srv04:~# ps u 3215
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root      3215  0.2  0.0  46568  1316 ?        SLl  Jan26  87:48 collectd -C /etc/collectd/collectd.conf -f
```
So, it really uses only 1 MByte of RAM. ;-)

Maikel-Predikkta:
```
$ ps u 31736
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     31736  0.1  5.2 865960 52324 ?        Ssl  Feb22   7:40 /usr/sbin/collectd
```

That is with these plugins:
```
$ grep LoadPlugin wormly.conf
LoadPlugin syslog
LoadPlugin cpu
LoadPlugin df
LoadPlugin disk
LoadPlugin interface
LoadPlugin load
LoadPlugin memory
LoadPlugin uptime
LoadPlugin vmem
LoadPlugin aggregation
LoadPlugin write_http
LoadPlugin "match_regex"
```

I guess I need to blame the loaded plugins. When I remove the Wormly configuration, it really uses only 1 MB of RAM. Do you think the typical use case is with a few plugins like this? Don't they count? ;-)

Well, I guess it's all a matter of perspective and that's why I thought it could be good to remove that little word. Your call! :-)

rpv-tomsk:
My 'tiny' home server also has a set of enabled plugins similar to your:

```
LoadPlugin syslog
LoadPlugin cpu
LoadPlugin df
LoadPlugin disk
LoadPlugin entropy
LoadPlugin interface
LoadPlugin load
LoadPlugin memory
LoadPlugin network
LoadPlugin processes
LoadPlugin swap
```
That is collectd-5.1 and it shows:

```
# ps u 3215
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root      3215  0.2  0.0  46568  1288 ?        SLl  Jan26  97:03 collectd -C /etc/collectd/collectd.conf -f
```

Other, production server has collectd-5.6 and resource usage is the following:

```
# ps u 8376
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root      8376  0.2  0.7 401684  2612 ?        SLl  Jan13 186:50 collectd -C /etc/collectd/collectd.conf -f
```

It also has a small set of plugins:

```
LoadPlugin syslog
<Plugin syslog>
        LogLevel info
</Plugin>
LoadPlugin cpu
LoadPlugin df
LoadPlugin disk
LoadPlugin entropy
LoadPlugin interface
LoadPlugin load
LoadPlugin memory
LoadPlugin network
LoadPlugin processes
LoadPlugin swap
LoadPlugin users
LoadPlugin xencpu
```


Compared to other agents, it really tiny....

```
#ps u xxxx 
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
nginx     xxxx  0.8  0.1 300192 65760 ?        S    Feb24  27:41 amplify-agent
```
Mostly that is due to the fact - amplify-agent is python-based.

Zabbix agent also shows tiny RAM usage.

Maikel-Predikkta:
Okay, that prompted me to investigate which plugin eats all my memory. The Debian default configuration seems to use 8 MB:
```
root      5981  1.0  0.8 781008  8708 ?        Ssl  22:23   0:00 /usr/sbin/collectd
```

The http_write plugin bumps it to 33 MB:
```
root      6056  9.0  3.3 863900 33584 ?        Ssl  22:24   0:00 /usr/sbin/collectd
```

And after it ran for a while, is uses 50 MB:
```
root      5827  0.1  5.0 865960 51204 ?        Ssl  22:13   0:00 /usr/sbin/collectd
```

I guess that's fair enough comparing it to curl which uses 27 MB on my system.

octo:
Yes, it's subjective, but it's also out self-promotion, and in a way to goal to which we aspire, so I think it's find being subjective in this case.

Best regards,
—octo

PR #2695, opened 2018-02-22
Replace gone link target with archived version

Maikel-Predikkta:
Since the old page doesn't exist any more, we can link to archive.org.

Other options:

- Delete the link.
- Copy the archived text to the current website.


rpv-tomsk:
Hi!

I see absolutely no problems with https://collectd.org/wiki/index.php/Submitting_patches page.

PR #2700, opened 2018-02-28
Fix crashes when some external APIs fail

ZhouyangJia:
Hi,

I analyzed the collectd source code and found some potential API bugs that may cause crashes. These crashes are mainly caused by insufficient error handling of API functions like PyType_Ready or pthread_create.

I think it's unsafe to assume the library function would be correct. It would be better if we could handle the error properly (at least printing a message).

Best,
Zhouyang

### DIFF
--- a/src/daemon/collectd.c
+++ b/src/daemon/collectd.c
@@ -71,7 +71,8 @@ static void sig_usr1_handler(int __attribute__((unused)) signal) {
    * so it should be done asynchronously */
   pthread_attr_init(&attr);
   pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
-  pthread_create(&thread, &attr, do_flush, NULL);
+  if (pthread_create(&thread, &attr, do_flush, NULL) != 0)
+    fprintf(stderr, "Can not create thread: %s.\n", strerror(errno));
   pthread_attr_destroy(&attr);
 }
 
--- a/src/python.c
+++ b/src/python.c
@@ -1257,16 +1257,34 @@ static int cpy_init_python(void) {
   Py_Initialize();
   python_sigint_handler = PyOS_setsig(SIGINT, cur_sig);
 
-  PyType_Ready(&ConfigType);
-  PyType_Ready(&PluginDataType);
+  if (PyType_Ready(&ConfigType) < 0) {
+    cpy_log_exception("python initialization");
+    return 1;
+  }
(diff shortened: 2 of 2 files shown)


src/python.c
```diff
@@ -1257,16 +1257,34 @@ static int cpy_init_python(void) {
   Py_Initialize();
   python_sigint_handler = PyOS_setsig(SIGINT, cur_sig);
 
-  PyType_Ready(&ConfigType);
-  PyType_Ready(&PluginDataType);
+  if (PyType_Ready(&ConfigType) < 0) {
+    cpy_log_exception("python initialization");
```
dago: This and the below error messages are all the same, this makes debugging harder than necessary, please change these to include the actual failure in the error message so each error can be tracked to the respective line.

dago:
@ZhouyangJia Would you mind rebasing the code resolving the conflicts and adjust to individual error messages? I would like to include this PR in the upcoming 5.11.0.

ZhouyangJia:
Hi, I add individual error messages in the error-handling code in #3402.
The code related to the fix in src/daemon/collectd.c has been deleted, so I remove that part.

Thanks

src/daemon/collectd.c
```diff
@@ -71,7 +71,8 @@ static void sig_usr1_handler(int __attribute__((unused)) signal) {
    * so it should be done asynchronously */
   pthread_attr_init(&attr);
   pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
-  pthread_create(&thread, &attr, do_flush, NULL);
+  if (pthread_create(&thread, &attr, do_flush, NULL) != 0)
+    fprintf(stderr, "Can not create thread: %s.\n", strerror(errno));
```
rubenk: At this point we're already forked so stderr is closed.
You're also using strerror() which is not async signal safe.

I don't think handling the error buys us anything here.

src/python.c
```diff
@@ -1257,16 +1257,34 @@ static int cpy_init_python(void) {
   Py_Initialize();
   python_sigint_handler = PyOS_setsig(SIGINT, cur_sig);
 
-  PyType_Ready(&ConfigType);
-  PyType_Ready(&PluginDataType);
+  if (PyType_Ready(&ConfigType) < 0) {
```
rubenk: PyType_Ready returns -1 on failure, please compare it with -1 directly.

PR #2701, opened 2018-02-28
virt: Fix tracking of VM state changes

rjablonx:
Introduced support for events that are sent directly from libvirt – it can be enabled when “PersistentNotification” parameter is set to “false”. This change is needed to handle properly all notifications for VMs state changes. In that mode collectd sends notification as soon as possible, right after it is received from libvirt (previously some state changes notifications could be not sent by collectd because of issues with caching domains data).
When “PersistentNotification“ parameter is set to “true”, then old implementation will be used – it handles state changes notification on its own in plugin code by caching VM states and comparing state changes but there are some known limitations (e.g there are problems with notifications for domains that were removed from cache before notification could be send).

Also introduced support for dispatching domain state information – previously information about current VM state could only be received using notification and it was not stored in CSV files. Now information about state with corresponding state reason is dispatched as ‘domain_state’,

### DIFF
--- a/src/daemon/plugin_mock.c
+++ b/src/daemon/plugin_mock.c
@@ -71,21 +71,36 @@ int plugin_register_data_set(const data_set_t *ds) { return ENOTSUP; }
 
 int plugin_dispatch_values(value_list_t const *vl) { return ENOTSUP; }
 
-int plugin_dispatch_notification(const notification_t *notif) { return ENOTSUP; }
+int plugin_dispatch_notification(const notification_t *notif) {
+  return ENOTSUP;
+}
 
 int plugin_notification_meta_add_string(notification_t *n, const char *name,
-                                        const char *value) { return ENOTSUP; }
+                                        const char *value) {
+  return ENOTSUP;
+}
 int plugin_notification_meta_add_signed_int(notification_t *n, const char *name,
-                                            int64_t value) { return ENOTSUP; }
+                                            int64_t value) {
+  return ENOTSUP;
+}
 int plugin_notification_meta_add_unsigned_int(notification_t *n,
-                                              const char *name, uint64_t value) { return ENOTSUP; }
+                                              const char *name,
+                                              uint64_t value) {
+  return ENOTSUP;
+}
 int plugin_notification_meta_add_double(notification_t *n, const char *name,
-                                        double value) { return ENOTSUP; }
+                                        double value) {
+  return ENOTSUP;
+}
 int plugin_notification_meta_add_boolean(notification_t *n, const char *name,
-                                         _Bool value) { return ENOTSUP; }
+                                         _Bool value) {
+  return ENOTSUP;
+}
 
 int plugin_notification_meta_copy(notification_t *dst,
-                                  const notification_t *src) { return ENOTSUP; }
+                                  const notification_t *src) {
+  return ENOTSUP;
+}
 
 int plugin_notification_meta_free(notification_meta_t *n) { return ENOTSUP; }
 
--- a/src/daemon/plugin_mock.c
+++ b/src/daemon/plugin_mock.c
@@ -71,38 +71,64 @@ int plugin_register_data_set(const data_set_t *ds) { return ENOTSUP; }
 
 int plugin_dispatch_values(value_list_t const *vl) { return ENOTSUP; }
 
-int plugin_dispatch_notification(const notification_t *notif) {
+int plugin_dispatch_notification(__attribute__((unused))
+                                 const notification_t *notif) {
   return ENOTSUP;
 }
 
-int plugin_notification_meta_add_string(notification_t *n, const char *name,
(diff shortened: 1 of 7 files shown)


src/virt.c
```diff
@@ -1578,6 +1817,145 @@ static int get_if_dev_stats(struct interface_device *if_dev) {
...
+    sfree(domains);
+    DEBUG(PLUGIN_NAME " plugin: notified state of %i persistent domains",
+          n_notified);
+  }
+#else
+  n = virConnectNumOfDomains(conn);
+  if (n > 0) {
+    int *domids;
+    /* Get list of domains. */
+    domids = malloc(sizeof(*domids) * n);
```
vmytnykx: use calloc instead

src/virt.c
```diff
@@ -1793,224 +2214,231 @@ static int lv_instance_include_domain(struct lv_read_instance *inst,
...
+  int *domids;
 
-    /* Get list of domains. */
-    domids = malloc(sizeof(*domids) * n);
-    if (domids == NULL) {
-      ERROR(PLUGIN_NAME " plugin: malloc failed.");
-      return -1;
-    }
+  /* Get list of domains. */
+  domids = malloc(sizeof(*domids) * n);
```
vmytnykx: same

rjablonx:
Hi guys,

Is there any possibility to get list of packages with their versions installed on images used for CI verification (e.g. output from dpkg -l)?
I'm asking because I'm trying to reproduce error on debian-wheezy (https://ci.collectd.org/job/pull-requests-build-on-wheezy-amd64-with-default-toolchain/1835/console) - but I was not able to get this kind of error during my checks.
I was using docker debian:wheezy image with collectd dependencies installed but on mine setup all testcases ended with PASS (All 17 tests passed including virt however tests-output looks different compared to CI console output(logs from each unit test are printed out directly on the screen instead file))
So in order to reproduce and fix that I think I need packages versions that are installed debian-wheezy image used for verification.

Thanks in advance,
Radek

rubenk:
Yes, it's an artifact in Jenkins:
https://ci.collectd.org/job/pull-requests-build-on-wheezy-amd64-with-default-toolchain/1835/artifact/dpkg-l.txt

rubenk:
Same for the test log btw: https://ci.collectd.org/job/pull-requests-build-on-wheezy-amd64-with-default-toolchain/1835/artifact/collectd-5.8.0.128.g8be6889/test-suite.log

rjablonx:
@rubenk Thanks for pointing me to the artifacts - I've also seen test-logs with memory leaks reports and it seems that this is old problem already mentioned in Makefile.am - libnl1 is leaking  on wheezy and there is hard to do something with it from collectd side.
On master all virt-unit tests are skipped because of that bug but I'll try to be more selective and skip them only if problematic libnl1 has been discovered during configure step.

Thanks,
Radek

rubenk:
Yes, sorry to put you through all that trouble.
Debian Wheezy goes EOL at the end of next month so this problem will go away soon.

rjablonx:
Hi guys,
is there any way to re-trigger travis-ci build? I've looked into logs and it seems that travis had some issues yesterday -  job died long before actual collectd sources build so this fail don't seem to be connected with last patch.

src/daemon/plugin_mock.c
```diff
@@ -71,6 +71,39 @@ int plugin_register_data_set(const data_set_t *ds) { return ENOTSUP; }
 
 int plugin_dispatch_values(value_list_t const *vl) { return ENOTSUP; }
 
+int plugin_dispatch_notification(const notification_t *notif) {
```
rubenk: can you mark the parameters as __attribute__((unused))?

src/virt_test.c
```diff
@@ -25,183 +25,71 @@
 #include "testing.h"
 #include "virt.c" /* sic */
 
-#include <unistd.h>
+static virDomainPtr *domains = NULL;
```
rubenk: no need to zero-initialize static variables.

src/virt.c
```diff
@@ -90,6 +108,14 @@
 
 #endif /* LIBVIR_CHECK_VERSION */
 
+/* structure used for aggregating notification-thread data*/
+typedef struct virt_notif_thread_s {
+  pthread_t event_loop_tid;
+  int domain_event_cb_id;
+  pthread_mutex_t active_mutex; /* protects 'is_active' member access*/
+  _Bool is_active;
```
rubenk: it is ok to use bool, true and false now.

rjablonx: yes it makes sense - it will look way cleaner than this '_Bool':) This and above reports will be fixed in next update
Thanks for review @rubenk

rjablonx:
Hi @rubenk,
I've just updated pull request in order to address issues from your last code review.

rubenk:
Thanks @rjablonx!

PR #2703, opened 2018-03-01
Fixes for ipmi plugin

MariuszX:
This PR contains three fixes for ipmi plugin:
- Fix crash in ipmi.so while interval set below 0,5
- Fix to ignore events from sensors on ignore list.
- Fix registration for events from specific sensors (e.g. 'System Airflow')

### DIFF
--- a/src/ipmi.c
+++ b/src/ipmi.c
@@ -794,11 +794,9 @@ entity_sensor_update_handler(enum ipmi_update_e op,
 
     if (st->sel_enabled) {
       int status = 0;
-      /* register threshold event if threshold sensor support events */
-      if ((ipmi_sensor_get_event_reading_type(sensor) ==
-           IPMI_EVENT_READING_TYPE_THRESHOLD) &&
-          (ipmi_sensor_get_threshold_access(sensor) !=
-           IPMI_THRESHOLD_ACCESS_SUPPORT_NONE))
+      /* register threshold event handler */
+      if (ipmi_sensor_get_event_reading_type(sensor) ==
+          IPMI_EVENT_READING_TYPE_THRESHOLD)
         status = ipmi_sensor_add_threshold_event_handler(
             sensor, sensor_threshold_event_handler, st);
       /* register discrete handler if discrete/specific sensor support events */
--- a/src/ipmi.c
+++ b/src/ipmi.c
@@ -1210,7 +1210,7 @@ static int c_ipmi_init(void) {
   }
 
   /* Don't send `ADD' notifications during startup (~ 1 minute) */
-  int cycles = 1 + (60 / CDTIME_T_TO_TIME_T(plugin_get_interval()));
+  int cycles = 1 + (int)(TIME_T_TO_CDTIME_T(60) / plugin_get_interval());
 
   st = instances;
   while (NULL != st) {
--- a/src/ipmi.c
+++ b/src/ipmi.c
@@ -388,8 +388,12 @@ static int sensor_list_add(c_ipmi_instance_t *st, ipmi_sensor_t *sensor) {
         ipmi_sensor_get_event_support(sensor));
 
   /* Both `ignorelist' and `sensor_name_ptr' may be NULL. */
-  if (ignorelist_match(st->ignorelist, sensor_name_ptr) != 0)
-    return 0;
+  if (ignorelist_match(st->ignorelist, sensor_name_ptr) != 0) {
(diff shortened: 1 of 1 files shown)


src/ipmi.c
```diff
@@ -388,8 +388,12 @@ static int sensor_list_add(c_ipmi_instance_t *st, ipmi_sensor_t *sensor) {
         ipmi_sensor_get_event_support(sensor));
 
   /* Both `ignorelist' and `sensor_name_ptr' may be NULL. */
-  if (ignorelist_match(st->ignorelist, sensor_name_ptr) != 0)
-    return 0;
+  if (ignorelist_match(st->ignorelist, sensor_name_ptr) != 0) {
```
vmytnykx: This logic you have moved into L785, so this code is duplicate and will never be called.

src/ipmi.c
```diff
@@ -798,9 +805,8 @@ entity_sensor_update_handler(enum ipmi_update_e op,
             sensor, sensor_discrete_event_handler, st);
 
       if (status) {
```
vmytnykx: brackets are not needed in this condition.

rpv-tomsk:
>Fix to ignore events from sensors on ignore list.

What all you think about moving this logic to a new configuration option, so that we  can independently disable notifications and metrics?

src/ipmi.c
MariuszX: done

src/ipmi.c
MariuszX: done

MariuszX:
I`ve uploaded changes suggested by Volodymyr.
About new config option, IMHO anyone interested in notification will also be interested in metrics which caused notification. The same in opposite direction.
New config option could also be implemented if requested/agreed.
So I`m waiting for comments....

vmytnykx:
Agree with @rpv-tomsk, that it would be great to separate the metrics and notification filtering configuration option as with current approach user will not be able, for instance, get all metric stats and disable some sort of notifications. So, what about new configuration option like ```SELIgnoreSelected``` which will be specific for notifications only?

kwiatrox:
The option to filter only notifications sounds good, but we should consider any backward compatibility issues. I think at best it should be added as new feature not necessarily part of this fix.
The change itself looks good to me.

rpv-tomsk:
If we add a new configuration option ('SELSensor', not 'SELIgnoreSelected'), it will be absolutely backwards compatible, because by default it will be empty and, as result, will pass all sensors like it is in current implementation. Regardless of backward compatibility, that will be a new feature, because that adds new option. Collectd policy is conservative here.

Also, wants to notice some about 'IgnoreSelected' option - I think we also need to implement it as separate 'SELIgnoreSelected' option. 
Your opinions?

MariuszX:
Agree. We will also need options for metrics only sensors MetricsSensor or similar and corresponding MetricsIgnoreSelected.

rpv-tomsk:
>We will also need options for metrics only sensors MetricsSensor or similar and corresponding MetricsIgnoreSelected.

Existing `Sensor` and `IgnoreSelected` are good for this. I think that does not need to be changed.

MariuszX:
Currently Sensor option selects both metrics and notifications, SELSensor will select notification only. How to handle case when user is only interested in metrics ?

rpv-tomsk:
In current trunk implementation, `Sensor` option does not affect SEL notifications.

vmytnykx:
As discussed.... once the cd660c5 commit is removed/moved into another PR as a feature request, I think the rest changes/fixes are good to be merged in scope of current PR.

MariuszX:
Commits related to 'Fix to ignore events from sensors on ignore list' reverted. So rest is ready to be merged.

kwiatrox:
Thanks @MariuszX , the change looks good now, and fixes the bugs. I also like the idea with new feature for SELSensor and SELIgnoreSelected options.
@rpv-tomsk do you think is should be backported to 5.6? If yes Mariusz needs to rebase it against 5.6 branch.

MariuszX:
Just checked. In case of 5.6 and 5.7 only reformatted 'Fix crash in ipmi.so while interval set below 0,5' is needed. Code related to second fix was introduced in 5.8

MariuszX:
Reformatted 'Fix crash in ipmi.so while interval set below 0,5'  is also needed in all 5.x branches (5.2, 5.3, 5.4, 5.5)

MariuszX:
Solaris build issues not related to this change. ( https://github.com/collectd/collectd/pull/2711 )

rpv-tomsk:
Thanks for your work!

PR #2704, opened 2018-03-01
Properly cleanup dropped MySQL connections.

dhrupadb:


dhrupadb:
FYI @igorpeshansky

src/mysql.c
```diff
@@ -270,12 +270,16 @@ static MYSQL *getconnection(mysql_database_t *db) {
   }
   db->is_connected = 0;
 
+  /* Close the old connection before initializing a new one. */
+  if (db->con != NULL) {
+    mysql_close(db->con);
+    db->con = NULL;
+  }
+
+  db->con = mysql_init(NULL);
```
igorpeshansky: You're turning this into an unconditional `mysql_init`, which is probably not what you intended. Maybe the following would be more efficient:
```c
  /* Close the old connection before initializing a new one. */
  if (db->con != NULL) {
    mysql_close(db->con);
  } else {
    db->con = mysql_init(NULL);
    if (db->con == NULL) {
      ERROR("mysql plugin: mysql_init failed: %s", mysql_error(db->con));
      return NULL;
    }
  }
```
? Or even (to minimize changes):
```c
  if (db->con == NULL) {
    db->con = mysql_init(NULL);
    if (db->con == NULL) {
      ERROR("mysql plugin: mysql_init failed: %s", mysql_error(db->con));
      return NULL;
    }
  } else {
    /* Close the existing connection before reconnecting. */
    mysql_close(db->con);
  }
```

rpv-tomsk:
Hi!

Thank you for your contribution!

Can you argue your solution by a reference to a documentation?

src/mysql.c
igorpeshansky: Never mind, what I said was silly. Should have read the docs better. `mysql_close` deallocates the connection object, so it needs to be initialized again. Your proposed code does it correctly, mine doesn't.

dhrupadb: Yep, the return on line 266 ensures that if we get this far the connection has dropped regardless and should be cleaned up.

dhrupadb:
Hi @rpv-tomsk, In operational scenarios where SSL connections are refreshed periodically (~ every 2 minutes), memory is leaked each time the connection is dropped and re-established. This patch fixes the memory leak issue.

dhrupadb:
Friendly ping. We'd like to get this merged as we want to cherry pick it into our fork.

igorpeshansky:
@octo @tokkee Thoughts?

rpv-tomsk:
Thanks for your work!

PR #2705, opened 2018-03-02
Notification nested metadata

abays:
Currently, notification metadata is restricted to a single level of key-value pairs.  It is not possible to properly nest another metadata dictionary as the value associated with a key.  The only option is to create a key/value pair with the value being a marshaled JSON string representation of the multi-level dictionary, riddled with extra escape characters (which quickly becomes chaotic as additional levels are introduced).  This unnecessarily increases the size of the outgoing data, and places additional processing burden on the third party application that receives that data, which is antithetical to the intended goal of low-latency event notifications.  We want the data to leave Collectd as cleanly as possible, and this patch allows for that for JSON output.  Legacy metadata functionality remains intact from existing plugins' point of view.  This patch merely provides the option for plugins to utilize nested notification metadata, should they so desire.

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -2287,20 +2287,52 @@ const data_set_t *plugin_get_ds(const char *name) {
   return ds;
 } /* data_set_t *plugin_get_ds */
 
-static int plugin_notification_meta_add(notification_t *n, const char *name,
-                                        enum notification_meta_type_e type,
-                                        const void *value) {
+/*
+ A semantic distinction in the signature names that follow:
+
+ "add" functions place the metadata object at the end of the
+ linked list of metadata objects belonging to the notification.
+
+ "append" functions place the metadata object at the end of the
+ linked list of metadata objects nested within another metadata
+ object of NM_TYPE_NESTED.
+*/
+
+static int plugin_notification_meta_append(notification_t *n,
+                                           notification_meta_t *m,
+                                           const char *name,
+                                           enum notification_meta_type_e type,
+                                           const void *value) {
+  // If n is passed and is not NULL, this metadata object will be append to the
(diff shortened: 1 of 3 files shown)


abays:
Here is an example of output received without support for nested metadata:

[{"labels":{"alertname":"collectd_connectivity_gauge","instance":"trex","connectivity":"eno2","type":"interface_status","severity":"OKAY","service":"collectd"},"annotations":{"summary":"","ves":"{\"domain\":\"stateChange\",\"eventId\":3,\"eventName\":\"interface eno2 up\",\"lastEpochMicrosec\":1520022192636007,\"priority\":\"high\",\"reportingEntityName\":\"collectd connectivity plugin\",\"sequence\":0,\"sourceName\":\"eno2\",\"startEpochMicrosec\":1520022192633363,\"version\":1.0,\"stateChangeFields\":{\"newState\":\"inService\",\"oldState\":\"outOfService\",\"stateChangeFieldsVersion\":1.0,\"stateInterface\":\"eno2\"}}"},"startsAt":"2018-03-02T20:23:12.635994076Z"}]

And here is an example with this patch:

[{"labels":{"alertname":"collectd_connectivity_gauge","instance":"trex","connectivity":"eno2","type":"interface_status","severity":"FAILURE","service":"collectd"},"annotations":{"summary":"","ves":{"domain":"stateChange","eventId":"2","eventName":"interface eno2 down","lastEpochMicrosec":"1520019322204489","priority":"high","reportingEntityName":"collectd connectivity plugin","sourceName":"eno2","version":"1","stateChangeFields":{"newState":"outOfService","oldState":"inService","stateChangeFieldsVersion":"1","stateInterface":"eno2"}}},"startsAt":"2018-03-02T19:35:22.204485230Z"}]

PR #2712, opened 2018-03-14
write_prometheus plugin: Add support for custom labels

cryptobioz:
Using the "Label" parameter in the write_prometheus configuration will add the specified label to all metrics exposed by the plugin.
This parameter can be used as many times as needed.

### DIFF
--- a/src/write_prometheus.c
+++ b/src/write_prometheus.c
@@ -59,6 +59,69 @@ static struct MHD_Daemon *httpd;
 
 static cdtime_t staleness_delta = PROMETHEUS_DEFAULT_STALENESS_DELTA;
 
+struct label_map_s {
+  char value[DATA_MAX_NAME_LEN];
+};
+typedef struct label_map_s label_map_t;
+
+static c_avl_tree_t *label_map = NULL;
+
+static int prom_add_label(const oconfig_item_t *ci) /* {{{ */
+{
+  label_map_t *map;
+  int status;
+  char *key;
+
+  if ((ci->values_num < 2) || (ci->values_num > 3) ||
+      (ci->values[0].type != OCONFIG_TYPE_STRING) ||
+      (ci->values[1].type != OCONFIG_TYPE_STRING)) {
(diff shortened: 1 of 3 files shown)


src/write_prometheus.c
```diff
@@ -167,7 +232,23 @@ static char *format_labels(char *buffer, size_t buffer_size,
              escape_label_value(value, sizeof(value), m->label[i]->value));
   }
 
-  strjoin(buffer, buffer_size, labels, m->n_label, ",");
+  /* Add custom labels */
```
rpv-tomsk: I think it has no much sense to build array of custom label values each time when `format_labels()` called.
It would be better to build string of custom labels once, preferrable in `prom_config()`, and then just append it in `format_labels()`. Also, no `calloc()/free()` will be introduced here. 

What do you think about such improvement?

mfournier:
NB: this implementation only works for plain text responses, not protocol-buffers.

I think we should discuss if we want to keep supporting protocol-buffers. Prometheus itself dropped support (see https://prometheus.io/docs/instrumenting/exposition_formats/ and https://github.com/RichiH/OpenMetrics/blob/master/markdown/protobuf_vs_text.md), but on the other hand the upcoming OpenMetrics standard would require protobuf support.

eero-t:
Why this would not be better implemented using value metadata, like is done with the Rieman plugin: https://github.com/collectd/collectd/pull/3158/files ?

That would allow per-metric labels, as requested in https://github.com/collectd/collectd/issues/2995.
 
To add metadata to metric, I think input plugins would need just to do this on submit:
```
  value_list_t vl = VALUE_LIST_INIT;
...
  if ((vl.meta = meta_data_create()) != NULL) {
    meta_data_add_string(vl.meta, "desc", "this is some random metric");
  }
```

eero-t:
"write_prometheus" v6 API PR with per-metric label support was just merged to "collectd-6.0" branch.

If you still want to be able to specify labels that are added to **all** metrics, it would make sense to rebase these changes to the v6 version of "write_prometheus".

PR #2725, opened 2018-03-23
utils_cache: fix pointer reference and meta_data ingestion

RafaelMarinheiro:
This changes a bad pointer reference in uc_iterator_get_values, which
would cause the grpc plugin to crash.

This also fixes uc_insert by making it properly add the meta_data to
the cache.

Both of these changes should make the grpc QueryValues call work as
expected.

### DIFF
--- a/src/daemon/utils_cache.c
+++ b/src/daemon/utils_cache.c
@@ -203,6 +203,10 @@ static int uc_insert(const data_set_t *ds, const value_list_t *vl,
   ce->interval = vl->interval;
   ce->state = STATE_OKAY;
 
+  if (vl->meta != NULL) {
+    ce->meta = meta_data_clone(vl->meta);
+  }
+
   if (c_avl_insert(cache_tree, key_copy, ce) != 0) {
     sfree(key_copy);
     ERROR("uc_insert: c_avl_insert failed.");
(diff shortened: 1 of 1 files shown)


src/daemon/utils_cache.c
```diff
@@ -203,6 +203,10 @@ static int uc_insert(const data_set_t *ds, const value_list_t *vl,
   ce->interval = vl->interval;
   ce->state = STATE_OKAY;
 
+  if (vl->meta != NULL) {
+    ce->meta = meta_data_clone(vl->meta);
```
tokkee: Hrm, I'm not sure what ce->meta is supposed to do but I think it was not meant for this purpose. It was introduced in 62e42f8dbbc99815f06d7204fc410bca9aece4ef but doesn't seem to be used since.

@octo can you comment on this?

tokkee: Friendly ping for @octo, do you recall history around this?

rpv-tomsk:
I can merge part of this PR, w/o `meta_data_clone()` added.

rpv-tomsk:
cc @mrunge 

Mattias, for your info. If I had PR merge permission, I could add such minor bugfix myself, w/o anybody disturbed and w/o long awaiting. But now I'm unable to do this and bugfixes are significally delayed. 
At practice, they are not delayed but stopped.
Do you think this is good?

Before you offered your help in PR merge, but after I lost merge permission, we worked together with @rubenk. He merged my PR's, and I merged his. But at some point he quiletly left project and do not respond to my letters now. Work with him stopped.

So, I have point of failure in task of PR merging. My work strongly depends on active developers, which quantity is near zero. So, my work also tends to zero, like The Project in overall.

What do you think?

mrunge:
Thank you @RafaelMarinheiro for your fix, sorry this took sooooo long to get it merged.

@rpv-tomsk Thank you Pavel for the pointer.

rpv-tomsk:
Are you sure we need to merge `meta_data_clone()` too?

PR #2726, opened 2018-03-23
text protocol: Add support for meta_data in PUTVAL

RafaelMarinheiro:
This adds the support for string meta_data in PUTVAL. This allows users
to send meta_data using either the unixsock plugin or the exec plugin.

The meta_data is specified using options of the form `meta:<KEY>=<VALUE>`.
As an example, meta:foo=bar will add an entry with key "foo" and value
"bar".

### DIFF
--- a/src/utils_cmd_putval.c
+++ b/src/utils_cmd_putval.c
@@ -40,11 +40,11 @@ static int is_quoted(const char *str, size_t len) {
     return 0;
   }
 
-  return (str[0] == '"' && str[len - 1] == '"') ||
-         (str[0] == '\'' && str[len - 1] == '\'');
+  return (str[0] == '"' && str[len - 1] == '"');
 }
 
-static int set_option(value_list_t *vl, const char *key, const char *value) {
+static int set_option(value_list_t *vl, const char *key, const char *value,
+                      cmd_error_handler_t *err) {
   if ((vl == NULL) || (key == NULL) || (value == NULL))
     return -1;
 
@@ -65,23 +65,30 @@ static int set_option(value_list_t *vl, const char *key, const char *value) {
     if (vl->meta == NULL) {
       vl->meta = meta_data_create();
       if (vl->meta == NULL) {
-        return 1;
+        return CMD_ERROR;
       }
     }
 
     if (is_quoted(value, value_len)) {
+      int metadata_err;
       const char *value_str = strndup(value + 1, value_len - 2);
       if (value_str == NULL) {
-        return 1;
+        return CMD_ERROR;
       }
-      meta_data_add_string(vl->meta, meta_key, value_str);
+      metadata_err = meta_data_add_string(vl->meta, meta_key, value_str);
       free((void *)value_str);
+      if (metadata_err != 0) {
+        return CMD_ERROR;
+      }
+      return CMD_OK;
     }
-    return 1;
+
+    cmd_error(CMD_NO_OPTION, err, "Non-string metadata not supported yet");
+    return CMD_NO_OPTION;
   } else {
-    return 1;
+    return CMD_ERROR;
   }
-  return 0;
+  return CMD_OK;
 } /* int set_option */
 
 /*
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
Can you please describe use-cases?

faxm0dem:
We use meta-data for puppet/facter information, like operating system name, version, hardware brand and model, etc. Makes data mining possible in metrics stored for instance in Elasticsearch

tokkee:
Ftr, this brings unixsock/exec in line with other write plugins that already expose metadata, e.g. everything using format_json (e.g. write_http) or the grpc plugin. While I'm not actually super-happy about that, I think consistency is a good argument in itself. I think for the future, we'll want to think about internal meta-data (e.g. as used by the network plugin) and external meta-data (metrics tags).

src/utils_cmd_putval.c
```diff
@@ -49,9 +58,29 @@ static int set_option(value_list_t *vl, const char *key, const char *value) {
...
+      }
+    }
 
+    if (is_quoted(value, value_len)) {
+      const char *value_str = strndup(value + 1, value_len - 2);
+      if (value_str == NULL) {
+        return 1;
+      }
+      meta_data_add_string(vl->meta, meta_key, value_str);
+      free((void *)value_str);
```
tokkee: This is missing a return 0, right? Also, I think we should print an error in the 'else' case:

    if (is_quoted(value, value_len)) {
      …
      return 0;
    }
    cmd_error(CMD_PARSE_ERROR, err, "Non-string metadata not supported yet");
    return CMD_PARSE_ERROR;

src/utils_cmd_putval.c
```diff
@@ -35,6 +35,15 @@
  * private helper functions
  */
 
+static int is_quoted(const char *str, size_t len) {
+  if (len < 2) {
+    return 0;
+  }
+
+  return (str[0] == '"' && str[len - 1] == '"') ||
+         (str[0] == '\'' && str[len - 1] == '\'');
```
tokkee: utils_cmds.c currently only supports double quotes. I think we should do the same here.

src/utils_cmd_putval.c
tokkee: Hrm, also, the caller currently ignores the return value of set_option, so that would have to be fixed as well :-/

RafaelMarinheiro: Done. I am using the values defined in utils_cmds.h [1] here.

[1]: https://github.com/collectd/collectd/blob/9ae06bc2026e041e937828379daa232d3c0afbf7/src/utils_cmds.h#L111

src/utils_cmd_putval.c
RafaelMarinheiro: Done.

rpv-tomsk:
Hi!

Thanks for your work. 

>While I'm not actually super-happy about that, I think consistency is a good argument in itself. 
>I think for the future, we'll want to think about internal meta-data (e.g. as used by the network plugin) and external meta-data (metrics tags).

I think we really have to think about this, because there were requests about similar features.
We have to teach `network` to transmit/receive meta then.

Ok, related to this issue, we need in updated documentation: `src/collectd-exec.pod`. 
Can you please do this?

rpv-tomsk:
>I think consistency is a good argument in itself.

Ok, question about consistency. )
This PR adds a feature into `cmd_parse_putval()`. But what about `cmd_create_putval()`?

dago:
This is forward-ported in #3381

src/utils_cmd_putval.c
```diff
@@ -49,10 +58,37 @@ static int set_option(value_list_t *vl, const char *key, const char *value) {
...
 
-  return 0;
+    if (is_quoted(value, value_len)) {
+      int metadata_err;
+      const char *value_str = strndup(value + 1, value_len - 2);
+      if (value_str == NULL) {
+        return CMD_ERROR;
+      }
+      metadata_err = meta_data_add_string(vl->meta, meta_key, value_str);
+      free((void *)value_str);
```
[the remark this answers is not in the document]
dago: Yes, it is, or this error will be thrown:
```
./../src/utils/cmds/putval.c:79:12: error: passing 'const char *' to parameter of type 'void *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]

      free(value_str);

           ^~~~~~~~~

/usr/include/stdlib.h:483:25: note: passing argument to parameter '__ptr' here

extern void free (void *__ptr) __THROW;
```

dago:
Now there are probably real failures in the testsuite that needs to be analyzed.

dago:
#3381 has been closed

src/utils_cmd_putval.c
rpv-tomsk: @RafaelMarinheiro , don't publish code which you don't use and don't checked.
... And don’t trust anyone.

PR #2729, opened 2018-03-27
modbus: Add scale and shift to modbus

cekstam:
Adding the "Scale" and "Shift" parameters from the snmp-plugin to the modbus plugin in order to be able to correct amplified data.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -4158,6 +4158,16 @@ supported.
 Sets the type instance to use when dispatching the value to I<collectd>. If
 unset, an empty string (no type instance) is used.
 
+=item B<Scale> I<Value>
+
+The values taken from collectd are multiplied by I<Value>. The field is optional
+and the default is B<1.0>.
+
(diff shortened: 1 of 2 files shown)


cekstam:
Commit 3127b45acf1af66b849cb83e02843b5efad18362 includes a run of `contrib/format.sh src/modbus.c`, but it seems to fail , and commit 8270e0d09971aecd4ea32360486f18ea578a93e2 includes a modbus.c done with clang-format version 6.0.0.

Any pointers on how to fix this?

rpv-tomsk:
cc @rubenk
Ruben, let's review two commits related to modbus plugin?

I think they are good enough.

It may be better to merge this after #2670, to simplify conflict resolve (however, nothing complicated).

rubenk:
Sure, go ahead.

I know nothing about modbus and can't test this so will have to trust that this works.

rpv-tomsk:
I had polished this too when doing merge.

Thanks for your work, @cekstam !

Issue #2727, opened 2018-03-24
collectd-mod-mqtt - hangs after 20 (qos=1) messages are sent


viulian:
*   Version of collectd: 5.8.0-4
*   Operating system / distribution: LEDE snapshot (17.01)

## Actual behavior

collectd-mod-mqtt plugin hangs when it has to publish more than 20 messages with QoS 1 (it hangs after the 20th). No other messages are being sent afterwards unless collectd is restarted. After a while, the broker will disconnect the client (5minutes I guess).

## Steps to reproduce

*   configure collectd with some plugins that produce output (memory, df, load, interface, etc) where there should be more than 20 measurements taken.
*   configure collectd-mod-mqtt plugin to publish the messages to a broker
*   start collectd.

Only 20 messages get sent then everything gets stuck.

I've narrowed it down to mqtt.c file not calling mosquitto_loop() after the call to mosquitto_publish(). It means that the client (collectd-mod-mqtt) will quickly fill out its inflight buffer, since only misquitto_loop() consumes the PUBACK from the broker. By default, libmosquitto has an inflight limit of 20 message (qos 1 or 2) - and without the confirmation from the broker, it will refuse to send more.

I have a small patch ready, but I haven't tested with QoS 2 for now, nor with TLS.

viulian:
#2730

mrunge:
This should be fixed in #2730

PR #2730, opened 2018-03-27
adding mosquitto_loop after mosquitto_publish, otherwise qos=1 messag…

viulian:
ChangeLog: collectd-mod-mqtt: hangs after 20 (qos=1) messages are sent

adding mosquitto_loop after mosquitto_publish, otherwise qos=1 messages would not be considered acknowledged by the broker.

A proposed fix for https://github.com/collectd/collectd/issues/2727

I haven't tested yet with messages of QoS=2 but now collectd-mod-mqtt is able to push more than 20 messages (without the mosquitto_loop reading the socket after mosquitto_publish, libmosquitto was unaware that the messages were acknowledged by the broker).

### DIFF
--- a/src/mqtt.c
+++ b/src/mqtt.c
@@ -445,6 +445,26 @@ static int publish(mqtt_client_conf_t *conf, char const *topic,
     return -1;
   }
 
+  // -1 means default timeout, 1000ms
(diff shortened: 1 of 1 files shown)


mrunge:
Thank you for your patch!

PR #2382, opened 2017-07-26
Plugin for  PCIe Advanced Error Reporting

kwiatrox:
The pcie_errors plugin collects PCI Express errors from Device Status in Capability
structure and from Advanced Error Reporting Extended Capability where available.
At every read it polls config space of PCI Express devices and dispatches
notification for every error that is set. It can be configured to parse log file
for system reported AER errors.
This plugin uses message log parser util so it depends on pull request #2154 .

PR #2733, opened 2018-03-28
New plugin to read PCIe errors

kwiatrox:
The pcie_errors plugin collects PCI Express errors from Device Status in Capability
structure and from Advanced Error Reporting Extended Capability where available.
At every read it polls config space of PCI Express devices and dispatches
notification for every error that is found. Notification is also sent after error is cleared.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6265,6 +6265,53 @@ Default: empty (monitor all bridges)
 
 =back
 
+=head2 Plugin C<pcie_errors>
+
+The I<pcie_errors> plugin collects PCI Express errors from Device Status in Capability
+structure and from Advanced Error Reporting Extended Capability where available.
+At every read it polls config space of PCI Express devices and dispatches
+notification for every error that is set. It checks for new errors at every read.
+The device is indicated in plugin_instance according to format "domain:bus:dev.fn".
+Errors are divided into categories indicated by type_instance: "correctable", and
+for uncorrectable errors "non_fatal" or "fatal".
+Fatal errros are reported as I<NOTIF_FAILURE> and all others as I<NOTIF_WARNING>.
+
+B<Synopsis:>
+
+  <Plugin "pcie_errors">
+    Source "sysfs"
+    AccessDir "/sys/bus/pci"
+    ReportMasked false
+    PersistentNotifications false
+  </Plugin>
+
+B<Options:>
+
+=over 4
+
+=item B<Source> B<sysfs>|B<proc>
+
+Use B<sysfs> or B<proc> to read data from /sysfs or /proc.
+The default value is B<sysfs>.
+
+=item B<AccessDir> I<dir>
+
+Directory used to access device config space. It is optional and defaults to
+/sys/bus/pci for B<sysfs> and to /proc/bus/pci for B<proc>.
+
+=item B<ReportMasked> B<false>|B<true>
+
+If true plugin will notify errors that are set to masked in Error Mask register.
+Such errors are not reported to the PCI Express Root Complex. Defaults to
+B<false>.
+
+=item B<PersistentNotifications> B<false>|B<true>
+
+If false plugin will dispatch notfication only on set/clear of error.
+The ones already reported will be ignored. Defaults to B<false>.
+
+=back
+
 =head2 Plugin C<perl>
 
 This plugin embeds a Perl-interpreter into collectd and provides an interface
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -6274,7 +6274,7 @@ notification for every error that is set. It checks for new errors at every read
 The device is indicated in plugin_instance according to format "domain:bus:dev.fn".
 Errors are divided into categories indicated by type_instance: "correctable", and
 for uncorrectable errors "non_fatal" or "fatal".
-Fatal errros are reported as I<NOTIF_FAILURE> and all others as I<NOTIF_WARNING>.
+Fatal errors are reported as I<NOTIF_FAILURE> and all others as I<NOTIF_WARNING>.
 
 B<Synopsis:>
 
@@ -6301,13 +6301,12 @@ Directory used to access device config space. It is optional and defaults to
 
 =item B<ReportMasked> B<false>|B<true>
 
-If true plugin will notify errors that are set to masked in Error Mask register.
-Such errors are not reported to the PCI Express Root Complex. Defaults to
-B<false>.
+If true plugin will notify about errors that are set to masked in Error Mask register.
+Such errors are not reported to the PCI Express Root Complex. Defaults to B<false>.
 
 =item B<PersistentNotifications> B<false>|B<true>
 
-If false plugin will dispatch notfication only on set/clear of error.
+If false plugin will dispatch notification only on set/clear of error.
 The ones already reported will be ignored. Defaults to B<false>.
 
 =back
--- a/src/pcie_errors.c
+++ b/src/pcie_errors.c
@@ -49,11 +49,10 @@
 #define PCIE_ECAP_OFFSET 0x100 /* ECAP always begin at offset 0x100 */
 
 typedef struct pcie_config_s {
-  _Bool use_sysfs;
-  _Bool notif_masked;
-  _Bool persistent;
+  bool use_sysfs;
+  bool notif_masked;
+  bool persistent;
   char access_dir[PATH_MAX];
-  _Bool config_error;
 } pcie_config_t;
 
 typedef struct pcie_device_s {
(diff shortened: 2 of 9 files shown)


rubenk:
How does this PR differ from #2382?

kwiatrox:
Hi @rubenk, this version does not depend on log parsing and is smaller. It has some unit tests added. I've closed the previous PR.

kwiatrox:
Hi @rubenk , @rpv-tomsk, any feedback or thoughts regarding this plugin? Any comments are welcome.

src/collectd.conf.pod
```diff
@@ -6051,6 +6051,53 @@ Default: empty (monitor all bridges)
...
+=head2 Plugin C<pcie_errors>
+
+The I<pcie_errors> plugin collects PCI Express errors from Device Status in Capability
+structure and from Advanced Error Reporting Extended Capability where available.
+At every read it polls config space of PCI Express devices and dispatches
+notification for every error that is set. It checks for new errors at every read.
+The device is indicated in plugin_instance according to format "domain:bus:dev.fn".
+Errors are divided into categories indicated by type_instance: "correctable", and
+for uncorrectable errors "non_fatal" or "fatal".
+Fatal errros are reported as I<NOTIF_FAILURE> and all others as I<NOTIF_WARNING>.
```
rubenk: s/errros/errors/

src/collectd.conf.pod
```diff
@@ -6051,6 +6051,53 @@ Default: empty (monitor all bridges)
...
+The default value is B<sysfs>.
+
+=item B<AccessDir> I<dir>
+
+Directory used to access device config space. It is optional and defaults to
+/sys/bus/pci for B<sysfs> and to /proc/bus/pci for B<proc>.
+
+=item B<ReportMasked> B<false>|B<true>
+
+If true plugin will notify errors that are set to masked in Error Mask register.
```
rubenk: notify about errors?

src/collectd.conf.pod
```diff
@@ -6051,6 +6051,53 @@ Default: empty (monitor all bridges)
...
+Such errors are not reported to the PCI Express Root Complex. Defaults to
+B<false>.
+
+=item B<PersistentNotifications> B<false>|B<true>
+
+If false plugin will dispatch notfication only on set/clear of error.
```
rubenk: s/notfication/notification/

src/pcie_errors.c
```diff
@@ -0,0 +1,779 @@
...
+/* Device Error Status */
+static pcie_error_t pcie_base_errors[] = {
+    {PCI_EXP_DEVSTA_CED, "Correctable Error"},
+    {PCI_EXP_DEVSTA_NFED, "Non-Fatal Error"},
+    {PCI_EXP_DEVSTA_FED, "Fatal Error"},
+    {PCI_EXP_DEVSTA_URD, "Unsupported Request"}};
+static const int pcie_base_errors_num = STATIC_ARRAY_SIZE(pcie_base_errors);
+
+/* Uncorrectable Error Status */
+static pcie_error_t pcie_aer_ues[] = {
```
rubenk: can you make this const?

src/pcie_errors.c
```diff
@@ -0,0 +1,779 @@
...
+  int mask;
+  const char *desc;
+} pcie_error_t;
+
+static llist_t *pcie_dev_list;
+static pcie_config_t pcie_config = {.access_dir = "", .use_sysfs = 1};
+static pcie_fops_t pcie_fops;
+
+/* Device Error Status */
+static pcie_error_t pcie_base_errors[] = {
```
rubenk: can you make this const?

src/pcie_errors.c
```diff
@@ -0,0 +1,779 @@
...
+#define PCIE_SEV_FATAL "fatal"
+#define PCIE_SEV_NOFATAL "non_fatal"
+
+#define PCIE_DEV(x) (((x) >> 3) & 0x1f)
+#define PCIE_FN(x) ((x)&0x07)
+
+#define PCIE_ECAP_OFFSET 0x100 /* ECAP always begin at offset 0x100 */
+
+typedef struct pcie_config_s {
+  _Bool use_sysfs;
```
rubenk: at your choosing, it's fine to use `bool` now.

src/pcie_errors.c
```diff
@@ -0,0 +1,779 @@
...
+    {PCI_ERR_UNC_ATOMEG, "Atomic egress blocked"},
+#endif
+#ifdef PCI_ERR_UNC_TLPPRE
+    {PCI_ERR_UNC_TLPPRE, "TLP prefix blocked"},
+#endif
+};
+static const int pcie_aer_ues_num = STATIC_ARRAY_SIZE(pcie_aer_ues);
+
+/* Correctable Error Status */
+static pcie_error_t pcie_aer_ces[] = {
```
rubenk: can you make this const?

src/pcie_errors.c
```diff
@@ -0,0 +1,779 @@
...
+  FILE *fd;
+  char file_name[PCIE_NAME_LEN];
+  char buf[PCIE_BUFF_SIZE];
+  unsigned int i = 0;
+  int ret = 0;
+
+  if (dev_list == NULL)
+    return -EINVAL;
+
+  snprintf(file_name, sizeof(file_name), "%s/devices", pcie_config.access_dir);
```
rubenk: please check for errors and truncation.

src/pcie_errors.c
```diff
@@ -0,0 +1,779 @@
...
+    unsigned int slot;
+    uint8_t bus, dev, fn;
+
+    if (sscanf(buf, "%x", &slot) != 1) {
+      ERROR(PCIE_ERRORS_PLUGIN ": Failed to read line %u from %s", i + 1,
+            file_name);
+      continue;
+    }
+
+    bus = slot >> 8U;
```
rubenk: declare the variables here please

src/pcie_errors.c
```diff
@@ -0,0 +1,779 @@
...
+static int pcie_list_devices_sysfs(llist_t *dev_list) {
+  DIR *dir;
+  struct dirent *item;
+  char dir_name[PCIE_NAME_LEN];
+  int ret = 0;
+
+  if (dev_list == NULL)
+    return -EINVAL;
+
+  snprintf(dir_name, sizeof(dir_name), "%s/devices", pcie_config.access_dir);
```
rubenk: check for errors and truncation please.

src/pcie_errors.c
```diff
@@ -0,0 +1,779 @@
...
+
+    if (status) {
+      ERROR(PCIE_ERRORS_PLUGIN ": Invalid configuration parameter \"%s\".",
+            child->key);
+      pcie_config.config_error = 1;
+      break;
+    }
+  }
+
+  return 0;
```
rubenk: please return an error on error.

kwiatrox: The idea here is to keep collectd running even if one plugin fails. If plugin config fails collectd stops loading. So error is stored in `pcie_config.config_error` and plugin returns error later on init, while the rest of the plugins can continue running until someone fixes the configuration and restart. Does it sound reasonable?

rubenk: I do understand your reasoning. However, if a plugin is misconfigured collectd shouldn't even start and the configuration should be fixed, making the whole system more predicatable. Configuration can then also be checked with collectd -t.

rpv-tomsk: We have to put this as a rule in developers documentation / style guide, and do some work to unify existing plugins behaviour. Yesterday I worked on `snmp` plugin and may say what it can start with wrong configuration (when `Data` block referenced from `Host` is not found, for example).

Maybe it's a time to do this?

rpv-tomsk:
>this version does not depend on log parsing and is smaller.

That is great! I don't like log parsing for such tasks....

src/pcie_errors.c
rubenk: Yes, good idea.

kwiatrox:
@rubenk, @rpv-tomsk All comments are addressed. I had to make rebase to use bool instead of _Bool.

kwiatrox:
Hi @rpv-tomsk,

Errors are usually recovered by hardware itself or software (driver), it depends on severity:
Corrected errors - Such errors are corrected by hardware and no software intervention is required.
Uncorrected Non-Fatal errors - Recovery from a non-fatal error depends on device-specific software associated with the requester that initiated the transaction.
Uncorrected Fatal errors - Recovery from fatal errors is done by resetting the component and link.

The recovery can fail, so we have the notification for cleared errors, because sometimes they are not cleared.

The information about errors is useful for higher layer, the persisting or not recovered errors on single device can indicate some problems with it.

rpv-tomsk:
Hi!

Thanks for explanations.

I had also put a comment about `types.db` above. 
I think that file should not be touched by this PR. Can you please check this / update PR?

kwiatrox:
Thank You for the fix,
Sorry for the delay, but I was busy with other assignments.

rpv-tomsk:

PR #2738, opened 2018-04-03
Disk pending operation data is missing

bufadu:
Hello,
I noticed that disk `pending_operation` datasource is missing on some of my servers. On these servers `in_progress` value is nearly always 0. I made a quick fix to allow `in_progress` value to be dispatched when equal to zero but I don't know if is it safe to compare with `NAN`..

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -819,7 +819,7 @@ static int disk_read(void) {
       if (read_merged || write_merged)
         ds->has_merged = 1;
 
-      if (in_progress)
+      if (in_progress != NAN)
         ds->has_in_progress = 1;
 
       if (io_time)


rpv-tomsk:
Hi!

On some 'not real' block devices (such as md raid) 'pending_operations' is _always_ equal to zero.
There is no practical use to send these zeroes as metrics and this is the reason for solution you see.  

Collectd does periodical polls, reading data from kernel. Not each of these polls will return non-zero value if you have fast hardware and small load which is not enough to fill kernel operations queue. 

After first occurence, when 'pending operations' is not zero, there is a flag set, so even zero values will be send as metric.

bufadu:
Ok, it makes sense. But depending of the server load, after collectd starts, you may have no pending operation dispatched until the value becomes non-zero. Once the value changes, collectd dispatches the value even if it's zero. So if you use rrd with collectd you will see holes in your graphs if collectd restarts.. 
More over, as you stated, some block devices have always pending operation equal to zero; but the value is actually provided by the kernel so why ignore it ?

That's my 2 cents. Up to you to choose which data collection strategy collectd should have. But IMHO, a monitoring system should be as agnostic as possible.

rpv-tomsk:
>But depending of the server load, after collectd starts, you may have no pending operation dispatched until the value becomes non-zero. 
...until the value becomes non-zero and Collectd _catch_ this value. This value go back to zero when no pending operations in queue.

>but the value is actually provided by the kernel so why ignore it?

Because it is not the actual value. It just placeholder in output format and _always_ is zero.

>But IMHO, a monitoring system should be as agnostic as possible.
Why we need to send unuseful (empty) metrics?

>So if you use rrd with collectd you will see holes in your graphs if collectd restarts.

On mine systems collectd restarts are rare.
If that metric has real data, then collectd catch non-zero values quickly and hole will be small.
If that metric has no real data then hole size is not critical.

Sometimes we need in compromises.

bufadu:
Ok. I close the pull request.
Thanks for your time !

PR #2740, opened 2018-04-04
Fix routeros unset radio-name showing up as "(null)"

melak:
This happens when the `radio-name` property is unset on a wireless
interface. Why this is done is beyond me, but it does seem to be a thing.


rubenk:
Thanks @melak!

PR #2742, opened 2018-04-10
ipmi plugin: Fix for segmentation fault for wrong data type in Sensor section.

MariuszX:
Incorrect, malicious data (i.e. numbers, true) in the Sensor section can cause a segmentation fault.
Error message is generated instead of crash after applying this patch.


rpv-tomsk:
Thanks!

PR #2745, opened 2018-04-16
Refactor collectd binary.

SeanCampbell:
Refactors Linux-specific code from collectd.c into collectd_linux.c. This makes it easier to use the non-Linux-specific code when we port to Windows.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2010,9 +2010,9 @@ install-exec-hook:
 	$(mkinstalldirs) $(DESTDIR)$(sysconfdir)
 	if test -e $(DESTDIR)$(sysconfdir)/collectd.conf; \
 	then \
-		$(INSTALL) -m 0640 $(srcdir)/src/collectd.conf $(DESTDIR)$(sysconfdir)/collectd.conf.pkg-orig; \
+		$(INSTALL) -m 0640 $(builddir)/src/collectd.conf $(DESTDIR)$(sysconfdir)/collectd.conf.pkg-orig; \
 	else \
-		$(INSTALL) -m 0640 $(srcdir)/src/collectd.conf $(DESTDIR)$(sysconfdir)/collectd.conf; \
+		$(INSTALL) -m 0640 $(builddir)/src/collectd.conf $(DESTDIR)$(sysconfdir)/collectd.conf; \
 	fi; \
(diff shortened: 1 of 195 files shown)


Makefile.am
```diff
@@ -193,6 +193,7 @@ endif
 
 
 collectd_SOURCES = \
+	src/daemon/collectd_linux.c \
```
tokkee: The code in this file isn't actually Linux specific but for all the (non-Windows) platforms currently supported by collectd. Since this file continues to include main(). I suggest to continue to call it collectd.c and special cases like Windows could then be collectd_windows.c. The code that you left in collectd.c / collectd.h would move to files like cmd.c / cmd.h and should only be used / linked into the main program (to avoid leaking symbols into other places).

src/daemon/collectd.c
```diff
@@ -86,13 +67,9 @@ static int init_hostname(void) {
     return 0;
   }
 
-  long hostname_len = sysconf(_SC_HOST_NAME_MAX);
-  if (hostname_len == -1) {
-    hostname_len = NI_MAXHOST;
-  }
-  char hostname[hostname_len];
+  char hostname[hostname_len()];
```
tokkee: Why did you introduce hostname_len() and call it twice? I think the code can continue to be inlined here.

src/daemon/collectd_linux.c
```diff
@@ -0,0 +1,274 @@
+/**
+ * collectd - src/collectd_linux.c
+ * Copyright (C) 2018  Google LLC
```
tokkee: The code in this file is copyright by the original authors.

src/daemon/collectd.c
SeanCampbell: This was to separate out some Windows vs. other OS behavior from the main logic of the program, but probably doesn't belong in this PR since it isn't actually needed yet.

SeanCampbell:
Made the changes you requested. Let me know if you want me to merge the two commits.

SeanCampbell:
Tests were breaking because of the collectd.h -> cmd.h renaming. Fixed those issues now.

tokkee:
Thank you! As discussed offline, I think there may be some wordsmithing that we wanna do but I think that can happen in a follow-up.

PR #2755, opened 2018-04-27
Add ability to turn on collectd "debug" feature in RPMs

dehotot:
Hi

This small PR adds the ability to build Debug-enabled collectd RPMs:

`$ rpmbuild -bb --with='debug' collectd.spec`

Without this flag, it does not appear possible to build 'debug' RPMs for use on RHELish systems.

The default behaviour has not changed — without the flag, the rpms are non-debug, as normal.


rubenk:
Looks good, thanks!

Issue #2680, opened 2018-02-09
virt plugin attempts to collect data from mcast interfaces


soeb:
*   Version of collectd: 5.8.0.106.g21df6c1
*   Operating system / distribution: CentOS 7

## Expected behavior

The virt plugin should not be attempting to collect data for multicast interfaces - this would be correct, as there is no data to collect (this will apply for anything without a source). For now I've had to resort to just filtering them out using _InterfaceDevice_ and _IgnoreSelected_.

## Actual behavior

The virt plugin attempts (and fails) to collect data from mcast interfaces/interfaces without a source.

Log contents:
```
virt plugin: virDomainInterfaceStats failed
virt failed to get interface stats for device (guestXXXmYY) in domain guest_XXX
```

Interfaces according to libvirt
```
# domiflist guest_XXX
Interface  Type       Source     Model       MAC
-------------------------------------------------------
guestXXX   bridge     vbrXXX     virtio      zz:zz:zz:zz:zz:zz
guestXXXmYY mcast      -          virtio      zz:zz:zz:zz:zz:zz
```

```
# domifstat guest_XXX guestXXXmYY
error: Failed to get interface stats guest_XXX guestXXXmYY
error: internal error: /proc/net/dev: Interface not found
```

## Steps to reproduce

*   Create a VM with an mcast interface defined for it (sample definition below)
*   Configure collectd with the virt plugin
*   Observe the results

```
    <interface type='mcast'>
      <mac address='zz:zz:zz:zz:zz:zz'/>
      <source address='239.0.0.1' port='10000'/>
      <target dev='guestXXXmYY'/>
      <model type='virtio'/>
      <driver name='vhost' ioeventfd='on'/>
      <alias name='net1'/>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x04' function='0x0'/>
    </interface>
```

PR #2757, opened 2018-04-30
Fix issue #2680 -- virt plugin mcast stats

abays:
Virt plugin will store interface type info, and then skip stats collection for that interface if its type is multicast

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -230,6 +230,8 @@ const char *domain_reasons[][DOMAIN_STATE_REASON_MAX_SIZE] = {
 };
 #endif /* HAVE_DOM_REASON */
 
+#define VIR_DOMAIN_INTERFACE_TYPE_MULTICAST "mcast"
+
 #define NR_CONFIG_KEYS ((sizeof config_keys / sizeof config_keys[0]) - 1)
 #define NANOSEC_IN_SEC 1e9
 
@@ -273,6 +275,7 @@ struct interface_device {
   char *path;       /* name of interface device */
   char *address;    /* mac address of interface device */
   char *number;     /* interface device number */
+  char *type;       /* interface type */
 };
 
 typedef struct domain_s {
@@ -302,7 +305,7 @@ static int add_block_device(struct lv_read_state *state, virDomainPtr dom,
 static void free_interface_devices(struct lv_read_state *state);
 static int add_interface_device(struct lv_read_state *state, virDomainPtr dom,
                                 const char *path, const char *address,
-                                unsigned int number);
+                                const char *type, unsigned int number);
 
 #define METADATA_VM_PARTITION_URI "http://ovirt.org/ovirtmap/tag/1.0"
 #define METADATA_VM_PARTITION_ELEMENT "tag"
@@ -1642,6 +1645,14 @@ static int lv_read(user_data_t *ud) {
 
   /* Get interface stats for each domain. */
   for (int i = 0; i < state->nr_interface_devices; ++i) {
+    if (strcmp(state->interface_devices[i].type,
+               VIR_DOMAIN_INTERFACE_TYPE_MULTICAST) == 0) {
+      DEBUG(PLUGIN_NAME " is skipping stat collection for multicast interface "
+                        "(%s) in domain %s",
+            state->interface_devices[i].path,
+            virDomainGetName(state->interface_devices[i].dom));
+      continue;
+    }
     int status = get_if_dev_stats(&state->interface_devices[i]);
     if (status != 0)
       ERROR(PLUGIN_NAME
@@ -1960,11 +1971,14 @@ static int refresh_lists(struct lv_read_instance *inst) {
       for (int j = 0; j < xml_interfaces->nodeNr; ++j) {
         char *path = NULL;
         char *address = NULL;
+        char *type = NULL;
         xmlNodePtr xml_interface;
 
         xml_interface = xml_interfaces->nodeTab[j];
         if (!xml_interface)
           continue;
+
+        type = (char *)xmlGetProp(xml_interface, (const xmlChar *)"type");
 
         for (xmlNodePtr child = xml_interface->children; child;
              child = child->next) {
@@ -1987,12 +2001,14 @@ static int refresh_lists(struct lv_read_instance *inst) {
              ignore_device_match(il_interface_devices, name, address) != 0))
           goto cont3;
 
-        add_interface_device(state, dom, path, address, j + 1);
+        add_interface_device(state, dom, path, address, type, j + 1);
       cont3:
         if (path)
           xmlFree(path);
         if (address)
           xmlFree(address);
+        if (type)
+          xmlFree(type);
       }
 
     cont:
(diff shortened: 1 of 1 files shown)


src/virt.c
```diff
@@ -1960,12 +1971,15 @@ static int refresh_lists(struct lv_read_instance *inst) {
...
         char *path = NULL;
         char *address = NULL;
+        char *type = NULL;
         xmlNodePtr xml_interface;
 
         xml_interface = xml_interfaces->nodeTab[j];
         if (!xml_interface)
           continue;
 
+        type = (char *)xmlGetProp(xml_interface, (const xmlChar *)"type");
```
atyronesmith: Do you need to check for error conditions here?

src/virt.c
```diff
@@ -2120,6 +2136,12 @@ static int add_interface_device(struct lv_read_state *state, virDomainPtr dom,
     return -1;
   }
 
+  type_copy = strdup(type);
+  if (!type_copy) {
+    sfree(type_copy);
```
atyronesmith: Should this be freeing type not type_copy?

src/virt.c
abays: No, but it does not an error check elsewhere.  Will fix.

src/virt.c
abays: Should be freeing address_copy and path_copy if they exist.  Nice find.

abays:
According to the reporter of issue #2680, this plugin shouldn't be collecting data for multicast interfaces, as they do not have a source device (and will thus cause the ```virDomainInterfaceStats``` libvirt call to fail).  Trying to do so produces errors like the following in the Collectd log:

```
virt plugin: virDomainInterfaceStats failed
virt failed to get interface stats for device (guestXXXmYY) in domain guest_XXX
```

Without that call, there are simply no metrics to gather.  So it would seem that we should always skip this type of interface (rather than even doing something like making metrics collection for it configurable).

rpv-tomsk:
Reporters of issue typically try to solve their own issue and that can't be applied in common form (typically).
If VM has interface, that interface can have traffic, so there can be statistics for that. 
Is I'm missing something?

abays:
Yes, there can be traffic for any given interface on a VM, but it appears to be a matter of whether it is possible to collect the statistics with the Libvirt API.  The function for gathering these interface stats is:

https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainInterfaceStats

The issue's reporter is saying that this function call will not work for multicast interfaces, so it should be avoided for those.  I'm no Libvirt expert, but if this function is the only option to collect the stats and it returns an error for multicast interfaces, then it indeed seems logical to skip the call for those interfaces.  This doesn't appear to be a special case related to the reporter's situation, so a common-form solution (always skipping multicast interfaces) would be appropriate.  I understand that some reported issues are due to user error or environmental configuration or what have you, but this problem seems to be a genuine bug in the way the plugin is trying to call the ```virDomainInterfaceStats``` API function.

rpv-tomsk:
>Yes, there can be traffic for any given interface on a VM, but it appears to be a matter of whether it is possible to collect the statistics with the Libvirt API. The function for gathering these interface stats is:

https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainInterfaceStats

Yes, and there is no comments about 'dont use this function for mcast interfaces'.

>The issue's reporter is saying that this function call will not work for multicast interfaces, so it should be avoided for those. 

That is only reporters opinion,  which is not backed by anything.

>I'm no Libvirt expert, but if this function is the only option to collect the stats and it returns an error for multicast interfaces, then it indeed seems logical to skip the call for those interfaces. 

There is no any comments in libvirt documentation about this function does not support multicast interfaces. There is desription:

```
This function returns network interface stats for interfaces attached to the domain.

The @device parameter is the network interface either by name or MAC address.

Domains may have more than one network interface. To get stats for each you should make multiple calls to this function.

Individual fields within the stats structure may be returned as -1, which indicates that the hypervisor does not support that particular statistic.
```

So, that is hypervisor-related behaviour. 

>This doesn't appear to be a special case related to the reporter's situation, so a common-form solution (always skipping multicast interfaces) would be appropriate. 

What is your opinion about 'special case' based on? I think reporter has only one hypervisor checked before reporting issue, there is nothing there to indicate otherwise. Why do we need to disable something in wide way based only on this report?

>I understand that some reported issues are due to user error or environmental configuration or what have you, but this problem seems to be a genuine bug in the way the plugin is trying to call the virDomainInterfaceStats API function.

I found nothing in `virDomainInterfaceStats` description what prohibits its use on multicast interfaces.
I found nothing what prohibits gathering stats from such interfaces, so when libvirt/hypervisor implementation will change - why do we need to return to this/similar PR again?

abays:
It may very well be a hypervisor issue.  I suppose there's no harm in trying to get the stats even though it will fail against the particular hypervisor the reporter was using (it just spams the logs, but that's not a big deal).  But the root cause appears to be a lack of an associated device in /proc/net/dev.  So it's not just the reporter's opinion, unfortunately, but I do agree now that a blanket skip of the interface is not the right approach.  I think, getting back to what the reporter said, there's more nuance than just the consideration of whether an interface is multicast or not.  As he said:

> The virt plugin attempts (and fails) to collect data from mcast interfaces/interfaces **without a source**.

Maybe that's the key here: the presence of a source.  It's plausible that a multicast interface without a source wouldn't work for the ```virDomainInterfaceStats``` call across different hypervisors.  Of course we'd need to test and check for this, but we can't assume that just because the documentation doesn't mention a problem with multicast/no-source that it must therefore always work.  I'll try to do some more digging.

rpv-tomsk:
My logic: If VM has interface, there _can be_ stats for this interface, independently of interface type or 'has source or not'.

rpv-tomsk:
>I agree. But there's also the consideration of whether you can collect those stats via the Libvirt API (which is what the plugin uses), given the particular interface. That is, there might be stats somewhere for the interface, but is the Libvirt API able to access them? 

Libvirt API _can_ provide stats for us. How libvirt should get these stats from hypervisor is only depends on libvirt implementation and is outside of Collectd.

>It would seem that a lack of a /proc/net/dev data for the interface prevents this call from working, which, again, could be hypervisor-specific, but we'll need to find out.

Looked in `libvirt-4.9.0-rc1`:

Some drivers - `lxc`, `qemu`, `openvz`, `libxl` - they use `virNetDevTapInterfaceStats()` from `src/util/virnetdevtap.c` and that function gathers data from `/proc/net/dev`.
So, it can fail with message `/proc/net/dev: Interface not found` when interface 'has no source'.

But there are other ways of gathering data too:

For `qemu` driver there is also another call, to `virNetDevOpenvswitchInterfaceStats`, which uses another datasource, not `/proc/net/dev`.
For `vz` driver there is another call too.

Whether libvirt will provide stats or return an error is the details of the libvirt implementation and may change at any time.

As `virDomainInterfaceStats` itself not prohibits gatherings stats from multicast (or other type) interfaces,
I see no reasons to add hardcoded ignore at code level.

abays:
> As virDomainInterfaceStats itself not prohibits gatherings stats from multicast (or other type) interfaces,
I see no reasons to add hardcoded ignore at code level.

I agree: the ignore is the wrong solution in this case.  Let's forget about multicast and the type of the interface in general.  Nonetheless, the problem still remains that the use of ```virDomainInterfaceStats``` will generate errors for interfaces without a source, at least for some hypervisors.  If that's fine and the constant reporting of the error is just the cost of doing business (given the uncertainty of the underlying hypervisor and the libvirt implementation, and the fact that there might be stats somewhere), perhaps we should just close/resolve the issue?  Otherwise, I'm curious as to whether we might modify the plugin logic to anticipate problems with ```virDomainInterfaceStats``` or even try some other call in other contexts (perhaps, for example, ```virNetDevOpenvswitchInterfaceStats``` if we somehow knew the interface was an OVS interface on the netdev data path).

abays:
Requires more research to find a proper solution, if any.  Closing.

rpv-tomsk:
>Nonetheless, the problem still remains that the use of virDomainInterfaceStats will generate errors for interfaces without a source, at least for some hypervisors.

Unfortunately, we have `refresh_lists()` updates all domain structures, so we can't just set automatic flag to not gather more stats if first `virDomainInterfaceStats` fails.

Issue #2758, opened 2018-04-30
Compile Issue on AIX 7.1


devops75:
I am trying to compile the source code on AIX 7.1....I was able to execute the build.sh and configure.
But when i try to execute the make command i see the following error. Any help/suggestion/guidance is appreciated.

ash-4.4$ make
        make  all-am
  CC       src/libcollectdclient/libcollectdclient_la-client.lo
  CC       src/libcollectdclient/libcollectdclient_la-network.lo
  CC       src/libcollectdclient/libcollectdclient_la-network_buffer.lo
  CC       src/libcollectdclient/libcollectdclient_la-network_parse.lo
In file included from src/libcollectdclient/network_parse.c:48:0:
./src/stdendian.h:148:2: error: #error Could not determine CPU byte order
 #error Could not determine CPU byte order
  ^~~~~
make: 1254-004 The error code from the last command is 1.

rubenk:
Looks like we broke this recently. cc @dago.

Ramachillara:
I tried compiling the branch 5.8 and I was able to compile on AIX, earlier I was trying to compile from Master , thanks all for your replies

rubenk:
@Ramachillara can you try master again? I did merge a fix in https://github.com/collectd/collectd/commit/4ee5bff6ae7203e18e56cce9542891287f7f8b2b

rpv-tomsk:
This issue expected to be fixed.
Closed due to lack of response and no actions is pending.
Feel free to reopen if you have new questions or proposals. 
Thanks for using Collectd.

PR #2761, opened 2018-05-03
Add endianness checks for AIX

dago:
Please try this to resolve issue #2758


rubenk:
I have no AIX either, but this looks like it would work. Thanks @dago!

PR #2762, opened 2018-05-03
Set the thread name before detaching the thread

mkauf:
``pthread_setname_np()`` does not work reliably when a thread has been created in a detached state. Set the thread name first, and then call ``pthread_detach()`` afterwards (if necessary).

This prevents error messages like this one:
```
set_thread_name("unixsock conn"): No such process
```

ChangeLog: Fix a race condition when setting thread names


mkauf:
@kwiatrox I rebased on the current master, now the format is OK. Can you please approve once again, after the tests have run?

mkauf:
@mrunge You have merged some PRs lately, can you please merge this PR too? It's already approved.

PR #2768, opened 2018-05-10
Disable new gcc8 warnings

kraj:
GCC seems to be not able to detect the checks for size are
already in place

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/libcollectdclient/network_parse.c
+++ b/src/libcollectdclient/network_parse.c
@@ -149,6 +149,11 @@ static int parse_int(void *payload, size_t payload_size, uint64_t *out) {
   return 0;
 }
 
+#pragma GCC diagnostic push
+#if defined(__GNUC__) && (__GNUC__ >= 8)
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
+#endif
+
 static int parse_string(void *payload, size_t payload_size, char *out,
                         size_t out_size) {
   char *in = payload;
@@ -160,6 +165,8 @@ static int parse_string(void *payload, size_t payload_size, char *out,
   strncpy(out, in, out_size);
   return 0;
 }
+
+#pragma GCC diagnostic pop
 
 static int parse_identifier(uint16_t type, void *payload, size_t payload_size,
                             lcc_value_list_t *state) {
--- a/src/write_sensu.c
+++ b/src/write_sensu.c
@@ -570,6 +570,11 @@ static char *sensu_value_to_json(struct sensu_host const *host, /* {{{ */
   return ret_str;
 } /* }}} char *sensu_value_to_json */
 
+#pragma GCC diagnostic push
+#if defined(__GNUC__) && (__GNUC__ >= 8)
+#pragma GCC diagnostic ignored "-Wstringop-overflow"
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
+#endif
 /*
  * Uses replace_str2() implementation from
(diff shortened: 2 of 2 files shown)


rubenk:
My guess is you're referring to these warnings:
```
  CC       src/libcollectdclient/libcollectdclient_la-network_parse.lo
In function ‘parse_string’,
    inlined from ‘parse_identifier’ at src/libcollectdclient/network_parse.c:168:7,
    inlined from ‘network_parse’ at src/libcollectdclient/network_parse.c:539:11:
src/libcollectdclient/network_parse.c:160:3: warning: ‘strncpy’ specified bound 64 equals destination size [-Wstringop-truncation]
   strncpy(out, in, out_size);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

Are you sure GCC is wrong? I haven't looked at the code deeply yet,, but disabling warnings should be a last resort.

rubenk:
This is a real issue:
```
  CC       src/write_sensu.lo
src/write_sensu.c: In function ‘replace_str’:
src/write_sensu.c:630:3: warning: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
   strncpy(r, p, strlen(p));
   ^~~~~~~~~~~~~~~~~~~~~~~~
```

Should use the size of the destination buffer there, not the source.

kraj:
Yes, actually the out_size seems not correct to me, it probably should be size of in buffer that should be fed

rubenk:
I'm not going to apply this, we'd appreciate PR's that fixes these issues.

rpv-tomsk:
```
static int parse_string(void *payload, size_t payload_size, char *out,
                        size_t out_size) {
  char *in = payload;

  if ((payload_size < 1) || (in[payload_size - 1] != 0) ||
      (payload_size > out_size))
    return EINVAL;

  strncpy(out, in, out_size);
  return 0;
}
```

>Yes, actually the out_size seems not correct to me, it probably should be size of in buffer that should be fed

I see no error here. My `man strncpy` says:

>char *strncpy(char *restrict s1, const char *restrict s2, size_t n);
> The  strncpy()  function  shall copy **not more than n bytes** (**bytes that follow a null byte are not copied**) from the array pointed to by s2 to the array pointed to by s1.  
If the array pointed to by s2 is a string that is shorter than n bytes, null bytes shall be appended to the copy in the array pointed to by s1, until n bytes in all are written.

So, it is allowed for `in` to be shorter than `out_size`, and explicit check for null byte (if `in` has equal length with `out`) exists too. The `out_size` is used here explicitly too, for null bytes to be appended after copied data ('zero-padding').

From other side, I don't understand why this code needs two copy operations, first in `parse_string` and second in `parse_identifier`.

rpv-tomsk:
The explanation I found:

>If the bound (third parameter of strncpy) is equal to the destination size, then it means that the program won't be able to append a NULL terminator if the source string has a size of 256. If there is no space to put a NULL terminator, then you risk having buffer overflows depending if your code expects it or not.

rubenk:
> If the bound (third parameter of strncpy) is equal to the destination size, then it means that the program won't be able to append a NULL terminator if the source string has a size of 256. If there is no space to put a NULL terminator, then you risk having buffer overflows depending if your code expects it or not.

Yes but as you rightly point out there's a check for if the input string is null-terminated. Don't think gcc is smart enough to see that. Probably best to use sstrncpy, in which case the input check can also be dropped.

PR #2773, opened 2018-05-15
write_redis: fix "max_set_duration" deletes unexpected data

takahashi-tsc:
The calculation result of the data range
when "max_set_duration" is used is incorrect,
because the (char []) type, i.e. string, variable "time" is used
as the number for the time calculation.
As a result, data in the wrong range is deleted.
With this change, the correct data is deleted
by calculation using a double type value.

### DIFF
--- a/src/write_redis.c
+++ b/src/write_redis.c
@@ -129,10 +129,10 @@ static int wr_write(const data_set_t *ds, /* {{{ */
   if (node->max_set_duration > 0) {
     /*
      * remove element, scored less than 'current-max_set_duration'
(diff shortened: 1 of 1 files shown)


takahashi-tsc:
I think that the previous commit (14d5571) includes a little mistake.
I also commit the fix. (Should I split the pull request?)

rubenk:
Hi Takahashi, thank you for your contribution.
I just fixed Travis on master. Please rebase your pull request on b4a2f3eb.

takahashi-tsc:
Thank you for your quick response. I rebased my request.

PR #2774, opened 2018-05-15
Fix collecdmon not start collectd

takahashi-tsc:
Because collectdmon cannot exit parse command line options loop,
it cannot execute collectd start processing.


rubenk:
Thanks again @takahashi-tsc!

PR #2737, opened 2018-03-30
Merge collectd-5.7 to collectd-5.8

rpv-tomsk:
I'm unable to do this in generic way, so will try this one.

```
# git push upstream collectd-collectd-5.8:refs/heads/collectd-5.8
Username for 'https://github.com':
Password for 'https://github.com':
Counting objects: 22, done.
Delta compression using up to 2 threads.
Compressing objects: 100% (8/8), done.
Writing objects: 100% (8/8), 1.26 KiB, done.
Total 8 (delta 7), reused 0 (delta 0)
remote: Resolving deltas: 100% (7/7), completed with 7 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/collectd-5.8.
remote: error: Required status check "pull-requests-github_trigger-aggregation" is expected.
To https://github.com/collectd/collectd.git
 ! [remote rejected] collectd-collectd-5.8 -> collectd-5.8 (protected branch hook declined)
error: failed to push some refs to 'https://github.com/collectd/collectd.git'
```

PR #2776, opened 2018-05-15
Unblock #2737

mfournier:


rpv-tomsk:
Hi Marc!
Thanks for this fix. But what should be done with #2742? It also waits for `pull-requests-github_trigger-aggregation`.

MariuszX:
@rpv-tomsk @rubenk I`ve retriggered #2742 as suggested by @mfournier in #2737 by using  (git commit --amend --no-edit && git push -f ... ). And it`s finally "green".

rpv-tomsk:
Thanks for update, #2742 merged now.

PR #2789, opened 2018-05-21
redis: [impr/fix] ability to select db for queries

skob:
we need to query not default db in redis


rpv-tomsk:
Hi! 

For this to be merged, can you please fix/improve related documentation and configuration files?
(`collectd.conf.pod` and `collectd.conf.in`)

src/redis.c
```diff
@@ -196,6 +200,10 @@ static int redis_config_node(oconfig_item_t *ci) /* {{{ */
       status = cf_util_get_int(option, &timeout);
       if (status == 0)
         rn.timeout.tv_usec = timeout;
+    } else if (strcasecmp("Db", option->key) == 0) {
+      status = cf_util_get_int(option, &db);
+      if (status == 0)
+        rn.db = db;
```
rpv-tomsk: I think we don't need to use intermediate variable here and may store value to `rn->db` directly.
See other cases of `cf_util_get_int` in other plugins - most of them store values directly.
Also you can unify this way and fix `Timeout` option too.

src/redis.c
```diff
@@ -426,6 +435,12 @@ static int redis_read(void) /* {{{ */
       freeReplyObject(rr);
     }
 
+    if ((rr = redisCommand(rh, "SELECT %d", rn->db)) == NULL) {
```
rpv-tomsk: I think we have to skip this if `rn->db == 0`. 
https://redis.io/commands/select explicitly says what new connections always use the database 0 and we always have new connections here.
What do you think?

src/redis.c
```diff
@@ -426,6 +435,12 @@ static int redis_read(void) /* {{{ */
...
+      WARNING("redis plugin: unable to switch to db `%d' on node `%s'.", rn->db,
+              rn->name);
+      goto redis_fail;
+    }
+
     if ((rr = redisCommand(rh, "INFO")) == NULL) {
```
rpv-tomsk: I think, `INFO` command shows stats for entire Redis instance, not for selected database, right?
Then new option `Database` better fits to `Query` block. That will be more flexible to use.
What do you think?

src/collectd.conf.pod
```diff
@@ -7135,6 +7136,10 @@ read function is blocking, you should keep this value as low as possible. Keep
 in mind that the sum of all B<Timeout> values for all B<Nodes> should be lower
 than B<Interval> defined globally.
 
+=item B<Db> I<Db>
+
+Database to perform queries
```
rpv-tomsk: This explanation is not expanded enough. Use other options as examples.

rpv-tomsk:
Note: https://redis.io/commands/select

src/redis.c
```diff
@@ -140,11 +143,14 @@ static redis_query_t *redis_config_query(oconfig_item_t *ci) /* {{{ */
...
     oconfig_item_t *option = ci->children + i;
 
+    rq->db = REDIS_DEF_DB;
     if (strcasecmp("Type", option->key) == 0) {
       status = cf_util_get_string_buffer(option, rq->type, sizeof(rq->type));
     } else if (strcasecmp("Instance", option->key) == 0) {
       status =
           cf_util_get_string_buffer(option, rq->instance, sizeof(rq->instance));
+    } else if (strcasecmp("Db", option->key) == 0) {
+      status = cf_util_get_int(option, &rq->db);
```
rpv-tomsk: Sorry, but I think we need to check option value to be positive here (so intermediate value will be needed).
What do you think?

src/redis.c
```diff
@@ -140,11 +143,14 @@ static redis_query_t *redis_config_query(oconfig_item_t *ci) /* {{{ */
   for (int i = 0; i < ci->children_num; i++) {
     oconfig_item_t *option = ci->children + i;
 
+    rq->db = REDIS_DEF_DB;
```
rpv-tomsk: This is wrong. The value will be overwritten when next option after `Database` will be proceeded.

Also, `calloc` initializes all bytes to 0. Do we really need REDIS_DEF_DB?
If you think yes, then just put this initialization above the loop.

src/collectd.conf.pod
```diff
@@ -7140,6 +7141,10 @@ than B<Interval> defined globally.
 The B<Query> block identifies a query to execute against the redis server.
 There may be an arbitrary number of queries to execute.
 
+=item B<Db> I<Db>
```
rubenk: Database is more clear than Db, don't you think?

src/collectd.conf.pod
```diff
@@ -7140,6 +7141,10 @@ than B<Interval> defined globally.
...
+
+Database to perform query
```
rubenk: to perform query on

src/collectd.conf.pod
```diff
@@ -7136,15 +7136,15 @@ read function is blocking, you should keep this value as low as possible. Keep
...
-Database to perform queries
-
 =item B<Query> I<Querystring>
 
 The B<Query> block identifies a query to execute against the redis server.
 There may be an arbitrary number of queries to execute.
 
+=item B<Db> I<Db>
+
+Database to perform query
```
rpv-tomsk: Too short and is out of style.

src/collectd.conf.pod
rpv-tomsk: Agreed, we use full word in other places.

src/collectd.conf.pod
rpv-tomsk: Allows to select the Redis logical database by its zero-based numeric index I<Database>.
Database with index 0 is used by default.

src/redis.c
```diff
@@ -319,10 +317,12 @@ static int redis_handle_query(redisContext *rh, redis_node_t *rn,
     return -1;
   }
 
-  if ((rr = redisCommand(rh, "SELECT %d", rq->db)) == NULL) {
-    WARNING("redis plugin: unable to switch to db `%d' on node `%s'.", rq->db,
-            rn->name);
-    return -1;
+  if (rq->database > 0) {
```
rpv-tomsk: Ouch, sorry for new and new comments, but this rule (`database > 0`) becomes wrong when we put `Database` to `Query` level.
If you set `Database 0` in some `Query` then database index from previous `Query` will be used, what will be unexpected and wrong behaviour.

Everything else looks good, with one minor note: Maybe it will be good to put new option in examples as commented, to emphasize it is not mandatory. On your selection.

rpv-tomsk:
Can you squash your commits on your side, or we may do it ourselves before merge?

src/redis.c
```diff
@@ -140,11 +142,18 @@ static redis_query_t *redis_config_query(oconfig_item_t *ci) /* {{{ */
...
+    &rq->database = 0;
+
     if (strcasecmp("Type", option->key) == 0) {
       status = cf_util_get_string_buffer(option, rq->type, sizeof(rq->type));
     } else if (strcasecmp("Instance", option->key) == 0) {
       status =
           cf_util_get_string_buffer(option, rq->instance, sizeof(rq->instance));
+    } else if (strcasecmp("Database", option->key) == 0) {
+      status = cf_util_get_int(option, &rq->database);
+      if (option[0].value.number < 0) {
```
rpv-tomsk: 1) Use `rq->database` here.
2) Add `WARNING('Message')` explaining what is happening.

skob: you right :-(
but i change in redis_handle_query already.

if it's problem — i can re-edit again :)

skob:
@rpv-tomsk  , it's ok now?
please tell me when I can wait merge :)
thank you!

rpv-tomsk:
Hi!

The placement of `rq->database = 0;` is wrong. Please see comments.

rpv-tomsk:
Thanks!

rpv-tomsk:
JFYI, @skob  - there was two commits related to your work - they fix 2 variables size to avoid its overflow after `REDIS_DEF_DB_COUNT` changed to 256.

skob:
oops
sorry

it is ok now ?

rpv-tomsk:
Yes, it's ok now )

Issue #2713, opened 2018-03-15
ipmi plugin: new feature, add options to filter SEL events.


MariuszX:
According to the request from https://github.com/collectd/collectd/pull/2703 we started development of this feature.

Requirements:
- new option 'SELSensor' - to select source of events
- new option 'SELIgnoreSelected' - to ignore events from selected sensors

Functionality will be similar to existing options 'Sensor' and 'IgnoreSelected', which are used now for filtering metrics. New options will allow independent filtering of SEL events.

Should we consider any other requirements?

rpv-tomsk:
This feature was implemented and merged.

PR #2796, opened 2018-05-24
ipmi plugin: Add SELSensor and SELIgnoreSelected config options.

MariuszX:
As requested and discussed in #2703 and #2713
New options to allow independent filtering of SEL events.
Functionality is similar to existing options 'Sensor' and 'IgnoreSelected',
which are used now for filtering metrics. 

Change-Id: I0fde54a25577e61a4c90a4ff52f62117540a4343
Signed-off-by: Mariusz Szafranski <mariuszx.szafranski@intel.com>

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -689,6 +689,9 @@
 #		NotifySensorNotPresent false
 #		NotifyIPMIConnectionState false
 #		SELEnabled false
+#		SELSensor "some_sensor"
+#		SELSensor "another_one"
+#		SELIgnoreSelected false
 #		SELClearEvent false
 #	</Instance>
 #	<Instance "remote">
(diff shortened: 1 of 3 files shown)


MariuszX:
Sensor names are hardware dependent. Sensor list for SELSensor can be acquired same way as for Sensor options by disabling filtering and enabling ex. NotifySensorAdd option. Then all available sensor names will appear in collectd log when added.

rpv-tomsk:
Thanks for detailed explanation!

rpv-tomsk:
>Then all available sensor names will appear in collectd log when added.

Which line(s) of code will log this data?

MariuszX:
src/ipmi.c lines 511-520 via notifications when NotifySensorAdd is enabled.
Additionally in debug build src/ipmi.c lines 381-389

rpv-tomsk:
So, generic user have no option to get full list of sensor names in form ready to be used in `SelSensor`, right? This makes this option unuseable.

MariuszX:
Same as for existing Sensor option. Both options Sensor and SELSensor share one list of available sensor names.

rpv-tomsk:
Agreed, we have `INFO()` for discrete sensors there.
Maybe we have to add `INFO()` for each handled sensor...

rpv-tomsk:
Thanks for your work!

MariuszX:
Thx.

PR #2799, opened 2018-05-26
RFE: Add plugin name field into plugin context use it to improve error reporting 

rpv-tomsk:
Superseedes #2011

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -997,12 +997,13 @@ int plugin_load(char const *plugin_name, bool global) {
    * type when matching the filename */
   status = snprintf(typename, sizeof(typename), "%s.so", plugin_name);
   if ((status < 0) || ((size_t)status >= sizeof(typename))) {
-    WARNING("plugin_load: Filename too long: \"%s.so\"", plugin_name);
+    daemon_log(LOG_WARNING, "plugin_load: Filename too long: \"%s.so\"",
+               plugin_name);
     return -1;
   }
 
   if ((dh = opendir(dir)) == NULL) {
-    ERROR("plugin_load: opendir (%s) failed: %s", dir, STRERRNO);
+    daemon_log(LOG_ERR, "plugin_load: opendir (%s) failed: %s", dir, STRERRNO);
     return -1;
   }
 
@@ -1012,16 +1013,19 @@ int plugin_load(char const *plugin_name, bool global) {
 
     status = snprintf(filename, sizeof(filename), "%s/%s", dir, de->d_name);
     if ((status < 0) || ((size_t)status >= sizeof(filename))) {
-      WARNING("plugin_load: Filename too long: \"%s/%s\"", dir, de->d_name);
+      daemon_log(LOG_WARNING, "plugin_load: Filename too long: \"%s/%s\"", dir,
+                 de->d_name);
       continue;
     }
 
     if (lstat(filename, &statbuf) == -1) {
-      WARNING("plugin_load: stat (\"%s\") failed: %s", filename, STRERRNO);
+      daemon_log(LOG_WARNING, "plugin_load: stat (\"%s\") failed: %s", filename,
+                 STRERRNO);
       continue;
     } else if (!S_ISREG(statbuf.st_mode)) {
       /* don't follow symlinks */
-      WARNING("plugin_load: %s is not a regular file.", filename);
+      daemon_log(LOG_WARNING, "plugin_load: %s is not a regular file.",
+                 filename);
       continue;
     }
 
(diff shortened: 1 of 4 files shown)


rpv-tomsk:
I'm in progress of work on this enhancement.

First commit, which originally was written by @octo, rebased to current master.

After it was applied, Collectd starts to log following output:

```
[2018-05-26 20:26:27] logfile plugin: plugin_load: plugin "logfile" successfully loaded.
[2018-05-26 20:26:27] logfile plugin: logfile: invalid loglevel [debug] defaulting to 'info'
[2018-05-26 20:26:27] csv plugin: plugin_load: plugin "csv" successfully loaded.
[2018-05-26 20:26:27] network plugin: plugin_load: plugin "network" successfully loaded.
[2018-05-26 20:26:27] routeros plugin: plugin_load: plugin "routeros" successfully loaded.
```

To restore correct logging new function `daemon_log` was added (second commit).
Propose to discuss draft of this.

@rubenk, can you please dig into this feature a bit, to discuss core code changes?

Next commits should change plugins code - almost every file will be touched, so I want this work to be merged shortly to avoid multiple rebases with conflicts resolve.

rpv-tomsk:
Added `bind` plugin as example of cleanup.

Bind plugin, before unfication: different behaviour of `QTypes` and `ParseTime` options:

[2018-05-26 22:37:59] bind plugin: Using string value `falses' for boolean option `ParseTime' is deprecated and will be removed in future releases. Use unquoted true or false instead.
[2018-05-26 22:37:59] bind plugin: Cannot parse string value `falses' of the `ParseTime' option as a boolean value.
[2018-05-26 22:37:59] bind plugin: The `QTypes' option needs exactly one boolean argument.


And after:


[2018-05-26 22:45:06] bind plugin: Using string value `falses' for boolean option `ParseTime' is deprecated and will be removed in future releases. Use unquoted true or false instead.
[2018-05-26 22:45:06] bind plugin: Cannot parse string value `falses' of the `ParseTime' option as a boolean value.
[2018-05-26 22:45:06] bind plugin: Using string value `trues' for boolean option `QTypes' is deprecated and will be removed in future releases. Use unquoted true or false instead.
[2018-05-26 22:45:06] bind plugin: Cannot parse string value `trues' of the `QTypes' option as a boolean value.

src/daemon/plugin.c
```diff
@@ -2234,8 +2239,35 @@ void plugin_log(int level, const char *format, ...) {
...
     le = le->next;
   }
+} /* void _daemon_log */
+
+void plugin_log(int level, const char *format, ...) {
+  char msg[1024] = "";
+
+  char const *name = plugin_get_ctx().name;
+  if (name != NULL)
+    snprintf(msg, sizeof(msg), "%s plugin: ", name);
```
rubenk: please check for error / truncation.

src/daemon/plugin.c
```diff
@@ -2234,8 +2239,35 @@ void plugin_log(int level, const char *format, ...) {
...
+
+  va_list ap;
+  va_start(ap, format);
+  vsnprintf(msg + strlen(msg), sizeof(msg) - strlen(msg), format, ap);
```
rubenk: likewise

src/daemon/plugin.c
```diff
@@ -2234,8 +2239,35 @@ void plugin_log(int level, const char *format, ...) {
...
+  msg[sizeof(msg) - 1] = 0;
```
rubenk: not strictly needed if the vnsprintf call succeeds.

src/daemon/plugin.c
```diff
@@ -2234,8 +2239,35 @@ void plugin_log(int level, const char *format, ...) {
...
+  va_end(ap);
+
+  _daemon_log(level, msg);
 } /* void plugin_log */
 
+void daemon_log(int level, const char *format, ...) {
+  char msg[1024] = "";
```
rubenk: just a thought, but why limit the length to 1024 bytes?

src/daemon/plugin.c
```diff
@@ -346,19 +346,22 @@ static void log_list_callbacks(llist_t **list, /* {{{ */
 static int create_register_callback(llist_t **list, /* {{{ */
                                     const char *name, void *callback,
                                     user_data_t const *ud) {
-  callback_func_t *cf;
 
-  cf = calloc(1, sizeof(*cf));
+  if ((name == NULL) || (callback == NULL))
```
rubenk: no need for the inner braces.

rubenk:
Hi @rpv-tomsk. I left some minor remarks, but otherwise this looks like a great improvement, go for it!

rpv-tomsk:
>go for it!

Ok, just as a last question. 
Maybe a softer implementation should go: change only `configfile.c` messages, without touching `INFO() / WARNINIG() / ERROR()` macro?

src/daemon/plugin.c
rpv-tomsk: What should we do on error ?

src/daemon/plugin.c
rpv-tomsk: >"640k ought to be enough for anybody" 

I don't know... Historical reasons..
)

src/daemon/plugin.c
rubenk: either pass it up to the caller, or just log it.

src/daemon/plugin.c
rubenk: would be nice to allocate it dynamically using alloca or vla's, stacks are large enough these days.

rubenk:
> Maybe a softer implementation should go: change only configfile.c messages, without touching INFO() / WARNINIG() / ERROR() macro?

I'm fine with whatever you prefer.

rpv-tomsk:
Closed in favor of already merged #2821.

PR #2802, opened 2018-05-30
Fix potential memory leaks found via scan-build

ajssmith:


rubenk:
Thanks again @ajssmith!

PR #2803, opened 2018-06-01
cpufreq plugin: Read number of p-state transitions and time spent in each p-state

rorysexton:
Extending cpufreq plugin to read the following additional metrics:
  - number of p-state(cpu frequency) transitions
  - time spent in each p-state

Both metrics are reported by the Linux cpufreq-stats driver:
https://www.kernel.org/doc/Documentation/cpu-freq/cpufreq-stats.txt

### DIFF
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -33,36 +33,43 @@ struct cpu_data_t {
   value_to_rate_state_t time_state[MAX_AVAIL_FREQS];
 } * cpu_data;
 
-/* Flags denoting capability of reporting stats. */
-unsigned report_time_in_state, report_total_trans;
+/* Flags denoting capability of reporting CPU frequency statistics. */
+static bool report_p_stats = false;
 
-static int counter_init(void) {
+static void cpufreq_stats_init(void) {
   cpu_data = calloc(num_cpu, sizeof(struct cpu_data_t));
   if (cpu_data == NULL)
-    return 0;
+    return;
 
-  report_time_in_state = 1;
-  report_total_trans = 1;
+  report_p_stats = true;
 
   /* Check for stats module and disable if not present. */
   for (int i = 0; i < num_cpu; i++) {
     char filename[PATH_MAX];
-    int status = snprintf(filename, sizeof(filename),
-                      "/sys/devices/system/cpu/cpu%d/cpufreq/"
-                      "stats/time_in_state",
-                      num_cpu);
-    if ((status < 1) || ((unsigned int)status >= sizeof(filename)))
-      report_time_in_state = 0;
 
-    status = snprintf(filename, sizeof(filename),
-                      "/sys/devices/system/cpu/cpu%d/cpufreq/"
-                      "stats/total_transitions",
-                      num_cpu);
-    if ((status < 1) || ((unsigned int)status >= sizeof(filename)))
-      report_total_trans = 0;
+    snprintf(filename, sizeof(filename),
+             "/sys/devices/system/cpu/cpu%d/cpufreq/stats/time_in_state", i);
+    if (access(filename, R_OK)) {
+      NOTICE("cpufreq plugin: File %s not exists or no access. P-State "
+             "statistics will not be reported. Check if `cpufreq-stats' kernel "
+             "module is loaded.",
+             filename);
+      report_p_stats = false;
+      break;
+    }
+
+    snprintf(filename, sizeof(filename),
+             "/sys/devices/system/cpu/cpu%d/cpufreq/stats/total_trans", i);
+    if (access(filename, R_OK)) {
+      NOTICE("cpufreq plugin: File %s not exists or no access. P-State "
+             "statistics will not be reported. Check if `cpufreq-stats' kernel "
+             "module is loaded.",
+             filename);
+      report_p_stats = false;
+      break;
     }
   }
-  return 0;
+  return;
 }
 
 static int cpufreq_init(void) {
@@ -72,9 +79,9 @@ static int cpufreq_init(void) {
 
   while (1) {
     int status = snprintf(filename, sizeof(filename),
-                      "/sys/devices/system/cpu/cpu%d/cpufreq/"
-                      "scaling_cur_freq",
-                      num_cpu);
+                          "/sys/devices/system/cpu/cpu%d/cpufreq/"
+                          "scaling_cur_freq",
+                          num_cpu);
     if ((status < 1) || ((unsigned int)status >= sizeof(filename)))
       break;
 
@@ -85,7 +92,7 @@ static int cpufreq_init(void) {
   }
 
   INFO("cpufreq plugin: Found %d CPU%s", num_cpu, (num_cpu == 1) ? "" : "s");
-  counter_init();
+  cpufreq_stats_init();
 
   if (num_cpu == 0)
     plugin_unregister_read("cpufreq");
@@ -127,49 +134,59 @@ static int cpufreq_read(void) {
 
     cpufreq_submit(i, "cpufreq", NULL, &v);
 
-    /* Read total transitions for cpu frequency */
-    if (report_total_trans) {
+    if (report_p_stats) {
+      /* Read total transitions for cpu frequency */
       snprintf(filename, sizeof(filename),
                "/sys/devices/system/cpu/cpu%d/cpufreq/stats/total_trans", i);
       if (parse_value_file(filename, &v, DS_TYPE_COUNTER) != 0) {
-        WARNING("cpufreq plugin: Reading \"%s\" failed.", filename);
+        ERROR("cpufreq plugin: Reading \"%s\" failed.", filename);
         continue;
       }
       cpufreq_submit(i, "counter", "transitions", &v);
-    }
 
-    /*
-     * Determine percentage time in each state for cpu during previous interval.
-     */
-    if (report_time_in_state) {
-      int j = 0;
-      cdtime_t now = cdtime();
-      char state[DATA_MAX_NAME_LEN];
-
+      /* Determine percentage time in each state for cpu during previous
+       * interval. */
       snprintf(filename, sizeof(filename),
                "/sys/devices/system/cpu/cpu%d/cpufreq/stats/time_in_state", i);
 
       FILE *fh = fopen(filename, "r");
-      if (fh == NULL)
+      if (fh == NULL) {
+        ERROR("cpufreq plugin: Reading \"%s\" failed.", filename);
         continue;
+      }
 
-      char buffer[DATA_MAX_NAME_LEN];  //128
+      int j = 0;
+      cdtime_t now = cdtime();
+      char buffer[DATA_MAX_NAME_LEN];
 
       while (fgets(buffer, sizeof(buffer), fh) != NULL) {
         unsigned long long time;
+        char state[DATA_MAX_NAME_LEN];
+
+        /*
+         * State time units is 10ms. To get rate of seconds per second
+         * we have to divide by 100. To get percents we have to multiply it
+         * by 100 back. So, just use parsed value directly.
+         */
         if (!sscanf(buffer, "%s%llu", state, &time)) {
-          WARNING("cpufreq plugin: Reading \"%s\" failed.", filename);
-          fclose(fh);
-          return 0;
+          ERROR("cpufreq plugin: Reading \"%s\" failed.", filename);
+          break;
         }
 
         if (j < MAX_AVAIL_FREQS) {
           gauge_t g;
-          if (value_to_rate(&g, (value_t){.counter = time}, DS_TYPE_COUNTER, now, &cpu_data[i].time_state[j]) != 0) {
-            continue;
+          if (value_to_rate(&g, (value_t){.counter = time}, DS_TYPE_COUNTER,
+                            now, &(cpu_data[i].time_state[j])) != 0) {
             j++;
+            continue;
           }
           cpufreq_submit(i, "percent", state, &(value_t){.gauge = g});
+        } else {
+          NOTICE("cpufreq plugin: Found too much frequency states (%d > %d). "
+                 "Plugin needs to be recompiled. Please open a bug report for "
+                 "this.",
+                 (j + 1), MAX_AVAIL_FREQS);
+          break;
         }
         j++;
       }
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -117,11 +117,11 @@ static void cpufreq_submit(int cpu_num, const char *type,
 }
 
 static int cpufreq_read(void) {
-  for (int i = 0; i < num_cpu; i++) {
+  for (int cpu = 0; cpu < num_cpu; cpu++) {
     char filename[PATH_MAX];
     /* Read cpu frequency */
     snprintf(filename, sizeof(filename),
-             "/sys/devices/system/cpu/cpu%d/cpufreq/scaling_cur_freq", i);
+             "/sys/devices/system/cpu/cpu%d/cpufreq/scaling_cur_freq", cpu);
 
     value_t v;
     if (parse_value_file(filename, &v, DS_TYPE_GAUGE) != 0) {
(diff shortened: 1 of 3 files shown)


src/cpufreq.c
```diff
@@ -82,9 +154,64 @@ static int cpufreq_read(void) {
...
+        continue;
+      }
+      fclose(fh);
+
+      cpufreq_submit(i, "transitions", NULL, v);
+    }
 
+    /* Determine time in state for cpu during previous interval
+     * Reported in 10mS as unit.
+     */
```
rpv-tomsk: If unit is 10ms, maybe we have to multiply value to get value in milliseconds?
Similar is done with frequency (kHz converted to Hz).
This is preferred to use.

src/cpufreq.c
```diff
@@ -25,8 +25,71 @@
 #include "common.h"
 #include "plugin.h"
 
+#define MAX_AVAIL_FREQS 20
```
rpv-tomsk: Why exactly this value is used?

src/cpufreq.c
```diff
@@ -25,8 +25,71 @@
...
+      report_time_in_state = 0;
+      return 0;
+    }
+    while (fgets(buffer, sizeof(buffer), fh) != NULL) {
+      if (!sscanf(buffer, "%s%lli", state, &t)) {
+        fclose(fh);
+        return 0;
+      }
+      t_data[i].time_prev[j] = t;
+      j++;
```
rpv-tomsk: This cycle does not check for MAX_AVAIL_FREQS limit, so value can exit over array bounds.

src/cpufreq.c
```diff
@@ -82,9 +154,64 @@ static int cpufreq_read(void) {
...
+               "/sys/devices/system/cpu/cpu%d/cpufreq/stats/total_trans", i);
+      fh = fopen(filename, "r");
+      if (fh == NULL)
+        continue;
+      while (fgets(buffer, sizeof(buffer), fh) != NULL) {
+        if (!sscanf(buffer, "%lli", &t)) {
+          fclose(fh);
+          return 0;
+        }
+        snprintf(buffer, sizeof(buffer), "%lli", t - t_data[i].transitions);
```
rpv-tomsk: I think you have to use DERIVE type for metrics.
So, delta will be calcultated inside of Collectd core.

Also, there is standard way of file parsing exists, without `snprintf('%lli')` and such ... .

This needs to be rewritten.
The same is for `time_in_state` metric.

rpv-tomsk:
Hi, @rorysexton.

Thanks for your work on Collectd. I left some remarks, code should be reworked to gets merged into upstream. If you want to continue this work feel free to ask questions.

src/cpufreq.c
rorysexton: fixed in next revision of patch

src/cpufreq.c
rorysexton: its used to create an array of frequencies that core can spend time in. Must be fixed size as its an element of another array allocated using calloc.

src/cpufreq.c
rorysexton: fixed in next revision of patch

src/cpufreq.c
rorysexton: Based on this comment I have made a couple of changes. 

transitions type is now COUNTER type and diff is calculated inside collectd core. It also uses the standard file parsing based on comment above.

For time_in_state metric parsing is a little more complex as the file is formatted like so:
2401000 335156
2400000 187794999
2300000 1
2200000 393122
Hence I need to parse this file manually to pull out the metrics. 
As per comment I've changed it to calculate the diff inside collectd core and report the percentage time in each frequency state using standard value_to_rate().

src/cpufreq.c
```diff
@@ -82,9 +142,58 @@ static int cpufreq_read(void) {
...
+
+    /* Read total transitions for cpu frequency */
+    if (report_total_trans) {
+      snprintf(filename, sizeof(filename),
+               "/sys/devices/system/cpu/cpu%d/cpufreq/stats/total_trans", i);
+      if (parse_value_file(filename, &v, DS_TYPE_COUNTER) != 0) {
+        WARNING("cpufreq plugin: Reading \"%s\" failed.", filename);
+        continue;
+      }
+      counter_t c = counter_diff(t_data[i].transition_prev, v.counter);
```
rpv-tomsk: IIRC, you haven't use `counter_diff` here, just do `cpufreq_submit` and all should work because `transitions` metric has `DERIVE` type.

src/cpufreq.c
```diff
@@ -82,9 +142,58 @@ static int cpufreq_read(void) {
...
+        }
+        cdtime_t now = cdtime();
+        gauge_t g;
+        if (j < MAX_AVAIL_FREQS) {
+          if (value_to_rate(&g, v, DS_TYPE_DERIVE, now,
+                            &t_data[i].time_state[j]) != 0) {
+            continue;
+            j++;
+          }
+          cpufreq_submit(i, "percent", state, &(value_t){.gauge = g});
```
rpv-tomsk: Will this really be `percent`  - value in 0-100 range?

src/cpufreq.c
```diff
@@ -82,9 +142,58 @@ static int cpufreq_read(void) {
...
+               "/sys/devices/system/cpu/cpu%d/cpufreq/stats/time_in_state", i);
+      fh = fopen(filename, "r");
+      if (fh == NULL)
+        continue;
+      while (fgets(buffer, sizeof(buffer), fh) != NULL) {
+        if (!sscanf(buffer, "%s%lli", state, &t)) {
+          fclose(fh);
+          return 0;
+        }
+        snprintf(time, sizeof(time), "%lli", t);
```
rpv-tomsk: IMO, this is overengineered solution.  Just assign `long long t`  to `v.derive`, if `value_to_rate` is really needed.

src/cpufreq.c
rpv-tomsk: Why exactly 20 ? What this value is based on?

src/cpufreq.c
```diff
@@ -25,8 +25,59 @@
...
+
+  /* Check for stats module and disable if not present. */
+  for (int i = 0; i < num_cpu; i++) {
+    char filename[256];
+    int status;
+    status = snprintf(filename, sizeof(filename),
+                      "/sys/devices/system/cpu/cpu%d/cpufreq/"
+                      "stats/time_in_state",
+                      num_cpu);
+    if ((status < 1) || ((unsigned int)status >= sizeof(filename)))
```
rpv-tomsk: >Check for stats module and disable if not present. 

Unfortunately, this check is not enough to check module presence.

rpv-tomsk:
@octo, I'm going to merge this. Can you please check spelling?
Thanks.

src/cpufreq.c
```diff
@@ -82,9 +132,67 @@ static int cpufreq_read(void) {
...
+      while (fgets(buffer, sizeof(buffer), fh) != NULL) {
+        unsigned long long time;
+        char state[DATA_MAX_NAME_LEN];
+
+        /*
+         * State time units is 10ms. To get rate of seconds per second
+         * we have to divide by 100. To get percents we have to multiply it
+         * by 100 back. So, just use parsed value directly.
+         */
+        if (!sscanf(buffer, "%s%llu", state, &time)) {
```
octo: *sscanf()* doesn't know the size of `state` and is prone to buffer overflows. Please don't use it.

src/cpufreq.c
```diff
@@ -82,9 +132,67 @@ static int cpufreq_read(void) {
...
+        if (j < MAX_AVAIL_FREQS) {
+          gauge_t g;
+          if (value_to_rate(&g, (value_t){.counter = time}, DS_TYPE_COUNTER,
+                            now, &(cpu_data[i].time_state[j])) != 0) {
+            j++;
+            continue;
+          }
+          cpufreq_submit(i, "percent", state, &(value_t){.gauge = g});
+        } else {
+          NOTICE("cpufreq plugin: Found too much frequency states (%d > %d). "
```
octo: Please indent error flow (see description above).

In this particular case:

```c
if (j >= MAX_AVAIL_FREQS) {
  NOTICE(…);
  break;
}
/* code for j < MAX_AVAIL_FREQS */
```

src/cpufreq.c
octo: I think so: the comment above says this is in 10ms increments, 10ms/1s is essentially a percentage.

src/cpufreq.c
```diff
@@ -82,9 +132,67 @@ static int cpufreq_read(void) {
...
+      snprintf(filename, sizeof(filename),
+               "/sys/devices/system/cpu/cpu%d/cpufreq/stats/time_in_state", i);
+
+      FILE *fh = fopen(filename, "r");
+      if (fh == NULL) {
+        ERROR("cpufreq plugin: Reading \"%s\" failed.", filename);
+        continue;
+      }
+
+      int j = 0;
```
octo: Name suggestion: `state_index`

src/cpufreq.c
```diff
@@ -82,9 +132,67 @@ static int cpufreq_read(void) {
     /* convert kHz to Hz */
     v.gauge *= 1000.0;
 
-    cpufreq_submit(i, v);
+    cpufreq_submit(i, "cpufreq", NULL, &v);
+
+    if (report_p_stats) {
```
octo: Please indent error flow; in other words, handle errors. Example:

```c
/* GOOD */
if (!looks_good) {
  handle_error();
  return -1;
}
rest_of_code();
```

instead of:

```c
/* BAD */
if (looks_good) {
  rest_of_code();
} else {
  handle_error();
  return -1;
}
```

In this particular case:

```c
if (!report_p_states) {
  continue;
}
/* rest of code */
```

(this definitely needs to go into the review comments doc)

src/cpufreq.c
rpv-tomsk: Replaced %s with %u.

rpv-tomsk:
Thanks for review.
PR updated, changes are verified, plugin is working.

src/cpufreq.c
```diff
@@ -168,27 +170,27 @@ static int cpufreq_read(void) {
...
-        if (j < MAX_AVAIL_FREQS) {
-          gauge_t g;
-          if (value_to_rate(&g, (value_t){.counter = time}, DS_TYPE_COUNTER,
-                            now, &(cpu_data[i].time_state[j])) != 0) {
-            j++;
-            continue;
-          }
-          cpufreq_submit(i, "percent", state, &(value_t){.gauge = g});
-        } else {
+        snprintf(state, sizeof(state), "%u", frequency);
```
octo: Please declare `state` just before this line.

src/cpufreq.c
```diff
@@ -132,34 +132,36 @@ static int cpufreq_read(void) {
...
       /* Read total transitions for cpu frequency */
       snprintf(filename, sizeof(filename),
-               "/sys/devices/system/cpu/cpu%d/cpufreq/stats/total_trans", i);
+               "/sys/devices/system/cpu/cpu%d/cpufreq/stats/total_trans", cpu);
       if (parse_value_file(filename, &v, DS_TYPE_COUNTER) != 0) {
         ERROR("cpufreq plugin: Reading \"%s\" failed.", filename);
         continue;
       }
-      cpufreq_submit(i, "counter", "transitions", &v);
+      cpufreq_submit(cpu, "counter", "transitions", &v);
```
octo: Please don't use the `counter` type. Maybe define a new `DERIVE` type called "transitions".

src/cpufreq.c
```diff
@@ -132,34 +132,36 @@ static int cpufreq_read(void) {
...
 
-    cpufreq_submit(i, "cpufreq", NULL, &v);
+    cpufreq_submit(cpu, "cpufreq", NULL, &v);
 
     if (report_p_stats) {
       /* Read total transitions for cpu frequency */
       snprintf(filename, sizeof(filename),
-               "/sys/devices/system/cpu/cpu%d/cpufreq/stats/total_trans", i);
+               "/sys/devices/system/cpu/cpu%d/cpufreq/stats/total_trans", cpu);
       if (parse_value_file(filename, &v, DS_TYPE_COUNTER) != 0) {
```
octo: Please use `DS_TYPE_DERIVE` instead of counter.

src/cpufreq.c
octo: This is not yet addressed.

src/cpufreq.c
rpv-tomsk: Why not? "counter" type is used widely enough.

Also, there is small problem with `DERIVE` type. Derive type is signed int64_t, while value in file (as provided by kernel) is unsigned uint64_t.  How to deal?

src/cpufreq.c
rpv-tomsk: The same is here, as described above.

Number of frequency transitions is a _counter_.

File `linux-3.2.78/drivers/cpufreq/cpufreq_stats.c` has following structure:
```
struct cpufreq_stats {
    unsigned int cpu;
    unsigned int total_trans;
    unsigned long long  last_time;
    unsigned int max_state;
    unsigned int state_num;
    unsigned int last_index;
    cputime64_t *time_in_state;
    unsigned int *freq_table;
#ifdef CONFIG_CPU_FREQ_STAT_DETAILS
    <------>unsigned int *trans_table;
#endif
};
```

When use DS_TYPE_DERIVE, `parse_value()` handles value as signed.

src/cpufreq.c
rpv-tomsk: Disagreed, that is not error flow. Imagine we have another `report_other_metrics` variable and another report.... 

If existing code is unacceptable, then we need to move this block of code into own function.
I this will be better solution.

What do you think?

src/cpufreq.c
rpv-tomsk: Updated.

src/cpufreq.c
octo: > Why not? "counter" type is used widely enough.

No, it's not:

```
cheery:collectd octo$ egrep -o 'ABSOLUTE|COUNTER|DERIVE|GAUGE' src/types.db | sort | uniq -c
   1 ABSOLUTE
  12 COUNTER
 154 DERIVE
 163 GAUGE
cheery:collectd octo$ grep COUNTER src/types.db
counter                 value:COUNTER:U:U
arc_counts              demand_data:COUNTER:0:U, demand_metadata:COUNTER:0:U, prefetch_data:COUNTER:0:U, prefetch_metadata:COUNTER:0:U
arc_l2_bytes            read:COUNTER:0:U, write:COUNTER:0:U
mysql_qcache            hits:COUNTER:0:U, inserts:COUNTER:0:U, not_cached:COUNTER:0:U, lowmem_prunes:COUNTER:0:U, queries_in_cache:GAUGE:0:U
mysql_threads           running:GAUGE:0:U, connected:GAUGE:0:U, cached:GAUGE:0:U, created:COUNTER:0:U
```

As you can see, with the exception of the `counter` type, all uses of `COUNTER` are legacy types that are no longer in active use.

> Also, there is small problem with `DERIVE` type. Derive type is signed int64_t, while value in file (as provided by kernel) is unsigned uint64_t. How to deal?

Ignore it, just cast to `derive_t`. This metrics increases by 100 every second, meaning it will overflow a signed integer after *(2^63 - 1) / 100* seconds – 2.9 billion years.

src/cpufreq.c
octo: That's fine, 2.9 billion years is well outside of my life-span so I'm not going to care ;)

src/cpufreq.c
octo: Even if it's not error flow: indent the short path, keep the longer / normal path at previous indentation level.

For a more in-depth discussion, see https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

rpv-tomsk:
PR updated

src/cpufreq.c
```diff
@@ -116,6 +116,69 @@ static void cpufreq_submit(int cpu_num, const char *type,
...
+    if (!sscanf(buffer, "%u%llu", &frequency, &time)) {
+      ERROR("cpufreq plugin: Reading \"%s\" failed.", filename);
+      break;
+    }
+
+    char state[DATA_MAX_NAME_LEN];
+    snprintf(state, sizeof(state), "%u", frequency);
+
+    if (state_index >= MAX_AVAIL_FREQS) {
+      NOTICE("cpufreq plugin: Found too much frequency states (%d > %d). "
```
octo: Nit, at your option: "found too **many** frequency states"

Issue #2805, opened 2018-06-04
virt: allows to build HosthameFormat with information from libvirt metadata


sileht:
Hi,

Some application that creates libvirt virtual machine, adds some metadata [1] about the instances.

Currently we can only use the "name" or "uuid". It would be nice to use any information from metadata section.

For example, Openstack nova generate this kind of metadata: 

```
  <metadata>
    <nova:instance xmlns:nova="http://openstack.org/nova/instance/1">
      <nova:package version="2014.2.3"/>
      <nova:name>demo1vm</nova:name>
        ....
    </nova:instance>
  </metadata>
```

Add I would like to get the `<nova:instance><nova:name>` attribute into `HostnameFormat`.

Does a contribution that allow this will be accepted ?

Like a new choice for `HostnameFormat` called `metadata` and a new option called `HostnameMetadataXPath` with xpath notation to get the attribute like: 
`/nova:instance/nova:name` ?

Regards,
sileht

[1] https://libvirt.org/formatdomain.html#elementsMetadata

atyronesmith:
Would something like the following be acceptable?

<Plugin "virt">
  Connection "qemu:///system"
  RefreshInterval 3600
  ExtraStats "cpu_util disk_err domain_state fs_info job_stats_background perf vcpupin"
  HostnameFormat "name friendly_name hostname"
</Plugin>

will produce...

"instance-000036e0" "test-gnocchi-3" "compute-01"

sileht:
@atyronesmith  `friendly_name` meaning is not clear, this doesn't mean anything for libvirt. (While I known it's the Openstack nova display_name in our context), so 'openstack' looks clearer for me, but that very specific.

My proposal is more generic and will work with any libvirt upper layer application that use metadata system of libvirt, it will looks like:
```
<Plugin "virt">
HostnameFormat "name metadata hostname"
HostnameMetadataXPath "/nova:instance/nova:name"
</Plugin>
```

And will produce the same output:
`instance-000036e0 test-gnocchi-3 compute-01`

I'm OK with the generic or the specific version.

sileht:
@atyronesmith, yes, we could do something to extract multiple metadata. But remember that in collectd `hostname` is always truncated to 62 chars, so can't really put many of them.

atyronesmith:
Yes. Names may be truncated...

How about going with your suggestion with
HostnameMetadataXPath an optional parameter 
defaulting to  "/nova:instance/nova:name"?

PR #2807, opened 2018-06-06
virt: allow read Hostname from libvirt metadata

sileht:
This change allows to read Hostname from Libvirt metadata API.

Applications like OVirt or Openstack Nova uses this metadata API to
store additional informations like the 'user facing' Guest name.

To do so, a new choice 'metadata' is added for 'HostnameFormat' and
'PluginInstanceFormat'. And two new options are also added to localize
the hostname into the Guest metadata: HostnameMetadataNS and
HostnameMetadataXPath.

Closes: #2805

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -9287,7 +9287,7 @@ be set to C<var_lib_libvirt_images_image1.qcow2>.
 Setting C<BlockDeviceFormatBasename true> will cause the I<type instance> to be
 set to C<image1.qcow2>.
 
-=item B<HostnameFormat> B<name|uuid|hostname|...>
+=item B<HostnameFormat> B<name|uuid|hostname|metadata...>
 
 When the virt plugin logs data, it sets the hostname of the collected data
 according to this setting. The default is to use the guest name as provided by
@@ -9299,6 +9299,9 @@ same guest across migrations.
 B<hostname> means to use the global B<Hostname> setting, which is probably not
 useful on its own because all guests will appear to have the same name. This is
 useful in conjunction with B<PluginInstanceFormat> though.
+
+B<metadata> means use information from guest's metadata. Use
+B<HostnameMetadataNS> and B<HostnameMetadataXPath> to localize this information.
 
 You can also specify combinations of these fields. For example B<name uuid>
 means to concatenate the guest name and UUID (with a literal colon character
@@ -9318,17 +9321,30 @@ setting B<name>.
 B<address> means use the interface's mac address. This is useful since the
 interface path might change between reboots of a guest or across migrations.
 
-=item B<PluginInstanceFormat> B<name|uuid|none>
+=item B<PluginInstanceFormat> B<name|uuid|metadata|none>
 
 When the virt plugin logs data, it sets the plugin_instance of the collected
 data according to this setting. The default is to not set the plugin_instance.
 
 B<name> means use the guest's name as provided by the hypervisor.
 B<uuid> means use the guest's UUID.
+B<metadata> means use information from guest's metadata.
 
 You can also specify combinations of the B<name> and B<uuid> fields.
 For example B<name uuid> means to concatenate the guest name and UUID
 (with a literal colon character between, thus I<"foo:1234-1234-1234-1234">).
+
+=item B<HostnameMetadataNS> B<string>
+
+When B<metadata> is used in B<HostnameFormat> or B<PluginInstanceFormat>, this
+selects in which metadata namespace we will pick the hostname. The default is
+I<http://openstack.org/xmlns/libvirt/nova/1.0>.
+
+=item B<HostnameMetadataXPath> B<string>
+
+When B<metadata> is used in B<HostnameFormat> or B<PluginInstanceFormat>, this
+describes where the hostname is located in the libvirt metadata. The default is
+I</instance/name/text()>.
 
 =item B<Instances> B<integer>
 
--- a/src/virt.c
+++ b/src/virt.c
@@ -734,75 +734,75 @@ char *metadata_get_hostname(virDomainPtr dom) {
       dom, VIR_DOMAIN_METADATA_ELEMENT, namespace, VIR_DOMAIN_AFFECT_CURRENT);
   if (metadata_str == NULL) {
     return NULL;
-  } else {
-    char *hostname = NULL;
-    xmlDocPtr xml_doc = NULL;
-    xmlXPathContextPtr xpath_ctx = NULL;
-    xmlXPathObjectPtr xpath_obj = NULL;
-    xmlNodePtr xml_node = NULL;
-
-    xml_doc = xmlReadDoc((xmlChar *)metadata_str, NULL, NULL, XML_PARSE_NONET);
-    if (xml_doc == NULL) {
-      ERROR(PLUGIN_NAME " plugin: xmlReadDoc failed to read metadata");
-      goto metadata_end;
-    }
-
-    xpath_ctx = xmlXPathNewContext(xml_doc);
-    if (xpath_ctx == NULL) {
-      ERROR(PLUGIN_NAME " plugin: xmlXPathNewContext(%s) failed for metadata",
-            metadata_str);
-      goto metadata_end;
-    }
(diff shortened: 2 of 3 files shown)


src/virt.c
```diff
@@ -761,6 +772,94 @@ static int lv_domain_info(virDomainPtr dom, struct lv_info *info) {
...
+  xmlXPathContextPtr xpath_ctx = NULL;
+  xmlXPathObjectPtr xpath_obj = NULL;
+
+  if (hm_xpath == NULL)
+    xpath_str = "/instance/name/text()";
+  else
+    xpath_str = hm_xpath;
+
+  if (hm_ns == NULL)
+  namespace = "http://openstack.org/xmlns/libvirt/nova/1.0";
```
anaudx: indentation is missing

src/virt.c
```diff
@@ -761,6 +772,94 @@ static int lv_domain_info(virDomainPtr dom, struct lv_info *info) {
...
+  else namespace = hm_ns;
```
anaudx: split into 2 lines

sileht: clang-format don't like that, but I have surround the code with {} to make clang-format happy and have a well indented code.

src/virt.c
sileht: clang-format don't like that, but I have surround the code with {} to make clang-format happy and have a well indented code.

sileht:
Hi, @leifmadsen, can you take a look to this PR ?

src/collectd.conf.pod
```diff
@@ -9076,18 +9079,31 @@ setting B<name>.
...
 
-=item B<PluginInstanceFormat> B<name|uuid|none>
+=item B<PluginInstanceFormat> B<name|uuid|metadata|none>
 
 When the virt plugin logs data, it sets the plugin_instance of the collected
 data according to this setting. The default is to not set the plugin_instance.
 
 B<name> means use the guest's name as provided by the hypervisor.
 B<uuid> means use the guest's UUID.
+B<metadata> means use an information from guest's metadata.
```
atyronesmith: use an information -> use information

sileht: done

abays:
Tried this out.  Cool functionality!

One question though: is it normal for the VM name to be used as the "host" in the Collectd value list or notification (regardless of which HostnameFormat option one uses)?  It seems like this should be the actual host itself, such as "compute-0" if you were running this on, say, an Openstack compute node in order to monitor Nova instances.  That being said, it looks like the virt plugin has always behaved this way.  Just curious as to why it seems to break the pattern that other plugins use.

Thanks for doing this, sileht!

sileht:
Yes, the virt plugin have always behaved this way. The main reason is to send metrics as if collectd was installed inside the VM but it's not. 

But if you really want  the host like other plugins, setting `HostnameFormat` to `hostname` should work.

src/virt.c
```diff
@@ -761,6 +772,96 @@ static int lv_domain_info(virDomainPtr dom, struct lv_info *info) {
...
+    } else if (xml_node->type == XML_ATTRIBUTE_NODE) {
+      hostname_src = (const char *)xml_node->children->content;
+    } else {
+      ERROR(PLUGIN_NAME " plugin: xmlXPathEval(%s) unsupported node type %d",
+            xpath_str, xml_node->type);
+      goto metadata_end;
+    }
+
+    hostname_len = strlen(hostname_src) + 1;
+    hostname = malloc(hostname_len);
```
rpv-tomsk: Why not just use `strdup()`  ?

rpv-tomsk:
Hi, @sileht !

Did you plan to continue this work? I have proposal regarding implementation: why we need to call metadata_get_hostname() for each init_value_list() call? 

I think we may optimize code a bit: add `plugin_instance` and `hostname` into `domain_t` and fill it only while `refresh_lists()` ?

What do you think about such rework? I can't test this well.

cc @abays

rpv-tomsk:
>From my experience, the hostname could change while the VM is running, let's make sure this is taken into account.  E.g. if we use the hostname to report some metric, the metric may change name.

Thanks for this notice!

By default, `refresh_lists()` called once per 60 seconds. That can be set by `RefreshInterval` option.
What do you think, that is acceptable delay between hostname change and metric name being changed? 

We also can update `hostname` and `plugin_instance` once per read cycle, not for each `init_value_list`.

src/virt.c
```diff
@@ -761,6 +772,96 @@ static int lv_domain_info(virDomainPtr dom, struct lv_info *info) {
...
+  } else {
+    namespace = hm_ns;
+  }
+
+  metadata_str = virDomainGetMetadata(dom, VIR_DOMAIN_METADATA_ELEMENT,
+                                      namespace, VIR_DOMAIN_AFFECT_CURRENT);
+  if (metadata_str) {
+    xml_doc = xmlReadDoc((xmlChar *)metadata_str, NULL, NULL, XML_PARSE_NONET);
+    if (xml_doc == NULL) {
+      ERROR(PLUGIN_NAME "plugin: xmlReadDoc failed to read metadata");
```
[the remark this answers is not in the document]
sileht: done

src/virt.c
```diff
@@ -761,6 +772,96 @@ static int lv_domain_info(virDomainPtr dom, struct lv_info *info) {
   return 0;
 }
 
+char *metadata_get_hostname(virDomainPtr dom) {
+  char *hostname = NULL;
```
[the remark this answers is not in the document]
sileht: done

src/virt.c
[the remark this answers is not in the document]
sileht: done

src/virt.c
```diff
@@ -761,6 +772,96 @@ static int lv_domain_info(virDomainPtr dom, struct lv_info *info) {
...
+
+  if (hm_ns == NULL) {
+    namespace = "http://openstack.org/xmlns/libvirt/nova/1.0";
+  } else {
+    namespace = hm_ns;
+  }
+
+  metadata_str = virDomainGetMetadata(dom, VIR_DOMAIN_METADATA_ELEMENT,
+                                      namespace, VIR_DOMAIN_AFFECT_CURRENT);
+  if (metadata_str) {
```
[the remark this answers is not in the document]
sileht: This is expected, I have added the error case to make it explicit.

src/virt.c
```diff
@@ -761,6 +772,96 @@ static int lv_domain_info(virDomainPtr dom, struct lv_info *info) {
...
+    xml_doc = xmlReadDoc((xmlChar *)metadata_str, NULL, NULL, XML_PARSE_NONET);
+    if (xml_doc == NULL) {
+      ERROR(PLUGIN_NAME "plugin: xmlReadDoc failed to read metadata");
+      goto metadata_end;
+    }
+    xpath_ctx = xmlXPathNewContext(xml_doc);
```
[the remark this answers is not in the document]
sileht: done

src/virt.c
```diff
@@ -761,6 +772,96 @@ static int lv_domain_info(virDomainPtr dom, struct lv_info *info) {
...
+    if (xpath_obj->type != XPATH_NODESET) {
+      ERROR(PLUGIN_NAME " plugin: xmlXPathEval(%s) unexpected return type %d "
+                        "(wanted %d) for metadata",
+            xpath_str, xpath_obj->type, XPATH_NODESET);
+      goto metadata_end;
+    }
+
+    // TODO(sileht): We can support || operator by looping on nodes here
+    if (xpath_obj->nodesetval == NULL || xpath_obj->nodesetval->nodeNr != 1) {
+      DEBUG(PLUGIN_NAME " plugin: xmlXPathEval(%s) return nodeset size=%i "
```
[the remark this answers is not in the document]
sileht: I changed it to WARNING. As this is not a collected error, but the xml attribute that the user have configured can't be found, so warn it.

src/virt.c
sileht: fixed.

src/virt.c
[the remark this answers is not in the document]
rpv-tomsk: >Please indent error flow; in other words, handle errors. Example:

That was not fixed.

rpv-tomsk:

PR #2810, opened 2018-06-08
Base port to Windows

SeanCampbell:


Makefile.am
```diff
@@ -228,7 +254,7 @@ collectd_SOURCES = \
 
 
 collectd_CFLAGS = $(AM_CFLAGS)
-collectd_CPPFLAGS = $(AM_CPPFLAGS)
+collectd_CPPFLAGS = $(AM_CPPFLAGS) $(LTDLINCL)
```
rubenk: what does this variable do? if it's for libltdl, we don't use that anymore.

src/daemon/common.c
```diff
@@ -267,7 +263,12 @@ int swrite(int fd, const void *buf, size_t count) {
   pfd.revents = 0;
   if (poll(&pfd, 1, 0) > 0) {
     char buffer[32];
-    if (recv(fd, buffer, sizeof(buffer), MSG_PEEK | MSG_DONTWAIT) == 0) {
+    if (recv(fd, buffer, sizeof(buffer),
+             MSG_PEEK
+#ifndef WIN32
+                 | MSG_DONTWAIT
```
rubenk: this looks a bit ugly, how about defining MSG_DONTWAIT to 0 for win32 like
```
#ifdef WIN32
#define MSG_DONTWAIT 0
#endif
```
and then keeping the recv call unchanged?

src/daemon/plugin.h
```diff
@@ -34,8 +34,13 @@
 #include "meta_data.h"
 #include "utils_time.h"
 
+#include <inttypes.h>
 #include <pthread.h>
 
+#ifdef WIN32
+#undef ERROR
```
rubenk: can you add a comment as to why you do this?

src/libcollectdclient/client.c
```diff
@@ -24,6 +24,14 @@
  *   Florian octo Forster <octo at collectd.org>
  **/
 
+#ifdef WIN32
+#include "gnulib_config.h"
+#include <winsock2.h>
+#else
+#include <sys/socket.h>
+#include <sys/un.h>
```
rubenk: moving these includes up will probably break some platform like OpenBSD. See 34471c08182f3ea9f977d8cea1e8934e59e01ef4 for an example.

src/libcollectdclient/client.c
```diff
@@ -452,9 +463,11 @@ static int lcc_open_netsocket(lcc_connection_t *c, /* {{{ */
     }
   }
 
-  struct addrinfo ai_hints = {.ai_family = AF_UNSPEC,
-                              .ai_flags = AI_ADDRCONFIG,
-                              .ai_socktype = SOCK_STREAM};
+  struct addrinfo ai_hints = {
+      .ai_family = AF_UNSPEC, .ai_flags = 0, .ai_socktype = SOCK_STREAM};
+#ifdef AI_ADDRCONFIG
+  ai_hints.ai_flags = AI_ADDRCONFIG;
```
rubenk: could you define AI_ADDRCONFIG to 0 if not previously defined? that would clean this up nicely.

src/libcollectdclient/collectd/network.h
```diff
@@ -60,7 +64,7 @@ int lcc_server_destroy(lcc_network_t *net, lcc_server_t *srv);
 
 /* Configure servers */
 int lcc_server_set_ttl(lcc_server_t *srv, uint8_t ttl);
-int lcc_server_set_interface(lcc_server_t *srv, char const *interface);
+int lcc_server_set_interface(lcc_server_t *srv, char const *interface_);
```
rubenk: I take it `interface` shadows something? How about calling it iface instead?

build.sh
```diff
@@ -18,34 +18,159 @@ EOF
...
-set -x
+    export CC="x86_64-w64-mingw32-gcc"
+
+    mkdir -p _build_aux
+
+    cd _build_aux
+    if [ -d "_gnulib" ]; then
+        echo "Assuming that gnulib is already built, because _gnulib exists."
+    else
+        git clone git://git.savannah.gnu.org/gnulib.git
```
rubenk: Ouch, cloning a git repo into a git repo, I'm not sure how I feel about this yet ;-) Won't git submodules work?

build.sh
```diff
@@ -18,34 +18,159 @@ EOF
...
+          time_r \
+          sys_stat \
+          fcntl-h \
+          sys_resource \
+          sys_wait \
+          setlocale \
+          strtok_r \
+          poll \
+          recv \
+          net_if
```
rubenk: could you sort this list?

build.sh
```diff
@@ -18,34 +18,159 @@ EOF
...
+        rm libgnu.a # get rid of it, to use libgnu.dll
+    fi
+    cd "${TOP_SRCDIR}"
+
+    set -x
+    autoheader \
+    && aclocal -I /usr/share/aclocal \
+    && $libtoolize --copy --force \
+    && automake --add-missing --copy \
+    && autoconf
```
rubenk: is there no autoreconf on cygwin that can replace the above 5 lines?

configure.ac
```diff
@@ -99,6 +99,10 @@ case $host_os in
     AC_DEFINE([KERNEL_SOLARIS], [1], [True if program is to be compiled for a Solaris kernel])
     ac_system="Solaris"
     ;;
+  *mingw32*)
+    AC_DEFINE([KERNEL_WIN32], [1], [True if program is to be compiled for a Windows kernel])
```
rubenk: this define looks unused in this PR, is that correct?

src/daemon/cmd_windows.c
```diff
@@ -0,0 +1,42 @@
...
+#include <stdio.h>
+#include <windows.h>
+
+int main(int argc, char **argv) {
+  WORD wVersionRequested;
+  WSADATA wsaData;
+  int err;
+  wVersionRequested = MAKEWORD(2, 2);
+
+  err = WSAStartup(wVersionRequested, &wsaData);
```
rubenk: could you combine this with line 31? We generally prefer C99-style declarations for newer code.

src/daemon/cmd_windows.c
```diff
@@ -0,0 +1,42 @@
...
+  if (err != 0) {
+    printf("WSAStartup failed with error: %d\n", err);
```
rubenk: does windows have a stderr to log this to?

build.sh
SeanCampbell: Hm, I hadn't heard of that before. I'll check this out.

src/libcollectdclient/collectd/network.h
SeanCampbell: Done.

build.sh
SeanCampbell: I was following how it builds it for other platforms, but autoreconf seems to work just as well and is simpler :)

configure.ac
SeanCampbell: Correct, it's not used in this PR, but I use it in some plugins I ported (in follow-up PRs). Would you prefer I wait to define it until it's used?

src/daemon/cmd_windows.c
SeanCampbell: Done.

src/daemon/cmd_windows.c
SeanCampbell: Yeah, this seems to work fine using ERROR as is the case with other platforms.

Makefile.am
SeanCampbell: Yup, I copied this over from an old commit and didn't realize it wasn't needed anymore. Removed.

src/daemon/common.c
SeanCampbell: Done.

src/daemon/plugin.h
SeanCampbell: Done.

src/libcollectdclient/client.c
SeanCampbell: Ah ok. Moved them down. gnulib_config.h needs to be before config.h, so I left the Windows includes at the top.

src/libcollectdclient/client.c
SeanCampbell: Done.

build.sh
rubenk: Ah indeed! We should probably just do that for all platforms then. Care to do a separate PR for that?

configure.ac
rubenk: Thanks for the explanation. No problem, just leave it in.

src/daemon/plugin.h
```diff
@@ -34,8 +34,14 @@
 #include "meta_data.h"
 #include "utils_time.h"
 
+#include <inttypes.h>
 #include <pthread.h>
 
+#ifdef WIN32
+// ERROR is previously defined in wingdi.h, included from windows.h.
```
rubenk: I do recall from a previous life that there was something like -DWIN32_LEAN_AND_MEAN which reduced the number of windows headers included. Would that help?

rubenk:
Thanks for the quick turnaround on the changes Sean. I left a few more minor comments but otherwise it looks good from my point of view.

I just have a general question, do you have any thoughts on how we can make sure we don't break Windows builds? We'll need to have some kind of CI.

build.sh
```diff
@@ -18,34 +18,155 @@ EOF
...
+    cp "${MINGW_ROOT}/bin/zlib1.dll" "${INSTALL_DIR}"
+    cp "${MINGW_ROOT}/bin/libwinpthread-1.dll" "${INSTALL_DIR}"
+    cp "${MINGW_ROOT}/bin/libdl.dll" "${INSTALL_DIR}"
+
+    echo "Done"
+}
+
+os_name="$(uname)"
+if test ${os_name:0:6} = "CYGWIN"; then
+    build_cygwin
```
rubenk: CI breaks on this:
```
+ checkbashisms -n build.sh
possible bashism in build.sh line 167 (${foo:3[:1]}):
if test ${os_name:0:6} = "CYGWIN"; then
+ exit 1
```

You'll need to use POSIX constructs, for example `if test "${string#*$substring}" != "$string"`

build.sh
SeanCampbell: That works really nicely here; updated. Thanks for pointing this out.

rubenk:
Perhaps @mfournier has some ideas as well, he's more familiar with our CI infra.

rubenk:
Just curious, did you manage to build these changes with something like mingw on Fedora, or only on cygwin?

SeanCampbell:
I've only built it on Cygwin at this point.

build.sh
```diff
@@ -18,34 +18,152 @@ EOF
...
+          canonicalize-lgpl \
+          fcntl-h \
+          getsockopt \
+          gettimeofday \
+          nanosleep \
+          netdb \
+          net_if \
+          poll \
+          recv \
+              regex \
```
rubenk: indentation is off

configure.ac
```diff
@@ -848,11 +854,17 @@ AC_CHECK_FUNCS([socket],
   [
     AC_CHECK_LIB([socket], [socket],
       [socket_needs_socket="yes"],
-      [AC_MSG_ERROR([cannot find socket() in libsocket])]
+      [
+        AC_CHECK_LIB(gnu, rpl_socket,
```
rubenk: please quote the arguments with []

rubenk:
@SeanCampbell could you add a few lines of documentation on how to build this for Windows? Not sure what's the best location, probably the README.

configure.ac
SeanCampbell: Done.

build.sh
SeanCampbell: Done.

SeanCampbell:
Sure thing, added that in.

README
```diff
@@ -1040,6 +1040,39 @@ To generate the `configure` script, you'll need the following dependencies:
...
+variable:
+
+$ export INSTALL_DIR="C:/some/other/install/directory"
+$ ./build.sh
+
+or:
+
+$ INSTALL_DIR="C:/some/other/install/directory" ./build.sh
+
+As of now, the only plugin tested to work on Windows is the `logfile' plugin.
```
tokkee: This is useful information but can easily go out of sync. We're also not yet tracking it for other OSes, but maybe we should. I'm wondering if we can have a good way to track and manage this information.

build.sh
```diff
@@ -18,34 +18,152 @@ EOF
...
+    : ${SBINDIR:="${INSTALL_DIR}"}
+    : ${SYSCONFDIR:="${INSTALL_DIR}"}
+    : ${LOCALSTATEDIR:="${INSTALL_DIR}"}
+    : ${DATAROOTDIR:="${INSTALL_DIR}"}
+    : ${DATADIR:="${INSTALL_DIR}"}
 
-set -x
+    echo "Installing collectd to ${INSTALL_DIR}."
+    TOP_SRCDIR=$(pwd)
+    MINGW_ROOT="/usr/x86_64-w64-mingw32/sys-root/mingw"
```
tokkee: Is this path name valid universally? If not, can it be auto-detected?

configure.ac
```diff
@@ -888,7 +900,12 @@ AC_CHECK_FUNCS([nanosleep], [],
     [
       AC_CHECK_LIB([posix4], [nanosleep],
         [nanosleep_needs_posix4="yes"],
-        [AC_MSG_ERROR([cannot find nanosleep])]
+        [
+          AC_CHECK_FUNCS([rpl_nanosleep],
+            [nanosleep_needs_gnulib="yes"],
```
tokkee: This seems to be unused. Do we need it?

src/daemon/common.c
```diff
@@ -69,11 +65,15 @@ extern kstat_ctl_t *kc;
 #endif
 
 /* AIX doesn't have MSG_DONTWAIT */
-#ifndef MSG_DONTWAIT
+#if !defined(MSG_DONTWAIT) && defined(MSG_NONBLOCK)
 #define MSG_DONTWAIT MSG_NONBLOCK
 #endif
 
-#if !HAVE_GETPWNAM_R
+#ifdef WIN32
```
tokkee: Nit: Putting this into an else branch seems cleaner.

tokkee:
:+1: from me. @rubenk do you want to merge once you're happy?

src/daemon/common.c
SeanCampbell: Done.

build.sh
SeanCampbell: Yep, looks like you can detect it.

configure.ac
SeanCampbell: I think this is leftover from changes I patched in from kele's port, so it's not needed anymore.

README
SeanCampbell: Removed.

SeanCampbell:
@rubenk Do you think any other changes are needed?

tokkee:
Since there were no further comments, I'll merge now.

Thank you for your work @SeanCampbell!

SeanCampbell:
Hi Pavel,

I opened a PR with the fix here: https://github.com/collectd/collectd/pull/2907

rpv-tomsk:
@SeanCampbell 

Hi, Sean!
I added some commits to fix some build issues, they revert some changes which was added in this PR.
Please check them and propose other solutions if they are needed.

Thanks!

PR #2821, opened 2018-06-16
RFE: Add plugin name field into plugin context use it to improve error reporting

rpv-tomsk:
Superseedes #2011 
Superseedes #2799

This PR is the next attempt to implement this enhancement.
Unlike previous, this one does not try to touch each plugin. 
Instead, only `configfile.c` functions are affected.

I think this is a shortest way to reach results.

cc @rubenk 

Ruben, what do you think?

### DIFF
--- a/src/daemon/plugin.h
+++ b/src/daemon/plugin.h
@@ -171,6 +171,7 @@ struct user_data_s {
 typedef struct user_data_s user_data_t;
 
 struct plugin_ctx_s {
+  char *name;
   cdtime_t interval;
   cdtime_t flush_interval;
   cdtime_t flush_timeout;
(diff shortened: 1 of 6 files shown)


rpv-tomsk:
TODO: look into `src/daemon/common.c`: `service_name_to_port_number()`, `parse_value()` and so on.

rpv-tomsk:
cc  @rubenk 

Ruben, what do you think about this PR in overall?
About functions/macro naming?

Is this good for go in ?

src/daemon/plugin.h
```diff
@@ -398,6 +399,15 @@ int parse_notif_severity(const char *severity);
 #define DEBUG(...) /* noop */
 #endif             /* ! COLLECT_DEBUG */
 
+/* These will log messages, prefixed by plugin name */
```
rubenk: s/These/This/

rubenk:
I left one minor remark, but otherwise LGTM!

PR #2830, opened 2018-06-18
Add Bind Address option

elieyal:
This change adds the option to do a Bind Address to an IP address instead of to an Interface Name only.
Now you can add the option for examples:
BindAddress 10.10.10.1
to the server section and it will bind to the IP address.


elieyal:
Will change the code and re-open.

PR #2827, opened 2018-06-18
Update network.c - Add Bind Address option

elieyal:
This allows the option to add a Bind Address to an IP address instead of the binding to the Interface Name

Issue #2818, opened 2018-06-13
Network plugin not sending data via specific interface


elieyal:
*  Version of collectd: 5.8.0-3
*  Operating system / distribution: CentOS 7.3.1611
*  On the other side there is an InfluxDB listening to UDP port 80

## Expected behavior

I have several servers that acts the same. all have:
eth0 is an internal NIC - not connected to the internet.
eth1 another internal NIC - not connected to the internet
eth2 is an external NIC - has access to the internet
lo

when the collectd start with no specific interface configured it selects the eth0. Although there is no connection to the internet, with tcpdump, I can see the packets being sent.
When redirecting the network plugin to a specific network interface and looking at the tcpdump I would see the packets being sent as well

## Actual behavior

In this case, I don't see ANY packets going out of the configured interface (or any of the interfaces)
I have tested with traceroute that there is a connection between the 2 servers (run traceroute on the collectd server and tcpdump to listen on the receiving interface where the influxdb is installed - I can see the packets arriving so it is not a FW or network issue). The packets are not even going out TO the interface.

## Steps to reproduce
*  step 1
Configure collectd.conf:

```
<Plugin network>
    <Server "influxdb" "80">
        Interface "eth2"
    </Server>
</Plugin>
```
* step 2
    Restart collectd

* step 3
    run tcpdump:
    tcpdump -i eth2 -p -n -s 1500 udp port 80

I have to say that I have another server with the same configuration (a copy with same configuration) and it works as expected. This is why this problem is a great mystery.

What I would like to have:

1. A bind-address capability instead of the interface name. it might resolve some issues
2. How can I compile Collectd myself with debug capabilities? I haven't found anything in the Readme's or on any website.

rpv-tomsk:
Hi!

>I have tested with traceroute that there is a connection between the 2 servers (run traceroute on the collectd server and tcpdump to listen on the receiving interface where the influxdb is installed - I can see the packets arriving so it is not a FW or network issue). The packets are not even going out TO the interface.

If you have correct routing between servers, then all should work without `Interface` specified.

>    A bind-address capability instead of the interface name. it might resolve some issues

Patches are welcome!

>    How can I compile Collectd myself with debug capabilities? I haven't found anything in the Readme's or on any website.

`./configure --enable-debug`

elieyal:
Ok, I managed to compile the source with debug and got the debug working.
With some network/system guys I managed to understand the issue with binding to the interface name instead of the IP address:
[...]

PR #2831, opened 2018-06-18
network plugin: Add BindAddress option

elieyal:
This change allows to Bind Address to an IP address instead of the binding to the Interface Name in the server section.
Example:
```
<Server "239.192.74.66" "25826">
                SecurityLevel Encrypt
                Username "user"
                Password "secret"
                BindAddress "10.9.25.59"
                ResolveInterval 14400
        </Server>
```
Credit to: Ofir Hermesh

Superseedes:  #2827
Issue: #2818


rpv-tomsk:
Please do not close issues, remove branches and so on. 
This makes review process much harder. 

Please look into #2827 for review comments. Please do updates here, but if you want to discuss/comment changes - do it in #2827.

If you have questions, feel free to ask.

Thanks for your work on Collectd project.

rpv-tomsk:
Hi!

Will you continue the work on this feature? If yes, when do you plan to do this?

elieyal:
Yes, I'm working on it, this is super important for me (and maybe others).
I'll be submitting a fix for all the comments (thank you for that) in a short while. I'm currently testing that everything will be working OK :-)

elieyal:
New commit was uploaded. Please review and comment.

src/network.c
```diff
@@ -2684,6 +2724,47 @@ static int network_config_set_interface(const oconfig_item_t *ci, /* {{{ */
...
+  *bind_address = malloc(sizeof(**bind_address));
+  (*bind_address)->ss_family = res->ai_family;
+  if (res->ai_family == AF_INET) {
+    struct sockaddr_in *addr = (struct sockaddr_in *)(*bind_address);
+    inet_pton(AF_INET, addr_text, &(addr->sin_addr));
+  } else if (res->ai_family == AF_INET6) {
+    struct sockaddr_in6 *addr = (struct sockaddr_in6 *)(*bind_address);
+    inet_pton(AF_INET6, addr_text, &(addr->sin6_addr));
+  } else {
+    ERROR("%s is an unknown address format %d\n", addr_text, res->ai_family);
```
rpv-tomsk: We need to inform upper level about error state.
`sfree(*bind_address); return -1; ` are needed.

src/network.c
```diff
@@ -1683,6 +1685,42 @@ static int network_set_interface(const sockent_t *se,
...
+    return 0;
+
+  DEBUG("fd %i: bind socket to address", se->data.client.fd);
+  char pbuffer[64];
+
+  if (ai->ai_family == AF_INET) {
+    struct sockaddr_in *addr =
+        (struct sockaddr_in *)(se->data.client.bind_addr);
+    inet_ntop(AF_INET, &(addr->sin_addr), pbuffer, 64);
+    DEBUG("binding client socket to ipv4 address: %s", pbuffer);
```
rpv-tomsk: Please add `network plugin:` prefix for all DEBUG/INFO/ERROR messages, here and below.

rpv-tomsk:
Hi!

Thanks for new commit. There is some new remarks added about the code.

elieyal:
I have committed a fix for the 2 remarks, I hope this will be sufficient :-)
please review and comment.
*Again, all credits goes to Ofir Hermesh for his great help on this.

src/network.c
```diff
@@ -2684,6 +2724,51 @@ static int network_config_set_interface(const oconfig_item_t *ci, /* {{{ */
...
+    return -1;
+  }
+
+  char addr_text[256];
+
+  if (cf_util_get_string_buffer(ci, addr_text, sizeof(addr_text)) != 0)
+    return -1;
+
+  int ret;
+  struct addrinfo hint, *res = NULL;
```
rubenk: please use designated initializers to initialize hint

src/network.c
```diff
@@ -2684,6 +2724,51 @@ static int network_config_set_interface(const oconfig_item_t *ci, /* {{{ */
...
+    struct sockaddr_in6 *addr = (struct sockaddr_in6 *)(*bind_address);
+    inet_pton(AF_INET6, addr_text, &(addr->sin6_addr));
+  } else {
+    ERROR("network plugin: %s is an unknown address format %d\n", addr_text,
+          res->ai_family);
+    sfree(*bind_address);
+    return -1;
+  }
+
+  return 0;
```
rubenk: you leak res, please call freeaddrinfo on it.

elieyal:
Rubenk, Thank you for the comments. I have fixed the issues you commented on.
Please approve this change.

src/network.c
```diff
@@ -2684,6 +2724,54 @@ static int network_config_set_interface(const oconfig_item_t *ci, /* {{{ */
...
+  struct addrinfo hint = {.ai_family = PF_UNSPEC,
+                          .ai_flags = AI_NUMERICHOST,
+                          .ai_protocol = IPPROTO_IP,
+                          .ai_socktype = SOCK_DGRAM};
+
+  ret = getaddrinfo(addr_text, NULL, &hint, &res);
+  if (ret) {
+    ERROR("network plugin: Bind address option has invalid address set: %s",
+          gai_strerror(ret));
+    freeaddrinfo(res);
```
rubenk: I'm not sure this is correct. Posix doesn't say anything about what res points to on failure (http://pubs.opengroup.org/onlinepubs/9699919799/functions/freeaddrinfo.html). That means res can still be NULL, and I don't think you can call `freeaddrinfo` with a NULL pointer. On the other hand, there's no guarantee that res is NULL on failure and you don't have to freeaddrinfo on it.

Let's just do what the rest of the code in our codebase does and remove this call to `freeaddrinfo`.

src/network.c
```diff
@@ -1683,6 +1685,42 @@ static int network_set_interface(const sockent_t *se,
...
+    if (bind(se->data.client.fd, (struct sockaddr *)addr, sizeof(*addr)) ==
+        -1) {
+      ERROR("network_plugin: failed to bind client socket (ipv4): %s",
+            STRERRNO);
+      return -1;
+    }
+  } else if (ai->ai_family == AF_INET6) {
+    struct sockaddr_in6 *addr =
+        (struct sockaddr_in6 *)(se->data.client.bind_addr);
+    inet_ntop(AF_INET, &(addr->sin6_addr), pbuffer, 64);
```
rubenk: This should be AF_INET6 I presume?

src/network.c
```diff
@@ -2684,6 +2724,54 @@ static int network_config_set_interface(const oconfig_item_t *ci, /* {{{ */
...
+  char addr_text[256];
+
+  if (cf_util_get_string_buffer(ci, addr_text, sizeof(addr_text)) != 0)
+    return -1;
+
+  int ret;
+  struct addrinfo *res = NULL;
+  struct addrinfo hint = {.ai_family = PF_UNSPEC,
+                          .ai_flags = AI_NUMERICHOST,
+                          .ai_protocol = IPPROTO_IP,
```
rubenk: IPPROTO_UDP seems more appropriate since that's what collectd uses.

src/network.c
```diff
@@ -2684,6 +2724,54 @@ static int network_config_set_interface(const oconfig_item_t *ci, /* {{{ */
...
+  }
+
+  char addr_text[256];
+
+  if (cf_util_get_string_buffer(ci, addr_text, sizeof(addr_text)) != 0)
+    return -1;
+
+  int ret;
+  struct addrinfo *res = NULL;
+  struct addrinfo hint = {.ai_family = PF_UNSPEC,
```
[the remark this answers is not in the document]
rubenk: I'll leave it up to you, but we call this struct ai_hints or hints elsewhere in the codebase

src/network.c
```diff
@@ -2684,6 +2724,54 @@ static int network_config_set_interface(const oconfig_item_t *ci, /* {{{ */
...
+
+  ret = getaddrinfo(addr_text, NULL, &hint, &res);
+  if (ret) {
+    ERROR("network plugin: Bind address option has invalid address set: %s",
+          gai_strerror(ret));
+    freeaddrinfo(res);
+    return -1;
+  }
+
+  *bind_address = malloc(sizeof(**bind_address));
```
rubenk: check malloc call for failure please

rubenk:
Thanks for your work on this @elieyal. I left some more comments, but it's starting to look good. One thing that's still missing is documentation, please document the new option in the manpage too.

elieyal:
Sure thing @rubenk. But are we all in agreement on these changes? Are these all the comments. 
Me and my college are trying our best but doing this small function back and forth... we will change anything you request to meet your standards but please let’s do this efficient. 
Thank you.

elieyal:
@rubenk, #2827 is closed.

rubenk:
@elieyal it's a small function but in the core of the daemon, so it's important to get it right. I have no further comments however.

elieyal:
@rubenk it is also important to me to make it good do we are on the same page. All comments are taken very seriously and will be fixed tomorrow morning. 
@rpv-tomsk any other comments from your side?

elieyal:
Hi,
I have committed all the requested changes.
@rubenk, please provide more details which file is to be modified for the manpage as I would like to update it with this new option as you requested (as it should be).

rubenk:
https://github.com/collectd/collectd/blob/master/src/collectd.conf.pod

elieyal:
Documentation also added.

src/collectd.conf.pod
```diff
@@ -5438,6 +5438,17 @@ behavior is to let the kernel choose the appropriate interface. Be warned
 that the manual selection of an interface for unicast traffic is only
 necessary in rare cases.
 
+=item B<BindAddress> I<IP Address>
+
+Set the outgoing IP address for IP packets. This option can be used instead of
+the I<Interface> option to explicitly define the IP address which will be used
+to send Packets to the remote server. This applies at least
+to IPv6 packets and if possible to IPv4. If this option is not applicable,
```
rubenk: > This applies at least to IPv6 packets and if possible to IPv4

I don't understand this sentence. Why would this not apply to IPv4?

src/collectd.conf.pod
```diff
@@ -5438,6 +5438,17 @@ behavior is to let the kernel choose the appropriate interface. Be warned
...
+undefined or a non-existent interface name is specified, the default
```
rubenk: I think that's too vague. What does it mean to not be applicable? I suggest something like 'if this option is not set'

src/collectd.conf.pod
elieyal: Since the BindAddress is the same as the Interface option, I did a copy/paste from the Interface option and modify it according to the usage (IP address instead of Interface). The explanation is the same so I left it as the Interface description. I don't know why the Interface option would not apply to IPv4 as it surely does but since they both do the same (except the type of bind) and someone wrote it for a reason, I didn't want to change it,

src/collectd.conf.pod
elieyal: This is copy/paste from the Interface Option and it acts exactly the same.
it's either the Interface option or BindAddress option or none. Since this is the description in the Interface option I left it as is because of the same functionality.

elieyal: You are correct about the Interface rather than the IP but since the default settings will be the interface as it was before the change the documentation as is, is correct. The default is binding to an interface selected by the kernel as it was before.

elieyal: You are correct about the Interface rather than the IP but since the default settings will be the interface as it was before the change the documentation as is, is correct. The default is binding to an interface selected by the kernel as it was before.

rpv-tomsk: >The default is binding to an interface selected by the kernel as it was before.

That is not true, IMHO. There is no _binding_ by default, so if you change routing after Collectd start, packets will flow by new routing rules.

https://linux.die.net/man/7/socket : 

>If the name is an empty string or the option length is zero, the socket device binding is removed.

So, there is no binding by default, as I understand.

elieyal: Ok. I will drop the lines below the third sentence

elieyal:
Changed descriptions as commented.

rpv-tomsk:
Just a note (for myself) `https://github.com/redis/hiredis/commit/37d25a392c9b9468e064a67c504939c9c4ea0031`

src/network.c
```diff
@@ -1683,6 +1685,42 @@ static int network_set_interface(const sockent_t *se,
...
 } /* }}} network_set_interface */
 
+static int network_bind_socket_to_addr(sockent_t *se,
+                                       const struct addrinfo *ai) {
+
+  if (se->data.client.bind_addr == NULL)
+    return 0;
+
+  DEBUG("network_plugin: fd %i: bind socket to address", se->data.client.fd);
+  char pbuffer[64];
```
rpv-tomsk: This variable currently used in `DEBUG()` call.
Then Collectd compiled w/o debug, this variable will be unused at practice - it will be filled by `inet_ntop()`, but not used later.

Mine proposal is to put it into `ERROR()` calls:

`ERROR("network plugin: failed to bind client socket (ipvX) to %s: %s", pbuffer, STRERRNO);`

This will use variable when non-debug version is compiled, and message become more informative.

elieyal: If you think that this is needed in non debug mode than I’ll change it.

elieyal:
Made the change, the format is not correct, will fix it in latertoday and do another commit.

elieyal:
@rpv-tomsk Once i'll finish with the debug to error in the last comment,
will it be released as a patch to 5.8.0? Will it be possible to have it release next week?

rpv-tomsk:
>will it be released as a patch to 5.8.0? 

Project policy allows only bugfix patches. This one is not a bugfix.
This is a feature and will be released in 5.9.

>Will it be possible to have it release next week?

:-) Subscribe to #2835 for news.

---
Your improvement was merged to 'master' branch.
Thanks for your work on Collectd project!

Issue #2219, opened 2017-03-15
collectd-tg producing unrealistic timestamps


bzed:
*   Version of collectd: 5.7.1-1.1 
*   Operating system / distribution: Debian testing

I'm running collectd-tg into a collectd which does nothing but receiving data via the network plugin and writing into a graphite port. Looking at the ngrep output gives me

```
collectd.conova.host0000.plugin042.derive-ti513721137.value 0.400000 147787.
collectd.conova.host0000.plugin008.gauge-ti1295265221.value 97.456085588783 147787.
collectd.conova.host0000.plugin021.derive-ti378479591.value 3.300000 147787.
collectd.conova.host0000.plugin021.gauge-ti1942227332.value 10.9819253906608 147787.
collectd.conova.host0000.plugin032.derive-ti445843840.value 6.100000 147787.
collectd.conova.host0000.plugin051.gauge-ti871939962.value 99.6843919623643 147787.
collectd.conova.host0000.plugin003.gauge-ti524540180.value 89.0014075208455 147787.
```

As we are doing rollups depending on the timestamp in the database, I can't use collectd-tg to load test the DB as 99% of the received lines will be dropped instantly..... 147787 is way too long ago... A good start would be a now() or just a few days ago.

faxm0dem:
I just stumbled on this while trying to stress-test riemann, and indeed the timestamp are epoch-1's

![collectd-tg-epoch](http://i.imgur.com/DA4ZIkG.png)

abays:
Pull request with fix can be found here:

https://github.com/collectd/collectd/pull/2837

rpv-tomsk:
Closed by #2837.

PR #2837, opened 2018-06-25
Use CLOCK_REALTIME for collectd-tg times (fixes issue 2219)

abays:
collectd-tg currently uses CLOCK_MONOTONIC when calling clock_gettime to acquire a time value for use in its lcc_value_list_t objects.  By definition, CLOCK_MONOTONIC "represents monotonic time since some unspecified starting point."  Apparently this "unspecified starting point" is some time further into the future than the standard Unix epoch (1-1-1970).  As a result, the timestamps that are sent out from collectd-tg are not standard epoch timestamps, and convert into human-readable dates much further into the past than the system's actual time.  Using CLOCK_REALTIME instead fixes this problem.

### DIFF
--- a/src/collectd-tg.c
+++ b/src/collectd-tg.c
@@ -104,7 +104,7 @@ static double dtime(void) /* {{{ */
 {
   struct timespec ts = {0};
 
-  if (clock_gettime(CLOCK_MONOTONIC, &ts) != 0)
+  if (clock_gettime(CLOCK_REALTIME, &ts) != 0)
     perror("clock_gettime");
 
   return (double)ts.tv_sec + (double)ts.tv_nsec / 1e9;


abays:
The CI stuff was complaining that the file wasn't clang-formatted.  If I remove that change on line 149, it will start complaining again.  Is that okay?

abays:
Okay, removed the change to 149.

abays:
Nevermind, it's still there.

abays:
All set.  Just the fix for this particular issue is included now.  Will make a separate PR for the clang formatting.

faxm0dem:
I understand. thanks!

rubenk:
Thanks @abays!

abays:
Glad to help out!

faxm0dem:
Fixes #2219

PR #2841, opened 2018-06-28
New plugin to read RAID events

mfijalko:
This plugin, named **mdevents**,  is responsible for gathering the events from RAID arrays that were written to syslog by mdadm utility (which is a user-space software for managing the RAIDs). Then, based on configuration provided by user, plugin will decide whether to send the collectd notification or not. 

Mdevents needs the syslog and mdadm to be present on a platform that collectd is launched.

Based on the naming of existing dpdk plugins (dpdkstat for metrics, dpdkevents for notifications) and the fact that there is already plugin called **md** for gathering metrics from RAIDs, giving this plugin name **mdevents** felt like the best option.

A slight change to ignorelist API was introduced - it is now possible to retrieve the count of nodes in array and the current **ignore** setting for given array.

ChangeLog: mdevents plugin: Generate notifications about events in RAID arrays.

[edit @kkepka:] added changelog

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6344,6 +6344,7 @@ plugin_irq="no"
 plugin_load="no"
 plugin_log_logstash="no"
 plugin_mcelog="no"
+plugin_mdevents="no"
 plugin_memory="no"
 plugin_multimeter="no"
 plugin_nfs="no"
@@ -6399,6 +6400,7 @@ if test "x$ac_system" = "xLinux"; then
   plugin_irq="yes"
   plugin_load="yes"
   plugin_mcelog="yes"
+  plugin_mdevents="yes"
   plugin_memory="yes"
   plugin_nfs="yes"
   plugin_numa="yes"
@@ -6799,6 +6801,7 @@ AC_PLUGIN([match_value],         [yes],                       [The value match])
 AC_PLUGIN([mbmon],               [yes],                       [Query mbmond])
 AC_PLUGIN([mcelog],              [$plugin_mcelog],            [Machine Check Exceptions notifications])
 AC_PLUGIN([md],                  [$have_linux_raid_md_u_h],   [md (Linux software RAID) devices])
+AC_PLUGIN([mdevents],            [$plugin_mdevents],          [Events from md (Linux Software RAID) devices])
 AC_PLUGIN([memcachec],           [$with_libmemcached],        [memcachec statistics])
 AC_PLUGIN([memcached],           [yes],                       [memcached statistics])
 AC_PLUGIN([memory],              [$plugin_memory],            [Memory usage])
@@ -7230,6 +7233,7 @@ AC_MSG_RESULT([    match_value . . . . . $enable_match_value])
 AC_MSG_RESULT([    mbmon . . . . . . . . $enable_mbmon])
 AC_MSG_RESULT([    mcelog  . . . . . . . $enable_mcelog])
 AC_MSG_RESULT([    md  . . . . . . . . . $enable_md])
+AC_MSG_RESULT([    mdevents  . . . . . . $enable_mdevents])
 AC_MSG_RESULT([    memcachec . . . . . . $enable_memcachec])
 AC_MSG_RESULT([    memcached . . . . . . $enable_memcached])
 AC_MSG_RESULT([    memory  . . . . . . . $enable_memory])
--- a/src/mdevents.c
+++ b/src/mdevents.c
@@ -1,0 +1,387 @@
+/**
+ * collectd - src/mdevents.c
+ *
+ * Copyright(c) 2018 Intel Corporation. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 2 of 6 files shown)


configure.ac
```diff
@@ -6328,6 +6335,7 @@ if test "x$ac_system" = "xLinux"; then
   plugin_fhcount="yes"
   plugin_fscache="yes"
   plugin_hugepages="yes"
+  plugin_mdevents="yes"
```
kwiatrox: This variable is not used. Check for `enable_mdevents == "yes"` and set `plugin_mdevents` basing on that condition.

configure.ac
```diff
@@ -6695,6 +6703,7 @@ AC_PLUGIN([hddtemp],             [yes],                     [Query hddtempd])
 AC_PLUGIN([hugepages],           [$plugin_hugepages],       [Hugepages statistics])
 AC_PLUGIN([intel_pmu],           [$with_libjevents],        [Intel performance monitor plugin])
 AC_PLUGIN([intel_rdt],           [$with_libpqos],           [Intel RDT monitor plugin])
+AC_PLUGIN([mdevents],            [$enable_mdevents],        [Events from md (Linux Software RAID) devices])
```
kwiatrox: Check for `plugin_mdevents` instead of `enable_mdevents`

configure.ac
```diff
@@ -7116,6 +7125,7 @@ AC_MSG_RESULT([    hddtemp . . . . . . . $enable_hddtemp])
 AC_MSG_RESULT([    hugepages . . . . . . $enable_hugepages])
 AC_MSG_RESULT([    intel_pmu . . . . . . $enable_intel_pmu])
 AC_MSG_RESULT([    intel_rdt . . . . . . $enable_intel_rdt])
+AC_MSG_RESULT([    mdevents  . . . . . . $enable_mdevents])
```
kwiatrox: Check for `plugin_mdevents` instead of `enable_mdevents`

src/mdevents.c
```diff
@@ -0,0 +1,428 @@
...
+
+#include <limits.h>
+#include <regex.h>
+#include <stdio.h>
+#include <string.h>
+
+#define MD_EVENTS_PLUGIN "mdevents"
+#define DAEMON_NAME "mdadm"
+
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
```
kwiatrox: In collectd already exists predefined macro: STATIC_ARRAY_SIZE

src/mdevents.c
```diff
@@ -0,0 +1,428 @@
...
+          "/var/log/syslog and /var/log/messages files are not present. Are "
+          "you sure that you have rsyslog utility installed on your system?\n");
+      return -1;
+    }
+  }
+
+  // monitor events only from point of collectd start
+  fseek(syslog_file, 0, SEEK_END);
+
+  if (md_events_compile_regex(&regex, regex_pattern))
```
kwiatrox: fclose(syslog_file)

src/mdevents.c
```diff
@@ -0,0 +1,428 @@
...
+    return -1;
+  }
+
+  // each element from matches array contains the indexes (start/end) within
+  // the string that we ran regexp on; use them to retrieve the substrings
+  md_events_copy_match(event.event_name, to_match, matches[EVENT]);
+  md_events_copy_match(event.md_device, to_match, matches[MD_DEVICE]);
+
+  // this one is not present in every event, regex API sets indexes to -1
+  // if the match wasn't found
```
kwiatrox: Maybe return 0 if match is not found in the current line.

src/mdevents.c
```diff
@@ -0,0 +1,428 @@
...
+    if (!syslog_file) {
+      MD_EVENTS_ERROR(
+          "/var/log/syslog and /var/log/messages files are not present. Are "
+          "you sure that you have rsyslog utility installed on your system?\n");
+      return -1;
+    }
+  }
+
+  // monitor events only from point of collectd start
+  fseek(syslog_file, 0, SEEK_END);
```
kwiatrox: Optionally check for `fseek` return value.

mfijalko:
> What do you think about implementation, which will not require syslog in the chain?
> Unfortunately, I found no such solution. Maybe we can ask `mdadm` developers to start implementing the library for such purposes?

Hi! First of all, sorry for that prolonged silence. Thanks for taking a look at this. I'm going to post a patch that addresses your comments.

Answering your question - from what I see, `mdadm` doesn't look like it is "alive", their github is silent for more or less half a year. But you know, we can give it a shot and post a feature request.

On the other hand, does having a syslog dependency is that painful? 

Thanks again and looking forward for your response!

rpv-tomsk:
This PR is something like or related to #3045

rpv-tomsk:
>On the other hand, does having a syslog dependency is that painful?

The longer chain - lower  reliability.

kkepka:
> This PR is something similar to or related to #3045

You are right @rpv-tomsk, functionally both plugins are watching files for particular messages to generate notification about.
The difference is that `mdevents` is to be just enabled and ready to go, while in `logparser` case user needs to provide definition of what want to monitor (via proper configuration). So it's trade of easy-to-use vs flexibility and being error prone on configuration side.

dago:
@mfijalko There is a failing test, would you mind having a look?

kkepka:
Red Hat check have more verbose UT logs.
From there we can see, that UT fails for init unit test when there is no `/var/log/syslog` or `/var/log/messages` available on the system, what is the case in ci env, so i'd say it's more unitests implementation / test env issue, rather then plugin itself.
I don't want to create any debt here, but as ci environment doesn't have dependency for this unit test case, would it make sense to remove it?

sunkuranganath:
Looks like all the changes requested has been resolved along with any of the UT/CI issues. Could we remove "changes requested"?

dago:
@sunkuranganath At least I cannot. @rpv-tomsk Can you resolve the changerequest?

PR #2856, opened 2018-07-10
Fix parsing option for avoiding making BaseDir

manio:
Parsing is broken since:
386ecab4 Reading configuration file is now done outside main().
So after mentioned commit we've got:

collectd: invalid option -- 'B'

This commit fixes this.


rpv-tomsk:
Ouch,...

This also be cherry-picked to 5.8.

Thanks!

rpv-tomsk:
Ouch, this can't be cherry-picked by me, so I have to do PR...

PR #2864, opened 2018-07-18
Add --with-libxml2 option to configure.

jimis:
It's now valid to disable building with libxml2, using --without-libxml2. It's
also possible to set a custom prefix --with-libxml2=/custom/path.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5920,25 +5920,41 @@ AC_SUBST([BUILD_WITH_LIBVARNISH_CFLAGS])
 AC_SUBST([BUILD_WITH_LIBVARNISH_LIBS])
 # }}}
 
-# pkg-config --exists 'libxml-2.0' {{{
-$PKG_CONFIG --exists 'libxml-2.0' 2>/dev/null
-if test $? -eq 0; then
-  with_libxml2="yes"
-else
-  with_libxml2="no (pkg-config doesn't know libxml-2.0)"
-fi
-
-if test "x$with_libxml2" = "xyes"; then
-  with_libxml2_cflags="`$PKG_CONFIG --cflags libxml-2.0`"
-  if test $? -ne 0; then
-    with_libxml2="no"
-  fi
-
-  with_libxml2_ldflags="`$PKG_CONFIG --libs libxml-2.0`"
-  if test $? -ne 0; then
-    with_libxml2="no"
-  fi
-fi
+# --with-libxml2 {{{
+AC_ARG_WITH(libxml2,
+  [AS_HELP_STRING([--with-libxml2@<:@=PREFIX@:>@], [Path to libxml2.])],
+  [
+    if test "x$withval" = "xno"; then
+      with_libxml2="no"
+    else if test "x$withval" = "xyes"; then
+      $PKG_CONFIG --exists 'libxml-2.0' 2>/dev/null
+      if test $? -eq 0; then
+        with_libxml2="yes"
+        with_libxml2_cflags="`$PKG_CONFIG --cflags libxml-2.0`"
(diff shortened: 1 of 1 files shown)


jimis:
Figured this one out because configure complained about unknown option `--without-libxml2`.

rpv-tomsk:
Hi, Dimitrios!
Thanks for your finding and fix provided!

At first look, it is good, however we need CI checks to pass before this PR gets merged.
At the moment, I see no required changes from your side, but they might appear after somebody's review.

rpv-tomsk:
CC @mfournier 
Marc, CI complains about new plugin appeared:

```
### Looking for any plugins previously unsupported on this platform ###
found this new plugin: pcie_errors.so
	linux-vdso.so.1 (0x00007ffe15f17000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f7f1b179000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f7f1b71e000)
Archiving artifacts
Skipping publisher since build result is FAILURE
Finished: FAILURE
```
Can you please take care, to CI checks pass again?

configure.ac
```diff
@@ -5904,32 +5904,29 @@ AC_SUBST([BUILD_WITH_LIBVARNISH_CFLAGS])
...
+      else
+        with_libxml2="no (pkg-config doesn't know libxml-2.0)"
+      fi
+    else
+      with_libxml2="yes"
+      with_libxml2_cflags="-I$withval/include"
+      with_libxml2_ldflags="-L$withval/lib"
+    fi; fi
+  ],
+  [with_libxml2="yes"]
```
rpv-tomsk: This implementations is wrong.
It breaks compilation when `--with-xml2` is not set explicitly.

rpv-tomsk:
FYI, I just moved `libvirt` part to it's own block in `configure.ac`, so please update your tree before work.

jimis:
Sorry I forgot about this. I hadn't noticed the `--with-xml2` parameter, so I implemented `--with-libxml2`. Is it doing the same thing? I'll have a look and get back to you.

jimis:
@rpv-tomsk  when you said `-with-xml2` was it a typo? You meant `--with-libxml2` ?

rpv-tomsk:
> @rpv-tomsk  when you said `-with-xml2` was it a typo? You meant `--with-libxml2` 

Yes, that was a typo. Thanks for update.

jimis:
Squashed the two commits, also amended with a comment line ` # }}}` that was missing at the end of the libxml2 stuff.

rubenk:
Looks good, thanks. A nice cleanup would be to just use the PKG_CHECK_MODULES macros instead of  calling $PKGCONFIG but I'll leave that up to you.

jimis:
I did a bit of reading around PKG_CHECK_MODULES and there is a big controversy around how to use it properly.

I did a small first effort of using it but it is not straightforward to support the traditional `--with-blah=path`, it needs a lot of code around it. So I'd rather we just merge this version at the time, thanks!

rubenk:
Where did you read that? I don't understand the controversy, it's just a simple macro after all.
We have a few existing checks that use PKG_CHECK_MODULES, look at the libprotobuf detection for instance (https://github.com/collectd/collectd/blob/master/configure.ac#L4588)

jimis:
A [first search](https://duckduckgo.com/?q=PKG_CHECK_MODULES) brought up mostly problems and "considered harmful" posts. It can be used properly with plenty of conditionals and boilerplate code around it, I understand, and I see that's the case with libprotobuf detection in collectd. I'll try and make use of it next time since I don't have time to dig into it ATM.

PR #2865, opened 2018-07-19
Support for interface statistics for bond ports for ovs_stats

abays:
These changes to the ovs_stats plugin allow it to report metrics for interfaces belonging to bond ports.  It takes the individual port's statistics and stores them in the plugin's port object for the bond.  When the bond port's values are to be dispatched, the associated interfaces' statistics are combined for each data point.


atyronesmith:
This is a good solution for the majority of cases (2 port bonds setup for 1:1 redundancy).  Should we address the N port (N < 8) case?

abays:
Please see PR #2880 for the latest version of this functionality

PR #2867, opened 2018-07-20
ovs_stats bond support with variable interface counts

abays:
These changes to the ovs_stats plugin allow it to report metrics for interfaces belonging to bond ports.  A new interface struct is introduced, which now stores the interface statistics.  The port struct has been updated to include a linked list of interface objects.  When the port data is to be dispatched, its associated interfaces are traversed and their data points are summed per index (which works for single-interface ports and multiple-interface bond ports alike).

### DIFF
--- a/src/ovs_stats.c
+++ b/src/ovs_stats.c
@@ -1118,18 +1118,18 @@ static int ovs_stats_plugin_read(__attribute__((unused)) user_data_t *ud) {
             for (interface_list_t *iface = port->iface; iface != NULL;
                  iface = iface->next) {
               memset(key_str, '\0', DATA_MAX_NAME_LEN);
-              snprintf(key_str, 5, "uuid%d", i);
+              snprintf(key_str, 6, "uuid%d", i);
               meta_data_add_string(meta, key_str, iface->iface_uuid);
 
               if (strlen(iface->ex_vm_id)) {
                 memset(key_str, '\0', DATA_MAX_NAME_LEN);
-                snprintf(key_str, 8, "vm-uuid%d", i);
+                snprintf(key_str, 9, "vm-uuid%d", i);
                 meta_data_add_string(meta, key_str, iface->ex_vm_id);
               }
 
               if (strlen(iface->ex_iface_id)) {
                 memset(key_str, '\0', DATA_MAX_NAME_LEN);
-                snprintf(key_str, 9, "iface-id%d", i);
+                snprintf(key_str, 10, "iface-id%d", i);
                 meta_data_add_string(meta, key_str, iface->ex_iface_id);
               }
 
--- a/src/ovs_stats.c
+++ b/src/ovs_stats.c
@@ -78,15 +78,22 @@ typedef enum iface_counter {
 #define PORT_NAME_SIZE_MAX 255
 #define UUID_SIZE 64
 
-typedef struct port_s {
-  char name[PORT_NAME_SIZE_MAX];      /* Port name */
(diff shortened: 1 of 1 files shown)


mrunge:
The CI issue seems to be a networking issue, not related to this PR. Any idea how to retrigger?

abays:
@mrunge I think someone needs to log in to Collectd's Jenkins and restart the build.  Otherwise I could commit another change to trigger a new build.

src/ovs_stats.c
```diff
@@ -230,17 +237,87 @@ static port_list_t *ovs_stats_get_port(const char *uuid) {
...
+      return iface;
+  }
   return NULL;
 }
 
+static interface_list_t *ovs_stats_get_interface(const char *uuid) {
+  if (uuid == NULL)
+    return NULL;
+
+  for (port_list_t *port = g_port_list_head; port != NULL; port = port->next) {
```
octo: You should lock `g_stats_lock` before iterating over `port->iface` or document that the caller needs to hold `g_stats_lock`.

src/ovs_stats.c
```diff
@@ -230,17 +237,87 @@ static port_list_t *ovs_stats_get_port(const char *uuid) {
...
+  }
+  return NULL;
+}
+
+static interface_list_t *ovs_stats_get_port_interface(port_list_t *port,
+                                                      const char *uuid) {
+  if (port == NULL || uuid == NULL)
+    return NULL;
+
+  for (interface_list_t *iface = port->iface; iface != NULL;
```
octo: You should lock `g_stats_lock` before iterating over `port->iface` or document that the caller needs to hold `g_stats_lock`.

src/ovs_stats.c
```diff
@@ -472,6 +548,46 @@ static int ovs_stats_update_port(const char *uuid, yajl_val port) {
...
+            // ifaces_list is [[ "uuid", "<some_uuid>" ], [ "uuid",
+            // "<another_uuid>" ], ... ]]
+            for (int i = 0; i < YAJL_GET_ARRAY(ifaces_list)->len; i++) {
+              yajl_val iface_tuple = YAJL_GET_ARRAY(ifaces_list)->values[i];
+
+              // iface_tuple is [ "uuid", "<some_uuid>" ]
+              iface_uuid_str =
+                  YAJL_GET_STRING(YAJL_GET_ARRAY(iface_tuple)->values[1]);
+
+              interface_list_t *iface =
```
octo: You should hold the lock when doing the lookup/create steps.

abays:
I can't really speak to the threading and the use of g_stats_lock here off the top of my head, because I didn't implement that part of this plugin's design.  It was the work of the original author, as far as I know.  My work here was to add the interface list type and use it within the context of the existing logic.  I understand that it's possible that I incorrectly added the new interface list stuff in the context of the locking, but my assumption was that the threading code was fine beforehand.  What I've essentially done is add interface list processing as a replacement for (or an addition to) the port list processing.  Given that the locking structure was already in place and that I merely added more data to be handled within it, is it possible that the plugin as it exists now might have the threading concerns mentioned by @octo?

By the way, this PR is still around because certain people were already relying on it when I created the latest incarnation of these changes, #2880.  That newer PR is more robust in that it allows for an indefinite number of interfaces to be associated with a bond port.  Should we close this PR (and another even older version of it) and focus on #2880 instead?

octo:
ACK, let's close in favor of #2880.

PR #2872, opened 2018-07-24
Create liststate command

saimon46:
This command extend the unix socket plugin to list the state of all Collectd metrics. It provide also a parameter to optionally filter by a given state.
I basically used the command listval and all the dependencies that it have creating liststate. I added to liststate the printing of the state near the metric, in plus there is also the filtering by parameter.
The command can be also used through "collectdctl" or directly to the socket.

This is the output without any parameter:
```
sbrundu@sbrundu-TTL-TEKNOPRO:~$ sudo socat - UNIX-CLIENT:/opt/collectd/var/run/collectd-unixsock 
LISTSTATE
19 Values found
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-docker0/if_dropped OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/interface-docker0/if_errors OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/interface-docker0/if_octets OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/interface-docker0/if_packets OKAY
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-eno1/if_dropped OKAY
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-eno1/if_errors OKAY
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-eno1/if_octets OKAY
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-eno1/if_packets OKAY
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-lo/if_dropped OKAY
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-lo/if_errors OKAY
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-lo/if_octets OKAY
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-lo/if_packets OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/load/load OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-buffered OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-cached OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-free OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-slab_recl OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-slab_unrecl OKAY
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-used ERROR
```

This instead filtering the state ERROR:
```
sbrundu@sbrundu-TTL-TEKNOPRO:~$ sudo socat - UNIX-CLIENT:/opt/collectd/var/run/collectd-unixsock
LISTSTATE state=ERROR
19 Values found
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-used ERROR
```

ChangeLog:Collectd: A new command liststate with filter by state


rpv-tomsk:
>This command extend the unix socket plugin to list the state of all Collectd metrics.

What is it - metric state, what it means? How it works?
How do you use your proposed feature at practice?

saimon46:
The metric state is just the internal state of a metric that collectd already have. This command allow the user to see it. This state is used by notifications mechanism and for instance is used by the Threshold plugin to know when a metric changed from OKAY to ERROR and trigger a notification. Without this command there is no way to have this type of notification.

rpv-tomsk:
We need documentation update, with explain of new feature, new commands and so on.
Project is not only the code, but and documentation too, otherwise people will be unable to use it.

Also, please update PR to pass CI checks.

Thanks for working on Collectd!

rpv-tomsk:
Ouch, please look which files you try to upload ..
You really haven't to upload some of them.

saimon46:
Yes I fixed. Thanks to advise me.
How can I update the documentation with this command?

rpv-tomsk:
>The metric state is just the internal state of a metric that collectd already have.

Is I understand correctly following:

- Collectd has a `cache`, which is able to store `metric state`;
- `threshold` plugin is able to change metric state by rules, defined in `threshold` configuration;
- None of other plugins do metric state changes;
- Your plugin just reports `metric state` from 'cache', and it is not able to report more details, such as warning/failure or 'min' or 'max' reached. (just because Collectd cache does not provides such data).

Sorry, I'm not familar with threshold and notifications of Collectd. (yet)

luisfdez:
Hi @rpv-tomsk ,

There are two data structures now in collectd:
* The avl tree for the cache: https://github.com/collectd/collectd/blob/430f41edba3d13ea1697319c94a2e104f98685ff/src/daemon/utils_cache.c#L79
* The avl tree for the thresholds: https://github.com/collectd/collectd/blob/430f41edba3d13ea1697319c94a2e104f98685ff/src/daemon/utils_threshold.h#L52

I'm not an expert on the internals of collectd, but I think that when all the notification logic was implemented, a state field was added to the internal cache to store what's the state of a given metric:
* https://github.com/collectd/collectd/blob/430f41edba3d13ea1697319c94a2e104f98685ff/src/daemon/utils_cache.c#L53

This command would just give visibility to that value in order to know the current state of a metric.

I don't think it needs to inspect things like warnings/failure min/max as this is a specifc concept of the threshold plugin.

I don't know if this is the proper verb/command representation to express the access to this information, but it is definitely useful.

saimon46:
Hi @rpv-tomsk,
I updated also the docs to complete the work on the new command.

Simone

rpv-tomsk:
cc @rubenk 
Hi,

Ruben, what do you think about this feature?
I have some doubts...

1) I still don't have/understand practical use-case of this feature.
2) This PR duplicates existing code to implement the feature.
    Maybe it should be implemented as extension of existing "listval" command of `unixsock`/ `collectctl`? (via command new arguments).
3) Commit 74de0e1c3b7f591fe30113078547c392ee9b32a9 ... 
    All new files are based on existing project files, so declared authorship is ... a bit incorrect and immodest. ;-)

rpv-tomsk:
Hi,

We have another work in progress, related to this issue: #2889 

Simone, can you please look into #2889 and put your comments?

rpv-tomsk:
>Without such a command, how do you know which metrics have transitioned to error state?

I don't use this feature yet. 
But how do you [plan to] use this? You should use some system to poll metrics and their states. What is the system which does this? Is it something opensource or self-developed project?

rpv-tomsk:
>I like this alternative, lighter to implement and no need to add new verbs to the plugin.

I will prefer the solution, when these two PR will be combined into some _solution_..

rpv-tomsk:
From my thinkings: I think `threshold` plugin (one which changes metric state in Collectd cache, and others if they exists), should store the message into cache too, so it can be used later by commands like `liststate`.

What do you think?

luisfdez:
Hi,

I'm out of the office these days so I'll be a bit slow answering.

> We have another work in progress, related to this issue: #2889

Related in the sense that they apply to notification/alarm side of collectd. Apart from that, they're two independent features.

> From my thinkings: I think threshold plugin (one which changes metric state in Collectd cache, and others if they exists), should store the message into cache too, so it can be used later by commands like liststate.

This is a comment for this PR or the WiP of alertname?

saimon46:
Hi All,

I think, as  Luis said, I can extend the command listval to have the new feature directly in listval, avoiding the creation of a new command that is redundant and most of the code is also a copy and paste. The only problem can be the output that it has to be compliant with the actual output.
So listval in the end can have a parameter that filter the output for every state. Without parameter the output is the same than the actual command. With the filter instead all the metrics are filtered by the state passed but without showing the state of every single metric.

What do you think?

saimon46:
Closing MR in favor of this one: #2908

PR #2874, opened 2018-07-24
virt: Add block info statistics for disk devices

rjablonx:
ChangeLog: virt: Add block info statistics for disk devices

Previously block info statistics(allocation, capacity, physical) were available through libvirt api but were missing in data from collectd-virt plugin.
virDomainBlockInfo statistics are now exposed through virt plugin as:
    -disk_allocation
    -disk_capacity
    -disk_physical

NOTE: newly added block info statistics can be fetched from libvirt only for devices with available 'source' property - because of this scenario additional logic for checking if source is available has been added to virt plugin
NOTE2: previously 'block stats' statistics were stored and described in source code as 'block info' - this pull request contains refactoring that renames 'block_info_*' family of functions to 'block_stats_*' as it should be named from beginning. Without that it would be confusing to have functions with 'block_info' names used for fetching completely different properties from libvirt

### DIFF
--- a/src/types.db
+++ b/src/types.db
@@ -51,6 +51,8 @@ df                      used:GAUGE:0:1125899906842623, free:GAUGE:0:112589990684
 df_complex              value:GAUGE:0:U
 df_inodes               value:GAUGE:0:U
 dilution_of_precision   value:GAUGE:0:U
+disk_allocation         value:GAUGE:0:U
+disk_capacity           value:GAUGE:0:U
 disk_error              value:GAUGE:0:U
 disk_io_time            io_time:DERIVE:0:U, weighted_io_time:DERIVE:0:U
 disk_latency            read:GAUGE:0:U, write:GAUGE:0:U
@@ -58,6 +60,7 @@ disk_merged             read:DERIVE:0:U, write:DERIVE:0:U
 disk_octets             read:DERIVE:0:U, write:DERIVE:0:U
 disk_ops                read:DERIVE:0:U, write:DERIVE:0:U
 disk_ops_complex        value:DERIVE:0:U
+disk_physical           value:GAUGE:0:U
 disk_time               read:DERIVE:0:U, write:DERIVE:0:U
 dns_answer              value:DERIVE:0:U
 dns_notify              value:DERIVE:0:U
--- a/src/virt.c
+++ b/src/virt.c
@@ -617,6 +617,9 @@ enum ex_stats {
   ex_stats_job_stats_completed = 1 << 8,
   ex_stats_job_stats_background = 1 << 9,
 #endif
+  ex_stats_disk_allocation = 1 << 10,
+  ex_stats_disk_capacity = 1 << 11,
+  ex_stats_disk_physical = 1 << 12
 };
 
 static unsigned int extra_stats = ex_stats_none;
@@ -644,6 +647,9 @@ static const struct ex_stats_item ex_stats_table[] = {
     {"job_stats_completed", ex_stats_job_stats_completed},
     {"job_stats_background", ex_stats_job_stats_background},
 #endif
+    {"disk_allocation", ex_stats_disk_allocation},
+    {"disk_capacity", ex_stats_disk_capacity},
+    {"disk_physical", ex_stats_disk_physical},
     {NULL, ex_stats_none},
 };
 
@@ -680,6 +686,12 @@ static void init_block_stats(struct lv_block_stats *bstats) {
   bstats->wr_total_times = -1;
   bstats->fl_req = -1;
   bstats->fl_total_times = -1;
+}
+
+static void init_block_info(virDomainBlockInfoPtr binfo) {
+  binfo->allocation = -1;
+  binfo->capacity = -1;
+  binfo->physical = -1;
 }
 
 #ifdef HAVE_BLOCK_STATS_FLAGS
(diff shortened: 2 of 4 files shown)


rpv-tomsk:
Hi!

What do you think, reporting of these metrics can be optional?
Maybe we have to add new selector to `ExtraStats` or implement `Report***` options?

Thanks for your plugin improvements.

rjablonx:
Hi @rpv-tomsk ,

thanks for review. This request came from barometer community - people were expecting that block info statistics(disk_allocation, disk capacity, disk_physical) will be always collected just like block stats (disk_octets, disk_ops) for same disk device.
I can make it optional - in 'optional' approach I would stick to ExtraStats selectors as there are currently heavily used for virt plugin and I would add "disk_allocation", "disk capacity", "disk_physical" options (the same way as "disk_time" is optional now for disk block devices). Those options could be added to default colllectd,conf as enabled but there will be always possibility to disable them if not needed.

rjablonx:
Hi @rpv-tomsk ,
Is there an option to merge this feature as is and then add 'optional' support for those parameters (as described in comment above) in separate pull request? Currently we have holiday season so delivering this 'optional-support' may delay up to 3-4 weeks from our side.

rjablonx:
Hi guys,
I've added optional-support for new virt parameters (as ExtraStats) - but I 've got problem with CI verification for last patch.
One of the CI jobs failed because of error in testing environment: "ERROR: This node is offline because Jenkins failed to launch the slave agent on it." I've tried to re-trigger CI jobs by "close-and reopen pull request" trick but this problematic job can't be restarted that way.
Could you re-trigger "pull-requests-github_trigger-aggregation" for me manually?

rjablonx:
Hi again,

is there possibility to trigger again "pull-requests-github_trigger-aggregation" on this pull request or do I have to re-upload entire pull request/add new commit? I'm asking because this PR was marked as 'build failed' because of CI infrastructure related error (build failed after ~4 secs because test-machine was offline).

abays:
I know this is a crude suggestion (and you probably already know this), but you could, say, add another whitespace somewhere in the file where it won't make a difference.  Then if you commit and push it should trigger the checks again.

anaudx:
+1, LGTM

rjablonx:
Hi @rpv-tomsk,

I've added support for dynamically enabling/disabling newly added virt disk stats as discussed some time ago (using ExtraStats: "disk_allocation", "disk capacity", "disk_physical"). Is it now ok to merge it or something else has to be added on top of it before merge?

src/types.db
```diff
@@ -51,13 +51,16 @@ df                      used:GAUGE:0:1125899906842623, free:GAUGE:0:112589990684
 df_complex              value:GAUGE:0:U
 df_inodes               value:GAUGE:0:U
 dilution_of_precision   value:GAUGE:0:U
+disk_allocation         value:DERIVE:0:U
```
rpv-tomsk: Hi!

Is these new types (all three) are correct?
At first look they should be GAUGE, not DERIVE.

DERIVE is intended for reporting of rate of value change, and that is not applicable to allocation/available capacity, I think.

Can you please comment this and (or) update PR if needed?

Thanks.
Sorry for review delay :-)

rjablonx:
Hi guys,
it took some time but I've finally added documentation and fixed types.db entries according to @rpv-tomsk comments.

rjablonx:
Hi guys,

I've just noticed that this pull request additionally helps for issue that we have discovered some time ago  - in code that is currently on 'master' branch when BlockDeviceFormat is set to 'source' then all block devices are always filtered out(maybe there is some issue with retrieving block device source-name).
One of patches in this pull request re-implemented fetching target-name and source-name so it seems that it helps also for this issue (and adds new block info statistics as expected:) )

anaudx:
+1 Nice work! LGTM

src/virt.c
```diff
@@ -1689,22 +1717,34 @@ static int get_disk_err(virDomainPtr domain) {
...
   if (!block_dev) {
     ERROR(PLUGIN_NAME " plugin: get_block_stats NULL pointer");
     return -1;
   }
 
-  struct lv_block_info binfo;
-  init_block_info(&binfo);
+  /* Block info statistics can be only fetched from devices with 'source'
+   * defined */
+  if (block_dev->has_source) {
```
rpv-tomsk: Hi,

I think we had to add additional condition check here:
`(extra_stats & (ex_stats_disk_allocation|ex_stats_disk_capacity|ex_stats_disk_physical))`

So, `virDomainGetBlockInfo` will be called only if it's result is really needed and will be used later.

What do you think?

rjablonx: yes, that makes sense. Thanks for suggestion:)
I will implement this change with hope that  I will be able to release updated version before code freeze.

mrunge: @rjablonx any chance to include the suggestion?

rjablonx: @mrunge, yes of course, it is internally implemented and currently it is under tests (we plan to release it during this week)

rjablonx:
Hi guys,
I've just added checking suggested by @rpv-tomsk - now block info statistics will be fetched from libvirt only when needed

rpv-tomsk:
Thanks for your work!

sunkuranganath:
Great to see this got merged almost immediately it was posted! Thanks for the updates and fast merges!

PR #2876, opened 2018-07-25
Fix typo in amqp1 JSON format error message

abays:
Fixes a formatting error log message that mistakenly mentions "graphite" when it should say "json" (as it is in the context of trying to output JSON data)


rpv-tomsk:
Merged, thanks!

Issue #2858, opened 2018-07-11
ovs_stats plugin does not collect dpdk interface statistics if under bond interface


mrunge:
*   Version of collectd:
*   Operating system / distribution:
collectd 5.8.0 / CentOS

## Expected behavior

copied from https://jira.opnfv.org/projects/BAROMETER/issues/BAROMETER-92?filter=allopenissues

The ovs stats plugin collects the right set of data in case the dpdk interfaces are mapped directly under the ovs bridge. But for the use case where a bonding interface is needed underneath the ovs bridge, ovs stats cannot pull the dpdk interface statistics.

More copy/pasting from the linked issue:

$ ovs-vsctl show

Bridge "br0"
Port "br0"
Interface "br0"
type: internal
Port dpdkbond <---------- Port Name
Interface "dpdk0" <---------- Associated Interface name
type: dpdk
options: {dpdk-devargs="0000:05:00.0"}
Interface "dpdk1" <---------- Associated Interface name
type: dpdk
options: {dpdk-devargs="0000:05:00.1"}

What I think is happening here is that OVS stats plugin in this case will request stats for the port with name ‘dpdkbond’ (or the uuid associated with the bonded port) rather than requesting stats for the associated interfaces  ‘dpdk0’ or ‘dpdk1’.

If this is the case there are 2 possible solutions for this

1. Extend OVS to provide aggregate statistics for bond ports, would require modification to OVS source. Not sure this is the correct approach as it does not just affect DPDK ports and will susceptible to much larger input and validation for the non DPDK usecases across the OVS community. I think there is also an argument against providing aggregate stats in the case where the bond is in different modes (load balncing, fail over etc). To my mind it would still be preferable to query the associated interfaces for clarity as regards how traffic is balanced or if a failure has occurred that warrants the back up interface being used.
2. Modify the ovs_stats plugin in collectd to detect if a port is a ‘bond’ port, if so then get the associated interfaces and use them when gathering stats (i.e. use interface name/UUIDs instead of port name/UUID). I think this is the better approach. The ovsdb-client already provides the ability to detect if a port is bonded and lists the associated interfaces with ‘ovsdb-client dump Port’, one would only have to parse for an entry in the ‘bond_active_slave’ column and if this is true then it’s a bonded port, parse the interface entries from the ‘interfaces’ column in that row. 

 

I think option number 2 would be the preferred option as it will work with the existing OVS behavior for bonded ports.

abays:
[...]

PR #2880, opened 2018-07-26
ovs_stats bonds support with separate interface stats

abays:
Third in a series of proposed changes to the ovs_stats plugin.

These changes to the ovs_stats plugin allow it to report metrics for interfaces belonging to bond ports. A new interface struct is introduced, which now stores the interface statistics. The port struct has been updated to include a linked list of interface objects. When the port data is to be dispatched, its associated interfaces are traversed and their data points are summed per index (which works for single-interface ports and multiple-interface bond ports alike).

This patch also adds an "InterfaceStats" config option to the ovs_stats plugin, which allows the plugin to report metrics for individual interfaces.  This is necessary to dispatch metrics for interfaces bundled within a bond port.  Otherwise, as above, only the port statistics will be sent for the bond, which is the sum of all associated interfaces for each data point.

This addresses issue #2858


rpv-tomsk:
Thanks for your work !

PR #2882, opened 2018-07-30
Add GNULIB_DIR to LDFLAGS in configure.ac on Windows

SeanCampbell:
When building on Cygwin, the linker needs to be pointed to the directory with libgnu.dll to build properly.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -779,8 +779,12 @@ AC_FUNC_STRERROR_R
 
 SAVE_CFLAGS="$CFLAGS"
 CFLAGS="-Wall -Werror"
-SAVE_LDFAGS="$LDFLAGS"
+SAVE_LDFLAGS="$LDFLAGS"
 LDFLAGS=""
+if test "x$ac_system" = "xWindows"; then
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -779,8 +779,12 @@ AC_FUNC_STRERROR_R
 
 SAVE_CFLAGS="$CFLAGS"
 CFLAGS="-Wall -Werror"
-SAVE_LDFAGS="$LDFLAGS"
+SAVE_LDFLAGS="$LDFLAGS"
 LDFLAGS=""
+if test "x$ac_system" = "xWindows"; then
+  # This is exported from build.sh
+  LDFLAGS="$LDFLAGS -L${GNULIB_DIR}"
```
tokkee: This will only be used for the strtok test. LDFLAGS is later reset to SAVE_LDFLAGS. Is that what you want?

SeanCampbell: Yes, "-L${GNULIB_DIR}" is added to LDFLAGS in build.sh, so it's saved with the other flags in SAVE_LDFLAGS.

PR #2889, opened 2018-08-03
WiP: Add optional AlertName attribute to the notifications

luisfdez:
This PR adds a new optional attribute to notifications named `AlertName`. We have now message as a field to describe the limits broken but we don't have a field to give a short, descriptive name for the alert/notification.

This field can be useful while integrating notifications with other systems like alertmanager and automation of notification handling in general.

I'm marking it as a work in progress as I have the feeling that a few things should be tested more and polished but it's close to a final implementation.

I would like as well to receive your feedback early before continuing on it. Do you think it's a sensible way of implementing it? any other plugin you think it would be worth adding this field in this first implementation? So far I focused in: threshold, unixsocket, logfile and write_http.


rpv-tomsk:
Can you please provide a configuration example, how did you use this feature in your setup?

luisfdez:
This feature allows you to give known names to the threshold settings.
```
 LoadPlugin "threshold"
 <Plugin "threshold"  
     <Type "load">
        AlertName "HighLoad"
        DataSource "midterm"
        FailureMax 4
     </Type>
   </Host>
 </Plugin>
```
Then, the field is propagated when the notification is sent to write plugins, notification callbacks, etc. So automation actions can be done based on that. I used as an example AlertManager, but this could be as well local actions on the node.

In the current imlemention I also added it to the PUTNOTIF verb, so you could do:

```
PUTNOTIF severity=warning time=1201094702 message="Custom notification script" alert_name=BrokenBlah
```

I hope this helps to understand the context of this feature.

Cheers,
Luis

faxm0dem:
I like the idea.

luisfdez:
@rpv-tomsk what's your input about this proposal after the example configuration given?

luisfdez:
Thanks @mrunge for the review. The first implementation I did of this features was based on the metadata functionality in the internal data structures and I think it could fit better on the module you propose.

I'll give it a try and see if I can make an implementation generic enough to satisfy both use cases.

atyronesmith:
Could this idea be merged / supplanted with 

https://github.com/collectd/collectd/pull/2705

Which adds the ability to include a hierarchical set of metadata...

PR #2895, opened 2018-08-10
Fix most truncation warnings from GCC 8.2 (WIP)

vincentbernat:
GCC 8.2 now warns about possible truncation in snprintf() and
strncat(). As for snprintf(), we introduce an helper ssnprintf() to
handle this correctly. It silently adds the missing '\0' at the end.
The change is mostly mechanical (some help from Coccinelle) and may be
a bit overzealous as in some case, it may be unneeded (GCC doesn't
detect all cases either).

vsnprintf() is also handled by putting the terminal null character in
case of overflow.

As for strncat(), the situation is far more grim. GCC seems to not
understand what this function is about. The canonical use of such a
function is `strncat(out, src, sizeof(out) - strlen(out) - 1)`. But
GCC won't accept that.

This PR is a WIP. Would such changes be acceptable? Moreover, I have
no clue on how to deal with the perfectly `strncat()` in
`snmp_agent.c`:

```
  CC       src/snmp_agent_la-snmp_agent.lo
In function ‘snmp_agent_build_name’,
    inlined from ‘snmp_agent_format_name.constprop’ at ../src/snmp_agent.c:1029:13,
    inlined from ‘snmp_agent_form_reply.isra.22’ at ../src/snmp_agent.c:1091:9:
../src/snmp_agent.c:964:9: error: ‘strncat’ output may be truncated copying between 0 and 127 bytes from a string of length 127 [-Werror=stringop-truncation]
         strncat(out, str, sizeof(out) - strlen(out) - 1);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
make[1]: *** [Makefile:6920: src/snmp_agent_la-snmp_agent.lo] Error 1
```

Apart from that, the result compiles without warnings and tests are
successful. It would need a second pass, but I've skipped that until I
get an informal ack on the change.

### DIFF
--- a/src/aggregation.c
+++ b/src/aggregation.c
@@ -198,15 +198,15 @@ static int agg_instance_create_name(agg_instance_t *inst, /* {{{ */
       sstrncpy(inst->ident.plugin_instance, AGG_FUNC_PLACEHOLDER,
                sizeof(inst->ident.plugin_instance));
     else if (strcmp("", tmp_plugin) != 0)
-      snprintf(inst->ident.plugin_instance, sizeof(inst->ident.plugin_instance),
-               "%s-%s", tmp_plugin, AGG_FUNC_PLACEHOLDER);
+      ssnprintf(inst->ident.plugin_instance, sizeof(inst->ident.plugin_instance),
+                "%s-%s", tmp_plugin, AGG_FUNC_PLACEHOLDER);
     else if (strcmp("", tmp_plugin_instance) != 0)
-      snprintf(inst->ident.plugin_instance, sizeof(inst->ident.plugin_instance),
-               "%s-%s", tmp_plugin_instance, AGG_FUNC_PLACEHOLDER);
+      ssnprintf(inst->ident.plugin_instance, sizeof(inst->ident.plugin_instance),
+                "%s-%s", tmp_plugin_instance, AGG_FUNC_PLACEHOLDER);
     else
-      snprintf(inst->ident.plugin_instance, sizeof(inst->ident.plugin_instance),
-               "%s-%s-%s", tmp_plugin, tmp_plugin_instance,
-               AGG_FUNC_PLACEHOLDER);
+      ssnprintf(inst->ident.plugin_instance, sizeof(inst->ident.plugin_instance),
+                "%s-%s-%s", tmp_plugin, tmp_plugin_instance,
+                AGG_FUNC_PLACEHOLDER);
   }
 
   /* Type */
--- a/src/amqp.c
+++ b/src/amqp.c
@@ -217,23 +217,23 @@ static char *camqp_strerror(camqp_config_t *conf, /* {{{ */
     if (r.reply.id == AMQP_CONNECTION_CLOSE_METHOD) {
       amqp_connection_close_t *m = r.reply.decoded;
       char *tmp = camqp_bytes_cstring(&m->reply_text);
-      snprintf(buffer, buffer_size, "Server connection error %d: %s",
-               m->reply_code, tmp);
+      ssnprintf(buffer, buffer_size, "Server connection error %d: %s",
+                m->reply_code, tmp);
       sfree(tmp);
     } else if (r.reply.id == AMQP_CHANNEL_CLOSE_METHOD) {
       amqp_channel_close_t *m = r.reply.decoded;
       char *tmp = camqp_bytes_cstring(&m->reply_text);
-      snprintf(buffer, buffer_size, "Server channel error %d: %s",
-               m->reply_code, tmp);
+      ssnprintf(buffer, buffer_size, "Server channel error %d: %s",
+                m->reply_code, tmp);
       sfree(tmp);
     } else {
-      snprintf(buffer, buffer_size, "Server error method %#" PRIx32,
-               r.reply.id);
+      ssnprintf(buffer, buffer_size, "Server error method %#" PRIx32,
+                r.reply.id);
     }
     break;
 
   default:
-    snprintf(buffer, buffer_size, "Unknown reply type %i", (int)r.reply_type);
+    ssnprintf(buffer, buffer_size, "Unknown reply type %i", (int)r.reply_type);
   }
 
   return buffer;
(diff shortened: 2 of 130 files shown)


vincentbernat:
I have replaced the litigious `strncat()` by `ssnprintf()` which provides the exact same behaviour (with the additional complexity of parsing a trivial format string), but GCC accepts this use as the return value of `snprintf()` is acted upon while there is no such way to do that with `strncat()`. An alternative would be to introduce `sstrncat()` that would work like `strlcat()` (and ignore the return value).

rubenk:
First, thanks for doing this!

As far as I know, snprintf() always NUL-terminates strings. Do you know of cases where that's not true?
We used to have a ssnprintf helper but we got rid of it for precisely that reason, it didn't have any benefit.

In some cases, silent truncation is the only option, but in others I'd prefer if we could return an error. We'll have to judge that case by case I guess.

rubenk:
I'd love to get rid of the strncat by the way.

rubenk:
Ironically, I may have introduced all these truncation warnings in be126043c2be20399d7670fe194645292018bde0.

vincentbernat:
Yes, `snprintf()` is safe. However, the opinion of GCC 8 is that you should handle truncation as well. The `ssprintf()` helper is only here to say "I don't care if it's truncated". We could handle all cases gracefully, but there are so many of them! Or we could handle only the ones GCC detect. This patch is doing things halfway because I lose hope at some point.

rubenk:
@octo I'd love your input on this, what do you think would be the best cause of action?
I'd love to check all the snprintf calls, but that's quite a lot of work, and in quite a few cases we have no way to return errors up the call chain. On the other hand, these warnings sometimes point to real bugs.

octo:
@rubenk Uff, that's a tough one.

I'd say the sanest way forward would be to re-introduce a wrapper, like @vincentbernat did (thanks by the way!), and make error handling easier – `if ((size_t)ret >= sizeof(buffer))` doesn't exactly roll off the tounge.

My recommendation would be to change the wrapper to this:

```c
/* ssnprintf returns zero on success, one if truncation occurred
   and a negative integer onerror. */
int ssnprintf(char *str, size_t sz, const char *format, ...) {
  va_list ap;
  va_start(ap, format);

  int ret = vsnprintf(str, sz, format, ap);

  va_end(ap);

  if (ret < 0) {
    return ret;
  }
  return (size_t)ret >= sz;
} /* int ssnprintf */
```

This would simplify (but not enforce) error handling. For example:

```c
if (ssnprintf(buf, sizeof(buf), ">>>%s<<<", var)) {
  handle_err();
  return;
}
```

Alternatively we could add two wrappers: one that treats trunction as an error and one that ignores it. The "truncation is an error" one could have the `warn_unused_result` function attribute and zero the buffer on trunction, like C11's `snprintf_s()` (that's sadly not implemented by GNU's libc).

zebity:
hi vincentbernat ,
can you advise on status of this fix. I did quick:

```
# find . -name "*.c" -exec grep -l snprintf {} \; | wc
    143     143    2854
```
and

```
find . -name "*.c" -exec grep -l ssnprintf {} \; | wc
      9       9     172

```

So have 143 *.c files having snprintf in them and with 9 of these having ssnprintf fix.
This compares with 130 files you have added into your fix branch.
Have you determined that the other files do not need fix. Given your statement about fix being mostly mechanical it would be pretty straight forward to do up updates automatically via awk or other script.

zebity:
Hi collectd'ers,
I had a go at building from git master and using ubuntu 19.04 using gcc 8.3.0 and did automatic sed edits (sed -e "s/snprintf/ssnprintf/g") on files to change: snprintf -> ssnprintf (as defined in existing /src/utils/common/common.h).
I made no other changes beyond doing the simple sed update.
This resulted in: 124 *.c files updated.
Next I did compiles and found that I was getting lot of undefined function errors as the "common.h" header file was not being included in existing code. My initial thought was to add include, but instead I just wound out the auto edited files and recompiled.
The result was reversal o: 62 *.c files back to their original content.
Finally I was left with only 5 *.c files that needed to be updated by hand editing them. I just put hack updates to add static _ssnprintf function (same code as here: https://github.com/collectd/collectd/pull/2895#issuecomment-437062237 but I deliberately added static directive and replicated it, as I just wanted to get clean compile going initially).
The following are the 3 files that required hand editing:

```
collectd/src/aggregation.c
collectd/src/protocols.c
collectd/src/sensors.c
collectd/src/collectdctl.c
collectd/src/threshold.c
```
Just posting as I believe that it could be that there is simpler way to get code updated to complie with gcc 8.2/3, which avoids having had edit such a large number of files.

zebity:
Collectd'ers,

I have now got unbuntu 19.04 build with GCC 8.3 compiling with a modules except:

```
    apple_sensors . . . . no
    aquaero . . . . . . . no (libaquaero5.h not found)
    df  . . . . . . . . . no
    gmond . . . . . . . . no (gm_protocol.h not found)
    gpu_nvidia  . . . . . no
    grpc  . . . . . . . . no (protoc3 not found)
    intel_pmu . . . . . . no (jevents.h not found)
    intel_rdt . . . . . . no (pqos.h not found)
    java  . . . . . . . . no (jar not found)
    lpar  . . . . . . . . no (AIX only)
    lvm . . . . . . . . . no (lvm2app.h not found)
    mic . . . . . . . . . no (MicAccessApi not found)
    mysql . . . . . . . . no (mysql_config failed)
    netapp  . . . . . . . no (netapp_api.h not found)
    onewire . . . . . . . no (owcapi.h not found)
    oracle  . . . . . . . no (ORACLE_HOME is not set)
    pf  . . . . . . . . . no
    python  . . . . . . . no
    routeros  . . . . . . no (routeros_api.h not found)
    sigrok  . . . . . . . no (pkg-config could not find libsigrok)
    tape  . . . . . . . . no
    xmms  . . . . . . . . no
    zone  . . . . . . . . no
```

Given that this pull appears to be gone quite, am asking if I should do new pull and put back all updates to get this on way to completed ?
Suggestions please.

vincentbernat:
Let's close in favor of #3153.

PR #2898, opened 2018-08-17
daemon: make plugin_dispatch_multivalue() obey write queue limits

### LINKED ISSUE #2897: Memory leak in plugin_dispatch_multivalue() as it does not obey the write queue limits
*   Version of collectd: 5.5.0 up to 5.8.0 and master
*   Operating system / distribution: Ubuntu 16.04 but all OSes are affected

## Expected behavior

With WriteQueueLimitLow and WriteQueueLimitHigh parameters set to the same value, lets say 10000, and with a single writer plugin which constantly fails to write data to its backend, collectd should stop allocating memory at a certain point - once the write queue is full.

## Actual behavior

Collectd allocates memory for values reported by plugins via plugin_dispatch_multivalue() which does not respect the limits of the write queue. The memory is still reachable as it sits in the write queue but the collectd process' memory consumption grows and grows indefinitely over time.

## Steps to reproduce

*   Prepare a config with
    * memory or swap plugin enabled (which calls plugin_dispatch_multivalue())
    * WriteQueueLimitLow/WriteQueueLimitHigh set to the same value - e.g. 10000
    * a writer plugin which fails to submit data to its backend, e.g. write_graphite with Graphite host set to an invalid IP address
*   run collectd

At some point collectd will start logging:

    [error] plugin_dispatch_values: Low water mark reached. Dropping 100% of metrics.

But the memory consumption of the collectd process will increase over time (verified with Valgrind).

IMO the problem is in [plugin_dispatch_multivalue()](https://github.com/collectd/collectd/blob/df135922b19890047ebf725e6b82cd1c174c6d52/src/daemon/plugin.c#L2102) which does not respect write queue limits, as done in [plugin_dispatch_values()](https://github.com/collectd/collectd/blob/df135922b19890047ebf725e6b82cd1c174c6d52/src/daemon/plugin.c#L2081)

The fix seems trivial so I'll provide a PR.

romanek-adam:
Fixes #2897.

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -146,6 +146,7 @@ static bool plugin_ctx_key_initialized;
 static long write_limit_high;
 static long write_limit_low;
 
+static pthread_mutex_t statistics_lock = PTHREAD_MUTEX_INITIALIZER;
 static derive_t stats_values_dropped;
 static bool record_statistics;
 
@@ -2076,7 +2077,6 @@ static bool check_drop_value(void) /* {{{ */
 
 EXPORT int plugin_dispatch_values(value_list_t const *vl) {
   int status;
-  static pthread_mutex_t statistics_lock = PTHREAD_MUTEX_INITIALIZER;
 
   if (check_drop_value()) {
     if (record_statistics) {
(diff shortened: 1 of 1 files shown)


romanek-adam:
Ci build failed due to a networking issue? Please run it once again.

PR #2899, opened 2018-08-17
powerdns plugin: no-packet-error should be counter, not gauge

XANi:
It is a counter of erroneous packets, and it always goes up so it should be returned as one (just like other counters returned by PDNS)
zaznaczenie_2018-08-17_001

### DIFF
--- a/src/powerdns.c
+++ b/src/powerdns.c
@@ -279,7 +279,7 @@ static statname_lookup_t lookup_table[] = /* {{{ */
         {"ipv6-questions", "dns_question", "incoming-ipv6"},
         {"malloc-bytes", "gauge", "malloc_bytes"},
         {"max-mthread-stack", "gauge", "max_mthread_stack"},
-        {"no-packet-error", "gauge", "no_packet_error"},
+        {"no-packet-error", "errors", "no_packet_error"},
         {"noedns-outqueries", "dns_question", "outgoing-noedns"},
         {"noping-outqueries", "dns_question", "outgoing-noping"},
         {"over-capacity-drops", "dns_question", "incoming-over_capacity"},


rpv-tomsk:
Ouch, disagreed.

I think this should be returned as metric of derive type. 

Also, it looks like other 'counters' should be reported in a same way.
The absolute value of these 'counter' metrics has no much sense. We are interested in _rate_ of errors/hits/drops, so metrics needs to be of 'derive' type (like `dns_question`, `dns_answer`).

XANi:
Well, yes, you're correct of course @rpv-tomsk but I wanted to fix one metric, not break stats of everyone using this plugin.

Changing all to derive would mean anyone that already used it would have their stats broken (as it would most likely be saved under different series) and need to update graphs to use new name

rpv-tomsk:
@octo, what do you think about such backward-incompatible change?

src/powerdns.c
```diff
@@ -279,7 +279,7 @@ static statname_lookup_t lookup_table[] = /* {{{ */
         {"ipv6-questions", "dns_question", "incoming-ipv6"},
         {"malloc-bytes", "gauge", "malloc_bytes"},
         {"max-mthread-stack", "gauge", "max_mthread_stack"},
-        {"no-packet-error", "gauge", "no_packet_error"},
+        {"no-packet-error", "counter", "no_packet_error"},
```
octo: Please use the `errors` type (which is a `DERIVE`) instead of `counter`.

XANi:
@octo sure, that makes more sense, changed type to `errors`

octo:
Cherry-picked into *collectd-5.7* as 373c41d2206603fec5c19019e0fec037ffcbeab4

Thanks @XANi!

PR #2913, opened 2018-08-29
Add support for humidity in sensors plugin

sarahkf:


rpv-tomsk:
Just a note: support of humidity sensors was added in lm_sensors-3.3.0 (released 2011-03-28).

rpv-tomsk:
Hi, Sarah!
Thanks for your contribution.

Can you please move your added code under `#if SENSORS_API_VERSION >= 0x402` condition?
That is not too hard to change for you, and while we have such conditionals in Collectd code we should use them, I think.

Thanks!

src/sensors.c
```diff
@@ -365,6 +365,7 @@ static int sensors_load_conf(void) {
       if ((feature->type != SENSORS_FEATURE_IN) &&
           (feature->type != SENSORS_FEATURE_FAN) &&
           (feature->type != SENSORS_FEATURE_TEMP) &&
+          (feature->type != SENSORS_FEATURE_HUMIDITY) &&
```
rpv-tomsk: Can you please move your added code under #if SENSORS_API_VERSION >= 0x402 condition?
Thanks.

sarahkf: This is fixed now!

rpv-tomsk:
Which equipment you use which provides humidity sensors?

sarahkf:
I am using a Si7021 sensor

rpv-tomsk:
>I am using a Si7021 sensor

How is it connected? Сan you please describe all in more detail?  Thanks.

src/sensors.c
```diff
@@ -367,6 +367,7 @@ static int sensors_load_conf(void) {
           (feature->type != SENSORS_FEATURE_TEMP) &&
 #if SENSORS_API_VERSION >= 0x402
           (feature->type != SENSORS_FEATURE_CURR) &&
+          (feature->type != SENSORS_FEATURE_HUMIDITY) &&
```
rpv-tomsk: Unfortunately, this was not enough.
As you can see, CI check still failing:

```
CC       src/sensors_la-sensors.lo
src/sensors.c: In function 'sensors_load_conf':
src/sensors.c:370: error: 'SENSORS_FEATURE_HUMIDITY' undeclared (first use in this function)
src/sensors.c:370: error: (Each undeclared identifier is reported only once
src/sensors.c:370: error: for each function it appears in.)
src/sensors.c:390: error: 'SENSORS_SUBFEATURE_HUMIDITY_INPUT' undeclared (first use in this function)
src/sensors.c: In function 'sensors_read':
src/sensors.c:525: error: 'SENSORS_FEATURE_HUMIDITY' undeclared (first use in this function)
make[1]: *** [src/sensors_la-sensors.lo] Error 1
```

That is caused by old `lm_sensors` version used in `Centos 6`. 
There is `3.1.1` version used, which has `SENSORS_API_VERSION` set to `0x420`.
The `lm-sensors-3.3.0`, there humidity sensors support firstly added, has `SENSORS_API_VERSION 0x431`. Therefore, we need its own condition for humidity sensors. 

Please, update the code. Thanks.

mtp401:
@rpv-tomsk We're using the SI7021 via Linux's IIO subsystem bridged to hwmon. It's hooked up via I2C.

src/sensors.c
sarahkf: This is fixed now

rpv-tomsk:
Thanks for your improvement and using of Collectd!

PR #2919, opened 2018-09-04
Fix segfault in libvirt plugin

rubenk:
The libvirt plugin refreshes its list of domains every RefreshInterval
seconds. If a domain crashes after getting the list, getting the stats
for its block devices returns an error. We then try to call
virDomainGetName on an invalid pointer. Fix this.

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -1636,7 +1636,7 @@ static int lv_read(user_data_t *ud) {
       ERROR(PLUGIN_NAME
             " failed to get stats for block device (%s) in domain %s",
             state->block_devices[i].path,
-            virDomainGetName(state->domains[i].ptr));
+            virDomainGetName(state->block_devices[i].dom));
   }
 
   /* Get interface stats for each domain. */


rpv-tomsk:
As I understand, that was just a copy-n-paste error, right?

>If a domain crashes after getting the list

Domain should be crashed or stopped domains behave the same?

Another question:  There is `dom->active` before `get_domain_metrics()` call.
Should we do the same check before `get_block_stats` and `get_if_dev_stats` to avoid error logging for stopped/crashed domains, or this is excessive check?

rubenk:
Yeah good point, I haven't seen this on stopped domains but that doesn't mean this couldn't happen.
It was certainly just a copy and paste error.

As for your second question, yes, I added a check (at least in 5.7) that checked if a domain was running. I'm not sure where that's gone, large parts of the plugin were changed since then.

I indeed noticed that fixing this issue results in large amounts of error logging when we try to get stats for a stopped domain, and this should be improved since they're not really errors.

PR #2924, opened 2018-09-11
dpdk: fix dpdk runtime config file path

kevin-laatz-intel:
The config path was changed in DPDK by the commit 'adf1d86  eal: move
runtime config file to new location’. As a result, the dpdk plugins won't
be able to link to the DPDK primary application after this change.

This commit adds a compile time check for the DPDK verison being used and
sets the DPDK runtime config file location accordingly.

Suggested-by: Michael Luo <michael.luo@intel.com>
Signed-off-by: Kevin Laatz <kevin.laatz@intel.com>

### DIFF
--- a/src/utils_dpdk.c
+++ b/src/utils_dpdk.c
@@ -43,7 +43,11 @@
 #include "common.h"
 #include "utils_dpdk.h"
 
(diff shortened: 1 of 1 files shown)


sunkuranganath:
PR looks good.

rpv-tomsk:
Thanks!

PR #2925, opened 2018-09-12
exec plugin: fix potential deadlock after fork()

danvratil:
POSIX forbids calling functions which are not thread-safe like
setenv() after fork() in multi-threaded programs. It happens to
work just fine with some libc implementations, but it causes a
deadlock in the child process with uClibc for example.

The proper way to pass new environment variables to the child
process is to update parent's environment before the fork(), so
that the child inherits the updated environment and then undo the
changes in the parent process again.

### DIFF
--- a/src/exec.c
+++ b/src/exec.c
@@ -240,11 +240,17 @@ static int exec_config(oconfig_item_t *ci) /* {{{ */
   return 0;
 } /* int exec_config }}} */
 
+#if !defined(HAVE_SETENV)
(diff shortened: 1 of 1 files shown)


octo:
Pulled into *collectd-5.8* as 33b2b04f702b3c30661521ddfe2d0f000c0fa5d9 and 0684aea7453186e6f90ce4b6b8c5bcca68f0d1b9. Thanks @danvratil!

PR #2942, opened 2018-10-09
ntpd: plug potential leak on error

rubenk:
Found by scan-build:

src/ntpd.c:833:5: warning: Potential memory leak
    ERROR("ntpd plugin: ntpd_do_query (REQ_GET_KERNEL) failed with status %i",
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./src/daemon/plugin.h:391:20: note: expanded from macro 'ERROR'
                   ^~~~~~~~~~
1 warning generated.

### DIFF
--- a/src/ntpd.c
+++ b/src/ntpd.c
@@ -832,11 +832,13 @@ static int ntpd_read(void) {
   if (status != 0) {
     ERROR("ntpd plugin: ntpd_do_query (REQ_GET_KERNEL) failed with status %i",
           status);
+    free(ik);
     return status;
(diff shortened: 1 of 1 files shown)


rubenk:
yes I think we should. Strange that scan-build didn't complain about that one. @rpv-tomsk care to do a PR?

That whole plugin needs a cleanup too btw.

PR #2946, opened 2018-10-09
gps: fix gps_read

ffontaine:
gps_read now takes three arguments since version 3.18 of gpsd: see
http://git.savannah.gnu.org/cgit/gpsd.git/commit/?id=6bba8b329fc7687b15863d30471d5af402467802

Fixes:
 - http://autobuild.buildroot.net/results/63a4f980d7cdb4b0823840f747589803fcd0d69f

### DIFF
--- a/src/gps.c
+++ b/src/gps.c
@@ -141,7 +141,11 @@ static void *cgps_thread(void *pData) {
         continue;
       }
 
+#if GPSD_API_MAJOR_VERSION > 6
+      if (gps_read(&gpsd_conn, NULL, 0) == -1) {
+#else
       if (gps_read(&gpsd_conn) == -1) {
+#endif
         WARNING("gps plugin: incorrect data! (err_count: %d)", err_count);
         err_count++;
 


src/gps.c
```diff
@@ -141,7 +141,7 @@ static void *cgps_thread(void *pData) {
         continue;
       }
 
-      if (gps_read(&gpsd_conn) == -1) {
```
rpv-tomsk: Hi!

Thanks for your work on Collectd.

Unfortunately, this patch can't be accepted in this form.

The `gpsd` version needs to be checked and conditionals added for both old and new `gps_read()` declarations to be supported.

ffontaine: Indeed, you're perfectly right, I sent this patch as a quick fix to buildroot failure but collectd needs to support old versions of gpsd, I will try to work on this and update the PR.

ffontaine:
The issue can also be fixed by #2947 so I close this PR.

PR #2947, opened 2018-10-09
gps plugin: fix build with newer gpsd

baruchsiach:
gpsd version 3.18 changed the prototype of gps_read(). Make the
gps_read() call depend on GPSD_API_MAJOR_VERSION to fix that.

This fixes build failures like:

gps.c: In function 'cgps_thread':
gps.c:144:11: error: too few arguments to function 'gps_read'
       if (gps_read(&gpsd_conn) == -1) {
		   ^~~~~~~~

### DIFF
--- a/src/gps.c
+++ b/src/gps.c
@@ -141,7 +141,12 @@ static void *cgps_thread(void *pData) {
         continue;
       }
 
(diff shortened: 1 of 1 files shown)


rpv-tomsk:

Issue #2951, opened 2018-10-12
`/proc/diskstats` changes in Linux 4.19


klausman:
*   Version of collectd: 5.8.0
*   Operating system / distribution: Linux 4.19-rc6

## Expected behavior

The disk plugin should export stats as it has done in the past

## Actual behavior

No disk stats are exported at all

## Steps to reproduce

- Run any of the 4.19-rc Linux kernels and probably any recent collectd version

## Cause

With 4.19, Linux gained extra fields in `/proc/diskstats`, indicating r/w stats for Discard operations on SSDs. From `linux/Documentation/iostats.txt`:

```
Field 12 -- # of discards completed
    This is the total number of discards completed successfully.

Field 13 -- # of discards merged
    See the description of field 2

Field 14 -- # of sectors discarded
    This is the total number of sectors discarded successfully.

Field 15 -- # of milliseconds spent discarding
    This is the total number of milliseconds spent by all discards (as
    measured from __make_request() to end_that_request_last()).
```

Collectd simply quits parsing if the number of fields is not exactly what it expects. I think that may be overly conservative, since the nuber of fields usually only gets added to, without rearrangement of existing fields.

I recommend making the checks for field numbers not X!=10 but X<10, that way, we don't lose stats in these cases, and can take time to implement parsing out extra stats if and when 4.19 becomes stable.

klausman:
Yes, that change seems to make disk stats work again. Thanks! I'll patch my local install until the change makes it into a release.

klausman:
Correction: it is not fixed.

But now I think the breakage is in the udev name mangling. Specifically, after a few loops of reading, `output_name` would be, say `sda4` on line 843, just before the libudev-specific block. Just before the `continue` in the _next_ libudev block (line 860), `ouput_name` is garbage, e.g. `ÐX`, indicating uninitialized memory. This name of course doesn't match the whitelist and this every disk gets skipped.

My provider for libudev is eudev 3.2.6 (https://github.com/gentoo/eudev). I'll try without having libudev enabled and see if that gives me metrics.

klausman:
Yes, not using libudev makes the disks appear again. My C is unfortunately not good enough to tell whether the uninitialized memory is a bug in eudev or in collectd.

octo:
Could you run this with debugging enabled and look for a message like:?

> disk plugin: renaming foo => bar

klausman:
I see this, once per poll interval:

```
[...]

PR #2952, opened 2018-10-12
disk plugin: Support Linux kernels with additional fields.

octo:
Linux 4.19 will add additional counters, which caused this plugin to bail out because it checked the number of fields exactly instead of ensuring that there are *at least* the number of fields expected.

Fixes: #2951

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -695,7 +695,9 @@ static int disk_read(void) {
 
     numfields = strsplit(buffer, fields, 32);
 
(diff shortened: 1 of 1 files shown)


octo:
Tested by @klausman in #2951

rubenk:
@octo yeah I'm all for dropping support for 2.4 kernels.

rpv-tomsk:
Hi! 
I have nothing against this.

PR #2955, opened 2018-10-15
disk plugin: Drop support for Linux 2.4.

octo:


rubenk:
Ah well, I still approve ;-)

PR #2961, opened 2018-10-18
Tree wide: Move utilities and libraries to src/utils/.

octo:
This is a first step that does not yet include *all* "utils_*.c" files, but we're getting there.

Most of this change was scripted, so if you spot something odd I'm all ears :)

ChangeLog: collectd: The organization of the source repository has been improved.


octo:
Hi @rubenk,

I re-generated this change on top of *master*. Could you PTAL? I'd like to merge this before merge conflicts accumulate again.

Best regards,
—octo

PR #2963, opened 2018-10-18
ovs_stats: add macro to populate counters list

teknoraver:
Use a stringification macro to populate the counter array. Doing this
instead of duplicating the tokens is less error prone.

Signed-off-by: Matteo Croce <mcroce@redhat.com>


rpv-tomsk:
Hi!

Your patch changes assignments:

```
  [rx_1024_to_1522_packets] = "rx_1024_to_1518_packets",
  [tx_1024_to_1522_packets] = "tx_1024_to_1518_packets",
```

So, your patch not only adds macro, but changes things.
With your patch another value will be assigned, but there are other occurences of 
`1024_to_1518_packets` below in code:

```
          ovs_stats_submit_two(devname, "if_packets", "1024_to_1518_packets",
                               port->stats[rx_1024_to_1522_packets],
                               port->stats[tx_1024_to_1522_packets], meta);

          ovs_stats_submit_two(devname, "if_packets", "1523_to_max_packets",
                               port->stats[rx_1523_to_max_packets],
                               port->stats[tx_1523_to_max_packets], meta);

```

Can you please clarify, why plugin used `1024_to_1518` in all these place and only one place has to be changed?

octo:
If the `tx_1024_to_1518_packets` → `tx_1024_to_1522_packets` change is on purpose, please cherry-pick into `collectd-5.8`. If the identifier is changed (and the metric name remains), then this should go into `master`.

teknoraver:
Hi all,

It seems that in the OpenFlow1.4 statistics the correct field is `1024_to_1522_packets`:
```
# ovs-ofctl -O OpenFlow14 dump-ports ovsbr0
OFPST_PORT reply (OF1.4) (xid=0x2): 3 ports
  port  guest2: rx pkts=0, bytes=0, drop=0, errs=0, frame=?, over=?, crc=?
           tx pkts=0, bytes=0, drop=0, errs=?, coll=?
           duration=132.213s
           rx rfc2819 1_to_64_packets=0, 65_to_127_packets=0,
           128_to_255_packets=0, 256_to_511_packets=0, 512_to_1023_packets=0,
           1024_to_1522_packets=0, 1523_to_max_packets=0,
```

So, the current `rx_1024_to_1518_packets` name is misleading and should be changed regardless of this patch.
I will send a separate PR for this

Thanks,

PR #2964, opened 2018-10-19
docs/review_comments.md: Start document with frequent review comments.

octo:


rpv-tomsk:
During recent work on `statsd` plugin an proposal appeared to add as rule:

- When Collectd plugin open listening socket, `INFO` message should be put in logs.

rpv-tomsk:
Ok, that needs to be put into another document, agreed. Let's add it too?

Another rule is needed: behaviour on plugin configuration error.

Example: https://github.com/collectd/collectd/pull/2891/files#diff-b4176ac9051daca66fe2917e4f4fa5f6R1114

Collectd has `collectd -t`, but implementations makes this flag unuseable.

rpv-tomsk:
We also need to do relating update of https://collectd.org/wiki/index.php/Coding_style .

PR #2970, opened 2018-10-24
src/utils_oauth.c: Renew OAuth tokens 30 seconds before they expire.

octo:
This solves sporious errors like this one:

> write_stackdriver plugin: POST https://monitoring.googleapis.com/v3/projects/{projectId}/timeSeries: API error 401: Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.

### DIFF
--- a/src/utils_oauth.c
+++ b/src/utils_oauth.c
@@ -425,9 +425,10 @@ static int new_token(oauth_t *auth) /* {{{ */
 
 static int renew_token(oauth_t *auth) /* {{{ */
 {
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Just after approve I get an idea, maybe it will be better to take`slack` into account in `new_token()` ?

octo:
I was thinking about that as well. I eventually decided that having accurate data in the data structures and applying "slack" when doing the check was cleaner.

rpv-tomsk:
I also had the same thought. =)

PR #2971, opened 2018-10-24
gps plugin: fix build with newer gpsd

rubenk:
gpsd version 3.18 changed the prototype of gps_read(). Make the
gps_read() call depend on GPSD_API_MAJOR_VERSION to fix that.

This fixes build failures like:

gps.c: In function 'cgps_thread':
gps.c:144:11: error: too few arguments to function 'gps_read'
       if (gps_read(&gpsd_conn) == -1) {
		   ^~~~~~~~

(cherry picked from commit 991a6d3fd38c2435d94de3853fda36b3330cf6ab)

### DIFF
--- a/src/gps.c
+++ b/src/gps.c
@@ -141,7 +141,12 @@ static void *cgps_thread(void *pData) {
         continue;
       }
 
-      if (gps_read(&gpsd_conn) == -1) {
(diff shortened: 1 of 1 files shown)


rubenk:
This patch fixes the build of the just released 5.8.1 on Fedora rawhide.

PR #2972, opened 2018-10-25
virt plugin: Remove optional virDomainGetCPUStats() from main flow

rpv-tomsk:
virDomainGetCPUStats() call is only required when `pcpu' extra statistics is requested.
While not supported in all configurations, this call may fail which causes failure in
reporting of other domain metrics.

virDomainGetCPUStats() was initially added in c7f23810d1f09f45cb4016da6eca6021d56bb0f7 (collectd-5.8.0)


rpv-tomsk:
cc @fromanirh

rpv-tomsk:
Updated.

PR #2976, opened 2018-10-26
Add a new UNKNOWN state as the initial state of metrics

luisfdez:
I am using intensively the collectd notifications and thresholds and I find that due to the lack of an initial _unknown_/_unset_ state for metrics there are some inconsistencies.

Some points I think this modification could improve:
* The docs says "_If a threshold is configured to be non-persistent a notification is created for each state change_". This is not 100% true as you will never receive a notification when a metric is initially considered OK, as it is collectd itself setting that default instead of the threshold plugin or any other plugin.
* Having this initial transition will help when collectd notifications are plugged to other external visualization tools.
* Setting by default `OK` to all the metrics, regarless of the presence or not of threshold definitions, doesn't look precise in particular if in the future this state can be queried by other entities.

It is likely that this PR will require a couple of iterations more but I would appreciate your input on the proposal first.

### DIFF
--- a/src/daemon/utils_cache.h
+++ b/src/daemon/utils_cache.h
@@ -31,9 +31,10 @@
 
 #include "plugin.h"
 
-#define STATE_OKAY 0
-#define STATE_WARNING 1
-#define STATE_ERROR 2
+#define STATE_UNKNOWN 0
+#define STATE_OKAY 1
+#define STATE_WARNING 2
+#define STATE_ERROR 3
 #define STATE_MISSING 15
 
 int uc_init(void);
--- a/src/daemon/utils_cache.c
+++ b/src/daemon/utils_cache.c
@@ -201,7 +201,7 @@ static int uc_insert(const data_set_t *ds, const value_list_t *vl,
   ce->last_time = vl->time;
   ce->last_update = cdtime();
   ce->interval = vl->interval;
-  ce->state = STATE_OKAY;
+  ce->state = STATE_UNKNOWN;
 
   if (c_avl_insert(cache_tree, key_copy, ce) != 0) {
     sfree(key_copy);
--- a/src/threshold.c
+++ b/src/threshold.c
@@ -309,7 +309,10 @@ static int ut_report_state(const data_set_t *ds, const value_list_t *vl,
   /* If the state didn't change, report if `persistent' is specified. If the
    * state is `okay', then only report if `persist_ok` flag is set. */
   if (state == state_old) {
-    if ((th->flags & UT_FLAG_PERSIST) == 0)
+    if (state == STATE_UNKNOWN) {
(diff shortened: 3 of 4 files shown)


octo:
Hi @luisfdez, thank you very much for your suggestion!

I think that's a good point. The users of `uc_get_state()` would need to be updated to handle this case gracefully. Currently the only uses are in `src/threshold.c` and `src/write_riemann.c`, so this should be fairly easy to do.

Best regards,
—octo

src/daemon/utils_cache.h
```diff
@@ -35,6 +35,7 @@
 #define STATE_WARNING 1
 #define STATE_ERROR 2
 #define STATE_MISSING 15
+#define STATE_UNKNOWN 99
```
octo: These defines should be purely internal and it should be harmless to renumber them – zero would be much more appropriate for `UNKNOWN` imho ;)

luisfdez: Good point,  I just updated it in the latest change. It actually helps to keep the logic while evaluating thresholds for unkown to be lower than OKAY.

I left MISSING in 15 as it is a _weird_ state, it is never reflect in a value present on the cache.

luisfdez:
Many thanks @octo  for the input.

I just updated the patch with a few modifictions:
* Renumber all the status but the MISSING state. After exploring the code I see that this state is a bit special as it is never represented in the cache.

A few more notes related to this change:
* While updating the patch I just realised that _MISSING_ is not a real state (in the sense of being present in the cache), so a transition from MISSING to OKAY will never happen. Said that, I think that this is never triggered: https://github.com/collectd/collectd/blob/92c3b2ed5f8e49737e29b11244585960a3478494/src/threshold.c#L365
* However, with the new state, a value that triggered a MISSING notifications will recover by going to UNKNOWN->OK and this will triggered a notification.
* While checking a single datasource I'm returning UNKNOWN as it could happen that the users sets an invalid datasource that doesn't exist and that would transition the metric to OKAY. This way the metric is kept as UNKOWN. This situation:
```
    <Plugin load>
      <Type load>
        DataSource fake
        FailureMax 2
        Persist true
        Hits 0
      </Type>
    </Plugin>
```
* Using the previous example, I also make sure that Persist=true doesn't trigger notifications for UNKNOWN=>UNKNOWN transitions like that example would produce.

In this model, transitions from UNKNOWN to either OKAY or FAILURE should never happen (this could only happen in principle if a type definition changes while collectd is running).

I will take another look to see if I find other conditions that should be polished.

luisfdez:
@octo , from my side I don't have anything else to add, so feel free to merge it.

Many thanks for taking a look into it.

I'll try to find some idle cycles to take a look to the MISSING state side of things.

Cheers,
Luis

octo:
Awesome, thanks @luisfdez!

PR #2981, opened 2018-10-29
exec plugin: remove a pseudo-security anti-feature

l29ah:
unreasonably preventing running programs as root

### DIFF
--- a/src/exec.c
+++ b/src/exec.c
@@ -474,10 +474,6 @@ static int fork_child(program_list_t *pl, int *fd_in, int *fd_out,
 
   uid = sp.pw_uid;
   gid = sp.pw_gid;
-  if (uid == 0) {
-    ERROR("exec plugin: Cowardly refusing to exec program as root.");
-    goto failed;
-  }
 
   /* The group configured in the configfile is set as effective group, because
    * this way the forked process can (re-)gain the user's primary group. */


octo:
Hi @l29ah, thank you very much for your PR! Could you please give a rationale why you think executing as root should be opened up? Name calling this restriction "anti feature" is not a convincing argument.

Thanks and best regards,
—octo

l29ah:
On Sun, Oct 28, 2018 at 11:53:29PM -0700, Florian Forster wrote:
> Hi @l29ah, thank you very much for your PR! Could you please give a rationale why you think executing as root should be opened up? Name calling this restriction "anti feature" is not a convincing argument.

Because there is data in a linux system that is only accessible to root. Having such a "feature" would require one to mess with a pseudo-user and `sudo` with more configuration and essentially the same result.
Could you please give a rationale why you think executing as root should be forbidden?

-- 
()  ascii ribbon campaign - against html mail
/\  http://arc.pasp.de/   - against proprietary attachments

octo:
I don't really care much one way or another, but to play devil's advocate:

*   forcing users to take extra steps makes it hard to *accidentally* run scripts as root that don't require privileges.
*   collectd should not execute anything as root that is loaded from a config file writable by non-root users. When using *sudo*, the security policy is defined in a separate file.
*   *sudo* logs its uses, making it auditable.

I'm not a security specialist and it is entirely possible that the consensus in the community is to drop such restrictions; and your use of the term "pseudo-security anti-feature" hints at that. If so, I'd expect there to be some rational (blog post, mailing list archive, …) that you could point us to…?

Issue #2171, opened 2017-02-10
feature request: add support for the --resolve option in curl plugin (CURLOPT_RESOLVE)?


bpkroth:
Hi, using collectd 5.6 and 5.7 on Debian.

We have a number of websites where I need to check IPv4 and IPv6 paths separately.

In some cases we're able to use the IP address in the URL attribute of the configuration and then send an appropriate Host value in a Header attribute.

Here's what an example conf might look like:

```
LoadPlugin "curl"
<Plugin "curl">
        <Page "https://f.q.d.n/">
                URL "https://10.1.2.3/"
                Header "Host: f.q.d.n"
                Header "Cache-Control: max-age=0"
                MeasureResponseTime true
                VerifyPeer false
                VerifyHost false
        </Page>
        <Page "https-6://f.q.d.n/">
                URL "https://fded::1234/"
                Header "Host: f.q.d.n"
                Header "Cache-Control: max-age=0"
                MeasureResponseTime true
                VerifyPeer false
                VerifyHost false
        </Page>
</Plugin>
```

The trouble is that this fails in strange ways when SNI support and/or redirects to different hostnames are involved (eg: `http://f.q.d.n/old/path` redirects to `https://new.f.q.d.n/new/path` for some or all paths being monitored).  For instance, the curl plugin continues to send the original Host attribute requested in the configuration (`f.q.d.n` in this case) which breaks SNI if the redirected hostname changes (note that the cli `curl` does not do this), or else the redirected hostname followed no longer uses the original IP requested (eg: switches from IPv4 to IPv6 or vise-versa).

Some combination of telling curl how to resolve hostnames to IPs for that series of requests (eg: via `CURLOPT_RESOLVE`) and/or adjusting the redirect behavior to optionally only allow following redirects within the same host would help solve this particular issue for us.

I haven't had a chance to look through the code much yet to see how feasible this is or get started on a patch myself yet, but I figured I'd post the request first and see if anyone had any other thoughts on the matter first.

Thanks,
Brian

shastah:
@bpkroth, I needed something similar and came up with #2986 - have a look at it, maybe it'll be useful in your case too.

bpkroth:
Sorry for the delay.  Thanks for reaching out.  I don't work on the project that had that particular issue anymore so I can't test it directly, but I'll pass it along in case it's still relevant (@uwkhahn).  At a quick skim it looked reasonable.  Thanks!

rpv-tomsk:
[...]

PR #2986, opened 2018-11-03
curl plugin: add AddressFamily

shastah:
In situations when hostnames in URLs resolve to both IPv4 and IPv6 addresses, sometimes it's useful to have separate statistics for both of these.

With this commit, within `<Page>` block you can set
```
  AddressFamily "ipv6"
```
or
```
  AddressFamily "ipv4"
```
to specifically use one or the other.

This might also solve issue #2171

ChangeLog: curl_* plugins: Ability to force specific AddressFamily has been added.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1795,6 +1795,7 @@ finance page and dispatch the value to collectd.
     <Page "stock_quotes">
       Plugin "quotes"
       URL "http://finance.google.com/finance?q=NYSE%3AAMD"
+      AddressFamily "any"
       User "foo"
       Password "bar"
       Digest false
@@ -1834,6 +1835,16 @@ Defaults to C<curl>.
 
 URL of the web site to retrieve. Since a regular expression will be used to
 extract information from this data, non-binary data is a big plus here ;)
+
+=item B<AddressFamily> I<Type>
+
+IP version to resolve URL to. Useful in cases when hostname in URL resolves
+to both IPv4 and IPv6 addresses, and you are interested in using one of them
+specifically.
+Use C<ipv4> to enforce IPv4, C<ipv6> to enforce IPv6. If C<libcurl> is compiled
+without IPv6 support, the latter will result in a warning, and fallback to C<any>.
+If C<Type> cannot be parsed, C<any> will be used as well (all IP versions that
+your system allows).
 
 =item B<User> I<Name>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1841,10 +1841,12 @@ extract information from this data, non-binary data is a big plus here ;)
 IP version to resolve URL to. Useful in cases when hostname in URL resolves
 to both IPv4 and IPv6 addresses, and you are interested in using one of them
 specifically.
-Use C<ipv4> to enforce IPv4, C<ipv6> to enforce IPv6. If C<libcurl> is compiled
-without IPv6 support, the latter will result in a warning, and fallback to C<any>.
-If C<Type> cannot be parsed, C<any> will be used as well (all IP versions that
-your system allows).
+Use C<ipv4> to enforce IPv4, C<ipv6> to enforce IPv6, or C<any> to keep the
+default behavior of resolving addresses to all IP versions your system allows.
+If C<libcurl> is compiled without IPv6 support, using C<ipv6> will result in
+a warning and fallback to C<any>.
+If C<Type> cannot be parsed, a warning will be printed and the whole B<Page>
+block will be ignored.
 
 =item B<User> I<Name>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -1959,6 +1959,7 @@ C<_stats> runtime statistics module of I<CouchDB>
 
   <Plugin curl_json>
     <URL "http://localhost:5984/_stats">
+      AddressFamily "any"
       Instance "httpd"
       <Key "httpd/requests/count">
         Type "http_requests"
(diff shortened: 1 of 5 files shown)


src/curl.c
```diff
@@ -437,7 +442,19 @@ static int cc_config_add_page(oconfig_item_t *ci) /* {{{ */
       status = cf_util_get_string(child, &page->plugin_name);
     else if (strcasecmp("URL", child->key) == 0)
       status = cf_util_get_string(child, &page->url);
-    else if (strcasecmp("User", child->key) == 0)
+    else if (strcasecmp("AddressFamily", child->key) == 0) {
+      status = cf_util_get_string(child, &af);
```
octo: Please define `af` here (i.e. in this block)

src/curl.c
```diff
@@ -437,7 +442,19 @@ static int cc_config_add_page(oconfig_item_t *ci) /* {{{ */
...
+      if (status != 0 || af == NULL) {
+        page->address_family = CURL_IPRESOLVE_WHATEVER;
+      } else if (strcasecmp("inet4", af) == 0 || strcasecmp("ipv4", af) == 0) {
```
octo: Please remove the undocumented `inet4` and `inet6` aliases. This opens up the possibility of inconsistent configurations for little benefit.

src/curl.c
```diff
@@ -437,7 +442,19 @@ static int cc_config_add_page(oconfig_item_t *ci) /* {{{ */
       status = cf_util_get_string(child, &page->plugin_name);
     else if (strcasecmp("URL", child->key) == 0)
       status = cf_util_get_string(child, &page->url);
-    else if (strcasecmp("User", child->key) == 0)
+    else if (strcasecmp("AddressFamily", child->key) == 0) {
+      status = cf_util_get_string(child, &af);
+      if (status != 0 || af == NULL) {
+        page->address_family = CURL_IPRESOLVE_WHATEVER;
```
octo: Please don't set `address_family` on failure. This is impossible to debug for users: for some reason, they run into an internal error and instead of an error message name resolution works slightly different – that's a nightmare to track down.

Instead, just `continue` – the `cf_util_get_…` functions all report an error when returning non-zero.

src/collectd.conf.pod
```diff
@@ -1835,6 +1836,16 @@ Defaults to C<curl>.
 URL of the web site to retrieve. Since a regular expression will be used to
 extract information from this data, non-binary data is a big plus here ;)
 
+=item B<AddressFamily> I<Type>
+
+IP version to resolve URL to. Useful in cases when hostname in URL resolves
+to both IPv4 and IPv6 addresses, and you are interested in using one of them
+specifically.
+Use C<ipv4> to enforce IPv4, C<ipv6> to enforce IPv6. If C<libcurl> is compiled
```
octo: I recommend to add the value `any` to explicitly configure the default behavior.

rpv-tomsk:
Hi @shastah,

Can you please add the same change into other `curl_*` plugins too, for them to be consistent?
Thanks!

src/curl.c
shastah: Done. Also for `curl_info`.

src/curl.c
shastah: I made this into a `WARNING()` and made all warning messages more verbose (including plugin instance name, for those who configure more than one `<Page>` block).
I also added explicit handling of `any` and unknown values for `AddressFamily`

src/collectd.conf.pod
shastah: Done.

shastah:
Hi @octo, @rpv-tomsk 
Thanks for your feedback - I think I fixed it all now. Can you let me know if this is acceptable for this plugin - if it is, then I'll do the same for other `curl_*` plugins.

src/curl.c
```diff
@@ -437,7 +440,34 @@ static int cc_config_add_page(oconfig_item_t *ci) /* {{{ */
...
+         * fall back to default - don't set status to non-zero.
+         */
+        curl_version_info_data *curl_info = curl_version_info(CURLVERSION_NOW);
+        if (curl_info->features & CURL_VERSION_IPV6)
+          page->address_family = CURL_IPRESOLVE_V6;
+        else
+          WARNING("curl plugin: IPv6 not supported by this libCURL. "
+                  "Using fallback `any'.");
+      } else {
+        WARNING("curl plugin: Unsupported value of `%s' "
```
rpv-tomsk: As you set `status = -1` here, I think we have to use `ERROR` instead of `WARNING` here.

rpv-tomsk:
>Can you let me know if this is acceptable for this plugin - if it is, then I'll do the same for other curl_* plugins.

Hi!  I think this is acceptable (with one minor note), so, please, do the same for other curl_* plugins and I will merge this.

src/curl.c
shastah: Actually, I did what most of this plugin's code does, `WARNING` with status `-1` is quite common all over the place
```
      WARNING("curl plugin: Option `%s' not allowed here.", child->key);
      status = -1;
```
```
      WARNING("curl plugin: `Regex' missing in `Match' block.");
      status = -1;
```
```
      WARNING("curl plugin: `Type' missing in `Match' block.");
      status = -1;
```
etc

rpv-tomsk: Yes, that is an open question, how to handle configuration errors in Collectd at all.

Ok, since there are many other similar places, that `WARNING` is ok.

shastah:
I've now implemented the same feature for curl_json and curl_xml - sorry it took me so long (life caught up with me).
Please review these changes and let me know if this is acceptable in this form, I'm happy to amend things if needed.

rpv-tomsk:
Thanks for your work!

PR #2990, opened 2018-11-08
virt plugin: Document Instances option

rpv-tomsk:
Documented implementation from 452324c4a5b6ec07575cfc5d5d339852b040509b

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -9330,12 +9330,6 @@ You can also specify combinations of the B<name> and B<uuid> fields.
 For example B<name uuid> means to concatenate the guest name and UUID
 (with a literal colon character between, thus I<"foo:1234-1234-1234-1234">).
 
-=item B<Instances> B<integer>
-
-How many read instances you want to use for this plugin. The default is one,
-and the sensible setting is a multiple of the B<ReadThreads> value.
-If you are not sure, just use the default setting.
-
 =item B<ExtraStats> B<string>
 
 Report additional extra statistics. The default is no extra statistics, preserving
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
cc @fromanirh

src/collectd.conf.pod
```diff
@@ -9386,11 +9380,38 @@ B<Note>: I<perf> metrics can't be collected if I<intel_rdt> plugin is enabled.
...
 in the lifecycle state of a domain. When set to true notifications will be sent
 for every read cycle. Default is false. Does not affect the stats being
 dispatched.
 
+=item B<Instances> B<integer>
+
+How many read instances you want to use for this plugin. The default is one,
+and the sensible setting is a multiple of the B<ReadThreads> value.
+
+This option is only useful then domains are specially tagged.
```
fromanirh: Well, I'd rephrase. it is *most* useful when domains are tagged, but it could very much be useful if they aren't.
The domain tagging feature is supposed to be built on top of this feature, which predated it.

fromanirh:
And thanks @rpv-tomsk for doing this!

fromanirh:
> @formanirh, don't you think there can be race when one thread does
> `refresh_lists() -> lv_clean_read_state() -> free_block_devices()/free_interface_devices()/ free_domains(state)` and another thread is inside of `lv_read()` and uses these structures?

@rpv-tomsk uhm it may be, but I need to check carefully the code.

rpv-tomsk:
>but I need to check carefully the code.

`refresh_lists()` is done per instance, but there is another issue then: When one of read threads updates `last_refresh` then other threads will not be updated. I think `last_refresh` should be instance variable, not global.

PR #2991, opened 2018-11-09
Add 'quorum' stat to ZooKeeper plugin

chris9871:
ChangeLog: plugin ZooKeeper: Add 'quorum' stat

Add 'quorum' stat to ZooKeeper plugin so that alerts are able to tell the difference between data lag and loss of quorum.

### DIFF
--- a/src/zookeeper.c
+++ b/src/zookeeper.c
@@ -176,6 +176,7 @@ static int zookeeper_read(void) {
   char *save_ptr;
   char *line;
   char *fields[2];
+  long followers = -1;
 
   if (zookeeper_query(buf, sizeof(buf)) < 0) {
     return -1;
@@ -218,15 +219,23 @@ static int zookeeper_read(void) {
     } else if (FIELD_CHECK(fields[0], "zk_approximate_data_size")) {
       zookeeper_submit_gauge("bytes", "approximate_data_size", atol(fields[1]));
     } else if (FIELD_CHECK(fields[0], "zk_followers")) {
-      zookeeper_submit_gauge("count", "followers", atol(fields[1]));
+      followers = atol(fields[1]);
+      zookeeper_submit_gauge("count", "followers", followers);
     } else if (FIELD_CHECK(fields[0], "zk_synced_followers")) {
       zookeeper_submit_gauge("count", "synced_followers", atol(fields[1]));
     } else if (FIELD_CHECK(fields[0], "zk_pending_syncs")) {
       zookeeper_submit_gauge("count", "pending_syncs", atol(fields[1]));
+    } else if (FIELD_CHECK(fields[0], "zk_server_state")) {
+      if (followers < 0) {
+        followers = 0;
+      }
     } else {
       DEBUG("Uncollected zookeeper MNTR field %s", fields[0]);
     }
   }
+  /* Reports 0 for followers, -1 for no zk_server_state, # when zk_followers present. */
+  /* Intended to be used for quorum detection by taking max for each time period. */
+  zookeeper_submit_gauge("count", "quorum", followers);
 
   return 0;
 } /* zookeeper_read */
--- a/src/zookeeper.c
+++ b/src/zookeeper.c
@@ -176,10 +176,11 @@ static int zookeeper_read(void) {
   char *save_ptr;
   char *line;
   char *fields[2];
-  long followers = -1;
+  long followers = 0;
 
   if (zookeeper_query(buf, sizeof(buf)) < 0) {
-    return -1;
+    zookeeper_submit_gauge("count", "quorum", -1);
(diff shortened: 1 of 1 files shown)


src/zookeeper.c
```diff
@@ -218,15 +219,24 @@ static int zookeeper_read(void) {
...
+        followers = 0;
+      }
     } else {
       DEBUG("Uncollected zookeeper MNTR field %s", fields[0]);
     }
   }
+  /* Reports 0 for followers, -1 for no zk_server_state, # when zk_followers
+   * present. Intended to be used for quorum detection by taking max for each
+   * time period. */
+  zookeeper_submit_gauge("count", "quorum", followers);
```
octo: I believe this code will do the following:

*   If `zk_server_state` is present, then
    *   Report the number of followers, potentially zero if the `followers` field was not found.
*   Else, if `zk_server_state` is missing, then
    *   Report the number of followers, potentially –1 if the `followers` field was not found.

What is the motivation behind this behavior? How is this an improvment over having the "folowers" metric only?

I think the intention might be to change the behavior to:

*   If `zk_server_state` equals `leader`, then
    *   Report the "followers" and "synced followers" metrics, even if they are zero.
*   Else
    *   do not report the "followers" and "synced followers" metrics.

I think this is much cleaner to implement by using some state and explicitly write down the logic outside of the loop:

```c
long followers = 0;
long synced_followers = 0;
bool is_leader = false;

// …

    } else if (FIELD_CHECK(fields[0], "zk_followers")) {
      followers = atol(fields[1]);
    } else if (FIELD_CHECK(fields[0], "zk_synced_followers")) {
      synced_followers = atol(fields[1]);
     } else if (FIELD_CHECK(fields[0], "zk_server_state")) {
      is_leader = !strcmp("leader", fields[1]);
    }

// …

// outside loop
if (is_leader) {
  zookeeper_submit_gauge("count", "followers", followers); 
  zookeeper_submit_gauge("count", "synced_followers", synced_followers); 
}
```

chris9871: Hi @octo!  Thanks for the comments!

The intent is to always report something for monitored hosts.  The reason why has to do with what happens upstream:

Host -> metrics staging -> metrics collection -> alerting

If there is a delay in metrics staging, it's impossible to determine if the lack of data at metrics collection is an alertable condition when doing alert checks.  With a '-1' or '0' it's possible to check for the max() for each data point and alert on a value < minimum acceptable quorum (with some inertia to allow for slight inconsistency on when the metric is reported/recieved).

Reporting whether there is a leader isn't really all that useful on top of the 'followers' gauge.  I'll look at the code organization with the above comment in mind.

src/zookeeper.c
```diff
@@ -176,9 +176,11 @@ static int zookeeper_read(void) {
   char *save_ptr;
   char *line;
   char *fields[2];
+  long followers = 0;
 
   if (zookeeper_query(buf, sizeof(buf)) < 0) {
-    return -1;
+    zookeeper_submit_gauge("count", "quorum", -1);
+    return 0;
```
chris9871: I don't entirely understand the implications of this change.  Does returning 0 here seem like it would cause any issues?

octo: If this returns non-zero, collectd treats this as "callback failed" and will start doing an exponential back-off, i.e. the function will be called less and less frequently until it recovers. If you return zero here, the function will be called periodically, even if `zookeeper_query()` fails. That may waste some resources, but since the plugin currently re-establishes a connection every time it reads metrics, I don't think that's an issue.

Keep in mind, though, that the connect timeout may be longer than the report interval, so you may get metrics less frequently than you'd expect in certain network condtions.

src/zookeeper.c
octo: Spot on, I have no experience with Zookeeper but I know distributed locking services in general.

Okay, so if I understand you correctly, you'd like to implement a metric that can be used to detect loss of quorum. The existing followers metric is not suitable for this because during such a situation followers is/may be still non-zero. Is that about right?

src/zookeeper.c
chris9871: I think the long timeout and additional resources are acceptable, as long as it isn't causing problems elsewhere in collectd.

src/zookeeper.c
chris9871: That's correct, I would like to implement a metric that can be used to detect loss of quorum.  And the existing metric is not suitable because:

The followers would be absent, which could be interpreted two ways: 1) the previously present value is the valid value (no alert); or 2) The data is missing so quorum is lost (alert, but with a high percentage of false alerts due to 'exception' latency conditions).  This is basically a 2 generals problem I guess.

So what I wanted was something that when present produces an alert when needed.

PR #3001, opened 2018-11-23
write_graphite plugin: Add option to reverse host

usev6:
ChangeLog: write_graphite plugin: Add option to reverse host


usev6:
My motivation for adding this option to the write_graphite plugin is, that currently I have to jump through some hoops to get all metrics in my preferred format in graphite: a tree like structure resembling DNS.

It's very convenient to use the real hostnames for configuring collectd (and those plugins that have a separate setting for the reported host -- like snmp or mysql). Unfortunately it's too costly to rewrite those metrics on the receiving side (graphite).

I'm aware that my PR might be a case of featuritis. I decided to open a PR nevertheless, because it's probably easier to get a feedback based on the code.

src/utils_format_graphite.c
```diff
@@ -34,6 +34,37 @@
 /* Utils functions to format data sets in graphite format.
  * Largely taken from write_graphite.c as it remains the same formatting */
 
+/* helper function for reverse_hostname */
+void reverse_string(char *r_host, int len) {
+  char t;
+  for (int i = 0; i < len / 2; i++) {
```
anaudx: Just a stylistic comment:
Index `j` could be declared and initialized in the `for` statement so that you can write:
`for (int i = 0, j = len - 1; i <= j; i++, j--) {`

src/utils_format_graphite.c
```diff
@@ -120,7 +151,14 @@ static int gr_format_name_tagged(char *ret, int ret_len, value_list_t const *vl,
   if (postfix == NULL)
     postfix = "";
 
-  gr_copy_escape_part(n_host, vl->host, sizeof(n_host), escape_char, 1);
+  if (flags & GRAPHITE_REVERSE_HOST) {
+    int len_host = strlen(vl->host);
+    char r_host[len_host + 1];
```
anaudx: The size of `r_host` array can be `DATA_MAX_NAME_LEN` as this is the size of `vl->host`, so that computation of `len_host` is not necessary.

src/utils_format_graphite.c
```diff
@@ -198,8 +236,16 @@ static int gr_format_name(char *ret, int ret_len, value_list_t const *vl,
 
   bool preserve_separator = (flags & GRAPHITE_PRESERVE_SEPARATOR);
 
-  gr_copy_escape_part(n_host, vl->host, sizeof(n_host), escape_char,
-                      preserve_separator);
+  if (flags & GRAPHITE_REVERSE_HOST) {
+    int len_host = strlen(vl->host);
+    char r_host[len_host + 1];
```
anaudx: Same comment as above.

usev6:
@anaudx Thanks for looking at this! I'll take a closer look and add the changes you suggested.

One question, though: Since there is a merge conflict due to renaming: Am I right in assuming that solving that merge conflict locally and pushing a new commit to my branch does leave your comments intact? Or would it be better to create a fresh PR?

usev6:
> One question, though: Since there is a merge conflict due to renaming: Am I right in assuming that solving that merge conflict locally and pushing a new commit to my branch does leave your comments intact? Or would it be better to create a fresh PR?

Well, I resolved the conflict and pushed a commit with suggestions from the code review. The review was dismissed by this, but at least the comments are still there.

src/utils/format_graphite/format_graphite.c
```diff
@@ -34,6 +34,36 @@
 /* Utils functions to format data sets in graphite format.
  * Largely taken from write_graphite.c as it remains the same formatting */
 
+/* helper function for reverse_hostname */
+void reverse_string(char *r_host, int len) {
+  char t;
+  for (int i = 0, j = len - 1; i < j; i++, j--) {
+    t = r_host[i];
```
rpv-tomsk: Hi,

Just a minor note: in this place, `char t` had to be declared inside of the loop.

usev6: > Just a minor note: in this place, `char t` had to be declared inside of the loop.

I was under the impression that keeping the declaration outside of the loop would be a tiny bit more efficient. I did some reading and learned that this is not the case. In fact, looking at the disassembled code with objdump showed no differences between the two versions. (I didn't use objdump before, so this was an interesting thing to do.) Many thanks for pointing it out to me!

rpv-tomsk: Compilers are very smart now.
Did you ever try to develop Atmel ATtiny13 microcontroller programs?
It has 1Kb program flash memory only, so ... .

I has one project with tiny13. When I looked into `objdump` output I found compiler optimizations which I ever can't think about. I think, If I will write assembler code, it will not be more than 1-3% smaller (by size) than gcc generated. But I will spent much-much more time to write and support such code.

Yes, that is an interesting experience. )

PR #3002, opened 2018-11-26
GenericJMX: Improve handling of edge cases

mheemskerk:
Improves handling of the following non-standard JMX attributes:

* JMX attributes that return a `javax.management.Attribute` instance
  instead of a scalar. While this is not the correct behaviour, it is
  a not uncommon mistake to make when implementing dynamic MBeans.
  Common tools like jconsole and and java mission control handle this
  gracefully.

* Attributes that use a '.' in the attribute name.  GenericJMX currently
  interprets this as path, loads the JMX attribute identified by the
  first path segment and attempts to read the field identified by the
  remainder. If the application exposes a JMX attribute that has a '.'
  in the attribute name, GenericJMX cannot read it because of the
  path splitting behaviour. This commit adds fallback behaviour that
  attempts to read the JMX attribute as is, if the path-splitting read
  failed.


dago:
@mheemskerk Would you mind rebasing and fixing the conflict and also running the formatter? I would like to include this PR in the upcoming 5.11 release.

mheemskerk:
Hi @dago, I'll try to find some time today to bring my PR up to date with master. Good to see some life in the collectd project!

PR #3004, opened 2018-11-28
Add MySQL channel replication stats

BrandonArp:
Find all replication channels and emit metrics for each individual channel.


bufadu:
This PR seems to have the same purpose than #2842 
Any hope to have this two PR validated anytime soon ?

Thanks !

PR #3018, opened 2018-12-06
Added input parameter to error messages

jkohen:
Print `str` in the `match_apply` error messages to help track down configuration issues. I'm seeing this error in the logs for a configuration someone else wrote, and I don't know what is causing the error.

### DIFF
--- a/src/utils_match.c
+++ b/src/utils_match.c
@@ -354,11 +354,11 @@ int match_apply(cu_match_t *obj, const char *str) {
   }
 
   if (status != 0) {
-    ERROR("utils_match: match_apply: match_substr failed.");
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Hi!

I don't think this is a good idea, as `str` can contain values of huge size.
For `match_substr()` there is two cases when it will return error, and these are not depend on configuration.

Please explain, how this patch helped you.

jkohen:
I see this message in my logs and it's not helpful to understand the
problem. What do you suggest to log? Would it be ok to log only the prefix
of `str`?

jkohen:
Thanks for your patience.

Issue #3026, opened 2018-12-18
processes plugin: Delay accounting behaves wrong on short-living processes


rpv-tomsk:
*   Version of collectd: master branch
*   Operating system / distribution: Linux
*   Kernel version (if applicable):  SMP Debian 3.2.82-1 x86_64 GNU/Linux

## Expected behavior

Expected to have no abnormal spikes on charts.

## Actual behavior

There is abnormal spikes when processes group has short-living processes (cron scripts, php-fpm workers). 
It works perfectly only when processes are long-running (mysql processes).

rrdtool dump has value, for example:
```
<!-- 2018-12-18 10:31:50 GMT / 1545129110 --> <row><v>1.2888634147e+09</v></row>
```

![image](https://user-images.githubusercontent.com/13328211/50181884-43b16400-0340-11e9-92e0-a55f5d4c741e.png)
![image](https://user-images.githubusercontent.com/13328211/50181907-5a57bb00-0340-11e9-84c2-f6e97aa93712.png)
![image](https://user-images.githubusercontent.com/13328211/50181931-70fe1200-0340-11e9-8301-004007be8df6.png)

@octo, what do you think?

rpv-tomsk:
Also, due to use of `value_to_rate`, plugin sends understated values as it losts values from first reading of process delay values.

When we have first read cycle completed, we have full list of processes which belong to their groups.

If new process found for group, we should account it's delay values immediately, not in next read cycle.
Otherwise this introduces huge inaccuracy - we lose data for first interval (when process just started) and for last interval (when process finished) both, while we can account first interval too.

Examples: 
Unpatched:
![image](https://user-images.githubusercontent.com/13328211/50209304-fd8eeb80-03a5-11e9-91e7-6106d00e714a.png)
Patched:
![image](https://user-images.githubusercontent.com/13328211/50209280-ee0fa280-03a5-11e9-8011-8faf15838184.png)

For MySQL there is some (small) difference appear, but I expect average values to be the same (charts will be updated soon):

Unpatched:
![image](https://user-images.githubusercontent.com/13328211/50209401-375ff200-03a6-11e9-8bf6-22f26778c5cf.png)
Patched:
![image](https://user-images.githubusercontent.com/13328211/50209427-45ae0e00-03a6-11e9-9c95-9830fd2eee17.png)

rpv-tomsk:
The patch used:
```
--- a/src/processes.c
+++ b/src/processes.c
[...]

PR #3027, opened 2018-12-19
processes plugin: Reimplement `delay_rate` calculation

rpv-tomsk:
Existing implementation ignores first readed value.
That introduces significant inaccuracy in metrics for short-living processes.

Issue: #3026

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -218,6 +218,9 @@ typedef struct procstat_entry_s {
   unsigned long id;
   unsigned char age;
 
+  cdtime_t now;
+  cdtime_t last_read;
+
   derive_t vmem_minflt_counter;
   derive_t vmem_majflt_counter;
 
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Hi, Florian!

>Hi @rpv-tomsk, a previous counter value and time stamp is required to calculate a rate. 
>
>The updated code will also skip the first value:
>
>* In `ps_update_delay`, `prev->last_read` is initially zero → `interval` will remain zero.
>* In `ps_update_delay_one`, the zero interval is checked and the function returns early.

That is true only for first read cycle (due to `last_submit` has zero value) and is expected.
For subsequent read cycles the `new->last_read = ps->last_submit;` assignment will work.

>Should it contain a bug (I'm not convinced it does) we should fix it there rather than
>re-implementing the logic in individual plugins.

The `value_to_rate()` ignores first readed value by `state->last_time == 0` check, and that is by it's design.

But here we need to assume what where was zero value at the moment of previous check (when currently processed process was not started yet), 
and we do calculate rate as `first value got for this new process` / `interval`.

So, function does not contain a bug, but just not suits for this task well.

>I'd much prefer to keep using the `value_to_rate()` function. 

To keep using the `value_to_rate()` function, we, probably, have to set `state->last_value = 0;` 
and `state->last_time = now() - interval;` before `value_to_rate()` call. 
But I do not have a such solution ready.

Regards,
Pavel.

dago:
Adjustment is needed in https://github.com/collectd/collectd/blob/5ebc182261321f0781f38d8645bfc7c74353434c/src/utils/common/common.c#L1384 to not skip the first entry

rpv-tomsk:
>Adjustment is needed in .. to not skip the first entry

As said before, that skip is by design.

Issue #2510, opened 2017-10-27
netlink plugin not working on CentOS, RHEL


mrunge:
*   Version of collectd: 5.5.0, 5.7.1 
it is reproducible with collectd-5.7.1-2.el7.x86_64 from EPEL

*   Operating system / distribution:
CentOS 7

## Expected behavior

(Description of the behavior / output that you expected)

## Actual behavior

```
Oct 27 08:40:02 localhost collectd[9317]: Initialization complete, entering read-loop.
Oct 27 08:40:02 localhost collectd[9317]: netlink plugin: link_filter_cb: IFLA_STATS64 mnl_attr_validate2 failed.
Oct 27 08:40:02 localhost collectd[9317]: netlink plugin: ir_read: mnl_socket_recvfrom failed.
Oct 27 08:40:02 localhost collectd[9317]: read-function of plugin `netlink' failed. Will suspend it for 20.000 seconds.

```
The config is pretty minimal:
```
LoadPlugin "netlink"
<Plugin "netlink">
 Interface "eth0"
</Plugin>
```

(Description of the behavior / output that you observed)

## Steps to reproduce

*   set up a fresh centos 7
*   yum install epel-release
*   yum install collectd-netlink
*  systemctl start collectd

smojtabai:
We are having this issue as well.

octo:
The docs have this to say:

> *   `mnl_attr_validate2()`
>     This function allows to perform a more accurate validation for attributes whose size is variable. If the size of the attribute is not what we expect, this functions returns -1 and errno is explicitly set.
> *   `mnl_socket_recvfrom()`
>     On error, it returns -1 and errno is appropriately set. If errno is set to `ENOSPC`, it means that the buffer that you have passed to store the netlink message is too small, so you have received a truncated message. To avoid this, you have to allocate a buffer of `MNL_SOCKET_BUFFER_SIZE` (which is 8KB, see `linux/netlink.h` for more information). Using this buffer size ensures that your buffer is big enough to store the netlink message without truncating it.

hontvari:
The same on Ubuntu 16.04, Collectd 5.5.1.

```
netlink plugin: link_filter_cb: IFLA_STATS64 mnl_attr_validate2 failed.
netlink plugin: ir_read: mnl_socket_recvfrom failed.
```

AlexZzz:
Hi! I faced with the same problem:
```
[...]

PR #3040, opened 2019-01-17
Backport: netlink plugin: Use of less strict rules in link_filter_cb()

AlexZzz:
Hi all! This is a bugfix backport request to collectd-5.8:

mnl_attr_validate2() function implements strict equality check of kernel and
userspace structures size. Additional counters was added to 4.6 Linux kernel,
sizes was changed and mismatch can occur.

This patch weakened validation.
Now Collectd just checks if structures, received from kernel space, has enough
data.

Closes: #2510

# Conflicts:
#	src/netlink.c


rpv-tomsk:
Thanks for reporting this!

PR #3043, opened 2019-01-22
WIP: Add TLS support for the amqp plugin

JAORMX:
This adds TLS support for the AMQP plugin. Namely, by adding the
following parameters that apply for both publishers and subscribers:

* CACert: the CA certificate to use for establishing trust (Setting this
          will enable TLS for the connection).

* ClientCert: The client certificate to use; this could be used if we
              want to enable mutually authenticated TLS.

* ClientKey: The client key to use; this could be used if we
              want to enable mutually authenticated TLS.

* VerifyPeer: Sets whether we want to verify the certificate of the
              other peer (defaults to true).

* VerifyHostname: Sets whether we want to verify the hostname of the
                  other peer. This would normally be done by verifying
                  the SAN extension or the CN of the peer certificate.
                  (defaults to true).

### DIFF
--- a/src/amqp.c
+++ b/src/amqp.c
@@ -36,6 +36,7 @@
 
 #include <amqp.h>
 #include <amqp_framing.h>
+#include <amqp_ssl_socket.h>
 
 #ifdef HAVE_AMQP_TCP_SOCKET_H
 #include <amqp_tcp_socket.h>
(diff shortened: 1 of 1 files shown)


JAORMX:
@rubenk @octo is TLS being tested for any other plugin? Seems Travis CI is using quite an old version of OpenSSL, which is not compatible with collectd.

PR #2624, opened 2017-12-13
Red Hat NFVPE Sysevent Plugin

abays:
ChangeLog: Sysevent plugin: A new plugin that monitors rsyslog for system events.

PR #3045, opened 2019-01-24
logparser:  Utility for filtering and parsing log messages

rjablonx:
ChangeLog: logparser: Add utility for filtering and parsing log messages

This utility simplifies filtering and parsing specific messages from given log file. It tracks log file using `utils_tail_match.h` API.

Main feature is automatic messages assembly based on set of user provided regular expressions containing distinguished expressions for recognizing start and the end of the message.
That makes assembling single message from multiple log lines quite easy. Utility supports multiple parsing jobs and also user defined message part validation flags (mandatory or not).


mrunge:
It looks like this PR is trying to achieve the same as https://github.com/collectd/collectd/pull/2624 (or the other way around)

src/utils_match.c
```diff
@@ -33,6 +33,7 @@
 
 #include <regex.h>
 
+#define UTILS_MATCH_FLAGS_REGEX 0x04
```
anaudx: Duplicates line 38

Makefile.am
```diff
@@ -1160,6 +1174,32 @@ logfile_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 logfile_la_DEPENDENCIES = $(COMMON_DEPS)
 endif
 
+if BUILD_PLUGIN_LOGPARSER
+pkglib_LTLIBRARIES += logparser.la
+logparser_la_SOURCES = src/logparser.c \
+	src/utils_message_parser.c src/utils_message_parser.h \
```
anaudx: Utils files `utils_*` are now stored in subdirectories of `src/utils/`.

src/collectd.conf.pod
```diff
@@ -3855,6 +3855,158 @@ B<Note>: There is no need to notify the daemon after moving or removing the
...
+=item B<PluginInstance> I<true>|I<String>
+
+If set to true, it sets plugin instance to string returned by regex. It can be
+overridden by user string.
+
+=item B<Tyoe> I<true>|I<String>
+
+Sets notification type using rules like B<PluginInstance>.
+
+=item B<TyoeInstance> I<true>|I<String>
```
anaudx: `TypeInstance`

src/collectd.conf.pod
```diff
@@ -3855,6 +3855,158 @@ B<Note>: There is no need to notify the daemon after moving or removing the
...
+Flag indicating if I<Match> item is mandatory for message validation. If set to
+true, whole message is discarded if it's missing. For false its presence is
+optional. Default value is set to true.
+
+=item B<PluginInstance> I<true>|I<String>
+
+If set to true, it sets plugin instance to string returned by regex. It can be
+overridden by user string.
+
+=item B<Tyoe> I<true>|I<String>
```
anaudx: `Type`

src/collectd.conf.pod
```diff
@@ -3855,6 +3855,158 @@ B<Note>: There is no need to notify the daemon after moving or removing the
...
+=over 4
+
+=item B<Logfile> I<File>
+
+The B<Logfile> block defines file to search. It may contain one or more
+B<Message> blocks which are defined below.
+
+=item B<FirstFullRead> I<true>|I<false>
+
+Set to true if want to read the file from the beginning first time.
```
anaudx: Reword to something like:
`Set to true if the file has to be read from the beginning on the first time`
Also add the behaviour when set to `false`.

src/collectd.conf.pod
```diff
@@ -3855,6 +3855,158 @@ B<Note>: There is no need to notify the daemon after moving or removing the
...
+
+=item B<Message> I<Name>
+
+B<Message> block contains matches to search the log file for.
```
anaudx: Also add something like:
`Each B<Message> block builds a notification message using matched elements if its mandatory B<Match> blocks were matched.`

src/collectd.conf.pod
```diff
@@ -3855,6 +3855,158 @@ B<Note>: There is no need to notify the daemon after moving or removing the
...
+
+=item B<DefaultPluginInstance> I<String>
+
+Sets the default plugin instance.
```
anaudx: Please reword to something like:
`Sets the default content of the plugin instance part of the message.`

src/collectd.conf.pod
```diff
@@ -3855,6 +3855,158 @@ B<Note>: There is no need to notify the daemon after moving or removing the
...
+
+=item B<DefaultType> I<String>
+
+Sets the default type.
+
+=item B<DefaultTypeInstance> I<String>
+
+Sets the default type instance.
```
anaudx: Please reword to something like:
`Sets the default content of the type instance part of the message.`

src/utils_tail.c
```diff
@@ -68,10 +68,11 @@ static int cu_tail_reopen(cu_tail_t *obj) {
...
   }
 
-  /* Seek to the end if we re-open the same file again or the file opened
-   * is the first at all or the first after an error */
+  /* Unless flag for rewinding to the start is set, seek to the end
+   * if we re-open the same file again or the file opened is the first at all
+   * or the first after an error */
   if ((obj->stat.st_ino == 0) || (obj->stat.st_ino == stat_buf.st_ino))
-    seek_end = 1;
+    seek_end = force_rewind ? false : true;
```
anaudx: Or simpler:
`seek_end = !force_rewind`

src/collectd.conf.pod
```diff
@@ -3855,6 +3855,158 @@ B<Note>: There is no need to notify the daemon after moving or removing the
...
+    </Logfile>
+  </Plugin>
+
+B<Options:>
+
+=over 4
+
+=item B<Logfile> I<File>
+
+The B<Logfile> block defines file to search. It may contain one or more
```
anaudx: Add information such as:
`Only one B<Logfile> block is allowed.`

kwiatrox: There can be more than one.

kwiatrox:
Some more explanation and use case:
The plugin can be used to look for patterns in logfiles (e.g. error messages in syslog) and generate notifications based on that information. Because it uses regular expressions in configuration file, the plugin is as much generic as possible. It can be used to get different errors and the notification can be formatted to contain required data obtained from message body.

For example, with provided default configuration, the plugin can look for AER errors in syslog, which can look like:
```
localhost kernel: [655739.977727] pcieport 0000:00:03.0: AER: Uncorrected (Non-Fatal) error received: id=0400
localhost kernel: [655739.993041] igb_uio 0000:04:00.0: PCIe Bus Error: severity=Uncorrected (Non-Fatal), type=Transaction Layer, id=0400(Receiver ID)
localhost kernel: [655740.008359] igb_uio 0000:04:00.0:   device [8086:10fb] error status/mask=00200000/00000000
localhost kernel: [655740.023499] igb_uio 0000:04:00.0:    [21] ACS Violation
localhost kernel: [655740.031056] igb_uio 0000:04:00.0: broadcast error_detected message
localhost kernel: [655740.038486] igb_uio 0000:04:00.0: AER: Device recovery failed
```

The plugin detects occurrence of the error and formats the notifications according to provided configuration. The details of notification can be adjusted using robust configuration options.

rpv-tomsk:
>The plugin detects occurrence of the error and formats the notifications according to provided configuration.
Have you any other interesting use-cases?

rpv-tomsk:
> =item B<FirstFullRead> I<true>|I<false>
> 
> Set to true if the file has to be parsed from the beginning on the first read.
> If false only subsequent writes to log file will be parsed.

I prefer to get strong explanation, why this feature is really needed.
While records are in file already as they was added in the past, you will get notifications like they occured right now.
Such configuration will behave badly on Collectd restart.

This does not costs changes in existing files/api you did.
IMHO.

kwiatrox:
@rpv-tomsk 
For the timestamp issue, the correct time for the past events can be stored within notification, in the example it is "incident time". The receiver of notification can use it to retrieve real time that the error has happened.

For the other example consider errors reported by mcelog other than memory errors. The mcelog plugin can read only memory errors, other types of errors are logged only to logfile. The example for IO error is below:
```
Dec 20 16:21:32 localhost kernel: [19093714.501970] mce: [Hardware Error]: Machine check events logged
Dec 20 16:21:32 localhost mcelog[43102]: Hardware event. This is not a software error.
Dec 20 16:21:32 localhost mcelog[43102]: MCE 0
Dec 20 16:21:32 localhost mcelog[43102]: CPU 0 BANK 1
Dec 20 16:21:32 localhost mcelog[43102]: MISC 0
Dec 20 16:21:32 localhost mcelog[43102]: TIME 1576855292 Fri Dec 20 16:21:32 2019
Dec 20 16:21:32 localhost mcelog[43102]: MCG status:
Dec 20 16:21:32 localhost mcelog[43102]: MCi status:
Dec 20 16:21:32 localhost mcelog[43102]: Corrected error
Dec 20 16:21:32 localhost mcelog[43102]: MCi_MISC register valid
Dec 20 16:21:32 localhost mcelog[43102]: MCA: BUS error: 0 0 Level-3 Generic Generic IO Request-did-not-timeout
Dec 20 16:21:32 localhost mcelog[43102]: Running trigger `bus-error-trigger'
Dec 20 16:21:32 localhost  mcelog[43102]: IO MCA reported by root port 0:00:00.0
Dec 20 16:21:32 localhost mcelog[43102]: Running trigger `iomca-error-trigger'
Dec 20 16:21:32 localhost mcelog[43102]: STATUS 8800000000000e0b MCGSTATUS 0
Dec 20 16:21:32 localhost mcelog[43102]: MCGCAP f000c14 APICID 0 SOCKETID 0
Dec 20 16:21:32 localhost mcelog[43102]: PPIN 360952d048c24c8b
Dec 20 16:21:32 localhost mcelog[43102]: MICROCODE *******
Dec 20 16:21:32 localhost mcelog[43102]: CPUID Vendor Intel Family * Model **
Dec 20 16:21:32 localhost mcelog: CPU 0 on socket 0 received Bus and Interconnect Errors in IO
Dec 20 16:21:32 localhost mcelog[43102]: <27>Dec 20 16:21:32 mcelog: CPU 0 on socket 0 received Bus and Interconnect Errors in IO
Dec 20 16:21:32 localhost mcelog: CPU 0 on socket 0 received IO MCA Errors from 0:00:00.0
Dec 20 16:21:32 localhost mcelog[43102]: <27>Dec 20 16:21:32 mcelog: CPU 0 on socket 0 received IO MCA Errors from 0:00:00.0
```

Regarding your question about `FirstFullRead` parameter. Some of the mce errors can be critical and lead to crush of the system, in such case the error can be read by logparser only after restart. As I explained the real time of event can be provided through one the message fields. This option can be false in most cases, but there is an use for it and I think it's presence does not do any harm.

rpv-tomsk:
Thanks for your explanation.
For my cases, I can find such plugin useful, to get notifications like you described.
But benefit from this plugin can be achived only when having good configuration, which is possible to get only having  enough experience or detailed documentation of kernel logged messages.

sunkuranganath:
We had this plugin tested to ensure it works as indicated.
If no more comments, we could approve this.

@kwiatrox If you get a chance, please capture the examples mentioned above in collectd wiki to make it easy.

dmitriklein:
Can logparser generate metrics like collectd_logparser_.. ? Where it sends messages to? How they can be captured? Thanks.

PR #3047, opened 2019-01-28
collection3: Improve index.cgi form fieldset border style

cjmayo:
- Fit to the form contents instead of filling the page width
- Relate better to the table below by aligning left-hand side and
  rounding corners

### DIFF
--- a/contrib/collection3/share/style.css
+++ b/contrib/collection3/share/style.css
@@ -1,3 +1,12 @@
+form {
+  display: flex;
+  margin-bottom: 10px;
+}
+
+fieldset {
+  margin-left: 0;
+}
+
 div.graph
 {
 }


rpv-tomsk:
Hi!

Thanks for your work on Collectd project!

I would prefer to merge this change as following:

```
--- style.css.orig      2014-12-15 09:34:14.000000000 +0000
+++ style.css   2019-01-30 07:46:25.707911373 +0000
@@ -1,3 +1,10 @@
+form {
+  display: flex;
+  margin-bottom: 10px;
+}
+fieldset {
+  margin-left: 0;
+}
 div.graph
 {
 }
```

I.e. w/o rounding corners. Your patch rounds only corners on form, but leaves images table corners straight. I would prefer to leave all the corners as straight to minimize changes.

rpv-tomsk:
Thanks!

PR #3055, opened 2019-02-01
Add OS X to the Travis build matrix.

dfandrich:


dfandrich:
It looks like this is working now. However, it shows a problem in the Prometheus plugin on OS X:

src/write_prometheus.c:763:50: error: use of undeclared identifier 'SOCK_CLOEXEC'
    fd = socket(ai->ai_family, ai->ai_socktype | SOCK_CLOEXEC, 0);

which makes sense since SOCK_CLOEXEC is Linux-specific. We could check this in as-is but it will cause

dfandrich:
Thanks for fixing the SOCK_CLOEXEC problem. I guess this PR will need to wait for a merge from collectd-5.7 before it makes sense to apply.

octo:
Hi, I merged 5.7 and 5.8 into *master* (#3066) and rebased this PR to it.

octo:
I have good news and bad: the `SOCK_CLOEXEC` issue is resolved, but it appears that Mac OS X really dislikes LDAP:

```
src/openldap.c:74:5: warning: 'ldap_unbind_ext_s' is deprecated: first deprecated in macOS 10.11 - use OpenDirectory Framework [-Wdeprecated-declarations]
    ldap_unbind_ext_s(st->ld, NULL, NULL);
    ^
```

rubenk:
We turn the deprecation errors into warnings, so the build shouldn't fail on the openldap plugin (https://github.com/collectd/collectd/blob/master/src/openldap.c#L34). At least it doesn't on my Mac. Something else must be going wrong. Could it be https://travis-ci.org/collectd/collectd/jobs/491360834#L2904 (nproc not found)?

rubenk:
The clang build fails but the gcc build succeeds, which is strange because clang and gcc are the same on OSX. Speaking of that, can we make it so that we only build once?

rubenk:
I meant it is the same compiler:
```
~ > /usr/bin/gcc --version
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/c++/4.2.1
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
~ > /usr/bin/clang --version
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

rubenk:
We should at least `brew install coreutils` for nproc.

octo:
@rubenk With the latest commit, the build fails with:

```
Error: coreutils 8.29 is already installed
```

Is `nproc` available on OS X? Is it in the path?

Since the homebrew install is *very* slow, maybe this is a bit of premature optimization and removing it is the safest bet for now.

rubenk:
It is, and I expect Travis to have set the PATH to include /usr/local/bin.

On my machine:
```
~ > echo $PATH
/Users/ruben/.cargo/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/Applications/Wireshark.app/Contents/MacOS:/Users/ruben/bin:/Users/ruben/node_modules/.bin:/Users/ruben/go/bin:/usr/local/opt/go/libexec/bin:/usr/local/opt/llvm/bin
~ > ls -l /usr/local/bin/nproc
lrwxr-xr-x 1 ruben admin 36 Feb 11 16:31 /usr/local/bin/nproc -> ../Cellar/coreutils/8.30_2/bin/nproc
~ > brew ls coreutils | grep nproc
/usr/local/Cellar/coreutils/8.30_2/bin/nproc
/usr/local/Cellar/coreutils/8.30_2/bin/gnproc
/usr/local/Cellar/coreutils/8.30_2/libexec/gnubin/nproc
/usr/local/Cellar/coreutils/8.30_2/libexec/gnuman/man1/nproc.1
/usr/local/Cellar/coreutils/8.30_2/share/man/man1/nproc.1
/usr/local/Cellar/coreutils/8.30_2/share/man/man1/gnproc.1

```

I am on 10.14 however, and Travis runs 10.13, but I can't imagine this matters.

rubenk:
> Since the homebrew install is _very_ slow, maybe this is a bit of premature optimization and removing it is the safest bet for now.

The homebrew addon for Travis should be faster than manually calling brew: https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-macos

octo:
I have no idea what I'm doing, but it succeeds without `nproc` :tada:

And now we iterate :)

rubenk:
Perfect, let's just merge this 👍

octo:
I'll give the homebrew addon a go. Also, I really like packages to be alphabetically sorted ;)

rubenk:
Great. I have a few other improvements in mind, like running `make distcheck` instead of make && make check, but we can add these on top.

octo:
With the addon, installing packages took ~160 seconds, which feels like an improvement :)

PR #3057, opened 2019-02-05
collection3: Add Type for smart_temperature

cjmayo:
collection3: Add Type for smart_temperature

### DIFF
--- a/contrib/collection3/etc/collection.conf
+++ b/contrib/collection3/etc/collection.conf
@@ -618,6 +618,13 @@ GraphWidth 400
   RRDFormat "%5.1lf%%"
   DSName "value Signal quality"
 </Type>
+<Type smart_temperature>
+  DataSources value
+  DSName value Temp
+  RRDTitle "Temperature ({instance})"
+  RRDVerticalLabel "°Celsius"
+  RRDFormat "%4.1lf°C"
+</Type>
 <Type snr>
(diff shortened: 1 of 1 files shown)


cjmayo:
Copied from Type temperature.

I don't understand what the ChangeLog check is asking for. I've ended up with the commit title duplicated in the commit body, that must need to go.

octo:
Hi @cjmayo,

thank you very much for your PR! The "ChangeLog" check is very new, so it is in our best interest to document it well. Could you tell me what you find confusing so we can improve? Did you click the "Details" link yet?

Best regards,
—octo

cjmayo:
I did click on Details - I think the only truly ambiguous thing there is if the body should contain **only** the one-line description.
I managed to miss that it is was talking about the PR and not the commit, maybe I was rushing, maybe body made me think of commit - although it looks like that is what the GitHub API calls field, the help article:
https://help.github.com/articles/about-pull-requests/
calls it a summary.

Anyway, I've fixed the commit body. Let's see how the ChangeLog check goes...

octo:
Hi @cjmayo, thanks for your feedback! I created #3061 to hopefully cause less confusion.

The "ChangeLog:" line needs to go into the PR description, not the commits. Changes in contrib/ we don't usually mention, so this PR is going to be excempt.

PR #3058, opened 2019-02-05
collection3: Output graph.cgi debug as UTF-8

cjmayo:
collection3: Output graph.cgi debug as UTF-8

Ensure text and symbols e.g. degrees Celsius are shown as intended.

### DIFF
--- a/contrib/collection3/bin/graph.cgi
+++ b/contrib/collection3/bin/graph.cgi
@@ -140,10 +140,10 @@ sub main
   if (param ('debug'))
   {
     print <<HTTP;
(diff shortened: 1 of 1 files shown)


cjmayo:
I don't understand what the ChangeLog check is asking for. I've ended up with the commit title duplicated in the commit body, that must need to go.

octo:
I don't understand why the status is shown as "expected", the payload delivered to Github set the status to "failure". For reference, the payload ID I used to verify this is 6200458401.

cjmayo:
Fixed the commit message.

PR #3061, opened 2019-02-07
CONTRIBUTING.md: Improve wording around ChangeLog; fix example.

octo:


CONTRIBUTING.md
```diff
@@ -53,18 +53,18 @@ the mailing list have a tendency to fall through the cracks.
...
 ```
-Foo plugin: A specific issue people had has been fixed.
+ChangeLog: Foo plugin: A specific issue people had has been fixed.
 ```
 
-Start with "Foo plugin" to give the reader context for the information. Other
-common prefixes are "collectd" for the core daemon and "Build system". Use past
-tense and passive voice the for remainder, e.g. "a bug has been fixed", "a
-feature has been added".
+The summary should start with "Foo plugin" to give the reader context for the
```
faxm0dem: aren't you contradicting yourself here (Line 61)?

CONTRIBUTING.md
```diff
@@ -53,18 +53,18 @@ the mailing list have a tendency to fall through the cracks.
 
 ### ChangeLog
 
-All PRs need to have a one-line description in the initial pull request body.
+All PRs need to have a one-line summary in the initial pull description.
```
faxm0dem: I'd really stress here that it's the PR message, which can be different to the commit message:

```
The PR description can be different to the last commit's message
```

CONTRIBUTING.md
mrunge: Personally, I would leave it this way. It is pretty clear.

@faxm0dem would you think 
The information should follow the following example:
`ChangeLog: <plugin name>: descriptive text.`
Other than `plugin name` "collectd" for the core daemon....

octo: I don't think it's a contradiction, but I see how this wording may be confusing. I've changed it now; do you like this better?

CONTRIBUTING.md
octo: I gave explicit instructions on how to edit the PR description. If possibe, I'd like to avoid negative documentation, i.e. document where things do *not* go. Do you think an example screenshot of a ChangeLog line would be helpful to contributors?

octo:
Updated, PTAL :)

faxm0dem:
changelog text now faxmodem-foolproof!

octo:
Thanks for your feedback @faxm0dem!

PR #3065, opened 2019-02-08
changed time format to no longer lose microseconds

btoneill:
ChangeLog: rrdcached plugin: Time resolution has been improved to microseconds.

Have been sending an int to rrdcached for the epoch value, it supports microseconds. You want to use microseconds. 

See https://lists.oetiker.ch/pipermail/rrd-developers/2003-March/001007.html for we want this

### DIFF
--- a/src/rrdcached.c
+++ b/src/rrdcached.c
@@ -66,8 +66,8 @@ static int value_list_to_string(char *buffer, int buffer_len,
 
   memset(buffer, '\0', buffer_len);
 
-  time_t t = CDTIME_T_TO_TIME_T(vl->time);
-  int status = snprintf(buffer, buffer_len, "%lu", (unsigned long)t);
+  double t = CDTIME_T_TO_DOUBLE(vl->time);
+  int status = snprintf(buffer, buffer_len, "%f", t);
   if ((status < 1) || (status >= buffer_len))
     return -1;
   int offset = status;
--- a/src/rrdcached.c
+++ b/src/rrdcached.c
@@ -66,8 +66,8 @@ static int value_list_to_string(char *buffer, int buffer_len,
 
   memset(buffer, '\0', buffer_len);
 
-  double t = CDTIME_T_TO_DOUBLE(vl->time);
-  int status = snprintf(buffer, buffer_len, "%.6f", t);
+  int status =
+      snprintf(buffer, buffer_len, "%.6f", CDTIME_T_TO_DOUBLE(vl->time));
   if ((status < 1) || (status >= buffer_len))
     return -1;
   int offset = status;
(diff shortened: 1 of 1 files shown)


btoneill:
Doesn't rrdcached itself require rrdtool 1.4?

I'll get the log line fixed.

Thanks,
Brian

octo:
Hey, I didn't see that the email you referenced was from 2003. If this feature has been in RRDtool for ~16 years, then the option is not required. Since you didn't update the description yet, I'm going to go ahead and do that.

Thanks and best regards,
—octo

src/rrdcached.c
```diff
@@ -74,8 +74,9 @@ static int value_list_to_string(char *buffer, int buffer_len,
 
   memset(buffer, '\0', buffer_len);
 
-  t = CDTIME_T_TO_TIME_T(vl->time);
-  status = snprintf(buffer, buffer_len, "%lu", (unsigned long)t);
+  t = CDTIME_T_TO_DOUBLE(vl->time);
+
+  status = snprintf(buffer, buffer_len, "%f", t);
```
octo: The PR description and the referenced email both talk about microseconds specifically, but this format may in fact be more precise than that. Please print as `%.6f` instead.

btoneill:
My understanding of the cdtime_t is that it only goes to microseconds. If I
am misunderstanding I can change it.

From utils_time.h:
/*
 * "cdtime_t" is a 64bit unsigned integer. The time is stored at a 2^-30
second
 * resolution, i.e. the most significant 34 bit are used to store the time
in
 * seconds, the least significant bits store the sub-second part in
something
 * very close to nanoseconds. *The* bing time in this
 * manner is that comparing times and calculating differences is as simple
as
 * it is with "time_t", i.e. a simple integer comparison / subtraction
works.
 */

octo:
> the sub-second part is
something very close to nanoseconds

One microsecond is 1000 nanoseconds.

btoneill:
I think I have updated with the requested change.

src/rrdcached.c
```diff
@@ -74,8 +74,9 @@ static int value_list_to_string(char *buffer, int buffer_len,
 
   memset(buffer, '\0', buffer_len);
 
-  t = CDTIME_T_TO_TIME_T(vl->time);
-  status = snprintf(buffer, buffer_len, "%lu", (unsigned long)t);
+  t = CDTIME_T_TO_DOUBLE(vl->time);
```
octo: `t` has the wrong type now (should be `double`, is `time_t`. I'd recommend to remove `t` completely and inline the conversion:

```c
status = snprintf(buffer, buffer_len, "%.6f", CDTIME_T_TO_DOUBLE(vl->time));
```

btoneill:
Do, i forgot that in my change there. My original change was in a different
branch and I copied it over.

I'l fix.

octo:
Thanks again for your PR @btoneill!

octo:
Rebased on *master*

PR #3069, opened 2019-02-11
Added feature to tail_csv to allow it to only submit the last line of the csv file

btoneill:
ChangeLog: tail_csv plugin: Added LastLineOnly option to only submit the last line found in the CSV file.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -8560,6 +8560,7 @@ B<Synopsis:>
        Interval 600
        Collect "snort-dropped"
        #TimeFrom 0
+       LastLineOnly false
    </File>
  </Plugin>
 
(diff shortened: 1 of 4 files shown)


src/collectd.conf.pod
```diff
@@ -8635,6 +8636,10 @@ Rather than using the local time when dispatching a value, read the timestamp
 from the field with the zero-based index I<Index>. The value is interpreted as
 seconds since epoch. The value is parsed as a double and may be factional.
 
+=item B<LastLineOnly> I<true>|I<false>
+
+Rather than sending all lines in the CSV file, we only enter the last one
```
anaudx: IMHO, the below proposed text would be more clear:
"If set to true, only the last line in the CSV file will be read. If set to false (the default), all lines will be read."

dago:
@btoneill I know it has been some time since you contributed this PR; but would you mind rebasing this PR, resolving the conflict and the code review?

Issue #2958, opened 2018-10-17
RFC: Automatic merging


octo:
CC: @rubenk, @rpv-tomsk, @tokkee 

Hi team,

as you may be aware, the `Automerge` label will cause @collectd-bot to automatically merge pull requests when certain conditions are met (CI succeeds, no pending reviews, …). I'm thinking about changing the behavior a bit and would like your thoughts on this:

1.  Require one "approved" review from one of the maintainers *instead* of the `Automerge` label. I.e. as soon as a maintainer approves a change, it will automatically get merged (assuming CI is successful).
1.  As a second step, I'd like to integrate with Github's [code ownership](https://blog.github.com/2017-07-06-introducing-code-owners/) concept. When, for example, a pull request only touches `src/dpdk{events,stat}.c`, then a (positive) review from one of the Intel folks working on this should be sufficient to trigger the merge. We should encourage contributors to add themselves as owners of new plugins.

**Caveats:**

*   You can now "resolve" code review comments in the UI, but as far as I see it's not possible to distinguish "resolved" from "unresolved" comments in the API. That means you'd have to ensure all comments are fixed before approving.
*   It's unclear how much of the code ownership we have to implement ourselves. I'm hoping that with the "Require review from Code Owner" option enabled a PR is simply marked as "not mergable" unless the condition is met. But the API docs are extremely vague with details like this.

What do you think?

Best regards,
—octo

rpv-tomsk:
Hi!

>Require one "approved" review from one of the maintainers instead of the Automerge label. I.e. as soon as a maintainer approves a change, it will automatically get merged (assuming CI is successful).

As for me, I used "Automerge" label when I have some small change to commit, but I want PR to pass CI checks. So, I make PR, give it "Automerge" label and bot cares about PR, not me. 

Next time, if I see what CI has failed, I will resolve that issue, or take no more care if CI succeeds. 
(Ok, I also check, if CI of `master` has not failed after merge of succeeded PR check, that can happen, unfortunately.)

>Require one "approved" review from one of the maintainers instead of the Automerge label. I.e. as soon as a maintainer approves a change, it will automatically get merged (assuming CI is successful).

At practice, there was cases when PR has approved maintainer review, but was not merged by anybody.
Proposed change will cover such cases, but most of the time, I think, when maintainer reviews PR, it already has CI checks completed, so he can just do merge directly. If maintainer doesn't want to take responsibility of merge, then he will not approve PR at all after such change. 

[...]

PR #3072, opened 2019-02-12
Add a unit test for the network plugin.

octo:
This is not the best of tests; it was just something I found lying around in my workspace and thought I'd upload it here in case somebody finds it useful.

ChangeLog: network plugin: A unit test has been added.


Makefile.am
```diff
@@ -1370,6 +1370,14 @@ network_la_CPPFLAGS += $(GCRYPT_CPPFLAGS)
 network_la_LDFLAGS += $(GCRYPT_LDFLAGS)
 network_la_LIBADD += $(GCRYPT_LIBS)
 endif
+
+test_plugin_network_SOURCES = src/network_test.c \
+			      src/daemon/configfile.c \
+			      src/daemon/types_list.c
+test_plugin_network_CPPFLAGS = $(AM_CPPFLAGS)
```
rubenk: This is not needed since you don't override the flags.

rubenk:
Looks good, thanks! The build failure looks real though, can you resolve that?

Makefile.am
octo: Fixed, thanks!

Makefile.am
```diff
@@ -1371,12 +1371,18 @@ network_la_LDFLAGS += $(GCRYPT_LDFLAGS)
...
-			      src/daemon/types_list.c
-test_plugin_network_CPPFLAGS = $(AM_CPPFLAGS)
-test_plugin_network_LDFLAGS = $(PLUGIN_LDFLAGS)
-test_plugin_network_LDADD = liboconfig.la libplugin_mock.la libmetadata.la
+test_plugin_network_SOURCES = \
+	src/network_test.c \
+	src/utils_fbhash.c \
+	src/daemon/configfile.c \
+	src/daemon/types_list.c
+test_plugin_network_LDFLAGS = $(PLUGIN_LDFLAGS) $(GCRYPT_LDFLAGS)
```
rubenk: You'll need `test_plugin_network_CPPFLAGS = $(AM_CPPFLAGS) $(GCRYPT_CPPFLAGS)` as well I think?

octo: Good catch, thanks!

rpv-tomsk:
Heh, I will not merge this, even if I want.
If something goes wrong after merge master, I will not be able to solve that issue.

 #2958

rpv-tomsk:
cc @mrunge 

>If something goes wrong after merge master, I will not be able to solve that issue.

That is really funny, I know:

```
Undefined			first referenced
 symbol  			    in file
inet_pton                           src/test_plugin_network-network_test.o  (symbol belongs to implicit dependency /lib/libnsl.so.1)
inet_ntop                           src/test_plugin_network-network_test.o  (symbol belongs to implicit dependency /lib/libnsl.so.1)
ld: fatal: symbol referencing errors. No output written to test_plugin_network
gmake[2]: *** [Makefile:6163: test_plugin_network] Error 2
```

Should @rjablonx  jump up and decide failed CI issue ?

rpv-tomsk:
I wish you good luck.

rpv-tomsk:
Ok, when I was "trusted contributor" I made PR exactly to fix CI problems.
I do that mostly because I can do PR, merge it without any delay, check for master branch CI results, apply new fixes and merge these fixes without any delay again.

For now, if any "untrusted contributor" will try to solve such issue, he should publish PR and sit and wait for somebody notice. Maybe stars come together and somebody pay attention on it to merge it to master.

When CI will fail again, the cycle (with mandatory sit-and-wait stage) will have to repeat again.

Good luck.

rpv-tomsk:
cc @faxm0dem @sunkuranganath

sunkuranganath:
@rpv-tomsk you have a unique way of communicating the issues :D 

I believe issues like these are not uncommon in widely used open source communities. However the resolution of issues like rights, CI, permissions, requests, etc., gets resolved fairly soon due to good communication.
We all work on this in our own time on top of our other responsibilities and want to make this work. 
As @mrunge describes in #2958 once we get to fix the issues I hope things will get a bit easier.

Please note that @mrunge @tokkee and myself are working on a Collectd release proposal that touches upon these things and get some direction/method to this madness. Will send out the draft in next few days as we are almost done with our updates. Hope to get your feedback to make this easy for you and rest of the community and establish some sense.

rpv-tomsk:
>We all work on this in our own time on top of our other responsibilities and want to make this work.

I'm speaking about the same - our's time is the most valuable resource, so we should not spent it on fighting administrative and CI issues.  Also, I think, nobody of us wants to spent time implementing feature and then throw out the result of the work just because lack of response. Yes, I know that is the problem 'code owners' tries to solve (add plugin owners ability to merge theirs changes), but that breaks things and puts them upside down.

rpv-tomsk:
Again: When I merge somebody's PR, I'm responsible for that merge. That is idea and goal of review process.

But in current situation, I _have no permission_ to fix issues, caused by that merge. 
So why do you ask me to be "(un)trusted maintainer"?

Another cause of this PR problem is different set of CI checks applied to 'master' and 'PR' branches.
Even if 'PR' branch check succeded, 'master' branch check can fail, which is what happened.

All these are administrative issues, which needs to be solved, but lasts for months and years.

PR #3074, opened 2019-02-13
Re-implement CI using cirrus-ci

mfournier:
Following the decision(*) to sunset our old Jenkins-based CI, here's a first attempt to provide a replacement using cirrus-ci's "free for opensource projects" plan. Not all checks we had with Jenkins have been re-implemented (yet), and the plan is to focus first on building/testing the code & submitted patches. ie: we'll take care of the nightly builds later.

One major difference is that all the tests are run in docker images we maintain (see https://github.com/collectd/ci-docker/blob/master/README.md). This will be much less cumbersome to work with than the VM templates we had previously: anyone with the docker engine installed can run tests locally, and build new images. And anyone can submit patches to https://github.com/collectd/ci-docker/ to improve existing images or add support for new linux distros.
Also, folks can enable cirrus-ci (via https://github.com/marketplace/cirrus-ci) on their own fork of collectd, and run the tests on their stuff before submitting a PR to the main repo.

For the record, cirrus-ci was chosen for it's support for freebsd, windows and macosx, although the config submitted here doesn't make use of these platforms yet. We initially considered(*) using cirrus-ci only for these platforms, and use travis-ci for the linux-based environments. But cirrus-ci's native support for custom docker images make integration easier than travis-ci.

@octo once this is merged, you'll have to install the cirrus-ci application from gihub's marketplace to the main repo. And once we're happy with the results, maybe consider making these checks mandatory for PRs ?

(*) discussed offline, during the contributor meetup in Brussels

PS: I'd like to grab the opportunity to thanks @exoscale and @digitalocean for sponsoring our CI for many years !

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,251 @@
+env:
+  LANG: C
+  DEFAULT_CONFIG_OPTS: --enable-debug --without-libstatgrab
+
+
+###
+# make distcheck and other sanity checks
+#
+relase_ready_task:
+  container:
+    image: collectd/ci:stretch_amd64
+  lint_script:
+    - /checks/check-bashisms.sh
+    - /checks/check-pod.sh
+  configure_script:
+    - ./build.sh
+    - ./configure $DEFAULT_CONFIG_OPTS
+  checks_script:
+    - make -s distcheck
+
+###
+# Default toolchain and build flags use in deb packages, on a range of Debian
+# and Ubuntu releases (+ Debian/unstable)
+# Most should succeed, and PRs shouldn't break them.
+#
+debian_default_toolchain_task:
+  matrix:
+    - allow_failures: true
+      skip_notifications: true
+      container:
+        image: collectd/ci:wheezy_amd64 # TODO: fix this platform
+    - allow_failures: false
+      container:
+        image: collectd/ci:jessie_amd64
+    - allow_failures: false
+      container:
+        image: collectd/ci:stretch_amd64
+    - allow_failures: false
+      container:
+        image: collectd/ci:stretch_i386
+    - allow_failures: false
+      container:
+        image: collectd/ci:trusty_amd64
+    - allow_failures: false
+      container:
+        image: collectd/ci:xenial_amd64
+    # debian/unstable is expected to fail
+    - allow_failures: true
+      skip_notifications: true
+      only_if: $CIRRUS_BRANCH == 'master'
+      container:
+        image: collectd/ci:sid_amd64
+  configure_script:
+    - ./build.sh
+    - gcc --version
+    - >
+      ./configure CC=gcc $DEFAULT_CONFIG_OPTS
+      CFLAGS="$(dpkg-buildflags --get CFLAGS)"
+      CPPLAGS="$(dpkg-buildflags --get CPPFLAGS)"
+      LDFLAGS="$(dpkg-buildflags --get LDFLAGS)"
+  build_script:
+    - make -sk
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -0,0 +1,251 @@
+env:
+  LANG: C
+  DEFAULT_CONFIG_OPTS: --enable-debug --without-libstatgrab
```
rubenk: Could you add --disable-dependency-tracking while you're at it? That shave about 30 seconds of a build on my machine.

.cirrus.yml
```diff
@@ -0,0 +1,251 @@
...
+
+
+###
+# make distcheck and other sanity checks
+#
+relase_ready_task:
```
rubenk: s/relase/release/ ?

.cirrus.yml
```diff
@@ -0,0 +1,251 @@
...
+  container:
+    image: collectd/ci:stretch_amd64
+  lint_script:
+    - /checks/check-bashisms.sh
+    - /checks/check-pod.sh
+  configure_script:
+    - ./build.sh
+    - ./configure $DEFAULT_CONFIG_OPTS
+  checks_script:
+    - make -s distcheck
```
rubenk: you'll want to add the configure options here too, so `make -s distcheck DISTCHECK_CONFIGURE_FLAGS="${DEFAULT_CONFIG_OPTS}"`

.cirrus.yml
```diff
@@ -0,0 +1,251 @@
...
+
+###
+# Default toolchain and build flags use in deb packages, on a range of Debian
```
rubenk: used

.cirrus.yml
```diff
@@ -0,0 +1,251 @@
...
+    - gcc --version
+    - >
+      ./configure CC=gcc $DEFAULT_CONFIG_OPTS
+      CFLAGS="$(dpkg-buildflags --get CFLAGS)"
+      CPPLAGS="$(dpkg-buildflags --get CPPFLAGS)"
+      LDFLAGS="$(dpkg-buildflags --get LDFLAGS)"
+  build_script:
+    - make -sk
+  tests_script:
+    - make -sk check
```
rubenk: do make distcheck here too? that will catch a few more things sooner.

rubenk:
Great work @mfournier! I left a few small nits but otherwise LGTM.

.cirrus.yml
mfournier: My understanding is that `make distcheck` is equivalent to running `make install && make uninstall` into a temporary directory, right ? In this case, the platform doesn't really make a different, or does it ?

Unlike the jenkins-based CI, here we run `release_ready_task` on any branch, PRs or not. So a `make distcheck` is always run.

So unless my understanding is incorrect, running `distcheck` more than once wouldn't bring anything. But happy to change this if I'm wrong ;-)

mfournier:
Thanks for the reviews ! I changed the details @rubenk pointed out.
Here's the output of a run on this updated patchset: http://cirrus-ci.com/build/5752304956342272

octo:
Awesome, thanks Marc! I'll let Ruben take another look and then merge it and install the app at the same time.

.cirrus.yml
rubenk: Yes it is equivalent. It could matter if make installs behaves differently on a different platform, for instance if we install specific files for FreeBSD, but I don't think we do.

octo:
FYI, I installed the Cirrus CI app from the marketplace.

PR #3084, opened 2019-02-18
WIP: Perl plugin: Refactor code ensuring aTHX exists into a new function

guillemj:
All these instances are doing the same, let's refactor into a single
function.

ChangeLog: Perl plugin: The code ensuring the ithread context exists got refactored.

### DIFF
--- a/src/perl.c
+++ b/src/perl.c
@@ -1274,6 +1274,20 @@ static c_ithread_t *c_ithread_create(PerlInterpreter *base) {
   return t;
 } /* static c_ithread_t *c_ithread_create (PerlInterpreter *) */
 
+static void c_ithread_ensure_context(void) {
+  c_ithread_t *t = NULL;
+  dTHX;
+
+  if (aTHX != NULL)
+    return;
+
+  pthread_mutex_lock(&perl_threads->mutex);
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Hi,

Are you sure this change is completely correct? 
Did you checked, compiled and carefully checked your changes? 

I doubt a little, because your newly-added function `c_ithread_ensure_context` declares new variable by using `dTHX` macro, while old variables are also defined in blocks, where you want to do 'cleanup'.

As result, as I see, the `aTHX` will be uninitialized in all these blocks, and I expect this updated code will crash.

guillemj:
Hey! Sorry yeah, I was fishing out old changes I've had locally for a while, from when we were using collectd, to avoid losing them. And i thought this one had been tested at the time, but apparently this was a WIP branch. Should have been obvious when it failed to build on the CI, but I guess I thought it was due to changes to the codebase. :(

I'll take another look and fix and test this when I've got a moment.

rpv-tomsk:
>I'll take another look and fix and test this when I've got a moment.

Ok, then please open new PR. 

Thanks for your work on Collectd.

PR #3091, opened 2019-02-19
Reduce the nr of allocations when parsing types.db

rubenk:
We allocate a number of times for each line in types.db.
Since plugin_register_dataset() makes a copy of the dataset anyway, we
can create the dataset on the stack instead of the heap.

This reduces the number of calloc calls on my system from 624 to 337
when running collectd -T with the default types.db.

Changelog: reduce the nr of allocations when parsing types.db


rubenk:
Thanks for the review and merging Pavel!

PR #3092, opened 2019-02-19
Don't hide errors when creating manpage

rubenk:
Apparently the default behaviour for pod2man changed at some point and
now it exits with a syntax error instead of exiting successfully with a
<POD ERRORS> section.

Now we do display the syntax issue:
[ruben@localhost collectd]$ make -j
  POD2MAN  src/collectd.conf.5
src/collectd.conf.pod around line 10361: You forgot a '=back' before '=head2'
POD document had syntax errors at /usr/bin/pod2man line 69.
make: *** [Makefile:9641: src/collectd.conf.5] Error 255

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2109,20 +2109,10 @@ am__v_POD2MAN_C_0 = @echo "  POD2MAN " $@;
 am__v_POD2MAN_C_1 =
 
 .pod.1:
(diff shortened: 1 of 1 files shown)


rubenk:
Hmm, older versions don't support the --errors flags. --errors=die is the default behaviour in new versions so I'll remove that, and we'll let CI catch the errors on newer platforms.

PR #3095, opened 2019-02-20
Build system: enable the "df" plugin when getmntent_r() is available.

octo:
Looks like the check result used to be stored in a variable of a different name.

I hope that this will fix the Cirrus CI builds.

ChangeLog: Build system: The "df" plugin is now built when `getmntent_r()` is available.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6308,7 +6308,7 @@ then
 	plugin_df="yes"
 fi
 
-if test "x$c_cv_have_getmntent_r" = "xyes"
+if test "x$have_getmntent_r" = "xyes"
 then
 	plugin_df="yes"
 fi


rubenk:
Ah oops! You could also use ac_cv_func_getmntent_r but it shouldn't matter.

octo:
I think explicitly initialized variables are easier to read.

rubenk:
I see something else going wrong too:
```
configure:17669: checking whether getmntent takes one argument
configure:17695: gcc -c -ggdb -Og -Wall  conftest.c >&5
conftest.c:132:20: fatal error: ./src/utils_mount.h: No such file or directory
  132 |           #include "./src/utils_mount.h"
      |                    ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

No idea why we include that header, we should probably remove that.

octo:
Probably fall-out from 6378ec288f34ff250b2971a1452338a2b34c240a. I'll see why that was included there.

octo:
`src/utils/mount/mount.h` (previously `src/utils_mount.h`) includes a long list of file system related headers. It's included in the configure script so that that configure script is guaranteed to behave like the (header included from the) df plugin.

rubenk:
Got it, thanks.

PR #3096, opened 2019-02-20
pod2man: remove --errors argument

rubenk:
Older versions of pod2man don't have this

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2109,10 +2109,10 @@ am__v_POD2MAN_C_0 = @echo "  POD2MAN " $@;
 am__v_POD2MAN_C_1 =
 
 .pod.1:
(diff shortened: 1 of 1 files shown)


octo:
Should we revert 8a5bcd989811719d0bc618fed3d6bc3552982da1 instead? I'd argue its good enough if the CI buidlers catch POD errors, but we should be informed somehow.

rubenk:
Before 8a5bcd989811719d0bc618fed3d6bc3552982da1 we were not informed of failures at all, probably because of the combination of || true, 2>&1 >/dev/null and a missing <POD ERRORS> section. This happened in https://github.com/collectd/collectd/pull/3019 and makes the failure hard to debug. It only failed later during make install because the man file wasn't created.

Not sure what the best solution is though. We could add autoconf checks to detect the error handling behaviour of pod2man?

octo:
Fair point. Let's go with this and make sure that syntax errors are reported by *some* builders at least.

PR #3107, opened 2019-02-21
lvm plugin: turn deprecation error into a warning

rubenk:
When you include lmv2app.h it warns that this is deprecated and you have
to use D-BUS.

This warning turns into an error since we build with -Werror by default.
We know using the lvm2 lib is deprecated but until someone writes a
replacement using D-BUS it's all we have.

So turn the error into a warning.

### DIFF
--- a/src/lvm.c
+++ b/src/lvm.c
@@ -26,7 +26,15 @@
 #include "plugin.h"
 #include "utils/common/common.h"
 
+#if GNUC > 4
+#pragma GCC diagnostic pop
+#pragma GCC diagnostic warning "-Wcpp"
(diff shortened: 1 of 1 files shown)


rubenk:
Hi Marc! I'd swear I'd fixed that issue, I'll look into it.

As for the second option, is there any platform at the moment where it doesn't spew a deprecation warning? If not, perhaps we should just remove the whole plugin?

mrunge:
@rubenk did you had a chance to get back to this?

dago:
LVM support was already removed in #3309

PR #3108, opened 2019-02-21
Cirrus improvements

rubenk:


rubenk:
Great idea, especially with Alpine being popular as a base for docker containers. And it's always nice to test against an alternative libc implementation.

PR #2908, opened 2018-08-24
Listval with state selection addon

saimon46:
This pull request is to extend the actual LISTVAL command providing also a parameter to optionally filter by a given state.

This is the output without any parameter:
```
sbrundu@sbrundu-TTL-TEKNOPRO:~$ sudo socat - UNIX-CLIENT:/opt/collectd/var/run/collectd-unixsock 
LISTVAL
19 Values found
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-docker0/if_dropped
1531484451.014 sbrundu-TTL-TEKNOPRO/interface-docker0/if_errors
1531484451.014 sbrundu-TTL-TEKNOPRO/interface-docker0/if_octets
1531484451.014 sbrundu-TTL-TEKNOPRO/interface-docker0/if_packets
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-eno1/if_dropped
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-eno1/if_errors
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-eno1/if_octets
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-eno1/if_packets
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-lo/if_dropped
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-lo/if_errors
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-lo/if_octets
1531484451.015 sbrundu-TTL-TEKNOPRO/interface-lo/if_packets
1531484451.014 sbrundu-TTL-TEKNOPRO/load/load
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-buffered
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-cached
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-free
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-slab_recl
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-slab_unrecl
1531484451.014 sbrundu-TTL-TEKNOPRO/memory/memory-used
```

This instead filtering the state ERROR:
```
sbrundu@sbrundu-TTL-TEKNOPRO:~$ sudo socat - UNIX-CLIENT:/opt/collectd/var/run/collectd-unixsock
LISTVAL state=ERROR
1 Values found
[...]

PR #3114, opened 2019-02-26
 Listval with state selection addon #2 to rebase

saimon46:
ChangeLog: Collectd: test


dago:
This look identical to #2908 so I suggest to close this one and continue with #2908 instead because of the conversation history.

Issue #3082, opened 2019-02-18
loading virt plugin segfaults


rubenk:
*   Version of collectd: master (f0a2bef2)
*   Operating system / distribution: Fedora rawhide

## Expected behavior

Normal output of collectd -T

## Actual behavior

```
[ruben@rubenkerkhof collectd]$ gdb --args ~/opt/collectd/sbin/collectd -T
GNU gdb (GDB) Fedora 8.2.50.20190120-15.fc30
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /home/ruben/opt/collectd/sbin/collectd...
(gdb) run
Starting program: /home/ruben/opt/collectd/sbin/collectd -T
Missing separate debuginfos, use: dnf debuginfo-install glibc-2.29-7.fc30.x86_64
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
option = ReadThreads; value = -1;
option = Hostname; value = localhost;
option = BaseDir; value = /home/ruben/opt/collectd/var/lib/collectd;
option = PIDFile; value = /home/ruben/opt/collectd/var/run/collectd.pid;
Done parsing `/home/ruben/opt/collectd/share/collectd/types.db'
Created new plugin context.
[2019-02-18 12:32:13] plugin_load: plugin "logfile" successfully loaded.
[2019-02-18 12:32:13] type = logfile, key = LogLevel, value = debug
[2019-02-18 12:32:13] type = logfile, key = File, value = STDOUT
[2019-02-18 12:32:13] type = logfile, key = Timestamp, value = false
type = logfile, key = PrintSeverity, value = true
[info] plugin_load: plugin "cpu" successfully loaded.
[info] plugin_load: plugin "interface" successfully loaded.
[info] plugin_load: plugin "virt" successfully loaded.
[...]

PR #3116, opened 2019-02-26
virt plugin: Fix segfaults on shutdown

rpv-tomsk:
If connection has failed then `lv_init' returns early without initializing `notif_thread'.
This causes segfault due to incorrect `pthread_join()` call on Collectd shutdown.

Closes: #3082 

Changelog: virt plugin: Fix segfaults on shutdown

cc @fromanirh
cc @rubenk

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -2055,6 +2055,8 @@ static int start_event_loop(virt_notif_thread_t *thread_data) {
     return -1;
   }
 
+  DEBUG(PLUGIN_NAME " plugin: starting event loop");
+
   virt_notif_thread_set_active(thread_data, 1);
   if (pthread_create(&thread_data->event_loop_tid, NULL, event_loop_worker,
                      thread_data)) {
@@ -2070,6 +2072,9 @@ static int start_event_loop(virt_notif_thread_t *thread_data) {
 
 /* stop event loop thread and deregister callback */
 static void stop_event_loop(virt_notif_thread_t *thread_data) {
+
+  DEBUG(PLUGIN_NAME " plugin: stopping event loop");
+
   /* Stopping loop */
   if (virt_notif_thread_is_active(thread_data)) {
     virt_notif_thread_set_active(thread_data, 0);
@@ -2311,8 +2316,6 @@ static int lv_init(void) {
 
   if (lv_connect() != 0)
     return -1;
-
-  DEBUG(PLUGIN_NAME " plugin: starting event loop");
 
   if (!persistent_notification) {
     virt_notif_thread_init(&notif_thread);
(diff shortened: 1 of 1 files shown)


src/virt.c
```diff
@@ -2059,7 +2059,9 @@ static int start_event_loop(virt_notif_thread_t *thread_data) {
   if (pthread_create(&thread_data->event_loop_tid, NULL, event_loop_worker,
                      thread_data)) {
     ERROR(PLUGIN_NAME " plugin: failed event loop thread creation");
+    virt_notif_thread_set_active(thread_data, 0);
```
fromanirh: good catch. I wonder if we should just initialize notif_thread before attempting to connect.

rpv-tomsk:
cc @octo 

Hi, there is a problem with Travis CI on this PR.

Thanks.

src/virt.c
anaudx: Wouldn't it be equivalent to only move the above instruction `virt_notif_thread_set_active(thread_data, 1);` just after `if (pthread_create(...)) {...}` statement ? The value of `thread_data->is_active` must be `false` before this statement because function `start_event_loop` is called after `virt_notif_thread_init`.

rpv-tomsk: Thanks for your review!

I think it is not equivalent. Started thread will check `is_active` and it has to be `true` on its start.

anaudx: I just noticed BTW in `lv_shutdown` that line `DEBUG(PLUGIN_NAME " plugin: stopping event loop");` should be inside the next `if (!persistent_notification)` statement.

rpv-tomsk: >event loop....

The same is about `starting event loop`.
I think, if change this, it is better to put these debug messages into `start_event_loop() / stop_event_loop()` then.

anaudx: Sorry @rpv-tomsk , I didn't see your answers to my above comments and I removed one of them to start a code review. Anyway, I agree with both your answers. Debug messages can be moved to mentioned functions.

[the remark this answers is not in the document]
anaudx: Please ignore this comment as regards to above comment https://github.com/collectd/collectd/pull/3116#discussion_r264725035

rpv-tomsk:
Ok, lets try to restart Travis CI check. )

anaudx:
+1

anaudx:
AFAIK, my approval should automatically merge this PR but this did not happen. I have no other possibility to merge it.

rpv-tomsk:
Thanks, guys! )

kkepka:
I guess to merge this we would need help from any other maintainer (like you did merge #3104 ), as it looks like _approve_ from _plugin owner_ doesn't have merge power at this moment.

mrunge:
Thank you for you work here.  I don't know why the approval did not merge the code here, based on the feedback in this thread, I'm going to do that.

rpv-tomsk:
>I guess to merge this we would need help from any other maintainer

I am disappointed with these changes that break and stop the development process.

mrunge:
It may be required to have the automerge label for these kind of PRs to be approved by CODEOWNERS?

anaudx:
The hint displayed over my approval icon says `Only reviews by reviewers with write access count toward mergeability`, which suggests that I (and following approvers of this PR) am missing write access.

PR #3117, opened 2019-03-01
virt: Fix parsing config file options

rjablonx:
ChangeLog: virt: Fix parsing config file options

This PR fixes issues related to parsing configuration options for virt plugin:
- fixed domain filtering option
- fixed interface filtering option
- added return statement to properly finish successful parsing
 ExtraStats' option value - without that, parsing of config file will
 fail if 'ExtraStats' option is used in the collectd config file
 because return code is the one for unrecognized parameter key.

- fixed parsing of RefreshInterval - previously error was thrown by
  mistake if value was not written between double quotes(i.e. "60" was
  fine but 60 was not)
- previously validation for boolean option value(such as value of
  'IgnoreSelected' parameter) in virt plugin was missing and bad input
  in config file could be unnoticed by user(in config file, boolean
  value is parsed from its string representation). Now validation is
  added and error is thrown whenever value can't be parsed directly as
  'true'|'yes'|'on' or 'false'|'no'|'off'.
- error messages have been added in all places where
   parsing failed in order to notify the user about the reason of
   collectd initialization failure.
- updated collectd.conf.pod with more info about filtering options

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -9252,13 +9252,51 @@ surrounded by I</.../> and collectd was compiled with support for regexps.
 
 The default is to collect statistics for all domains and all their devices.
 
-Example:
+B<Note:> B<BlockDevice> and B<InterfaceDevice> options are related to
+corresponding B<*Format> options. Specifically, B<BlockDevice> filtering depends
+on B<BlockDeviceFormat> setting - if user wants to filter block devices by
+'target' name then B<BlockDeviceFormat> option has to be set to 'target' and
+B<BlockDevice> option must be set to a valid block device target
+name("/:hdb/"). Mixing formats and filter values from different worlds (i.e.,
+using 'target' name as B<BlockDevice> value with B<BlockDeviceFormat> set to
+'source') may lead to unexpected results (all devices filtered out or all
+visible, depending on the value of B<IgnoreSelected> option).
+Similarly, option B<InterfaceDevice> is related to B<InterfaceFormat> setting
+(i.e., when user wants to use MAC address as a filter then B<InterfaceFormat>
+has to be set to 'address' - using wrong type here may filter out all of the
+interfaces).
+
+B<Example 1:>
+
+Ignore all I<hdb> devices on any domain, but other block devices (eg. I<hda>)
+will be collected:
 
  BlockDevice "/:hdb/"
  IgnoreSelected "true"
+ BlockDeviceFormat "target"
 
-Ignore all I<hdb> devices on any domain, but other block devices (eg. I<hda>)
-will be collected.
+B<Example 2:>
+
+Collect metrics only for block device on 'baremetal0' domain when its
+'source' matches given path:
+
+ BlockDevice "baremetal0:/var/lib/libvirt/images/baremetal0.qcow2"
+ BlockDeviceFormat source
+
+As you can see it is possible to filter devices/interfaces using
+various formats - for block devices 'target' or 'source' name can be
+used.  Interfaces can be filtered using 'name', 'address' or 'number'.
+
+B<Example 3:>
+
+Collect metrics only for domains 'baremetal0' and 'baremetal1' and
+ignore any other domain:
+
+ Domain "baremetal0"
+ Domain "baremetal1"
+
+It is possible to filter multiple block devices/domains/interfaces by
+adding multiple filtering entries in separate lines.
 
 =item B<BlockDeviceFormat> B<target>|B<source>
 
@@ -9288,6 +9326,11 @@ Setting C<BlockDeviceFormat target> will cause the I<type instance> to be set
 to C<sda>.
 Setting C<BlockDeviceFormat source> will cause the I<type instance> to be set
 to C<var_lib_libvirt_images_image1.qcow2>.
+
+B<Note:> this option determines also what field will be used for
+filtering over block devices (filter value in B<BlockDevice>
+will be applied to target or source). More info about filtering
+block devices can be found in the description of B<BlockDevice>.
 
 =item B<BlockDeviceFormatBasename> B<false>|B<true>
 
@@ -9338,6 +9381,11 @@ B<address> means use the interface's mac address. This is useful since the
 interface path might change between reboots of a guest or across migrations.
 
 B<number> means use the interface's number in guest.
+
+B<Note:> this option determines also what field will be used for
+filtering over interface device (filter value in B<InterfaceDevice>
+will be applied to name, address or number).  More info about filtering
+interfaces can be found in the description of B<InterfaceDevice>.
 
 =item B<PluginInstanceFormat> B<name|uuid|metadata|none>
 
--- a/src/virt.c
+++ b/src/virt.c
@@ -1155,6 +1155,11 @@ static int lv_init_ignorelists() {
 /* Validates config option that may take multiple strings arguments.
  * Returns 0 on success, -1 otherwise */
 static int check_config_multiple_string_entry(const oconfig_item_t *ci) {
+  if (ci == NULL) {
+    ERROR(PLUGIN_NAME " plugin: ci oconfig_item can't be NULL");
+    return -1;
+  }
+
   if (ci->values_num < 1) {
     ERROR(PLUGIN_NAME
           " plugin: the '%s' option requires at least one string argument",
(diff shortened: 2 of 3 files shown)


src/virt.c
```diff
@@ -2687,12 +2768,26 @@ static void free_domains(struct lv_read_state *state) {
 
 static int add_domain(struct lv_read_state *state, virDomainPtr dom,
                       bool active) {
-
   int new_size = sizeof(state->domains[0]) * (state->nr_domains + 1);
+  const char *domname = virDomainGetName(dom);
```
rpv-tomsk: IMHO, that is not the best solution to put ignorelist check into `add_domain`.
It is not intuitive that `add_domain` is smart and can decide to skip adding.

I think that will be better to move this check into another function and call it before `add_domain`.

Can you please update plugin code?

Many thanks!

src/virt.c
```diff
@@ -1138,12 +1151,25 @@ static int lv_init_ignorelists() {
...
+    if (!IS_TRUE(_value) && !IS_FALSE(_value)) {                               \
+      ERROR(PLUGIN_NAME " plugin: Wrong value for boolean parameter %s: %s",   \
+            _str_param_name, _value);                                          \
+      return -1;                                                               \
+    }                                                                          \
+  } while (0)
+
 static int lv_config(const char *key, const char *value) {
-  if (virInitialize() != 0)
+  if (virInitialize() != 0) {
```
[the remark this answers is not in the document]
rjablonx: Hi,
thanks for review. Indeed, at first glance it seems to be unnecessary - I'll try to remove that and I'll do tests to check if collectd will be working fine without that.

src/virt.c
rjablonx: it was tempting to add logic here because all needed fields were already fetched in add_domain but it is no problem to get domain name separately and skip domain before 'add_domain' call. I'll refactor this

src/collectd.conf.pod
```diff
@@ -9252,13 +9252,51 @@ surrounded by I</.../> and collectd was compiled with support for regexps.
...
+B<Note:> B<BlockDevice> and B<InterfaceDevice> options are related to
+corresponding B<*Format> options. Specifically, B<BlockDevice> filtering depends
+on B<BlockDeviceFormat> setting - if user wants to filter block devices by
+'target' name then B<BlockDeviceFormat> option has to be set to 'target' and
+B<BlockDevice> option must be set to a valid block device target
+name("/:hdb/"). Mixing formats and filter values from different worlds (i.e.,
+using 'target' name as B<BlockDevice> value with B<BlockDeviceFormat> set to
+'source') may lead to unexpected results (all devices filtered out or all
+visible, depending on the value of B<IgnoreSelected> option).
+Simiralarly, option B<InterfaceDevice> is related to B<InterfaceFormat> setting
```
rubenk: Similarly

src/virt.c
```diff
@@ -1187,16 +1232,21 @@ static int lv_config(const char *key, const char *value) {
     return 0;
   }
   if (strcasecmp(key, "BlockDeviceFormatBasename") == 0) {
+    VALIDATE_BOOLEAN_VALUE(value, "BlockDeviceFormatBasename");
```
rubenk: Can't you use cf_util_get_boolean here?

rjablonx: Initially I've tried to use 'cf_util_get_boolean' but it demands oconfig_item_t * as a parameter and it is not available from this context

rubenk: You could register lv_config as a complex config function (which I'd argue it is by now)

rpv-tomsk: >You could register lv_config as a complex config function (which I'd argue it is by now)
I had the same thoughts.

rjablonx: Hi guys,
is it possible to integrate this PR as is and to send refactoring-patch for lv_config function withing next two weeks as separate PR(to change lv_config->complex config/remove probably not needed virInitialize()/refactor skipping domains logic) ?
I'm asking because before sending next patch we have to do full test cycle to be sure that we don't introduce any regressions(+some time for internal approval chain) and we are afraid that we might miss 5.9 release deadline with all bugfixes from current PR.

rpv-tomsk: >is it possible to integrate this PR as is and to send refactoring-patch for lv_config function withing next two weeks as separate PR

>>From my POV "maybe" and "in the future" are not allowed for such small changes.

I have no strong requirement to redo lv_config as a complex config function, 
but two weeks have passed and no changes published.

rjablonx: Hi @rpv-tomsk, code with all required changes is already prepared (refactoring to use complex_config + removed virInitialize + refactored skipping domains + typo fixes) and we are waiting for internal green light before we can publish changes.

rjablonx:
Hi guys,

I've just uploaded required changes(refactoring to use complex_config + removed virInitialize + refactored skipping domains + typo fixes)

src/virt.c
```diff
@@ -1166,254 +1152,331 @@ static int lv_init_ignorelists() {
...
+        ERROR(PLUGIN_NAME " plugin: unknown InterfaceFormat: %s", format);
+        sfree(format);
+        return -1;
+      }
+
+      sfree(format);
+      continue;
+    } else if (strcasecmp(c->key, "Instances") == 0) {
+      if (cf_util_get_int(c, &nr_instances) != 0) {
+        ERROR(PLUGIN_NAME " plugin: could not get 'Instances' parameter");
```
rpv-tomsk: Hi,

Additional error reporting is not needed after `cf_util_get_int()`.
It has such reporting inside.

src/virt.c
```diff
@@ -1166,254 +1152,331 @@ static int lv_init_ignorelists() {
...
+              nr_instances, NR_INSTANCES_MAX);
+        return -1;
+      }
+
+      DEBUG(PLUGIN_NAME " plugin: configured %i instances", nr_instances);
+      continue;
+    } else if (strcasecmp(c->key, "ExtraStats") == 0) {
+      char *ex_str = NULL;
+
+      if (cf_util_get_string(c, &ex_str) != 0 || ex_str == NULL) {
```
rpv-tomsk: Hi,

Additional error reporting is not needed after cf_util_get_string().
It has such reporting inside. Also, there is no need to duplicate `ex_str == NULL` check.

src/virt.c
```diff
@@ -1166,254 +1152,331 @@ static int lv_init_ignorelists() {
...
 
-  if (strcasecmp(key, "PersistentNotification") == 0) {
-    persistent_notification = IS_TRUE(value);
-    return 0;
-  }
+      /* ExtraStats parsed successfully*/
+      continue;
+    } else if (strcasecmp(c->key, "PersistentNotification") == 0) {
+      if (cf_util_get_boolean(c, &persistent_notification) != 0) {
+        ERROR(PLUGIN_NAME
```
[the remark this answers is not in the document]
rjablonx: Hi @rpv-tomsk,
Yes, I know that cf_util_get.. family of functions has internal reporting - I just wanted to make clearly visible to user that failed option is related to virt plugin, Not sure if unnecessary - prefer more info than less info in case of errors but can remove if required.

src/virt.c
rjablonx: Hi,
same as above.

src/virt.c
rjablonx: yes, seems that ex_str checking for null is redundant here

src/virt.c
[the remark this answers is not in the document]
anaudx: @rpv-tomsk May we use `DEBUG` message instead of `ERROR` message in these cases ?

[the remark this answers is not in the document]
rpv-tomsk: Why do you need this DEBUG? `cf_util_get_boolean` has ERROR reporting with plugin name included in log messages.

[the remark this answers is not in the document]
anaudx: We have no particular reason, this was just to output more contextual info to the user. But if you think this is redundant or too much information, we can remove them. Also, in order to be consistent, this means removing such ERROR messages in several other places throughout the plugin (that existed prior to this PR).

anaudx:
hi all,
I just uploaded requested changes (removing ERROR messages and unnecessary NULL pointer checks)

rpv-tomsk:
Hi, Team.

I will not merge none of PR untill it will be possible for me to merge my PR too.

Thanks for your attention.

mrunge:
Can we please move any

> Hi, Team.
> 
> I will not merge none of PR untill it will be possible for me to merge my PR too.
> 
> Thanks for your attention.

@rpv-tomsk 

Pavel, I don't know what makes you think your PRs are not greatly appreciated.

To me, it looks like there is a config issue here with the automation, so that code-owners approval does not lead to merging code.
For now, I'm happy to be the proxy here and to merge code, once it is approved by code owners.

If I happen to miss an open PR, please feel free to add me to the PR, and I'll take care of it.

Thank you all for your contributions here.

PR #3125, opened 2019-03-12
mysql: [impr] enable uptime metric

skob:
ChangeLog: mysql plugin: added uptime metric

### DIFF
--- a/src/mysql.c
+++ b/src/mysql.c
@@ -905,8 +905,8 @@ static int mysql_read(user_data_t *ud) {
 
     } else if (strncmp(key, "Slow_queries", strlen("Slow_queries")) == 0) {
       derive_submit("mysql_slow_queries", NULL, val, db);
-    } else if (strncmp(key, "Uptime", strlen("Uptime")) == 0) {
-      gauge_submit("mysql_uptime", NULL, val, db);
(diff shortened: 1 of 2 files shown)


src/mysql.c
```diff
@@ -905,6 +905,8 @@ static int mysql_read(user_data_t *ud) {
 
     } else if (strncmp(key, "Slow_queries", strlen("Slow_queries")) == 0) {
       derive_submit("mysql_slow_queries", NULL, val, db);
+    } else if (strncmp(key, "Uptime", strlen("Uptime")) == 0) {
```
rpv-tomsk: Hi,

This line is wrong. Mysql output has `Uptime` and `Uptime_since_flush_status` variables, and this check matches them both.

src/types.db
```diff
@@ -165,6 +165,7 @@ mysql_sort              value:DERIVE:0:U
 mysql_sort_merge_passes value:DERIVE:0:U
 mysql_sort_rows         value:DERIVE:0:U
 mysql_slow_queries      value:DERIVE:0:U
+mysql_uptime            value:GAUGE:0:4294967295
```
rpv-tomsk: Existing `uptime` type should be used.

rpv-tomsk:

PR #3131, opened 2019-03-28
Fix apache plugin BusyWorkers/IdleWorkers update

zuckschwerdt:
Apache http mod_status added a second set of `BusyWorkers`, `IdleWorkers` in https://github.com/apache/httpd/commit/6befc1893bea72320388b84fd879a4533e29b244

E.g.
```
...
DurationPerReq: 8.53412
BusyWorkers: 1
IdleWorkers: 49
Processes: 2
Stopping: 0
BusyWorkers: 1
IdleWorkers: 49
ConnsTotal: 2
...
```

This will result in an attempt to submit those twice in the same interval which errors with
- apache_connections.rrd: illegal attempt to update
- apache_idle_workers.rrd: illegal attempt to update

This patch adds a guard to only submit one value per update.

ChangeLog: apache plugin: BusyWorkers/IdleWorkers update has been fixed.

### DIFF
--- a/src/apache.c
+++ b/src/apache.c
@@ -509,6 +509,12 @@ static int apache_read_host(user_data_t *user_data) /* {{{ */
   char *ptr = st->apache_buffer;
   char *saveptr = NULL;
   char *line;
+  /* Apache http mod_status added a second set of BusyWorkers, IdleWorkers in
+   * https://github.com/apache/httpd/commit/6befc18
+   * For Apache 2.4.35 and up we need to ensure only one key is used.
+   * S.a. https://bz.apache.org/bugzilla/show_bug.cgi?id=63300
+   */
+  int apache_connections_submitted = 0, apache_idle_workers_submitted = 0;
   while ((line = strtok_r(ptr, "\n\r", &saveptr)) != NULL) {
     ptr = NULL;
     char *fields[4];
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
>added a second set of BusyWorkers, IdleWorkers

I think you have to report this as a bug in 'Apache HTTPD' project.
If these 'sets' has same value, then only one set has to exists.
If these sets has different values, then we need clear way to distingush them, and explanation of their values.

zuckschwerdt:
The apache httpd commit linked shows only very few changed lines. I don't see how that was a mistake. Also it's dated from "7 Aug 2018" -- somebody would have noticed?

Anyway, at least make a prominent note that apache 2.4.38 and up will not work with this plugin. The amount of failure messages is a hint but that might be to late.

No strong feelings if you close this PR though ;)

zuckschwerdt:
I'll raise an issue with httpd for this. The would still be needed for apache 2.4.35 to at least 2.4.40

rpv-tomsk:
Please update this PR with your results.

rpv-tomsk:
Fast solution like proposed by you is not right way to solve this issue. 
That is bad in the long run. 
Reasoning:

>If these 'sets' has same value, then only one set has to exists.
>If these sets has different values, then we need clear way to distingush them, and explanation of their values.

We need clear explanation of fields in status report, to make an informed decision how to handle them.
Right now I see no such good reasoning to merge your solution.

>Anyway, at least make a prominent note that apache 2.4.38 and up will not work with this plugin. The amount of failure messages is a hint but that might be to late.

That change was included in httpd-2.4.35.

zuckschwerdt:
Yes, sorry, 2.4.35, edited above for clarity. No problem, close if you feel this is bad practice.

For reference issue at Apache Bugzilla: https://bz.apache.org/bugzilla/show_bug.cgi?id=63300

I still feel this should be addressed somehow in collectd, otherwise this plugin will not work with at least the last 5 versions, since apache httpd 2.4.35.

I'll put this fix downstream with some distros I use in the meantime.

mrunge:
I have an output on a Fedora machine

```
ServerVersion: Apache/2.4.38 (Fedora)
ServerMPM: event
Server Built: Jan 23 2019 12:39:31
CurrentTime: Tuesday, 02-Apr-2019 14:09:32 CEST
RestartTime: Tuesday, 02-Apr-2019 14:07:17 CEST
ParentServerConfigGeneration: 1
ParentServerMPMGeneration: 0
ServerUptimeSeconds: 135
ServerUptime: 2 minutes 15 seconds
Load1: 0.77
Load5: 0.51
Load15: 0.34
Total Accesses: 4
Total kBytes: 11
Total Duration: 1
CPUUser: .03
CPUSystem: .05
CPUChildrenUser: 0
CPUChildrenSystem: 0
CPULoad: .0592593
Uptime: 135
ReqPerSec: .0296296
BytesPerSec: 83.437
BytesPerReq: 2816
DurationPerReq: .25
BusyWorkers: 1
IdleWorkers: 74
Processes: 3
Stopping: 0
BusyWorkers: 1
IdleWorkers: 74
ConnsTotal: 0
ConnsAsyncWriting: 0
ConnsAsyncKeepAlive: 0
ConnsAsyncClosing: 0
```
With that, httpd in 2.4.38 is not affected by the described issue

zuckschwerdt:
Thanks for verifying. Your output has this issue:
```
...
DurationPerReq: .25
BusyWorkers: 1  <--
IdleWorkers: 74  <--
Processes: 3
Stopping: 0
BusyWorkers: 1  <--
IdleWorkers: 74  <--
ConnsTotal: 0
...
```

mrunge:
Dang, I missed this, you are right.

zuckschwerdt:
Well, I built an intrusive debugging setup to catch the source of that error, then one day blinked and couldn't believe how I'd missed that all the time ;)

dago:
@zuckschwerdt Can you please add an elaborate comment on the issue near the workaround in the sourcecode explaining the rationale for it and the affected Apache versions.

zuckschwerdt:
@dago comment added, does it explain sufficiently?

PR #3135, opened 2019-04-04
SNMP: apply options 'Scale' and 'Shift' to all values, not just to ga…

alex2grad:
…uge-values

The options 'Scale' and 'Shift' are applied only to gauge-values now.
This patch removes this restriction to apply these options to all types
of values.
This patch also applies these options for SNMP values of ANS.1 type
"OCTET STRING".

### DIFF
--- a/src/collectd-snmp.pod
+++ b/src/collectd-snmp.pod
@@ -245,22 +245,18 @@ traffic.
 
 =item B<Scale> I<Value>
 
-The gauge-values returned by the SNMP-agent are multiplied by I<Value>.  This
+The values returned by the SNMP-agent are multiplied by I<Value>.  This
 is useful when values are transferred as a fixed point real number. For example,
 thermometers may transfer B<243> but actually mean B<24.3>, so you can specify
 a scale value of B<0.1> to correct this. The default value is, of course,
 B<1.0>.
 
-This value is not applied to counter-values.
-
 =item B<Shift> I<Value>
 
-I<Value> is added to gauge-values returned by the SNMP-agent after they have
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
Hi,

I think `Shift` can't be applied to DS_TYPE_DERIVE inside of plugin as constant value will not change derivative value. And DS_TYPE_ABSOLUTE and DS_TYPE_COUNTER are treated as a bit deprecated, AFAIK.

So, the question is: what is your use-case?

alex2grad:
> So, the question is: what is your use-case?
Getting CyberPower PDU stat, one of the OID 
is DERIVE type which returns apparent power measured in 0.1 kilowatt-hours.
So I need 'Scale 0.1' to correctly get this data.

BTW, looking at snmp_agent.c there is function snmp_agent_get_value_from_ds_type
which use scale and shift on all types.
===
static int snmp_agent_get_value_from_ds_type(const value_t *val, int type,
                                             double scale, double shift,
                                             long *value) {
  switch (type) {
  case DS_TYPE_COUNTER:
    *value = (long)((val->counter * scale) + shift);
    break;
  case DS_TYPE_ABSOLUTE:
    *value = (long)((val->absolute * scale) + shift);
    break;
  case DS_TYPE_DERIVE:
    *value = (long)((val->derive * scale) + shift);
    break;
  case DS_TYPE_GAUGE:
    *value = (long)((val->gauge * scale) + shift);
    break;
  case TYPE_STRING:
    break;
  default:
    ERROR(PLUGIN_NAME ": Unknown data source type: %i", type);
    return -EINVAL;
  }
 
  return 0;
}
===

Issue #3134, opened 2019-04-03
df is wrong when using systemd


btoneill:
*   Version of collectd: 5.8
*   Operating system / distribution: CentOS 7
*   Kernel version (if applicable):

## Expected behavior
df module should include mounted filesystems

## Actual behavior

df module includes all mounted filesystems as well as also including /etc and /usr as mounted filesystems, even if they aren't.

## Steps to reproduce

*   Run collectd with df module
*    See output

btoneill:
The cause of this is known, but, not sure what the proper fix is.

In CentOS 7, /etc/mtab is a symlink to /proc/self/mounts.
The systemd service file for collectd has "ProtectSystem=full"
ProtectSystem=full causes /etc and /usr to be mounted ro for the process and shows up as a ro mount in /proc/${PID}/mounts for the collectd pid
collectd parses /etc/mtab, which then gives us the systemd protected filesytem mounts.

Not sure what the best option to fix this is.

Maybe have it look for /proc/1/mounts instead of /etc/mtab? That is pretty ugly tho...
Maybe an option to use a different file for COLLECTD_MNTTAB as an option to the df module?

btoneill:
Or, the other option is remove ProtectSystem=full, not sure if that's a better solution or note. It gets around the problem, but, it does get rid of the added little bit of security that systemd gives you.

btoneill:
Looks like not all that much actually uses this value on CentOS 7 at least.

$ grep -R ProtectSystem .
./multi-user.target.wants/NetworkManager.service:ProtectSystem=true
./multi-user.target.wants/collectd.service:ProtectSystem=full
./dbus-org.bluez.service:ProtectSystem=full
./bluetooth.target.wants/bluetooth.service:ProtectSystem=full
./dbus-org.freedesktop.NetworkManager.service:ProtectSystem=true

rubenk:
> df module includes all mounted filesystems as well as also including /etc and /usr as mounted filesystems, even if they aren't.

Are you sure they aren't? I think systemd mounts them in a private mount namespace. I think collectd does what it's supposed to do here. You can always exclude /etc/ and /usr.

btoneill:
I don't think we want it to add in /etc and /usr. If I change ProtectSystem=false in the collectd.service file it no longer mounts /etc and /usr and collectd no longer adds them. 

[...]

PR #3136, opened 2019-04-05
Remove ProtectSystem from service file

mrunge:
This prevents showing eg. /etc if that's a separate file system.

Resolves: https://github.com/collectd/collectd/issues/3134


rubenk:
Thanks Matthias.

PR #3137, opened 2019-04-11
Add support for RS485 to modbus plugin

dangowrt:
ChangeLog: Allow setting up RS485 mode for Modbus-RTU

### DIFF
--- a/src/modbus.c
+++ b/src/modbus.c
@@ -97,6 +97,12 @@ enum mb_conntype_e /* {{{ */
   MBCONN_RTU }; /* }}} */
 typedef enum mb_conntype_e mb_conntype_t;
 
+enum mb_uarttype_e /* {{{ */
+{ UARTTYPE_RS232,
+  UARTTYPE_RS422,
+  UARTTYPE_RS485 }; /* }}} */
+typedef enum mb_uarttype_e mb_uarttype_t;
+
 struct mb_data_s;
 typedef struct mb_data_s mb_data_t;
 struct mb_data_s /* {{{ */
@@ -126,8 +132,9 @@ struct mb_host_s /* {{{ */
   char host[DATA_MAX_NAME_LEN];
   char node[NI_MAXHOST]; /* TCP hostname or RTU serial device */
   /* char service[NI_MAXSERV]; */
-  int port;     /* for Modbus/TCP */
-  int baudrate; /* for Modbus/RTU */
+  int port;               /* for Modbus/TCP */
+  int baudrate;           /* for Modbus/RTU */
+  mb_uarttype_t uarttype; /* UART type for Modbus/RTU */
   mb_conntype_t conntype;
 
   mb_slave_t *slaves;
@@ -386,6 +393,22 @@ static int mb_init_connection(mb_host_t *host) /* {{{ */
     host->connection = NULL;
     return status;
   }
+
+#if defined(linux) && LIBMODBUS_VERSION_CHECK(2, 9, 4)
+  switch (host->uarttype) {
+  case UARTTYPE_RS485:
+    if (modbus_rtu_set_serial_mode(host->connection, MODBUS_RTU_RS485))
+      DEBUG("Modbus plugin: Setting RS485 mode failed.");
+    break;
+  case UARTTYPE_RS422:
+    /* libmodbus doesn't say anything about full-duplex symmetric RS422 UART */
+    break;
+  case UARTTYPE_RS232:
+    break;
+  default:
+    DEBUG("Modbus plugin: Invalid UART type!.");
+  }
+#endif /* defined(linux) && LIBMODBUS_VERSION_CHECK(2, 9, 4) */
 
   return 0;
 } /* }}} int mb_init_connection */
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
Hi,

Please note CI messages and I will do review of your work.

Thanks for your contribution!

dangowrt:
@rpv-tomsk `contrib/format.sh src/modbus.c` only makes changes outside of my contribution, so I guess this should be an additional commit fixing existing style problems.
```
diff --git a/src/modbus.c b/src/modbus.c
index 59ce27eb..91b1c833 100644
--- a/src/modbus.c
+++ b/src/modbus.c
@@ -1066,8 +1066,7 @@ static int mb_config_add_host(oconfig_item_t *ci) /* {{{ */
                                  /* callback = */ mb_read,
                                  /* interval = */ interval,
                                  &(user_data_t){
-                                     .data = host,
-                                     .free_func = host_free,
+                                     .data = host, .free_func = host_free,
                                  });
   } else {
     host_free(host);
```

dangowrt:
@rpv-tomsk CI tests now pass
All checks have passed
10 successful checks

dangowrt:
@rpv-tomsk CI tests now pass
All checks have passed
10 successful checks

src/modbus.c
```diff
@@ -983,11 +1003,26 @@ static int mb_config_add_host(oconfig_item_t *ci) /* {{{ */
...
       status = cf_util_get_string_buffer(child, host->node, sizeof(host->node));
-      if (status == 0)
+      if (status == 0) {
         host->conntype = MBCONN_RTU;
+        host->uarttype = UARTTYPE_RS232;
+      }
     } else if (strcasecmp("Baudrate", child->key) == 0)
       status = cf_util_get_int(child, &host->baudrate);
-    else if (strcasecmp("Interval", child->key) == 0)
+    else if (strcasecmp("UARTType", child->key) == 0) {
```
rpv-tomsk: This configuration directive should be processed with respect to `LEGACY_LIBMODBUS` define.

```
#if !LEGACY_LIBMODBUS && LIBMODBUS_VERSION_CHECK(2,9,4)
      [... your code ...]
#else
      ERROR("Modbus plugin: Option `UARTType' not supported. Please "
            "upgrade libmodbus to at least 2.9.4");
      return -1;
#endif
```

This also should check `#if defined(linux)` too.

src/modbus.c
```diff
@@ -983,11 +1003,26 @@ static int mb_config_add_host(oconfig_item_t *ci) /* {{{ */
...
+      if (status != 0)
+        break;
+      if (strncmp(buffer, "RS485", 6) == 0)
+        host->uarttype = UARTTYPE_RS485;
+      else if (strncmp(buffer, "RS422", 6) == 0)
+        host->uarttype = UARTTYPE_RS422;
+      else if (strncmp(buffer, "RS232", 6) == 0)
+        host->uarttype = UARTTYPE_RS232;
+      else
+        break;
```
rpv-tomsk: Please add complain if option has wrong value.

src/modbus.c
```diff
@@ -369,6 +376,19 @@ static int mb_init_connection(mb_host_t *host) /* {{{ */
       ERROR("Modbus plugin: Creating new Modbus/RTU object failed.");
       return -1;
     }
+    switch (host->uarttype) {
+    case UARTTYPE_RS485:
+      if (modbus_rtu_set_serial_mode(host->connection, MODBUS_RTU_RS485))
+        DEBUG("Modbus plugin: Setting RS485 mode failed.");
+    /* fall-through */
```
rpv-tomsk: Use of `/* fall-through */` looks strange here.

rpv-tomsk: The `modbus_rtu_set_serial_mode` appeared only in  libmodbus-2.9.4 and available only on Linux platform.
You need to add needed conditionals to handle this correctly.

dangowrt: I changed them to 'break' instead, it doesn't really matter at all in that place obviously.

dangowrt:
@rpv-tomsk Thanks for the quick review. I adressed all your comments and run-time-tested the result.

src/modbus.c
```diff
@@ -387,6 +394,22 @@ static int mb_init_connection(mb_host_t *host) /* {{{ */
     return status;
   }
 
+#if defined(linux) && LIBMODBUS_VERSION_CHECK(2, 9, 4)
```
rpv-tomsk: Can you please update this line to match l1016:
`#if defined(linux) && !LEGACY_LIBMODBUS && LIBMODBUS_VERSION_CHECK(2, 9, 4)`

I think that will be more correct.

dangowrt: This is inside a whole section of  `#ifdef !LEGACY_LIBMODBUS` because `mb_init_connection()` is declared once for `LEGACY_LIBMODBUS` and once for `!LEGACY_LIBMODBUS`. I added my code only to the function declared in the `!LEGACY_LIBMODBUS` case.

dangowrt: Do you still like me do add this (redundant) `!LEGACY_LIBMODBUS`?

rpv-tomsk: No, thanks, I simply missed that upper 'if'.

rpv-tomsk:
Thanks for your work on Collectd !

PR #3138, opened 2019-04-11
exec: fix a race condition when setting environment variables

### LINKED ISSUE #3041: Bug in Exec: missing environment variables in many of children processes
*   Operating system / distribution: built from source
*   Kernel version (if applicable): not applicable

## Expected behavior

All children processes should have environment variables COLLECTD_INTERVAL and COLLECTD_HOSTNAME accessible

## Actual behavior

Some of children processes are missing COLLECTD_INTERVAL and COLLECTD_HOSTNAME. Which processes are missing variables is always different.

## Steps to reproduce

Configure a number of external programs running with Exec plugin (let it be just a simple shell script echo-ing environment variables to a file) and start collectd.

## Notes
If my reading of the code is correct there is a possibility of race in [fork_child](https://github.com/collectd/collectd/blob/collectd-5.8.1/src/exec.c#L441) when setting ([1](https://github.com/collectd/collectd/blob/collectd-5.8.1/src/exec.c#L497)) and unsetting ([2](https://github.com/collectd/collectd/blob/collectd-5.8.1/src/exec.c#L541), [3](https://github.com/collectd/collectd/blob/collectd-5.8.1/src/exec.c#L565)) environment variables which is likely to be the cause.

mkauf:
Pass the environment as a parameter to the child process
and don't modify the environment of the parent process.

Follow-up to 0684aea7

Closes #3041

Changelog: exec: fix a race condition when setting environment variables

### DIFF
--- a/src/exec.c
+++ b/src/exec.c
@@ -26,6 +26,9 @@
 #define _DEFAULT_SOURCE
 #define _BSD_SOURCE /* For setgroups */
 
+/* _GNU_SOURCE is needed in Linux to use execvpe */
+#define _GNU_SOURCE
+
 #include "collectd.h"
 
 #include "plugin.h"
@@ -42,6 +45,8 @@
 #ifdef HAVE_SYS_CAPABILITY_H
 #include <sys/capability.h>
 #endif
+
+extern char **environ;
 
 #define PL_NORMAL 0x01
 #define PL_NOTIF_ACTION 0x02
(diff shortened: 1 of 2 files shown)


src/exec.c
```diff
@@ -486,17 +456,42 @@ static int fork_child(program_list_t *pl, int *fd_in, int *fd_out,
...
+    size_t envp_idx;
+    for (envp_idx = 0; environ[envp_idx] != NULL && envp_idx < env_size;
+         ++envp_idx) {
+      envp[envp_idx] = environ[envp_idx];
+    }
+
+    /* Add the collectd environment variables */
+    envp[envp_idx++] = interval_buf;
+    envp[envp_idx++] = hostname_buf;
+    envp[envp_idx++] = NULL;
```
kwiatrox: Minor issue: envp_idx is not used any more after that line, so increment is not necessary here.

mkauf: A bot already merged this PR, I don't know how to proceed.

mrunge: Submit a new PR then?

PR #3139, opened 2019-04-12
 lua: fix lua registration callback for more then one lua script

feckert:
This is already applied to NetBSD:
 http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/sysutils/collectd/patches/patch-src_lua.c?annotate=1.1

Closes #2379

Changelog: lua: fix lua registration callback for more then one lua script


feckert:
@octo ping

PR #3151, opened 2019-05-13
Libvirt plugin: split memory report into dedicated types

rpv-tomsk:
Changelog: libvirt plugin: split memory report into dedicated types

As documented (https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainMemoryStats),
statistics elements represents following:

1)
```
VIR_DOMAIN_MEMORY_STAT_SWAP_IN: The total amount of data read from swap space (in kb). 
VIR_DOMAIN_MEMORY_STAT_SWAP_OUT: The total amount of memory written out to swap space (in kb). 
```

These elements are about swap usage, not used amount, and should not be submitted as GAUGE type (which `memory` type is).

So, move these elements values to `swap_io` type which is DERIVE.

2) 
```
VIR_DOMAIN_MEMORY_STAT_MAJOR_FAULT: The number of page faults that required disk IO to service. 
VIR_DOMAIN_MEMORY_STAT_MINOR_FAULT: The number of page faults serviced without disk IO.
```

These elements are `number of page faults`, not bytes. They also should be submitted as DERIVE type `ps_pagefaults`.

3)
``` 
VIR_DOMAIN_MEMORY_STAT_LAST_UPDATE: Timestamp of the last statistic
```

That is timestamp and I see no reasons to report it as `memory` type.
Maybe it should be reported as some type, or used somehow else,  but in this PR its data simply rejected.

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -926,6 +926,10 @@ static void memory_stats_submit(gauge_t value, virDomainPtr dom,
     return;
   }
 
+  /* Skip 'last_update' reporting as that is not memory but timestamp */
+  if (tag_index == 9)
+    return;
+
   submit(dom, "memory", tags[tag_index], &(value_t){.gauge = value}, 1);
 }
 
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
Many Thanks!

rubenk:
You too!

Issue #3145, opened 2019-05-04
dns plugin: Rewriting numeric qtypes to strings depends on symbols that have been removed from newer glibc versions  


usev6:
*   Version of collectd: 5.8 (and older)
*   Operating system / distribution: Linux / CentOS 7 (probably other OS/distros that use glibc, too)

## Expected behavior
Metric names for qtypes should have the name of the resource record type -- e.g. ```dns_qtype.A```

## Actual behavior
Metric names for qtypes contain the numeric id of the resource record type with a ```#``` prepended -- e.g. ```dns_qtype.#1```

## Steps to reproduce
I've encountered this behaviour (numeric ids are no longer rewritten to the names of the resource record types) after an upgrade from collectd-5.8.0-3.el7 to collectd-5.8.0-4.el7 on CentOS 7. I've opened a bug report in the Fedora EPEL about the problem with the new package: https://bugzilla.redhat.com/show_bug.cgi?id=1706350

I think the underlying reason is in the code base of collectd, namely that the code in ```src/utils/dns/dns.c``` looks for the symbols ```__NAMESER``` and ```__BIND``` and falls back to the numeric type ids if those are not defined (this happens in qtype_str, cmp. https://github.com/collectd/collectd/blob/55cf383915/src/utils/dns/dns.c#L667). Since newer versions of glibc have removed those symbols (cmp. https://sourceware.org/git/?p=glibc.git;a=commit;h=006768c72a), the checks should be adjusted. I don't know how to do that in a clean way. Maybe just looking for ```HAVE_ARPA_INET_H``` would be good enough for the basic resource record types?

Please take my analysis with some salt -- I might be wrong about the cause of the problem.

usev6:
I've continued to wonder how a clean fix would look like. I don't really understand why the code in ```qtype_str``` uses the enum values from ```arpa/nameser.h``` like it does. Maybe you could help me to understand that?

Even with the initial commit of the (at that time new) dns plugin, ```qtype_str``` used symbols from ```arpa/nameser.h``` (or ```arpa/nameser_compat.h```) for the different cases of the switch statement: https://github.com/collectd/collectd/commit/1e7b5f4e24#diff-2706c7320cd0104beeb5a589d34db8b0R800

The next function in that file, ```opcode_str``` just used integers instead: https://github.com/collectd/collectd/commit/1e7b5f4e24#diff-2706c7320cd0104beeb5a589d34db8b0R854

[...]

PR #3156, opened 2019-05-15
dns plugin: Do not use headers from glibc

rpv-tomsk:
glibc has removed __NAMESER and __BIND symbols from their codebase.

There is no benefits of use defines from glibc, so replaced them by numbers.

Closes: #3145

Changelog: dns plugin: Do not use headers from glibc

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -635,16 +635,6 @@ fi
 
 # }}}
 
-# For the dns plugin
-AC_CHECK_HEADERS([arpa/nameser.h])
-AC_CHECK_HEADERS([arpa/nameser_compat.h], [], [],
-  [[
-    #if HAVE_ARPA_NAMESER_H
-    # include <arpa/nameser.h>
-    #endif
-  ]]
-)
-
 AC_CHECK_HEADERS([net/if_arp.h], [], [],
   [[
     #if HAVE_SYS_SOCKET_H
--- a/src/utils/dns/dns.c
+++ b/src/utils/dns/dns.c
@@ -79,12 +79,6 @@
 #if HAVE_ARPA_INET_H
 #include <arpa/inet.h>
 #endif
-#if HAVE_ARPA_NAMESER_H
-#include <arpa/nameser.h>
-#endif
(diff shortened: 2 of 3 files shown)


rpv-tomsk:
Sorry, Team,

I'm not able to do anything with repeated failings on Travis CI and will not do nothing to push this PR into upstream.

Let there be another dead PR?

usev6:
If I read the output of the failing travis job correctly, there was something with the system clock. The before_script ran for -358.01 seconds (line 501):

> $ autoreconf -vif         before_script  -358.01s

It looks like the system clock was corrected (backwards) while autoreconf was running. Probably the ```AM_SANITY_CHECK``` macro discovered this problem later, when it compared timestamps of two files.

If this specific problem has happened before on mac os x, it might make sense to try to execute a ```sudo sntp -sS pool.ntp.org``` (and wait a bit?) before starting autoreconf.

Whatever, I don't think that this failure was related to the changes in the pull request -- so to me that doesn't sound like a good reason to let this PR become dead.

rpv-tomsk:
>If I read the output of the failing travis job correctly...
>It looks like the system clock was corrected...
>that doesn't sound like a good reason to let this PR become dead.

That is not the first this problem occurence, but nobody takes care.
As for now, I see no any reason to do 'force push' to restart failed CI check: unfortunately, after #2958 I have no permission to merge my own pull requests. As there is no interest/actions from other Team members and I will do nothing too, sorry.

usev6:
Hi Pavel,

I've read through #2958 and I understand your above comments better now. Regardless of the discussion about rights to merge pull requests (I'm not in a position to comment on this), I'd like to thank you for looking at my issue and opening this PR. I'm quite sure a lot of people highly appreciate your work on collectd.

Best regards
Christian

rpv-tomsk:
As you all can see, nobody cares.

I think this is good example of #2958 problem.

I'm already thinking about to press "Leave Team" button, seriously.
That gives me no benefits, only problems.
But, I think, the team doesn't care about that too.

PR #3159, opened 2019-05-19
check_uptime: New plugin, based on new cache_event callback.

rpv-tomsk:
Hi,

This is implementation of `check_uptime` plugin, based on new `cache_event` callback.
Plugin checks for cache events for `uptime` metric, and sends notification when metric state changed.

When new `uptime` metric appears in cache, OK notification sent:
```
Host: hs01
Plugin: mysql
Plugin instance: local
Type: uptime

Message: Service is running. Uptime now: 4 sec.
```

When new value for `uptime` metric is less than previous value, WARNING notification sent:

```
Host: hs01
Plugin: mysql
Plugin instance: local
Type: uptime

Message: Service just restarted. Uptime was: 15 min 9 sec. Uptime now: 2 sec.
```

When no new updates for metric, cache entry expires. This results in FAILURE notification:

```
Host: hs01
Plugin: mysql
Plugin instance: local
Type: uptime

Message: Service is down. Uptime was: 3 min 32 sec.
```

For `uptime` plugin there is small hardcode, so notification message is about 'Host', not 'Service':
```
2019-05-19 21:43:24 - OKAY@hs01

Host: hs01
Plugin: uptime
Type: uptime

Message: Host is running. Uptime now: 35 day(s) 6 hour(s) 25 min 21 sec.
```

New `cache_event` callback implemented for efficiency.
There is no need to send all metrics to plugin via `write` callback, so plugin does not need to check all that data stream. When new metric appears in cache, `cache_event` callback called once. If plugin confirms its interest in that metric, then it will receive further UPDATE / EXIPRED events.

To avoid race conditions, `cache_event` callbacks are registered in array and mask was added into cache element structure to signal if callback should be called.  This allows to avoid malloc/memcpy operations in process of callbacks calling at the cost of limit: only 32 callbacks can be registered (can be easily expanded to 64).

Based on this new callback, `threshold` plugin can be reworked in efficient way too.

Will be glad to get all your feedbacks.

Changelog: check_uptime: New plugin, based on new cache_event callback.


rpv-tomsk:
cc @faxm0dem

rpv-tomsk:
Thanks for review.

PR #1215, opened 2015-08-10
Modify network.c to generate rates.

ThomasADavis:
Update to #949 to current git master.

Fixes several of the items listed, including several cut&paste errors.

PR #3162, opened 2019-05-21
write_influxdb_udp: new plugin to send values to InfluxDB using line protocol via udp.

carlospeon:
Hello:

This is a try to improve collectd and influxdb integration, allowing you to write to influxdb using tags and sending rates for derive datasources. It uses udp influxdb line protocol [1]. Its based on network plugin and #1215.

It's possible to send rates to influxdb using others plugins, but usually not using influxdb tags adequately (#1107)

Regards,
Carlos.

[1] https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_tutorial/

ChangeLog: write_influxdb_udp: new plugin to send values to InfluxDB using line protocol via udp.


src/write_influxdb_udp.c
```diff
@@ -0,0 +1,729 @@
...
+  struct sockaddr_storage *bind_addr;
+};
+
+struct sockent_server {
+  int *fd;
+  size_t fd_num;
+};
+
+typedef struct sockent {
+#define SOCKENT_TYPE_CLIENT 1
```
rpv-tomsk: As I see, you use only CLIENT.

Please remove all dead/unused code, related to SERVER.

carlospeon: Thank you for review.

I think it's done.

CI build failed with:
`checking whether build environment is sane... configure: error: newly created file is older than distributed files!
Check your system clock
The command "./configure" exited with 1.`

rpv-tomsk: >CI build failed with:

That is known issue (for me), unfortunately nobody cares.
If you want, you can do `git commit --amend` + `git push -f ... `.

carlospeon: Thank you for the CI workaround :)

src/write_influxdb_udp.c
```diff
@@ -0,0 +1,701 @@
...
+    int optname;
+
+    if (IN_MULTICAST(ntohl(addr->sin_addr.s_addr)))
+      optname = IP_MULTICAST_TTL;
+    else
+      optname = IP_TTL;
+
+    if (setsockopt(se->client.fd, IPPROTO_IP, optname, &wifxudp_config_ttl,
+                   sizeof(wifxudp_config_ttl)) != 0) {
+      ERROR("network plugin: setsockopt (ipv4-ttl): %s", STRERRNO);
```
rpv-tomsk: Please review error messages.

src/write_influxdb_udp.c
```diff
@@ -0,0 +1,701 @@
...
+      return -1;
+  }
+
+  return 0;
+} /* int bind_socket_to_addr */
+
+static sockent_t *sockent_create() {
+  sockent_t *se;
+
+  se = calloc(1, sizeof(*se));
```
rpv-tomsk: New codestyle expects variables declared on first use, if possible.

src/write_influxdb_udp.c
```diff
@@ -0,0 +1,701 @@
...
+static bool wifxudp_config_store_rates;
+
+static sockent_t *sending_socket;
+
+/* Buffer in which to-be-sent network packets are constructed. */
+static char *send_buffer;
+static char *send_buffer_ptr;
+static int send_buffer_fill;
+static cdtime_t send_buffer_last_update;
+static value_list_t send_buffer_vl = VALUE_LIST_INIT;
```
rpv-tomsk: Unused variable.

src/write_influxdb_udp.c
```diff
@@ -0,0 +1,701 @@
...
+  if (status == 0) {
+    /* no real values to send (nan) */
+    free(values);
+    return 0;
+  }
+  if (status < 0) {
+    ERROR("write_influxdb_udp plugin: write_influxdb_point failed.");
+    free(values);
+    return -1;
+  } else if (status > wifxudp_config_packet_size) {
```
rpv-tomsk: I think this condition is not possible to occur.
`write_influxdb_point()` returns `-1` if it has no space in `values + values_len` defined buffer.
And `values_len <= wifxudp_config_packet_size` always.

src/write_influxdb_udp.c
```diff
@@ -0,0 +1,701 @@
...
+
+  if (wifxudp_config_packet_size - send_buffer_fill < status)
+    flush_buffer();
+
+  memcpy(send_buffer_ptr, values, status);
+  send_buffer_fill += status;
+  send_buffer_ptr += status;
+  send_buffer_last_update = cdtime();
+
+  if (wifxudp_config_packet_size - send_buffer_fill < 120)
```
rpv-tomsk: What is magic of this `120` value?

src/write_influxdb_udp.c
```diff
@@ -0,0 +1,701 @@
...
+    flush_buffer();
+
+  pthread_mutex_unlock(&send_buffer_lock);
+
+  free(values);
+
+  return (status < 0) ? -1 : 0;
```
rpv-tomsk: This condition never occurs too.

src/write_influxdb_udp.c
```diff
@@ -0,0 +1,701 @@
...
+      ERROR("write_influxdb_udp plugin: "
+            "sendto failed: %s. Closing sending socket.",
+            STRERRNO);
+      sockent_client_disconnect(sending_socket);
+      return;
+    }
+
+    break;
+  } /* while (42) */
+} /* void network_send_buffer_plain */
```
rpv-tomsk: Please, do careful code review yourself.

src/write_influxdb_udp.c
```diff
@@ -0,0 +1,701 @@
...
+
+    if (ds->ds[i].type == DS_TYPE_GAUGE) {
+      if (isnan(vl->values[i].gauge))
+        continue;
+      if (have_values)
+        BUFFER_ADD(",");
+      BUFFER_ADD("%s=%lf", ds->ds[i].name, vl->values[i].gauge);
+      have_values = true;
+    } else if (wifxudp_config_store_rates) {
+      if (rates == NULL)
```
rpv-tomsk: I slightly misunderstand this piece of code. What is the goal?
The `csv` plugin (where all it started) documentation says: `If set to B<true>, convert 'counter' values to rates. If set to B<false> (thedefault) counter values are stored as is`. But `csv` and  all its code copy-pasts  touched all types except GAUGE, not only COUNTER. There is mismatch between implementation and documentation which is misleading.

src/write_influxdb_udp.c
carlospeon: Sure. Thank you.

src/write_influxdb_udp.c
carlospeon: Thank you. It should be done.

src/write_influxdb_udp.c
carlospeon: Removed. Thank you.

src/write_influxdb_udp.c
carlospeon: Absolutely right.

src/write_influxdb_udp.c
carlospeon: It's the measured average point size in my testing. As plugin, type, instance, and timestamps have the usual values, the differences with others setups are mostly hostnames, so it should not be very different.

Due the nature of this protocol, it probably should be used throught the loopback interface and with a bigger than usual packet size, so it seemed a good approach to me, but it's OK to remove also.

src/write_influxdb_udp.c
carlospeon: Yes, done.

src/write_influxdb_udp.c
carlospeon: Sorry :$

src/write_influxdb_udp.c
rpv-tomsk: Cite from https://mailman.verplant.org/pipermail/collectd/2014-April/006164.html

```
One attention point is that,
based on CSV plugin, I followed a rule of only calculating rates for
values which are not of GAUGE type. I don't fully understand the
reason for that, but once the data is the same for any output plugin,
I assumed it was right for Network too.
```

1) Can you work on documentation update?

https://mailman.verplant.org/pipermail/collectd/2014-April/006164.html

```
One attention point is that,
based on CSV plugin, I followed a rule of only calculating rates for
values which are not of GAUGE type. I don't fully understand the
reason for that, but once the data is the same for any output plugin,
I assumed it was right for Network too.
```

2) 

Do you have much metrics in your system? 

I have proposals to change `StoreRate` implementations, based on #3159 introduced callback, or to add new `write_rate` callback - that all is to avoid `uc_get_rate` call (which does search for value in cache while we can get value w/o such search).

That would be good to make CPU usage compare, if you have much metrics :-)
That will argue changes.

carlospeon: Updated documentation to include derive and absolute. Thank you.

src/write_influxdb_udp.c
rpv-tomsk: I think we need to add some comment about choosed value and goals.

rpv-tomsk: And another question: Why you don't use principle from `network` plugin, it write values to output buffer directly, w/o intermediate buffer + memcpy(), and does retry if data does not fit to buffer.
That avoids `memcpy()` for each written value, in cost of retry.

src/write_influxdb_udp.c
rpv-tomsk: >If set to B<true>, convert absolute, counter and derive values to rates.

heh, that is not so such simple change. ))
That is not true - derive is not converted to rate again. Derive is already rate.

rpv-tomsk: Please review this my opinion:

```When StoreRate is false, then raw metric values sent via wire.
For DERIVE types, what means raw value is sent, not its change, like for GAUGE, COUNTER and ABSOLUTE.

If StoreRate is true, then change of metric value is sent for DERIVE, COUNTER and ABSOLUTE types - i.e. rate.
GAUGE type always sent as-is.```

You should fully understand how things works (look into `src/daemon/utils_cache.c uc_update()` code where `values_gauge` calculated and do review. )

File: src/write_influxdb_udp.c (hunk 6 above)
Comment (carlospeon) [reply src/write_influxdb_udp.c]: I though that in network plugin the size of the value part was predictable, but in the end you are right, I think it can be done. Please let me some time to rewrite and test it.

File: src/write_influxdb_udp.c (hunk 9 above)
Comment (carlospeon) [reply src/write_influxdb_udp.c]: > That is not true - derive is not converted to rate again. Derive is already rate.
> (my this comment is not true too ... ))) )

I've looked at uc_update at utils_cache.c but I'm probably missing something.

`case DS_TYPE_DERIVE: {
      derive_t diff = vl->values[i].derive - ce->values_raw[i].derive;

      ce->values_gauge[i] =
          ((double)diff) / (CDTIME_T_TO_DOUBLE(vl->time - ce->last_time));
      ce->values_raw[i].derive = vl->values[i].derive;
    } break;` 

> So, from my POV, documentation needs more explanations about Collectd internals.

Sure, that's true. I think it's a common problem in several write plugins. I've looked at csv and write_tsdb and the documentation and implementation also differs.

Comment (rpv-tomsk) [reply src/write_influxdb_udp.c]: >I've looked at uc_update at utils_cache.c but I'm probably missing something.

Let's discuss. Ask your questions )

Comment (carlospeon) [reply src/write_influxdb_udp.c]: >     2. Another question. Do you have much metrics in your system?

Several thousands with an interval of 10. Metrics with rates are the usual ones: cpu, interface, and some custom made (the last not in all clients).

> That would be good to make CPU usage compare, if you have much metrics :-)
> That will argue changes.

Sure, we can see it.

Comment (rpv-tomsk) [reply src/write_influxdb_udp.c]: Please take a look:
https://github.com/collectd/collectd/pull/3164/files

That is a draft, just to remove cache scan.
It uses two malloc+memcpy to provide values for write plugin.
If we introduce new callback or option to `plugin_register_write()`, we can avoid one of them.

that is draft, completely unchecked (only compilation succeded), so I can miss something with pointers and sizes, please be patient and fix that if found.

File: src/write_influxdb_udp.c (hunk 6 above)
Comment (carlospeon) [reply src/write_influxdb_udp.c]: I've committed a new implementation writing directly to udp buffer.

File: src/write_influxdb_udp.c (hunk 9 above)
Comment (carlospeon) [reply src/write_influxdb_udp.c]: > Please review this my opinion:
> 
> ```
> When StoreRate is false, then raw metric values sent via wire.
> For DERIVE types, what means raw value is sent, not its change, like for GAUGE, 
> COUNTER and ABSOLUTE.
> 
> If StoreRate is true, then change of metric value is sent for DERIVE, COUNTER 
> and ABSOLUTE types - i.e. rate. GAUGE type always sent as-is.
> ```

I committed this, but I skipped

>```
>For DERIVE types, what means raw value is sent, not its change, like for GAUGE, 
> COUNTER and ABSOLUTE.
>```

Just because I didn't understand it properly. Maybe we should change the option `StoreRates` for something like:

```
StoreAsGauges. If true store derive, counter and absolute types as gauges.
```

> Please take a look:
> https://github.com/collectd/collectd/pull/3164/files

I did a quick test:
kernel: traps: reader#2[25562] general protection ip:7f14b82aa59c sp:7f14b065a588 error:0 in libc-2.17.so[7f14b8225000+1c2000]
systemd: collectd.service: main process exited, code=killed, status=11/SEGV

not enough time to debug it yet.

rpv-tomsk: This needs `gdb` ...
I checked that code on `csv` plugin, and it worked. 
Did you update all binaries and libraries to new versions?

rpv-tomsk: >I committed this, but I skipped
That is ok. )
>Just because I didn't understand it properly. 

Setup output to `csv` files and look for stored values.
With 'StoreRate false' you will see 'raw' values. 
The same values are sent "to wire" by write plugins.

>StoreAsGauges. If true store derive, counter and absolute types as gauges.

Maybe, I don't understand word "gauge" properly. Sorry, I'm not native speaker.

If true, then Derive will be sent/stored as derive, really )
If false, Derive is sent as 'counter' which is raw value.

And derive of that value calculated/displayed later. For example, rrd database stores raw values.
That may also be good, if your time-series-database saves these raw values:  while it displays chart based on calculated derive, it also can show raw values for some point of time.

---
>Maybe we should change the option StoreRates for something like: StoreAsGauges.

I don't think such renaming is good idea. It is better to have option name unified between plugins.

carlospeon: > This needs `gdb` ...
> I checked that code on `csv` plugin, and it worked.
> Did you update all binaries and libraries to new versions?

Oh, I see... all plugins need to be recompiled and copied.

Here are user and system cpu times for collectd process:

Screenshot from 2019-05-26 18-50-49

values before gap are from master branch and after gap from store-rate branch. The last one seems a bit higher.

Overall cpu usage seems unaffected (server has 2x12 physical cores)

rpv-tomsk: >The last one seems a bit higher.

Ouch, unexpectedly...

carlospeon: > Turns out that introduced additional 'malloc+memcpy' (which can be avoided, though) takes more cpu than cache search.

Probably a more extensive testing is needed.

Do you think this PR is suitable/ready for commit? 

On the testing side I'm running it successfully ( n * 1000 hosts ==> collectd == loopback, packetsize 16k ==> influxdb) and collectd memory is plain at 800MB.

carlospeon:
Hello @rpv-tomsk:

Is this PR acceptable? anything more I should do?

Regards,
Carlos.

dago:
@carlospeon Can you please document the new plugin in `src/collectd.conf.in` ?

carlospeon:
> @carlospeon Can you please document the new plugin in `src/collectd.conf.in` ?

Thank you @dago, I think it's done.

systemcrash:
Hello people, sorry to 'hijack' this PR conversation. But does this PR regard just the 'older' v1.x Line protocol, or does it also cover [the 'newer' v2](https://docs.influxdata.com/influxdb/v2.0/reference/syntax/line-protocol/)? 

They seem the same, just want to confirm 😄

PR #3163, opened 2019-05-24
virt plugin: improvements

rpv-tomsk:
Changelog: virt plugin: improvements and fixes.

1)
```
virt plugin: Added ExtraStats selector 'memory' 
This allows to disable virDomainMemoryStats calls when hypervisor/drivers
does not supports that function or it does not provide additional details.
```

2)
```
virt plugin: Added ExtraStats selector 'vcpu'  
This allows to disable virDomainGetVcpus calls when hypervisor/driver
does not supports that function.
```

3)
```
virt plugin: Do not request cpu maps when not required
```

4)
```
virt plugin: Disable ExtraStats selectors when unsupported by libvirt 
When function is not supported by driver, log filled by continuous messages like:

libvirt: Domain Config error : this function is not supported by the connection driver: virDomainGetDiskErrors

libvirt API allows us to detect such cases and disable unsupported calls.
```

5)
```
virt plugin: Added timeout to event_loop thread
As per virEventRunDefaultImpl() documentation, this function will block
forever if there are no registered event handlers.

This leads to Collectd is unable to correctly stop event_loop thread
if libvirtd was restarted.
```
https://libvirt.org/html/libvirt-libvirt-event.html#virEventRunDefaultImpl

6)
```
virt plugin: Added connection state check via virConnectIsAlive()

Before this fix, if libvirt daemon was stopped, Collectd starts to spam logs with lot of "failed" messages.
With this fix it correctly re-establishes connection.
```

7)
```
virt plugin: Fixed typos and inconsistencies
```

image

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1684,7 +1684,7 @@
 #	InterfaceFormat name
 #	PluginInstanceFormat name
 #	Instances 1
-#	ExtraStats "cpu_util disk disk_err domain_state fs_info job_stats_background pcpu perf vcpupin disk_physical disk_allocation disk_capacity"
+#	ExtraStats "cpu_util disk disk_err domain_state fs_info job_stats_background pcpu perf vcpu vcpupin disk_physical disk_allocation disk_capacity memory"
 #	PersistentNotification false
 #</Plugin>
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -9520,6 +9520,9 @@ Requires libvirt API version I<1.2.9> or later.
 a domain. Only one type of job statistics can be collected at the same time.
 Requires libvirt API version I<1.2.9> or later.
 
+=item B<memory>: report statistics about memory usage details, provided
+by libvirt virDomainMemoryStats() function.
+
 =item B<pcpu>: report the physical user/system cpu time consumed by the hypervisor, per-vm.
 Requires libvirt API version I<0.9.11> or later.
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -9531,6 +9531,8 @@ metrics they must be enabled for domain and supported by the platform. Requires
 libvirt API version I<1.3.3> or later.
 B<Note>: I<perf> metrics can't be collected if I<intel_rdt> plugin is enabled.
 
+=item B<vcpu>: report domain virtual CPUs utilisation.
+
 =item B<vcpupin>: report pinning of domain VCPUs to host physical CPUs.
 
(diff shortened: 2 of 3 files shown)


rpv-tomsk:
This code is used in my production.
Can you please review/merge it?

Thanks.

rpv-tomsk:
Many thanks for your review, I will push commits with these fixed now.

>As per virConnGetLastError documentation, it may be better to use virGetLastError (or even simpler virGetLastErrorCode as only the error code is used)

virGetLastErrorCode is too new. )

>* This method is not protected against access from multiple
> * threads. In a multi-threaded application, always use the
> * global virGetLastError() API which is backed by thread
> * local storage.

Can you please explain me, what "which is backed by thread local storage" means and how it works?

anaudx:
> virGetLastErrorCode is too new. )

What is the rule to be used to decide what is too new? Is that up to a particular version of libvirt?
> Can you please explain me, what "which is backed by thread local storage" means and how it works?

I understand that each thread has its own "private" storage but I don't know the details how this works, unfortunately.

rpv-tomsk:
>LGTM

You should add review and approve it, then merge can be done.

Thanks.

rpv-tomsk:
>I understand that each thread has its own "private" storage but I don't know the details how this works, unfortunately.

I carefully looked into libvirt source and found implementation in `src/util/virerror.c`, function `virLastErrorObject()` which uses `virThreadLocalGet`.

rpv-tomsk:
>What is the rule to be used to decide what is too new? Is that up to a particular version of libvirt?

I have Debian 7 in use, that uses libvirt-1.2.9, so that is my own need, at first.
At second, `virGetLastErrorCode()` introduced in `libvirt-4.5` only, while current `Debian stable` uses only `libvirt-3.0`...

Also, plugin code checks for versions up to 0.9.x and there was no cleanup for these defines.
There is no much overhead in `virGetLastError()` )

Thanks for your careful review!

rpv-tomsk:
>LGTM

Thanks, but, unfortunately, that is not enough to merge. Merging is still blocked. :(

Development is stopped by these 'strict rules'.

anaudx:
@rpv-tomsk Many thanks for your detailed answers.
About the merge issue, I don't know how can I help (I am not the author of these strict rules).
It looks like I am still missing some write access so that my approval wan automatically merge the PR.
We already reported this issue in previous PR (#3116) but without effect.
@octo @mrunge Can you fix the issue and grant that my approval will effectively merge the PR?

mrunge:
In my understanding, there is an issue with the automation. I'm happy to pull the trigger here, but your approval should be enough to merge this, @anaudx

anaudx:
@mrunge Thank you!

PR #3164, opened 2019-05-26
utils_cache: How to avoid uc_get_rate cache scan, draft.

rpv-tomsk:


rpv-tomsk:
cc @sradco

Hi,
As you implemented 'write' plugin recently, invite you to discuss the question, should we do some work to optimize `uc_get_rate` or not? The main goal is avoid cache search, and choose the way how to provide 'rates' to write plugins.

PR #3169, opened 2019-06-07
Replace illegal dot characters in prometheus keys with underscores

zarath:
ChangeLog: write_prometheus: Replace illegal dot characters in prometheus keys with underscores


zarath:
ChangeLog: write_prometheus: Replace illegal dot characters in prometheus keys with underscores

eero-t:
FYI: collectd v6 branch got proper support for this in #4206, which converts _all_ characters not accepted by Prometheus.

Unless this will do something similar, and not just handle dots, I think this can be closed.

Issue #3166, opened 2019-06-03
collectd fails to build with Python 3.8


stratakis:
*   Version of collectd: 5.8.1
*   Operating system / distribution:  Fedora Rawhide

Currently on Fedora we are testing rebuilding the distro against Python 3.8 and collectd is one of the packages that failed to build from source.

This happens when the --enable-python option is given, due to `python3-config --libs` [invocation](https://github.com/collectd/collectd/blob/dfb17826ceb8223cca9009b7b9f4194252ff994d/configure.ac#L4774) requiring also the `--embed` flag, with Python 3.8, in order to link the application to libpython.

As this is applicable only for 3.8, using the same flag on lesser versions will make the compilation fail as well, so a conditional would be required there. [More info](https://docs.python.org/3.8/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build).

Full build log attached.

[build.log.gz](https://github.com/collectd/collectd/files/3248966/build.log.gz)

rpv-tomsk:
Thanks for your report!
Patches are welcome.

PR #3170, opened 2019-06-07
Add Python 3.8 compatibility

stratakis:
From Python 3.8 and onwards C extensions are no longer
linked to libpython so in order to embed python within
an application the --embed flag needs to be added to
python3-config --libs. A fall back is provided as the
command will error out on previous python versions.

References: https://bugs.python.org/issue36721

Resolves: #3166 

ChangeLog: Add Python 3.8 compatibility

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4771,7 +4771,7 @@ if test "$PYTHON_CONFIG" != ""; then
   if test $? -ne 0; then
     with_libpython="no"
   fi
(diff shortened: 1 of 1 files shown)


rpv-tomsk:

Self-Perfection:
I suspect this fix is incomplete and fetching `$LIBPYTHON_LDFLAGS` from `python-config` has to be fixed as well.

relevant bugreport: https://bugs.archlinux.org/task/64533

PR #3178, opened 2019-06-18
regex match: Fix unexpected match with empty meta data

ashie:
ChangeLog: regex match: Fix unexpected match with empty meta data

If a user set

  MetaData "foo" "bar"

and a metric does not have meta data (vl->meta == NULL), it causes
unexpected match.

Existance of MetaData config should be checked before meta data in
a metric.

See also: #1930

Signed-off-by: Takuro Ashie <ashie@clear-code.com>


rpv-tomsk:
Thanks!

PR #3182, opened 2019-06-19
Fix return value or loglevel for several plugins

### LINKED ISSUE #3180: plugin load fails to parse ReportRelative config option
When using the option `ReportRelative` in the load plugin on 5.9.0, the config parsing fails, and collectd refuses to start:

```
> cat /tmp/collectd.conf
LoadPlugin load
<Plugin "load">
  ReportRelative true
</Plugin>
> collectd -C /tmp/collectd.conf -T
plugin_load: plugin "load" successfully loaded.
Error: Parsing the config file failed!
```

faxm0dem:
Some plugins' RC was != 0 and started to fail due to stricter verification introduced by 3b9c7b2.
This commit fixes those return values. For some plugins, fix verbosity of error message: non-zero rc
should be error, not warning.

Fixes #3180

What I chose to do:

* fix RC when I judged it to be non-fatal (load, logfile)
*  fix loglevel from WARNING to ERROR in other cases (barometer, openvpn, ted)
* add log message (syslog)

ChangeLog: fix incorrect return codes or log level for several plugin config callbacks

### DIFF
--- a/src/daemon/configfile.c
+++ b/src/daemon/configfile.c
@@ -267,7 +267,8 @@ static int dispatch_loadplugin(oconfig_item_t *ci) {
 
   /* default to the global interval set before loading this plugin */
   plugin_ctx_t ctx = {
-      .interval = cf_get_default_interval(), .name = strdup(name),
+      .interval = cf_get_default_interval(),
+      .name = strdup(name),
   };
   if (ctx.name == NULL)
     return ENOMEM;
@@ -413,10 +414,11 @@ static int dispatch_block_plugin(oconfig_item_t *ci) {
   /* Hm, no complex plugin found. Dispatch the values one by one */
   for (int i = 0, ret = 0; i < ci->children_num; i++) {
     if (ci->children[i].children == NULL) {
-      oconfig_item_t *child = ci -> children + i;
+      oconfig_item_t *child = ci->children + i;
       ret = dispatch_value_plugin(name, child);
       if (ret != 0) {
-        ERROR("Plugin %s failed to handle option %s, return code: %i", name, child -> key, ret);
+        ERROR("Plugin %s failed to handle option %s, return code: %i", name,
+              child->key, ret);
         return ret;
       }
     } else {
@@ -478,9 +480,9 @@ static int cf_ci_replace_child(oconfig_item_t *dst, oconfig_item_t *src,
     return 0;
   }
 
-  temp = realloc(dst->children,
-                 sizeof(oconfig_item_t) *
-                     (dst->children_num + src->children_num - 1));
+  temp =
+      realloc(dst->children, sizeof(oconfig_item_t) *
+                                 (dst->children_num + src->children_num - 1));
   if (temp == NULL) {
     ERROR("configfile: realloc failed.");
     return -1;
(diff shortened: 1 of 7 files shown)


src/daemon/configfile.c
```diff
@@ -414,8 +414,10 @@ static int dispatch_block_plugin(oconfig_item_t *ci) {
   for (int i = 0, ret = 0; i < ci->children_num; i++) {
     if (ci->children[i].children == NULL) {
       ret = dispatch_value_plugin(name, ci->children + i);
-      if (ret != 0)
+      if (ret != 0) {
+        ERROR("dispatch for plugin %s returned non-zero code : %i", name, ret);
```
rpv-tomsk: I think we need smth more human-readable, like "Plugin `%s` failed to handle option `%s`, return code: %i".

src/logfile.c
```diff
@@ -53,7 +53,7 @@ static int logfile_config(const char *key, const char *value) {
     if (log_level < 0) {
       log_level = LOG_INFO;
       ERROR("logfile: invalid loglevel [%s] defaulting to 'info'", value);
-      return 1;
+      return 0;
```
rpv-tomsk: That was 'error' by code design. Developer explicitly choosed to return "1".
Now you want to return "0", that does message not so critical as it was before.
I think message log level should be changed to .... WARNING or NOTICE ("project" should have policy about such case).

faxm0dem: I agree with warning

src/daemon/configfile.c
faxm0dem: done

src/ted.c
```diff
@@ -243,7 +243,7 @@ static int ted_config(const char *key, const char *value) {
 
     tmp = atoi(value);
     if (tmp < 0) {
-      WARNING("ted plugin: Invalid retry count: %i", tmp);
+      ERROR("ted plugin: Invalid retry count: %i", tmp);
```
rpv-tomsk: I think we need to fix a bug here: output "%s" + value.

```

**Comment by mrunge** [comment]
Can you please re-run clang-format?

**Comment by faxm0dem** [comment]
I reran the clang lint, everything ok now? I can squash everything

**Comment by rpv-tomsk** [comment]
As you can see, that is not ok and CI still failing.
Use `contrib/format.sh` as proposed by CI.

**Comment by faxm0dem** [comment]
sorry about that - should be okay now

**Comment by rpv-tomsk** [comment]
Further reading:

https://github.com/orgs/collectd/teams/trusted-contributors/discussions/1

#2958

**Comment by mrunge** [comment]
Hmm, automerge doesn't work?

**Comment by rpv-tomsk** [comment]
oops ? :-))) unexpectedly, right? :-))))

**Comment by rpv-tomsk** [comment]
Team prevents other people from project developing.

Live with it and enjoy.

**Comment by mrunge** [comment]
So, automerge issues have nothing to do with this patch. I'll merge it now manually. Let's take the discussion from here.
I see your point @rpv-tomsk , let's see what we can do to fix this.

File: src/syslog.c [hunk 4]
```diff
@@ -56,7 +57,8 @@ static int sl_config(const char *key, const char *value) {
   } else if (strcasecmp(key, "NotifyLevel") == 0) {
     notif_severity = parse_notif_severity(value);
     if (notif_severity < 0)
```
kwiatrox: Missing brackets for if? "Return 1" is outside of if scope.

PR #3189, opened 2019-06-24
network plugin: Add a new metadata "network:ip_address"

ashie:
ChangeLog: network plugin: Add a new metadata "network:ip_address"

Although the telemetry data already have the "host" identifier, it can be determined by clients. This additional metadata will help you to detect spoofing at server side.

It will also help you to filter the data by sender's address.


ashie:
Sorry, now I notice that it's missing the implementation for encrypted cases.
I'll address them.

Issue #1301, opened 2015-10-12
Testsuite for Solaris when building 32 bit breaks


dago:
When building the latest version on Solaris 10 with 32 bit the test suite does not compile:
```
gmake  test_common test_meta_data test_utils_avltree test_utils_heap test_utils_time test_utils_subst
gmake[4]: Entering directory '/home/dam/work/collectd/src/daemon'
  CC       common_test.o
In file included from collectd.h:31:0,
                 from common.h:31,
                 from common_test.c:28:
../../src/config.h:1622:0: error: "_FILE_OFFSET_BITS" redefined [-Werror]
 #define _FILE_OFFSET_BITS 64
 ^
In file included from /usr/include/inttypes.h:20:0,
                 from ../../src/testing.h:27,
                 from common_test.c:27:
/opt/csw/lib/gcc/i386-pc-solaris2.10/4.9.2/include-fixed/sys/feature_tests.h:196:0: note: this is the location of the previous definition
 #define _FILE_OFFSET_BITS 32
 ^
cc1: all warnings being treated as errors
Makefile:986: recipe for target 'common_test.o' failed
```

This has probably to do with largefiles as described in lf64(5) being different in the standard GCC headers and the determined width.

mfournier:
@dago are you able to retest current master by any chance ? It looks like #1442 should have fixed this. Thanks !

dago:
Sure, this is integrated in our buildbot:
  https://buildfarm.opencsw.org/buildbot/waterfall?category=collectd
If you go back in history you see that is passes on the 57dfc4722bb1c4d7f735c71aa113b9f6bbdce964  commit:
  https://buildfarm.opencsw.org/buildbot/builders/collectd-solaris10-i386
That means the issue is fixed now, but not with the commit from #1442.

Feel free to add a link to our buildbot for CollectD to your development page:
  https://collectd.org/dev-info.shtml

Best regards -- Dago

mfournier:
Great, thanks @dago !

dago:
The issue reappeared in the latest master:
```
  CCLD     write_http.la
gmake  test_format_json test_utils_latency test_utils_vl_lookup test_utils_mount test_plugin_ceph
gmake[4]: Entering directory '/home/dam/work/collectd/src'
  CC       utils_format_json_test.o
In file included from ./daemon/collectd.h:31:0,
                 from utils_format_json_test.c:28:
[...]

PR #3193, opened 2019-06-27
Add standard include early or _FILE_OFFSET_BITS will have definition …

dago:
…mismatches on Solaris.
This basically is a reappearance of issue #1301.

Changelog: Add standard include early or _FILE_OFFSET_BITS will have definition mismatches on Solaris.

### DIFF
--- a/src/daemon/globals.c
+++ b/src/daemon/globals.c
@@ -21,8 +21,8 @@
  * DEALINGS IN THE SOFTWARE.
  **/
 
-#include "globals.h"
 #include "utils/common/common.h"
+#include "globals.h"
 
 #if HAVE_KSTAT_H
 #include <kstat.h>
--- a/src/daemon/globals.c
+++ b/src/daemon/globals.c
@@ -21,8 +21,14 @@
  * DEALINGS IN THE SOFTWARE.
  **/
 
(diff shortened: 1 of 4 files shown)


rpv-tomsk:
Hi,

Thanks for your work on this issue. However, CI asks to run `contrib/format.sh`. Can you please do this?

dago:
After running `contrib/format.sh` on the updated files I notice that the formatter undoes my changes (among other probably useful things):
```
diff --git a/src/daemon/globals.c b/src/daemon/globals.c
index 5409294..a76a44c 100644
--- a/src/daemon/globals.c
+++ b/src/daemon/globals.c
@@ -21,8 +21,8 @@
  * DEALINGS IN THE SOFTWARE.
  **/
 
-#include "utils/common/common.h"
 #include "globals.h"
 
 #if HAVE_KSTAT_H
 #include <kstat.h>
```

rpv-tomsk:
Then needs to add 
```
// clang-format off
[ ..  includes ..]
// clang-format on
```

around these includes, and comment about explicit include order is required.

rpv-tomsk:
Thanks for your work!

I also did all the best I could on this issue.

PR #3195, opened 2019-06-28
Use GCC-specific flags only when compiling with GCC

dago:
configure unconditionally adds gcc-specific W-style flags which break compilation when compiling with other compilers like Oracle Studio. The PR adds conditionals around the W-style flags to only be applied when compiling with GCC.

Changelog: Build system: Add conditionals in configure around the W-style flags to only be applied when compiling with GCC.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -777,8 +777,10 @@ AC_CHECK_FUNCS_ONCE([ \
 
 AC_FUNC_STRERROR_R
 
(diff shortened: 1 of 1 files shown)


mrunge:
Could you please fix the ChangeLog, see https://github.com/collectd/collectd/blob/master/docs/CONTRIBUTING.md#changelog ?

dago:
Sure, ChangeLog added.

PR #3203, opened 2019-07-02
Add missing definitions for libnetsnmpagent

dago:
Without these lines `--with-libnetsnmpagent=<dir>` will not set `CPPFLAGS` and `LDFLAGS` properly. The structure is now the same as it is already the case for the corresponding libnetsnmp library.

ChangeLog: Build system: Add missing definitions for libnetsnmpagent in configure.ac


mrunge:
Thanks again!

PR #3204, opened 2019-07-04
UBI plugin: collect some stats about the UBIFS

pticon:
ChangeLog: The new "ubifs" plugin reports block state for flash memory devices with UBIFS filesystem

This plugin collects some stats about the UBIFS:
 * bad physical underlying blocks (bad_peb_count)
 * max erase counter value (max_ec)

Signed-off-by: Pierre Lebleu <pme.lebleu@gmail.com>


ChangeLog
```diff
@@ -74,6 +74,7 @@
 	  Thanks to Pavel Rochnyak. #2570, #2673
 	* Write Syslog plugin: The new "write_syslog" plugin writes value
 	  lists as syslog messages. Thanks to Shirly Radco. #3019
+	* Write Ubi plugin: The new "read_ubi" plugin reads value for UBIFS.
```
rpv-tomsk: Hi,

Your plugin is not part of `Collectd-5.8.1`, so this line is wrong.
Moreover, please don't touch Changelog, that file is updated semi-automatically while new release preparation, based on "Changelog" line in PR description.

rpv-tomsk: The next:

`Write Ubi plugin: The new "read_ubi" plugin reads value for UBIFS.`

"Write Ubi" was written based on "Write Syslog" or similar. 
If you read "Write Syslog" line carefully, you find that plugin writes values to syslog.
Your plugin does not write metrics, it _reads_ them.

`The new "read_ubi" plugin reads value for UBIFS.`

1) Your plugin is not "read_ubi".
2) "reads value for UBIFS" is not informative.

Probably, Changelog line should be something like "The new "ubifs" plugin reports block state for flash memory devices with UBIFS filesystem".   As you more familar with this, you can write this shorter and  more precise.

configure.ac
```diff
@@ -6420,6 +6420,7 @@ plugin_tcpconns="no"
 plugin_ted="no"
 plugin_thermal="no"
 plugin_turbostat="no"
+plugin_ubi="no"
```
rpv-tomsk: You also need to add `plugin_ubi="yes"` few lines below, enabling it for Linux by default, like others.

configure.ac
```diff
@@ -6913,6 +6914,7 @@ AC_PLUGIN([thermal],             [$plugin_thermal],           [Linux ACPI therma
 AC_PLUGIN([threshold],           [yes],                       [Threshold checking plugin])
 AC_PLUGIN([tokyotyrant],         [$with_libtokyotyrant],      [TokyoTyrant database statistics])
 AC_PLUGIN([turbostat],           [$plugin_turbostat],         [Advanced statistic on Intel cpu states])
+AC_PLUGIN([ubi],                 [$plugin_ubi],               [UBI statistics])
```
rpv-tomsk: UBI or UBIFS ?

src/ubi.c
```diff
@@ -0,0 +1,148 @@
...
+
+#define SYS_PATH "/sys/class/ubi/"
+
+/*
+ * Device attributes
+ */
+#define DEV_BAD_COUNT                                                          \
+  "bad_peb_count" // Count of bad physical eraseblocks on the underlying MTD
+                  // device.
+#define MAXIMUM_ERASE "max_ec" // Current maximum erase counter value
```
rpv-tomsk: Did this counter changes frequently or rarely?
Do you have sysfs documentation for all possible counters and their values for UBI/MTD?

From my understanding, this counter is very static. Is I'm correct?

rpv-tomsk: I think, `reserved_for_bad` is also very useful for your plugin, because it provides us number of eraseblocks reserved for bad block handling. If it becomes zero, there will be troubles when the next bad block appears. So plugin might watch this value too.

rpv-tomsk:
Notes
https://unix.stackexchange.com/questions/44949/ubi-bad-block-management
https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-ubi

configure.ac
pticon: Not sure if we want to enable by default, because it only makes sense for the NAND device using the UBIFS filesystem. What do you think ?

rpv-tomsk: We do not enable it by default, but _compile_ by default )

pticon: Ok, thanks!

ChangeLog
pticon: Ok, thank you for the clarification.

configure.ac
pticon: Let's call the plugin UBI

src/ubi.c
pticon: According to the documentation:
_bad_peb_count_: count of bad physical eraseblocks
_max_ec_: current highest erase counter value

Those 2 counters *should* not move a lot and that's why it is good idea to monitor them.

Indeed, when the '_bad_peb_count_' starts to increase, something wrong is probably happening because your device is potentially dying.

Concerning '_max_ec_', this counter will increase by sure but it should not increase too fast, otherwise, it means you are using your flash too fast.

pticon: The counter 'reserved_for_bad' will not move, it is static. There is no need to monitor it

rpv-tomsk:
>Should I drop this pull request and create an other one with an atomic commit ?

I don't think you should.

You should wait for someone who is trusted contributor to merge this PR.
I'm not this one.

Thanks for your work.

pticon:
@octo What do you think about this ?

feckert:
@dago @pticon 

I was about to create a plugin for OpenWrt for the [luci app-statistics](https://github.com/openwrt/luci/tree/master/applications/luci-app-statistics). Unfortunately it's not possible to create a valid picture, because the lines are always 0. The value **max_ec** should **68**. The value **bad_peb_count** is 0 but that is OK, because I have no badblock on my ubi nand.

Screenshot_2020-06-23 VR2-106149 - Diagramme - LuCI

From my point of view the used data type COUNTER is wrong. I do not want to get the rate But this is calculated with the type COUNTER. Better would be the tpye GAUGE. Or am I missing something here?
https://collectd.org/wiki/index.php/Data_source. 

```
rrdtool info ubi/max_ec-ubi0.rrd
filename = "ubi/max_ec-ubi0.rrd"
rrd_version = "0001"
step = 30blob:https://192.168.0.53/77f43508-37b5-4ca9-897b-744f5b5692fa
last_update = 1592910344
ds[value].type = "COUNTER"
ds[value].minimal_heartbeat = 60
ds[value].min = 0.0000000000e+00
ds[value].max = NaN
ds[value].last_ds = "68"
ds[value].value = 0.0000000000e+00
ds[value].unknown_sec = 0
rra[0].cf = "AVERAGE"
rra[0].rows = 144
rra[0].pdp_per_row = 1
rra[0].xff = 1.0000000000e-01
rra[0].cdp_prep[0].value = NaN
rra[0].cdp_prep[0].unknown_datapoints = 0
rra[1].cf = "AVERAGE"
rra[1].rows = 144
rra[1].pdp_per_row = 20
rra[1].xff = 1.0000000000e-01
rra[1].cdp_prep[0].value = 0.0000000000e+00
rra[1].cdp_prep[0].unknown_datapoints = 0
rra[2].cf = "AVERAGE"
rra[2].rows = 144
rra[2].pdp_per_row = 140
rra[2].xff = 1.0000000000e-01
rra[2].cdp_prep[0].value = 0.0000000000e+00
rra[2].cdp_prep[0].unknown_datapoints = 0
rra[3].cf = "AVERAGE"
rra[3].rows = 144
rra[3].pdp_per_row = 620
rra[3].xff = 1.0000000000e-01
rra[3].cdp_prep[0].value = 0.0000000000e+00
rra[3].cdp_prep[0].unknown_datapoints = 0
rra[4].cf = "AVERAGE"
rra[4].rows = 144
rra[4].pdp_per_row = 7320
rra[4].xff = 1.0000000000e-01
rra[4].cdp_prep[0].value = 0.0000000000e+00
rra[4].cdp_prep[0].unknown_datapoints = 2249
```

feckert:
@pticon @dago 
I have staged a [pull request](https://github.com/collectd/collectd/pull/3486) to fix this and so I think that's right.

Issue #3112, opened 2019-02-22
PowerDNS plugin Bug : unreachables


apandiyan:
*   Version of collectd: collectd-5.8.1-1.el7.x86_64 (Epel)
*   Operating system / distribution: CentOS 7.6.1810
*   Kernel version (if applicable): 3.10.0-957

## Expected behavior

PowerDNS plugin should send statistics to InfluxDB

## Actual behavior

PowerDNS plugin doesn't send statistics to InfluxDB and it provides info in the log that

Feb 22 18:39:50 ns1 collectd[18738]: powerdns plugin: submit: Not found in lookup table: unreachables = 34677;

## Steps to reproduce

*   Install PDNS (pdns-4.1.6-1pdns.el7.x86_64) from Epel
*   Install collectd (collectd-5.8.1-1.el7.x86_64) from Epel
*   Configure powerdns plugin as mentioned in wiki document (https://collectd.org/wiki/index.php/Plugin:PowerDNS)

rubenk:
Please share your configuration here

apandiyan:
Hi,

This is my configuration.

--xx--

Hostname    "localhost"
BaseDir     "/var/lib/collectd"
PIDFile     "/var/run/collectd.pid"
PluginDir   "/usr/lib64/collectd"
TypesDB     "/usr/share/collectd/types.db"
LoadPlugin syslog
LoadPlugin cpu
LoadPlugin cpufreq
LoadPlugin cpusleep
LoadPlugin df
LoadPlugin disk
LoadPlugin ethstat
LoadPlugin interface
LoadPlugin load
LoadPlugin memory
LoadPlugin network
LoadPlugin powerdns
LoadPlugin processes
LoadPlugin protocols
LoadPlugin swap
LoadPlugin uptime
LoadPlugin users
<Plugin network>
        Server "INFLUXDB_SERVER" "PORT"
</Plugin>
<Plugin "powerdns">
  <Server "auth_dns">
    Collect "latency"
    Collect "udp-answers" "udp-queries"
    Collect "recursing-answers" "recursing-questions"
    Collect "packetcache-hit" "packetcache-miss"
    Collect "query-cache-hit" "query-cache-miss"
    Collect "packetcache-size" "qsize-q"
    Collect "corrupt-packets" "servfail-packets"
    Collect "tcp-answers" "tcp-queries"
    Collect "udp4-answers" "udp4-queries"
    Collect "udp6-answers" "udp6-queries"
[...]

PR #3205, opened 2019-07-05
powerdns: Added 'unreachables' reporting

rpv-tomsk:
Changelog: powerdns: Added 'unreachables' reporting
Issue: #3112

Wiki https://collectd.org/wiki/index.php/Plugin:PowerDNS provides configuration with 'unreachables' reported, but that is not supported by plugin.
This PR fixes that.


rpv-tomsk:
This is very serious change, so it requires to be reviewed by as many as possible reviewers.

Thanks for revoking merge permission.

PR #3207, opened 2019-07-08
Fix bug that leads to CPPFLAGS gets overridden with CFLAGS when libxm…

dago:
…ms is enabled

ChangeLog: Build system: Fix bug that leads to CPPFLAGS gets overridden with CFLAGS when libxmms is enabled
ChangeLog: Network plugin: Fix typo in message and formatting

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5769,7 +5769,7 @@ if test "x$with_libxmms" = "xyes"; then
 fi
 
 if test "x$with_libxmms" = "xyes"; then
-  SAVE_CPPFLAGS="$CFLAGS"
+  SAVE_CPPFLAGS="$CPPFLAGS"
   CPPFLAGS="$with_xmms_cflags"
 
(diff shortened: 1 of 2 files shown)


rpv-tomsk:
What makes you to do these useless appruves? =)

dago:
Why is this ChangeLog check still in waiting state? I added the respective line in the description right from the start this time.

mrunge:
Unfortunately, I don't have a clue how to trigger  the checks again, other than adding new commits here.

If you want to add a useful commit, you could fix the typo in https://github.com/collectd/collectd/blob/master/src/network.c#L384

dago:
Ok, done, let's see if this works now.

mrunge:
Thank you again for your work.

I'll just go ahead and merge this, even when we don't know why the ChangeLog check did not report anything.

PR #3224, opened 2019-07-18
Move trusted-contributors down the list

mrunge:
since the last matching pattern takes the most precedence.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,7 +4,6 @@
 # These owners will be the default owners for everything in the repo. Unless a
 # later match takes precedence, # @trusted-contributors will be requested for
 # review when someone opens a pull request.
-*       @trusted-contributors
 
 /src/intel_pmu.c	@kwiatrox @sunkuranganath
 /src/intel_rdt.c	@kwiatrox @sunkuranganath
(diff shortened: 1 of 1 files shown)


rjablonx:
when I read more and more about this codeowners concept, then I think that this change will effectively disable every line above *trusted-contributors (as last pattern will probably match everything, also parts matched earlier).
So maybe there is another way to improve current behavior but directly in github options - I discovered that undesired effect when every code owner has to accept change on entire tree can be disabled in github project options: https://help.github.com/en/articles/enabling-required-reviews-for-pull-requests in section  "Require review from Code Owners" (but this change requires admin right on github project)

sunkuranganath:
See that this is already merged now. Question is would 'code owners' still be able to merge changes relevant to the files they are assigned?
Earlier I used to see option to "merge" under review. Now I dont see it any more.

rpv-tomsk:
>@rpv-tomsk Could you please check if this merge helps you?

Sorry, I can't check this as I'm in @collectd/untrusted-contributors , not in @collectd/trusted-contributors.

rpv-tomsk:
Thanks for your efforts.

PR #3227, opened 2019-07-22
postgresql: Enable to treat meta data at writer

ashie:
ChangeLog: Enable to handle meta data at writer

Although users may want to handle some meta data at SQL world but postgresql plugin doesn't provide a way to do it. For example a user may want to write `network:username` to DB.

This patch add a new optional argument `$10` on executing a SQL command to pass a JSON string which stores meta data. `$10` is added only when `MetaData` option is added to `<Writer>` so that users of old versions can safely upgrade to this version.


dago:
Other plugins already use `libjansson` for JSON writing, maybe for the sake of consistency it would be good to use that here too?

ashie:
> Other plugins already use libjansson for JSON writing, maybe for the sake of consistency it would be good to use that here too?

It's reasobale, I'll address it.

ashie:
Hmm, libjansson is optional and only dpdk_telemetry plugin uses it.
On the other hand there is built-in API to format JSON (src/utils/format_json), and some plugins uses it.
Although former one has better API, I don't want to make this feature optional. So I'll choose later one (add a wrapper function to expose `meta_data_to_json()` to public then use it).

ashie:
In addition, I'll update the document & sample schema to clarify that `$10` is optional.

ashie:
> So I'll choose later one (add a wrapper function to expose meta_data_to_json() to public then use it).

I added a new function `format_json_meta_data()` for it, but it's not enough for my usage yet.
Because I need to select json keys to dump. I'll add this feature to this function.

ashie:
TODO list:
* [ ] Add `selected_keys` argument to `format_json_meta_data()`
* [ ] Replace `metadata_to_json()` with `format_json_meta_data()`
* [ ] Update the document
* [ ] Update the document & sample schema to clarify that $10 is optional.

ashie:
@collectd/trusted-contributors @dago or someone.
I've completed to rewrite this pull request.
Could you please review again if there is still a chance to merge it?

kenhys:
NOTE: CI build/experimental fails because of the following issue, it has nothing to do with this PR change.

fix(epics): Fix include path -- at least on Debian unstable.
https://github.com/collectd/collectd/pull/4298

kenhys:
> NOTE: CI build/experimental fails because of the following issue, it has nothing to do with this PR change.

I've predicate that It takes a longer time to be fixed (Build experimental issue See #4298)

Could you review @octo ? (except above issue)

src/utils/format_json/format_json.c
```diff
@@ -397,6 +404,28 @@ static int format_json_value_list_nocheck(char *buffer, /* {{{ */
...
+  char temp[temp_size];
+  int status;
+
+  status = meta_data_to_json(temp, sizeof(temp), meta, metadata_keys,
+                             metadata_keys_num);
+  if (status != 0)
+    return status;
+  temp_size = strlen(temp);
+
+  memcpy(buffer + (*ret_buffer_fill), temp, temp_size + 1);
```
eero-t, line 422: Reads one item past `temp` array end which is (eventually) going to trigger compiler warnings (and fail builds with `-Werror`).

Looking at existing code, `format_json_value_list_nocheck()` seems to have the same bug.

eero-t (commented):
At this point, I think new features should really go to v6 branch, not any more to the v5 `main` one.  Unfortunately `postgresql` plugin has been only partially ported to new v6 APIs...

src/utils/format_json/format_json.c
ashie: Ah, although it will be guarded just above `if (status != 0)` (`-ENOMEM`) in actual, certainly it's not good.
Probably we should use another variable and check length before calling `memcpy()`.

eero-t: No, it's OK after all, I'm just blind. I somehow managed to missed the strlen() call. :-/

Issue #3232, opened 2019-07-26
rrdtool is broken in 5.9.1


faxm0dem:
```
rrdtool plugin: failed to build values string
```

faxm0dem:
other plugins test for the return code of `ssnprintf`, e.g. nginx, postgresql, etc. so this is critical

faxm0dem:
See #3153

rpv-tomsk:
The reason is:

New function specification:
```
/* ssnprintf returns zero on success, one if truncation occurred
   and a negative integer onerror. */
int ssnprintf(char *str, size_t sz, const char *format, ...) {
  va_list ap;
  va_start(ap, format);

   int ret = vsnprintf(str, sz, format, ap);

   va_end(ap);

   if (ret < 0) {
    return ret;
  }
  return (size_t)ret >= sz;
} /* int ssnprintf */
```

Replaced function specification:

```
int snprintf (char *__s, size_t__n, const char *__fmt, ...)

Like sprintf(), but instead of assuming s to be of infinite size, 
no more than n characters (including the trailing NUL character) will be converted to s.

Returns the number of characters that would have been written to s if there were enough space.
```

*Returns the number of characters that would have been written to s if there were enough space.*

While new returns 0.

Sorry for missing this. I was in reviewers too, but has not checked that carefully enough (spent too much time fighting against Team new rules and checks).

Also affected plugin: `src/threshold.c`

Proposed patch:
-   if (ret < 0) {
-   return ret;
-  }
-  return (size_t)ret >= sz;
+ return ret;

But I don't know if this will recover GCC complaints...

faxm0dem:
your proposed change fixes rrdtool. I'm checking if it fixes postgresql just to make sure.
@sunkuranganath will it still build on RHEL8?

sunkuranganath:
@faxm0dem I dont have RHEL environment setup in our labs. 
So have to defer your question to @mrunge

mrunge:
Should be fixed by https://github.com/collectd/collectd/pull/3237

zebity:
Hi @mrunge & @rpv-tomsk ,

been away on break and saw that introduction of ssnprintf wrapper, broke some other code, my apologies.

Code snippet for wrapper was taken from here: https://github.com/collectd/collectd/pull/2895#issuecomment-437062237

Also be aware that the same code snippet is in: https://github.com/collectd/collectd/blob/master/src/collectdctl.c only this time in the static function:

[...]

PR #3233, opened 2019-07-26
WIP: fix rrdtool

faxm0dem:
this works around #3232 it's not a proper fix of course just a proof that 8f6aa697 is the root cause


rpv-tomsk:
=))) ouch-ouch, too many files touched ))

faxm0dem:
Yeah this was based against 5.9.1 I'll try your proposed patch Monday

mrunge:
In my understanding, this should be fixed with https://github.com/collectd/collectd/pull/3237

PR #3234, opened 2019-07-27
Remove including sys/sysctl.h on glibc systems

kraj:
Changelog: Remove sys/sysctl.h which is removed in Glibc 2.30

### DIFF
--- a/src/contextswitch.c
+++ b/src/contextswitch.c
@@ -26,7 +26,7 @@
 #include "plugin.h"
 #include "utils/common/common.h"
 
-#ifdef HAVE_SYS_SYSCTL_H
+#if defined(HAVE_SYS_SYSCTL_H) && !defined(__GLIBC__)
 #include <sys/sysctl.h>
 #endif
 
--- a/src/cpu.c
+++ b/src/cpu.c
@@ -60,7 +60,7 @@
 
 #if (defined(HAVE_SYSCTL) && HAVE_SYSCTL) ||                                   \
     (defined(HAVE_SYSCTLBYNAME) && HAVE_SYSCTLBYNAME)
-#ifdef HAVE_SYS_SYSCTL_H
+#if defined(HAVE_SYS_SYSCTL_H) && !defined(__GLIBC__)
 #include <sys/sysctl.h>
 #endif
 
--- a/src/memory.c
+++ b/src/memory.c
@@ -28,7 +28,7 @@
 #include "plugin.h"
 #include "utils/common/common.h"
 
-#ifdef HAVE_SYS_SYSCTL_H
+#if defined(HAVE_SYS_SYSCTL_H) && !defined(__GLIBC__)
 #include <sys/sysctl.h>
 #endif
 #ifdef HAVE_SYS_VMMETER_H
--- a/src/processes.c
+++ b/src/processes.c
@@ -87,7 +87,7 @@
 #if HAVE_MACH_VM_PROT_H
 #include <mach/vm_prot.h>
 #endif
-#if HAVE_SYS_SYSCTL_H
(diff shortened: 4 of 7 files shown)


mrunge:
Thank you for this PR! Would you be able to add a ChangeLog like described here? https://github.com/collectd/collectd/blob/master/docs/CONTRIBUTING.md#changelog

rubenk:
Sorry, but I don't think this is the right fix. If glibc doesn't ship this header then configure should detect that and not define HAVE_SYS_SYSCTL_H. Is that not the case?

kraj:
> Sorry, but I don't think this is the right fix. If glibc doesn't ship this header then configure should detect that and not define HAVE_SYS_SYSCTL_H. Is that not the case?

@rubenk yes that would be right but so far in 2.30 there is a warning and eventually, in future versions, it will be gone. So this was an attempt to squash the deprecation warning which breaks the build due to `-Werror`

rubenk:
In that case, can you detect this at ./configure time?

zebity:
Hi @kraj I stumped on your PR by way of raising issue: https://github.com/collectd/collectd/issues/3338

I have looked across Ubuntu / Fedora Linux , FreeBSD (sysctl originated in BSB 4.4) and MacOS and also the impacted collectd code.

I believe the resolution to this set of problem is relatively simple and does not involve changing the configure process.

The underlying cause is that in Linux the now preferred way to get kernel parameter data is via the process file system (procfs), which is why some of the code checks for procfs.

For BSD variants use of sysctl continues as the function and header are provided by OS code not glibc.

MacOS (Darwin) is a BSD mutant.

The reason the src/contextswitch.c PR works is because it is doing a check against sysctl && sysctlbyname . sysctlbyname is only available in BSD.

So in summary a clean way to fix this set of code which should work is to:
1. Check for procfs first (the Linux way)
   do procfs based version of stuff
2. Otherwise check for sysctl (the BSD way)
   do sysctl based version of stuff
3. Are there other cases ??
    Need to look at code a bit more...

The existing code across the various modules is very hard to understand as its such a mix of conditional compilations.

I believe a consistent approach can be applied across all of them, but each will need to be looked at carefully.

Thanks very much for helping with collectd, it needs all the help it can get ;-)

Cheers,

Zebity

qingling128:
Hi, we are running into the same issue on Ubuntu 1910 when building Collectd. Is this something that can be expected to be fixed in the short term? We are a bit concerned about using a workaround like disabling all warnings as we might miss other stuff revealed by the warnings.

zebity:
Hi qingling128,

I submitted a pull request some time ago that addressed this: https://github.com/collectd/collectd/pull/3342

Unfortunately this is been sitting unapproved for over a month now, so is in limbo land.

While annoying at the moment it can be worked around and is by doing your Ubuntu collectd build using the debian packaging make updates: https://github.com/collectd/pkg-debian/pull/18

The good news is that 5.9.2 collectd upstream updates have now flowed through from Debian to Canonical / Ubuntu and now part of Focal Fossa (20.04 LTS), which will is targeted for release around March of this year.

If you are doing Ubuntu / Debian related updates then I am sure that @bzed would welcome you onboard to support the Debian maintenance work: https://salsa.debian.org/debian/pkg-collectd

I backported his Debian build updates back into my pkg_debian pull request, but I think we all recognise that this tree is now inactive and so best to managed via the debian package.

Finally remember that the libc sysctl deprecation message will vanish in the "fullness of time" and this little area of frustration will vanish ;-)

Cheers,

Zebity

zebity:
@kraj now that https://github.com/collectd/collectd/pull/3342 has been merged which took up with thread, I believe this pull request can be closed. This helps keep repository current.

@qingling128 thanks for providing little push to get updates merged.

cheers,

Zebity

kraj:
sure

PR #3243, opened 2019-08-08
Changelog: processes plugin: fix number of running processes

badevos:
ChangeLog: processes plugin: fix number of running processes 

scanning /proc/*/stat AND computing other process stats takes too much
time. Consequently, the number of running processes based on the occurences
of 'R' as character indicating the running state is typically zero. Due
to processes are actually changing state during the evaluation of it's
stat(s).

The 'procs_running' number in /proc/stat on the other hand is more
accurate, and can be retrieved in one single 'read' call.

In the unlikely case that procs_running can not be retrieved from
/proc/stat, (read interrupted prior to reading up to 'procs_running'
the value '1' assumed, since this plugin is at least being executed.

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -1429,6 +1429,29 @@ static int ps_read_process(long pid, process_entry_t *ps, char *state) {
   return 0;
 } /* int ps_read_process (...) */
 
+static int procs_running(void) {
+  char buffer[4096] = {};
+
+  char *running;
+
+  ssize_t status;
+
+  status = read_file_contents("/proc/stat", buffer, sizeof(buffer) - 1);
+  if (status <= 0) {
+    return -1;
+  }
+
+  running = strstr(buffer, "procs_running");
+  if (!running) {
+    WARNING("procs_running not found, assuming 1");
+    return 1;
+  }
+
+  running += 14;
+
+  return atoi(running);
+}
+
 static char *ps_get_cmdline(long pid, char *name, char *buf, size_t buf_len) {
   char *buf_ptr;
   size_t len;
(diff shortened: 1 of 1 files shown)


faxm0dem:
Thanks for this idea!

I wouldn't return 1 if the file isn't readable. Either retry or return undef

badevos:
Hi,

I updated the code and return '-1' on error. 
I agree, it's up to the higher level application to interprete the data, not collectd.

Let me know what to do next?

Kr,
Bart.

lurkr2:
add string to first comment https://github.com/collectd/collectd/pulls#issue-305459063
`ChangeLog: processes plugin: fix number of running processes`

> Hi,
> 
> I updated the code and return '-1' on error.
> I agree, it's up to the higher level application to interprete the data, not collectd.
> 
> Let me know what to do next?
> 
> Kr,
> Bart.

dago:
@badevos Can you please add a more elaborate comment instead of `/* get procs_running from /proc/stat */` to explain the rationale about this change? Probably the text in the first box of this PR would make a good explanation.

src/processes.c
```diff
@@ -1429,6 +1429,41 @@ static int ps_read_process(long pid, process_entry_t *ps, char *state) {
   return 0;
 } /* int ps_read_process (...) */
 
+static int procs_running(void) {
+  char buffer[4096] = {};
```
volth: 4096 is not enough

On my laptop, `/proc/stat` is 5200 bytes long and "procs_running" substring is behind 4096's byte from the beginning, so `collectd` filled the logs with "procs_running not found"

PR #3246, opened 2019-08-09
zfs_arc plugin: Fix skipping of leading lines

jplitza:
Probably due to a merge gone wrong, the first two lines were skipped twice, skipping four lines in total and thus missing the "hits" and "misses" counts.

ChangeLog: zfs_arc plugin: A bug that caused the first to values to be skipped was fixed.


rubenk:
Yes indeed, thanks for the fix!

PR #3247, opened 2019-08-09
zfs_arc plugin: Read dbuf_size, dnode_size and bonus_size

### LINKED ISSUE #2843: zfs_arc plugin: Reading kstat value "other_size" failed.
*   Version of collectd: 5.8.0
*   Operating system / distribution: Arch Linux

## Expected behavior
ZFS on Linux 0.7 no longer has other_size in /proc/spl/kstat/zfs/arcstats, it got replaced with dbuf_size, dnode_size and bonus_size.
https://github.com/zfsonlinux/zfs/commit/25458cbef9e59ef9ee6a7e729ab2522ed308f88f

zfs_arc plugin should read those when other_size is missing.

## Actual behavior
```
parse_value: Failed to parse string as derive: "data".
zfs_arc plugin: `llist_search` failed for key: 'other_size'.
zfs_arc plugin: Reading kstat value "other_size" failed.
```

## Steps to reproduce
*   Install ZFS on Linux >= 0.7
*   LoadPlugin zfs_arc

jplitza:
Fixes #2843

ChangeLog: zfs_arc plugin: New cache values are now read on Linux

### DIFF
--- a/src/zfs_arc.c
+++ b/src/zfs_arc.c
@@ -268,9 +268,16 @@ static int za_read(void) {
   za_read_gauge(ksp, "mfu_size", "cache_size", "mfu_size");
   za_read_gauge(ksp, "mru_ghost_size", "cache_size", "mru_ghost_size");
   za_read_gauge(ksp, "mru_size", "cache_size", "mru_size");
-  za_read_gauge(ksp, "other_size", "cache_size", "other_size");
   za_read_gauge(ksp, "p", "cache_size", "p");
   za_read_gauge(ksp, "size", "cache_size", "arc");
+
+  /* The "other_size" value was replaced by more specific values in ZFS on Linux
+   * version 0.7.0 (commit 25458cb)
+   */
+  if (za_read_gauge(ksp, "dbuf_size", "cache_size", "dbuf_size") != 0 ||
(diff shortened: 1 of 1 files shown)


src/zfs_arc.c
```diff
@@ -285,10 +285,20 @@ static int za_read(void) {
...
   za_read_gauge(ksp, "mru_ghost_size", "cache_size", "mru_ghost_size");
   za_read_gauge(ksp, "mru_size", "cache_size", "mru_size");
-  za_read_gauge(ksp, "other_size", "cache_size", "other_size");
   za_read_gauge(ksp, "p", "cache_size", "p");
   za_read_gauge(ksp, "size", "cache_size", "arc");
 
+  /* The "other_size" value was replaced by more specific values in ZFS on Linux
+   * version 0.7.0 (commit 25458cb)
+   */
+  static int other_size_avail = 1;
```
rubenk: can you make this a bool?

rubenk: or even better, drop the use of the static variable, and just read other_size if you can't read one of the other 3 counters,

jplitza: I was mimicking the logic from the "l2_size_avail" right below this, because I assumed that to be the desired approach. But sure, here you go.

PR #3249, opened 2019-08-10
Allow oracle collectd queries to run at specific interval

azhark-google:


mrunge:
Thank you for your contribution. It would be much easier to review the changes, if you could reduce them to the required minimum and would leave out unrelated changes.
Also, adding a description or referencing a bug would be helpful.

PR #3252, opened 2019-08-13
Fix a few issues found with LGTM

rubenk:
Changelog: fix a few issues found with LGTM


rubenk:
Thanks for the review. No problem at all, I've been AFK for much longer!

PR #3257, opened 2019-08-19
write_http: Remove curl default http response to stdout

### LINKED ISSUE #3105: Chatty messages (write_http)
*   Operating system / distribution: RHEL 6.7

## Expected behavior

collectd (write_http) should only log errors to stderr (or whatever logging plugin you are using) if the loglevel is set to err

## Actual behavior

With the syslog plugin configured to LogLevel err, I see hundreds of these messages in /var/log/messages

{"text":"Success","code":0}

## Steps to reproduce
1. Install collectd
2. configure write_http (in this case sending data to Splunk)
3. restart collectd, since write_http only seems to log these messages when you restart the service

duckfez:
ChangeLog:  write_http:  Remove curl default HTTP response to stdout

Should fix #3105.  Adds a callback to the calls to curl to write to a response buffer.  The response buffer is only dumped to the collectd logs in DEBUG or when there's an error.  Tested with output to Splunk.

### DIFF
--- a/src/write_http.c
+++ b/src/write_http.c
@@ -90,7 +90,6 @@ struct wh_callback_s {
 
   char *response_buffer;
   size_t response_buffer_size;
-  size_t response_buffer_fill;
 
   int data_ttl;
   char *metrics_prefix;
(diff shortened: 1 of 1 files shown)


duckfez:
Replaced with #3263 which is the same patch cleaned up a little more.

PR #3258, opened 2019-08-21
Fix bug in postgresql plugin (conninfo)

lurkr2:
Error processing connection parameters.
The connection parameters specified in the config were not processed correctly, and therefore were not taken into account.
The snprintf function returns the number of bytes of a string, unlike
ssnprintf which returns bool.Bug in processing connection parameters.
Connection parameters specified in the config were not taken into account

ChangeLog: Postgresql plugin: Connection parameters not taken.

### DIFF
--- a/src/postgresql.c
+++ b/src/postgresql.c
@@ -58,7 +58,7 @@
  * is ignored. */
 #define C_PSQL_PAR_APPEND(buf, buf_len, parameter, value)                      \
   if ((0 < (buf_len)) && (NULL != (value)) && ('\0' != *(value))) {            \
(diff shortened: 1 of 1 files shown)


mrunge:
Thank you for your contribution here. I am quite certain, the issue has already been fixed by https://github.com/collectd/collectd/commit/c80ef3c943ed11f49d8d2c0440823d43ef0d7cd5

5.9.1 has some known issues I hope to fix with upcoming 5.9.2 (to be released asap)

lurkr2:
thanx for work

PR #3263, opened 2019-08-22
write_http no longer writes http response to stdout

### LINKED ISSUE #3105: Chatty messages (write_http)

duckfez:
Should fix #3105.  Replaces PR #3257.

Registers a callback with libcurl to write the HTTP response to a buffer instead of stdout.  Depending on logging settings, this will either be discarded or written to the log.  By design, I'm only keeping the first 4K of the HTTP response, as any additional data is probably not helpful.  Tested with Splunk HEC, and seems to work there - not tested with other HTTP targets.

ChangeLog: write_http: Remove libcurl default HTTP response to stdout


src/write_http.c
```diff
@@ -807,6 +845,16 @@ static int wh_config_node(oconfig_item_t *ci) /* {{{ */
     wh_callback_free(cb);
     return -1;
   }
+
+  cb->response_buffer_size = WRITE_HTTP_RESPONSE_BUFFER_SIZE;
+  cb->response_buffer = malloc(cb->response_buffer_size);
```
kwiatrox: Is there a reason to not make response_buffer as array of given size, its length seems to be fixed anyway? It would simplify things in other places too.

duckfez: I think that would be fine.  It's been a while, but I think I was originally trying for being able to grow the response_buffer dynamically if the response coming back wouldn't fit.  I gave up, realizing that for most use cases the first few hundred bytes of the response are all that they need, and even then only in the case of an error.  

I also thought about trying to allocate a response buffer as big as the Content-Length header, but I don't think that value is necessarily available to me here.

src/write_http.c
```diff
@@ -91,6 +98,24 @@ typedef struct wh_callback_s wh_callback_t;
...
+  int len = 0;
+
+  DEBUG("write_http plugin: curl callback writing %zu bytes of output.", nmemb);
+  if (nmemb > cb->response_buffer_size - 1)
+    len = cb->response_buffer_size - 1;
+  else
+    len = nmemb;
+
+  memcpy(cb->response_buffer, ptr, len);
+  cb->response_buffer[cb->response_buffer_size] = '\0';
```
dago: This produces a memory corruption as it writer the `\0` past the end of the allocated storage. This should probably be
```
cb->response_buffer[len] = '\0';
```

duckfez:
Updated per suggestions made in review.  The response buffer is now an array at 1K, and I've hopefully fixed the out-of-bounds accesses.  My apologies it's taken so long to get this turned around.

mrunge:
Thank you!

PR #3264, opened 2019-08-22
configurable plugin instance in gpu_cuda plugin

rdietric:
The plugin instance can now be configured via collectd.conf. Currently, the GPU name is used. However, this does not allow us to distinguish between GPUs of the same type, e.g. in workstations or computing clusters, which often contain multiple GPUs of the same type. 

Hence, I propose to configure the plugin instance of the gpu_cuda plugin via the two booleans 'InstanceByGPUIndex' and 'InstanceByGPUName'. The GPU index is unique for every NVIDIA GPU in the system. Furthermore, this is more in line with the cpu plugin, which uses the core as plugin instance. I set the default to "GpuId-GpuName".

I also replaced MAX_DEVNAME_LEN with NVML_DEVICE_NAME_BUFFER_SIZE from nvml.h.

ChangeLog: GPU NVML plugin: configurable plugin instance by GPU name and/or GPU index.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -667,6 +667,8 @@
 #   GPUIndex 0
 #   GPUIndex 2
 #   IgnoreSelected false
+#   InstanceByGPUIndex true
+#   InstanceByGPUName true
 #</Plugin>
 
 #<Plugin grpc>
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3280,6 +3280,16 @@ B<GPUIndex> entries are collected. For greater clarity, setting IgnoreSelected
 without any GPUIndex directives will result in B<no> statistics being
 collected.
 
+=item B<InstanceByGPUIndex>
+
+If set to false, the GPU ID will not be part of the plugin instance. The default
+is 'GPU ID'-'GPU name'
+
+=item B<InstanceByGPUName>
+
+If set to false, the GPU name will not be part of the plugin instance. The 
+default is 'GPU ID'-'GPU name'
+
 =back
 
 =head2 Plugin C<grpc>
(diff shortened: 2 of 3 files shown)


rdietric:
The build error does not seem to be related to my changes. I tried to reproduce it on my local machine with clang, but it builds without errors. Should I push a dummy change to trigger CI once more? PR #3273 had a similar issue.

dago:
I see. Question: why are there **_any_** PRs in 5.9 branch not merged into master? Shouldn't we do this as starter for 5.12?

mrunge:
@dago the way the community did this so far is: fix issues in released branches and merge released branches to master. Apparently, this did not happen here...

@kkepka good catch, would you be able to propose patches for these issues? Previously, one could simply merge them, once they went through a review, but that is not the case anymore.

dago:
@mrunge But the development has changed since then? Nowadays PRs are merged to master and releases are snapshotted (or branched for backports of security releases), right?

mrunge:
@dago it is still encouraged to push fixes to released branches. In reality, that does not happen, as folks propose patches for master.

In *my* personaly view, we should keep it that way (patches to master, cherry-pick to released branches), but the last time we spoke about that, I was the only one arguing for that workflow. I believe, we have that "original" workflow documented somewhere, but I currently can not find the correct link right now.

kkepka:
> @kkepka good catch, would you be able to propose patches for these issues? Previously, one could simply merge them, once they went through a review, but that is not the case anymore.

sure, created PR with mentioned fixes #3393

dago:
@kkepka #3393 has been merged, do I get you right that the code still does not compile even with #3393 applied?

kkepka:
That's right @dago. However just checked and I can observe same issue with current master, so it may be something with my test build env or issue is out there in build (and seems not related to this patch from @rdietric)

dago:
@rpv-tomsk Thanks for reviewing the PR and approving the request!

PR #3277, opened 2019-09-16
Use sstrncpy instead of strncpy in sysevent plugin

abays:
ChangeLog: sysevent plugin: use sstrncpy instead of strncpy

### DIFF
--- a/src/sysevent.c
+++ b/src/sysevent.c
@@ -472,7 +472,7 @@ static int read_socket() {
     } else {
       DEBUG("sysevent plugin: writing %s", buffer);
 
-      strncpy(ring.buffer[ring.head], buffer, sizeof(buffer));
+      sstrncpy(ring.buffer[ring.head], buffer, sizeof(buffer));
       ring.timestamp[ring.head] = cdtime();
       ring.head = next;
     }


abays:
Certainly.  It fixes the following compilation error:

```src/sysevent.c: In function ‘read_socket’:
src/sysevent.c:475:53: error: argument to ‘sizeof’ in ‘strncpy’ call is the same expression as the source; did you mean to use the size of the destination? [-Werror=sizeof-pointer-memaccess]
  475 |       strncpy(ring.buffer[ring.head], buffer, sizeof(buffer));```

**Comment by rubenk** [comment]
Why don't you use the length of the destination to limit the amount of bytes copied?

**Comment by abays** [comment]
Using the destination results in this compilation error instead:

```src/sysevent.c: In function 'read_socket':
src/sysevent.c:475:53: error: argument to 'sizeof' in '__builtin_strncpy' call is the same expression as the destination; did you mean to provide an explicit length? [-Werror=sizeof-pointer-memaccess]
       strncpy(ring.buffer[ring.head], buffer, sizeof(ring.buffer[ring.head]));
```

Elsewhere in Collectd I've always used `sstrncpy`.  Not using it here initially was an oversight on my part (and for whatever reason it used to compile without complaint).  I also recall either someone here (or some Collectd documentation) recommending the use of `sstrncpy` over `strncpy` do to security concerns.

rubenk:
In that case we should try to understand the compilation error. Switching to sstrncpy is merely papering over the issue. If you look at its implementation, you'll see that it just ensures nul-termination, but it doens't prevent writing more than can fit into the destination buffer.
```
  char *sstrncpy(char *dest, const char *src, size_t n) {
    strncpy(dest, src, n);
    dest[n - 1] = '\0';

    return dest;
  } /* char *sstrncpy */
```

abays:
In this case, using `sstrncpy` isn't ignoring the issue after all, because all buffers used in the plugin are of the same size, set here:

```static int listen_buffer_size = 4096;```

The call to `recvfrom` prior to this `strncpy` uses this buffer size, and we handle any error received from that call (which would include the buffer size being too small).

Anyhow, another option is just to use `listen_buffer_size` as the third argument to the `strncpy` call.  But I've seen Octo recommending the use of `sstrncpy` over `strncpy` elsewhere, so I'm thinking we should use that here nonetheless.

rubenk:
Mind you, I'm not opposed to the use of sstrncpy, but it's a pity it suppresses the compiler warning which I think is right.

@abays how did you try to pass the destination size?
If you did `strncpy(ring.buffer[ring.head], buffer, sizeof(ring.buffer[ring.head]));` that will indeed trigger a warning, sizeof(ring.buffer[ring.head] is the size of a pointer to char.

mrunge:
@rubenk I'm sorry if I stepped over your toes, that hasn't been my intention here. I think you are right.

PR #3282, opened 2019-09-18
Check for sysevent plugin init calloc failures

abays:
ChangeLog: sysevent: Add a few missing calloc result checks in the sysevent_init function


mrunge:
Ah, I believe the ChangeLog line may only be one single line.

PR #3237, opened 2019-07-29
fix ssnprintf wrapper

faxm0dem:
This should fix #3232 #3235 and #3236 
ChangeLog: Fixes failing plugins after introduction of snprintf wrapper

Issue #3223, opened 2019-07-17
Feedback Requested: Collectd Release Process Proposal


sunkuranganath:
Hello Collectd’ers,

In order to bring together a structure and known cadence to Collectd releases, @mrunge , @tokkee and I have been working together to bring forward to the community a release process proposal.

The idea is to share the burden of releasing, tagging and maintaining releases and establish few guidelines for community ownership.
We are looking for feedback and progress towards arriving at a common ground. Appreciate your help.

Please go through following proposal shared via Google docs link [1]. 

[1] https://docs.google.com/document/d/1Ha9zPyKa_1pBar58LUPRf4EV2ap1sMcrOtRz3qauogo/edit?usp=sharing

rpv-tomsk:
Collectd has no active developers, has no active maintainers, has no publicaly active Owner. face it.
Team did all possible to terminate development, while you  try to do the opposite.  I think, your try will have no success (you simply has no appropriate permissions on GitHub project), don't waste your time, guys.

I think, you will have nothing good to release in nearest, simply because nobody is interested in PR's merging, no matter how useful the changes.  Only destroying changes accepted freely.

sunkuranganath:
Please check https://github.com/collectd/collectd/pull/3224

zebity:
Hi collectd'ers,

agree with core objective of seeing to get "know cadence" going on collectd.

Here are my thoughts and outline of issues which I am seeing on this:

1. Need to define clearly (or simply restate [versioning](https://en.wikipedia.org/wiki/Software_versioning) principles are. To my mind it should be something like (taking 5.9.2 as example)>
Major (ie 5) - reflects change to core collectd driver code
Minor (ie 9) - reflects additions / removals if "significant" (and given passions and effort put into plug-ins they are likely to be all considered significant)
Patch (ie 2) - reflects fixes, plug-in enhancements, beautification and other maintenance activities

2. Need to ensure that CI process defines clear test objectives, currently this process appears to be focussed on testing that things compile, rather than run.
It should be that a bump in Major, Minor needs to be functionally verified across range of distribution (see point 3), while patch release can go through a more limited set of validations

3. Need to define they set of targets that must be covered for Major / Minor release.
My observation is that there are 3 primary platforms for which there is plug-in code for:
Linux variations, BSD, Solaris and Windows. I believe that release type requires clear definition of verification quality steps. Here is possible set:
[...]

PR #3283, opened 2019-09-18
Consistent returns from ssnprintf/__ssnprintf wrappers, add ssnprintf2 for optional error handling.

zebity:
ChangeLog: collectd: Consistent ssnprintf/__ssnprintf returns, new ssnprintf2 for error handling.

@mrunge cherry picked fix for ssnprintf wrapper https://github.com/collectd/collectd/pull/3237 (mea culpa) back into master, thank you @faxm0dem . Applied same update to __ssnprintf wrapper in src/collectdctl.c for return result consistency and ensured opening comments consistent with code.

As per original suggestion from @octo (https://github.com/collectd/collectd/pull/2895) I have also added second wrapper ssnprintf2 into src/utils/common/common.c . This has the same return behaviour of original suggestion (which resulted in problem), so that in future if we want to add error handling for truncation (rather than ignore it) then this can be done on case by case basis.

The degree of impact of adding truncation error handling is to large to do in single hit, as I don't want to risk breaking code base a second time.

Thanks to all for continued updates into collectd.


mrunge:
Thank you, I'll try to get back to this asap.

src/utils/common/common.c
```diff
@@ -89,8 +89,7 @@ char *sstrncpy(char *dest, const char *src, size_t n) {
   return dest;
 } /* char *sstrncpy */
 
-/* ssnprintf returns zero on success, one if truncation occurred
-   and a negative integer onerror. */
+/* ssnprintf returns result from vsnprintf conistent with snprintf */
```
rubenk: so how is this function different then from snprintf?

src/utils/common/common.c
```diff
@@ -99,11 +98,24 @@ int ssnprintf(char *str, size_t sz, const char *format, ...) {
 
   va_end(ap);
 
+  return ret;
+} /* int ssnprintf */
+
+/* ssnprintf2 returns zero on success, one if truncation occurred
+   and a negative integer on error. */
+int ssnprintf2(char *str, size_t sz, const char *format, ...) {
```
rubenk: It seems that this does exactly the same as snprintf and ssnprintf, or am I missing something?

mrunge: The return value is different, and I assume the idea is to provide both.

rubenk: you can detect truncation with snprintf just fine, so I don't see how this helps.

rubenk:
I'm sorry for perhaps sounding so negative, but this has all turned into a bit of a mess.
The original issue as I see it is two-fold:
* gcc started warning for potential truncation.
* we turn all warnings into errors, which breaks peoples builds.

Some of those truncation issues are real and we should fix them, some of them are not. Adding wrappers around them will never get them fixed. Admittingly checking every call site is a lot of work.

What I'd rather see is that we stop building with -Werror by default (most distros already turn that of). We could then turn -Werror on in CI, but turn the truncation errors into warnings (-Wno-error=stringop-truncation or something like that). We'll still see the warnings and can fix them in due time.

How does that sound?

mrunge:
I agree, this sounds like a good plan to go forward.

So the next step(s) would be to revert the ssnprintf changes and to fix potential truncation.

Ideally, we'd remove ssnprintf and at the same time, fixing truncation warnings. I'd prefer to keep warnings as errors.

src/utils/common/common.c
zebity: Hi @rubenk the ssnprintf wrapper was introduced to get around GCC 8.2/.3 warning, only, there is no functional difference.
The original suggested fix included changing return to allow handling of truncation, however as this broke stuff.
So I to allow for future handling of truncation, case I added ssnprintf2 case.
Also to ensure that the other case that had the alternated return case (in collectdctl.c), I updated this to be consistent with snprintf.
As per discussion below, this was all driven by compiler behaviour than feature/fix need and yes the amount of code touched was greater than desired.
Cheers.

zebity:
@ubenk & @mrunge,

sounds like a plan.

Wrapper was put in place due to number of places that -Werror broke build.

Remember that if you keep -Werror on in CI then all the builds will fail :-(

The amount of work required to get rid of warnings (assuming change the compile flags) is around approximately 67 files that where updated originally. It was the large number of files being changed that blocked getting this done.

I suspect that most if these files are dormant in terms of change so getting them fixed on an "as needs" basis will be slow process.

Cheers,

zebity

zebity:
@mrunge @rubenk @rpv-tomsk 

a quick find / grep indicates that there are 67 files with ssnprintf wrapper and 19 of these actually appear to use the return value.

So if any rework is going to be done then it would make sense to start with ones that use the return value, as these are potentially sensitive to truncation.

On different note, the number of plugins (there are around 60 libraries required to build for Debian package) and potential rate of change across these means that the release rate will have turn up substantially to allow collectd to keep up with changes forced to keep plugins current.

I think this is reflected by the very number of forked repositories.

Cheers,

Zebity.

mrunge:
Thank you @zebity . What do you mean by "release rate"? Release rates for collectd? You may have seen https://github.com/collectd/collectd/pull/3318 ?

zebity:
@mrunge yes to #3318  and I commend and thank you for your hard graft in keeping the wheels turning.

I was just doing some "out loud thinking...". I have been looking at the GitHub statistics and it would be really helpful to have some sense of rate of change across the entire code base (ie change with core vs change within pluggins and bug fix / currency driven change vs feature change.
I dwell on this due to my dear old debian packaging still languishing at 8.x ... :-(

Also I am a bit stunned by the 1000 or so forks... are these all mostly dormant or is there a whole lot of work happening outside of main branch ?

As I said I so just "thinking out load".

zebity:
@rjablonx @mrunge I ask if you could please review and merge this request.

I have now wound out anything that is pre-emptive of immediate need (ie ssnprintf2 proposal), so change now just ensures that collectdctl.c is aligned with @faxm0dem fix (https://github.com/collectd/collectd/pull/3237).

I will not do any further changes related to the ssnprintf wrappers via this request.
This is In line with need to keep regular cadence as per discussion thread https://github.com/collectd/collectd/issues/3223
I believe that any further work on add/remove/tweeking build flags should be tackled and managed the basis of individual merge requests and not a bulk change.

I think that is is consistent with approach of letting primary owner / maintainers of the various 67 files updated to be able to be involved in any further remediation.

For the initial identified set of 29, I believe the best approach would be equivalent 29 individual pull requests, which can run there own course and with their own priority.
While this might seen onerous in terms of impact to release process, if we get greater level of automation going then it should allow ongoing smoother regular patch level releases, starting from new 5.10.0 release.

I will get good to get this initial bulk change (and associated formatting clean up) done and dusted.

Cheers,

Zebity

zebity:
@mrunge many thanks.

FYI, I have updated test server to Ubuntu 19.10 so next step will be to test current 5.9.2 build against that and ensure all is ok on this update.

Cheers,

Zebity

PR #3292, opened 2019-10-02
New Capabilities plugin

kwiatrox:
ChangeLog: capabilities plugin: New plugin to read static platform data

Read decoded SMBIOS data using dmidecode and expose it through httpd.
Uses libmicrohttpd and libjansson.
It is a new approach to abandoned util #2000 which was using dmidecode to read BIOS information.

Signed-off-by: Kamil Wiatrowski <kamilx.wiatrowski@intel.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -750,6 +750,32 @@ bind_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 bind_la_LIBADD = $(BUILD_WITH_LIBCURL_LIBS) $(BUILD_WITH_LIBXML2_LIBS)
 endif
 
+if BUILD_PLUGIN_CAPABILITIES
+pkglib_LTLIBRARIES += capabilities.la
+capabilities_la_SOURCES = src/capabilities.c \
+                          src/utils/dmi/dmi.c \
+                          src/utils/dmi/dmi.h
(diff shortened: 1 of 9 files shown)


rubenk:
How is this approach different from #2000?

kwiatrox:
> How is this approach different from #2000?

Most important differences:
Old solution #2000: added only BIOS data to global variables, was part of core collectd.
New plugin: separate plugin (thus can be enabled or disabled), read wider range of data, format data as JSON, expose data for users using libmicrohttpd.

paweltox:
Why did you implement exposing data to user with libmicrohttpd instead of using existing write_http plugin?

kwiatrox:
> Why did you implement exposing data to user with libmicrohttpd instead of using existing write_http plugin?

write_http is one of the collectd write plugins that works in typical collectd data model, it sent metrics and notifications using http.
The capabilities plugin does not collect any dynamic data into metrics or notifications, but only exposes static data that does not change during operation.
It provides micro http webserver and works in pull model. The similar pull model is for example used in write_prometheus plugin.

JoshHilliker:
This would be extremely helpful.   It would be great to get this out as soon as possible.

dago:
This is further tracked in PR #3384

dago:
Closing as #3384 has been merged.

Issue #3289, opened 2019-09-30
[Regression] Plugin exec no longer reads from all configured entries


iain-buclaw-sociomantic:
*   Version of collectd: 5.9.1
*   Operating system / distribution: Linux
*   Kernel version (if applicable):

## Expected behavior

All configured Plugin exec entries to have their values read and sent.

## Actual behavior

Only a small random number of exec entries work.  Downgrading to 5.8 fixes the problem.

## Steps to reproduce
```
LoadPlugin exec

<Plugin exec>
	Exec "user" "/path/to/script" "--instance" "A"
	Exec "user" "/path/to/script" "--instance" "B"
	Exec "user" "/path/to/script" "--instance" "C"
	Exec "user" "/path/to/script" "--instance" "D"
	Exec "user" "/path/to/script" "--instance" "E"
</Plugin>
```

Restarting or Stop/Starting collectd will cause some entries to have metrics collected, others not, but never all to be collected, despite all entries showing up in `ps faux`.

mrunge:
In general, 5.9.1 should not be used.

iain-buclaw-sociomantic:
I haven't tried 5.9.0 yet, I skipped that version, though I can trigger a build and test it.

mrunge:
I was hoping to get to tag 5.9.2 today.

iain-buclaw-sociomantic:
Just tried 5.9.0, and it looks like the same problem, so some sort of bisect between 5.8.1 and 5.9.0 is required.

mrunge:
IMHO we should be clear with https://github.com/collectd/collectd/pull/3290

mrunge:
could you please give 5.9.2 (on the releases page) a try? Ideally that'd get picked up from distributions as well.

iain-buclaw-sociomantic:
Built and running on one server.  I'll leave it till tomorrow to test other configurations.

iain-buclaw-sociomantic:
Nope, still happens, although I've gotten further in finding out what the precise problem is.

Metrics are being collected and sent over the network, however the `COLLECTD_HOSTNAME` is either getting lost, or is not exported properly before forking.  I've found a new metric called `localhost.xxx.yyy` - which is the default hostname used if none is provided to the exec script.

Working version is 5.8.0, I've noticed that 5.8.1 exists, and I can reproduce the problem there as well, so hopefully that means there's very little between the two that could be the problem.

iain-buclaw-sociomantic:
[...]

PR #3295, opened 2019-10-04
Issue #3289: Add execvpe() wrapper, call it instead of setenv/unsetenv.

iain-buclaw-sociomantic:
ChangeLog: Exec plugin: Fix race condition in fork_child, where environment variables got unset before calling the exec program. #3289

execvpe() is a GNU extention, and not widely supported across other platforms, so the implementation has been made internal for the benefit of the exec plugin.

Because fork_child is called from multiple threads concurrently, it is not safe to use setenv()/unsetenv() to manipulate global shared state.
 
The bug witnessed is code being executed in the following order:

- Thread 0: setenv("COLLECTD_HOSTNAME")
- Thread 0: fork()
- Thread 1: setenv("COLLECTD_HOSTNAME")
- Thread 0: unsetenv("COLLECTD_HOSTNAME")
- Thread 1: fork()
- Thread 1: execvp(file, argv)

Leaving the program executed in thread 1 without COLLECTD_HOSTNAME and COLLECTD_INTERVAL being set.

Fixes: #3289

### DIFF
--- a/src/exec.c
+++ b/src/exec.c
@@ -245,47 +245,6 @@ static int exec_config(oconfig_item_t *ci) /* {{{ */
   return 0;
 } /* int exec_config }}} */
 
-#if !defined(HAVE_SETENV)
-static char env_interval[64];
-// max hostname len is 255, so this should be enough
-static char env_hostname[300];
-#endif
-
-static void set_environment(void) /* {{{ */
-{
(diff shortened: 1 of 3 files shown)


mkauf:
Please compare / merge with this PR: #3138

mrunge:
the other PR has been merged.

PR #3302, opened 2019-10-12
Add new plugin for Nagios Livestatus internal statistics

riton:
Thie P.R proposes a new plugin for collecting Nagios / Naemon Livestatus internal statistics.

I've added a basic documentation in the _pod_ file and the sample `collectd.conf` file.
I may have missed a few places where I should have added stuff.

Regards


riton:
A bug has been found in production with the current version of this plugin.
We're currently working on a fix.

riton:
The parsing bug has (apparently) been fixed and we've updated the data types to match the ones used by the Livestatus code.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+             "requests_rate");
+  L_DISPATCH(ls_collectd_dispatch_counter, status->service_checks,
+             "service_checks");
+  L_DISPATCH(ls_collectd_dispatch_count, status->service_checks_rate,
+             "service_checks_rate");
+
+  return rc;
+} /* int ls_collectd_dispatch */
+
+static int ls_read(void) /* {{{ */
```
rpv-tomsk: Does that socket allows persistent connections?

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+  L_DISPATCH(ls_collectd_dispatch_counter, status->host_checks, "host_checks");
+  L_DISPATCH(ls_collectd_dispatch_count, status->host_checks_rate,
+             "host_checks_rate");
+  L_DISPATCH(ls_collectd_dispatch_counter, status->livecheck_overflows,
+             "livecheck_overflows");
+  L_DISPATCH(ls_collectd_dispatch_count, status->livecheck_overflows_rate,
+             "livecheck_overflows_rate");
+  L_DISPATCH(ls_collectd_dispatch_counter, status->livechecks, "livechecks");
+  L_DISPATCH(ls_collectd_dispatch_count, status->livechecks_rate,
+             "livechecks_rate");
```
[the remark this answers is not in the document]
riton: Indeed, this seems like a good idea to drop the `_rate` metrics.
On the other hand, I do not understand why `DERIVE` would be more suitable than `COUNTER` here. Can you explain ?

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+
+  return vl;
+} /* value_list_t ls_collectd_init_vl */
+
+static int ls_collectd_dispatch_count(gauge_t value,
+                                      const char *plugin_instance) {
+  value_list_t vl = ls_collectd_init_vl();
+
+  sstrncpy(vl.type, "count", sizeof(vl.type));
+  sstrncpy(vl.plugin_instance, (char *)plugin_instance,
```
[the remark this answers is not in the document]
riton: Ok, see what you mean and seems like a good idea.

src/livestatus.c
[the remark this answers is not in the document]
rpv-tomsk: Use of `DERIVE` will calculate and provide 'rate' metric, based on collectd read interval.
So, that would be current rate, not 'average rate since launch' or 'rate for unknown period [defined in nagios daemon]'.

src/livestatus.c
riton: [Yes it does](https://checkmk.com/cms_livestatus.html#Keeping%20a%20connection%20alive%20(KeepAlive)) but I didn't want to dedicate a connection in the Livestatus backend for the collectd monitoring.
A more flexible solution may be to allow the end-user to configure this and decide if he wants collectd to use a persistent connection.
Would it be better for you ?

src/livestatus.c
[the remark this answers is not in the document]
faxm0dem: collectd will also compute the rate for `COUNTER`. I think the latter is more adapted to the case at hand, as it has the advantage of not missing a value during counter reset

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+
+  for (int j = 0; j < LIVESTATUS_EXPECTED_FIELDS_RESP_NB; j++) {
+    fields[j] = malloc(sizeof(char *) * 32);
+    if (fields[j] == NULL) {
+      ERROR("livestatus plugin: malloc failed: %s", STRERRNO);
+      rc = -1;
+      goto free_all_fields;
+    }
+
+    memset(fields[j], 0x0, 32);
```
kwiatrox: You can use calloc instead of malloc and memset will not be needed.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+  }
+
+  pchar = (char *)lresponse;
+  char *wstart = (char *)lresponse;
+  int exit = 0;
+  while (!exit) {
+    if (*pchar == '\0' || *pchar == '\n') {
+      exit = 1;
+    } else {
```
kwiatrox: 'else if (*pchar != ';')' to avoid one unnecessary level of nesting

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+      if (*pchar != ';') {
+        pchar++;
+        continue;
+      }
+    }
+
+    strncpy(fields[i], wstart, pchar - wstart);
+    wstart = pchar + 1;
+    i++;
+    pchar++;
```
kwiatrox: You can do it in one line: wstart = ++pchar;

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+  fields = NULL;
+
+  return rc;
+} /* int ls_parse */
+
+static int ls_read_parse(const int sockfd, livestatus_status_t *lstatus) {
+  ssize_t bread = -1;
+  char buffer[IO_BUFFER_SIZE];
+
+again:
```
kwiatrox: goto is sometimes used for error handling, but here it is more typical use for a while loop.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+  memset(&buffer, 0x0, sizeof(buffer));
```
kwiatrox: You probably mean memset(buffer, 0x0, sizeof(buffer)); (without &)

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+  buffer[IO_BUFFER_SIZE - 1] = '\0';
+
+  return ls_parse((const char *)&buffer, lstatus);
+} /* int ls_read_parse */
+
+static int ls_send_request(const int sockfd) {
+  int written = -1;
+  char request[512];
+
+  memset(&request, 0x0, sizeof(request));
```
kwiatrox: You probably mean 'request' without &.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+
+  snprintf(request, sizeof(request) - 1, "GET status\nColumns: %s\n\n",
+           LIVESTATUS_QUERY_COLUMNS);
+
+again:
```
kwiatrox: goto is sometimes used for error handling, but here it is more typical use for a while loop.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+} /* int ls_parse */
+
+static int ls_read_parse(const int sockfd, livestatus_status_t *lstatus) {
+  ssize_t bread = -1;
+  char buffer[IO_BUFFER_SIZE];
+
+again:
+  memset(&buffer, 0x0, sizeof(buffer));
+
+  bread = read((int)sockfd, &buffer, IO_BUFFER_SIZE - 1);
```
kwiatrox: buffer without &

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+    if (errno != EINTR) {
+      ERROR("livestatus plugin: reading from socket: %s", STRERRNO);
+      return -1;
+    }
+    goto again;
+  }
+
+  buffer[IO_BUFFER_SIZE - 1] = '\0';
+
+  return ls_parse((const char *)&buffer, lstatus);
```
kwiatrox: buffer without &

dago:
@riton We are currently working on the 5.11.0 release and I noticed there are still a lot of open code discussions. Would you mind having a look so that this plugin may go into the release?

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+  }
+
+  if (strcasecmp("OnFailureMaxRetry", key) == 0) {
+    char *max_retry_str = strdup(value);
+    if (max_retry_str == NULL) {
+      ERROR("livestatus plugin: strdup failed: %s", STRERRNO);
+      return 1;
+    }
+
+    livestatus_obj.max_retry = atoi(value);
```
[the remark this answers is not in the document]
riton: I've pushed a new version that relies on `strtol` to handle user input. Is this new solution okay for you ?

src/livestatus.c
riton: Modified as requested

src/livestatus.c
riton: Fixed.

src/livestatus.c
riton: Fixed.

src/livestatus.c
riton: Fixed.

src/livestatus.c
riton: Fixed.

src/livestatus.c
riton: Fixed.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+  }
+
+  if (strcasecmp("OnFailureBackOffSeconds", key) == 0) {
+    char *backoff_str = strdup(value);
+    if (backoff_str == NULL) {
+      ERROR("livestatus plugin: strdup failed: %s", STRERRNO);
+      return 1;
+    }
+
+    livestatus_obj.backoff_sec = atoi(value);
```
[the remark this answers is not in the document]
riton: I've pushed a new version that relies on strtol to handle user input. Is this new solution okay for you ?

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+#include "collectd.h"
+#include "plugin.h"
+#include "utils/common/common.h"
+
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <sys/un.h>
+
+#define IO_BUFFER_SIZE 4096
+#define LIVESTATUS_FIELD_SEP ';'
```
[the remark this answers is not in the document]
riton: Fixed.

riton:
Hi @dago , sorry for the delay. 

@kwiatrox thank you very much for the review, much appreciated.

Makefile.am
```diff
@@ -1174,6 +1174,20 @@ java_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(JAVA_LDFLAGS)
 java_la_LIBADD = $(JAVA_LIBS)
 endif
 
+if BUILD_PLUGIN_LIVESTATUS
+pkglib_LTLIBRARIES += livestatus.la
+livestatus_la_SOURCES = src/livestatus.c
+livestatus_la_CFLAGS = $(AM_CFLAGS)
+livestatus_la_LDFLAGS = $(PLUGIN_LDFLAGS)
+livestatus_la_LIBADD =
```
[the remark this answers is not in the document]
riton: Fixed.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+  return 0;
+} /* int ls_init */
+
+static int ls_config(const char *key, const char *value) /* {{{ */
+{
+  if (strcasecmp("LivestatusSocketFile", key) == 0) {
+    char *socket_file = strdup(value);
+    if (socket_file == NULL) {
+      ERROR("livestatus plugin: strdup failed: %s", STRERRNO);
+      return 1;
```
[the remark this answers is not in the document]
riton: Fixed.

src/livestatus.c
riton: I've migrated to a `do{...}while(errno == EINTR)` pattern. Is this okay for you ?

src/livestatus.c
riton: I've migrated to a `do{...}while(errno == EINTR)` pattern. Is this okay for you ?

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+    }
+
+    livestatus_obj.socket_file = socket_file;
+  }
+
+  if (strcasecmp("OnFailureMaxRetry", key) == 0) {
+    char *max_retry_str = strdup(value);
```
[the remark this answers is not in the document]
riton: Fixed.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+    if (max_retry_str == NULL) {
+      ERROR("livestatus plugin: strdup failed: %s", STRERRNO);
+      return 1;
+    }
+
+    livestatus_obj.max_retry = atoi(value);
+  }
+
+  if (strcasecmp("OnFailureBackOffSeconds", key) == 0) {
+    char *backoff_str = strdup(value);
```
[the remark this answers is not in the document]
riton: Fixed.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+  if (sfd < 0) {
+    ERROR("livestatus plugin: creating client socket: %s", STRERRNO);
+    return -1;
+  }
+
+  memset(&sun, 0x0, sizeof(sun));
+  sun.sun_family = AF_UNIX;
+  strcpy(sun.sun_path, sockfile);
+
+again:
```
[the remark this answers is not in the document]
riton: I've migrated to a `do{...}while(errno == EINTR)` pattern. Is this okay for you ?

src/livestatus.c
riton: Fixed.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+      .service_checks_rate = strtod(fields[18], NULL),
+  };
+
+  return status;
+} /* livestatus_status_t c_to_livestatus_status */
+
+static int ls_init(void) {
+  livestatus_obj.socket_file = "/var/cache/naemon/live";
+  livestatus_obj.max_retry = 20;
+  livestatus_obj.backoff_sec = 1;
```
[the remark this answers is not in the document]
riton: Damn, okay, wasn't expecting this.
What do you suggest, should I migrate all the _default values  initialization_ to the _config callback_ ?

src/livestatus.c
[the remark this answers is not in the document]
octo: The difference between `COUNTER` and `DERIVE` is that `COUNTER` tries to detect an overflow. If, in fact, the counter was reset (application got restarted, switch port went down/up), this "overflow" can result in huge and incorrect spikes. A `DERIVE` (with a minimum value of zero) will interpret a reset as a *negative* rate and dismiss the value.

With 64bit counters, overflow are very very rare and resets several orders of magnitude more likely. Hence, we recommend to use `COUNTER` only if you know exactly what you are doing. An example would be reading a 32 bit octet counter from a 1 GBit/s interface every 30 seconds: the 32 bit counters may overflow every ~34 seconds under full load.

src/livestatus.c
octo: If reconnecting works for the backend, I'd say that's fine and would prefer to keep it simple. If establishing many new connections causes trouble, we can add such an option later.

src/livestatus.c
```diff
@@ -0,0 +1,463 @@
...
+static int ls_config(const char *key, const char *value) /* {{{ */
+{
+  if (strcasecmp("LivestatusSocketFile", key) == 0) {
+    char *socket_file = strdup(value);
+    if (socket_file == NULL) {
+      ERROR("livestatus plugin: strdup failed: %s", STRERRNO);
+      return 1;
+    }
+
+    livestatus_obj.socket_file = socket_file;
```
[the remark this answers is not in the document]
riton: Nice catch. How would you like me to handle this ? When should I `free` the value ?

I've looked at the _zookeeper_ plugin and `sfree()` is called in the _config callback_.
Such approach does not free the value if the plugin is unregistered, am I right ?
Thanks in advance for your suggestions.

riton:
Okay, I've addressed multiple problems raised by @kwiatrox . Hope my modifications will be okay.

_Note_: Is this me or the `./contrib/format.sh src/livestatus.c` produces weirdly formated code ?
A lot of functions definitions appears to be _indented_.

src/livestatus.c
```diff
@@ -0,0 +1,479 @@
...
+    int sfd = -1;
+
+    sfd = socket(AF_UNIX, SOCK_STREAM, 0);
+    if (sfd < 0) {
+      ERROR("livestatus plugin: creating client socket: %s", STRERRNO);
+      return -1;
+    }
+
+    memset(&sun, 0x0, sizeof(sun));
+    sun.sun_family = AF_UNIX;
```
octo, line 203: Define `sun` here and initialize it:

```c
struct sockaddr_un sun = {
  .sun_family = AF_UNIT,
  // other fields are automatically zero initialized.
};
```

https://github.com/collectd/collectd/blob/master/docs/review_comments.md#initialize-variables
https://github.com/collectd/collectd/blob/master/docs/review_comments.md#define-variables-on-first-use

src/livestatus.c
```diff
@@ -0,0 +1,479 @@
...
+    strcpy(sun.sun_path, sockfile);
```
octo, line 204: Please use `sstrncpy()` instead, toavoid buffer overflows.

src/livestatus.c
```diff
@@ -0,0 +1,479 @@
...
+    *sockfd = sfd;
+
+    return 0;
+  } /* unix_connect */
+
+  static int ls_parse(const char *lresponse, livestatus_status_t *lstatus) {
+    int rc = -1;
+    int i = 0;
+    char **fields = NULL;
+    char *pchar = NULL;
```
octo, line 225: Please define variables on first use, here and below.

https://github.com/collectd/collectd/blob/master/docs/review_comments.md#define-variables-on-first-use

src/livestatus.c
```diff
@@ -0,0 +1,479 @@
...
+    buffer[IO_BUFFER_SIZE - 1] = '\0';
+
+    return ls_parse((const char *)buffer, lstatus);
+  } /* int ls_read_parse */
+
+  static int ls_send_request(const int sockfd) {
+    int written = -1;
+    char request[512];
+
+    memset(request, 0x0, sizeof(request));
```
octo, line 313: No need to use *memset* here:

```c
char request[512] = "";
```

src/livestatus.c
```diff
@@ -0,0 +1,479 @@
...
+
+    snprintf(request, sizeof(request) - 1, "GET status\nColumns: %s\n\n",
```
octo, line 315: Please use `ssnprintf` (note the extra "s") instead. *sprintf(3)* may not null-terminate the result. This is especially problematic when using the `%s` format.

src/livestatus.c
```diff
@@ -0,0 +1,479 @@
...
+
+    return plugin_dispatch_values(&vl);
+  } /* int ls_collectd_dispatch_count */
+
+  static int ls_collectd_dispatch_counter(counter_t value,
+                                          const char *plugin_instance) {
+    value_list_t vl = ls_collectd_init_vl();
+
+    sstrncpy(vl.type, "counter", sizeof(vl.type));
+    sstrncpy(vl.plugin_instance, (char *)plugin_instance,
```
octo, line 361: The cast to `char *` is not necessary – the second argument to `sstrncpy` is a `const char *`.

src/livestatus.c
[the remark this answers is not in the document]
octo: Initialize the static variable:

```c
static livestatus_t livestatus_obj = {
  .socket_file = "/var/cache/naemon/live",
  .max_retry = 20,
  .backoff_sec = 1,
};
```

src/livestatus.c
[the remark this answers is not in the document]
octo: Note that, if `.socket_file` is initialized to a string literal, freeing the literal will segfault. What we typically do instead is initialize the field to `NULL` and, on use, do something like:

```c
(obj.socket_file != NULL) ? obj.socket_file : DEFAULT_SOCKET_FILE
```

src/livestatus.c
```diff
@@ -0,0 +1,479 @@
...
+    }
+  }
+
+  if (strcasecmp("OnFailureBackOffSeconds", key) == 0) {
+    livestatus_obj.backoff_sec = ls_strtoi(value, NULL, 10);
+    if (errno != 0) {
+      ERROR("livestatus plugin: strtoi failed for OnFailureBackOffSeconds: %s",
+            STRERRNO);
+      return -1;
+    }
```
octo, line 186: A closing curly brace is missing here. This is causing the formatting issues. It will also lead to this code not compiling …

octo:
> _Note_: Is this me or the `./contrib/format.sh src/livestatus.c` produces weirdly formated code ?
> A lot of functions definitions appears to be _indented_.

That's a missing curly. I've left a comment in the code where it is missing.

src/livestatus.c
riton: Hi @octo . I do not have any problem using `ssnprintf()` here.
But I'm wondering if there is a real advantage over the `snprintf()` function I'm using here ?
A quick look at the `ssnprintf()` function shows me that this is a simple wrapper on top of `vsnprintf()`. What would be the benefit ?

src/livestatus.c
riton: Fixed.

src/livestatus.c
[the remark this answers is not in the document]
riton: Way better ! Thanks.

src/livestatus.c
riton: Fixed.

src/livestatus.c
riton: I've recreated my dev environment, and obviously something was wrong here :) Thanks.

src/livestatus.c
riton: Should be fixed now.

src/livestatus.c
riton: Fixed.

src/livestatus.c
riton: Should be fixed now.

src/livestatus.c
octo: We introduced `ssnprintf` and `sstrncpy` to ensure that the result is null terminated. Later, we realized that `snprintf` already does null termination. That's probably why the function doesn't do anything different from vanilla `snprintf` and should be removed.

So with that in mind, never mind, feel free to use `snprintf` directly. Sorry for the noise ;)

src/livestatus.c
```diff
@@ -0,0 +1,468 @@
+/**
+ * collectd - src/livestatus.c
+ *
+ * Copyright (C) 2019 IN2P3 Computing Centre, IN2P3, CNRS
```
octo, line 4: At your option: consider bumping the date.

src/livestatus.c
```diff
@@ -0,0 +1,468 @@
...
+
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <sys/un.h>
+
+#define IO_BUFFER_SIZE 4096
+#define LIVESTATUS_FIELD_SEP ';'
+#define LIVESTATUS_PLUGIN_NAME "livestatus"
+
+typedef uint64_t lcounter_t;
```
octo, line 40: What the benefit of defining a new type over using `derive_t`?

(Note: `derive_t` is a signed `int64_t`, so the below `strtoull` calls would need to become `strtoll` instead.)

src/livestatus.c
```diff
@@ -0,0 +1,468 @@
...
+  return 0;
+} /* unix_connect */
+
+static int ls_parse(const char *lresponse, livestatus_status_t *lstatus) {
+  int rc = -1;
+  int i = 0;
+  char **fields = NULL;
+  char *pchar = NULL;
+
+  fields = malloc(sizeof(char **) * LIVESTATUS_EXPECTED_FIELDS_RESP_NB);
```
octo: I the readability of this function can be greatly improved using `strarray_add` and `strarray_free` from `"common.h"`:

```c
char **fields = NULL;
size_t fields_num = 0;
// …
while (read fields) {
  char *value = strtok(); // see below
  strarray_add(&fields, &fields_num, value);
}
// …
strarray_free(fields, fields_num);
return rc;
```

src/livestatus.c
```diff
@@ -0,0 +1,468 @@
...
+  if (fields == NULL) {
+    ERROR("livestatus plugin: malloc failed: %s", STRERRNO);
+    return -1;
+  }
+
+  for (int j = 0; j < LIVESTATUS_EXPECTED_FIELDS_RESP_NB; j++) {
+    fields[j] = calloc(1, sizeof(char *) * 32);
```
octo, line 230: **Obsolete, see above**

This allocation is off in a number of ways.

*Type mismatch:* the `sizeof(char *)` gives the impression that a `char **` is allocated, but `fields[j]` is a `char *`. So instead of 32 `char *` this seems to allocate 256 `char` (assuming 8 byte pointers). Was this meant to allocate a string with 31 characters (+ null byte)?

*Magic literal:* Where is the `32` literal coming from? Please add a comment to document the reasoning.

*Unusual use of `calloc`:* The first argument to `calloc` is the number of elements, doing the multiplication "manually" seems off.

src/livestatus.c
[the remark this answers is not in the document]
riton: Hi @octo . I've hooked into the `shutdown` callback to free the memory (very much inspired from the `chrony` plugin).
Question: Since _collectd_ can dynamically unregister plugins, maybe that relying on the `shutdown` callback is not enough. What's your advice ?

riton:
@octo @kwiatrox Thanks for your review and advices.

I've pushed some modifications and fixed the conversations that I knew how to fix. I've left opened the ones where I still have questions.

Regards

src/livestatus.c
```diff
@@ -0,0 +1,468 @@
...
+    }
+  }
+
+  pchar = (char *)lresponse;
+  char *wstart = (char *)lresponse;
+  int exit = 0;
+  while (!exit) {
+    if (*pchar == '\0' || *pchar == '\n') {
+      exit = 1;
+    } else if (*pchar != LIVESTATUS_FIELD_SEP) {
```
octo, line 245: This re-implements string tokenization, which is error prone and unnecessarily verbose. Please use *strtok_r(3)* instead.

```c
char *res = strdup(lresponse); // skip, if "lresponse" doesn't really need to be "const".
char *ptr = res;
char *saveptr = NULL;
while ((char *f = strtok_r(ptr, LIVESTATUS_FIELD_SEP, &saveptr)) != NULL) {
  ptr = NULL;
  strarray_add(&fields, &fields_num, f);
}
// use "fields"
strarray_free(fields, fields_num);
sfree(res);
```

src/livestatus.c
```diff
@@ -0,0 +1,468 @@
...
+};
+typedef struct livestatus_status_s livestatus_status_t;
+
+static livestatus_t livestatus_obj = {
+    .socket_file = "/var/cache/naemon/live",
+    .max_retry = 20,
+    .backoff_sec = 1,
+};
+
+livestatus_status_t c_to_livestatus_status(const char **fields) {
```
octo, line 107: Please pass a `size_t fields_num` and verify that the array is large enough here. Keep in mind that we're also trying to be defensive in face of future changes. In other words, we should add this check even if the below code (currently) guarantees that there are at least 19 fields.

src/livestatus.c
```diff
@@ -0,0 +1,483 @@
...
+  }
+
+  *sockfd = sfd;
+
+  return 0;
+} /* unix_connect */
+
+static int ls_parse(const char *lresponse, livestatus_status_t *lstatus) {
+
+  char **fields = malloc(sizeof(char **) * LIVESTATUS_EXPECTED_FIELDS_RESP_NB);
```
octo, line 222: I the readability of this function can be greatly improved using `strarray_add` and `strarray_free` from `"common.h"`:

```c
char **fields = NULL;
size_t fields_num = 0;
// …
while (read fields) {
  char *value = strtok(); // see below
  strarray_add(&fields, &fields_num, value);
}
// …
strarray_free(fields, fields_num);
return rc;
```

src/livestatus.c
```diff
@@ -0,0 +1,483 @@
...
+  }
+  if (y < min) {
+    errno = ERANGE;
+    return min;
+  }
+  return y;
+} /* static long ls_strtol_subrange */
+
+// Borrowed from https://stackoverflow.com/a/29378380
+static int ls_strtoi(const char *s, char **endptr, int base) {
```
octo, line 149: Why is this restricting us to the `int` (usually 32 bit, but that's not guaranteed) range? The downstream code is using 64 bit counters consistently, so why reject counter values despite being able to handle them?

src/livestatus.c
```diff
@@ -0,0 +1,483 @@
...
+} /* static int ls_config */
+
+static int unix_connect(const char *sockfile, int *sockfd) {
+  struct sockaddr_un sun = {
+      .sun_family = AF_UNIX,
+  };
+  int rc = -1;
+  int sfd = -1;
+
+  sfd = socket(AF_UNIX, SOCK_STREAM, 0);
```
octo, line 197: `int sfd = socket(…`

Here and below.

https://github.com/collectd/collectd/blob/master/docs/review_comments.md#define-variables-on-first-use

src/livestatus.c
octo: *Obsolete:* You pushed new commits while I was reviewing. I've copied this comment and created a new comment attached to the new location.

PR #3303, opened 2019-10-12
buddyinfo plugin: new plugin for memory fragmentation info

asafkahlon:
ChangeLog: buddyinfo plugin: New plugin for memory fragmentation info.

Signed-off-by: Asaf Kahlon <asafka7@gmail.com>

### DIFF
--- a/src/buddyinfo.c
+++ b/src/buddyinfo.c
@@ -46,7 +46,6 @@ static int buddyinfo_config(const char *key, const char *value) {
     ignorelist = ignorelist_create(1);
     if (ignorelist == NULL) {
       ERROR("buddyinfo plugin: ignorelist_create failed");
-      sfree(ignorelist);
       return -ENOMEM;
     }
   }
--- a/src/buddyinfo.c
+++ b/src/buddyinfo.c
@@ -41,13 +41,23 @@ static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
 static ignorelist_t *ignorelist;
 
 static int buddyinfo_config(const char *key, const char *value) {
-  if (ignorelist == NULL)
+  if (ignorelist == NULL) {
     ignorelist = ignorelist_create(1);
+    if (ignorelist != NULL) {
+      sfree(ignorelist);
+      return ENOMEM;
+    }
+  }
 
-  if (strcasecmp(key, "Zone") == 0)
-    ignorelist_add(ignorelist, value);
-  else
+  if (strcasecmp(key, "Zone") == 0) {
+    if (ignorelist_add(ignorelist, value)) {
+      ERROR("buddyinfo plugin: cannot add value to ignorelist");
+      return -1;
+    }
+  } else {
+    ERROR("buddyinfo plugin: invalid option: %s", key);
     return -1;
+  }
 
   return 0;
 }
@@ -87,14 +97,17 @@ static int buddyinfo_read(void) {
       continue;
 
     numfields = strsplit(dummy, fields, BUDDYINFO_FIELDS);
-    if (numfields != BUDDYINFO_FIELDS)
+    if (numfields != BUDDYINFO_FIELDS) {
+      WARNING("line %s doesn't contain %d orders, skipping...", buffer,
+              MAX_ORDER);
       continue;
+    }
 
     zone = fields[1];
     for (int i = 1; i <= MAX_ORDER; i++) {
       ssnprintf(pagesize_kb, sizeof(pagesize_kb), "%dKB",
                 NUM_OF_KB(pagesize, i - 1));
-      buddyinfo_submit(zone, pagesize_kb, atoll(fields[i + 1]));
+      buddyinfo_submit(zone, pagesize_kb, (int)atoll(fields[i + 1]));
     }
   }
 
@@ -102,8 +115,16 @@ static int buddyinfo_read(void) {
   return 0;
 }
 
+static int buddyinfo_shutdown(void) {
+  ignorelist_free(ignorelist);
+
+  return 0;
+}
+
 void module_register(void) {
+
   plugin_register_config("buddyinfo", buddyinfo_config, config_keys,
                          config_keys_num);
   plugin_register_read("buddyinfo", buddyinfo_read);
+  plugin_register_shutdown("buddy", buddyinfo_shutdown);
 }
--- a/src/buddyinfo.c
+++ b/src/buddyinfo.c
@@ -32,7 +32,8 @@
 #include <unistd.h>
 
 #define MAX_ORDER 11
-#define BUDDYINFO_FIELDS MAX_ORDER + 2 // "zone" + Name + (MAX_ORDER entries)
+#define BUDDYINFO_FIELDS                                                       \
+  MAX_ORDER + 4 // "Node" + node_num + "zone" + Name + (MAX_ORDER entries)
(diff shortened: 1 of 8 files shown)


mrunge:
Thank you for your contribution to collectd!

Before going through this in more detail, you have copyrights in there from people not being you. Is that correct?

asafkahlon:
> Before going through this in more detail, you have copyrights in there from people not being you. Is that correct?

Oops :)
I'm sorry, I probably missed that when I did a naive copy-paste from one of the other plugins. I'll push a correction for that.

src/buddyinfo.c
```diff
@@ -0,0 +1,109 @@
...
+#define NUM_OF_KB(pagesize, order) ((pagesize) / 1024) * (1 << (order))
+
+static const char *config_keys[] = {"Zone"};
+static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
+
+static ignorelist_t *ignorelist;
+
+static int buddyinfo_config(const char *key, const char *value) {
+  if (ignorelist == NULL)
+    ignorelist = ignorelist_create(1);
```
kwiatrox: Add check after ignorelist_create if ignorelist != NULL

src/buddyinfo.c
```diff
@@ -0,0 +1,109 @@
...
+
+  if (strcasecmp(key, "Zone") == 0)
+    ignorelist_add(ignorelist, value);
```
kwiatrox: Check if returned 0 (on success)

src/buddyinfo.c
```diff
@@ -0,0 +1,109 @@
...
+  else
+    return -1;
```
kwiatrox: Add ERROR about wrong option.

src/buddyinfo.c
```diff
@@ -0,0 +1,109 @@
...
+
+    numfields = strsplit(dummy, fields, BUDDYINFO_FIELDS);
+    if (numfields != BUDDYINFO_FIELDS)
+      continue;
+
+    zone = fields[1];
+    for (int i = 1; i <= MAX_ORDER; i++) {
+      ssnprintf(pagesize_kb, sizeof(pagesize_kb), "%dKB",
+                NUM_OF_KB(pagesize, i - 1));
+      buddyinfo_submit(zone, pagesize_kb, atoll(fields[i + 1]));
```
kwiatrox: cast from long long to int

src/buddyinfo.c
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  fclose(fh);
+  return 0;
+}
+
+void module_register(void) {
+  plugin_register_config("buddyinfo", buddyinfo_config, config_keys,
+                         config_keys_num);
+  plugin_register_read("buddyinfo", buddyinfo_read);
```
kwiatrox: You need also plugin_register_shutdown to clean ignorelist with `ignorelist_free`.

src/collectd.conf.pod
```diff
@@ -1455,6 +1455,20 @@ By default no detailed zone information is collected.
...
+
+The B<buddyinfo> plugin collects information by reading "/proc/buddyinfo".
+This file contains information about the number of available contagious
+physical pages at the moment.
+
+=over 4
+
+=item B<Zone> I<ZoneName>
+
+Zone to collect info about. Will collect all zones by default.
```
kwiatrox: Optional: leaving the string empty like `Zone ""` should also result in default behavior.

src/buddyinfo.c
```diff
@@ -0,0 +1,109 @@
...
+    return -1;
+  }
+
+  while (fgets(buffer, sizeof(buffer), fh) != NULL) {
+    if (!(dummy = strstr(buffer, "zone")))
+      continue;
+
+    numfields = strsplit(dummy, fields, BUDDYINFO_FIELDS);
+    if (numfields != BUDDYINFO_FIELDS)
+      continue;
```
kwiatrox: Is it normal and expected to have less fields?, if not add some WARNING/INFO here for easier debugging of problems.

asafkahlon:
> Thank you for your contribution.
> 
> 1. Why do you omit Node info? There can be more than one and depending on the use case it may be beneficial to add this info to plugin-instance. My example below:
> 
> ```
> Node 0, zone   Normal   1618   2566   3331   3607   1383    776    756    295     44     33  30127
> Node 1, zone   Normal    934   2174   2477   1909   1377    861    400    268    144     78  30827
> ```
> 
You're right. The current implementation doesn't contain node filtering.
I can add it to this PR if it's really needed, but I thought zone filtering is more important as a first stage and nodes can be added on a different PR (again, can be added here if needed).
BTW, I already used the zone name for plugin-instance so I'm not sure where this field should go...

> 2. What is the use case? You might consider use of /proc/pagetypeinfo instead of buddyinfo to get more detailed information.

The use case is giving the ability to track and debug the fragmentation status on the system.
In my opinion pagetypeinfo contains many information that maybe won't interest the average user (maybe it worth a separate plugin...?)

Anyway, thanks for the review!
I fixed your comments about the code.

src/buddyinfo.c
```diff
@@ -0,0 +1,130 @@
...
+
+static const char *config_keys[] = {"Zone"};
+static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
+
+static ignorelist_t *ignorelist;
+
+static int buddyinfo_config(const char *key, const char *value) {
+  if (ignorelist == NULL) {
+    ignorelist = ignorelist_create(1);
+    if (ignorelist != NULL) {
```
kwiatrox: if (ignorelist == NULL) {
  ERROR("buddyinfo plugin: ignorelist_create failed");
  return -ENOMEM; // be consistent  and return negative numbers on error
}

Please test your commits before submitting. In earlier comment I refered to (ignorelist != NULL) as a positive case.

src/buddyinfo.c
```diff
@@ -0,0 +1,130 @@
...
+      WARNING("line %s doesn't contain %d orders, skipping...", buffer,
+              MAX_ORDER);
+      continue;
+    }
+
+    zone = fields[1];
+    for (int i = 1; i <= MAX_ORDER; i++) {
+      ssnprintf(pagesize_kb, sizeof(pagesize_kb), "%dKB",
+                NUM_OF_KB(pagesize, i - 1));
+      buddyinfo_submit(zone, pagesize_kb, (int)atoll(fields[i + 1]));
```
kwiatrox: If biggest expected output fits the int, why not use `atoi`?

kwiatrox:
I think it is not much work to add node info to metrics. You can use a format like: 'Node0/DMA32', 'Node0/Normal', 'Node1/Normal' etc. for plugin-instance. I do not insist on filtering option.

src/buddyinfo.c
```diff
@@ -0,0 +1,135 @@
...
+static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
+
+static ignorelist_t *ignorelist;
+
+static int buddyinfo_config(const char *key, const char *value) {
+  if (ignorelist == NULL) {
+    ignorelist = ignorelist_create(1);
+    if (ignorelist == NULL) {
+      ERROR("buddyinfo plugin: ignorelist_create failed");
+      sfree(ignorelist);
```
kwiatrox: sfree has no effect and is not required here

Issue #3284, opened 2019-09-19
make check fails


mrunge:
*   Version of collectd: 5.9.1.git
*   Operating system / distribution: Fedora
*   Kernel version (if applicable): doesn't matter

## Actual behavior

running make check fails:
```
Testing strunescape ...
ok 87 - status == 0
ok 88 - buffer = "foo   bar"
ok 89 - status == 0
ok 90 - buffer = "      foo
"
ok 91 - status == 0
ok 92 - buffer = "With "quotes""
not ok 93 - status != 0
FAILURE.

```

that fails here:
```
  /* Backslash at buffer end */                                                
  strncpy(buffer, "\\t3\\56", sizeof(buffer));                                 
  status = strunescape(buffer, 4);                                             
  OK(status != 0);                                                             
  OK(buffer[0] == '\t');                                                       
  OK(buffer[1] == '3');                                                        
  OK(buffer[2] == 0);                                                          
  OK(buffer[3] == 0);                                                          
  OK(buffer[4] == '5');                                                        
  OK(buffer[5] == '6');                                                        
  OK(buffer[6] == '7');                                                        
                               
```

## Steps to reproduce

*   clone collectd from github
*   ./build.sh
*   configure
* make check

mrunge:
This issue was closed by #3306

PR #3305, opened 2019-10-14
Fix warnings and make check pass

mrunge:
ChangeLog: tests, fixes check
Resolves: https://github.com/collectd/collectd/issues/3284

### DIFF
--- a/src/utils/common/common_test.c
+++ b/src/utils/common/common_test.c
@@ -198,7 +198,7 @@ DEF_TEST(escape_slashes) {
   for (size_t i = 0; i < STATIC_ARRAY_SIZE(cases); i++) {
     char buffer[32];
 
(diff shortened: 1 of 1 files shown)


src/utils/common/common_test.c
```diff
@@ -198,7 +198,7 @@ DEF_TEST(escape_slashes) {
   for (size_t i = 0; i < STATIC_ARRAY_SIZE(cases); i++) {
     char buffer[32];
 
-    strncpy(buffer, cases[i].str, sizeof(buffer));
+    strcpy(buffer, cases[i].str);
```
rubenk: What if the source exceeds the buffer?

mrunge: It doesn't, since the source is defined in the test case itself.
I agree, it's not ideal in any case. I'm open to better suggestions here.

PR #3309, opened 2019-10-14
Remove LVM plugin

rubenk:
ChangeLog: Remove LVM plugin, liblvm2app has been deprecated


rpv-tomsk:
Why do you completely remove code,  instead of disable it by default?

rubenk:
It's easier ;-)
Disabled code is broken code, this is only for master so by the time we ship this it's 2030 and every distro removed liblvm2app.

mrunge:
I'm not opposed to remove broken code. Otherwise it'd hint, we are still "supporting" it.

larsf:
This didn't make it into 5.10

PR #3330, opened 2019-10-29
Avoid $< (implied source) in non-inference rules

efuss:
ChangeLog: build system: Avoid $< (implied source) in non-inference rules

Spell out input files in the collectd.grpc.pb.cc, collectd.pb.cc and types.pb.cc targets instead of using `$<`, which may only be defined for inference rules in non-GNU (but POSIX-compilent) versions of make.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2217,13 +2217,16 @@ BUILT_SOURCES += collectd.grpc.pb.cc collectd.pb.cc types.pb.cc
 
 collectd.grpc.pb.cc: $(srcdir)/proto/collectd.proto $(srcdir)/proto/types.proto
 	$(V_PROTOC)$(PROTOC) -I$(srcdir)/proto \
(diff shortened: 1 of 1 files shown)


rubenk:
This breaks the build on MacOS, see https://travis-ci.org/collectd/collectd/jobs/604583453#L31282:
```
ERROR: files left in build directory after distclean:

./types.grpc.pb.cc

./types.grpc.pb.h

make[1]: *** [distcleancheck] Error 1
```

efuss:
> This breaks the build on MacOS
Oops, I'll look into this after my vacation.
Could you provide me with a log of a working macOS build (before my modification)?

rubenk:
All builds are at https://travis-ci.org/collectd/collectd/builds

mrunge:
@efuss thank you for this fix. I believe it will pass CI once you rebase it. There has been a known issue with Mac builds and grpc (that's what fails).

Once rebased, that part is disabled in this case.

dago:
This has been fixed with #3376

PR #3331, opened 2019-10-30
Create internal library for utils_llist

jkohen:
We plan to use this library in the write_gcm plugin in the https://github.com/Stackdriver/collectd fork. I hope it's OK with you to make this change upstream before we are ready to contribute the Stackdriver plug-in into this codebase.

ChangeLog: no user-visible changes; internal refactoring.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -136,6 +136,7 @@ noinst_LTLIBRARIES = \
 	libheap.la \
 	libignorelist.la \
 	liblatency.la \
+	libllist.la \
 	liblookup.la \
 	libmetadata.la \
 	libmount.la \
@@ -244,8 +245,6 @@ collectd_SOURCES = \
 	src/daemon/utils_cache.h \
 	src/daemon/utils_complain.c \
 	src/daemon/utils_complain.h \
-	src/daemon/utils_llist.c \
-	src/daemon/utils_llist.h \
 	src/daemon/utils_random.c \
 	src/daemon/utils_random.h \
 	src/daemon/utils_subst.c \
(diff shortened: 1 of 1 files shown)


jkohen:
> Why do you need this?
> 
> From my understanding, all `utils_llist` functions are provided by collectd binary, and there is no need to include it into each plugin separately. There at least 6 plugins which uses `utils_list` now, and they do not include `utils_list` as separate library.

I'm using it in a utils library and its test, which don't link against the collectd binary: https://github.com/Stackdriver/collectd/pull/158/files#diff-c949f93d03f44a4217d7a138f9e2e54aR1845-R1869

I'm happy to change that if you have any suggestions. I followed the pattern used by the libformat_json and its test, which also don't link against the collectd binary: https://github.com/collectd/collectd/blob/master/Makefile.am#L424-L445

rpv-tomsk:
Ok, so your change is required to provide library to for tests binaries compile...
Sorry, I mistakenly thought you removed utils_llist functions from collectd binary [to be included into each plugin separately].

I think, this PR has nothing bad/wrong and it is good to merge.

jkohen:
@rpv-tomsk thanks! Can you approve it?

rpv-tomsk:
>Can you approve it?

Sorry, I have no such permissions.

jkohen:
@mrunge thanks for the approval. Can you merge it? I don't think I have access to do that.

rpv-tomsk:
Possible, you have to fall to your knees to get your PR merged.

I completely do not understand the behavior of Collectd Team members.
It shows that developers are not needed in the project.

mrunge:
> @mrunge thanks for the approval. Can you merge it? I don't think I have access to do that.

Hi, yes, of course. I simply forgot to press merge. Sorry about that.

PR #3333, opened 2019-10-31
Implement the NFS plugin for NetBSD

efuss:
ChangeLog: NFS plugin: An implementation for NetBSD (using sysctl) has been added.

Implement the NFS plugin for NetBSD using sysctl.
As NetBSD doesn't collect separate statistics for NFSv2 and NFSv3 and doesn't (yet) support NFSv4, this is v3 only.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -99,6 +99,10 @@ case $host_os in
     AC_DEFINE([KERNEL_SOLARIS], [1], [True if program is to be compiled for a Solaris kernel])
     ac_system="Solaris"
     ;;
+  *mingw32*)
+    AC_DEFINE([KERNEL_WIN32], [1], [True if program is to be compiled for a Windows kernel])
+    ac_system="Windows"
+    ;;
   *)
     ac_system="unknown"
     ;;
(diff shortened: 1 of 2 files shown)


mrunge:
Thank you for your contribution here. Please address the check issues, like clang-format and the changelog.

efuss:
I addessed some of the clang-format issues.

But clang-format also complains about `#endif` comments I didn't introduce and wants to re-order some `#include`s into a form that _does not compile_.

efuss:
I also added the ChangeLog: entry to the initial comment, sorry again for having missed that in the first place.

rpv-tomsk:
>But clang-format also complains about #endif comments I didn't introduce and wants to re-order some #includes into a form that does not compile.

That is completely normal case when technical limitations are put by "CollectD Team" which makes "Developers life interesting". Unfortunately, nobody cares about this.

In my POV, in normal community, when CI (clang-format) rules changes, the person who does this is responsible to convert codebase to new rules to do not hurt people. But this is not the case.

Here clang-format rules silently changes again and again - Team is not sure yet, which clang-rules he want. And I'm unsure, if Team want to have developers at all.

dago:
The issue has been fixed with the PR #3377

PR #3334, opened 2019-10-31
Native (sysctl-based) disk plugin implementation for NetBSD

efuss:
ChangeLog: Disk plugin: A native (sysctl-based) implementation for NetBSD was added.

Natively implement the disk plugin on NetBSD using `sysctl()`.
Based on prior work by Håvard Eidnes, <he@NetBSD.org>.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -99,6 +99,10 @@ case $host_os in
     AC_DEFINE([KERNEL_SOLARIS], [1], [True if program is to be compiled for a Solaris kernel])
     ac_system="Solaris"
     ;;
+  *mingw32*)
+    AC_DEFINE([KERNEL_WIN32], [1], [True if program is to be compiled for a Windows kernel])
+    ac_system="Windows"
+    ;;
   *)
     ac_system="unknown"
     ;;
@@ -111,6 +115,7 @@ AM_CONDITIONAL([BUILD_FREEBSD], [test "x$ac_system" = "xFreeBSD"])
 AM_CONDITIONAL([BUILD_LINUX], [test "x$ac_system" = "xLinux"])
 AM_CONDITIONAL([BUILD_OPENBSD], [test "x$ac_system" = "xOpenBSD"])
 AM_CONDITIONAL([BUILD_SOLARIS], [test "x$ac_system" = "xSolaris"])
+AM_CONDITIONAL([BUILD_WIN32], [test "x$ac_system" = "xWindows"])
 
 if test "x$ac_system" = "xSolaris"; then
   AC_DEFINE([_POSIX_PTHREAD_SEMANTICS], [1], [Define to enforce POSIX thread semantics under Solaris.])
@@ -160,6 +165,7 @@ AC_CHECK_HEADERS_ONCE([ \
   kstat.h \
   kvm.h \
   libgen.h \
+  locale.h \
   mntent.h \
   mnttab.h \
   netdb.h \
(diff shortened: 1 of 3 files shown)


mrunge:
Could you please address the check issues?

zebity:
@efuss given that you are looking at NetBSD, could this be extended to include the other BSD variants (FreeBSD, OpenBSD & MacOS (Darwin) ?

It would be good to get broader coverage. I can help with FreeBSD validation if needed.

Cheers,

Zebity

efuss:
> @efuss given that you are looking at NetBSD, could this be extended to include the other BSD variants (FreeBSD, OpenBSD & MacOS (Darwin) ?
@zebity I'm afraid no. I'm writing this as part of my job, and I guess I can't justify to my employer (Mathematisches Institut der Universität Bonn) spending my working time writing code for an OS we don't run.

efuss:
I fixed a stupid error (`#ifdef` instead of `#if`) I made when porting the patch from the stable version I use to HEAD.

I also addressed some of the clang-format issues. But clang-format also complains about indentation of `#endif` comments I didn't introduce and would like to normalize the line wrapping of two `disk_submit()` calls into a form I find harder to understand.

I edited the initial comment to include a ChangeLog: entry, sorry for having missed that.

zebity:
@efuss no point in fighting with clang-format as it validation is applied automatically and there is not way to stop it.
I had a glance through code today and saw that there was already specific case in code for KERNEL_FREEBSD.
My question is does it make sense to treat the KERNEL_FREEBSD, KERNEL_NETBSD and optionally OpenBSD differently ?
Cheers,

Zebity

rpv-tomsk:
>@efuss no point in fighting with clang-format as validation is applied automatically and there is not way to stop it.

Stupid position. When there is  "No way" to make developers happy, then there is "no way" for developers to work at all.

If you don't want to think about developers, developers will not think about project.

See #3333: 

>But clang-format also complains about #endif comments I didn't introduce and wants to re-order some #includes into a form that does not compile.

Should I highlight in bold "*clang-format ... wants to re-order some #includes into a form that does not compile*" for you to understand?

Why do you think it is normal that the team put responsibility on the developers for changes that they did not make?

rpv-tomsk:
For @efuss 
At some point in my work on Collectd, I found that the CI rules have changed. Code that has already been formatted (several PR was merged before, with CI passed) required new changes. Absolutely no response was received from the team. Complete ignore.  I don’t think it’s normal to make developers work on formatting again and again.

efuss:
> no point in fighting with clang-format as it validation is applied automatically and there is not way to stop it.
If that's indeed the case, someone in the know has to find a way to work around it. Applying clang-format to my code _breaks it_.

> I had a glance through code today and saw that there was already specific case in code for KERNEL_FREEBSD.
> My question is does it make sense to treat the KERNEL_FREEBSD, KERNEL_NETBSD and optionally OpenBSD differently ?
Yes. NetBSD doesn't have libgeom or devstat.

efuss:
> Why do you think it is normal that the team put responsibility on the developers for changes that they did not make?
Well, the main point (or so I think) is that, when the formatting rules change, the formatting changes needed to satisfy the new rules should go into one single seperate commit to keep the revision history clean.

zebity:
@efuss ouch , if clang-format breaks code then we need to find way to work around this.

I say “work around” not because I agree with formatting CI policy, rather I see it as an immovable object and all protests about it go unheard by people who can change it.

On change front, my point on consistent BSD approach was to to suggest if they could not all leverage your sysctl based mechanism as this is available on all BSD variants. This would reduce conditional compile variations.

But this could break something ... , so you have to look at details of variations.

Good luck,

Zebity.

efuss:
> On change front, my point on consistent BSD approach was to to suggest if they could not all leverage your sysctl based mechanism as this is available on all BSD variants. This would reduce conditional compile variations.
Feel free to test the sysctl based code on other BSDs. I would be happy to have contributed to consistent code.

dago:
This has been fixed in #3379

dago:
Closing as #3379 has been merged

PR #3344, opened 2019-11-16
Fix linking with Python 3.8

Self-Perfection:
Since Python 3.8 --embed flag needs to be provided to python-config to embed python.
Reference: https://bugs.python.org/issue36721

This was partially addressed in #3170 but only fixed building and not linking with python.

ChangeLog: Fix linking with Python 3.8

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4711,7 +4711,7 @@ if test "$PYTHON_CONFIG" != ""; then
   if test $? -ne 0; then
     with_libpython="no"
   fi
-  LIBPYTHON_LDFLAGS="`${PYTHON_CONFIG} --ldflags`"
+  LIBPYTHON_LDFLAGS="`${PYTHON_CONFIG} --ldflags --embed`" || LIBPYTHON_LDFLAGS="`${PYTHON_CONFIG} --ldflags`"
   if test $? -ne 0; then
     with_libpython="no"
   fi


configure.ac
```diff
@@ -4711,7 +4711,7 @@ if test "$PYTHON_CONFIG" != ""; then
   if test $? -ne 0; then
     with_libpython="no"
   fi
-  LIBPYTHON_LDFLAGS="`${PYTHON_CONFIG} --ldflags`"
+  LIBPYTHON_LDFLAGS="`${PYTHON_CONFIG} --ldflags --embed`" || LIBPYTHON_LDFLAGS="`${PYTHON_CONFIG} --ldflags`"
```
rubenk: please redirect stderr of the first command to /dev/null, otherwise an error shows up on all systems where we don't have python 3.8.

Self-Perfection: I have checked `python2-config --unsupported-flag` and `python3-config --unsupported-flag` behaviour on couple of systems. And everywhere `python-config` sent error message to stdout. So on systems with python older than 3.8 error message would be stored in `$LIBPYTHON_LDFLAGS` and not displayed.

On the other hand, future versions of python might correctly give error to stderr if something went wrong unrelated to `--embed` flag.

Therefore it seems to me current variant, without devnulling of stderr, is better.

Do you insist on stderr redirection anyway?

mrunge: That makes sense to me and is also consistent with the other part of python3.8 work

PR #3345, opened 2019-11-20
redis module: add collect master_repl_offset and slave_repl_offset

lurkr2:
For Redis plugin - collect replication offsets
ChangeLog: redis plugin: add collect master_repl_offset and slave_repl_offset

### DIFF
--- a/src/redis.c
+++ b/src/redis.c
@@ -394,6 +394,49 @@ static int redis_handle_info(char *node, char const *info_line,
   return 0;
 } /* }}} int redis_handle_info */
 
+static int redis_slave_info(char *node, char const *info_line) /* {{{ */
+{
+  char *saveptr = NULL;
+  char *str = strstr(info_line, "connected_slaves");
+  char slave[] = "slave";
+  if (str != NULL) {
+    str = strtok_r(str, ":\n", &saveptr);
+    while (str != NULL) {
+      // slaveXXX:ip=XXXX,port=XXXX,state=online,offset=XXXX,lag=XXXX
+      if (strncmp(str, slave, sizeof(slave) - 1) == 0 &&
+          isdigit((unsigned char)*(str + sizeof(slave) - 1))) {
+        value_t val;
+        static char slavename[64];
+        char *ip;
+        char *port;
+        char *lag;
+        do {
+          strtok_r(NULL, "=", &saveptr);
+          ip = strtok_r(NULL, ",", &saveptr);
+          strtok_r(NULL, "=", &saveptr);
+          port = strtok_r(NULL, ",", &saveptr);
+          ssnprintf(slavename, sizeof(slavename), "slave_%s_%s", ip, port);
+          strtok_r(NULL, ",", &saveptr);
+          strtok_r(NULL, ",", &saveptr);
+          strtok_r(NULL, "=", &saveptr);
+          lag = strtok_r(NULL, "\n", &saveptr);
+          if (parse_value(lag, &val, DS_TYPE_DERIVE) != 0) {
+            WARNING("redis plugin: Unable to parse lag value of slave `%s'.",
+                    slavename);
+            return -1;
+          }
+          redis_submit(node, "lag", slavename, val);
+          str = strtok_r(NULL, ":\n", &saveptr);
+        } while (strncmp(str, slave, sizeof(slave) - 1) == 0 &&
+                 isdigit((unsigned char)*(str + sizeof(slave) - 1)));
+        break;
+      }
+      str = strtok_r(NULL, ":\n", &saveptr);
+    }
+  }
+  return 0;
+} /* }}} int redis_slave_info */
+
 static int redis_handle_query(redis_node_t *rn, redis_query_t *rq) /* {{{ */
 {
   redisReply *rr;
@@ -689,6 +732,8 @@ static void redis_read_server_info(redis_node_t *rn) {
 
   if (rn->report_cpu_usage)
     redis_cpu_usage(rn->name, rr->str);
+
+  redis_slave_info(rn->name, rr->str);
 
   freeReplyObject(rr);
 } /* void redis_read_server_info */
--- a/src/redis.c
+++ b/src/redis.c
@@ -682,6 +682,10 @@ static void redis_read_server_info(redis_node_t *rn) {
                     "total_net_input_bytes", DS_TYPE_DERIVE);
   redis_handle_info(rn->name, rr->str, "total_bytes", "output",
                     "total_net_output_bytes", DS_TYPE_DERIVE);
+  redis_handle_info(rn->name, rr->str, "current", "master_repl_offset",
+                    "master_repl_offset", DS_TYPE_GAUGE);
+  redis_handle_info(rn->name, rr->str, "current", "slave_repl_offset",
+                    "slave_repl_offset", DS_TYPE_GAUGE);
 
   redis_keyspace_usage(rn, rr->str);
 
--- a/src/redis.c
+++ b/src/redis.c
@@ -399,40 +399,40 @@ static int redis_slave_info(char *node, char const *info_line) /* {{{ */
   char *saveptr = NULL;
   char *str = strstr(info_line, "connected_slaves");
   char slave[] = "slave";
-  if (str != NULL) {
-    str = strtok_r(str, ":\n", &saveptr);
-    while (str != NULL) {
-      // slaveXXX:ip=XXXX,port=XXXX,state=online,offset=XXXX,lag=XXXX
-      if (strncmp(str, slave, sizeof(slave) - 1) == 0 &&
-          isdigit((unsigned char)*(str + sizeof(slave) - 1))) {
-        value_t val;
-        static char slavename[64];
-        char *ip;
-        char *port;
-        char *lag;
-        do {
-          strtok_r(NULL, "=", &saveptr);
-          ip = strtok_r(NULL, ",", &saveptr);
-          strtok_r(NULL, "=", &saveptr);
-          port = strtok_r(NULL, ",", &saveptr);
-          ssnprintf(slavename, sizeof(slavename), "slave_%s_%s", ip, port);
-          strtok_r(NULL, ",", &saveptr);
-          strtok_r(NULL, ",", &saveptr);
-          strtok_r(NULL, "=", &saveptr);
-          lag = strtok_r(NULL, "\n", &saveptr);
-          if (parse_value(lag, &val, DS_TYPE_DERIVE) != 0) {
-            WARNING("redis plugin: Unable to parse lag value of slave `%s'.",
-                    slavename);
-            return -1;
-          }
-          redis_submit(node, "lag", slavename, val);
-          str = strtok_r(NULL, ":\n", &saveptr);
-        } while (strncmp(str, slave, sizeof(slave) - 1) == 0 &&
-                 isdigit((unsigned char)*(str + sizeof(slave) - 1)));
-        break;
-      }
-      str = strtok_r(NULL, ":\n", &saveptr);
+  if (str != NULL)
+    return 0;
+  str = strtok_r(str, ":\n", &saveptr);
+  while (str != NULL) {
+    // slaveXXX:ip=XXXX,port=XXXX,state=online,offset=XXXX,lag=XXXX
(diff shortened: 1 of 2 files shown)


src/redis.c
```diff
@@ -682,6 +682,10 @@ static void redis_read_server_info(redis_node_t *rn) {
                     "total_net_input_bytes", DS_TYPE_DERIVE);
   redis_handle_info(rn->name, rr->str, "total_bytes", "output",
                     "total_net_output_bytes", DS_TYPE_DERIVE);
+  redis_handle_info(rn->name, rr->str, "current", "master_repl_offset",
```
mkobyli: What do you think about something like this: "current_repl_offset" as type and "master" as type_instance? In my opinion "current" as a type can be too generic.

lurkr2: i`m select current because this type already exists in types.db :)

rpv-tomsk: `current` in types.db is in terms of electrical engineering meaning, so please choose another type.

lurkr2: maybe "flow" or "gauge" or "records" ?

rpv-tomsk:
Can you please provide charts example built based on these metrics?

src/redis.c
rpv-tomsk: No flow, no gauge.
These offsets are in records or bytes?

I think you might add new type `offset` and set type instance to `master_repl` or `slave_repl`.

BUT:

Another question: why do you need these metrics? 
From my POV, they are not informative, their value grow monotonically and can't be placed on charts.

Just looked into my servers, results are:

```
role:slave
slave_repl_offset:58271319441
master_repl_offset:58271319441
```

```
role:master
connected_slaves:1
...
master_repl_offset:58272087142
```

So I think these raw values are not useful in wild.
Please argue, why they are needed in form you try to propose.

lurkr2: Offset measured in records.
I didn’t want to add a new type in types.db as this will increase the file size and give absolutely nothing, the name of the metric fully reflects the essence.
My case:
Compare two or more offsets, between master and slave/slaves per fixed interval,
the result will be lagg master-slave(s) in ops(records)
If lagg > x or slaves < n - send alarm message.

lurkr2: > Can you please provide charts example built based on these metrics?

stack: grafana<-graphite<-collectd
https://notify.ekb-info.ru/uploads/2019-11-22-13-52-37-tsqnalxdapbenddx.png
https://notify.ekb-info.ru/uploads/2019-11-22-13-55-40-fwtmzkbdncmijcnf.png

rpv-tomsk: >Compare two or more offsets, between master and slave/slaves per fixed interval,
the result will be lagg master-slave(s) in ops(records)

Then this needs to be done on Collectd level, so its `threshold` plugin could be used to sent notifications.

Collectd is not agent used only for graphana.
Personally, I dislike implementations when Collectd is used to sent raw metrics which do not represent nothing by itself, for future calculations which could be done on Collector side too.

rpv-tomsk:
>Compare two or more offsets, between master and slave/slaves per fixed interval,
the result will be lagg master-slave(s) in ops(records)

Then this needs to be done on Collectd level, so its `threshold` plugin could be used to sent notifications.

Collectd is not agent used only for graphana.
Personally, I dislike implementations when Collectd is used to sent raw metrics which do not represent nothing by itself, for future calculations which could be done on Collector side too.

lurkr2:
> > Compare two or more offsets, between master and slave/slaves per fixed interval,
> > the result will be lagg master-slave(s) in ops(records)
> 
> Then this needs to be done on Collectd level, so its `threshold` plugin could be used to sent notifications.
> 
> Collectd is not agent used only for graphana.
> Personally, I dislike implementations when Collectd is used to sent raw metrics which do not represent nothing by itself, for future calculations which could be done on Collector side too.

it is impossible to perform the transformations that you are talking about on side collectd, since these are 2 or more different machines (master and more slaves).
IMHO, his is the essence of monitoring the replication lag, ask the master, ask the slave, and make a comparison on the monitoring side.
Do you have ideas on how to do this on one machine in push model?

rpv-tomsk:
As I said before, I see all required data on master:

```
role:master
connected_slaves:1
slave0:ip=192.168.1.200,port=6379,state=online,offset=58272087142,lag=1
master_repl_offset:58272087142
```

lurkr2:
> As I said before, I see all required data on master:
> 
> ```
> role:master
> connected_slaves:1
> slave0:ip=192.168.1.200,port=6379,state=online,offset=58272087142,lag=1
> master_repl_offset:58272087142
> ```

a good idea, let's do it :)

rpv-tomsk:
You can compare charts produced by both methods... 

At next step it can be implemented notification about slave status change (that should be done on plugin level).

dago:
@lurkr2 Would you mind changing the code as you suggested?

lurkr2:
> @lurkr2 Would you mind changing the code as you suggested?

No free time now :(, maybe later, sorry

dago:
@lurkr2 Sure, I'll set this to pending action then, feel free to pick up work on this any time.

lurkr2:
> @lurkr2 Sure, I'll set this to pending action then, feel free to pick up work on this any time.

thanx

lurkr2:
ChangeLog: redis plugin: add collect slaves lag for master

rpv-tomsk:
Hi, 
Can you please show example of charts for new metric?

rpv-tomsk:
I want to ask maintainers to ask such examples for each newly-added metric/ metrics of new plugin.

src/redis.c
```diff
@@ -394,6 +394,49 @@ static int redis_handle_info(char *node, char const *info_line,
   return 0;
 } /* }}} int redis_handle_info */
 
+static int redis_slave_info(char *node, char const *info_line) /* {{{ */
+{
+  char *saveptr = NULL;
+  char *str = strstr(info_line, "connected_slaves");
+  char slave[] = "slave";
+  if (str != NULL) {
```
rpv-tomsk: Hi, can I please ask you to reformat your code, to get fast return.

```
if (str == NULL) 
   return 0;

... rest of code...
```

That is preferred and more readable.
If that is possible in do { } block, please define variable on first use and not in the beginning of block.
https://github.com/collectd/collectd/blob/master/docs/review_comments.md

Thanks.

lurkr2:
> Hi,
> Can you please show example of charts for new metric (new implementation)?

of course

src/redis.c
```diff
@@ -394,6 +394,49 @@ static int redis_handle_info(char *node, char const *info_line,
...
+  while (str != NULL) {
+    // slaveXXX:ip=XXXX,port=XXXX,state=online,offset=XXXX,lag=XXXX
+    if (strncmp(str, slave, sizeof(slave) - 1) == 0 &&
+        isdigit((unsigned char)*(str + sizeof(slave) - 1))) {
+      value_t val;
+      static char slavename[64];
+      char *ip;
+      char *port;
+      char *lag;
+      do {
```
rpv-tomsk: Looking into `redis_db_stats()` I found there is more precise implementation.

Could you please rework this piece of code to not depend on fields order, i.e. using `strstr(str, 'offset')`.
That also would be more clear to understand which field you use, `offset` or `lag`.

As a future update, you can check `state=online` and implement notifications about state change.

Unfortunately, currently I have no Redis in cluster setup to make tests.

src/redis.c
```diff
@@ -394,6 +394,49 @@ static int redis_handle_info(char *node, char const *info_line,
...
 } /* }}} int redis_handle_info */
 
+static int redis_slave_info(char *node, char const *info_line) /* {{{ */
+{
+  char *saveptr = NULL;
+  char *str = strstr(info_line, "connected_slaves");
+  char slave[] = "slave";
+  if (str != NULL)
+    return 0;
+  str = strtok_r(str, ":\n", &saveptr);
```
rpv-tomsk: It seems you don't need to include `:` here and in L435.

rpv-tomsk:
Please try this parser implementation:
```
#include <stdio.h>
#include <string.h>

#define WARNING printf
#define ERROR printf
#define ssnprintf snprintf

static int redis_slave_info(char const *info_line)
{
  char *line = strstr(info_line, "connected_slaves");
  if (line == NULL)
    return 0;

  char *ptr = line;
  char *saveptr;
  while ((line = strtok_r(ptr, "\n", &saveptr)) != NULL) {
    ptr = NULL;
    // slaveXXX:ip=XXXX,port=XXXX,state=online,offset=XXXX,lag=XXXX

    const char slave[] = "slave";
    if (strncmp(line, slave, sizeof(slave) - 1) != 0 || !isdigit((unsigned char)*(line + sizeof(slave) - 1)))
      continue;

    char *values = strstr(line, ":");
    if (values == NULL) {
        ERROR("redis plugin: not found ':' separator in line '%s'", line);
        continue;
    }
    values++;

    char *ip = NULL;
    char *port = NULL;
    char *lag = NULL;

    char *field;
    char *saveptr_field = NULL;
    while ((field = strtok_r(values, "=", &saveptr_field)) != NULL) {
      values = NULL;

      char **savevalue = NULL;
      if (strcmp(field, "ip") == 0)
        savevalue = &ip;
      else if (strcmp(field, "port") == 0)
        savevalue = &port;
      else if (strcmp(field, "lag") == 0)
        savevalue = &lag;

      if ((field = strtok_r(NULL, ",", &saveptr_field)) == NULL) //value
        continue;

      if (savevalue)
        *savevalue = field; //save value
    }

    if (ip == NULL) {
        ERROR("redis plugin: not found slave ip in line '%s'", line);
        continue;
    }
    if (port == NULL) {
        ERROR("redis plugin: not found slave port in line '%s'", line);
        continue;
    }
printf("Found IP: %s\n", ip);
printf("Found PORT: %s\n", port);

    static char slavename[64];
    ssnprintf(slavename, sizeof(slavename), "slave_%s_%s", ip, port);

    if (!lag) {
      WARNING("redis plugin: Unable to find lag value of slave `%s'.", slavename);
      continue;
    }

/*
    value_t val;
    if (parse_value(lag, &val, DS_TYPE_DERIVE) != 0) {
      WARNING("redis plugin: Unable to parse lag value of slave `%s'.", slavename);
      continue;
    }

    redis_submit(node, "lag", slavename, val);
*/
  }
  return 0;
} /* }}} int redis_slave_info */

int main()
{
    char str[] = "connected_slaves:0\nline:value\n\nslave000:ip=1.2.3.4,port=5678,state=online,offset=9876,lag=432\nslave001:ip=1.2.3.5,port=5679,state=online,offset=9876,lag=432\n";
    redis_slave_info(str);
    return (0);
}
```

Also, please explain how value 64 was used in `static char slavename[64];`.
This might be incorrect/baseless assumption.

rpv-tomsk:
Due to possible duplicate of data (slave lag can be reported from master and from slaves), I think this feature (report) should be configurable.
I propose `ReportSlavesLag` option. New option requires to be documented.

src/redis.c
rpv-tomsk: An argue to use strict parser: look into `sentinelRefreshInstanceInfo` implementation in `redis/src/sentinel.c` code:
```
/* New format. */
                ip = strstr(l,"ip="); if (!ip) continue;
                ip += 3; /* Now ip points to start of ip address. */
                port = strstr(l,"port="); if (!port) continue;
                port += 5; /* Now port points to start of port number. */
                /* Nul term both fields for easy access. */
                end = strchr(ip,','); if (end) *end = '\0';
                end = strchr(port,','); if (end) *end = '\0';
```

rpv-tomsk:
Also I propose to discuss another redis features:
1) Notification sent on Redis status change (master <> slave)
2) Send 'slave delay' metric based on `master_link_down_since_seconds` and `master_last_io_seconds_ago` values.

src/redis.c
lurkr2: thanks, I will see

lurkr2:
I think it is right to split the “INFO” request for a redis server into a partition (server, memory, statistics, replication, processor) in order to avoid blocking the redis server when moving data to the slaves (affected more than 32 GB of memory)

what do you think of it?

rpv-tomsk:
>blocking the redis server when moving data to the slaves

Hmm, when that blocking is possible?

rpv-tomsk:
Looking into `redis-4.0.10/src/server.c` code, I see what choosing section of INFO command affects only output of data. All data for full INFO command built/exists anyway.

Issue #3347, opened 2019-11-20
collectd 5.10, some plugins didn't register a configuration callback


seblu:
*   Version of collectd: 5.10.0
*   Operating system / distribution: Arch Linux
*   Kernel version (if applicable): 5.3.0

## Rationale

After Upgrading to from 5.9.2 to 5.10.0, my collectd daemons was not able to start anymore.
Some plugins reports they « didn't register a configuration callback ».

In my configuration, only the following plugins are concerned:
- contextswitch
- cpufreq
- entropy
- uptime
- users

Removing them from the configuration file and the daemon starts.

### config

```
AutoLoadPlugin true
<Plugin "contextswitch">
</Plugin>
<Plugin "cpufreq">
</Plugin>
<Plugin "uptime">
</Plugin>
<Plugin "users">
</Plugin>
...
```

### log output
```
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "syslog" successfully loaded.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "conntrack" successfully loaded.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "contextswitch" successfully loaded.
nov. 20 16:03:30 collectd[21343]: Found a configuration for the `contextswitch' plugin, but the plugin didn't register a configuration callback.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "cpu" successfully loaded.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "cpufreq" successfully loaded.
nov. 20 16:03:30 collectd[21343]: Found a configuration for the `cpufreq' plugin, but the plugin didn't register a configuration callback.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "df" successfully loaded.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "disk" successfully loaded.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "entropy" successfully loaded.
nov. 20 16:03:30 collectd[21343]: Found a configuration for the `entropy' plugin, but the plugin didn't register a configuration callback.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "ethstat" successfully loaded.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "hddtemp" successfully loaded.
nov. 20 16:03:30 collectd[21343]: plugin_load: plugin "interface" successfully loaded.
[...]

PR #3217, opened 2019-07-13
daemon: Check if plugin actually loaded before reporting configuration issues

rpv-tomsk:
Changelog: daemon: Check if plugin actually loaded before reporting configuration issues

1.1)

Case: Plugin not loaded, "complex" configuration found.

Old Behaviour:  

  * dispatch_value_plugin() not called. 
  * Reported message (for each block in plugin configuration):

  "There is a `%s' block within the configuration for the %s plugin. The plugin either only expects "simple" configuration statements "
  "or wasn't loaded using `LoadPlugin'. Please check your configuration."
   

  * dispatch_block_plugin() returns 0 (daemon continues to run).

1.2) 

Case: Plugin not loaded, "simple" configuration found.

Old Behaviour: 
 
  * called dispatch_block_plugin()->dispatch_value_plugin()->cf_dispatch()
  * Reported message:

  "Found a configuration for the `%s' plugin, but the plugin isn't loaded or didn't register a configuration callback.",

  * dispatch_block_plugin() returns -1 (daemon fails to start).

2)

Case: Plugin supports only "simple" configuration, but complex one configured.

Old behaviour:

  * Reported message (for each block in plugin configuration):

  "There is a `%s' block within the configuration for the %s plugin. The plugin either only expects "simple" configuration statements "
  "or wasn't loaded using `LoadPlugin'. Please check your configuration."

  * daemon continues to run.

3.1) 

Case: Plugin loaded but does not register configuration callback. "simple" configuration found

Old behaviour: 

  * Reported message:

  "Found a configuration for the `%s' plugin, but the plugin isn't loaded or didn't register a configuration callback.",

  * dispatch_block_plugin() returns -1 from cf_dispatch(), (daemon fails to start).

3.2) 

Case: Plugin loaded but does not register configuration callback. "complex" configuration found

Old behaviour: 

  * Reported message:

  "There is a `%s' block within the configuration for the %s plugin. The plugin either only expects "simple" configuration statements "
  "or wasn't loaded using `LoadPlugin'. Please check your configuration."

  * daemon continues to run.
  

-----

As you can see, message and behaviour depends on configuration "complexity", and strange a bit - daemon internals easily knows, is plugin was loaded or not.

New behaviour uses that knowledge to not confuse users:

   
1) 

Case: Plugin was not loaded:

  * Reported message (once per plugin configuration):
     "There is configuration for the `%s' plugin, but the plugin isn't loaded. Please check your configuration."

[...]

PR #3349, opened 2019-11-23
daemon: Fix regression caused by #3217

rpv-tomsk:
Empty <Plugin> section can be used to autoload plugin,
which does not supports configuration callback.

Closes: #3347

Changelog: daemon: Fix regression caused by #3217


rubenk:
Thanks Pavel :-)

PR #3357, opened 2019-12-19
Host plugin: Add new plugin to detect new and missing hosts.

minfrin:
ChangeLog: Host plugin: Add new plugin to detect new and missing hosts.

This PR adds a new plugin that keeps track of hosts that have been seen by collectd, and sends notifications when a new host is seen, or an existing is no longer seen for a period of time, respectively.

It depends on liblmdb to provide the underlying index, chosen for low cost.

### DIFF
--- a/contrib/redhat/collectd.spec
+++ b/contrib/redhat/collectd.spec
@@ -78,6 +78,7 @@
 %define with_gmond 0%{!?_without_gmond:1}
 %define with_gps 0%{!?_without_gps:1}
 %define with_hddtemp 0%{!?_without_hddtemp:1}
+%define with_host 0%{!?_without_host:1}
 %define with_hugepages 0%{!?_without_hugepages:1}
 %define with_interface 0%{!?_without_interface:1}
 %define with_ipc 0%{!?_without_ipc:1}
@@ -509,6 +510,16 @@ Requires:	%{name}%{?_isa} = %{version}-%{release}, hddtemp
 %description hddtemp
 The HDDTemp plugin collects the temperature of hard disks. The temperatures are
 provided via SMART and queried by the external hddtemp daemon.
+%endif
+
+%if %{with_host}
+%package host
+Summary:        Host plugin for collectd
+Group:          System Environment/Daemons
+Requires:       %{name}%{?_isa} = %{version}-%{release}, liblmdb
+%description host
+The host plugin sends notifications when a host is seen for the first time,
+and when a host has not been seen for a configurable interval.
 %endif
 
 %if %{with_intel_pmu}
@@ -1333,6 +1344,12 @@ Collectd utilities
 %define _with_hddtemp --disable-hddtemp
 %endif
 
+%if %{with_host}
+%define _with_host --enable-host
+%else
+%define _with_host --disable-host
+%endif
+
 %if %{with_hugepages}
 %define _with_hugepages --enable-hugepages
 %else
@@ -2055,6 +2072,7 @@ Collectd utilities
 	%{?_with_gps} \
 	%{?_with_grpc} \
 	%{?_with_hddtemp} \
+        %{?_with_host} \
 	%{?_with_hugepages} \
 	%{?_with_intel_pmu} \
 	%{?_with_intel_rdt} \
@@ -2646,6 +2664,11 @@ fi
 %{_libdir}/%{name}/hddtemp.so
 %endif
 
+%if %{with_host}
+%files host
+%{_libdir}/%{name}/host.so
+%endif
+
 %if %{with_intel_pmu}
 %files intel_pmu
 %{_libdir}/%{name}/intel_pmu.so
@@ -2904,6 +2927,9 @@ fi
 %doc contrib/
 
 %changelog
+* Thu Dec 19 2019 Graham Leggett <minfrin@sharp.fm> - 5.9.2-2
+- Add new host plugin
+
 * Mon Oct 14 2019 Ruben Kerkhof <ruben@rubenkerkhof.com> - 5.9.2-2
 - Remove lvm plugin, liblvmapp has been deprecated upstream
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3449,6 +3449,28 @@ TCP-Port to connect to. Defaults to B<7634>.
 
 =back
 
+=head2 Plugin C<host>
+
+The host plugin keeps track of hosts that have been recently seen, and sends
+notifications when a host is seen for the first time, and subsequently when
+a host has gone missing after not being seen for a period of time.
+
+=over 4
+
+=item B<HostPath> I<path/to/database>
+
+Path and filename of the LMDB hosts database. Defaults to B<hosts>.
+
+=item B<LostInterval> I<Seconds>
+
+Number of seconds to pass before a host is notified as missing. Defaults to B<30>.
+
+=item B<DelayInterval> I<Seconds>
+
+Number of seconds after startup to suppress missing notifications. Defaults to B<0>.
+
+=back
+
 =head2 Plugin C<hugepages>
 
 To collect B<hugepages> information, collectd reads directories
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3463,11 +3463,11 @@ Path and filename of the LMDB hosts database. Defaults to B<hosts>.
 
 =item B<LostInterval> I<Seconds>
 
-Number of seconds to pass before a host is notified as missing. Defaults to B<30>.
+Number of seconds to pass before a host is notified as missing. Defaults to B<60>.
 
 =item B<DelayInterval> I<Seconds>
 
-Number of seconds after startup to suppress missing notifications. Defaults to B<0>.
+Number of seconds after startup to suppress missing notifications. Defaults to B<60>.
 
 =back
 
--- a/src/host.c
+++ b/src/host.c
@@ -41,9 +41,9 @@
 #define PLUGIN_NAME "host"
 
 #define DEFAULT_HOST_PATH "hosts"
-#define DEFAULT_LOST_INTERVAL 30
+#define DEFAULT_LOST_INTERVAL 60
 #define DEFAULT_THREAD_INTERVAL 2
-#define DEFAULT_DELAY_INTERVAL 0
+#define DEFAULT_DELAY_INTERVAL 60
 
 static const char *config_keys[] = {"HOSTPATH", "LOSTINTERVAL", "DELAYINTERVAL"};
 static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
@@ -72,6 +72,12 @@ static void *host_thread(void *arg) /* {{{ */
   }
 
   pthread_mutex_lock(&host_lock);
+
+  ts_wait.tv_sec = tv_begin.tv_sec + delay_interval + thread_interval;
+  ts_wait.tv_nsec = 0;
+
+  pthread_cond_timedwait(&host_cond, &host_lock, &ts_wait);
+
   while (host_thread_loop > 0) {
     MDB_txn *txn;
     MDB_dbi dbi;
@@ -140,8 +146,7 @@ static void *host_thread(void *arg) /* {{{ */
       host = key.mv_data;
       tv_then = data.mv_data;
 
-      if ((tv_now.tv_sec - tv_then->tv_sec >= lost_interval) &&
-              (tv_now.tv_sec - tv_begin.tv_sec >= delay_interval)) {
+      if (tv_now.tv_sec - tv_then->tv_sec >= lost_interval) {
 
         notification_t n;
         char message[NOTIF_MAX_MSG_LEN];
@@ -421,7 +426,7 @@ static int host_write(const data_set_t *ds, const value_list_t *vl,
           ssnprintf(message, sizeof(message), "Host is found: %.*s",
                   (int)sizeof(vl->host), vl->host);
           notification_init(&n, NOTIF_OKAY, message, vl->host, PLUGIN_NAME,
-                            NULL, "host", "lost");
+                            NULL, "host", "found");
           n.time = cdtime();
 
           plugin_dispatch_notification(&n);
--- a/src/host.c
+++ b/src/host.c
@@ -1,0 +1,466 @@
+/**
+ * collectd - src/host.c
+ * Copyright (C) 2019       Graham Leggett
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ * Authors:
+ *   Graham Leggett <minfrin at sharp.fm>
+ *
+ **/
+
+/*
+ * This plugin keeps track of hosts that have been seen by collectd, and sends
+ * notifications when a new host is seen, and when a host has not been seen.
+ *
+ */
+
+#include "collectd.h"
+
+#include "plugin.h"
+#include "utils/common/common.h"
+
+#include <lmdb.h>
+
+#define PLUGIN_NAME "host"
+
+#define DEFAULT_HOST_PATH "hosts"
+#define DEFAULT_LOST_INTERVAL 30
+#define DEFAULT_THREAD_INTERVAL 2
+#define DEFAULT_DELAY_INTERVAL 0
+
+static const char *config_keys[] = {"HOSTPATH", "LOSTINTERVAL", "DELAYINTERVAL"};
+static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
+static char *host_path = DEFAULT_HOST_PATH;
+static int lost_interval = DEFAULT_LOST_INTERVAL;
+static int delay_interval = DEFAULT_DELAY_INTERVAL;
+static int thread_interval = DEFAULT_THREAD_INTERVAL;
+
+static pthread_mutex_t host_lock = PTHREAD_MUTEX_INITIALIZER;
+static pthread_cond_t host_cond = PTHREAD_COND_INITIALIZER;
+static int host_thread_loop;
+static pthread_t host_thread_id;
+
+static MDB_env *env;
+
+static void *host_thread(void *arg) /* {{{ */
+{
+  struct timeval tv_begin;
+  struct timeval tv_now;
+  struct timespec ts_wait;
+
+  if (gettimeofday(&tv_begin, NULL) < 0) {
+    ERROR("host plugin: gettimeofday failed: %s", STRERRNO);
+    host_thread_loop = 0;
+    return (void *)0;
+  }
+
+  pthread_mutex_lock(&host_lock);
+  while (host_thread_loop > 0) {
+    MDB_txn *txn;
+    MDB_dbi dbi;
+    MDB_val key, data;
+    MDB_cursor *cursor;
+    int rc;
+
+    if (gettimeofday(&tv_now, NULL) < 0) {
+      ERROR("host plugin: gettimeofday failed: %s", STRERRNO);
+      host_thread_loop = 0;
+      break;
+    }
+
+    pthread_mutex_unlock(&host_lock);
+
+    rc = mdb_txn_begin(env, NULL, 0, &txn);
+    if (rc) {
+      ERROR("mdb_txn_begin returned: %s (%d)", mdb_strerror(rc), rc);
+      goto skip;
+    }
+
+    rc = mdb_dbi_open(txn, NULL, MDB_CREATE, &dbi);
+    if (rc) {
+      ERROR("mdb_dbi_open returned: %s (%d)", mdb_strerror(rc), rc);
+      mdb_txn_abort(txn);
+      goto skip;
+    }
+
+    rc = mdb_cursor_open(txn, dbi, &cursor);
+    if (rc) {
+      ERROR("mdb_cursor_open returned: %s (%d)", mdb_strerror(rc), rc);
+      mdb_txn_abort(txn);
+      goto skip;
+    }
+
+    while ((rc = mdb_cursor_get(cursor, &key, &data, MDB_NEXT)) == 0) {
+      notification_t n;
+      struct timeval *tv_then;
+      const char *host;
+
+      /* sanity check - remove any entries that have mismatched sizes,
+       * this might happen after an upgrade.
+       */
+
+      if (key.mv_size != sizeof(n.host) ||
+              data.mv_size != sizeof(struct timespec)) {
+
+          rc = mdb_cursor_del(cursor, 0);
+          if (rc) {
+            ERROR("mdb_cursor_del returned: %s (%d)", mdb_strerror(rc), rc);
+            mdb_cursor_close(cursor);
+            mdb_txn_abort(txn);
+            goto skip;
+          }
+
+          continue;
+      }
+
+      /* too old? if so, we've lost the host.
+       *
+       * also - stay silent for the first delay_interval after startup so we
+       * don't send a flurry of stale alerts.
+       *
+       */
+
+      host = key.mv_data;
+      tv_then = data.mv_data;
+
+      if ((tv_now.tv_sec - tv_then->tv_sec >= lost_interval) &&
+              (tv_now.tv_sec - tv_begin.tv_sec >= delay_interval)) {
+
+        notification_t n;
+        char message[NOTIF_MAX_MSG_LEN];
+
+        rc = mdb_cursor_del(cursor, 0);
+        if (rc) {
+          ERROR("mdb_cursor_del returned: %s (%d)", mdb_strerror(rc), rc);
+          mdb_cursor_close(cursor);
+          mdb_txn_abort(txn);
+          goto skip;
+        }
+
+        ssnprintf(message, sizeof(message), "Host not seen for %d seconds: %.*s",
+                (int)(tv_now.tv_sec - tv_then->tv_sec), (int)sizeof(n.host), host);
+        notification_init(&n, NOTIF_FAILURE, message, host, PLUGIN_NAME,
+                          NULL, "host", "lost");
+        n.time = cdtime();
+
+        plugin_dispatch_notification(&n);
+
+      }
+
+    }
+    mdb_cursor_close(cursor);
+
+    rc = mdb_txn_commit(txn);
+    if (rc) {
+      ERROR("mdb_txn_commit returned: %s (%d)", mdb_strerror(rc), rc);
+      goto skip;
+    }
+
+skip:
+    pthread_mutex_lock(&host_lock);
+
+    if (host_thread_loop <= 0)
+      break;
+
+    ts_wait.tv_sec = tv_now.tv_sec + thread_interval;
+    ts_wait.tv_nsec = 0;
+
+    pthread_cond_timedwait(&host_cond, &host_lock, &ts_wait);
+    if (host_thread_loop <= 0)
+      break;
+  } /* while (host_thread_loop > 0) */
+
+  pthread_mutex_unlock(&host_lock);
+
+  return (void *)0;
+} /* }}} void *host_thread */
+
+static int start_thread(void) /* {{{ */
+{
+  int status;
+
+  pthread_mutex_lock(&host_lock);
+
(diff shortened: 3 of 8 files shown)


rpv-tomsk:
Collectd code was always implemented to use minimal amount of resources to reach the goal.
As I see, your implementation has too much overkills.
Instead of using liblmdb you can/should use internal structures such as `avltree`. 
Recent version of Collectd has `check_uptime` plugin which does similar (but not the same) task and uses cache_events instead of handling each metric write.

Consider looking on it.

minfrin:
Looking at avltree, this looks like an in memory cache with no persistence across restarts, is that correct? If this is so this is not useful to us, as we have to detect hosts that have disappeared during times when collectd is not running, for example during restarts of collectd or reboots of the machine running collectd. The notifications sent by the host plugin will get engineers out of bed, so false messages cannot be sent.

The choice of lmdb was based on performance - it has extremely cheap reads with no locking, which is what most calls to write will result in - a single read and exit. I didn't see any formal persist-to-disk key/value store available in collectd, but happy to use it if available and preferred.

Not coming up with anything about check_uptime in the docs, https://collectd.org/wiki/index.php?search=check_uptime comes up blank. Looking in the source, I see the check_uptime wants to track uptime of a machine. The host plugin has no interest in uptime, all it cares about is whether a host is visible to collectd and sending data, any data. Whether the machine went down, resetting uptime, or whether a network cable got unplugged for 20 seconds, both scenarios need to trigger a notification.

Our current (cloud) monitoring tells us a machine is gone after at least ten minutes, which is largely useless. Collectd and the host plugin can tell us if a machine has gone after ten seconds, which is a vast improvement.

rpv-tomsk:
>Our current (cloud) monitoring tells us a machine is gone after at least ten minutes, which is largely useless. 
>Collectd and the host plugin can tell us if a machine has gone after ten seconds, which is a vast improvement.

Such statements make me wonder. Anyway, these notifications went to _people_. 
From my POV, there is no real difference, if that notification will be read after 5-10 minutes or after 10 seconds.

If you critically depend on "10 seconds" I have bad news for you.
And I feel sorry for the people who have to respond/react in these 10 seconds.

rpv-tomsk:
>*The host plugin has no interest in uptime*, [cut] 
>Whether the machine went down, *resetting uptime*, or [cut], both scenarios need to trigger a notification.

So you have already decided whether you need to handle uptime value or not? 
: ))))

rpv-tomsk:
>Looking at avltree, this looks like an in memory cache with no persistence across restarts, is that correct?

Yes, there is no persistence. But your plugin documentation also does not explains why somebody has to configure <HostPath path/to/database>.

rpv-tomsk:
1) Look on https://github.com/collectd/collectd/blob/master/docs/review_comments.md for variables declaration and initialization.
2) Your code should use "fast return". Get invalid cases out of the way first, then add the "real" body of the method. That is much easier to read.
3) Carefully review and polish your code. The code is still too raw.
4) "host" plugin name is too common. Also, some variables should be renamed too, to be more intuitive.

minfrin:
> Such statements make me wonder. Anyway, these notifications went to _people_.
> From my POV, there is no real difference, if that notification will be read after 5-10 minutes or after 10 seconds.
> 
> If you critically depend on "10 seconds" I have bad news for you.
> And I feel sorry for the people who have to respond/react in these 10 seconds.

I don't understand this comment?

We are not a startup. We have service level agreements with large clients, and there are penalties for downtime. A system that waits 10 to 20 minutes to tell us a machine is down (existing cloud solution), or a system that takes 10 seconds to tell us a machine is down (collectd with this plugin) gives our people an extra 10 to 20 minutes to react to an issue and that is exactly what our support people want.

Can you explain in more detail what you meant by "critically depend on 10 seconds"?

> So you have already decided whether you need to handle uptime value or not?
> : ))))

Again, I am not understanding this comment.

Uptime to me is the uptime value that a box announces as it's uptime, ie the equivalent of SNMP OID sysUpTime / 1.3.6.1.2.1.1.3. Are you talking about the same thing?

Most specifically I am not interested in sysUpTime, I am interested in a box going silent, for any reason. The host plugin will run on a machine that collates collectd messages from many hosts. As soon as a host goes silent, we want to know about it.

> Yes, there is no persistence. But your plugin documentation also does not explains
> why somebody has to configure <HostPath path/to/database>.

I have updated the comment in 0ca4713, can you confirm this makes more sense?

minfrin:
> Look on https://github.com/collectd/collectd/blob/master/docs/review_comments.md
> for variables declaration and initialization.

I put a lot of effort into matching the style of the existing code, using the ping plugin as a template. I'm assuming this is a new initiative? I am happy to change it - can you verify that 
8dcaf65 works for you?

minfrin:
> Your code should use "fast return". Get invalid cases out of the way first, then add
> the "real" body of the method. That is much easier to read.

Currently there are no "invalid" cases, the code assumes that all values of vl->host are valid. I am assuming this is what may have tripped you up.

host_write is then split into two paths, the fast path, and then the slow path.

The fast path involves doing a read to check if our host already exists, and whether we saw the host in the same second as stored in the kv store. 99.9% of all cases will be matched and bypassed here. LMDB is lock free on all read only transactions, and is as a result extremely fast. We can discard almost all checks without locking at the first step.

As soon as the host has been seen in a different second to the one stored, or if the get picked up nothing, we go down the slower write path. This path uses a write to update the second in which we last saw the host, and given we only ever do this once per second, this is very cheap.

I have added explanatory comments in f7798e6, can you confirm if this makes sense?

minfrin:
> Carefully review and polish your code. The code is still too raw.

Can you be specific on what requires polishing?

A significant amount of time and effort has been spent making this code match the existing code, both in terms of style, density of formatting, etc. If you would like something different I am happy to change it.

minfrin:
> "host" plugin name is too common. Also, some variables should be renamed too, to
> be more intuitive.

I struggled with the name of this plugin. It was named for the vl->host variable that we keep track of.

Perhaps the "presence" plugin? Happy to rename it to anything you suggest.

Do you have specific examples of variables that should be renamed? They were chosen to match the ping plugin, but given that plugin doesn't match the required style it was probably a poor choice. Again, happy to change anything you need changing.

minfrin:
> Even if you would do all these, that does not warrant your plugin will be accepted [at all/fast].

Is there a mechanism by which plugins can be built externally to collectd? Both the plugin.h and utils/common/common.h headers are internal. Forking the entire codebase for a single plugin is not ideal, and we would like to avoid anything that would make end user's lives difficult.

rpv-tomsk:
> gives our people an extra 10 to 20 minutes to react to an issue and that is exactly what our support people want.

I don't see any practical difference between "10 extra minutes" and "9 extra minutes".

If your hosts fails so often then you did something wrong.

You wrote:
>Collectd and the host plugin can tell us if a machine has gone **after ten seconds**, which is a vast improvement.

Again: I feel sorry for the people who have to respond/react in these 10 seconds.  This is a very nervous and stressful job.


>I am not interested in sysUpTime

Then you can lose some sporadic reboots.

>I have updated the comment in 0ca4713, can you confirm this makes more sense?

That is much better now.


>I put a lot of effort into matching the style of the existing code, using the ping plugin as a template.
>A significant amount of time and effort has been spent making this code match the existing code

I already found these plugins correlate much.
But existing Collectd code is not ideal too.


>> Your code should use "fast return". 

>Currently there are no "invalid" cases


That is about code style. Code can be readable better.
 
https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

>Can you be specific on what requires polishing?

You want to publish your plugin "just to be included upstream" or for real people?
`HostPath "hosts"` - not intuitive naming. Do you know about `/etc/hosts` file? 
`LostInterval 30`  - that is really interval? Value in collectd.conf.in does not match default in code.

`DelayInterval` - is missing in `collectd.conf.in`. Do you really think option name is intuitive? This is about delay between "host down" detected and "notification sent", isn't it?

The same is about variables naming.

> `if (tv_now.tv_sec - tv_then->tv_sec >= lost_interval + thread_interval) {`

Your business at huge risk!! 20% error!!!
This code will send notification only after 12 seconds, not after expected 10 !!!!

This bug only emphasizes that you really has no need in such accuracy.

>As soon as the host has been seen in a different second to the one stored, 

You don’t really need that kind of accuracy.
Common solution is update at 1/2 of threshold interval.

If your check threshold is 60 seconds, it's enough to update once per 30 seconds. 

So, you can reduce locking/writes 5 times for your "10 seconds" configuration (update LMDB for each host only once per 5 seconds), 30 times for "60 seconds" configuration and 150 times for configuration with threshold of 300 sec, etc.

>Perhaps the "presence" plugin? Happy to rename it to anything you suggest.

"presence" is also too common. I think something like "check_host_presence" will be more specific and intuitive.

>Is there a mechanism by which plugins can be built externally to collectd?

I don't know about solution. But you can find it or propose yours.

>Update the defaults to match collectd's native interval. 

VERY strange action. That should not work good in real environments. 
Your threshold must be greater twice than native interval, at least.
Please, just return 60 seconds value back and do not put nothing you have not tested enough.

minfrin:
> I don't see any practical difference between "10 extra minutes" and "9 extra minutes".
> 
> If your hosts fails so often then you did something wrong.

> Again: I feel sorry for the people who have to respond/react in these 10 seconds. This
> is a very nervous and stressful job.

Our people have no need to respond/react in 10 seconds, and our hosts do not fail every 10 minutes, I suspect we have misunderstood each other on the core problem. I will explain the problem in detail.

We have a 2 hour SLA. If we are down for 2 hours or more, our customers get their money back.

In the cloud, we have no visibility over how our machines are physically provisioned. When physical hosts fail, our virtual hosts go dark without warning. This does not happen often, but it does happen, once every few months. The cloud provider provides a monitoring solution that will tell us when a box is down, but this solution takes between 10 and 20 minutes to tell us. This means that of the 120 minutes our people have available to investigate and fix, 10 to 20 minutes of that time is wasted.

We also deploy collectd, and all hosts pass stats to the monitoring machines continuously. If a machine disappears for any reason, collectd will know about it within 10 seconds. This means that of the 120 minutes our people have available to investigate and fix, just 10 seconds is wasted, which is no time at all.

The difference comes about because the cloud based solution is based on polling, while collectd is based on streaming. For this reason detecting the disappearance of a host is far more accurately done with collectd, and therefore this plugin.

minfrin:
> Then you can lose some sporadic reboots.

By design, because we persist our hosts list, we will catch all reboots, except for ones that happen completely within the time collectd itself was restarting. We accept this limitation.

> That is about code style. Code can be readable better.
> 
> https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

Unfortunately that topic on stackexchange was closed for being an opinion.

The code paths are not just simple "every non zero response is an error". We have to react to and handle result codes like MDB_NOTFOUND, and we have to make sure the monitoring thread does not exit until explicitly requested. You'll see a similar structure in the ping plugin.

> You want to publish your plugin "just to be included upstream" or for real people?

Real people.

minfrin:
> HostPath "hosts" - not intuitive naming. Do you know about /etc/hosts file?
> LostInterval 30 - that is really interval? Value in collectd.conf.in does not match default in code.
> 
> DelayInterval - is missing in collectd.conf.in. Do you really think option name is intuitive? This
> is about delay between "host down" detected and "notification sent", isn't it?

These names were chosen to match the brevity of the existing directives, but am happy to make them longer if you would like them to be different. Over at apache/httpd we typically prefix directives with the name of the module, but I notice collectd doesn't do that, nor does collectd seem to support directives outside of <Plugin> scope.

The HostPath is the path to the host datastore, which by default is called hosts in the /var/lib/collectd directory. Obviously datastores don't go into /etc, so /etc/hosts won't be a problem. Would you prefer something like CheckPresencePath?

LostInterval is indeed an interval. It is the time beyond which a host is considered lost. Do you recommend a different terminology for this?

DelayInterval is the delay we introduce after startup so that we give hosts a window of time to check back in and remind us they're still there. Do you have an improved suggestion for this one?

rpv-tomsk:
StateDatabase (PersistentXXXX)
StartupDelay  (StartupGrace ?)
HostTimeout

rpv-tomsk:
>You'll see a similar structure in the ping plugin.
I'm absolutely not interested in "similar structure" in something.

As I already said, existing code is not ideal. But new code should be.

minfrin:
> > if (tv_now.tv_sec - tv_then->tv_sec >= lost_interval + thread_interval) {
> Your business at huge risk!! 20% error!!!
> This code will send notification only after 12 seconds, not after expected 10 !!!!
> 
> This bug only emphasizes that you really has no need in such accuracy.

This is not a bug, the behaviour is intended.

As you test the code, comment out the thread_interval, and see what happens - you now have a race condition where the code detects the host is missing, sends a notification and removes the host, and then almost immediately the host is detected, a notification is sent that the host is back, and the host is stored again. This happens every ten seconds. The 2 second delay covers the 1 second interval during which we ignore hosts we have seen in the lasts second, and an extra second for safety. This works great.

As to "Your business at huge risk!! 20% error!!!", I have already explained how our people don't have to respond in 10 seconds, so there is no business risk.

> > As soon as the host has been seen in a different second to the one stored,
> You don’t really need that kind of accuracy.
> Common solution is update at 1/2 of threshold interval.

As explained above this is not about accuracy, but rather about avoiding a race condition.

rpv-tomsk:
>This works great.

But not match documentation. 

I can setup configuration where will be no races even having `ReadInterval 10`  and `HostTimeout 10`.
Just send more metrics from your hosts and they will be some kind of "normally distributed" - i.e will come more than once every 10 seconds - that will be enough to race not occur.

If you need "works great" for your setup when your hosts sent small amount of metrics, then setup that explicitly in config, by declaring `HostTimeout` to 12 seconds.

However, that has nothing good to sent notification on _one_ lost metric, ie declaring `HostTimeout` equal to `ReadInterval`. It should be at least twice greater to be _confirmed state_, not accidental.

minfrin:
> > Perhaps the "presence" plugin? Happy to rename it to anything you suggest.
> "presence" is also too common. I think something like "check_host_presence" will
> be more specific and intuitive.

Almost all plugins are one word names, with a few plugins with prefixes like "write" that are two word names. I would prefer "check_presence", as it fits better with the existing code, and follows your check_uptime plugin.

Will update.

rpv-tomsk:
>Would you prefer something like CheckPresencePath?

```
<Plugin host>
    CheckPresencePath "hosts"
    LostInterval 30
    DelayInterval 30
</Plugin>
```

Imagine seeing the configuration for the first time.
Please answer: "How presence is checked using that path?" 
There will be no fast answer.

Then compare that with:

```
<Plugin check_host_presence>
    StateDatabase "/var/lib/collectd/hosts"
    StartupDelay 60
    HostTimeout 60
</Plugin>
```

Just looking here, you know:
- plugin checks host presence
- it uses state database in some file
- it has startup delay

HostTimeout is not enough intuitive, and possibly can be improved.

From my POV, plugin development should be started from designing its configuration and options.
That is people will use.

rpv-tomsk:
>I would prefer "check_presence",

One more proposal: `check_hostpresence`. )

src/host.c
```diff
@@ -0,0 +1,519 @@
...
+static pthread_t host_thread_id;
+
+static MDB_env *env;
+
+static void *host_thread(void *arg) /* {{{ */
+{
+  struct timeval tv_begin = {0};
+
+  if (gettimeofday(&tv_begin, NULL) < 0) {
+    ERROR("host plugin: gettimeofday failed: %s", STRERRNO);
```
rpv-tomsk: Here you added prefix to messages, later there are many cases w/o prefix.

minfrin:
> > Update the defaults to match collectd's native interval.
> VERY strange action. That should not work good in real environments.
> Your threshold must be greater twice than native interval, at least.
> Please, just return 60 seconds value back and do not put nothing you have not tested enough.

All testing was done by explicitly setting the values to match collectd's native interval of 10, the defaults were not used. The value of 60 was from an earlier bug that was fixed.

A host will be been last seen between 0 and 10 seconds ago, and each host sends multiple values during that period, not just a single ping. For that reason 10 plus the 2 seconds to cover the race condition covers the full range.

Happy to set it to 20, but the way the code works right now this isn't necessary.

rpv-tomsk:
You probably missed my comment:

> >This works great.

>But not match documentation. 

>>This is not a bug, the behaviour is intended.

>When something does not match documentation - that is a bug.

minfrin:
> StateDatabase (PersistentXXXX)
> StartupDelay (StartupGrace ?)
> HostTimeout

I would rather go for datastore rather than database, but these work:

StateDatastore
StartupDelay
HostTimeout

Lunch calls, then I'll send an update.

minfrin:
> From my POV, plugin development should be started from designing its
> configuration and options.
> That is people will use.
> The more simple it is, the less questions will be asked, and the more often it will be used.

Agreed, and that was how the plugin was designed. The config directives made perfect sense to me, however I'm happy to change them.

Can you review a9054d6 for me?

minfrin:
> As I already said, existing code is not ideal. But new code should be.
> 
> The such is completely unacceptable here.

Thank you for editing the comment to highlight where the issue lay. Would it be possible to use inline comments to highlight these issues, so that comments are placed in context.

I have updated the code to match your preferred style, can you review 1ab5f2e for me?

minfrin:
> I can setup configuration where will be no races even having ReadInterval 10 and HostTimeout 10.
> Just send more metrics from your hosts and they will be some kind of "normally distributed" - i.e > will come more than once every 10 seconds - that will be enough to race not occur.

In collectd's default configuration, this is not the case, which is why the race condition was discovered. We are very careful not to send false notifications, which is why this is a big deal for us.

> If you need "works great" for your setup when your hosts sent small amount of metrics,
> then setup that explicitly in config, by declaring HostTimeout to 12 seconds. Be honest
> with yourself.

What this means is that our end users have to know and understand that the HostTimeout needs to be 2 seconds bigger than Interval, and it is very likely people aren't going to care about this distinction and just make them equal. This is intentionally done to prevent bugs being raised and then closed as not-a-bug.

> However, that has nothing good to sent notification on one lost metric, ie declaring
> HostTimeout equal to ReadInterval. It should be at least twice greater to be confirmed
> state, not accidental.

As I said, happy to make it 20. Can you verify e60cd94 for me?

src/host.c
minfrin: I have made them consistent in dcd1bb5, can you verify for me?

minfrin:
> >I would prefer "check_presence",
> One more proposal: check_hostpresence. )

That option was still too long, it looked really out of place.

I have renamed it to check_presence across the board, can you verify 22a6fef for me?

src/check_presence.c
```diff
@@ -0,0 +1,533 @@
...
+    return 0;
+  } else
+    return -1;
+} /* int presence_config */
+
+void module_register(void) {
+  plugin_register_init(PLUGIN_NAME, presence_init);
+  plugin_register_config(PLUGIN_NAME, presence_config, config_keys,
+                         config_keys_num);
+  plugin_register_write(PLUGIN_NAME, presence_write, NULL);
```
rpv-tomsk: I think that is possible to move `register_write()` (and possibly `register_shutdown()`) into init phase and do not register it if not configured.
This will not waste CPU time.

src/check_presence.c
```diff
@@ -0,0 +1,533 @@
...
+  }
+
+  pthread_mutex_lock(&presence_lock);
+  memset(&presence_thread_id, 0, sizeof(presence_thread_id));
+  pthread_mutex_unlock(&presence_lock);
+
+  return status;
+} /* }}} int stop_thread */
+
+static int presence_init(void) {
```
rpv-tomsk: There is no way to check if your plugin working or not (i.e is configured or not) w/o doing "active tests".
I think that would be good to notify user "plugin is not configured and will not do any checks /send any notifications" (please rephrase!!!) if `state_datastore == NULL`.

src/check_presence.c
```diff
@@ -0,0 +1,533 @@
...
+  } else {
+
+    /* existing database, do the get */
+
+    rc = mdb_get(txn, dbi, &key, &data);
+  }
+
+  int add = 0;
+
+  if (MDB_NOTFOUND == rc) {
```
rpv-tomsk: This code can be much more readable if you move up `add = 1` into L377 and
move block  `mdb_get failed` upper, below `mdb_get()`  call.

Then there will be no need in this second `if` construct.

src/collectd.conf.pod
```diff
@@ -1548,6 +1548,39 @@ at all, B<all> cgroups are selected.
...
+
+The B<check_presence> plugin keeps track of hosts that have been recently seen,
+and sends notifications when a host is seen for the first time, and subsequently
+when a host has gone missing after not being seen for a period of time. The
+duration after which a host is considered missing is controlled with the
+B<HostTimeout> option.
+
+To ensure that no notifications are missed or duplicated, the hosts are
+persisted in an LMDB key/value store, which is kept across collectd restarts.
+The hosts plugin provides a grace period after starting up for hosts to tell
```
rpv-tomsk: Old plugin name.

src/collectd.conf.pod
```diff
@@ -1548,6 +1548,39 @@ at all, B<all> cgroups are selected.
...
+automatically to ensure there are no race conditions where a host might be
+declared missing but then immediately found.
+
+=over 4
+
+=item B<StateDatastore> I<path/to/database>
+
+Path and filename of the LMDB hosts key/value store. Defaults to B<check-presence>.
+
+=item B<HostTimeout> I<Seconds>
```
rpv-tomsk: Option description is missing. Please add it.

src/collectd.conf.pod
```diff
@@ -1548,6 +1548,39 @@ at all, B<all> cgroups are selected.
...
+
+=item B<StartupDelay> I<Seconds>
+
+Number of seconds after startup to suppress missing notifications. Defaults to B<20>,
+which is twice the value of Interval.
```
rpv-tomsk: "which is twice the value of Interval" - needs to be rephrased or removed or code reworked.

Your defaults is hardcoded and do not depend on `Interval` configured value.
Do not fool people - This option description is ambiguous and may be misunderstood..

src/check_presence.c
```diff
@@ -0,0 +1,533 @@
...
+
+  } else {
+
+    /* have we seen host in the same second as we did our last put? if so
+     * it is good enough, exit cheaply
+     */
+
+    struct timeval *tv_then = data.mv_data;
+
+    if (data.mv_size == sizeof(struct timespec) && ((tv_then = data.mv_data)) &&
```
rpv-tomsk: You don’t really need that kind of accuracy.
Common solution is update at 1/2 of threshold interval.

If your check threshold is 60 seconds, it's enough to update once per 30 seconds.

So, you can reduce locking/writes 5 times for your "10 seconds" configuration (update LMDB for each host only once per 5 seconds), 30 times for "60 seconds" configuration and 150 times for configuration with threshold of 300 sec, etc.

It is unfortunate that you ignored this comment in the comments.

src/check_presence.c
```diff
@@ -0,0 +1,533 @@
...
+#include "plugin.h"
+#include "utils/common/common.h"
+
+#include <lmdb.h>
+
+#define PLUGIN_NAME "check_presence"
+
+#define DEFAULT_STATE_DATASTORE "check-presence"
+#define DEFAULT_HOST_TIMEOUT 20
+#define DEFAULT_THREAD_INTERVAL 2
```
rpv-tomsk: There is absolutely no need in scanning database each 2 seconds if your `HostTimeout 60`.
That is enough, say, to do such scan once per 10 seconds, then you will get notifications for hosts unreachable for 60-70seconds with average value of 65 seconds. This will reduce cpu usage 5 times.

I.e propose to made this configurable with default value of 1/2 of `HostTimeout`.

And of course, remove `thread_interval` from "avoid race conditions" code.

src/check_presence.c
```diff
@@ -0,0 +1,533 @@
...
+
+  if (presence_thread_loop != 0) {
+    pthread_mutex_unlock(&presence_lock);
+    return 0;
+  }
+
+  presence_thread_loop = 1;
+  int status = plugin_thread_create(&presence_thread_id, /* attr = */ NULL,
+                                    presence_thread,
+                                    /* arg = */ (void *)0, "host");
```
rpv-tomsk: Outdated plugin name.

src/collectd.conf.pod
```diff
@@ -1548,6 +1548,39 @@ at all, B<all> cgroups are selected.
...
+duration after which a host is considered missing is controlled with the
+B<HostTimeout> option.
+
+To ensure that no notifications are missed or duplicated, the hosts are
+persisted in an LMDB key/value store, which is kept across collectd restarts.
+The hosts plugin provides a grace period after starting up for hosts to tell
+us they are still present, catching hosts that have gone missing during our
+restart. The grace period is controlled with the B<StartupDelay> option.
+
+An additional two seconds are added on top of B<HostTimeout> and B<StartupDelay>
```
rpv-tomsk: This is a crutch. Please remove it from code and documentation.

contrib/redhat/collectd.spec
```diff
@@ -2026,6 +2043,7 @@ Collectd utilities
 	%{?_with_bind} \
 	%{?_with_ceph} \
 	%{?_with_cgroups} \
+    %{?_with_check_presence} \
```
rpv-tomsk: This line looks unaligned.

src/check_presence.c
```diff
@@ -0,0 +1,533 @@
...
+      /* too old? if so, we've lost the host.
+       *
+       * also - stay silent for the first delay_interval after startup so we
+       * don't send a flurry of stale alerts.
+       *
+       */
+
+      const char *host = key.mv_data;
+
+      struct timeval *tv_then = {0};
```
rpv-tomsk: Why do not merge these two lines into one?

src/check_presence.c
rpv-tomsk: Something strange I see here in `&& ((tv_then = data.mv_data))`.

Can you please carefully and critically review your code yourself?

minfrin: It'll be another week or two before I can work on this code again.

In the mean time can you fully describe what "something strange" means on that line?

In C a single "=" is an assignment operator, and in this case we don't dare try to assign the variable until we've checked the length. This is a defensive technique - don't assign possible invalid values to variables, not even temporarily.

The double brackets are a signal to the compiler that the assignment is intended.

dago:
@minfrin Would you mind reviewing the comments from the code review?

mrunge:
@minfrin Hi, would you have the cycles to get back to this?

mrunge:
Friendly ping here?

minfrin:
Drowning in bugs right now.

This work is now blocking work on one of the bugfixes, and git's handling of forks is awful. Let me look at this now.

src/check_presence.c
minfrin: Fixed.

For the future, don't assume the person isn't word blind to the thing you see. @dago 's review highlighted the problem.

src/check_presence.c
minfrin: Assuming you're asking for them to be merged - done.

contrib/redhat/collectd.spec
minfrin: It was indeed - fixed.

src/check_presence.c
minfrin: Fixed.

src/collectd.conf.pod
minfrin: Added.

src/collectd.conf.pod
minfrin: Fixed.

octo:
Thanks for the update, @minfrin!

@mrunge and @dago, does one of you have the time to do a review?

src/collectd.conf.pod
minfrin: The race condition protection has been removed, and the the limit on values documented.

src/check_presence.c
minfrin: CheckInterval added as recommended.

src/check_presence.c
minfrin: The network guys who wanted this disagreed.

Your suggestion of a CheckInterval option below resolves this - it will allow people who need accuracy to get it, while allowing people who don't want it to not.

Implemented as below.

src/collectd.conf.pod
minfrin: The phrase has been removed.

src/check_presence.c
minfrin: This suggestion breaks the first pass through this code.

There are three possible code paths:

- The very first time we see a host, we need to ignore the failure mdb_dbi_open(), and skip the get of mdb_get().
- All other times, we must mdb_dbi_open() and then mdb_get().
- An error has occurred.

In addition, code above "int add = 0;" is about open/get, while the code below is about whether we follow the fast path and skip the writing of the host.

These phases are kept separate from one another.

src/check_presence.c
minfrin: True - good to see this is an option.

Done.

src/check_presence.c
minfrin: Done.

minfrin:
All updated, appreciate the quick response.

minfrin:
Quick ping, any news?

minfrin:
Another quick ping @octo, any news?

minfrin:
Hi @octo @mrunge and @dago, happy new year!

Is it possible to take a look at this one?

contrib/redhat/collectd.spec
```diff
@@ -2084,7 +2100,7 @@ Collectd utilities
 	%{?_with_capabilities} \
 	%{?_with_ceph} \
 	%{?_with_cgroups} \
-	%{?_with_check_uptime} \
+	%{?_with_check_presence} \
```
dago, line 2103: Why is check_uptime removed here?

dago (commented):
Looks good to me

dago:
Looks like the CI is blocked again, @mrunge would you mind having a look please?

mrunge:
centos 6 CI test should be get retired asap, see https://github.com/collectd/collectd/pull/3784 Unfortunately, that is blocked by travis (not responding anymore). I am out of powers here :-/

contrib/redhat/collectd.spec
minfrin: That was an edit that should have been a duplicate line, I've just fixed this one.

dago:
I'd say this is ready to merge when the CI is unbroken again.

minfrin:
Any news on the CI?

I have other patches that are also blocked behind CI failures that seem unrelated to the changes being made.

PR #3359, opened 2019-12-19
Factor out read_text_file_contents to ensure that data read from text files is a valid C string.

igorpeshansky:
Also fixes a potential unterminated string in `thermal_procfs_device_read`.

ChangeLog: collectd: Factored out read_text_file_contents for reading text files

 and used it to fix a potential unterminated string in the `thermal` plugin.

/cc @jkohen 

[edit @mrunge]: Broke the ChangeLog line, since CI fails on long changelog lines.

### DIFF
--- a/src/battery.c
+++ b/src/battery.c
@@ -349,11 +349,9 @@ static int sysfs_file_to_buffer(char const *dir, /* {{{ */
 
   snprintf(filename, sizeof(filename), "%s/%s/%s", dir, power_supply, basename);
 
-  status = (int)read_file_contents(filename, buffer, buffer_size - 1);
+  status = (int)read_text_file_contents(filename, buffer, buffer_size);
   if (status < 0)
     return status;
-
-  buffer[status] = '\0';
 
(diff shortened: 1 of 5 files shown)


igorpeshansky:
Note to the maintainers: this is a semantics-preserving refactoring, except for the potential bug fix in the `thermal` plugin. Since the bug hasn't been reported, AFAICS, I'm not sure what would be the value of adding it to the release notes, but I'm happy to add a ChangeLog line if necessary. Otherwise please tag as `Unlisted Change`.

mrunge:
Thank you!

The ChangeLog feature helps us with writing a changelog, which is also about to highlight and thank contributors.

If you don't want to be included there, I'm happy to flag this as unlisted feature. However, I'd prefer to add and list your contribution :) (I'd leave the ChangeLog question to you...)

igorpeshansky:
@mrunge That's fair. I assumed the ChangeLog was for user-visible changes only... Done.

This is ready for review.

PR #3360, opened 2019-12-23
Support for Varnish 6 and remove Varnish 2, 3 and 5

ismaelpuerto:
ChangeLog: varnish plugin: Support for Varnish 6, deprecated 2, 3 and 5.
ChangeLog: varnish plugin: Added counters for vmod-goto.

I keep Varnish 4 because Varnish Software keep support for Varnish Plus 4. Deleted Varnish 2, 3 and 5 because does't maintened: https://varnish-cache.org/releases/

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5908,23 +5908,13 @@ if test "x$with_libvarnish" = "xyes"; then
   SAVE_CPPFLAGS="$CPPFLAGS"
   CPPFLAGS="$CPPFLAGS $with_libvarnish_cflags"
 
-  $PKG_CONFIG --atleast-version=5.2 'varnishapi' 2>/dev/null
+  $PKG_CONFIG --atleast-version=6 'varnishapi' 2>/dev/null
   if test $? -eq 0; then
-    AC_DEFINE([HAVE_VARNISH_V5], [1], [Varnish 5 API support])
+    AC_DEFINE([HAVE_VARNISH_V6], [1], [Varnish 6 API support])
   else
     AC_CHECK_HEADERS([vapi/vsc.h],
       [AC_DEFINE([HAVE_VARNISH_V4], [1], [Varnish 4 API support])],
-      [
-        AC_CHECK_HEADERS([vsc.h],
-          [AC_DEFINE([HAVE_VARNISH_V3], [1], [Varnish 3 API support]) ],
-          [
-            AC_CHECK_HEADERS([varnishapi.h],
-              [AC_DEFINE([HAVE_VARNISH_V2], [1], [Varnish 2 API support])],
-              [with_libvarnish="no (found none of the varnish header files)"]
-            )
-          ]
-        )
-      ]
+      [with_libvarnish="no (found none of the varnish header files)"]
     )
   fi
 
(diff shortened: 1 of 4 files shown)


dago:
@ismaelpuerto There definitely a failing CI test. Would you mind having a look?

ismaelpuerto:
Hello @dago, fail test of "trusty" because use old version of varnish, Ubuntu trusty born at 2014. Technically, trusty is a version not supported without subscription. https://wiki.ubuntu.com/Releases

dago:
@ismaelpuerto I see, I am a bit reluctant to throw out support for old versions when there is the possibility some it is still used in the field. However, I could imagine that people using varnish on platforms will probably run newer OS versions and people with older OSes probably do so for other datacenter non-web stuff.

mrunge:
I'd be good with dropping unsupported versions.

dago:
@mrunge I am now merging this, but we need to take out building on "trusty" then. Can you tell me where this can be changed in the CI?

kkepka:
if you would like to remove building on trusty at all, then quickshot would be on [L38-40 in .cirrus.yml](https://github.com/collectd/collectd/blob/master/.cirrus.yml#L40), but i can double check it on side if you want

dago:
@kkepka We can continue to build on trusty, but then Varnish needs to be disabled. Would you be willing to adjust the configuration?

kkepka:
ok, just `--disable-varnish` there is not enough, as it fails on `/checks/check-built-plugins.sh`, will look for right place to disable test for varnish on trusty

dago:
@kkepka Can you please follow up on the CI issue with trusty? The broken CI makes it hard to see real build issues.

kkepka:
yes i can, getting into it right now

kkepka:
short backtrace of reason for https://github.com/collectd/ci-docker/pull/3:
- ci fails with missing varnish plugin on trusty
- failing script [check-built-plugins.sh](https://github.com/collectd/ci-docker/blob/trusty_amd64/checks/check-built-plugins.sh) uses `SUPPORTED_PLUGIN_LIST` env var to check if supported plugins have been built
- content of this var is set in [Dockerfile](https://github.com/collectd/ci-docker/blob/trusty_amd64/Dockerfile) during ci container image build by travis on per target branch (`trusty_amd64`) image/distro

rubenk:
Can we revert this for now, at least until we managed to merge 5.10 and 5.11 back into master and released 5.11?

mrunge:
@rubenk would you open a PR to revert or shall I?

rubenk:
@mrunge please do

mrunge:
Submitted https://github.com/collectd/collectd/pull/3412

rubenk:
Ubuntu 18.04 still has varnish 5 and is widely used (also for Travis). Can we keep support for Varnish 5 a little longer?

mrunge:
https://github.com/collectd/collectd/pull/3414 has all reverts minus the noop commit https://github.com/collectd/collectd/pull/3360/commits/039ccb92bca8950066531131db7133ac567e3769 from @dago

PR #3362, opened 2020-01-07
Add support for Interface regexps in ethstat plugin

iuliaandreeaoprea:
This change allows users to properly define and use wildcard based interface naming when using the ethstat plugin. You can also specify explicit interfaces, if this is desired.

Example:
```
<Plugin ethstat>
    Interface "/^ens[0-9]?$/"
    Interface "eth0"
    [...]
</Plugin>
```
ChangeLog: ethstat plugin: Add support for Interface regexps in ethstat plugin.

### DIFF
--- a/src/ethstat.c
+++ b/src/ethstat.c
@@ -26,8 +26,8 @@
 
 #include "plugin.h"
 #include "utils/avltree/avltree.h"
-#include "utils/ignorelist/ignorelist.h"
 #include "utils/common/common.h"
+#include "utils/ignorelist/ignorelist.h"
 #include "utils_complain.h"
 
 #if HAVE_SYS_IOCTL_H
@@ -62,11 +62,12 @@ static c_avl_tree_t *value_map;
 
 static bool collect_mapped_only;
 
-static int ethstat_check_if_existing_interface(char **existing_interfaces, char *entry) {
+static int ethstat_check_if_existing_interface(char **existing_interfaces,
+                                               char *entry) {
   char *aux;
   int idx;
 
-  for(idx = 0; idx < valid_interfaces_num; ++idx) {
+  for (idx = 0; idx < valid_interfaces_num; ++idx) {
     aux = strdup(existing_interfaces[idx]);
     if (strcmp(aux, entry) == 0)
       return 0;
@@ -91,7 +92,7 @@ static int ethstat_get_matching_interface(const char *entry) {
     return 1;
   }
 
-  #if HAVE_GETIFADDRS
+#if HAVE_GETIFADDRS
   ignorelist_add(ethstat_ignorelist, entry);
 
   if (getifaddrs(&if_list) != 0) {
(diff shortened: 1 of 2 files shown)


src/ethstat.c
```diff
@@ -41,38 +42,108 @@
...
 static bool collect_mapped_only;
 
+static int ethstat_check_if_existing_interface(char **existing_interfaces,
+                                               char *entry) {
+  char *aux;
+  int idx;
+
+  for (idx = 0; idx < valid_interfaces_num; ++idx) {
+    aux = strdup(existing_interfaces[idx]);
+    if (strcmp(aux, entry) == 0)
```
rpv-tomsk: It seems you like memory leaks much. I don't see why you need to do `strdup` at all.

src/ethstat.c
```diff
@@ -41,38 +42,108 @@
...
+}
+
+static int ethstat_get_matching_interface(const char *entry) {
+  size_t len;
+  struct ifaddrs *if_list = NULL;
+  char **v_tmp;
+  ignorelist_t *ethstat_ignorelist = NULL;
+
+  if (ethstat_ignorelist == NULL)
+    ethstat_ignorelist = ignorelist_create(/* invert = */ 1);
```
rpv-tomsk: `ignorelist_free` call is missing.

src/ethstat.c
```diff
@@ -41,38 +42,108 @@
...
+
+  len = strlen(entry);
+  if (len == 0) {
+    return 1;
+  }
+
+#if HAVE_GETIFADDRS
+  ignorelist_add(ethstat_ignorelist, entry);
+
+  if (getifaddrs(&if_list) != 0) {
```
rpv-tomsk: `freeifaddrs()` is missing for `if_list`.

rpv-tomsk:
Regarding your update:
Typically, Collectd plugin configuration supports changes in system which occur later after Collectd start. This plugin also support interfaces which could appear later. Your update introduces regression - if interface is missing on Collectd start, it will never be picked up. (Yes, current implementation will complain if configured interface is not found, but when interface will appear - it's stats will be submitted).

I think such work needs to be a bit deeper.

1) To avoid regression but without support of interfaces added later, you could check if option value is regex or not.
2) I think, preferred way of improving this plugin is to add dynamic update of interfaces list (i.e. do it once per ... minute etc). This allow to implement `IgnoreSelected` option too.  Let's discuss this.
3) You missed updating plugin documentation.

dago:
@iuliaandreeaoprea Would you mind resolving the issues brought up in the code review?

mrunge:
@iuliaandreeaoprea thank you for putting the code up for review. Would you be able to resolve the suggestions pointed out?

mrunge:
@iuliaandreeaoprea Hi, this is nearly there. Would you be able to address the comments above?

Issue #3322, opened 2019-10-25
exec plugin buffer overflow under certain conditions


kwiatrox:
*   Version of collectd: 5.9.2, also master
*   Operating system / distribution: Ubuntu 18.04
*   Kernel version (if applicable): N/A

## Expected behavior

Exec plugin can work with file descriptors bigger than 1024. This situation happen when limit for collectd is increased by `ulimit -n` and other plugin used all the smaller FDs (e.g. intel_pmu plugin).

## Actual behavior

Buffer overflow -> core dump
```
[2019-10-24 14:22:08] [info] Initialization complete, entering read-loop.
[2019-10-24 14:22:08] [debug] Created new plugin context.
[2019-10-24 14:22:08] [debug] intel_pmu: pmu_read:416
[2019-10-24 14:22:08] [debug] Created new plugin context.
[2019-10-24 14:22:08] [debug] plugin_read_thread: read-function of the `exec' plugin took 0.000256 seconds.
[2019-10-24 14:22:08] [debug] plugin_read_thread: Effective interval of the `exec' plugin is 10.000 seconds.
[2019-10-24 14:22:08] [debug] plugin_read_thread: Next read of the `exec' plugin at 1571919738.133.
[2019-10-24 14:22:08] [debug] Created new plugin context.
*** buffer overflow detected ***: /opt/collectd/sbin/collectd terminated
Aborted (core dumped)
```

Parts of core dump:
```
#4  0x00007f1ea8501d21 in __GI___fortify_fail (msg=msg@entry=0x7f1ea8583905 "buffer overflow detected") at fortify_fail.c:44
#5  0x00007f1ea84ffa10 in __GI___chk_fail () at chk_fail.c:28
#6  0x00007f1ea8501c0a in __fdelt_chk (d=<optimized out>) at fdelt_chk.c:25
#7  0x00007f1ea7fbdc25 in exec_read_one (arg=0x5564d9a89120) at src/exec.c:602
        __d = <optimized out>
        pl = 0x5564d9a89120
        fd = 2309
        fd_err = 2311
        __PRETTY_FUNCTION__ = "exec_read_one"
(...)
```

## Steps to reproduce

*   change ulimit
`ulimit -n 4096`
*   setup intel_pmu plugin to monitor all cores and events, assuming that
    "num_of_cores" * "num_of_events" >= 1024
*   enable exec plugin with at least one exec to run
*   start collectd

## Root cause

[...]

PR #3363, opened 2020-01-07
exec plugin: replace select with poll

kwiatrox:
ChangeLog: exec plugin: replace select with poll

Replace select with poll to allow big File Descriptors values on systems which allow more than 1024 open FDs. Other than that there is no any change in functionality.
This fixes issue #3322.

### DIFF
--- a/src/exec.c
+++ b/src/exec.c
@@ -38,6 +38,7 @@
 #include "utils/cmds/putval.h"
 
 #include <grp.h>
+#include <poll.h>
 #include <pwd.h>
 #include <signal.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


rubenk:
Thank you. Did you test this on all OS's we support?

kkepka:
I think verification of this patch may not been so comprehensive and broad in terms of multiple OS's, but will let @kwiatrox to let know about details where it has been checked.

Out of curiosity - what is actually the list of "all OS's we support"?
few of them are mentioned on https://collectd.org/features.shtml , like `Linux, Solaris, Mac OS X, AIX, FreeBSD, NetBSD, and OpenBSD`, but is there any more specific e.g. particular linux distributions and ver of releases?

kwiatrox:
> Thank you. Did you test this on all OS's we support?

I had a chance to test it only on Ubuntu so far.

sunkuranganath:
I would assume at this point we would go with CI checks. If CI passes and there are no pending review comments, I would think it can be merged.
As per our meetup discussions, exec plugin is part of "core plugins" list. So we require a release manager or core reviewers to accept this bug fix. Adding core reviewers (that I know of) to the review.
The sooner we accept this bug fix the better.

dago:
Ok, I'll merge this one, but we need to keep in mind that strict release testing on all supported platforms is done.

PR #3367, opened 2020-01-24
types.db: iwinfo: 'stations value:GAUGE:0:256'

westurner:
Add `stations value:GAUGE:0:256` to types.db

https://github.com/grafana/grafana/issues/7425 mentions that this is necessary for the `collectd-mod-iwinfo` that ships with OpenWRT.


rpv-tomsk:
That should be done by maintainers of OpenWRT/collectd-mod-iwinfo/etc.

Collectd upstream code has no metrics named `stations`, so it should not have such type in `types.db`.
Upstream `types.db` could not cover all world use-cases.

dago:
Closing as this is purely downstream-specific.

PR #3376, opened 2020-02-18
Efuss fix 3313

dago:
ChangeLog: build system: Avoid $< (implied source) in non-inference rules


dago:
This is the rebase of #3330

PR #3377, opened 2020-02-18
Fix of "Implement the NFS plugin for NetBSD #3333"

dago:
ChangeLog: NFS plugin: An implementation for NetBSD (using sysctl) has been added.

### DIFF
--- a/src/nfs.c
+++ b/src/nfs.c
@@ -27,6 +27,15 @@
 #include "plugin.h"
 #include "utils/common/common.h"
 
+#if KERNEL_NETBSD
+#include <sys/param.h>
+#include <sys/mount.h>
+#include <sys/sysctl.h>
+#include <nfs/rpcv2.h>
+#include <nfs/nfsproto.h>
+#include <nfs/nfs.h>
+#endif
+
 #if HAVE_KSTAT_H
 #include <kstat.h>
 #endif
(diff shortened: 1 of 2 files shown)


src/nfs.c
```diff
@@ -596,6 +612,34 @@ static int nfs_read(void) {
 }
   /* #endif KERNEL_LINUX */
 
+#elif KERNEL_NETBSD
+static int nfs_read(void) {
+  struct nfsstats ns;
+  size_t size = sizeof(ns);
+  int mib[] = {CTL_VFS, 2, NFS_NFSSTATS};
```
kwiatrox: Might add a comment about the meaning of 2.

src/nfs.c
```diff
@@ -596,6 +612,34 @@ static int nfs_read(void) {
...
+  value_t values[nfs3_procedures_names_num];
+  int i;
+
+  /* NetBSD reports v2 statistics mapped to v3 and doen't yet support v4 */
+  if (!report_v3)
+    return 0;
+  if (sysctl(mib, 3, &ns, &size, NULL, 0) != 0)
```
kwiatrox: I would use STATIC_ARRAY_SIZE(mib) for length to avoid magic numbers.

src/nfs.c
```diff
@@ -596,6 +612,34 @@ static int nfs_read(void) {
...
 
+#elif KERNEL_NETBSD
+static int nfs_read(void) {
+  struct nfsstats ns;
+  size_t size = sizeof(ns);
+  int mib[] = {CTL_VFS, 2, NFS_NFSSTATS};
+  value_t values[nfs3_procedures_names_num];
+  int i;
+
+  /* NetBSD reports v2 statistics mapped to v3 and doen't yet support v4 */
```
kwiatrox: Maybe add a WARNING or INFO if user tries to config only v2 or v4. For now it will silently omit the parameters and leave the user without any information. This comment is probably the only place where this information is provided.

src/nfs.c
dago: That is documented in line 623
```
  /* NetBSD reports v2 statistics mapped to v3 and doen't yet support v4 */
```

src/nfs.c
dago: Updated.

src/nfs.c
dago: Warnings added.

PR #3378, opened 2020-02-18
Reorder conditions so that on Linux the Linux-specific code is execut…

dago:
…ed instead of the more generic code for getifaddrs()

ChangeLog: interface plugin: Reorder so that Linux-specific code is actually executed


mrunge:
@dago Somehow, the ChangeLog line can only be one line here, otherwise the check currently fails.

dago:
@mrunge I adjusted the ChangeLog line, however, the test was green even before

PR #3380, opened 2020-02-19
Remove section about self-approve patches

mrunge:
We moved away from that model and every commit should go
through a review.

### DIFF
--- a/docs/maintainer_guide.md
+++ b/docs/maintainer_guide.md
@@ -24,14 +24,3 @@ using your commit rights to merge the pull request.
 
 You have write access to the *collectd/collectd* repository. Please use it
 responsibly.
-
-#### Own work
-
-Open *pull requests* for your own changes, too:
-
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
I wish you good luck, while you doing rule changes by fact of technical changes.
Also, I wish you good luck waiting for minor changes review and recommend to do not approve changes until at least three reviewers will approve PR.

mrunge:
I am changing the documentation to reflect the current state.

rpv-tomsk:
By this, you state what current state is something that should really be.
But, maybe, current state should be changed to match existing rules, and not rules should be changed.

rpv-tomsk:
@octo, I think you lost control on project.

rpv-tomsk:
>We moved away from that model and every commit should go through a review.

Then you also must remove following section too:

```
### Repository access

You have write access to the *collectd/collectd* repository. Please use it
responsibly.
```

PR #2352, opened 2017-07-08
openvpn: reworked status file parser

rpv-tomsk:
* Added detection of status fields index by parsing 'HEADER' lines.
   This allows to support status files from latest OpenVPN-2.4 and possible from future versions;
 * Fixed a bug with OpenVPN status file version 3 where delimiter is 'tab' (not space or comma);
 * Status file version detection was moved from configuration callback to read callback;
 * Now 'plugin_register_complex_read()' is used for registering read callbacks.

Closes: #2128

PR #3381, opened 2020-02-19
Forward-port and rebase to master of "text protocol: Add support for meta_data in PUTVAL #2726"

dago:
ChangeLog: text protocol: Add support for meta_data in PUTVAL

### DIFF
--- a/src/utils/cmds/cmds_test.c
+++ b/src/utils/cmds/cmds_test.c
@@ -57,213 +57,122 @@ static struct {
 } parse_data[] = {
     /* Valid FLUSH commands. */
     {
-        "FLUSH",
-        NULL,
-        CMD_OK,
-        CMD_FLUSH,
+        "FLUSH", NULL, CMD_OK, CMD_FLUSH,
     },
     {
-        "FLUSH identifier=myhost/magic/MAGIC",
-        NULL,
-        CMD_OK,
-        CMD_FLUSH,
+        "FLUSH identifier=myhost/magic/MAGIC", NULL, CMD_OK, CMD_FLUSH,
     },
     {
-        "FLUSH identifier=magic/MAGIC",
-        &default_host_opts,
-        CMD_OK,
-        CMD_FLUSH,
+        "FLUSH identifier=magic/MAGIC", &default_host_opts, CMD_OK, CMD_FLUSH,
     },
     {
-        "FLUSH timeout=123 plugin=\"A\"",
-        NULL,
-        CMD_OK,
-        CMD_FLUSH,
+        "FLUSH timeout=123 plugin=\"A\"", NULL, CMD_OK, CMD_FLUSH,
     },
     /* Invalid FLUSH commands. */
     {
         /* Missing hostname; no default. */
-        "FLUSH identifier=magic/MAGIC",
-        NULL,
-        CMD_PARSE_ERROR,
-        CMD_UNKNOWN,
+        "FLUSH identifier=magic/MAGIC", NULL, CMD_PARSE_ERROR, CMD_UNKNOWN,
     },
     {
         /* Missing 'identifier' key. */
-        "FLUSH myhost/magic/MAGIC",
-        NULL,
-        CMD_PARSE_ERROR,
-        CMD_UNKNOWN,
+        "FLUSH myhost/magic/MAGIC", NULL, CMD_PARSE_ERROR, CMD_UNKNOWN,
     },
     {
         /* Invalid timeout. */
-        "FLUSH timeout=A",
-        NULL,
-        CMD_PARSE_ERROR,
-        CMD_UNKNOWN,
+        "FLUSH timeout=A", NULL, CMD_PARSE_ERROR, CMD_UNKNOWN,
     },
     {
         /* Invalid identifier. */
-        "FLUSH identifier=invalid",
-        NULL,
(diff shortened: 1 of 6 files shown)


dago:
@mrunge I found another issue: both `src/utils/cmds/putval.c` and the existing `src/onewire.c` uses `strndup` which is at least on Solaris not available. There is already a conditional include on gnulib for some socket stuff, should I adjust the code to pull in `strndup` from gnulib also if necessary?

dago:
I added `sstrndup` to `common.c` which seems useful in the future.

mrunge:
CI error is genuine

```
./../src/utils/cmds/putval.c:79:12: error: passing 'const char *' to parameter of type 'void *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]

      free(value_str);

           ^~~~~~~~~

/usr/include/stdlib.h:483:25: note: passing argument to parameter '__ptr' here

extern void free (void *__ptr) __THROW;
```

dago:
In the original PR there was a cast to `(void *)` but rpm-tomsk questioned if this would be necessary, in my environment it worked, but obviously it is needed. I removed the erranous commit.

dago:
Now there are probably real failures in the testsuite that needs to be analyzed.

dago:
@mrunge This should IMHO be fine for merging now.

src/utils/common/common.c
```diff
@@ -160,6 +160,28 @@ char *sstrdup(const char *s) {
...
 } /* char *sstrdup */
 
+char *sstrndup(const char *s, size_t n) {
+  char *r;
+  size_t sz;
+
+  if (s == NULL)
+    return NULL;
+
+  sz = strlen(s);
```
kwiatrox: Afaik the implementation of strndup uses strnlen and supports also not null-terminated strings. To be in line with that we would need to add sstrnlen probably.

dago: Nope, the string must be null terminated as e.g. explained in https://en.cppreference.com/w/c/experimental/dynamic/strndup:

> Returns a pointer to a null-terminated byte string, which contains copies of at most size bytes from the string pointed to by str. If the null terminator is not encountered in the first size bytes, it is added to the duplicated string.

I got the inspiration for the code from gnulib at https://github.com/danielj7/gnulib/blob/master/lib/strndup.c with a replacement of `strnlen()`.

dago: Dang, you are right. Thanks for catching this, I'll fix it.

src/utils/cmds/putval.c
```diff
@@ -49,10 +58,37 @@ static int set_option(value_list_t *vl, const char *key, const char *value) {
...
+      vl->meta = meta_data_create();
+      if (vl->meta == NULL) {
+        return CMD_ERROR;
+      }
+    }
 
-  return 0;
+    if (is_quoted(value, value_len)) {
+      int metadata_err;
+      const char *value_str = sstrndup(value + 1, value_len - 2);
```
rpv-tomsk: What the ... are you doing here?

Instead of inventing square wheels, you really need: 

```
-      const char *value_str = sstrndup(value + 1, value_len - 2);
+      char *value_str = strdup(value + 1);
+      value_str[value_len - 2] = '\0';
```

I'm chocked seeing ` exit(3);` in `sstrndup` implementation.

dago: So you are probably as shocked to see `exit(3)` in the `sstrdup` implementation from Octo three years ago...

src/utils/cmds/cmds_test.c
```diff
@@ -204,6 +204,18 @@ static struct {
         CMD_OK,
         CMD_PUTVAL,
     },
+    {
+        "PUTVAL myhost/magic/MAGIC meta:KEY=\"string_value\" 1234:42",
```
rpv-tomsk: This test is invalid.
In this case `set_option()` receives `string_value` in `value`  variable.
`is_quoted()` returns false, no metadata added.

src/utils/cmds/cmds_test.c
```diff
@@ -204,6 +204,18 @@ static struct {
...
+        NULL,
+        CMD_OK,
+        CMD_PUTVAL,
+    },
+    {
+        "PUTVAL myhost/magic/MAGIC meta:KEY='string_value' 1234:42",
```
rpv-tomsk: This test is invalid too.
In this case `set_option()` receives `'string_value'` in `value` variable.
`is_quoted()` returns false for single quotes. No metadata added.

src/utils/cmds/putval.c
rpv-tomsk: That is not reason to copy errors.

rpv-tomsk:
Please stop merging broken code.

src/utils/cmds/putval.c
rpv-tomsk: Accidentally found comment from @octo in https://github.com/collectd/collectd/pull/2352

>That doesn't make sense, since the difference between strdup() and sstrdup() 
>is that the latter exits. I'd prefer to switch to strdup() here 
>(and leave the error handling). 
>Unfortunately I haven't found the energy to remove the uses of sstrdup() globally yet.

This only confirms my vision,  what no `exit()` should be called in _monitoring_ system in case of OOM.

PR #3390, opened 2020-02-29
I think in this case we need sfree(cb);

klebertarcisio:
Hi guys,

I believe that in this case we need to call the sfree() function.

Regards.

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -1306,7 +1306,7 @@ EXPORT int plugin_register_flush(const char *name, plugin_flush_cb callback,
             .data = cb,
             .free_func = plugin_flush_timeout_callback_free,
         });
-
+    sfree(cb);
     sfree(flush_name);
     return status;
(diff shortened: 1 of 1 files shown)


src/daemon/plugin.c
```diff
@@ -1306,7 +1306,7 @@ EXPORT int plugin_register_flush(const char *name, plugin_flush_cb callback,
             .data = cb,
             .free_func = plugin_flush_timeout_callback_free,
         });
-
+    sfree(cb);
```
dago: I think freeing it here is wrong as the data is passed to `plugin_register_complex_read` and freed after use by calling `plugin_flush_timeout_callback_free`, but I may not understand the code well enough.

klebertarcisio: exactly, my bad. In this case it is unnecessary to call sfree() again.

dago:
Closing as suggested code is wrong.

PR #3392, opened 2020-03-02
call sfree()

klebertarcisio:
Hi guys, I believe that in this case we need to call sfree ().

Regards.


src/rrdtool.c
```diff
@@ -643,6 +643,8 @@ static int rrd_cache_insert(const char *filename, const char *value,
     DEBUG("rrdtool plugin: (rc->last_value = %" PRIu64 ") "
           ">= (value_time = %" PRIu64 ")",
           rc->last_value, value_time);
+    sfree(rc->values);
```
rpv-tomsk: No, you wrong here.
If `rc->last_value >= value_time` then `rc` is cached value, not newly `malloc()-ated`.
It still in cache and does not needs to be freed.

PR #3394, opened 2020-03-03
Add a FieldSeaparator option to tail_csv plugin

### LINKED ISSUE #2893: Plugin tail_csv should allow user specified separator
Plugin `tail_csv` currently supports only files where the separator is comma (`,`). It should be possible to specify a different character as delimiter for each file being processed.

There is an open PR #948 but that needs some rework before it can be merged. This issue can be used to collect feedback from community/core-team if such a feature is desirable.

redbaron4:
This PR adds a new option to tail_csv plugin called `FieldSeparator`. The value for this option is a single character. This option is used to define a custom separator. Default will be ',' so the PR is backward compatble.

Closes #2893 

ChangeLog: tail_csv plugin: Adds a FieldSeparator option

### DIFF
--- a/src/tail_csv.c
+++ b/src/tail_csv.c
@@ -47,6 +47,7 @@ struct instance_definition_s {
   char *plugin_name;
   char *instance;
   char *path;
+  char field_separator;
   cu_tail_t *tail;
   metric_definition_t **metric_list;
   size_t metric_list_len;
@@ -157,7 +158,7 @@ static int tcsv_read_buffer(instance_definition_t *id, char *buffer,
   /* Count the number of fields. */
   metrics_num = 1;
   for (i = 0; i < buffer_size; i++) {
-    if (buffer[i] == ',')
+    if (buffer[i] == id->field_separator)
       metrics_num++;
   }
 
@@ -179,7 +180,7 @@ static int tcsv_read_buffer(instance_definition_t *id, char *buffer,
   metrics[0] = ptr;
   i = 1;
   for (ptr = buffer; *ptr != 0; ptr++) {
-    if (*ptr != ',')
+    if (*ptr != id->field_separator)
       continue;
 
     *ptr = 0;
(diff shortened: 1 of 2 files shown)


src/tail_csv.c
```diff
@@ -277,6 +278,28 @@ static int tcsv_config_get_index(oconfig_item_t *ci, ssize_t *ret_index) {
...
 }
 
+static int tcsv_config_get_separator(oconfig_item_t *ci, char *ret_char) {
+  size_t len_opt;
+
+  if ((ci->values_num != 1) || (ci->values[0].type != OCONFIG_TYPE_STRING)) {
+    WARNING("tail_csv plugin: The \"%s\" config option needs exactly one "
+            "string argument.",
+            ci->key);
+    return -1;
```
rpv-tomsk: You return `-1` here, so reporting should be ERROR.

src/tail_csv.c
```diff
@@ -277,6 +278,28 @@ static int tcsv_config_get_index(oconfig_item_t *ci, ssize_t *ret_index) {
...
+  }
+
+  len_opt = strlen(ci->values[0].value.string);
```
rpv-tomsk: Policy ask us to declare variables at first use.

redbaron4: Thanks for looking at this @rpv-tomsk , the variable has been declared at L282 unless I am missing something here?

rpv-tomsk: https://collectd.org/review-comments#initialize-variables

redbaron4:
@rpv-tomsk Ah OK I see it now. Do you want me to open another pull request to fix these issues. I can mark it as a WIP so that merging is hopefully not done until it has been satisfactorily reviewed.

src/tail_csv.c
dago: Other invalid configuration options are also reported as `WARNING` and should then for consistency also be changed to `ERROR`, however, I see some more occurrences of `WARNING` in other plugins for ignored options, so this looks overall consistent me unless I am missing something.

rpv-tomsk: Don't look into plugins code, maintainers often do merge without deep analysis.
Look into `src/daemon/configfile.c` that maintained by "core maintainers" which are more qualified. 
The policy is simple - if function returns error, then message should be `ERROR` too, otherwise `WARNING`.

dago: From what I can see in the code `ERROR` is always followed by `return -1`, but there are occurrences in the code where `return -1` is preceded by `WARNING`, so I would say causality goes only in one direction here, e.g.
https://github.com/collectd/collectd/blob/6079accece816abc4ff74724d90ac499847a6246/src/daemon/configfile.c#L414-L417
However, from looking at the actual usages in the `tail_csv.c` plugin these should very well be errors and warnings should probably only allowed for backwards compatibility which is not the case here.

PR #3398, opened 2020-03-04
write_sensu: add the IncludeSource option

mbakhoff:
ChangeLog: write_sensu: add the IncludeSource option

It might make sense to enable the new option by default because it's probably what most users expect the plugin to do.


mbakhoff:
Thank you for the review!

Issue #3400, opened 2020-03-06
Release tag missing


bzed:
@mrunge please push the release tag.

mrunge:
Hi,
what release tag is missing? 5.11 is not released yet...

ashie:
Hmm, `./version-gen.sh` points `5.9.2.550.g62ae49f` while 5.10.0 is [already released](https://collectd.org/wiki/index.php/Version_5.10).

It seems that the tag name for 5.10.0 is wrong:

```
$ git tag

5.10.0
collectd-2.0.1
collectd-3.10.0
...
collectd-5.9.2
svn-trunk
```

(`collectd-5.10.0` is missing, there is `5.10.0` instead)

See also: #3293

mrunge:
> Hmm, `./version-gen.sh` points `5.9.2.550.g62ae49f` while 5.10.0 is [already released](https://collectd.org/wiki/index.php/Version_5.10).
> 

see https://github.com/collectd/collectd/blob/collectd-5.10/version-gen.sh ?


I'll cherry-pick the changes to master as well.

mrunge:
Sorry for the delay here, we'll fix the missing tag later today while we do the release for 5.11.0.

mrunge:
The tag https://github.com/collectd/collectd/tree/collectd-5.10.0 has been pushed, along with collectd-5.11.0.

PR #3405, opened 2020-03-11
Merge collectd 5.10 back into master

mrunge:
this fixes also one mentioned issue in https://github.com/collectd/collectd/issues/3400

### DIFF
--- a/src/varnish.c
+++ b/src/varnish.c
@@ -631,6 +631,33 @@ static int varnish_monitor(void *priv,
     else if (strcmp(name, "thread_queue_len") == 0)
       return varnish_submit_gauge(conf->instance, "workers", "queue_length",
                                   "threads", val);
+    else if (strcmp(name, "n_wrk") == 0)
+      return varnish_submit_gauge(conf->instance, "workers", "threads",
+                                  "worker", val);
+    else if (strcmp(name, "n_wrk_create") == 0)
+      return varnish_submit_derive(conf->instance, "workers", "total_threads",
+                                   "created", val);
+    else if (strcmp(name, "n_wrk_failed") == 0)
+      return varnish_submit_derive(conf->instance, "workers", "total_threads",
+                                   "failed", val);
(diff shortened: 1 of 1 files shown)


mrunge:
Varnish fails to build.

https://travis-ci.org/github/collectd/collectd/jobs/660965154#L31893

dago:
I think this is an artefact of #3360 and should have been fixed with https://github.com/collectd/ci-docker/pull/3

mrunge:
Right, so shall I skip this for now? I'd say, let's drop it in CI, but I'd leave that to any reviewer to decide.

kkepka:
just quick note without diving into details (yet) - https://github.com/collectd/ci-docker/pull/3 have only effect in cirrusci checks (as trusty base image)
here build has failed on travis check based on bionic, from logs, we can see varnish in ver 5.2.1-1

mkobyli:
In my opinion if it possible we should update varnish to latest stable and supported version (6.0) on CI

rpv-tomsk:
>mrunge wants to merge 29 commits into collectd:master from mrunge:master

1) Why do you open your PR from your own repo?
I think, such PR should be opened directly from collectd repo, w/o any third-party repos.

2) collectd-5.11 is already branched. By branching model, used in Collectd, you should merge collectd-10 into collectd-11 and then merge collectd-11 into master.

mrunge:
I am following the documentation mentioned here: https://collectd.org/wiki/index.php/Submitting_patches

rubenk:
Normally this should go via 5.11 indeed, to preserve a linear history but in practice it doesn't matter much as long as you merge 5.10 in 5.11 too.

rpv-tomsk:
I'm happy to see how fast work on Collectd project is done now.

mrunge:
fails due to varnish compile issues

rubenk:
Shall we handle #3408 first and then merge 5.11 into master?

mrunge:
Makes sense to me

rubenk:
No need for this anymore.

PR #3407, opened 2020-03-12
intel_pmu plugin: add check for lib symbol in configure.ac

kwiatrox:
ChangeLog: Improved check for libjevents which is dependency for intel_pmu plugin

Add a check to provide better information for users who would try to compile the plugin with the old libjevents version.


kwiatrox:
Thanks for fast response @mrunge !

PR #3409, opened 2020-03-13
Merge 5.10 branch into 5.11

rpv-tomsk:
Unlike # 3408 and # 3406, this PR was created using github webui and using only Collectd official repo branches, without any third-party repo.


rpv-tomsk:
Hmm, very strange. This way tries to update files `Makefile.am`, `configure.ac`, `src/turbostat.c` while these changes already applied into `collectd-5.10` branch.

PR #3410, opened 2020-03-13
amqp: allow multiple hosts for failover

mbakhoff:
ChangeLog: amqp plugin: allow multiple hosts to support failover

With this implementation all the hosts must use the same port number. Specifying a different port for each host would be tricky wrt config backwards compatibility and it's not clear how important it is.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -540,7 +540,7 @@ B<Synopsis:>
    # Send values to an AMQP broker
    <Publish "some_name">
      Host "localhost"
-#    Host "amqp1.example.com" "amqp2.example.com" "amqp3.example.com"
+     Host "fallback-amqp.example.com"
      Port "5672"
      VHost "/"
      User "guest"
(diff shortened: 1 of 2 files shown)


src/collectd.conf.pod
```diff
@@ -540,6 +540,7 @@ B<Synopsis:>
    # Send values to an AMQP broker
    <Publish "some_name">
      Host "localhost"
+#    Host "amqp1.example.com" "amqp2.example.com" "amqp3.example.com"
```
octo, line 543: I find this comment a bit distracting. How about amending the existing line, e.g. to:

```
Host "localhost"
Host "fallback-amqp.example.com"
```

mbakhoff:
Thank you @octo for the review! 

I considered round robin and thought that using random might balance the load better. It was also a bit easier to implement. I'm not aware of any use cases where setting a preference for a clustered broker node is useful. However if you think that round robin would be better, then I can make the change. Should be quite simple and I don't have a strong preference.

src/collectd.conf.pod
mbakhoff: good idea. pushed

mbakhoff:
Thank you!

PR #3413, opened 2020-03-13
Remove duplicate plugin author from AUTHORS

rubenk:


AUTHORS
```diff
@@ -71,9 +71,6 @@ Aneesh Puttur <aputtur at redhat.com>
 Andy Smith <ansmith at redhat.com>
  - AMQP 1.0 plugin.
  
-Aneesh Puttur <aputtur at redhat.com>
```
mrunge: could you please keep him here and remove the entry between both Andies? (6 lines above) That'd leave the list sorted alphabetically by name

rubenk: Ah yes of course, hold on.

PR #3420, opened 2020-03-16
Revert "Forward-port and rebase to master of "text protocol: Add support for meta_data in PUTVAL #2726""

rpv-tomsk:
Changelog: Reverts collectd/collectd#3381

Feature, added in #3381, is not working.
Implementation broken. Tests are wrong.

`sstrndup()` implementation is wrong (it should never call `exit()`)


rpv-tomsk:
This shit is never ending: `Please run: contrib/format.sh src/utils/common/common.c`.

How this possible after tree-wide formatting, and for PR which only removes code?

PR #3422, opened 2020-03-16
network plugin: Optimize metadata "network:ip_address" adding

rpv-tomsk:
Changelog: network plugin: Optimize metadata "network:ip_address" adding
Changelog: network plugin: Optimize network_receive()
Changelog: network plugin: Add client address into error messages

There is no need in some `memcpy` and `memset` calls.

Issue: #3191

PR #3428, opened 2020-03-21
network plugin: Reworked fix of data race accessing `sending_sockets`

rpv-tomsk:
Data race conditions explanation in #3310 is a bit incorrect.

The `network_send_buffer()` where `se->lock` was added in
c44c159065daf8bc7ab6c03287f281d317b1d5fd, is called by chain `network_write() -> flush_buffer()`.

The `network_write()` could be called from multiply threads, but it has `send_buffer_lock`, so all calls to `network_send_buffer()` are serialized and there is no data race.

The issue could be caused by second possible call path, `network_notification()->network_send_buffer()`. There is no lock, so race can occur.

Notifications are sent rarely, so lock on each socket entry is excessive.

Related: #3310
Reworks: c44c159065daf8bc7ab6c03287f281d317b1d5fd

Changelog: network plugin: Reworked fix of data race accessing `sending_sockets`

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -142,7 +142,6 @@ typedef struct sockent {
   } data;
 
   struct sockent *next;
-  pthread_mutex_t lock;
 } sockent_t;
 
 /*                      1 1 1 1 1 1 1 1 1 1 2 2 2 2 2 2 2 2 2 2 3 3
@@ -1133,7 +1132,7 @@ static int parse_part_sign_sha256(sockent_t *se, /* {{{ */
 
   return 0;
 } /* }}} int parse_part_sign_sha256 */
-  /* #endif HAVE_GCRYPT_H */
+/* #endif HAVE_GCRYPT_H */
 
 #else  /* if !HAVE_GCRYPT_H */
 static int parse_part_sign_sha256(sockent_t *se, /* {{{ */
@@ -1288,7 +1287,7 @@ static int parse_part_encr_aes256(sockent_t *se, /* {{{ */
 
   return 0;
 } /* }}} int parse_part_encr_aes256 */
-  /* #endif HAVE_GCRYPT_H */
+/* #endif HAVE_GCRYPT_H */
 
 #else  /* if !HAVE_GCRYPT_H */
 static int parse_part_encr_aes256(sockent_t *se, /* {{{ */
@@ -1568,7 +1567,6 @@ static void sockent_destroy(sockent_t *se) /* {{{ */
 
     sfree(se->node);
     sfree(se->service);
-    pthread_mutex_destroy(&se->lock);
 
     if (se->type == SOCKENT_TYPE_CLIENT)
       free_sockent_client(&se->data.client);
@@ -1697,7 +1695,7 @@ static int network_set_interface(const sockent_t *se,
       ERROR("network plugin: setsockopt (bind-if): %s", STRERRNO);
       return -1;
     }
-      /* #endif HAVE_IF_INDEXTONAME && SO_BINDTODEVICE */
+    /* #endif HAVE_IF_INDEXTONAME && SO_BINDTODEVICE */
 
 #else
     WARNING("network plugin: Cannot set the interface on a unicast "
@@ -1887,7 +1885,6 @@ static sockent_t *sockent_create(int type) /* {{{ */
   se->service = NULL;
   se->interface = 0;
   se->next = NULL;
-  pthread_mutex_init(&se->lock, NULL);
 
   if (type == SOCKENT_TYPE_SERVER) {
     se->data.server.fd = NULL;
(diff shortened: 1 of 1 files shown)


rpv-tomsk:
CC @yousong

rpv-tomsk:
Can you please do AddressSanitizer check again?

rpv-tomsk:
@yousong

Kindly ask to excuse ours release-managers for missing Changelog records. 
I hope they will fix that, and will be more accurate in future.

Your work on Collectd and your contribution is very important.
Thanks.

yousong:
Thanks for contacting me.  It has been a while since my last try on this and please do correct me in case my statements below are misleading or incorrect, etc..

I do have "threshold" enabled to generate notifications for network disruption events.  In extreme cases, the event can happen almost on every interval for all day long.

So the call chain could be 

```
ut_check_threshold: threshold write cb
  plugin_dispatch_notification
    network_notification: network notification cb
      network_send_buffer
```

"network_write()" accumulates data and sends them when the shared buffer is full.  "send_buffer_lock" protects this global "send_buffer".

On the other hand, I guess "network_notification()" intends to be responsive.  It uses a local scratch buffer and sends it out immediately.

So I guess even though the "send_buffer_lock" can indeed serialize things here, it's a misuse of the lock.  Or we can rename the lock to another name to reflect its new purpose.  Lock for each sockent can make sending to each remote server potentially parallel in a sequel way across different threads.

The other thing is that I have "network" plugin write to influxdb.  Now I look at it, the part I can't explain is that I see no notification data recorded in influxdb.

rpv-tomsk:
>Lock for each sockent can make sending to each remote server potentially parallel in a sequel way across different threads.

>On the other hand, I guess "network_notification()" intends to be responsive. It uses a local scratch buffer and sends it out immediately.

Yes, you are correct in this, but I don't think costs of locking on socket while sending metrics is really required. I think, use of same lock will not significally delay in notification sent.

If you really think we need locking on each sockent, maybe it will be enough to lock only when doing 'reconnect' in `sockent_client_connect()`? That will do things even more responsive.

What do you think?

rpv-tomsk:
>So I guess even though the "send_buffer_lock" can indeed serialize things here,
>it's a misuse of the lock. Or we can rename the lock to another name to reflect its new purpose.

I agree with this. Can you propose new name, if we agreed to use this lock in this way?

yousong:
> > Lock for each sockent can make sending to each remote server potentially parallel in a sequel way across different threads.
> 
> > On the other hand, I guess "network_notification()" intends to be responsive. It uses a local scratch buffer and sends it out immediately.
> 
> Yes, you are correct in this, but I don't think costs of locking on socket while sending metrics is really required. I think, use of same lock will not significally delay in notification sent.

Delay is not an issue with udp datagrams.  It's also not a computing hotspot.  I agree that a single lock for coordination purpose should suffice.  As for the name, I think "send_lock" should do, i.e. lock for send buffer and send sockets.

> 
> If you really think we need locking on each sockent, maybe it will be enough to lock only when doing 'reconnect' in `sockent_client_connect()`? That will do things even more responsive.

No, when we use mutex, it's meant for exclusive access, read and write.  rwlock is a bit overkill here.

rpv-tomsk:
Thanks, will update lock name soon.
If you interested, you can review #3422 (As a whole, as well as individual commits).

src/network.c
```diff
@@ -1887,7 +1886,6 @@ static sockent_t *sockent_create(int type) /* {{{ */
   se->service = NULL;
   se->interface = 0;
   se->next = NULL;
-  pthread_mutex_init(&se->lock, NULL);
```
mrunge: was this removed completely by accident?

src/network.c
```diff
@@ -1568,7 +1568,6 @@ static void sockent_destroy(sockent_t *se) /* {{{ */
 
     sfree(se->node);
     sfree(se->service);
-    pthread_mutex_destroy(&se->lock);
```
mrunge: this was also removed.

rpv-tomsk:
While you unable to read existing explanations I see no point to spent time on new.

rpv-tomsk:
Just press "approve" and "merge" like new maintainers always do, don't try to think.

rpv-tomsk:
Oh, sorry. You do "approve-n-merge" only for trusted maintainer's PR, while I'm untrusted,  sorry, I forgot.

mrunge:
> While you unable to read existing explanations I see no point to spent time on new.

The question was real, there is no need for you to be rude.

rpv-tomsk:
We completely revert https://github.com/collectd/collectd/commit/c44c159065daf8bc7ab6c03287f281d317b1d5fd. 
The `se->lock` lock completely removed, it was introduced in reverted commit.

To mitigate race we now use existing `send_buffer_lock` in `network_notification()`. 
In second commit that lock was renamed to better match new its role.

src/network.c
rpv-tomsk: As you can see, that field was fully removed from structure.

Issue #2486, opened 2017-10-16
[RFC] Per-(write)-plugin threads


octo:
This issue is supposed to be used to brainstorm / design a change to the *write threads* so that there is one write thread per plugin (or one thread pool per plugin).

In the current implementation, there is a central queue in which all *value lists* are enqueued. Write threads dequeue *value lists* and call all registered write callbacks sequentially. The problem with this implementation is that one typical failure mode of write callbacks is to time out: many establish or use a network connection and when there is a problem it can take several seconds before the appropriate system calls return. All write threads are then stuck in the same write plugin, the queue stalls and eventually overflows and metrics are lost. (#2104)

Most write plugins use locks in fairly central locations, making large parts effectively single-threaded. As determined by @rpv-tomsk:

> Most of write plugins do not intended their write callback to be called from multiple threads and will lock parallel calls.
>
> These will lock and behave as single-threaded: 
>
> `network.c`: will lock at global `send_buffer_lock` (only one callback can be registered)
> `snmp_agent.c`: will lock at global `g_agent->lock` (only one callback can be registered)
> `write_prometheus.c`: will lock at global `metrics_lock` (only one callback can be registered).
>
> Partially locked:
>
> `rrdtool.c`: Global lock at `rrd_cache_insert` after data prepared and `stat()` called, only one callback registered - treat it as a single-threaded.
>
> These will lock and are single-threaded per callback by a fact:
>
> `write_http.c`: has a callback lock (multiple callbacks can be registered)
> `write_sensu.c`: has a callback lock (multiple callbacks can be registered)
> 
> Partially locked (prepare data then do locks) (per-callback lock, multiple callbacks can be registered):
> 
> `write_graphite.c`: format_graphite() then locked send 
> `write_mongodb.c`: wm_create_bson() then locked send 
> `write_redis.c`: Prepare data with format_values() and then do locked send 
> `write_riemann.c`: Prepare data then do locked send  
> `write_tsdb.c`: Prepare messages then do locked send (each message send protected by a cb lock) 
> `amqp.c`: Prepare data, then do locked publish 
> `mqtt.c`: Prepare data, then do locked publish
> 
> No locks:
> 
> `csv.c`: Allows calls from multiple threads to be processed without lock (single callback registered)
> `write_kafka.c`: Allows calls from multiple threads to be processed without lock (multiple callbacks can be registered)
> `write_log.c`: Allows calls from multiple threads to be processed without lock (single callback registered)
[...]

PR #3432, opened 2020-03-23
amqp1: Add options to limit send queue length

ryan-mccabe:
ChangeLog: amqp1 plugin: Add options to limit send queue length.

Add two options to control send queue length and behavior: SendQueueLimit and SendQueueDropNew

If SendQueueLimit is set to an integer greater than 0, the send queue in the
amqp1 plugin will be constrained to that number of entries to avoid unbounded
memory use in the case that there is no amqp1 connection. If the
SendQueueDropNew parameter is set to true, new messages will be dropped when
the queue is full and a new message is enqueued, otherwise the oldest message
in the queue will be dropped. The default is for the oldest message to be
dropped.

Signed-off-by: Ryan McCabe <rmccabe@redhat.com>

### DIFF
--- a/src/amqp1.c
+++ b/src/amqp1.c
@@ -63,6 +63,7 @@ typedef struct amqp1_config_transport_s {
   char *password;
   char *address;
   int retry_delay;
+  int sendq_limit;
 } amqp1_config_transport_t;
 
 typedef struct amqp1_config_instance_s {
@@ -341,6 +342,15 @@ static int encqueue(cd_message_t *cdm,
   }
 
   pthread_mutex_lock(&send_lock);
+  if (transport->sendq_limit > 0 &&
+      DEQ_SIZE(out_messages) >= transport->sendq_limit) {
+    cd_message_t *evict;
+
+    DEBUG("amqp1 plugin: dropping oldest message because sendq is full");
+    evict = DEQ_HEAD(out_messages);
+    DEQ_REMOVE_HEAD(out_messages);
+    cd_message_free(evict);
+  }
   DEQ_INSERT_TAIL(out_messages, cdm);
   pthread_mutex_unlock(&send_lock);
 
@@ -697,6 +707,8 @@ static int amqp1_config_transport(oconfig_item_t *ci) /* {{{ */
       status = cf_util_get_int(child, &transport->retry_delay);
     else if (strcasecmp("Instance", child->key) == 0)
       amqp1_config_instance(child);
+    else if (strcasecmp("SendQueueLimit", child->key) == 0)
+      status = cf_util_get_int(child, &transport->sendq_limit);
     else
       WARNING("amqp1 plugin: Ignoring unknown "
               "transport configuration option "
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -860,6 +860,12 @@ When the AMQP1 connection is lost, defines the time in seconds to wait
 before attempting to reconnect. Defaults to 1, which implies attempt
 to reconnect at 1 second intervals.
 
+=item B<SendQueueLimit> I<SendQueueLimit>
+If there is no AMQP1 connection, the plugin will continue to queue
+messages to send, which could result in unbounded memory consumption. This
+parameter is used to limit the number of messages in the outbound queue to
+the specified value. The default value is 0, which disables this feature.
+
 =back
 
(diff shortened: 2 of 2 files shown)


src/amqp1.c
```diff
@@ -308,9 +310,25 @@ static int encqueue(cd_message_t *cdm,
                     amqp1_config_instance_t *instance) /* {{{ */
 {
   /* encode message */
-  pn_message_t *message = pn_message();
+  pn_message_t *message;
```
rpv-tomsk: Hi,

Please define variables on first use
https://github.com/collectd/collectd/blob/master/docs/review_comments.md

src/collectd.conf.pod
```diff
@@ -860,6 +860,18 @@ When the AMQP1 connection is lost, defines the time in seconds to wait
 before attempting to reconnect. Defaults to 1, which implies attempt
 to reconnect at 1 second intervals.
 
+=item B<SendQueueLimit> I<SendQueueLimit>
+If there is no AMQP1 connection, the plugin will continue to queue
+messages to send, which could result in unbounded memory consumption. This
+parameter is used to limit the number of messages in the outbound queue to
+the specified value. The default value is 0, which disables this feature.
+
+=item B<SendQueueDropNew> I<SendQueueDropNew>
```
rpv-tomsk: Why we need this option and behaviour configurable?

src/amqp1.c
ryan-mccabe: Ok, I will adjust that

src/collectd.conf.pod
ryan-mccabe: I thought somebody may want this behavior, but it's possible there is no real use case for it. Chatted with Matthias a bit, and he couldn't think of one. I can remove it for the time being, and if somebody ever needs it, it's easy to add back.

rpv-tomsk:
I can invite you to work on a broader kernel level solution.
Please take a look of comments in #2480. I'm open to discuss this feature and work on it.

src/collectd.conf.pod
rpv-tomsk: I think we don't need to introduce features just for their presence.
After release that should be supported "forever", as we prefer not to introduce possible regressions.

ryan-mccabe:
> I can invite you to work on a broader kernel level solution.
> Please take a look of comments in #2480 (and possible #2486 too).
> I'm open to discuss this feature and work on it.

This is something I may have some interest in working on, but I don't think it would solve the problem this patch intends to solve, as the data that is being queued causing the unbounded memory usage is not value list data, but amqp1 traffic/state.

rpv-tomsk:
>This is something I may have some interest in working on, but I don't think it would solve the problem this patch intends to solve, as the data that is being queued causing the unbounded memory usage is not value list data, but amqp1 traffic/state.

I'm not only about "exponential backoff" (as issue/pr named), but also about implementing unsent value list data queue "on kernel level". So, then any write plugin should return '0' when it unable to send more data, and can appear new mech which will drop or aggregate pending values (so, limiting queue size).

ryan-mccabe:
> I'm not only about "exponential backoff" (as issue/pr named), but also about implementing unsent value list data queue "on kernel level". So, then any write plugin should return '0' when it unable to send more data, and can appear new mech which will drop or aggregate pending values (so, limiting queue size).

Yes, I reread the proposal, and I think that would be a useful project, but I'm not sure it would solve things completely for this plugin. This plugin will keep amqp1 messages on the queue until they are acked, for example, if presettle is not enabled. If you have a receiver that is never acking (or perhaps is just very slow to ack) messages that it has received, the internal message queue could still grow large and consume a lot of memory, so it's more than simply not being able to write out value list data for this plugin.

rpv-tomsk:
>This plugin will keep amqp1 messages on the queue until they are acked, for example, if presettle is not enabled.

Why should the message remain in the internal queue in this case?
It should be resent? When?

src/collectd.conf.pod
kkepka: > don't need to introduce features just for their presence.

+1, and if at all would like to introduce some feature to test out (with a good reason), then it could/should be clearly marked as experimental (subject to change/remove)

kkepka:
Overall change looks good.
Could you (or anyone) let us know if / with what steps it has been verified on runtime? (in case anyone else also would like to check)

PR #3433, opened 2020-03-24
Fix memory leak in write_riemann

mrunge:


rubenk:
Why didn't you merge 5.9 into master?

mrunge:
Merging this into master was apparently forgotten, so, this was done right now and we should make sure the fix also lands on other branches, such as 5.11 and 5.10.

PR #3436, opened 2020-03-24
Fix memory leak in write_riemann

mrunge:


rpv-tomsk:
You should merge this into collectd-5.10 and then merge 5.10 into 5.11.

kkepka:
@rpv-tomsk you mean for more clear tree history? that make sense +1

rubenk:
Why would you want to merge this into 5.11? It's not that we're going to branch 5.12 of off the 5.11 branch.

mrunge:
I'd want to merge this fix into 5.11, since it was not on master, and thus not in 5.11, i.e. 5.11 was released without this fix.

rubenk:
Yes but since it's already relesased, what's the point?

mrunge:
don't we do a 5.11.1 for bugfixes?

rubenk:
That would be 5.12 then

faxm0dem:
what's the point in keeping a bugfix number in version then ?
I think it's sad if we don't keep bugfix releases

mrunge:
I was planning to do bugfix releases if needed.

Issue #1721, opened 2016-05-24
disk plugin: Incorrect value disk_time when use UdevNameAttr option.


letonphat1988:
Hi all,
I'm using Disk plugin with UdevNameAttr option to collect my HDD system. But i'm getting incorrect value for disk_time counter which in minutes or more not miliseconds. I tried use sda/sdb instead and everything's correct. 
Note: my system's idle, i haven't run anything yet sothat  highvalue's impossible
this's my config for Disk plugin
![image](https://cloud.githubusercontent.com/assets/16660339/15502289/388853b4-21dd-11e6-9be1-ca06e46769f2.png)
These're the records after i use UdevNameAttr
![image](https://cloud.githubusercontent.com/assets/16660339/15502212/c29cb410-21dc-11e6-99b5-d76d14fdbeae.png)
And these're the records if i use device name (sda, sdb,...)
![image](https://cloud.githubusercontent.com/assets/16660339/15502434/11186aa2-21de-11e6-8e0c-ab68bb4f76a6.png)

Do you think this's a bug ? or please give me some advices to debug.

rpv-tomsk:
`UdevNameAttr` does changes nothing in submitted values, it only affects output-name for block device.

1) Please show output of

udevadm info --query=all --name=/dev/sda

2) If you can, please build debug version of Collectd from sources (with configure --enable-debug) and run built Collectd in minimal configuration (with only required plugins - `disk`, `logfile` and ... plugin you use to submit values to display).

rpv-tomsk:
Issue confirmed.

rpv-tomsk:
> UdevNameAttr "ID_PATH"

Do not use that. It renames different block devices to the same name:

[2016-05-25 00:27:35] disk plugin: renaming sda2 => pci-0000:00:1f.2-scsi-0:0:0:0
[2016-05-25 00:27:35] disk plugin: renaming sda3 => pci-0000:00:1f.2-scsi-0:0:0:0

That causes all issues.

letonphat1988:
Hi rpv-tomsk, sorry about my delay. I missed the github's email notification.
About this case, I checked again and we're having a big conflict between paritions with UdevNameAttr.  But in my system, I only want to get the counters of whole disk and don't need the counters per partitions.
Do you think we can solve this issue by ignoring all partitions ?

rpv-tomsk:
[...]

PR #3441, opened 2020-04-19
Provide an udev rule with ID_SERIAL based, disk plugin usable ID attribute.

### LINKED ISSUE #823: Disk plugin with UdevNameAttr needs classifiers
UdevNameAttr option of disk plugin is very useful with storage systems. However relying on UdevNameAttr causes problems with partitions because most ID attributes ares exactly same for partitions and disks, for example ID_SERIAL and ID_SCSI_COMPAT are all identical for the disk and it's partitions. The main problem here is that when using ID_SCSI_COMPAT or ID_SERIAL you get multiple updates to the same file and flaky/skewed/erroneous statistics.

This necessitates some sort of additional classification by DEVTYPE and MINOR udev properties, for example if we'd use MINOR when DEVTYPE is "partition" appended to ID_SCSI_COMPAT we'd get a sane formatting like (suppose the disk has 1 partition):

```
{{ID_SCSI_COMPAT}}-part{{MINOR}}
=> SATA_ST3000DM001-1CH_W1F3P4AY-part1
```

And when we have DEVTYPE "disk" we can just use ID_SCSI_COMPAT by itself.

maciejsszmigiero:
Disk plugin UdevNameAttr option is very useful for making sure that the
same disk always ends in the same RRD file no matter which particular
device node it ends being assigned by the OS.

However, using the most fitting ID_SERIAL udev attribute for this purpose
results in data about the whole disk and each partition being mixed
together incorrectly, since this attribute has exactly the same value for
the whole disk device node and its particular partition nodes.

Let's introduce an ID_COLLECTD attribute instead which differentiates
between the whole disk and its particular partitions in a new udev rule
file provided in "contrib" directory.
To warn users about the issue let's also add a relevant note to the
UdevNameAttr option description.

Based on Stewart Adam's workaround from #823

Also included in this PR is a small commit that removes execute permission bits from two .c files that had them set needlessly.

ChangeLog: Disk plugin: Provide an udev rule with ID_SERIAL based attribute so disk and partition data don't get mixed incorrectly in udev-based disk naming.
Fixes: #823


igittigitthub:
Thank you very much! It's working for me.

This will also close the issue #1721 (`UdevNameAttr "ID_PATH"`)

Mini tutorial:
1. Put here your new rule `/etc/udev/rules.d/99-storage-collectd.rules`
1. `udevadm trigger` (the command `udevadm control --reload` is not enough)
1. Test it: `udevadm info --query=all --name=/dev/sda | grep "ID_COLLECTD"`
1. Change `UdevNameAttr "DEVNAME"` to `UdevNameAttr "ID_COLLECTD"` in your collectd configuration `/etc/collectd/collectd.conf`
1. `service collectd restart`
1. Enjoy your new fancy disk names

PR #3446, opened 2020-04-27
Add target in category for varnish 6

### LINKED ISSUE #2206: varnish: please support SMF statistics
Please enhance the varnish plugin to collect SMF (cache on disk) statistics. In the following example two on disk cache locations called **DISK_IMG** and **DISK_IMG2** are configured. So the plugin either needs to find the statistic names on its own from the API or there needs to be a way to configure the disk images which should be monitored.

```
SMF.DISK_IMG.c_req                       13346         0.02 Allocator requests
SMF.DISK_IMG.c_fail                          0         0.00 Allocator failures
SMF.DISK_IMG.c_bytes                 281821184       462.83 Bytes allocated
SMF.DISK_IMG.c_freed                 235171840       386.22 Bytes freed
SMF.DISK_IMG.g_alloc                       846          .   Allocations outstanding
SMF.DISK_IMG.g_bytes                  46649344          .   Bytes outstanding
SMF.DISK_IMG.g_space               10690768896          .   Bytes available
SMF.DISK_IMG.g_smf                         862          .   N struct smf
SMF.DISK_IMG.g_smf_frag                      0          .   N small free smf
SMF.DISK_IMG.g_smf_large                    16          .   N large free smf
SMF.DISK_IMG2.c_req                      21769         0.04 Allocator requests
SMF.DISK_IMG2.c_fail                         0         0.00 Allocator failures
SMF.DISK_IMG2.c_bytes               1631899648      2680.04 Bytes allocated
SMF.DISK_IMG2.c_freed                   405504         0.67 Bytes freed
SMF.DISK_IMG2.g_alloc                    21763          .   Allocations outstanding
SMF.DISK_IMG2.g_bytes               1631494144          .   Bytes outstanding
SMF.DISK_IMG2.g_space               9105924096          .   Bytes available
SMF.DISK_IMG2.g_smf                      21777          .   N struct smf
SMF.DISK_IMG2.g_smf_frag                     0          .   N small free smf
SMF.DISK_IMG2.g_smf_large                   14          .   N large free smf
```

Cactusbone:
ChangeLog: varnish plugin: Add target in category for varnish 6

Fixes #2206

I've limited it to varnish 6 because it's the only one I tested, but it should probably works for varnish 5 too.

This basically improve category precision by adding "target" information.
for SMA, targets are s0 or Transient, both of which have large difference, Transient being a small cache (around 20MB here), and s0 the large one (configured via malloc option). examples and explanation here : https://blog.anine.io/varnishstat-varnish-performance-tuning-part-ii-2/

For SMF, it should work the same way based on #2206 example (not tested)

I've also included backend name in VBE stats, for them to be of any use when you have more than one backend.

It's been a while since I've coded in C, so please triple check it :)

### DIFF
--- a/src/varnish.c
+++ b/src/varnish.c
@@ -858,38 +858,45 @@ static int varnish_monitor(void *priv,
                                    "bitmap", "happy_hprobes", val);
     */
     if (strcmp(name, "bereq_hdrbytes") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "total_bytes",
-                                   "bereq_hdrbytes", val);
+      return varnish_submit_derive_with_target(conf->instance, "vbe",
+                                               stat_target, "total_bytes",
+                                               "bereq_hdrbytes", val);
     else if (strcmp(name, "bereq_bodybytes") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "total_bytes",
-                                   "bereq_bodybytes", val);
+      return varnish_submit_derive_with_target(conf->instance, "vbe",
+                                               stat_target, "total_bytes",
+                                               "bereq_bodybytes", val);
     else if (strcmp(name, "bereq_protobytes") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "total_bytes",
-                                   "bereq_protobytes", val);
+      return varnish_submit_derive_with_target(conf->instance, "vbe",
+                                               stat_target, "total_bytes",
+                                               "bereq_protobytes", val);
     else if (strcmp(name, "beresp_hdrbytes") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "total_bytes",
-                                   "beresp_hdrbytes", val);
+      return varnish_submit_derive_with_target(conf->instance, "vbe",
+                                               stat_target, "total_bytes",
+                                               "beresp_hdrbytes", val);
     else if (strcmp(name, "beresp_bodybytes") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "total_bytes",
-                                   "beresp_bodybytes", val);
+      return varnish_submit_derive_with_target(conf->instance, "vbe",
+                                               stat_target, "total_bytes",
+                                               "beresp_bodybytes", val);
     else if (strcmp(name, "beresp_protobytes") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "total_bytes",
-                                   "beresp_protobytes", val);
+      return varnish_submit_derive_with_target(conf->instance, "vbe",
+                                               stat_target, "total_bytes",
+                                               "beresp_protobytes", val);
     else if (strcmp(name, "pipe_hdrbytes") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "total_bytes",
-                                   "pipe_hdrbytes", val);
+      return varnish_submit_derive_with_target(conf->instance, "vbe",
+                                               stat_target, "total_bytes",
+                                               "pipe_hdrbytes", val);
     else if (strcmp(name, "pipe_out") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "total_bytes",
-                                   "pipe_out", val);
+      return varnish_submit_derive_with_target(
+          conf->instance, "vbe", stat_target, "total_bytes", "pipe_out", val);
     else if (strcmp(name, "pipe_in") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "total_bytes",
-                                   "pipe_in", val);
+      return varnish_submit_derive_with_target(
+          conf->instance, "vbe", stat_target, "total_bytes", "pipe_in", val);
     else if (strcmp(name, "conn") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "connections",
-                                   "c_conns", val);
+      return varnish_submit_derive_with_target(
+          conf->instance, "vbe", stat_target, "connections", "c_conns", val);
     else if (strcmp(name, "req") == 0)
-      return varnish_submit_derive(conf->instance, "vbe", "http_requests",
-                                   "b_reqs", val);
+      return varnish_submit_derive_with_target(
+          conf->instance, "vbe", stat_target, "http_requests", "b_reqs", val);
   }
 
   /* All Stevedores support these counters */
--- a/src/varnish.c
+++ b/src/varnish.c
@@ -101,8 +101,9 @@ typedef struct user_config_s user_config_t; /* }}} */
 static bool have_instance;
 
 static int varnish_submit(const char *plugin_instance, /* {{{ */
-                          const char *category, const char *type,
-                          const char *type_instance, value_t value) {
+                          const char *category, const char *target,
+                          const char *type, const char *type_instance,
+                          value_t value) {
   value_list_t vl = VALUE_LIST_INIT;
 
   vl.values = &value;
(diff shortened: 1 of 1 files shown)


src/varnish.c
```diff
@@ -127,36 +134,106 @@ static int varnish_submit_gauge(const char *plugin_instance, /* {{{ */
...
+  const int MAX_TOKENS = 4; // we expect 4 maximum tokens
+  char *token_array[MAX_TOKENS];
+  int i = 0;
+  int token_count = -1;
+  char *buffer = strdup(pt->name);
+  char *running = buffer;
+  do {
+    token_array[i] = strsep(&running, ".");
+    if (token_array[i] == NULL) {
+      token_count = i - 1;
```
octo, line 209: This is very surprising and hard to read: if you have four tokens, the tokens will be in `token_array[0..3]` and `token count` will be 2 – neither the last element nor the actual count. Please either set the token count to the actual count, i.e. `i + 1`, or rename the variable to `last_index` and set it to `i`.

src/varnish.c
```diff
@@ -127,36 +134,106 @@ static int varnish_submit_gauge(const char *plugin_instance, /* {{{ */
...
+  char targetbuff[DATA_MAX_NAME_LEN];
+
+  sstrncpy(namebuff, token_array[token_count], sizeof(namebuff));
+  name = namebuff;
+
+  // stat_category is available at token_array[0], if we ever get stats with the
+  // same name for different categories;
+
+  if (token_count > 1) {
+    sstrncpy(targetbuff, token_array[token_count - 1], sizeof(targetbuff));
```
octo, line 230: This line and the comment above contradict each other. Which one is it, first element of `token_array` (`token_array[0]`) or third-to-last element, i.e. `token_array[-3]`?

src/varnish.c
```diff
@@ -127,36 +134,106 @@ static int varnish_submit_gauge(const char *plugin_instance, /* {{{ */
...
+} /* }}} int varnish_submit_derive_with_target */
+
 #if HAVE_VARNISH_V3 || HAVE_VARNISH_V4 || HAVE_VARNISH_V5 || HAVE_VARNISH_V6
 static int varnish_monitor(void *priv,
                            const struct VSC_point *const pt) /* {{{ */
 {
   uint64_t val;
   const user_config_t *conf;
   const char *name;
+  //  const char *stat_category;
```
octo, line 183: This comment is not helpful, please remove.

src/varnish.c
```diff
@@ -127,36 +134,106 @@ static int varnish_submit_gauge(const char *plugin_instance, /* {{{ */
...
 
   if (pt == NULL)
     return 0;
 
   conf = priv;
 
-#if HAVE_VARNISH_V5 || HAVE_VARNISH_V6
+#if HAVE_VARNISH_V6
+  /*
+   stats examples:
```
octo, line 193: For each of these, please specify what `name` and `stat_target` would be set to.

src/varnish.c
Cactusbone: comment above is to indicate the first token is a category, and can be used later. 
here we're computing second to last element to get the stat target (ie s0, transient, or a disk name).  I should add a comment to better separate, or remove the previous comment completely

src/varnish.c
```diff
@@ -127,36 +134,106 @@ static int varnish_submit_gauge(const char *plugin_instance, /* {{{ */
...
+   MAIN.threads
+   SMA.s0.c_req
+   SMA.Transient.c_req
+   VBE.vclName.backendName.req // where vclName changes every time a new vcl is
+   loaded
+  */
+  const int MAX_TOKENS = 4; // we expect 4 maximum tokens
+  char *token_array[MAX_TOKENS];
+  int i = 0;
```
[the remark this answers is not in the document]
Cactusbone: I'm trying to split the string by dot, and get the last token as name, and previous token as target. but I also need to know how many tokens were there to know if target exists (and is not the first token which is category).

At first I also wanted to check for category, so was trying to get the full token array, but since I'm not checking for it, this is overkill.

So v5 code might work to get the two last tokens, but I'll need to count tokens as well.

src/varnish.c
Cactusbone: when we get a null token, we're one index too far, hence the minus one. but you're right, this is an index and not a count :)

src/varnish.c
Cactusbone: Great idea ! I'll do the necessary changes and update tomorrow :)

src/varnish.c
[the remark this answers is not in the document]
octo: How about this:

```c
char *tokens[4] = {NULL};
size_t tokens_num = 0;
char *ptr = buffer;
char *saveptr = NULL;
while ((token = strtok_r(ptr, ".", &saveptr)) != NULL) {
  ptr = NULL;
  if (tokens_num < STATIC_ARRAY_SIZE(tokens)) {
    tokens[tokens_num] = token;
  }
  tokens_num++;
}
if ((tokens_num < 2) || (tokens_num > STATIC_ARRAY_SIZE(tokens))) {
  // handle error
}

// Note how all these assignments are close to each other,
// making it easy to grasp what is going on.
// "array[len - 1]" is a very common pattern for "give me the last element",
// "array[len - 2]" is also easy to read in this context.
category = tokens[0];
name = tokens[tokens_num - 1];
if (tokens_num >= 3) {
  target = tokens[tokens_num - 2];
}
```

*Caveat: typed from memory, mistakes likely*

Cactusbone:
Just though of it, but I'm not sure CI builds run against varnish 5 or varnish 6, ubuntu bionic (18.04) ships with varnish 5.2 and ubuntu focal (20.04) with 6.2.

src/varnish.c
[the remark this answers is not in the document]
Cactusbone: Thanks a lot, all assignments are close to each other, but since i need to copy strings, it's not as compact as your solution :)

src/varnish.c
```diff
@@ -127,36 +134,101 @@ static int varnish_submit_gauge(const char *plugin_instance, /* {{{ */
...
+} /* }}} int varnish_submit_derive_with_target */
+
 #if HAVE_VARNISH_V3 || HAVE_VARNISH_V4 || HAVE_VARNISH_V5 || HAVE_VARNISH_V6
 static int varnish_monitor(void *priv,
                            const struct VSC_point *const pt) /* {{{ */
 {
   uint64_t val;
   const user_config_t *conf;
   const char *name;
+  const char *stat_target = NULL;
```
octo, line 183: This needs to be moved down, into the `#if HAVE_VARNISH_V6` block.

src/varnish.c
```diff
@@ -127,36 +134,101 @@ static int varnish_submit_gauge(const char *plugin_instance, /* {{{ */
...
                                  uint64_t derive_value) {
-  return varnish_submit(plugin_instance, category, type, type_instance,
+  return varnish_submit(plugin_instance, category, NULL, type, type_instance,
                         (value_t){
                             .derive = (derive_t)derive_value,
                         });
 } /* }}} int varnish_submit_derive */
 
+static int
+varnish_submit_derive_with_target(const char *plugin_instance, /* {{{ */
```
octo, line 166: This function needs to be guarded by `#if HAVE_VARNISH_V6`.

src/varnish.c
```diff
@@ -127,36 +134,101 @@ static int varnish_submit_gauge(const char *plugin_instance, /* {{{ */
...
                                 uint64_t gauge_value) {
-  return varnish_submit(plugin_instance, category, type, type_instance,
+  return varnish_submit(plugin_instance, category, NULL, type, type_instance,
                         (value_t){
                             .gauge = (gauge_t)gauge_value,
                         });
 } /* }}} int varnish_submit_gauge */
 
+static int
+varnish_submit_gauge_with_target(const char *plugin_instance, /* {{{ */
```
octo, line 144: This function needs to be guarded by `#if HAVE_VARNISH_V6`.

src/varnish.c
Cactusbone: Usage is guarded behind #if HAVE_VARNISH_V4 || HAVE_VARNISH_V5 || HAVE_VARNISH_V6

idea was to define target only in V6, but functions are used for other versions as well to avoid too much duplication

Cactusbone: ha yeah, ubuntu trusty ships with varnish 3, which is not in this short list.

src/varnish.c
Cactusbone: guarded with #if HAVE_VARNISH_V4 || HAVE_VARNISH_V5 || HAVE_VARNISH_V6

src/varnish.c
Cactusbone: guarded with #if HAVE_VARNISH_V4 || HAVE_VARNISH_V5 || HAVE_VARNISH_V6

src/varnish.c
Cactusbone: guarded with #if HAVE_VARNISH_V4 || HAVE_VARNISH_V5 || HAVE_VARNISH_V6

PR #3448, opened 2020-04-28
Zookeeper missing mntr fields

hightea:
Add missing `zk_open_file_descriptor_count` and `zk_max_file_descriptor_count` mntr fields to handled fields for 3.4 version.
There was, I think, a typo resulting in checking 3 times the `zk_ephemerals_count` field instead of these two missing fields.

Also add missing fields for the 3.5 zookeeper version : `zk_last_proposal_size`, `zk_min_proposal_size` and `zk_max_proposal_size`
Note : list of fields in the mntr 4LW can be found here [http://zookeeper.apache.org/doc/r3.5.7/zookeeperAdmin.html#sc_4lw](http://zookeeper.apache.org/doc/r3.5.7/zookeeperAdmin.html#sc_4lw). The current documentation (for the 3.6.0) is not up-to-date

ChangeLog: zookeeper plugin: add missing mntr fields

PR #3449, opened 2020-04-28
Zookeeper - Fix missing Mntr fields

hightea:
Add missing `zk_open_file_descriptor_count` and `zk_max_file_descriptor_count` mntr fields to handled fields for 3.4 version.
There was, I think, a typo resulting in checking 3 times the `zk_ephemerals_count` field instead of these two missing fields.

ChangeLog: zookeeper plugin: fix missing mntr fields


hightea:
Hi @octo 
As you advised me in https://github.com/collectd/collectd/pull/3448#issuecomment-620635324 I created two PR : This one and #3450 (which you quickly merged, thanks !)

I see that there are no reviewer for the current PR; Is there anything I can do ? (I'm not aware of your process for the maintenance branches)

Thanks in advance,
Théophane

hightea:
Thanks !

Issue #3442, opened 2020-04-19
[mqtt] reconnect is not resubscribing to the queue


XANi:
*   Version of collectd: 5.8.1, 5.11.0 (debian unstable and stable)
*   Operating system / distribution: Debian Buster

A simple restart of mosquitto causes collectd to permanently lose subscription. In logs:

    mqtt_connect_broker: mosquitto_connect failed: Connection refused

My config is just
```
LoadPlugin "mqtt"
<Plugin "mqtt">
   <Subscribe "bunny">
     Host "127.0.0.1"
     Port "1883"
     User "xxx"
     Password "xxx"
     CleanSession true
   </Subscribe>
</Plugin>
```
+ bog-standard mosquitto config

from Wireshark of the traffic I can see connection is established but the queue subscription is not happening

PR #3452, opened 2020-05-01
MQTT: Resubscribe after a reconnect

maciejsszmigiero:
After losing our connection to a MQTT broker our subscription will be gone, too, so we need to subscribe to our topic once again after reconnecting, otherwise collectd will not receive any MQTT messages any more.

ChangeLog: MQTT plugin: Resubscribe after a reconnect
Fixes: #3442

### DIFF
--- a/src/mqtt.c
+++ b/src/mqtt.c
@@ -458,20 +458,19 @@ static int publish(mqtt_client_conf_t *conf, char const *topic,
     return -1;
   }
 
-  #if LIBMOSQUITTO_MAJOR == 0
-    status = mosquitto_loop(conf->mosq, /* timeout = */ 1000 /* ms */);
-  #else
-    status = mosquitto_loop(conf->mosq,
-                            /* timeout[ms] = */ 1000,
-                            /* max_packets = */ 1);
-  #endif
+#if LIBMOSQUITTO_MAJOR == 0
+  status = mosquitto_loop(conf->mosq, /* timeout = */ 1000 /* ms */);
+#else
+  status = mosquitto_loop(conf->mosq,
+                          /* timeout[ms] = */ 1000,
+                          /* max_packets = */ 1);
+#endif
 
   if (status != MOSQ_ERR_SUCCESS) {
     c_complain(LOG_ERR, &conf->complaint_cantpublish,
                "mqtt plugin: mosquitto_loop failed: %s",
-               (status == MOSQ_ERR_ERRNO)
-                   ? STRERRNO
-                   : mosquitto_strerror(status));
+               (status == MOSQ_ERR_ERRNO) ? STRERRNO
+                                          : mosquitto_strerror(status));
     /* Mark our connection "down" regardless of the error as a safety
      * measure; we will try to reconnect the next time we have to publish a
      * message */
--- a/src/mqtt.c
+++ b/src/mqtt.c
@@ -238,6 +238,19 @@ static void on_message(
   sfree(vl.values);
 } /* void on_message */
 
+static int mqtt_subscribe(mqtt_client_conf_t *conf) {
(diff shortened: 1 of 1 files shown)


maciejsszmigiero:
clang-format issues are outside the code parts affected by this change so they won't fit this PR very well.

src/mqtt.c
```diff
@@ -238,6 +238,20 @@ static void on_message(
   sfree(vl.values);
 } /* void on_message */
 
+static int mqtt_subscribe(mqtt_client_conf_t *conf) {
```
octo, line 241: I'm not convinced this needs its own function: the function is ~13 lines long and saves ~4 lines. I think the code would be shorter and easier to read without it.

octo:
> 
> 
> clang-format issues are outside the code parts affected by this change so they won't fit this PR very well.

Just put those cosmetic fixes in a separate commit – the check doesn't negotiate ;)

maciejsszmigiero:
Thanks for your quick review Florian, will add a comment fixing the clang-format reported things in a respin.

src/mqtt.c
maciejsszmigiero: Well, I think it's also about the code readability and maintainability.
Calling `mqtt_subscribe(conf)` two times looks nicer to me than calling `mosquitto_subscribe(conf->mosq, /* message_id = */ NULL, conf->topic, conf->qos)` two times.
And duplicating the code would duplicate not only the `mosquitto_subscribe() `call but also the whole error message.

octo: Agreed, it looks nicer. I'd argue the (less nice) code using `mosquitto_subscribe()` directly would be more readable, because all relevant information is immediately visible. Using the wrapper obfuscates this information.

I don't want to spend too much time on something with little (if any) gain, so let's go with the function.

src/mqtt.c
```diff
@@ -238,6 +238,20 @@ static void on_message(
...
+  int status;
+
+  status = mosquitto_subscribe(conf->mosq, /* message_id = */ NULL, conf->topic,
```
octo, line 244: Declare and initialize`status` simultaneously.

```c
int status = mosquitto_subscribe(…
```

https://collectd.org/review-comments#initialize-variables

src/mqtt.c
```diff
@@ -238,6 +238,20 @@ static void on_message(
...
+                               conf->qos);
+  if (status != MOSQ_ERR_SUCCESS) {
+    ERROR("mqtt plugin: Subscribing to \"%s\" failed: %s", conf->topic,
+          mosquitto_strerror(status));
+    return -1;
```
octo, line 249: Since `MOSQ_ERR_SUCCESS == 0`, we can just do `return status;` below.

src/mqtt.c
```diff
@@ -238,6 +238,20 @@ static void on_message(
...
+  }
```
octo, line 250: Both call sites call `mosquitto_disconnect(conf->mosq);` on failure. Should this be deduplicated here, too?

maciejsszmigiero: Okay.

src/mqtt.c
maciejsszmigiero: Well, that's mixing two return value classes: 0 / -1 and mosquitto error codes.

From just reading of this code it isn't clear that `MOSQ_ERR_SUCCESS` means zero (one has to dig into a mosquitto header file for this), so for readability and future ABI changes compatibility all `mqtt_subscribe()` callers (two) would need to check for this particular constant too.

So it's better to "translate" it immediately to zero in the single place in `mqtt_subscribe()` so callers can use the 0 / -1 status values as in the most other places in this code file.

PR #3455, opened 2020-05-04
Skip non-existing, but configured sockets

mrunge:
ChangeLog: ceph: skip non-existing ceph sockets

Otherwise collectd will continue to work, but e.g the unixsock plugin does not.

### DIFF
--- a/src/ceph.c
+++ b/src/ceph.c
@@ -205,6 +205,7 @@ enum cstate_t {
   CSTATE_READ_VERSION,
   CSTATE_READ_AMT,
   CSTATE_READ_JSON,
+  CSTATE_NOT_EXISTING,
 };
 
 enum request_type_t {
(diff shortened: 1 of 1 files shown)


octo:
Hi Matthias,

thanks for this PR. I'm a bit confused how the "unixsock plugin" relates to this change?

Best regards,
—octo

mrunge:
In our tests, we have seen that the unixsock plugin stops working as a side-effect of the ceph plugin complaining about errors. What I did is: load the unixsock plugin, load the ceph plugin and e.g configure the csv plugin to have a write plugin. If there is no ceph socket available (but configured), then collectd will misbehave.

octo:
Thanks Matthias!

I've tracked it down to the *ceph plugin* initializing a file descriptor to zero, then calling `close(0)` if establishing the connection fails. #3459 fixes this.

Due to this bug, the *socket(2)* call in the *unixsock plugin* returned zero. While the *unixsock plugin* was calling *accept(2)* on FD 0, the *ceph plugin* closed FD 0 again. This causes *accept(2)* to return `EBADF`.

I'd prefer to merge #3459 *instead* of this PR: using *access(2)* (very slightly) changes the behavior, because the permission check is done with the real user ID, not the effective user ID. What do you think?

Best regards,
—octo

mrunge:
https://github.com/collectd/collectd/pull/3459 is a cleaner approach. I've merged that one. Thank you for your help!

PR #3463, opened 2020-05-13
mysql: report slave_io_running and slave_sql_running

miesi:
ChangeLog: mysql plugin: report slave_io_running and slave_sql_running

added reporting of the fields Slave_IO_Running and Slave_SQL_Running as 0 and 1. Needed to catch more cases when replications broke.

I have added an entry to types.db, please advise me which existing type should be used.

### DIFF
--- a/src/mysql.c
+++ b/src/mysql.c
@@ -472,20 +472,13 @@ static int mysql_read_slave_stats(mysql_database_t *db, MYSQL *con) {
     unsigned long long counter;
     double gauge;
 
-    if (row[SLAVE_SQL_RUNNING_IDX] == NULL) {
-      gauge_submit("slave_running", "sql", 1, db);
-    } else if (strcasecmp(row[SLAVE_SQL_RUNNING_IDX], "yes") != 0) {
-      gauge_submit("slave_running", "sql", 1, db);
-    } else {
-      gauge_submit("slave_running", "sql", 0, db);
-    }
-    if (row[SLAVE_IO_RUNNING_IDX] == NULL) {
-      gauge_submit("slave_running", "io", 1, db);
-    } else if (strcasecmp(row[SLAVE_IO_RUNNING_IDX], "yes") != 0) {
-      gauge_submit("slave_running", "io", 1, db);
-    } else {
-      gauge_submit("slave_running", "io", 0, db);
-    }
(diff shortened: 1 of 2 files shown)


src/types.db
```diff
@@ -249,6 +249,7 @@ slurm_last_cycle_depth  value:GAUGE:0:U
 slurm_cycle_depth       value:DERIVE:0:U
 slurm_job_stats         value:DERIVE:0:U
 slurm_queue_length      value:DERIVE:0:U
+slave_running           value:GAUGE:0:U
```
octo, line 252: We're aiming at metric types that are a bit less specific than "slave running" (though, admittedly, we haven't consistently done a great job in the past). My recommendation would be to add a new type `bool value:GAUGE:0:1` and report these metrics as:

*   bool-slave-sql-running
*   bool-slave-io-running

src/mysql.c
```diff
@@ -472,6 +472,21 @@ static int mysql_read_slave_stats(mysql_database_t *db, MYSQL *con) {
     unsigned long long counter;
     double gauge;
 
+    if (row[SLAVE_SQL_RUNNING_IDX] == NULL) {
+      gauge_submit("slave_running", "sql", 1, db);
+    } else if (strcasecmp(row[SLAVE_SQL_RUNNING_IDX], "yes") != 0) {
+      gauge_submit("slave_running", "sql", 1, db);
+    } else {
+      gauge_submit("slave_running", "sql", 0, db);
+    }
```
octo, line 481: I think this could be condensed to (here and below):

```c
gauge_submit("bool", "slave-sql-running",
    (row[SLAVE_SQL_RUNNING_IDX] != NULL) && (strcasecmp(row[SLAVE_SQL_RUNNING_IDX], "yes") == 0),
    db);
```

This works, because a boolean in C is an integer that can only hold 0 and 1. It can implicitly we widened to a double, which is what this is taking advantage of. For the different metric name, see below.

src/types.db
miesi: Okay, will push this change to my fork later today.

src/mysql.c
miesi: I have adjusted the code with your recommendation and built rpm packages I'm now going to test.

PR #3478, opened 2020-06-07
[WIP] cgroups: Include the entire cgroup hierarchy recursively

### LINKED ISSUE #938: Recursive cgroups collection
Right now, the cgroup plugin only read cpuacct.stat from top-level directory, and does not walk into slices, for example.
Since systemd creates slices like this, the directory hierarchy can be like this:

    # pwd
    /sys/fs/cgroup/cpu,cpuacct/system.slice/system-ssh.slice
    # ls -l
    total 0
    drwxr-xr-x   7 root root 0 févr. 13 13:19 .
    drwxr-xr-x 111 root root 0 févr. 13 14:51 ..
    -rw-r--r--   1 root root 0 févr. 12 13:50 cgroup.clone_children
    -rw-r--r--   1 root root 0 févr. 12 13:50 cgroup.procs
    -r--r--r--   1 root root 0 févr. 12 13:50 cpuacct.stat
    -rw-r--r--   1 root root 0 févr. 12 13:50 cpuacct.usage
    -r--r--r--   1 root root 0 févr. 12 13:50 cpuacct.usage_percpu
    -rw-r--r--   1 root root 0 févr. 12 13:50 cpu.shares
    -rw-r--r--   1 root root 0 févr. 12 13:50 notify_on_release
    drwxr-xr-x   2 root root 0 févr. 12 14:58 ssh@23697-127.0.0.1:22-127.0.0.1:44011.service
    drwxr-xr-x   2 root root 0 févr. 12 20:27 ssh@23706-2a01:e35:2ec7:ae40::2:1:22-2a01:e35:2ec7:ae40:21f:d0ff:fe80:127d:46803.service
    drwxr-xr-x   2 root root 0 févr. 13 01:00 ssh@23712-127.0.0.1:22-127.0.0.1:58745.service
    drwxr-xr-x   2 root root 0 févr. 13 08:48 ssh@23747-127.0.0.1:22-127.0.0.1:34885.service
    drwxr-xr-x   2 root root 0 févr. 13 09:33 ssh@23751-127.0.0.1:22-127.0.0.1:36196.service
    -rw-r--r--   1 root root 0 févr. 12 13:50 tasks

The statistics are currently not collected for all ssh@… directories.

### LINKED ISSUE #2293: cgroup plugin missing cgroups directly under /sys/fs/cgroup/cpu
cgroups created directly under /sys/fs/cgroup/cpu (/sys/fs/cgroup/cpu/bar instead of /sys/fs/cgroup/cpu/foo/bar) are not picked up by the cgroup plugin.

systemd v232 with default layout:
```
│ ├─/sys/fs/cgroup                        tmpfs                                     tmpfs      ro,nosuid,nodev,noexec,mode=755
│ │ ├─/sys/fs/cgroup/systemd              cgroup                                    cgroup     rw,nosuid,nodev,noexec,relatime,xattr,release_agent=/usr/lib/systemd/systemd-cgroups-agent,name=systemd
│ │ ├─/sys/fs/cgroup/cpu,cpuacct          cgroup                                    cgroup     rw,nosuid,nodev,noexec,relatime,cpu,cpuacct
│ │ ├─/sys/fs/cgroup/devices              cgroup                                    cgroup     rw,nosuid,nodev,noexec,relatime,devices
│ │ ├─/sys/fs/cgroup/blkio                cgroup                                    cgroup     rw,nosuid,nodev,noexec,relatime,blkio
│ │ └─/sys/fs/cgroup/pids                 cgroup                                    cgroup     rw,nosuid,nodev,noexec,relatime,pids

```

gerow:
This can potentially break current cgroups users as it shifts existing metrics
to different names.

Fixes #938
Fixes #2293

ChangeLog: cgroups plugin: Include entire cgroup hierarchy recursively.

Hi folks, I'm marking this WIP since I'm not sure if it's mergable as-is since it ends up moving metric names around. For example, what was previously under the cgroup `apache2.service` will now show up under `/system.slice/apache2.service`.

I figure it's easiest if I just start with code and figure out how best to deal with that. If moving metrics around like this is considered OK then great, but otherwise it might make sense to gate this move behind a config (or do something else entirely).

### DIFF
--- a/src/cgroups.c
+++ b/src/cgroups.c
@@ -48,13 +48,14 @@ cgroups_submit_one(char const *plugin_instance, char const *type_instance,
   plugin_dispatch_values(&vl);
 } /* void cgroups_submit_one */
 
+static int read_cpuacct_walk(const char *, const char *, void *);
+
 /*
- * This callback reads the user/system CPU time for each cgroup.
+ * Read cpuacct info starting from dirname. This function is called recursively
+ * for child cgroups.
  */
-static int read_cpuacct_procs(const char *dirname, char const *cgroup_name,
-                              void *user_data) {
+static int read_cpuacct(const char *dirname, const char *cgroup_name) {
   char abs_path[PATH_MAX];
-  struct stat statbuf;
   char buf[1024];
   int status;
 
@@ -63,20 +64,8 @@ static int read_cpuacct_procs(const char *dirname, char const *cgroup_name,
   if (ignorelist_match(il_cgroup, cgroup_name))
     return 0;
 
-  snprintf(abs_path, sizeof(abs_path), "%s/%s", dirname, cgroup_name);
+  snprintf(abs_path, sizeof(abs_path), "%s/cpuacct.stat", dirname);
 
-  status = lstat(abs_path, &statbuf);
-  if (status != 0) {
-    ERROR("cgroups plugin: stat (\"%s\") failed.", abs_path);
-    return -1;
-  }
-
-  /* We are only interested in directories, so skip everything else. */
-  if (!S_ISDIR(statbuf.st_mode))
-    return 0;
-
-  snprintf(abs_path, sizeof(abs_path), "%s/%s/cpuacct.stat", dirname,
-           cgroup_name);
   fh = fopen(abs_path, "r");
   if (fh == NULL) {
     ERROR("cgroups plugin: fopen (\"%s\") failed: %s", abs_path, STRERRNO);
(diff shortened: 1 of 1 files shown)


gerow:
Another thing to note is that with this change the sum of all the cgroups metrics will now be more than the total cpu time used across the entire machine since some of the groups are supersets of others.

Also, I'm realizing that I'm allocating a 1k buffer on the stack each recursion which can easily be eliminated. That should probably be fixed.

gerow:
I'm going to close this PR for now, it isn't backwards compatible with existing cgroups users and probably doesn't make sense to merge. I'll just write a custom plugin and come back with proposals after figuring out if something else makes sense.

PR #3479, opened 2020-06-11
processes: increase reading buffer for /proc/stat

mhumpula:
If machine has high number of CPUs or interrupts or both, the /proc/stat
can easily grow over current buffer size 4kB. On my current machine (64
cores, ~300 interrupts) the /proc/stat has 12kB of data. And since the
`procs_running` line is after the interrupts it is never read.

Increasing the size to 64KB so it will accommodate even heavier machines.

It would be nice if this would be more dynamic to don't allocate large buffer on smaller machines. I have no idea about the current smallest supported setup by the collectd. Optionally the part could be reprogrammed to read the whole file in 4kB chunks, something similar is already done in `read_fork_rate`.

ChangeLog: Process plugin: Increase reading buffer for /proc/stat

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -757,7 +757,7 @@ static int ps_init(void) {
     pset_list_len = 0;
     return -1;
   }
-    /* #endif HAVE_THREAD_INFO */
+  /* #endif HAVE_THREAD_INFO */
 
 #elif KERNEL_LINUX
   pagesize_g = sysconf(_SC_PAGESIZE);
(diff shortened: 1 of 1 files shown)


mrunge:
Thank you for your contribution. This looks okay to me. Would you be able to run 
```
contrib/format.sh src/processes.c 
```
and push that change? Apparently, this is required for CI to pass here.

mhumpula:
I didn't wont to change non related lines, but ok. It's pushed reformated now.

PR #3480, opened 2020-06-17
intel_pmu: fix crash after specifying incorrect HardwareEvents

kwiatrox:
ChangeLog: intel_pmu plugin: fix the possible crash on plugin init

If HardwareEvents group in configuration consisted only of events that can't be resolved collectd was crashing during init. Added a check to end the group only if some events were actually added to the group.


kwiatrox:
That is a fast approval! Thanks @octo .

octo:
You got lucky ;)

PR #3481, opened 2020-06-18
strbuf: Library that implements printing to a dynamically allocated buffer.

octo:
I'm planning on using this code to build the string representation of a label based metric. This certainly will have uses all throughout the daemon though. Looking for "ADD" or "ADDF" macros should uncover places that might benefit from this code.

ChangeLog: collectd: The `strbuf` utility library has been added. It simplifies building strings iteratively.

### DIFF
--- a/src/utils/strbuf/strbuf.c
+++ b/src/utils/strbuf/strbuf.c
@@ -196,3 +196,82 @@ int strbuf_printf(strbuf_t *buf, char const *format, ...) {
 
   return 0;
 }
+
+int strbuf_printn(strbuf_t *buf, char const *s, size_t n) {
+  if ((buf == NULL) || (s == NULL))
+    return EINVAL;
+  if (n == 0) {
+    return 0;
+  }
+
+  size_t s_len = strnlen(s, n);
+  int status = strbuf_resize(buf, s_len);
+  if (status != 0)
+    return status;
+
+  size_t bytes = strbuf_avail(buf);
+  if (bytes == 0)
+    return ENOSPC;
+
+  if (bytes > s_len)
+    bytes = s_len;
+
+  memmove(buf->ptr + buf->pos, s, bytes);
+  buf->pos += bytes;
+  buf->ptr[buf->pos] = 0;
+
+  return 0;
+}
+
+int strbuf_print_escaped(strbuf_t *buf, char const *s, char const *need_escape,
+                         char escape_char) {
+  if ((buf == NULL) || (s == NULL) || (need_escape == NULL) ||
+      (escape_char == 0)) {
+    return EINVAL;
+  }
+
+  size_t s_len = strlen(s);
+  while (s_len > 0) {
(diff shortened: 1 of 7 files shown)


src/utils/strbuf/strbuf.h
```diff
@@ -0,0 +1,66 @@
...
+  &(strbuf_t) { .ptr = NULL }
+#define STRBUF_NEW_STATIC(b)                                                   \
+  &(strbuf_t) { .ptr = b, .size = sizeof(b), .fixed = 1 }
+#define STRBUF_DELETE(buf)                                                     \
+  do {                                                                         \
+    if (!buf->fixed) {                                                         \
+      free(buf->ptr);                                                          \
+    }                                                                          \
+    *buf = (strbuf_t){.ptr = NULL};                                            \
+  }
```
igorpeshansky, line 47: These macros are public. Should you be testing them as well?

octo: Excellent point, thank you! I've added tests and even found an error with the `STRBUF_DESTROY` macro.

octo:
I added an update to `src/utils/common/common.c` which showcases how the `strbuf`library can be used to simplify existing code. While there, I also improved the code coverage of the "common" library.

src/utils/strbuf/strbuf.c
```diff
@@ -0,0 +1,198 @@
...
+    if (tmp >= 1)
+      cached_pagesize = (size_t)tmp;
+    else
+      cached_pagesize = 1024;
+  }
+
+  return cached_pagesize;
+}
+
+/* strbuf_resize resizes an dynamic buffer to ansure that "need" bytes can be
```
sunkuranganath, line 56: 'to ensure'

octo: Done, thank you!

octo:
While using this library to simplify existing code, I noticed two more common use-cases:

1.  Limited print: provide a function that takes a size, printing only that number of characters. When extracting part of a string, printing everything up to the null byte forces users to copy the relevant part into a temporary buffer. The new `strbuf_printn` function addresses this use-case.
1.  Escaped print: provide a function that escapes the string before adding it to the buffer. When creating output strings, certain characters commonly need to be escaped, e.g. a `"` (quote) may need to be printed as `\"`. The new `strbuf_print_escaped` takes a list of characters that need to be escaped and an escape character, ensuring that each to-be-escaped character is preceded by the escape character.

    I'm a bit conflicted whether letting the caller provide a list of "forbidden" characters and an escape character is the best approach here. The existing `escape_string` function always uses `\` (backslash) to escape and has a fixed set of characters it escapes, including `\n` (newline) and `"` (quote). I.e. the flexibility may not be needed. On the other hand, code such as `format_json` have very specific escaping rules and implement their own functions for this. What do you think?

Thanks and best regards,
—octo

mrunge:
Looks good to me!

octo:
Thanks Matthias!

PR #3485, opened 2020-06-23
Use hostname_g in parse_identifier for a missing hostname in the identifier

YmrDtnJu:


YmrDtnJu:
I am aware of the failed check. I would change that if the PR is principally accepted.

PR #3486, opened 2020-06-24
ubi: fix data source type

feckert:
ChangeLog: ubi: fix data source type from counter to gauge

Currently the values are stored in the ubi plugin as data source type
`counter`. But this makes no sense, because the values change very slowly
and I don't want to know the rate of change. It is better to store the
value as data source type `gauge`. Then I can see the current value.


mrunge:
Thank you for your contribution!

PR #3491, opened 2020-06-29
network.c: process the full receive list at dispatch thread in each i…

carlospeon:
Hello:

Improve dispatch thread performance. Slightly related to #3490.

Regards.
Carlos.

ChangeLog: network.c: process the full receive list at dispatch thread in each iteration.

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -1133,7 +1133,7 @@ static int parse_part_sign_sha256(sockent_t *se, /* {{{ */
 
   return 0;
 } /* }}} int parse_part_sign_sha256 */
-  /* #endif HAVE_GCRYPT_H */
+/* #endif HAVE_GCRYPT_H */
 
 #else  /* if !HAVE_GCRYPT_H */
 static int parse_part_sign_sha256(sockent_t *se, /* {{{ */
@@ -1288,7 +1288,7 @@ static int parse_part_encr_aes256(sockent_t *se, /* {{{ */
 
   return 0;
 } /* }}} int parse_part_encr_aes256 */
-  /* #endif HAVE_GCRYPT_H */
+/* #endif HAVE_GCRYPT_H */
 
 #else  /* if !HAVE_GCRYPT_H */
 static int parse_part_encr_aes256(sockent_t *se, /* {{{ */
@@ -1697,7 +1697,7 @@ static int network_set_interface(const sockent_t *se,
       ERROR("network plugin: setsockopt (bind-if): %s", STRERRNO);
       return -1;
     }
-      /* #endif HAVE_IF_INDEXTONAME && SO_BINDTODEVICE */
+    /* #endif HAVE_IF_INDEXTONAME && SO_BINDTODEVICE */
 
 #else
     WARNING("network plugin: Cannot set the interface on a unicast "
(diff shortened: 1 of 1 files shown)


octo:
Hi @carlospeon,

thank you very much for your PR!

Do you have any profiling or other data that shows what kind of speedup we can expect from this? I wonder how this change would interact with #3492. Maybe each thread should only claim 1/*n* of the receive queue (where *n* is the number of dispatch threads)?

Best regards,
—octo

carlospeon:
> Hi @carlospeon,
> 
> thank you very much for your PR!
> 

Hi @octo, you are welcome. Glad to reach you.

> Do you have any profiling or other data that shows what kind of speedup we can expect from this? 

I don't have any number to show. I went into this piece of code after watching an overloaded system; I noticed that the dispatch thread didn't use all (single) cpu available while the receive queue grew and grew. I thought that the this was due the mutex overhead, so dispatching packets in "batchs" seemed a good optimization to me.

In the one_dispath_thread world this patch was though, I didn't found any good reason to unlock the mutex while there are still packets pending to dispatch, so processing the whole queue looked so straight forward that I didn't feel the need of measure the improvement.

> I wonder how this change would interact with #3492. Maybe each thread should only claim 1/_n_ of the receive queue (where _n_ is the number of dispatch threads)?

In the scenario of _n_ dispatch threads, the argument of the less the the mutex is requested the better the cpu will be used should still be valid, and this PR should work aswell. I have the perception that the receive thread is fast enough to feed multiple dispatch threads without the need of balancing packets between them, but It's just a though and again no numbers here.

On the other hand splitting the queue in chunks of _queue_lenght/n_ size seems a more "academic" approach, pretty easy to code and with low overhead. Maybe, just in the case of a healthy system with _queue_lenght_ close to 0, is not worth splitting the queue, but to be honest I don't have a strong opinion about what should be done and which option is better in performance if one.

> 
> Best regards,
> —octo

Best regards,
Carlos.

octo:
Hi @carlospeon, thanks for the clarification!

@manuelluis, I think this and your PR (#3492) try to tackle the same issue, that the dispatch thread can't keep up with the receive thread. However, it's unclear to me how both, the lock and the CPU cycles, can be a bottleneck simultaneously. Do you happen to have any profiling data? Alternatively, can you give this PR a try to see if it improves the performance enough for your situation?

The background is that this PR is significantly easier than #3492. If it solved the issue, I'd prefer to go with it.

Thanks and best regards,
—octo

src/network.c
```diff
@@ -2204,55 +2204,54 @@ static int sockent_add(sockent_t *se) /* {{{ */
...
-      if (i < se->data.server.fd_num)
-        break;
+      /* Look for the correct `sockent_t' */
+      sockent_t *se = listen_sockets;
+      while (se != NULL) {
+        size_t i;
 
-      se = se->next;
-    }
+        for (i = 0; i < se->data.server.fd_num; i++)
```
octo, line 2232: Please declare `i` inside the for loop:

```c
for (size_t i = 0; …
//   ^^^^^^
```

manuelluis:
@octo  the #3492 was in production for more than two years (now, I have a faster machine) to solve the problem that the dispatch thread cannot handle all traffic, it was near the 100% before this modification.

I can try to quantify the impact of the locking in the dispatch thread, I still have this old machine.

With this patch the metric `queue_length` are not accurate, It will be zero but there is still data to process in the dispatch thread. We need to add the pending messages in the dispatch queue.

octo:
@manuelluis Awesome, thank you!

You're right, the queue length is then inaccurate. Maybe we can use an atomic int to update the queue length lock-free …?

src/network.c
carlospeon: I think we shouldn't do that:
```
while (se != NULL) {
  size_t i;

  for (i = 0; i < se->data.server.fd_num; i++)
    if (se->data.server.fd[i] == ent->fd)
      break;

  if (i < se->data.server.fd_num)
    break;

  se = se->next;
}
```
It's the same code that main but indented as now run inside the loop that process whole queue.

carlospeon:
Hello, @octo, I'm closing this due to inactivity and because this idea of batch processing was evolved and extended to plugin.c and this PR is outdated. If interested please ping me.

PR #3503, opened 2020-07-10
intel_rdt: correct value metrics reported by the plugin

paweltox:
ChangeLog: intel_rtd plugin: Fixed nan values in csv output.


sunkuranganath:
This change introduces rdt_submit function that simplifies & removes duplication of data type submission, while supporting updated libpqos version 4.
As code owner of this plugin, I have already went through review & updates to this plugin internally. This includes code formatting, scale factor support and testing on RDT enabled hardware. 
Approving this code to be merged.

PR #3504, opened 2020-07-13
Add a Count option to snmp plugin

efuss:
ChangeLog: A count option was added allowing to gather the number of table entries (matching given criteria) rather than their values.

Add a `Count` option to the snmp plugin allowing to gather the number of table entries (matching given criteria) rather than their values.
This is especially useful in combination with the recently introduced Filter options.

Example: Count number of total/connected stations on a (Lancom) Access Point.

### DIFF
--- a/src/collectd-snmp.pod
+++ b/src/collectd-snmp.pod
@@ -30,6 +30,22 @@ collectd-snmp - Documentation of collectd's C<snmp plugin>
       #FilterValues "1", "2"
       Values "IF-MIB::ifInOctets" "IF-MIB::ifOutOctets"
     </Data>
+    <Data "lancom_stations_total">
+        Type "counter"
+        PluginInstance "stations_total"
+        Table true
+        Count true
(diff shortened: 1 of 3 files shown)


efuss:
> I have a small nit.
Which is?

mrunge:
> I have a small nit.
> Which is?

Oh, yes! Mentioning this helps. Sorry about that.

@octo dropped absolute counters in https://github.com/collectd/collectd/commit/97374e85e8be6319c02726ba56ad06669fba077e , it may be wise to just remove that from this patch as well.

efuss:
> @octo dropped absolute counters
But that's in the collectd-6.0 branch, no?
In main, I'll need to cope with someone requesting DS_TYPE_ABSOLUTE.

mrunge:
yes, that's correct, that's in 6.0. So, for 6.0 there is some change required.

efuss:
> So, for 6.0 there is some change required.
So what am I expected to do? Create a second Pull Request for collectd-6?
I guess it would be far easier to make that trivial change within a merge to -6 once my branch has been merged, no?

mrunge:
I'd update the PR and drop DS_TYPE_ABSOLUTE right now. We removed it, since it was not widely used, if it was used at all.

Anyways, that can be done in another PR, I'm good with merging this here.

Issue #3094, opened 2019-02-20
[Feature Request/Version 6.0]: Adding Labels to Collectd Metrics


srivasta:
*   Version of collectd: Targeted for 6.0
*   Operating system / distribution: Any/All
*   Kernel version (if applicable):

## Expected behavior

The data collected by collectd is essentially a time series (also known as a metric). Metrics have names that identify the time series (usually based on some combination of where they are being 2 we as, like hostname, or what the time series represents, like load average or memory usage). Currently, collectd uses a  fixed 3-5 tuples (three of them are fixed and mandatory, and only two fields are flexible and up to the plugin author) are too inflexible (a service might be better served by sharding metrics by component, or region, or other notional dimensions than the hard coded set in collectd). Using more descriptive, and user defined, set of labels  to create the identity of the metric might be more natural for users.  Labels make selecting, filtering and grouping much easier. Most modern monitoring systems have mechanisms to query for metrics based on arbitrary labels as long as the metrics carry the labels with them. This functionality is currently missing from the collectd data model, and this proposal is aimed at adding that. Examples are detailed in the background section

The proposed change is in a design document at https://docs.google.com/document/d/173gGP3tUD3yfN2NNHxCv0BsKsacfDlSyoaQIn7MqLtQ/edit?usp=sharing

The document is still a work in progress; I am modifying the wire protocol section to be in line with the https://openmetrics.io/ recommendations.

srivasta:
I have updated the design document with a new binary format section using the openmetrics suggested formats, and moved the previous binary protocl segment into the alternative considered section. the openmetrics standard is very much a work in progress, I have considered just the public facing information in this draft.

sunkuranganath:
Thanks Manoj for creating this.

srivasta:
https://docs.google.com/document/d/1RZ7YAO6lWcDDvevzxFGVDgv4UQpdgWoT7OCP9Vpb2CU/edit?usp=sharing

srivasta:
I have an initial stalking horse commit that implements the data structures in the design docs, and has unit tests to the avl tree used in the structures. I think we need some discussion on how to convert the current value_list_t 5-tupples --> name + labels structure in the new format. Also, how should we deal with metadata lists in the new format?

https://github.com/srivasta/collectd/tree/metric-label-support

srivasta:
I have an initial proposal for how to convert from the current value_list_t data structure to the proposed new single data source metrics_t format 

[...]

PR #3506, opened 2020-07-14
[Epic] Metric label support

octo:
This PR changes the way metrics are identified. Since this is a fundamental change, it is quite complex and large. I'll try to summarize how the new API works.

Each metric is identified by a *metric name* and a *label set* (a list of key/value pairs, keys must be unique). I'd like to add "resource labels", similar to [the OpenTelemetry "Resource" type](https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/resource/v1/resource.proto), but that's still to do.

## Data structures

There are two central data structures:

*   `metric_family_t` holds the metric type, metric name, and a list of metrics.
*   `metric_t` holds the value, a label set, collection time, interval, and a pointer to the parent *metric family*.

Read plugins generate and dispatch a `metric_family_t` to the daemon, write plugins receive `metric_family_t`. So in many regards, `metric_family_t` takes the role of the previous `value_list_t`.

Other functions take a `metric_t` instead, e.g. `uc_get_rate()` which converts a counter value to a rate, and some formatting / serialization functions.

## Plugin status

*→ [collectd 6 project plan](https://docs.google.com/spreadsheets/d/1ss4NJeJ00CwAmGgIRGQ2EJM9LFifPPrJcq_c0tGN_vY/edit#gid=741037669)*

The following plugins have been updated and work with label metric (on Linux):

*   cpu
*   memory
*   write\_http
*   write\_log
*   write\_stackdriver

There are 173 plugins total.

*   84 compile successfully
*   45 fail to compile
*   44 are unknown (can't build on my development machine)

## Compiling

For the time being, I recommend to compile with:

```sh
./configure --disable-all-plugins --enable-cpu --enable-memory --enable-write_http […]
#           ^^^^^^^^^^^^^^^^^^^^^
```

Issue: #3094


Makefile.am
```diff
@@ -439,28 +436,33 @@ libplugin_mock_la_SOURCES = \
 	src/daemon/utils_time.h
 
 libplugin_mock_la_CPPFLAGS = $(AM_CPPFLAGS) -DMOCK_TIME
-libplugin_mock_la_LIBADD = libcommon.la libignorelist.la $(COMMON_LIBS)
+libplugin_mock_la_LIBADD = libcommon.la libignorelist.la libmetadata.la $(COMMON_LIBS)
+#libplugin_mock_la_LIBADD = libcommon.la libignorelist.la $(COMMON_LIBS)
```
mrunge, line 440: Can we get rid of this line? Looks like an oversight.

octo: Done.

mrunge (approved):
Thank you!

Let's get this merged and handle any fall-out once we discover issues.

octo:
Thanks Matthias!

PR #3508, opened 2020-07-17
Implement metric_t and metric_family_t.

octo:
*   `metric_family_t` holds the metric type, metric name, and a list of metrics.
*   `metric_t` holds the value, a label set, collection time, interval, and a pointer to the parent *metric family*.

This another break-out from #3506, trying to make the change easier to review.

ChangeLog: collectd: The new `metric_t` and `metric_family_t` data structures hold metrics that are identified by a label set.

### DIFF
--- a/src/daemon/metric.c
+++ b/src/daemon/metric.c
@@ -284,7 +284,7 @@ char const *metric_label_get(metric_t const *m, char const *name) {
   return set->value;
 }
 
-int metric_list_add(metric_list_t *metrics, metric_t m) {
+static int metric_list_add(metric_list_t *metrics, metric_t m) {
   if (metrics == NULL) {
     return EINVAL;
   }
@@ -314,6 +314,20 @@ int metric_list_add(metric_list_t *metrics, metric_t m) {
   metrics->num++;
 
   return 0;
+}
+
+static void metric_list_reset(metric_list_t *metrics) {
+  if (metrics == NULL) {
+    return;
+  }
+
+  for (size_t i = 0; i < metrics->num; i++) {
+    metric_reset(metrics->ptr + i);
+  }
+  free(metrics->ptr);
+
+  metrics->ptr = NULL;
+  metrics->num = 0;
 }
 
 static int metric_list_clone(metric_list_t *dest, metric_list_t src,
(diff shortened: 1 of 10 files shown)


src/daemon/metric.h
```diff
@@ -0,0 +1,185 @@
...
+ * Metric
+ */
+/* forward declaration since metric_family_t and metric_t refer to each other.
+ */
+struct metric_family_s;
+typedef struct metric_family_s metric_family_t;
+
+/* metric_t is a metric inside a metric family. */
+typedef struct {
+  metric_family_t *family; /* for family->name and family->type */
```
srivasta, line 93: /* The metric struct/object is coupled with the metric family, and I can't imagine very many cases of altering this coupling (which variant of the value union is to be used depends on the metric_family. Are there other reasons for the creation of a different structure? */

srivasta, line 93: /* Pointer to the containing parent metric)family_t for family->name and family->type */

src/daemon/metric.h
```diff
@@ -0,0 +1,185 @@
...
+  value_t value;
+  cdtime_t time; /* TODO(octo): use ms or µs instead? */
+  cdtime_t interval;
+  /* TODO(octo): target labels */
+  meta_data_t *meta;
+} metric_t;
+
+/* metric_identity writes the identity of the metric "m" to "buf". An example
+ * string is:
+ *
```
srivasta, line 105: Is the metric identity specification documented somewhere?

src/daemon/metric.c
```diff
@@ -0,0 +1,675 @@
...
+
+int value_marshal_text(strbuf_t *buf, value_t v, metric_type_t type) {
+  switch (type) {
+  case METRIC_TYPE_GAUGE:
+  case METRIC_TYPE_UNTYPED:
+    return strbuf_printf(buf, GAUGE_FORMAT, v.gauge);
+  case METRIC_TYPE_COUNTER:
+    return strbuf_printf(buf, "%" PRIu64, v.counter);
+  default:
+    ERROR("Unknown metric value type: %d", type);
```
srivasta, line 53: At this point we don't know what the value type is, and we can't be sure that it is an int. 
The printf function uses the format specifiers to figure out what to pop off the stack. So when it sees %d, it pops off 4 bytes and interprets them as an int, which is wrong (the binary representation of (float)3.0 is not the same as (int)3).

You'll need to either use the %f format specifiers or cast the arguments to int.

src/daemon/metric.h
octo: I added "backreference" to make clear that metric_family_t and metric_t refer to reach other. I don't think we need to explain *what* ("Pointer to […] metric_family_t"), just *why* this is here.

src/daemon/metric.h
octo: Added a note that this uses the OpenMetrics / Prometheus exposition format.

src/daemon/metric.c
octo: Well no, but actually yes.

This line is not printing the value, but the value's type. `type` is a `metric_type_t`, which is an `enum`. [Section 6.7.2.2](http://c0x.coding-guidelines.com/6.7.2.2.html) of the C11 standard states:

> The  identifiers  in  an  enumerator  list  are  declared  as  constants  that  have  type `int` and may   appear   wherever   such are permitted.

*and*

> Each enumerated type shall be compatible with char, a signed integer type, or an unsigned integer type.

If I read this correctly, then `size(METRIC_TYPE_GAUGE) == 4` because the *identifier*, i.e. `METRIC_TYPE_GAUGE`, is treated like an `int`. However, `sizeof((metric_type_t){METRIC_TYPE_GAUGE})`, which allocates a `metric_type_t` on the stack and takes its size, may yield other sizes, e.g. 1 or 2 bytes, because the *type*, i.e. `metric_type_t`, may have different storage sizes.

I've added a cast to `int` to be on the save side, though I think that compilers will likely warn about a problem caused by printf format and argument mismatch.

src/daemon/metric.h
```diff
@@ -0,0 +1,185 @@
...
+ * ENOENT. The returned pointer may not be valid after a subsequent call to
+ * "metric_label_set". */
+char const *metric_label_get(metric_t const *m, char const *name);
+
+/* metric_reset frees all labels and meta data stored in the metric and resets
+ * the metric to zero. */
+int metric_reset(metric_t *m);
+
+/* metric_marshal_text writes a text representation of "m" to the buffer "buf".
+ * An example output is:
```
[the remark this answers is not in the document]
octo: This one is a bit trickier, because it doesn't implement all that OpenMetrics / Prometheus specify, just what `metric_marshal_text` produces.

`metric_identity` is used, e.g. to look up the metric in the cache (see `src/daemon/utils_cache.c`). Arguably `metric_marshal_text` and `metric_family_unmarshal_text` could be moved to an OpenMetrics / Prometheus specific plugin. What do you think?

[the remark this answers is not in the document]
srivasta: I like that, since that leaves open the possibility of marshaling/unmarshaling to a different wire protocol in the future, if needed, without increasing complexity a whole lot. And we can add a comment in the documentation in the plugin poninting to the spec, and stating that only the relevant bitsof the spec are implemented here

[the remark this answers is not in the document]
octo: I removed these two functions – well, partially. I added `metric_parse_identity()`, which is used e.g. when a metric expires: we get the string representation from the cache and need to convert it to a `metric_t*` or `metric_family_t*`.

PR #58, opened 2012-03-18
Protocols plugin: Add Solaris support.

octo:
ChangeLog: protocols plugin: Add Solaris support

Import of Cosmin Ioiart's changes.

PR #3515, opened 2020-08-04
Protocol plugin solaris support

dago:
ChangeLog: protocols plugin: Add Solaris support

This is a forward-port of PR #58

### DIFF
--- a/src/protocols.c
+++ b/src/protocols.c
@@ -22,6 +22,7 @@
  *
  * Authors:
  *   Florian octo Forster <octo at collectd.org>
+ *   Cosmin Ioiart <cioiart at gmail.com>
  **/
 
 #include "collectd.h"
@@ -30,13 +31,25 @@
 #include "utils/common/common.h"
 #include "utils/ignorelist/ignorelist.h"
 
-#if !KERNEL_LINUX
+#if !(KERNEL_LINUX || KERNEL_SOLARIS)
 #error "No applicable input method."
 #endif
 
+#if KERNEL_LINUX
 #define SNMP_FILE "/proc/net/snmp"
 #define NETSTAT_FILE "/proc/net/netstat"
+#endif
 
+/*
+ * On Solaris, all the key/value pairs are read via kstat
+ */
+#if HAVE_KSTAT_H
+#include <kstat.h>
+#endif
+
+#if KERNEL_SOLARIS
+#include <string.h>
+#endif
 /*
  * Global variables
  */
@@ -64,6 +77,7 @@ static void submit(const char *protocol_name, const char *str_key,
 
   vl.values = &value;
   vl.values_len = 1;
+  sstrncpy(vl.host, hostname_g, sizeof(vl.host));
   sstrncpy(vl.plugin, "protocols", sizeof(vl.plugin));
   sstrncpy(vl.plugin_instance, protocol_name, sizeof(vl.plugin_instance));
   sstrncpy(vl.type, "protocol_counter", sizeof(vl.type));
@@ -71,6 +85,8 @@ static void submit(const char *protocol_name, const char *str_key,
 
   plugin_dispatch_values(&vl);
 } /* void submit */
+
+#if KERNEL_LINUX
 
 static int read_file(const char *path) {
   FILE *fh;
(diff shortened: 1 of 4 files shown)


dago:
CI seems to be broken as git repo access itself fails.

mrunge:
CI failed, because it could not git clone from github. That should be temporary.

dago:
In the code the function `get_kstat_value_to_string` is used, it was already in the original PR #58 and the implementation neither exists in collectd nor Solaris libkstat - I don't even get a single hit on Google! @octo can you remember how this was meant? Probably this should be rewritten to use functions from libkstat instead.

src/protocols.c
```diff
@@ -64,6 +77,7 @@ static void submit(const char *protocol_name, const char *str_key,
 
   vl.values = &value;
   vl.values_len = 1;
+  sstrncpy(vl.host, hostname_g, sizeof(vl.host));
```
octo, line 80: This is done automatically by *plugin_dispatch_values()* is the field is empty.

src/protocols.c
```diff
@@ -30,13 +31,25 @@
...
+
+/*
+ * On Solaris, all the key/value pairs are read via kstat
+ */
+#if HAVE_KSTAT_H
+#include <kstat.h>
+#endif
 
+#if KERNEL_SOLARIS
+#include <string.h>
```
octo, line 51: `<string.h>` is already included via `"collectd.h"`.

src/protocols.c
```diff
@@ -170,19 +186,75 @@ static int read_file(const char *path) {
...
+  kstat_t *ksp_chain = NULL;
+
+  if (mod_name == NULL)
+    return -1;
+
+  if (kc == NULL)
+    return -1;
+  for (ksp_chain = kc->kc_chain; ksp_chain != NULL;
+       ksp_chain = ksp_chain->ks_next) {
+    if (strncmp(ksp_chain->ks_module, mod_name, sizeof(ksp_chain->ks_module)) ==
```
octo, line 207: Is `ks_module` terminated with a nul byte? If so, use *strcmp()* instead -- *strncmp()* implies (to me) that the code is trying to do a prefix match.

src/protocols.c
```diff
@@ -170,19 +186,75 @@ static int read_file(const char *path) {
...
+            0 &&
+        ksp_chain->ks_type == KSTAT_TYPE_NAMED) {
+      kstat_named_t *kn = NULL;
+      kstat_read(kc, ksp_chain, kn);
+      kn = (kstat_named_t *)ksp_chain->ks_data;
+      for (int i = 0; (kn != NULL) && (i < ksp_chain->ks_ndata); i++, kn++) {
+        char value[16];
+        char name[256];
+
+        if (kn == NULL) {
```
octo, line 217: This is never true, because `kn != NULL` is already asserted by the condition in the `for` loop.

src/protocols.c
```diff
@@ -170,19 +186,75 @@ static int read_file(const char *path) {
...
+       ksp_chain = ksp_chain->ks_next) {
+    if (strncmp(ksp_chain->ks_module, mod_name, sizeof(ksp_chain->ks_module)) ==
+            0 &&
+        ksp_chain->ks_type == KSTAT_TYPE_NAMED) {
+      kstat_named_t *kn = NULL;
+      kstat_read(kc, ksp_chain, kn);
+      kn = (kstat_named_t *)ksp_chain->ks_data;
+      for (int i = 0; (kn != NULL) && (i < ksp_chain->ks_ndata); i++, kn++) {
+        char value[16];
+        char name[256];
```
octo, line 215: I recommend initializing these buffers to zero, e.g.:

```c
char name[256] = "";
```

That way, if *get_kstat_value_to_string()* does not initialize the buffer due to a failure, the following code will still behave in a well-defined manner.

octo (commented):
Thanks @dago, this looks mostly good, just some nit-picky coding style remarks :)

src/protocols.c
dago: OK, removed, I didn't review this part and just forward-ported it - yeah, no excuse!

dago:
@octo This works for me now, but two CI builds are failing, probably not because of my changes.
Would you mind taking a look at my code changes?

PR #3522, opened 2020-08-14
Create a benchmark for the distribution

Lana243:
ChangeLog: distbench: Distribution benchmark was implemented. The `run-all.sh` script runs all the implementations and create the necessary csv files


distbench/Makefile.am
```diff
@@ -1,7 +1,11 @@
...
 
 # If you need special libraries, e.g. libm (math lib), declare them like so:
-# xxx_LIBS = -lm
+bkjg_LDFLAGS = -lm
+
+margalit_LDFLAGS = -lm
+
+sshmidt_LDFLAGS = -lm
 
 bkjg_SOURCES = main.c function.h bkjg.c
```
octo, line 10: The `function.h` file has been deleted. The "release ready" check fails, because it it still mentioned here.

Lana243: Oh, thanks, I see. I fixed it

PR #3526, opened 2020-08-17
Improve SIGCHLD handling

efuss:
ChangeLog: SIGCHLD handling has been improved, allowing Perl and Python plugins to gather child processes exit status.

Don't explicitly (`SIG_IGN`) ignore the `SIGCHLD` signal. Contrary to implicitly (`SIG_DFL`) ignoring it, explicitly igoring `SIGCHLD` causes `wait()` et. al. to always raise `ECHILD`.
This behaviour seems neither to be widely known nor prominently documented everywhere. On NetBSD, wait(2) documents it; in POSIX, it's hidden in the `_exit`/`_Exit` description.

In the exec plugin, install a sigaction rather than a traditional handler.
When signalled, inspect the `siginfo's` `si_pid` field and only act on the signal and reap the child's status if the pid is one of the program list's pids.

This allows other plugins that fork child processes (Perl and Python plugins) to get their children's status via `wait()` et. al.

I fail to track down where the `SIG_IGN` was introduced so I can't find a possible rationale for it.

Child process status handling still appears dodgy to me. In `exec_read_one()`, there's a `waitpid()` which seems to race with the `SIGCHLD` handler; I fail to see where the `pl->status` field set by the `SIGCHLD` handler gets examined or acted upon.

This change has only been lightly tested. The Perl plugin is now able to correctly retrieve a sub-processes status; some artificial programs called by the Exec plugin (exit successfully/unsuccessfully; loop) appear to behave as before. Someone really using the Python or Exec plugins should double-check this.

### DIFF
--- a/src/daemon/cmd.c
+++ b/src/daemon/cmd.c
@@ -170,7 +170,7 @@ int main(int argc, char **argv) {
   /*
    * fork off child
    */
-  struct sigaction sig_chld_action = {.sa_handler = SIG_IGN};
+  struct sigaction sig_chld_action = {.sa_handler = SIG_DFL};
 
   sigaction(SIGCHLD, &sig_chld_action, NULL);
 
--- a/src/exec.c
+++ b/src/exec.c
@@ -97,8 +97,8 @@ static pthread_mutex_t pl_lock = PTHREAD_MUTEX_INITIALIZER;
 /*
  * Functions
  */
-static void sigchld_sigaction(int __attribute__((unused)) sig, siginfo_t *info, void __attribute__((unused)) *ctx)
-{
+static void sigchld_sigaction(int __attribute__((unused)) sig, siginfo_t *info,
+                              void __attribute__((unused)) * ctx) {
   program_list_t *pl;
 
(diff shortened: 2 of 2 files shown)


efuss:
The Travis CI build fails with
> ../../src/daemon/collectd.c:339:10: error: duplicate case value '102'

which is a place I didn't touch.

kwiatrox:
@efuss, problem with CI was caused by another issue. It is fixed by #3536.

efuss:
Closing to re-trigger the Travis CI build.

efuss:
Re-opening to (hopefully) re-trigger the Travis CI build.

efuss:
Ping? I'm running this (with both perl plugins using pipes and exec plugins concurrently) ever since.

eero-t:
> Ping? I'm running this (with both perl plugins using pipes and exec plugins concurrently) ever since.

This PR was before I was involved in this project, but could you rebase it to `main`?

Or port it to the collectd-6.0 branch, and its (greatly refactored) plugin version?

germnansk (approved):

PR #3531, opened 2020-08-19
Add support for the distribution metrics to the collectd core

bkjg:
ChangeLog:  Distribution Metrics: add support for distribution metrics to the collectd core.

Addded support for the distribution metrics to the uitls_cache.[ch] files. Implemented uc_get_percentile, uc_get_percentile_by_name, uc_get_last_time, uc_get_last_update functions. Added distribution metric structure to the cache_entry_s structure. Added distribution_sub function to the distribution.[ch]. Added support for distribution metrics to the uc_update function, uc_insert, uc_get_rate and uc_get_rate_by_name functions. Added fixes in uc_update for checking null pointers.


Makefile.am
```diff
@@ -151,6 +151,7 @@ check_LTLIBRARIES = \
 
 
 check_PROGRAMS = \
+    test_utils_cache \
```
Lana243, line 154: Small issue with tab sizes. You can run the format script to fix it 😊

src/daemon/distribution.c
```diff
@@ -336,3 +336,49 @@ double distribution_squared_deviation_sum(distribution_t *dist) {
...
+  }
+
+  if (d1->num_buckets != d2->num_buckets) {
+    return EINVAL;
+  }
+
+  pthread_mutex_lock(&d1->mutex);
+  pthread_mutex_lock(&d2->mutex);
+
+  if (d1->total_sum < d2->total_sum) {
```
Lana243, line 353: I suppose it would be easier to consider that the first argument is always greater than the second one in order to avoid cases analysis and code duplication

Lana243 (commented):
Great job, Barbara!

Makefile.am
bkjg: I fixed it! Thank you!

src/daemon/distribution.c
bkjg: There is a possibility that the distribution will be reseted and this function should be able to handle this case. The first argument is the distribution that was updated more recently, however due to the reset, it still could have smaller counters.

PR #3543, opened 2020-08-27
Fix network test

dago:
ChangeLog: Avoid empty initialization when gcrypt is absent as it leads to compile error on Sun Studio

### DIFF
--- a/src/network_test.c
+++ b/src/network_test.c
@@ -224,17 +224,7 @@ static int decode_string(char const *in, uint8_t *out, size_t *out_size) {
 }
 
 DEF_TEST(parse_packet) {
-  sockent_t se
-#if HAVE_GCRYPT_H
-      = {.data.server =
-             (struct sockent_server){
-                 .cypher = NULL,
-                 .userdb = NULL,
-                 .security_level = SECURITY_LEVEL_NONE,
-             },
-        }
-#endif
-  ;
+  sockent_t se = {0};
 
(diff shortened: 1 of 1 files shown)


src/network_test.c
```diff
@@ -224,16 +224,17 @@ static int decode_string(char const *in, uint8_t *out, size_t *out_size) {
 }
 
 DEF_TEST(parse_packet) {
-  sockent_t se = {
-      .data.server =
-          (struct sockent_server){
+  sockent_t se
```
kwiatrox, line 227: It is not fully equivalent as `se`  is not nulled and `.data_server` can contain random data.
You can use memset to zero the `se` or try using the multi layer default initializer to circumvent the compilation errors.
Something like:
```
#if HAVE_GCRYPT_H
se = //current code
#else
se = {{}}; // or se = {{{}}} whatever is necessary;
#endif
```

octo: ACK. Looking into this, we don't need to differentiate between the with and without gcrypt cases. Simply initializing to `{0}` should be sufficient. The gcrypt specific code only initialized the gcrypt fields to zero, so that can easily be removed.

I pushed an additional commit simplifying that.

octo (approved):

dago:
LGTM and works cleanly in my picky environment, Thanks @octo ! I tend to make my changes with minimal changes, probably I should aim for a more holistic approach like your fix 👍

PR #3545, opened 2020-08-27
Extended format_json.c to map distribution metrics to json output.

elene-margalit:
ChangeLog: Extended function format_metric to call the new function format_metric_distribution if the argument is a distribution. Function format_metric_distribution formats the distribution metric to into json according to the prom2json mapping of Histograms to json.


src/daemon/metric.c
```diff
@@ -41,13 +43,64 @@
 /* Metric names must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*` */
 #define VALID_NAME_CHARS VALID_LABEL_CHARS ":"
 
+int distribution_count_marshal_text(strbuf_t *buf, distribution_t *dist) {
+
```
bkjg, line 47: In your opinion: deleting these empty lines before after function name would make code more readable.

bkjg (commented):
It looks great, Elene!

src/daemon/metric.c
elene-margalit: Thank you Barbara for your suggestion, I agree with you.

PR #3547, opened 2020-08-28
Fix memory leak in curl plugin

bkjg:
When AddressFamily field was set in config file in configuration for curl plugin, memory was allocated and pointer was stored in af variable. At the end of block of code, this memory wasn't freed. 

ChangeLog: curl plugin: Fix memory leak

### DIFF
--- a/src/curl.c
+++ b/src/curl.c
@@ -467,6 +467,7 @@ static int cc_config_add_page(oconfig_item_t *ci) /* {{{ */
                 child->key, page->instance);
         status = -1;
       }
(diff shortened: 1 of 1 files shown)


kwiatrox (approved):
Thank you for the fix @bkjg , LGTM!

octo:
Thank you so much Barbara! I've cherry-picked this fix into the *collectd-5.11* branch as 318257446c995869bdf9c5fb51e1903b0e6e7b08.

PR #3551, opened 2020-08-31
Finalized formatting distribution metrics to JSON

elene-margalit:
ChangeLog: Changed function distribution_marshal_text to format distribution metrics according to Prometheus text format of Histograms. Adapted unit test of this function to the new changes.


octo (approved):
Thank you Elene!

PR #3559, opened 2020-09-01
daemon: Fix the build on Solaris.

octo:
It caused problems on Solaris, which also has a "label_t" type and was not used
in the API.


octo:
@dago The main problem appears to be that *strnlen(3)* is not found by the linker. [The Oracle docs](https://docs.oracle.com/cd/E86824_01/html/E54766/strnlen-3c.html) give me the impression that it should in principle exist on Solaris. Do you know what I need to do to get this to work?

dago:
@octo The linked docs are for Solaris 11.3 while the Solaris CI runs on Solaris 10 which does not have `strnlen`. That means a replacement like the one from Gnulib is needed https://www.gnu.org/software/gnulib/manual/html_node/strnlen.html

octo:
So, if *strnlen(3)* is not available in the standard library, it may be necessary to link with gnulib, similarly to this special case for the *socket* call? https://github.com/collectd/collectd/blob/main/configure.ac#L860

dago:
To my understanding Gnulib was only meant to be linked on MinGW and that it was also required to avoid Gnulib linkage due to license issues. But yes: the absence of `strnlen` could be solved easily with Gnulib as could be most of the other replacement functions in `src/utils/common/common.c`. Alternatively a replacement implementation of `strnlen` could be provided in `common.c`.

octo:
Alright, I have added a very simple *strnlen* implementation. It now fails with:

```
"/opt/csw/include/statgrab.h", line 118: warning: dubious reference to struct typedef: sg_error_details
"/opt/csw/include/statgrab.h", line 119: warning: dubious reference to enum typedef: sg_error
"/opt/csw/include/statgrab.h", line 120: warning: dubious reference to struct typedef: sg_error_details
"/opt/csw/include/statgrab.h", line 194: warning: dubious reference to struct typedef: sg_cpu_stats
"/opt/csw/include/statgrab.h", line 194: warning: dubious reference to struct typedef: sg_cpu_stats
"/opt/csw/include/statgrab.h", line 213: warning: dubious reference to enum typedef: sg_cpu_percent_source
"/opt/csw/include/statgrab.h", line 214: warning: dubious reference to struct typedef: sg_cpu_stats
"/opt/csw/include/statgrab.h", line 307: warning: dubious reference to struct typedef: sg_fs_stats
"/opt/csw/include/statgrab.h", line 307: warning: dubious reference to struct typedef: sg_fs_stats
"/opt/csw/include/statgrab.h", line 323: warning: dubious reference to struct typedef: sg_disk_io_stats
"/opt/csw/include/statgrab.h", line 323: warning: dubious reference to struct typedef: sg_disk_io_stats
"/opt/csw/include/statgrab.h", line 344: warning: dubious reference to struct typedef: sg_network_io_stats
"/opt/csw/include/statgrab.h", line 344: warning: dubious reference to struct typedef: sg_network_io_stats
"/opt/csw/include/statgrab.h", line 384: warning: dubious reference to struct typedef: sg_page_stats
"/opt/csw/include/statgrab.h", line 384: warning: dubious reference to struct typedef: sg_page_stats
"/opt/csw/include/statgrab.h", line 461: warning: dubious reference to enum typedef: sg_process_count_source
"/opt/csw/include/statgrab.h", line 462: warning: dubious reference to struct typedef: sg_process_stats
"src/daemon/collectd.c", line 64: syntax error before or at: {
"src/daemon/collectd.c", line 66: undefined symbol: str
"src/daemon/collectd.c", line 66: cannot dereference non-pointer type
"src/daemon/collectd.c", line 67: warning: improper pointer/integer combination: arg #1
"src/daemon/collectd.c", line 89: undefined symbol: str
"src/daemon/collectd.c", line 89: warning: improper pointer/integer combination: op "="
"src/daemon/collectd.c", line 90: warning: improper pointer/integer combination: arg #2
"src/daemon/collectd.c", line 90: warning: improper pointer/integer combination: arg #2
"src/daemon/collectd.c", line 90: warning: improper pointer/integer combination: arg #2
"src/daemon/collectd.c", line 135: warning: using out of scope declaration: init_hostname
"src/daemon/collectd.c", line 235: warning: using out of scope declaration: setlocale
"src/daemon/collectd.c", line 250: warning: using out of scope declaration: sg_init
"src/daemon/collectd.c", line 255: warning: using out of scope declaration: sg_str_error
"src/daemon/collectd.c", line 255: warning: using out of scope declaration: sg_get_error
"src/daemon/collectd.c", line 259: warning: using out of scope declaration: sg_drop_privileges
"src/daemon/collectd.c", line 260: warning: using out of scope declaration: sg_str_error
"src/daemon/collectd.c", line 260: warning: using out of scope declaration: sg_get_error
"src/daemon/collectd.c", line 272: undefined symbol: loop
"src/daemon/collectd.c", line 391: undefined symbol: loop
"src/daemon/collectd.c", line 64: warning: static function called but not defined: init_hostname()
cc: acomp failed for src/daemon/collectd.c
gmake[1]: *** [Makefile:8684: src/daemon/collectd-collectd.o] Error 2
```

It looks to me as if the *libstatgrab* header is to blame here, as the code the compiler complains about is "immediately" following that include and I don't see anything wrong with the code the compiler points to.

octo:
Alright, the daemon builds now. It currently fails at the *aggregation plugin*, which is expected.

@dago, could you add the `-k` flag to `make` so it keeps building despite errors? That will make it much easier to address build issues.

Also, is it possible to exclude plugins that are known to not build? This is basically only relevant in the *collectd-6.0* branch.

dago:
@octo The general scheduler now has an additional `-k` flag added to `make` and there is also a special scheduler dedicated to the `collectd-6.0` branch which has an additional `--disable-zookeeper` just as example added. Just drop me a note about the specific configure options so I can adjust them properly.

dago:
I did some checking and what does not work is applying the special configure arguments for the `collectd-6.0` branch to pull requests, even when they are created against that branch. The reason is that from the changeset in buildbot the applied branch is set to the PR. At the moment I don't have a good solution for this.

octo:
Would it be possible to add flags from a file in the repository? E.g. (assumes bash):

```bash
./configure $(test -e solaris-ci.flags && egrep '^--((enable|disable)-([a-z_-]+))$' solaris-ci.flags)
```

This would allow us to disable known-to-be-broken plugins similar to the lists we have in `.cirrus.yml` and `.travis.yml`: https://github.com/collectd/collectd/blob/7e20042fef419ffa6fcca2468055967fda1884ec/.cirrus.yml#L5-L48

dago:
I know added specific steps in the CI build to honour `solaris-ci.flags`. Can you give it a try to verify that it actually works as expected?

dago:
There were still some issues with the argument extraction which should be fixed now. I'll keep an eye on the commits and the CI status of this PR.

octo:
Thanks @dago, let's see if the new commit works 👍

octo:
Looks like the builder picked up the flags file. I'll tweak it when I have time to hopefully "fix" the Solaris build of the *collectd-6.0* branch.

octo:
Hi @dago,

this mostly seems to work, thanks! The `--disable-target_v5upgrade` line seems to be ignored, maybe because the digit in the plugin name?

The only remaining issue is that Solaris apparently prints `NAN` as `-nan` instead of `nan`, breaking two unit tests. I don't know how I can fix this, unfortunately :\

Best regards,
—octo

dago:
Hi @octo,

I adjusted the option filter string so numbers are allowed too now. Regarding the NaN issue I don't have a good solution either - I also faced the issue in the testsuite from rrdtool.

Best regards -- Dago

octo:
It works! Could you please review @dago?

dago:
Sure, anything specific that I should verify? Or just a basic test getting it to run on Solaris with some basic metrics?

octo:
@dago Just need an approving review to submit. If you'd check that I didn't do anything obviously stupid that'd be appreciated, though I'd like to think that's unlikely ;)

src/utils/strbuf/strbuf.c
```diff
@@ -200,6 +200,17 @@ int strbuf_printf(strbuf_t *buf, char const *format, ...) {
   return 0;
 }
 
+#if !HAVE_STRNLEN
+static size_t strnlen(const char *s, size_t maxlen) {
```
dago, line 204: This is already in `common.c`, why is this defined again here?

src/utils/format_graphite/format_graphite.c
```diff
@@ -35,15 +35,23 @@
 /* Utils functions to format data sets in graphite format.
  * Largely taken from write_graphite.c as it remains the same formatting */
 
-static int gr_format_values(strbuf_t *buf, metric_t const *m, gauge_t rate) {
-  if (m->family->type == DS_TYPE_GAUGE)
-    return strbuf_printf(buf, GAUGE_FORMAT, m->value.gauge);
-  else if (rate != -1)
-    return strbuf_printf(buf, "%f", rate);
```
dago, line 42: This case seems to be missing in the new code, is this intended?

src/daemon/plugin.c
```diff
@@ -1779,14 +1772,10 @@ EXPORT int plugin_read_all_once(void) {
     old_ctx = plugin_set_ctx(rf->rf_ctx);
 
     if (rf->rf_type == RF_SIMPLE) {
-      int (*callback)(void);
-
-      callback = rf->rf_callback;
+      int (*callback)(void) = (void *)rf->rf_callback;
```
dago, line 1775: Shouldn't the cast be for the exact type instead of `(void *)` ?

src/daemon/plugin.c
```diff
@@ -1707,13 +1704,9 @@ EXPORT int plugin_init_all(void) {
...
-    callback_func_t *cf;
-    plugin_init_cb callback;
-    plugin_ctx_t old_ctx;
-
-    cf = le->value;
-    old_ctx = plugin_set_ctx(cf->cf_ctx);
-    callback = cf->cf_callback;
+    callback_func_t *cf = le->value;
+    plugin_ctx_t old_ctx = plugin_set_ctx(cf->cf_ctx);
+    plugin_init_cb callback = (void *)cf->cf_callback;
```
dago, line 1709: Same here as below regarding the type: shouldn't the cast be more specific instead of `(void *)`?

src/daemon/metric.h
```diff
@@ -64,16 +64,6 @@ typedef struct {
...
-/* label_t represents a key/value pair. It is similar to label_pair_t, except
- * that is has const fields. label_t is used in function arguments to prevent
- * the called function from modifying its argument. Internally labels are
- * stored as label_pair_t to allow modification, e.g. by targets in the "filter
- * chain". */
-typedef struct {
-  char const *name;
-  char const *value;
-} label_t;
-
```
dago, line 76: Why have there been two types `label_pair_t` and `label_t` when they have the same fields? But I guess it is ok to get rid of `label_t` and replace with `label_pair_t` in all occasions like you did.

src/daemon/metric.c
```diff
@@ -66,7 +66,10 @@ static label_pair_t *label_set_read(label_set_t labels, char const *name) {
     return NULL;
   }
 
-  label_t label = {
+  struct {
+    char const *name;
+    char const *value;
+  } label = {
```
dago, line 72: Why is this explicit instead of `label_pair_t`?

src/daemon/metric.h
octo: `label_t`'s fields were `const`. It was used as an argument to signal that the argument would not be modified. Almost all use-cases outside of unit tests had been removed though.

src/daemon/plugin.c
octo: The C standard says that all pointer types can be case to `void*`, and `void*` can be cast to any pointer type. Direct pointer conversion sometimes leads to compilers complaining, so I felt it safest to use the standard-approved method.

In retrospect, it didn't remove the warnings emitted by the Sun compiler, and I think it didn't have any effect on the result. If you prefer, I can revert those changes …?

src/daemon/plugin.c
octo: Discussed above.

src/utils/format_graphite/format_graphite.c
octo: Yes. This code previously (erroneously) used `DS_TYPE_…` instead of the values defined by the `metric_type_t` enum, e.g. `METRIC_TYPE_GAUGE`. This has been fixed in the process.

src/utils/strbuf/strbuf.c
octo: *strbuf* can't easily depend on *common*, because that would result in a circular dependency.

The cleanest solution would –imho– be to create a new package containing only the string functions, e.g. `strjoin` and friends. I didn't want to blow this change up though. How about we file that as a feature request and leave as is for now?

src/daemon/metric.c
octo: `name` is a `char const *` and can't be assigned to `(label_pair_t).name`, which is a `char *` (i.e. not const). I preferred to keep the argument `const`. Should I add a comment to clarify?

octo:
Hi @dago, thank you for the thorough review! I think I've responded to all your comments. Please let me know which changes you'd like to see :+1: 

Best regards, —octo

src/utils/strbuf/strbuf.c
dago: Ok, sounds reasonable.

src/daemon/plugin.c
dago: As this is for enhances compatibility this sounds reasonable.

src/daemon/metric.c
dago: That is fine, just wondering.

dago (approved):
Looks good to merge

PR #3560, opened 2020-09-01
Rewrite problematic language where possible

mrunge:
This patch leaves out rewrite master/slave from mysql and also
master from snmp.

This repaces blacklist with blocklist, whitelist with allowlist.

ChangeLog: collectd: Fix problematic use of language where possible.

### DIFF
--- a/docs/BUILD.dpdkstat.md
+++ b/docs/BUILD.dpdkstat.md
@@ -217,7 +217,7 @@ See also: http://dpdk.org/doc/guides/prog_guide/multi_proc_support.html
 
  *  The same PCI device configuration should be passed to the primary process as
     the secondary process uses the same port indexes as the primary.
- *  A deny-list / allow-list of NICs isn't supported yet.
+ *  A blocklist / allowlist of NICs isn't supported yet.
  *  Plugin initialization time depends on read interval. It requires 5 read
     cycles to set up internal buffers and states. During that time no statistics
     are submitted.
--- a/docs/BUILD.dpdkstat.md
+++ b/docs/BUILD.dpdkstat.md
@@ -217,7 +217,7 @@ See also: http://dpdk.org/doc/guides/prog_guide/multi_proc_support.html
 
  *  The same PCI device configuration should be passed to the primary process as
     the secondary process uses the same port indexes as the primary.
- *  A blacklist / whitelist of NICs isn't supported yet.
+ *  A deny-list / allow-list of NICs isn't supported yet.
  *  Plugin initialization time depends on read interval. It requires 5 read
     cycles to set up internal buffers and states. During that time no statistics
     are submitted.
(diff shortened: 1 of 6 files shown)


docs/BUILD.dpdkstat.md
```diff
@@ -217,7 +217,7 @@ See also: http://dpdk.org/doc/guides/prog_guide/multi_proc_support.html
 
  *  The same PCI device configuration should be passed to the primary process as
     the secondary process uses the same port indexes as the primary.
- *  A blacklist / whitelist of NICs isn't supported yet.
+ *  A deny-list / allow-list of NICs isn't supported yet.
```
igorpeshansky, line 220: I believe the prevalent terminology is `s/blacklist/blocklist/`, and `s/whitelist/allowlist/` (no dashes in "allowlist") — see, e.g., [this article](https://www.theregister.com/2019/09/03/chromium_microsoft_offensive/).

mrunge: Thank you for the quick review, I'll update the PR

igorpeshansky (dismissed):
LGTM :shipit:

mrunge:
For master/slave (for mysql), would everybody agree with using primary/replica.

This is a change included in mariadb https://mariadb.com/kb/en/standard-replication, I've added that to this change here as well.

src/mysql.c
```diff
@@ -463,40 +463,40 @@ static int mysql_read_slave_stats(mysql_database_t *db, MYSQL *con) {
...
     mysql_free_result(res);
     return -1;
   }
 
-  if (db->slave_stats) {
+  if (db->replica_stats) {
     unsigned long long counter;
     double gauge;
 
-    gauge_submit("bool", "slave-sql-running",
```
octo, line 477: Changing the metric names breaks backwards compatibility. Can we ear-mark removing this language from the metric identity for version 6 please?

octo (changes requested):

octo:
Thank you so much for working on this Matthias! Unfortunately changing the metric identity is problematic, as it breaks backwards compatibility. In the upcoming version 6 release we should be able to replace those terms in the metric identities, too.

src/mysql.c
```diff
@@ -193,11 +193,11 @@ static int mysql_config_database(oconfig_item_t *ci) /* {{{ */
     else if (strcasecmp("ConnectTimeout", child->key) == 0)
       status = cf_util_get_int(child, &db->timeout);
     else if (strcasecmp("MasterStats", child->key) == 0)
-      status = cf_util_get_boolean(child, &db->master_stats);
+      status = cf_util_get_boolean(child, &db->primary_stats);
```
mrunge, line 196: I shouldn't change this here, since it's still used in mysql/mariadb in older releases.

src/mysql.c
mrunge: you are right, this is fixed in the next commit.

octo (approved):
Thanks @mrunge!

PR #3562, opened 2020-09-02
Ping plugin

Lana243:
Ping plugin used to calculate an average latency. I modified that plugin to use a distribution_t for calculating latency instead.

### DIFF
--- a/src/daemon/distribution.c
+++ b/src/daemon/distribution.c
@@ -343,6 +343,22 @@ double distribution_squared_deviation_sum(distribution_t *dist) {
   return squared_deviation_sum;
 }
 
+double distribution_stddev(distribution_t *dist) {
+    if (dist == NULL || dist->num_buckets == 0) {
+        errno = EINVAL;
+        return NAN;
+    }
+    if (dist->num_buckets == 1) {
+        return 0.0;
+    }
+    pthread_mutex_lock(&dist->mutex);
+    double stddev = sqrt(((((double)dist->num_buckets) * dist->total_square_sum) -
+                             (dist->total_sum * dist->total_sum)) /
+                            ((double)(dist->num_buckets * (dist->num_buckets - 1))));
+    pthread_mutex_unlock(&dist->mutex);
+    return stddev;
+}
+
 int distribution_reset(distribution_t *dist) {
   if (dist == NULL) {
     return errno = EINVAL;
--- a/src/daemon/distribution.c
+++ b/src/daemon/distribution.c
@@ -319,7 +319,7 @@ double distribution_total_sum(distribution_t *dist) {
 
 uint64_t distribution_total_counter(distribution_t *dist) {
   if (dist == NULL) {
-    return NAN;
+    return EINVAL;
   }
   return dist->tree[0].bucket_counter; // should I add mutex here?
 }
@@ -331,7 +331,7 @@ double distribution_squared_deviation_sum(distribution_t *dist) {
   double mean = distribution_average(dist);
   pthread_mutex_lock(&dist->mutex);
   double squared_deviation_sum =
-      mean * mean * distribution_total_counter(dist) -
+      mean * mean * (double) distribution_total_counter(dist) -
       2 * mean * dist->total_sum + dist->total_square_sum;
   pthread_mutex_unlock(&dist->mutex);
   return squared_deviation_sum;
--- a/src/daemon/distribution.c
+++ b/src/daemon/distribution.c
@@ -317,7 +317,7 @@ double distribution_total_sum(distribution_t *dist) {
   return dist->total_sum; // should I add mutex here?
 }
 
-double distribution_total_counter(distribution_t *dist) {
+uint64_t distribution_total_counter(distribution_t *dist) {
   if (dist == NULL) {
     return NAN;
   }
--- a/src/daemon/distribution.c
+++ b/src/daemon/distribution.c
@@ -210,13 +210,10 @@ static void update_tree(distribution_t *dist, size_t node_index, size_t left,
     update_tree(dist, right_child, mid + 1, right, gauge);
 }
 
-void distribution_update(distribution_t *dist, double gauge) {
+int distribution_update(distribution_t *dist, double gauge) {
   if (dist == NULL)
-    return;
-  if (gauge < 0) {
-    errno = EINVAL;
-    return;
-  }
+    return EINVAL;
+  assert(gauge >= 0);
   pthread_mutex_lock(&dist->mutex);
   update_tree(dist, 0, 0, dist->num_buckets - 1, gauge);
   dist->total_sum += gauge;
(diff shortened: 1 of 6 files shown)


src/utils/common/common.c
```diff
@@ -1136,8 +1136,14 @@ metric_family_t *plugin_value_list_to_metric_family(value_list_t const *vl,
   fam->name = name.ptr;
   name = (strbuf_t){0};
 
-  fam->type =
-      (dsrc->type == DS_TYPE_GAUGE) ? METRIC_TYPE_GAUGE : METRIC_TYPE_COUNTER;
+  switch (dsrc->type) {
```
bkjg, line 1139: You missed the breaks in switch cases. Without them you will go to the next case and your family type in all cases will be METRIC_TYPE_UNTYPED

bkjg (commented):
Thank you! That looks great, Svetlana!

src/daemon/distribution.c
```diff
@@ -210,18 +211,16 @@ static void update_tree(distribution_t *dist, size_t node_index, size_t left,
...
+int distribution_update(distribution_t *dist, double gauge) {
+  if (dist == NULL || gauge < 0) {
+    return errno = EINVAL;
   }
   pthread_mutex_lock(&dist->mutex);
   update_tree(dist, 0, 0, dist->num_buckets - 1, gauge);
   dist->total_sum += gauge;
   dist->total_square_sum += gauge * gauge;
   pthread_mutex_unlock(&dist->mutex);
+  return 0;
```
elene-margalit, line 223: Why did you change the return type to int?

src/daemon/distribution.c
```diff
@@ -337,14 +344,42 @@ double distribution_squared_deviation_sum(distribution_t *dist) {
   return squared_deviation_sum;
 }
 
+double distribution_stddev(distribution_t *dist) {
```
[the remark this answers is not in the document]
elene-margalit: I think it calculates the standard deviation of the gauges.

elene-margalit (commented):
Great work Svetlana!

src/daemon/distribution.c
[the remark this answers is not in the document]
Lana243: Yes, Elene is right. Ths function calculates the standard deviation of gauges. It is also written in distribution.h file. Do you think it'll be better to change the name? Could you propose some other name?

src/daemon/distribution.c
Lana243: I change it in order to return error code if smth went wrong. That case errno can't be overwritten by smth else

src/daemon/distribution.c
```diff
@@ -337,14 +344,42 @@ double distribution_squared_deviation_sum(distribution_t *dist) {
...
+    dist->tree[i].bucket_counter = 0;
+  }
+  return 0;
+}
+
 /* TODO(bkjg): add tests for this function */
 int distribution_sub(distribution_t *d1, distribution_t *d2) {
   if (d1 == NULL || d2 == NULL) {
-    return EINVAL;
+    return errno = EINVAL;
```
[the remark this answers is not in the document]
Lana243: Hm, I was a bit confused with it because when we have functions that has return values then we set errno to smth. In other case when our functions are voids we just return the error code and don't set errno. I thought that it would be better to have one approach to all functions but I'm not sure. If you think that it's odd I'll remove it

src/daemon/distribution.c
```diff
@@ -337,14 +344,42 @@ double distribution_squared_deviation_sum(distribution_t *dist) {
...
+  }
+  dist->total_sum = 0;
+  dist->total_square_sum = 0;
+  for (size_t i = 0; i < tree_size(dist->num_buckets); i++) {
+    dist->tree[i].bucket_counter = 0;
+  }
+  return 0;
+}
+
```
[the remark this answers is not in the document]
Lana243: Sure! Thanks for your comment. I've fixed it

src/ping.c
```diff
@@ -82,6 +81,11 @@ static double ping_interval = 1.0;
 static double ping_timeout = 0.9;
 static int ping_max_missed = -1;
 
+/* Private variables for distribution for latency */
+static size_t num_buckets = 100;
+#define PING_DEF_DISTRIBUTION                                                  \
+  distribution_new_linear(num_buckets, ping_timeout / num_buckets)
```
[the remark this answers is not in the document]
Lana243: I think it's unnecessary as ping_timeout is double so the result of division will be casted to double

src/utils/common/common.c
Lana243: Thanks a lot, Barbara! I didn't know that switches in C work this way. I'll fix it

src/ping.c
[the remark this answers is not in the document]
bkjg: I think that some compilers can complain about it

bkjg (commented):
That looks great, Svetlana! 😃

src/ping.c
```diff
@@ -565,8 +574,38 @@ static int ping_config(const char *key, const char *value) /* {{{ */
   return 0;
 } /* }}} int ping_config */
 
-static void submit(const char *host, const char *type, /* {{{ */
-                   gauge_t value) {
+static void submit_distribution(const char *host, const char *type,
+                                distribution_t *dist) {
+  metric_family_t *fam = calloc(1, sizeof(*fam));
+  if (fam == NULL) {
+    free(fam);
```
bkjg, line 581: Why are you freeing metric_family_t in case then fam is NULL? I think this line is redundant

src/ping.c
```diff
@@ -565,8 +574,38 @@ static int ping_config(const char *key, const char *value) /* {{{ */
   return 0;
 } /* }}} int ping_config */
 
-static void submit(const char *host, const char *type, /* {{{ */
-                   gauge_t value) {
+static void submit_distribution(const char *host, const char *type,
```
bkjg, line 577: What does in case when host or type or dist are null?

src/ping.c
```diff
@@ -631,28 +661,13 @@ static int ping_read(void) /* {{{ */
...
     /* Calculate drop rate. */
     droprate = ((double)(pkg_sent - pkg_recv)) / ((double)pkg_sent);
 
-    submit(hl->host, "ping", latency_average);
-    submit(hl->host, "ping_stddev", latency_stddev);
-    submit(hl->host, "ping_droprate", droprate);
+    submit_gauge(hl->host, "ping_droprate", droprate);
+    submit_distribution(
+        hl->host, "ping_distribution_latency",
+        dist_latency); // dist_latency will be destroyed in submit_distribution
```
bkjg, line 670: Just to let you know:
I could argue if dist_latency is really destroyed in submit_distribution. As you can see, in submit_distribution you create a temporary metric with this distribution, nextly you append it to the metric family and at the end you free the metric_family. However I think there is a bug inside metric_reset function which metric_family_free calls and there is a place where a distribution_destroy call is missing. Also I think metric_family_metric_append should clone the distribution instead using the same.

I will fix these bugs in another pull request

bkjg (commented):
Thank you, Svetlana!

PR #3570, opened 2020-09-09
Utils cache: Start value

Lana243:
Add start_value and start_time to cache_entry_t. That will allow us to get start values not from meta data but from cache_entry_t itself. This way we don't need to store distributions in meta data which is a bit complicated as meta data stores scalar values.

ChangeLog: Utils cache: Add start_value and start_time


elene-margalit (commented):
Well done Svetlana!

src/daemon/utils_cache.c
```diff
@@ -42,7 +42,7 @@
 
 typedef struct cache_entry_s {
   char name[6 * DATA_MAX_NAME_LEN];
-  distribution_t *values_distribution;
+  distribution_t *distribution_increase;
```
elene-margalit, line 45: What is the meaning of this attribute?

src/daemon/distribution.h
```diff
@@ -118,6 +118,17 @@ double distribution_squared_deviation_sum(distribution_t *dist);
...
+
+/** This function subtracts d2 from d1 if arguments are correct
+ *  if arguments are NULL pointers or the structure of distributions is
+ * different then EINVAL is returned if distributions have the same structure
+ * but the second distribution is not less then the first then ERANGE is
+ * returned in case of success returns 0
+ *  This function holds both mutexes for d1 and d2. Be sure that you call the
+ * function with the same order of arguments everytime **/
 int distribution_sub(distribution_t *d1, distribution_t *d2);
 #endif // COLLECTD_DISTRIBUTION_H
```
elene-margalit, line 134: I think a new line at the end of file needs to be added here.

src/daemon/distribution.c
```diff
@@ -340,48 +341,91 @@ double distribution_squared_deviation_sum(distribution_t *dist) {
...
   }
-
   if (d1->num_buckets != d2->num_buckets) {
-    return EINVAL;
+    comparison[0] = EINVAL;
+    return comparison;
+  }
+  for (size_t i = 0; i < tree_size(d1->num_buckets); i++) {
+    if (d1->tree[i].maximum !=
+        d2->tree[i].maximum) { // will it work with doubles?
```
elene-margalit, line 367: That's a good question. After some googling I found out that a so called epsilon can be used to compare doubles:

"A better way to compare would be fabs(a-b) < epsilon where is epsilon is a small enough value, like 0.000001." (Taken from Reddit)

But I don't know for sure what the best approach is.

src/daemon/utils_cache.c
Lana243: distribution_increase shows us how the distribution has been updated since last time when we saw it. Using it will able to calculate rate. This change was just to rename values_distribution as it was unclear what exactly it means

src/daemon/distribution.c
Lana243: I agree with octo. I'll leave it like that and add a comment

src/daemon/distribution.c
```diff
@@ -340,48 +341,91 @@ double distribution_squared_deviation_sum(distribution_t *dist) {
...
+    if (res != cur_res) {
+      if (cur_res == 0)
+        continue;
+      if (res == 0) {
+        res = cur_res;
+        continue;
+      }
+      comparison[0] = ERANGE;
+      break;
+    }
```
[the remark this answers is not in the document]
Lana243: Yes, sure! Thanks a lot for your comment. It simplify code a lot!😊

src/daemon/distribution.c
```diff
@@ -340,48 +341,91 @@ double distribution_squared_deviation_sum(distribution_t *dist) {
...
+static int compare_longint(uint64_t a, uint64_t b) {
+  return a == b ? 0 : a < b ? -1 : 1;
+}
+
+/** return a pointer to array int[2].
+ * The first value is error code or 0 if succeed.
+ * The second value is result of comparison
+ */
+static int *distribution_cmp(distribution_t *d1, distribution_t *d2) {
+  int *comparison = calloc(2, sizeof(*comparison));
```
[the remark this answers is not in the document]
Lana243: Thanks for suggestion! I changed it using first option

src/daemon/utils_cache.c
```diff
@@ -350,17 +359,25 @@ static int uc_update_metric(metric_t const *m) {
   }
 
   case METRIC_TYPE_DISTRIBUTION: {
-    int status =
-        distribution_sub(ce->values_raw.distribution, m->value.distribution);
+    distribution_destroy(ce->distribution_increase);
+    ce->distribution_increase = distribution_clone(m->value.distribution);
+    status = distribution_sub(ce->distribution_increase,
```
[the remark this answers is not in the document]
Lana243: Thanks for the comment. Fixed it.

src/daemon/utils_cache.c
```diff
@@ -649,6 +667,47 @@ int uc_get_value(metric_t const *m, value_t *ret) {
...
+    if (ce->state == STATE_MISSING) {
+      status = -1;
+    } else {
+      *ret_start_value = typed_value_clone(ce->start_value).value;
+      *ret_start_time = ce->start_time;
+    }
+  } else {
+    DEBUG("utils_cache: uc_get_start_value_by_name: No such value: %s", name);
+    status = -1;
+  }
```
octo, line 690: Consider inverting the logic here: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

src/daemon/utils_cache.c
```diff
@@ -143,25 +148,29 @@ static int uc_insert(metric_t const *m, char const *key) {
   case DS_TYPE_COUNTER:
     ce->values_gauge = NAN;
     ce->values_raw.counter = m->value.counter;
-    ce->values_distribution = NULL;
+    ce->distribution_increase = NULL;
+    ce->start_value = create_typed_value(m->value, METRIC_TYPE_COUNTER);
```
octo, line 152: *create_typed_value()* doesn't seem to exist. Potentially this could be a composite literal like so:

```c
ce->start_value = (typed_value_t){m->value, METRIC_TYPE_COUNTER};
```

src/daemon/utils_cache.c
```diff
@@ -641,7 +644,7 @@ int uc_get_value_by_name(const char *name, value_t *ret_values) {
     if (ce->state == STATE_MISSING) {
       status = -1;
     } else {
-      *ret_values = ce->values_raw;
+      *ret_values = typed_value_clone(ce->values_raw).value;
```
octo, line 647: At your option, a `value_t value_clone(value_t v)` might be easier to read.

Lana243: Unfortunately, we can't do it like this as then we won't pass any information about type to clone function. What we can do is to create a function `value_t value_clone(typed_value_t)`. Do you think it'd be better?

src/daemon/utils_cache.c
```diff
@@ -147,30 +147,30 @@ static int uc_insert(metric_t const *m, char const *key) {
...
+    ce->values_raw = typed_value_create(m->value, METRIC_TYPE_GAUGE);
     ce->distribution_increase = NULL;
-    ce->start_value = create_typed_value(m->value, METRIC_TYPE_GAUGE);
+    ce->start_value = typed_value_create(m->value, METRIC_TYPE_GAUGE);
     break;
 
   case DS_TYPE_DERIVE:
     ce->values_gauge = NAN;
-    ce->values_raw.derive = m->value.derive;
+    // ce->values_raw.derive = m->value.derive;
```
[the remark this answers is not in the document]
Lana243: Thanks for your comment. I add derives with `METRIC_TYPE_COUNTER` type

src/daemon/utils_cache.c
Lana243: Create a `typed_value_create` function. Using such composite literal here won't work as if type is distribution then we need to clone it

octo (approved):
Thanks @Lana243!

PR #3571, opened 2020-09-09
write_influx_udp: build influxdb points outside de mutex.

carlospeon:
Hello:

This is a PR to move the build of the influxdb points outside de mutex. Extra memory and a copy operation are needed but this allow write threads to build points in paralell which is convinient in high load environments.

Regards,
Carlos.

ChangeLog: write_influx_udp: build influxdb points outside de mutex.


octo (approved):
More efficient and shorter / cleaner. Great work @carlospeon, thank you!

PR #3574, opened 2020-09-14
replace mysql_config by mariadb_config

mrunge:
to enable building the mysql plugin on debian.

ChangeLog: Build system: Update dependency for the mysql/mariadb plugin

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3527,7 +3527,8 @@ AC_SUBST([BUILD_WITH_LIBMOSQUITTO_LIBS])
 # }}}
 
 # --with-libmysql {{{
-with_mysql_config="mysql_config"
+# mariadb_config provides the same output like mysql_config
+with_mysql_config="mariadb_config"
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -3538,8 +3539,8 @@ AC_ARG_WITH([libmysql],
     else
       if test -f "$withval" && test -x "$withval"; then
         with_mysql_config="$withval"
-      else if test -x "$withval/bin/mysql_config"; then
-        with_mysql_config="$withval/bin/mysql_config"
```
octo, line 3542: I'd prefer if we checked for *mariadb_config* in addition to *mysql_config*. I think we should check for *mariadb_config* first, to prefer the newer system.

octo:
With collectd/ci-docker#8 MySQL once again builds on Debian Buster. I think it still makes sense to prepare for a future without MySQL compatibility packages.

PR #3578, opened 2020-09-15
Fix memory leaks in write_http plugin

bkjg:
Fix memory leaks in write_http plugin

ChangeLog: fix memory leaks in write_http plugin

### DIFF
--- a/src/write_http.c
+++ b/src/write_http.c
@@ -290,7 +290,10 @@ static int wh_flush(cdtime_t timeout,
     return ENOMEM;
   }
 
-  return wh_post(cb, json);
+  int status = wh_post(cb, json);
+  free((char*)json);
+
+  return status;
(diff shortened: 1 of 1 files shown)


src/write_http.c
```diff
@@ -290,7 +290,10 @@ static int wh_flush(cdtime_t timeout,
     return ENOMEM;
   }
 
-  return wh_post(cb, json);
+  int status = wh_post(cb, json);
+  free((char *)json);
```
octo, line 294: I'd prefer to drop the `const` keyword from the `json` variable ­– some compilers are bound to complain about disregarding the const keyword here. When done, you can remove the cast here.

octo (commented):
Thanks for finding and fixing these Barbara :-D

octo (approved):

PR #3579, opened 2020-09-15
Fix bug inside write_http plugin

bkjg:
Fix the probable bug inside write_http plugin. For loop by i variable every time takes the same metric from the metric family. Due to the iteration by i, it looks like we should take ith element every time.

ChangeLog: write_http plugin: A bug when iterating over metrics in a metric family has been fixed.


octo (approved):
Thanks Barbara!

PR #3581, opened 2020-09-16
cirrus.yml: Add address_sanitizer_task.

octo:


octo:
This fails, because `make check` involves Valgrind, and Valgirnd and AddressSanitizer step on each other's toes.

PR #3582, opened 2020-09-16
Utils cache destroy

Lana243:
ChangeLog: Utils Cache: Fix memory leaks by adding new uc_destroy function


octo (approved):
Great work Svetlana, thank you!

PR #3760, opened 2020-09-25
configfile.c: Fix memory leaks in case of failure

bkjg:
ChangeLog: configfile.c: Fix memory leaks in case of failure


octo (approved):
Great catch, thanks Barbara!

octo:
Cherry-picked into *collectd-5.12* as a3f5e2de54d5f30197dda612113eec50cd9bbae0.

PR #3763, opened 2020-09-29
[collectd 6] write_influxdb_udp: migration to v6.0

carlospeon:
Initial version for v6.0. 

Issue #3746

Changelog: write_influxdb_udp.c: migration to v6.0

### DIFF
--- a/src/write_influxdb_udp.c
+++ b/src/write_influxdb_udp.c
@@ -326,22 +326,32 @@ static void flush_buffer(void) {
   write_influxdb_udp_init_buffer();
 }
 
+static void fill_send_buffer(char const *buffer, size_t len) {
+  assert(len <= wifxudp_config_packet_size);
+
+  pthread_mutex_lock(&send_buffer_lock);
+  if (wifxudp_config_packet_size - send_buffer_fill < len)
+    flush_buffer();
+  memcpy(send_buffer_ptr, buffer, len);
+  send_buffer_fill += len;
+  send_buffer_ptr += len;
+  send_buffer_last_update = cdtime();
+  pthread_mutex_unlock(&send_buffer_lock);
+}
+
 static int wifxudp_escape_string(char *buffer, size_t buffer_size,
                                  const char *string) {
 
   if ((buffer == NULL) || (string == NULL))
     return -EINVAL;
 
-  if (buffer_size < 3)
-    return -ENOMEM;
-
   int dst_pos = 0;
 
 #define BUFFER_ADD(c)                                                          \
   do {                                                                         \
     if (dst_pos >= (buffer_size - 1)) {                                        \
-      buffer[buffer_size - 1] = '\0';                                          \
-      return -ENOMEM;                                                          \
+      buffer[buffer_size - 1] = 0;                                             \
+      return dst_pos++;                                                        \
     }                                                                          \
     buffer[dst_pos] = (c);                                                     \
     dst_pos++;                                                                 \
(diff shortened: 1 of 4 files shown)


src/write_influxdb_udp.c
```diff
@@ -365,136 +365,131 @@ static int wifxudp_escape_string(char *buffer, size_t buffer_size,
...
-    if (ds->ds[i].type == DS_TYPE_GAUGE) {
-      if (isnan(vl->values[i].gauge))
-        continue;
-      if (have_values)
-        BUFFER_ADD(",");
-      BUFFER_ADD("%s=%lf", ds->ds[i].name, vl->values[i].gauge);
-      have_values = true;
+    if (fam->type == METRIC_TYPE_GAUGE) {
+      if (!isnan(metric.value.gauge)) {
+        BUFFER_ADD("value=%lf", metric.value.gauge);
```
octo, line 411: Sorry for picking on code you haven't introduced: `%lf` is uncommon, because `gauge_t` is essentially a double. I'd prefer to use `GAUGE_FORMAT` here, or `%f` if the format produced by `GAUGE_FORMAT` can't be parsed by InfluxDB.

src/write_influxdb_udp.c
```diff
@@ -365,136 +365,131 @@ static int wifxudp_escape_string(char *buffer, size_t buffer_size,
...
+        BUFFER_ADD("value=%lf", rate);
+        have_values = true;
+      }
+    } else {
+      switch (fam->type) {
+      case METRIC_TYPE_COUNTER:
+        BUFFER_ADD("value=%" PRIi64 "i", metric.value.counter);
+        have_values = true;
+        break;
+      case METRIC_TYPE_UNTYPED: /* derive ¿? */
```
octo, line 432: No, basically a gauge. I'm debating with myself whether we should remove `METRIC_TYPE_UNTYPED`, it seems like a Prometheus legacy that we don't need to introduce.

octo (commented):
Hi @carlospeon,

thank you so much for your PR! I've left some comments inline.

We're always looking for contributors to help keep collectd going. Would you consider (co)owning this plugin?

Thanks and best regards,
--octo

src/write_influxdb_udp.c
```diff
@@ -365,136 +365,131 @@ static int wifxudp_escape_string(char *buffer, size_t buffer_size,
...
-  BUFFER_ADD(" ");
-  for (size_t i = 0; i < ds->ds_num; i++) {
-    if ((ds->ds[i].type != DS_TYPE_COUNTER) &&
-        (ds->ds[i].type != DS_TYPE_GAUGE) &&
-        (ds->ds[i].type != DS_TYPE_DERIVE)) {
-      sfree(rates);
-      return -1;
+    for (size_t j = 0; j < metric.label.num; j++) {
+      label_pair_t label = metric.label.ptr[j];
+      if (strcmp(label.name, "") != 0 && strcmp(label.value, "") != 0) {
```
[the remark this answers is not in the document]
dago: Theoretically yes, but what I have found while inspecting the kstat code is that metrics with an empty name _can_ be reported. If all reader modules consistently skip empty metrics your statement is true, but is this the case?

carlospeon:
> Hi @carlospeon,
> 
> thank you so much for your PR! I've left some comments inline.
> 
> We're always looking for contributors to help keep collectd going. Would you consider (co)owning this plugin?
> 
> Thanks and best regards,
> --octo

Hello! Yes, its OK for me.

Best regards.
Carlos.

src/write_influxdb_udp.c
```diff
@@ -365,136 +365,131 @@ static int wifxudp_escape_string(char *buffer, size_t buffer_size,
...
+  if (fam == NULL)
+    return EINVAL;
+
   char buffer[NET_DEFAULT_PACKET_SIZE];
+  size_t idx = 0;
 
-  int status = write_influxdb_point(buffer, NET_DEFAULT_PACKET_SIZE, ds, vl);
+  while (idx < fam->metric.num) {
+    int status =
+        write_influxdb_point(buffer, NET_DEFAULT_PACKET_SIZE, fam, &idx);
```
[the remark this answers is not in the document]
carlospeon: Absolutly. It needs some deep changes.

src/write_influxdb_udp.c
```diff
@@ -365,136 +365,131 @@ static int wifxudp_escape_string(char *buffer, size_t buffer_size,
...
-    BUFFER_ADD_ESCAPE(vl->type);
-  }
-  if (strcmp(vl->type_instance, "") != 0) {
-    BUFFER_ADD(",type_instance=");
-    BUFFER_ADD_ESCAPE(vl->type_instance);
-  }
+  for (; *idx < fam->metric.num; (*idx)++) {
+    int offset = metric_offset;
+    have_values = false;
+    BUFFER_ADD_ESCAPE(fam->name);
```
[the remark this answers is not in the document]
carlospeon: OK, let me take a look.

src/write_influxdb_udp.c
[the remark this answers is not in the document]
carlospeon: This comes from bad exec plugins and other software sending metrics directly to collectd with empty plugin_instance, type, etc. If it does not apply in v6 it can be removed ofc.

src/write_influxdb_udp.c
carlospeon: Ok, changed to `GAUGE_FORMAT`

src/write_influxdb_udp.c
carlospeon: OK, expecting no gauges I guess.

src/write_influxdb_udp.c
```diff
@@ -365,136 +365,131 @@ static int wifxudp_escape_string(char *buffer, size_t buffer_size,
...
+    BUFFER_ADD(" ");
 
-    if (ds->ds[i].type == DS_TYPE_GAUGE) {
-      if (isnan(vl->values[i].gauge))
-        continue;
-      if (have_values)
-        BUFFER_ADD(",");
-      BUFFER_ADD("%s=%lf", ds->ds[i].name, vl->values[i].gauge);
-      have_values = true;
+    if (fam->type == METRIC_TYPE_GAUGE) {
```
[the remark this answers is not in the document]
carlospeon: OK, updated.

src/write_influxdb_udp.c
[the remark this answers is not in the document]
octo: That label names are not empty is ensured in `label_set_create`. An empty label value is treated as "delete this label" in `metric_label_set`. There are other ways to allocate the structure, circumventing these checks, but it should be quite rare.

On the other hand, these checks here don't really hurt either, so if you prefer we can keep them in as well.

src/write_influxdb_udp.c
octo: Correct, `METRIC_TYPE_COUNTER` is taking the role of derives.

src/write_influxdb_udp.c
```diff
@@ -326,22 +326,30 @@ static void flush_buffer(void) {
   write_influxdb_udp_init_buffer();
 }
 
+static void fill_send_buffer(char const *buffer, size_t len) {
```
octo, line 329: I think that `len <= wifxudp_config_packet_size` is ensured by the calling code, but it's not apparent when reading this function. I'd recommend adding an assertion to document this fact.

src/write_influxdb_udp.c
```diff
@@ -364,137 +372,120 @@ static int wifxudp_escape_string(char *buffer, size_t buffer_size,
...
+    if (!isnan(metric.value.gauge)) {
+      BUFFER_ADD("value=" GAUGE_FORMAT, metric.value.gauge);
       have_values = true;
-    } else if (wifxudp_config_store_rates) {
-      if (rates == NULL)
-        rates = uc_get_rate_vl(ds, vl);
-      if (rates == NULL) {
+    }
+    break;
+  default:
```
octo, line 421: I recommend to use `case METRIC_TYPE_COUNTER` specifically. We're doing to merge the distribution metrics soon, and that would make this switch-case easier to discover, and the code can fail gracefully if it gets called with a distribution metric that it does not yet support.

src/write_influxdb_udp.c
```diff
@@ -364,137 +372,120 @@ static int wifxudp_escape_string(char *buffer, size_t buffer_size,
...
-      return -1;
-    }
 
-    if (ds->ds[i].type == DS_TYPE_GAUGE) {
-      if (isnan(vl->values[i].gauge))
-        continue;
-      if (have_values)
-        BUFFER_ADD(",");
-      BUFFER_ADD("%s=%lf", ds->ds[i].name, vl->values[i].gauge);
+  switch (metric.family->type) {
```
octo, line 413: This is relatively complex:

```c
switch (metric.family->type) {
  // …
  default:
    if (wifxudp_config_store_rates) {
      // …
    } else {
      switch (metric.family->type) {
        case METRIC_TYPE_COUNTER: // …
      }
    }
}
```

I think this could be simplified to:

```c
if (wifxudp_config_store_rates ||
    (metric.family->type == METRIC_TYPE_GAUGE) ||
    (metric.family->type == METRIC_TYPE_UNTYPED)) {
  // …
  // uc_get_rate handles gauge gracefully
  BUFFER_ADD();
  return offset;
}
switch (metric.family->type) {
  case METRIC_TYPE_COUNTER:
    // …
}
```

src/write_influxdb_udp.c
```diff
@@ -364,137 +372,120 @@ static int wifxudp_escape_string(char *buffer, size_t buffer_size,
...
-      if (rates == NULL)
-        rates = uc_get_rate_vl(ds, vl);
-      if (rates == NULL) {
+    }
+    break;
+  default:
+    if (wifxudp_config_store_rates) {
+      gauge_t rate;
+      int ret = uc_get_rate(&metric, &rate);
+      if (ret != 0) {
```
octo, line 425: `ret` is basically unused. This could be simplified to:

```c
if (uc_get_rate(&metric, &rate) != 0) {
```

octo (commented):
Thanks for your update, @carlospeon!

src/write_influxdb_udp.c
carlospeon: I agree.

src/write_influxdb_udp.c
carlospeon: Ok, done!

src/write_influxdb_udp.c
carlospeon: Ok.

src/write_influxdb_udp.c
carlospeon: Ok, done!

src/write_influxdb_udp.c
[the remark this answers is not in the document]
dago: In the optimal case the place where empty values are removed should be defined explicitly, like "no read plugin may return empty values" and if they still do it is a bug. Introducing late checks on individual writer plugins result in behaviour deviances depending on the writer plugin which should in my opinion be avoided.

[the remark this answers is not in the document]
carlospeon: Removed

src/write_influxdb_udp.c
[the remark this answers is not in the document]
carlospeon: > At your option: it'd be great if we could replace the `BUFFER_ADD...` macros with the functions provided in `"utils/strbuf/strbuf.h"` I think `wifxudp_escape_string` could be replaced by `strbuf_print_escaped`.

You are probably right but I'm a bit confused how a full buffer is managed and I would prefer to stay for now.

````
/* Ensure the escape sequence is not truncated. */
    if (buf->fixed && (strbuf_avail(buf) < 2)) {
      return 0;
    }
````

carlospeon:
Hello @mrunge! Could you please review this PR?

Regards.

mrunge (approved):
thank you, looks good to me.

PR #3768, opened 2020-10-13
Avoid deadlocks.

pj1031999:
ChangeLog: Distribution data structure: Avoid deadlocks.

- Introduce `distribution_lock` for locking two distribution_t objects.
- Introduce `distribution_unlock` for unlocking two distribution_t objects.
- Fix the lock order in `distribution_equal`.
- Fix the lock order in `distribution_sub`.

### DIFF
--- a/src/daemon/distribution.c
+++ b/src/daemon/distribution.c
@@ -38,6 +38,61 @@ struct distribution_s {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


bkjg (approved):
Thank you for the fix @pj1031999 , LGTM!

PR #3770, opened 2020-10-26
Collectd 6.0 zone plugin

dago:
Porting the zone plugin to 6.0 metric format


dago:
The output is not well tested, please do not merge yet.

PR #3773, opened 2020-11-12
Compare both PID and process start time to avoid issues with immediate PID reuse on Linux

drNowak:
When PID is reused on Linux between two cycles, the data from the old process will not be cleared and new process will reuse the previous data. This is causing issues like negative CPU.

We have seen those issues at scale in Google Cloud, and this patch fixed the issue completely. This fix is tested at scale for 6 months already.

ChangeLog: processes.c: fix negative CPU reporting caused by PID reuse on Linux.

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -177,6 +177,9 @@
 typedef struct process_entry_s {
   unsigned long id;
   char name[PROCSTAT_NAME_LEN];
+  // The time the process started after system boot.
+  // Value is in jiffies.
+  unsigned long long starttime;
 
   unsigned long num_proc;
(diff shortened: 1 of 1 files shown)


drNowak:
I have run contrib/format.sh src/processes.c and added commit.

octo (approved):
Thanks Krzysztof!

vincele (commented):
LGTM

drNowak:
Hey guys, I cannot merge myself, please merge.

octo:
Rebased on `main`, now waiting for CI results.

PR #3776, opened 2020-11-26
lua plugin: Use a global lock to call Lua's API

kenhys:
ChangeLog:  Since vanilla Lua isn't multi-thread-safe, a global lock should be used
when calling their functions. The previous code causes unstable behavior
especially on loading Lua's C modules with both read & write threads.

```
e.g.)

  collectd.register_read(function()
    require('cqueues')
    return 0
  end)

  collectd.register_write(function()
    return 0
  end)
```

Note that you need to add the following config to your collectd.conf to
load C modules:

```
  <LoadPlugin lua>
    Globals true
  </LoadPlugin>
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,7 +9,7 @@ env:
 #
 release_ready_task:
   container:
-    image: collectd/ci:stretch_amd64
+    image: collectd/ci:buster_amd64
   lint_script:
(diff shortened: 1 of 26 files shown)


kenhys:
I've created against wrong branch, so I' ll close it.

PR #3778, opened 2020-11-27
lua: support to detect LuaJIT automatically

kenhys:
ChangeLog: lua: add fallback to detect LuaJIT

Before:
  LIBLUA_PKG_CONFIG_NAME=luajit ./configure --enable-lua

After:
  ./configure --enable-lua

In the previous versions, for enabling LuaJIT,
it needs to specify LIBLUA_PKG_CONFIG_NAME=luajit explicitly.

In this commit, add fallback to detect LuaJIT without specifying
LIBLUA_PKG_CONFIG_NAME=luajit explicitly by default (just use
--enable-lua).

Note that if both of lua and luajit exists, then lua is used.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3221,7 +3221,14 @@ else
                                     [
                                       PKG_CHECK_MODULES([LUA], [lua51],
                                         [with_liblua="yes"],
-                                        [with_liblua="no (pkg-config cannot find liblua)"]
+                                        [
+                                          PKG_CHECK_MODULES([LUA], [luajit],
+                                            [
+                                              with_liblua="yes"
+                                            ],
+                                            [with_liblua="no (pkg-config cannot find libluajit)"]
+					  )
+					]
                                       )
                                     ]
                                   )


kenhys:
Verified with:

```
sudo apt install -y libluajit-5.1-dev
git clone https://github.com/collectd/collectd.git
cd collectd
./build.sh
./configure --prefix=/tmp/local --disable-all-plugins --enable-lua --enable-cpu
make install
```

Use conf and sample lua:
```
$ cat /legacy/collectd/test-minimum.conf
PluginDir   "/tmp/local/lib/collectd"

LoadPlugin cpu

<LoadPlugin lua>
  Globals true
</LoadPlugin>

<Plugin lua>
  BasePath "/legacy/collectd/"
  Script "test-minimum.lua"
</Plugin>
```

```
$ cat /legacy/collectd/test-minimum.lua
-- without PR#3779

local PLUGIN_NAME = "Lua minimum plugin"

function read()
   collectd.log_info(PLUGIN_NAME .. ": read")
   return 0
end

function write(metrics)
   collectd.log_info(PLUGIN_NAME .. ": write")
   for key,value in pairs(metrics) do
      print(key,value)
   end
   return 0
end

collectd.register_read(read)
collectd.register_write(write)
```

Then execute `/tmp/local/sbin/collectd -f -C test-minimum.conf`

kenhys:
@octo could you afford to eview this PR?

octo (approved):
Thanks @kenhys!

PR #3781, opened 2020-12-02
gcc compat: Use alloca explicitly instead of VLA

weltling:
ChangeLog: gcc compat: Use alloca explicitly instead of VLA

When -Wall is used, compilation can fail with errors like below:

```
src/libcollectdclient/network_buffer.c:273:11: error: unbounded use of
variable-length array [-Werror=vla-larger-than=]
  273 |   uint8_t pkg_values_types[vl->values_len];
      |           ^~~~~~~~~~~~~~~~
src/libcollectdclient/network_buffer.c:274:11: error: argument to
variable-length array may be too large [-Werror=vla-larger-than=]
  274 |   value_t pkg_values[vl->values_len];
      |           ^~~~~~~~~~
src/libcollectdclient/network_buffer.c: In function
‘lcc_network_buffer_add_value’:
src/libcollectdclient/network_buffer.c:274:11: note: limit is 16 bytes,
but argument may be as large as 18446744073709551615
```

This is fixed by explicitly using `alloca` which is equivalent to the compiler magic behind the VLA. Alternatively, these case can be moved to the heap, whereby in most cases it seems to be not necessary. Alternatively, consumer need to enforce `-Wno-vla-larger-than` or it needs to be added to the defaults in the makefile.

Signed-off-by: Anatol Belski <anbelski@linux.microsoft.com>


mrunge (dismissed):

mrunge:
Hi, and thank you for your patch!

Unfortunately, our CI is currently blocked, and I don't have any powers to unblock it.
Basically, two things need to happen: 
1. drop centos 6 (patch already up)
2. replacement of travis e.g with github actions, see https://github.com/collectd/collectd/pull/3787

weltling:
Thanks for checking the patch! Sure the QA needs to pass prior a merge. I'm around anytime once CI is fixed, to refresh it for the next round.

Thanks!

mrunge (approved):
Thank you, looks good to me

PR #3785, opened 2020-12-05
github actions: use GitHub Actions for checking build

kenhys:
ChangeLog: Because of new pricing model,
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
It may be better to use another CI/CD such as GitHub Actions.

PR #3784, opened 2020-12-05
cirrus: drop CentOS6 (EOL)

kenhys:
ChangeLog: It had been reached EOL at November 30th, 2020.
There is no security updates available anymore.

ref. https://wiki.centos.org/About/Product

PR #3787, opened 2020-12-08
[DNM] Drop travis, add github actions

mrunge:
This is based on the patches https://github.com/collectd/collectd/pull/3785 and https://github.com/collectd/collectd/pull/3784 and also drops travis.yml.

This is WIP, since there is no integration with solaris or mac builds (yet)

ChangeLog: Drop Travis, use Github actions


.github/workflows/linux.yml
```diff
@@ -0,0 +1,96 @@
+name: Linux
+on:
+  - push
+  - pull_request
+jobs:
+  buildtest::
```
leifmadsen, line 6: Extra colon

leifmadsen (approved):

.github/workflows/linux.yml
mrunge: Thank you, good catch

PR #3790, opened 2020-12-08
dhcpleases: add dhcpleases plugin

PolynomialDivision:
Changelog: dhcpleases: add plugin for counting current dhcp leases

The plugin is useful for the Freifunk Community. Currently, we use the exec-plugin. With that dhcpleases plugin we have native collectd support to measure this important statistic.


PolynomialDivision:
I'm very new to this project. Could u help me to understand why some tests are failing? I suspect that `configure.ac` is wrong, or?

mrunge:
Hi, 

thank you for your PR!

CI is currently a bit (?) blocked by travis. Also CentOS 6 reached it's end-of-life and should be removed.

PolynomialDivision:
> Hi,
> 
> thank you for your PR!
> 
> CI is currently a bit (?) blocked by travis. Also CentOS 6 reached it's end-of-life and should be removed.

Hey. :)
I'm just chaning the configure.ac right now to see if the buildchain then succeeds. I now used the olsrd plugin as template.

PolynomialDivision:
I have no idea why https://github.com/collectd/collectd/pull/3790/checks?check_run_id=1517390398 this fails. :/

mrunge:
> I have no idea why https://github.com/collectd/collectd/pull/3790/checks?check_run_id=1517390398 this fails. :/

I do, there is a PR up for that to disable it. CentOS 6 should go away, this has nothing to do with your patch.

mrunge:
This is the PR solving the CentOS 6 issue: https://github.com/collectd/collectd/pull/3784

PolynomialDivision:
Okay. Thanks for your help! :D I think I was able to fix my other snmp6 PR, too. :)
I will test everything again.

PolynomialDivision:
I made the dhpleasefile now configurable. :)

BKPepe:
Would it great if you can do `git format-patch` from your pull requests, which you sent to upstream and include it here as there is missing header, commit description and so on.

pmelange:
@PolynomialDivision I think this module looks promising and I look forward to using it.  One feature which may be "nice to have" is something like an Offset option.  

By this, I mean the following:

If I have a network which has a /24 DHCP range.  I also know that X of the devices connected to the network are access points and Y of the devices are managed switches.  In this case, I would set "Offset" to X+Y expecting then the number of DHCP leases reported to be the actual number of users.  This would be a far more informational statistic to me.

Another option, which also would address this idea, is to have the ability to ignore specific MAC or hostnames when counting.

manuelluis:
How about making a generic plugin to count lines in a file that match specific patterns?

Something like this:

```
<Plugin "linecount">
  <File "/var/test/test.txt">
    Instance "test"
    Include "regex"
    Exclude "regex"
  </File>
</Plugin>
```

vincele:
I'd use a dhcpleases plugin if it is merged.

Issue #3216, opened 2019-07-13
collectd -t should test config


rpv-tomsk:
*   Version of collectd: 5.8.0

## Expected behavior

Collectd should test config and report if it is broken/has errors.

## Actual behavior

Daemon exits without config test.

## Steps to reproduce

run `collectd -t`
It exits immediately w/o any output.

Whissi:
I just wanted to report a similar issue: collectd 5.9.1 isn't starting here with errors like
```
Plugin syslog failed to handle option LogLevel, return code: 1
```
or
```
Plugin iptables failed to handle option Chain, return code: 1
```
However, configtest reports no problem.

faxm0dem:
@whissi due to #3232 too

Whissi:
I am not sure if this is the same problem. Sure: Once the mentioned bug is fixed, the reported plugins are working. However, isn't the problem that the plugins will report an error for whatever reason and the configtest doesn't notice?

I.e. see https://github.com/collectd/collectd/issues/3236#issuecomment-516355640 ... even with the patch applied, `collectd -t` didn't report an error. After recompiling collectd with `--enable-debug` which seems like making loglevel `debug` available, it worked.

faxm0dem:
you're absolutely right, thanks for clarifying this

kkepka:
Using bisect and totally not valid config
```
$ cat /opt/collectd/etc/collectd.conf
badconfig
```
found that this behavior has changed probably with 386ecab
```
$ /opt/collectd/sbin/collectd -t
$ echo $?
0
```
one earlier 6b64979 have build issue on Ubuntu16.04 and next one 6cde759 was working as following:
```
$ /opt/collectd/sbin/collectd -t
Parse error in file `/opt/collectd/etc/collectd.conf', line 2 near `<newline>': syntax error, unexpected EOL
yyparse returned error #1
configfile: Cannot read file `/opt/collectd/etc/collectd.conf'.
Unable to read config file /opt/collectd/etc/collectd.conf.
Error: Reading the config file failed!
Read the logs for details.
$ echo $?
1
```
I'm not fully familiar with daemon yet, so just sharing this find here for now and hope to get back later to investigate logic that was affected. Starting point most probably will be [src/daemon/collectd.c#L402-L403](https://github.com/collectd/collectd/blob/master/src/daemon/collectd.c#L402-L403) and verifying if code responsible for testing config is being executed somewhere before.

isality:
[...]

PR #3793, opened 2020-12-17
Fix broken "-t" option of collectd

ashie:
Changelog: Fix broken "-t" option of collectd command

"-t" option should test config file but it always returns 0 with no
message even if a broken config file is specified.

Signed-off-by: Takuro Ashie <ashie@clear-code.com>

### DIFF
--- a/src/daemon/collectd.c
+++ b/src/daemon/collectd.c
@@ -395,9 +395,6 @@ struct cmdline_config init_config(int argc, char **argv) {
 
   read_cmdline(argc, argv, &config);
 
-  if (config.test_config)
-    exit(EXIT_SUCCESS);
-
   if (optind < argc)
     exit_usage(EXIT_FAILURE);
 
@@ -405,6 +402,9 @@ struct cmdline_config init_config(int argc, char **argv) {
 
   if (configure_collectd(&config) != 0)
     exit(EXIT_FAILURE);
+
+  if (config.test_config)
+    exit(EXIT_SUCCESS);
 
   return config;
 }


ashie:
Example broken collectd.conf:
```
foobar
```

before:
```console
$ ./collectd -t -C collectd.conf; echo $?
0
```

after:
```console
$ ./collectd -t -C config.txt; echo $?
configfile: stat (config.txt) failed: No such file or directory
Configuration file config.txt is empty.
Error: Parsing the config file failed!
1
```

faxm0dem:
Thanks ! I'm really surprised I never noted this before.
Just FTR I'm investigating which commit introduced this bug.

ashie:
At least collectd-4.0 doesn't seem broken.

mrunge:
Thank you.
This fixes https://github.com/collectd/collectd/issues/3216

mrunge:
> Thanks ! I'm really surprised I never noted this before.
> Just FTR I'm investigating which commit introduced this bug.

Hi Fabien,

you forgot that you worked on this, I see your comments on https://github.com/collectd/collectd/issues/3216 .... ;-)

faxm0dem:
Indeed, thanks for reminding me :-)

faxm0dem:
I just tested this and it seems to solve the issue: return 1 status when parser is failing.
I also tested a valid config, which still works.

mrunge (approved):

eero-t:
Can somebody else  than PR author re-trigger CI run (now that CI is fixed)?

systemcrash:
Just ran into this recently - good catch!

PR #3795, opened 2020-12-22
[collectd 6] lua: migrate the Lua plugin to v6.0

kenhys:
ChangeLog: lua: migrate the Lua plugin to v6.0

Since v6, value_list_t is deprecated and metric_family_t is introduced.

metric_family_t is mapped to the following Lua table:

```
  {
    name => ...,
    help => ...,
    unit => ...,
    type => ...,
    resource => {
      key1 => value1,
      ...,
      keyN => valueN
    }
    metric => {
     [1] => {
       label => {
         key1 => value1,
         ...,
         keyN => valueN
       }
       value => ...,
       time => ...,
       interval => ...,
       meta => {
         key1 => value1,
         ...,
         keyN => valueN
       }
     },
     ...
     [N] => {
       ...
     }
  }
```

Closes: #3654


kenhys:
Rebased with recent collectd-6.0 branch:

Verified with:

```
sudo apt install -y libluajit-5.1-dev
git clone https://github.com/kenhys/collectd.git
cd collectd
git checkout -b v6-lua origin/v6-lua
export LIBLUA_PKG_CONFIG_NAME=luajit
./build.sh
./configure --prefix=/tmp/local --disable-all-plugins --enable-lua --enable-cpu
make install
```

Use conf and sample lua:

```
$ cat /legacy/collectd/test-v6.conf
PluginDir   "/tmp/local/lib/collectd"

LoadPlugin cpu

<LoadPlugin lua>
  Globals true
</LoadPlugin>

<Plugin lua>
  BasePath "/legacy/collectd/"
  Script "test-v6.lua"
</Plugin>
```

Use test-v6.lua something like:
(put inspect.lua under same directory)

```
$ cat /legacy/collectd/test-v6.lua
local inspect = require("inspect")
local PLUGIN_NAME = "Lua minimum plugin"

function read()
   collectd.log_info(PLUGIN_NAME .. ": read")
   return 0
end

# verified with only "LoadPlugins cpu"

function write(metrics)
   collectd.log_info(PLUGIN_NAME .. ": write")
   -- print(inspect(metrics))
   
   assert(metrics["name"] == "system.cpu.time", "name must be system.cpu.time")
   assert(#metrics["resource"] == 3, "resource must contains 3 elements")
   assert(metrics["resource"][1]["host.id"], "resource must contain host.id")
   assert(metrics["resource"][2]["host.name"], "resource must contain host.name")
   assert(metrics["resource"][3]["service.name"], "resource must contain service.name")
   assert(metrics["type"] == 0, "type must be 0")
   assert(metrics["unit"] == "s", "unit must be 's'")
   assert(#metrics["metric"] > 0, "metric must be contained")
   for index, metric in pairs(metrics["metric"]) do
     assert(metric["interval"] == 10, "interval must be 10")
     assert(metric["time"] > 0, "time must be greater than 0")
     assert(metric["value"] >= 0, "value must be greater than 0")
     assert(#metric["label"] == 2, "label must contain cpu and state")
     assert(metric["label"][1]["cpu"], "label must contain cpu")
     assert(metric["label"][2]["state"], "label must contain state")
   end
   return 0
end

collectd.register_read(read)
collectd.register_write(write)
```

/tmp/local/sbin/collectd -f -C test-v6.conf should work without error.

kenhys:
@octo could you afford to review this PR?

src/utils_lua.c
```diff
@@ -274,38 +235,200 @@ int luaC_pushvalue(lua_State *L, value_t v, int ds_type) /* {{{ */
...
+    label_set_t will be mapped to the following Lua table:
+
+    label = {
+      [1] => { name => value },
+      ...
+      [N] => { ... }
+    }
+   */
   lua_newtable(L);
+  if (label.num > 0) {
```
octo, line 250: You can remove this `if` – if `label.num` is zero the loop will simply be skipped.

src/utils_lua.c
```diff
@@ -274,38 +235,200 @@ int luaC_pushvalue(lua_State *L, value_t v, int ds_type) /* {{{ */
...
+int luaC_pushlabelset(lua_State *L, label_set_t label) /* {{{ */
 {
+  /*
+    label_set_t will be mapped to the following Lua table:
+
+    label = {
+      [1] => { name => value },
+      ...
+      [N] => { ... }
+    }
```
octo, line 247: Why did you chose this approach? I'd find the following more intuitive:

```
    label = {
      name0 => value0,
      name1 => value1,
      …
    }
```

In other words, the order is not important and not guaranteed to remain by collectd.

If we were to add non-string label values in the future, would that influence the decision here?

src/utils_lua.c
```diff
@@ -274,38 +235,200 @@ int luaC_pushvalue(lua_State *L, value_t v, int ds_type) /* {{{ */
...
+      continue;
+    }
+    lua_newtable(L);
+    lua_pushinteger(L, meta_type);
+    lua_setfield(L, -2, "type");
+    char *string;
+    int64_t i64value;
+    uint64_t u64value;
+    double dvalue;
+    bool bvalue;
```
octo, line 303: Please declare these inside the switch cases. Also initialize them to zero.

src/utils_lua.c
```diff
@@ -274,38 +235,200 @@ int luaC_pushvalue(lua_State *L, value_t v, int ds_type) /* {{{ */
...
-  lua_pushstring(L, vl->host);
-  lua_setfield(L, -2, "host");
+int luaC_pushmetadata(lua_State *L, meta_data_t *meta) /* {{{ */
+{
+  /*
+    meta_data_t will be mapped to the following Lua table:
+
+    meta => {
+      [1] => {
+        type => ...,
```
octo, line 271: I'm not very familiar with Lua. From a quick look, it looks like Lua can determine the type at runtime. We we really need to export this field? If not, we could export the simpler `{ name0 => value0, name1 => value1, … }` table mentioned above.

src/utils_lua.c
```diff
@@ -274,38 +235,200 @@ int luaC_pushvalue(lua_State *L, value_t v, int ds_type) /* {{{ */
...
+      [1] => {
+        label => ...,
+        value => ...,
+        time => ...,
+        interval => ...,
+        },
+      ...
+      [N] => {
+       ...
+      }
```
octo, line 355: Thank you, this is super helpful!

src/utils_lua.c
```diff
@@ -274,38 +235,200 @@ int luaC_pushvalue(lua_State *L, value_t v, int ds_type) /* {{{ */
...
+      name => ...,
+      help => ...,
+      unit => ...,
+      type => ...,
+      resource => {
+        [1] => { name => value }
+        ...,
+        [N] => { ... }
+      },
+      metric => {
```
octo, line 393: Maybe abbreviate this to `metric => luaC_pushmetriclist(),` to keep the comment succinct.

src/utils_lua.c
```diff
@@ -274,38 +235,200 @@ int luaC_pushvalue(lua_State *L, value_t v, int ds_type) /* {{{ */
   return 0;
 } /* }}} int luaC_pushvalue */
 
-int luaC_pushvaluelist(lua_State *L, const data_set_t *ds,
-                       const value_list_t *vl) /* {{{ */
+int luaC_pushlabelset(lua_State *L, label_set_t label) /* {{{ */
```
octo, line 238: Not in the header file ⇒ make `static`

Dito for all other new functions except `luaC_pushmetricfamily`.

octo (commented):
Overall this looks great, thanks @kenhys! 

My biggest concern is that label sets and meta data is mapped to a *list* of one-element maps, rather than just a map; see discussion inline.

Best regards and a happy new year 🎆 
—octo

src/utils_lua.c
kenhys: I 'm not sure that label set returns unique key or not.
If label set returns  `label[N-1]={name="foo", value="will be lost"}, label[N] = { name= "foo", vaule="something"}` in C world, in above example, it would be `label = { foo = "something" }` as outcome.

src/utils_lua.c
kenhys: Thanks, fixed now.

src/utils_lua.c
kenhys: Fixed it.

src/utils_lua.c
kenhys: Thanks. Fixed it.

src/utils_lua.c
kenhys: Fixed it. Thanks!

src/utils_lua.c
kenhys: No need  to export for Lua world. I've removed. thanks!

src/utils_lua.c
octo: Label keys are unique within a label set, so this overwrite behavior should not happen. I think that `label = { foo = "something", bar = "something else" }` is easier to work with.

src/utils_lua.c
```diff
@@ -274,38 +235,170 @@ int luaC_pushvalue(lua_State *L, value_t v, int ds_type) /* {{{ */
...
+    meta => {
+      [1] => {
+        key => ...,
+        value => ...,
+      },
+      ...
+      [N] => {
+        ...
+      }
+    }
```
octo, line 276: As with labels: metadata keys are unique, and I think we should simplify this to:

```lua
meta => {
  key0 => "value0",
  key1 => 42,
  …
}
```

src/utils_lua.c
kenhys: Thanks, I've got it.

src/utils_lua.c
kenhys: Fixed it.

eero-t:
Some test code for this functionality would also be nice, but that could be a separate PR.

octo:
+1, tests would be nice to have, but I won't block on them.

I do think that we need to update the documentation a bit though. Could you document the updated behavior in `src/collectd-lua.pod`? I think it would be ideal to add a "Data Types" section and there talk about how `metric_family_t` is mapped to Lua. I think with that change this should be good to go.

Thanks for your continued work and best regards
—octo

kenhys:
I've added explanation about data types and rebased with collectd-6.0.

kenhys:
Except "Labels", CI has passed.

octo (approved):
Looks greak, thanks @kenhys!

PR #3802, opened 2021-01-06
nut.c: add support for ups.realpower

f00b4r0:
ChangeLog: plugin nut: add support for ups.realpower

Some UPS report a global "realpower" value corresponding to actual Watts
instead of VA as part of the "ups." key.
The Eaton Ellipse series is such an example: where the device does not
report current values but only voltages, watts and VA.
This commit makes that value available in reported statistics using the
"watt-ups" type_instance, matching 64321ae

```
% upsc ups
battery.charge: 100
battery.charge.low: 10
battery.runtime: 942
battery.type: PbAc
device.mfr: EATON
device.model: Ellipse PRO 650 
device.serial:
device.type: ups
driver.flag.pollonly: enabled
driver.name: usbhid-ups
input.voltage: 233.0
input.voltage.extended: no
outlet.1.desc: PowerShare Outlet 1
outlet.1.id: 2
outlet.1.status: on
outlet.1.switchable: no
outlet.2.desc: PowerShare Outlet 2
outlet.2.id: 3
outlet.2.status: on
outlet.2.switchable: no
outlet.desc: Main Outlet
outlet.id: 1
outlet.switchable: no
output.frequency: 50.0
output.frequency.nominal: 50
output.voltage: 229.0
output.voltage.nominal: 230
ups.beeper.status: enabled
ups.delay.shutdown: 20
ups.delay.start: 30
ups.firmware: 01.04.0012
ups.load: 24
ups.mfr: EATON
ups.model: Ellipse PRO 650 
ups.power: 147
ups.power.nominal: 650
ups.productid: ffff
ups.realpower: 110
ups.serial:
ups.status: OL
ups.timer.shutdown: 0
ups.timer.start: 0
ups.vendorid: 0463
```

### DIFF
--- a/src/nut.c
+++ b/src/nut.c
@@ -326,6 +326,8 @@ static int nut_read(user_data_t *user_data) {
         nut_submit(ups, "percent", "load", value);
       else if (strcmp("ups.power", key) == 0)
         nut_submit(ups, "power", "ups", value);
+      else if (strcmp("ups.realpower", key) == 0)
+        nut_submit(ups, "power", "watt-ups", value);
       else if (strcmp("ups.temperature", key) == 0)
         nut_submit(ups, "temperature", "ups", value);
     }


f00b4r0:
While I'm there, purely for my information, is there a specific reason not to simplify the string comparisons like e.g.:

From:
```
    } else if (strncmp("output.", key, 7) == 0) {
      if (strcmp("output.current", key) == 0)
        nut_submit(ups, "current", "output", value);
      else if (strcmp("output.frequency", key) == 0)
        nut_submit(ups, "frequency", "output", value);
      ...
```

To:
```
    } else if (strncmp("output.", key, 7) == 0) {
      key += 7;
      if (strcmp("current", key) == 0)
        nut_submit(ups, "current", "output", value);
      else if (strcmp("frequency", key) == 0)
        nut_submit(ups, "frequency", "output", value);
      ...
```
?
It looks like this was the intent of the code, otherwise I'm not sure what's the reason for the `strncmp()` cases, since `strcmp()` will break at first mismatch character anyway?

That'd be a first step optimization that would make this routine more efficient and the resulting binary smaller (reading the code it seems there's further room for optims, so maybe there's a good reason to keep it that way?).

Cheers

eero-t:
@mrunge you merged https://github.com/collectd/collectd/pull/3889 which is superset of the changes here, so this can be closed.

@f00b4r0 if you want to optimize the lookup, replace it  with bsearch() on a static const sorted array of the keys & corresponding metric names, and make PR out of that.

PR #3818, opened 2021-01-12
[collectd 6] cpusleep: migration to v6.0

manuelluis:
ChangeLog: cpusleep plugin: migration to v6.0

### DIFF
--- a/src/cpusleep.c
+++ b/src/cpusleep.c
@@ -36,15 +36,23 @@
 #include "utils/common/common.h"
 #include <time.h>
 
-static void cpusleep_submit(derive_t cpu_sleep) {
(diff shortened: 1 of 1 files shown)


manuelluis:
Maybe use `cpusleep_seconds_total ` instead of `cpusleep_milliseconds_total` ?

elfiesmelfie:
I tested this in a CentOS 8 container, and both versions reported the same metrics.

> Maybe use `cpusleep_seconds_total ` instead of `cpusleep_milliseconds_total` ?

I think cpusleep_milliseconds_total is fine.

mrunge (approved):

PR #3819, opened 2021-01-13
[collectd 6] processes: migration to v6.0

manuelluis:
ChangeLog: processes plugin: migration to v6.0

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -905,84 +905,84 @@ static void ps_submit_state(gauge_t *proc_state) {
 }
 
 /* submit info about specific process (e.g.: memory taken, cpu usage, etc..) */
-static void ps_metric_append_proc_list(metric_family_t *fams_proc[],
+static void ps_metric_append_proc_list(metric_family_t *fams_proc,
                                        procstat_t *ps) {
   metric_t m = {0};
 
   metric_label_set(&m, "name", ps->name);
 
   m.value.gauge = ps->vmem_size;
-  metric_family_metric_append(fams_proc[FAM_PROC_VMEM_SIZE], m);
+  metric_family_metric_append(&fams_proc[FAM_PROC_VMEM_SIZE], m);
 
   m.value.gauge = ps->vmem_rss;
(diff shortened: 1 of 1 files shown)


octo:
Hi Manuel,

thank you so much for your work. I see that you have created a bunch of PRs, migrating various plugins to collectd 6. I appreciate your work very much and would like to work through the PRs to merge them. Sorry I haven't been very responsive in the past, but I hope that I'll be able to spend some more time and review your work.

I'm starting with this plugin, because I think it's quite fundamental but also very complex. If you have any plugins we should treat with priority, please let me know!

I have rebased this branch on top of the current `collectd-6.0` to allow our builders to run successfully.

Best regards
—octo

src/processes.c
```diff
@@ -346,6 +346,58 @@ int getargs(void *processBuffer, int bufferLen, char *argsBuffer, int argsLen);
...
+  FAM_PROC_IO_DISKR,
+  FAM_PROC_IO_DISKW,
+  FAM_PROC_FILE_HANDLES,
+  FAM_PROC_FILE_HANDLES_MAPPED,
+  FAM_PROC_CTX_VOLUNTARY,
+  FAM_PROC_CTX_INVOLUNTARY,
+  FAM_PROC_DELAY_CPU,
+  FAM_PROC_DELAY_BLKIO,
+  FAM_PROC_DELAY_SWAPIN,
+  FAM_PROC_DELAY_FREEPAGES,
```
octo, line 374: I think a bunch of these should be combined, so that we end up with the following families:

* vmem
* cpu (usage/utilization)
* processes
* threads
* memory faults
* I/O
* file handles
* context switches
* delay

I think a good rule of thumb is: if metrics use the same unit and could conceivably be aggregated together, then they should be in the same metric family.

octo (commented):
Overall this is a great start. Thanks again @manuelluis! There are two things we need to change before we can merge this PR:

1. We need to combine some of the metrics into fewer metric families. The `metric_family_append` function is a great asset when doing that.
2. The metric families should be named so conform to OpenTelemetry's naming conventions as much as possible.

Best regards,
—octo

PR #3821, opened 2021-01-13
[collectd 6] protocols: migration to v6.0

manuelluis:
ChangeLog: protocols plugin: migration to v6.0


kwiatrox (approved):
Looks good and works well in similar way to old implementation.

mrunge (approved):
Thank you

PR #3822, opened 2021-01-13
exec plugin: remove check against running scripts as root.

FeepingCreature:
*Every other plugin* can run with full root privileges. Collectd even asks to be run as root! Python, Perl - is there any reason Bash scripts are inherently less safe than Python scripts? The python plugin doesn't even *let you* drop privileges. This check adds extra effort for no documented reason, especially on docker where by default only a root account is created. At least, this limit needs to be better justified.

### DIFF
--- a/src/collectd-exec.pod
+++ b/src/collectd-exec.pod
@@ -287,11 +287,6 @@ to make use of collectd's more powerful interface.
 
 =item *
 
-The user, the binary is executed as, may not have root privileges, i.E<nbsp>e.
-must have an UID that is non-zero. This is for your own good.
-
-=item *
-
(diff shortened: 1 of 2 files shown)


mrunge:
Thank you for your PR.

I disagree with this. I haven't really seen a case where root access itself is **required**. This check is only a tiny bar to prevent you shooting yourself in the foot.

The python plugin should be fixed.

FeepingCreature:
I guess my problem is that I really don't see what sort of foot-shooting you're envisioning here. I have a hard time seeing a scenario where an attacker can make you run a collectd script as root but not approximately just as easily run any ordinary script as root.

PR #3832, opened 2021-01-27
Fix #3831 by adding commas to the infiniband types

### LINKED ISSUE #3831: ib_octets and ib_packets format inconsistent with other types
In collectd 5.12 we got the infiniband plugin.

The ib_octets and ib_packets types in types.db have two values, delimited by a space. However, other types use a comma and space.

Interestingly, the types.db documentation states:
> Each line consists of two or more fields delimited by spaces and/or horizontal tabs

So these lines are technically correct and all the existing lines are wrong - however tools like bucky read types.db and get parse errors.

I think we should stick with comma+space, and then consider either making that the standard, OR, as part of major release change the convention to match the standard.

Additionally, I wonder if we need ib_octets and ib_packets - presumably if_octets/packets would be suitable? Looking at the documentation and collectd contributing guidelines we are discouraged from adding new types.

I'll send in a PR for adding the comma, in any case.

nward:
ChangeLog: infiniband: Add missing commas  to the infiniband types.

See #3831 for a deeper discussion of this.

### DIFF
--- a/src/types.db
+++ b/src/types.db
@@ -119,8 +119,8 @@ ib_errors               value:DERIVE:0:U
 ib_lid                  value:GAUGE:0:32767
 ib_lid_mask_count       value:GAUGE:0:7
 ib_link_error_recovery  value:DERIVE:0:U
-ib_octets               rx:DERIVE:0:U tx:DERIVE:0:U
-ib_packets              rx:DERIVE:0:U tx:DERIVE:0:U
+ib_octets               rx:DERIVE:0:U, tx:DERIVE:0:U
+ib_packets              rx:DERIVE:0:U, tx:DERIVE:0:U
 ib_phys_state           value:GAUGE:0:15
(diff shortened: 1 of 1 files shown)


mrunge:
Thank you for your PR. This looks good to me. The CI issue is known and will hopefully get fixed rather sooner than later.

mrunge (approved):

elfiesmelfie:
mrunge: this looks good to me, and considering the only thing that was blocking appears to be el6, which is out-of-support and is being/has been removed from CI, I'd recommend merging main into this branch to update it, and re-triggering the CI.

elfiesmelfie (approved):
This change looks good to me, and it *should* pass CI if el6 is removed.

mrunge (approved):

mrunge:
Thank you for your PR.

PR #3833, opened 2021-01-27
[collectd 6] buddyinfo: migration to v6.0

manuelluis:
ChangeLog: buddyinfo plugin: migration to v6.0

### DIFF
--- a/src/buddyinfo.c
+++ b/src/buddyinfo.c
@@ -63,35 +63,23 @@ static int buddyinfo_config(const char *key, const char *value) {
   return 0;
 }
 
-static void buddyinfo_submit(const char *zone_fullname, const char *zone,
-                             const char *size, const int freepages) {
-  value_list_t vl = VALUE_LIST_INIT;
-  value_t value = {.gauge = freepages};
-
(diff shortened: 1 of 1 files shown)


src/buddyinfo.c
```diff
@@ -63,36 +63,24 @@ static int buddyinfo_config(const char *key, const char *value) {
   return 0;
 }
 
-static void buddyinfo_submit(const char *zone_fullname, const char *zone,
-                             const char *size, const int freepages) {
-  value_list_t vl = VALUE_LIST_INIT;
-  value_t value = {.gauge = freepages};
-
-  if (ignorelist_match(ignorelist, zone) != 0)
```
kwiatrox, line 71: It looks like the whole functionality with ignorelist for zone was removed. Now the plugin has the configuration that does nothing. Probably it should be kept after migration. Other than that the change looks OK.

manuelluis: Fixed.

kwiatrox (approved):

mrunge (approved):

mrunge:
Thank you Kamil and Manuel.

PR #3840, opened 2021-02-08
collection3: fix inappropriate usage of CGI:param

kenhys:
ChangeLog: When using collection3 as a CGI, the following error is sent to logs
repeatedly.

  FastCGI sent in stderr: "CGI::param called in list context from
  /usr/share/doc/collectd-core/examples/collection3/lib/Collectd/Graph/Common.pm
  line 529, this can lead to vulnerabilities. See the warning in
  "Fetching the value or values of a single named parameter" at
  /usr/share/perl5/CGI.pm line 412"

This is caused inappropriate usage of param(), it should be handled as
a scalar or should be treated by multi_param() explicitly.

Signed-off-by: Kentaro Hayashi <kenhys@gmail.com>

### DIFF
--- a/contrib/collection3/lib/Collectd/Graph/Common.pm
+++ b/contrib/collection3/lib/Collectd/Graph/Common.pm
@@ -526,7 +526,7 @@ sub get_selected_files
   for (qw(hostname plugin plugin_instance type type_instance))
   {
     my $part = $_;
-    my @temp = param ($part);
+    my @temp = multi_param ($part);
     if (!@temp)
     {
       next;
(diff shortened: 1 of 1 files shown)


systemcrash:
my Perl is a bit rusty, but looks Ok.

kenhys:
FYI: 

collectd-core: CGI::param called in list context raise error in Collectd/Graph/Common.pm
version graph
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=982294

Fixed in collectd/5.12.0-7 and merged into Debian bullseye or later.
https://packages.debian.org/search?suite=bullseye&searchon=names&keywords=collectd

eero-t:
@octo `Common.pm` was your code originally, maybe you could take a look?

(Patch looks still the same in latest Debian sid, against collectd 5.12.0-15.)

octo (approved):
I haven't used the CGI module in a long time, but with a quick look in the CGI module docs the change looks good to me. Thanks @kenhys!

jnahmias:
fixes #2355

PR #3844, opened 2021-02-12
New plugin for reading diskstats

kwiatrox:
ChangeLog: New plugin providing some specific calculated disks metrics.

New Diskstats plugin is optimized for providing data allowing to determine fitness factors of workload to particular types of SSD drives (e.g. based on NAND TLC/QLC media, or Intel 3DXP or Z-NAND). The new plugin precisely calculates a number of derived metrics (like MB read/s, MB written/s, MB discarded/s, average IO latency time, average read/write IO latency time, average queue depth) on a client side providing additional accuracy.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -996,6 +996,25 @@ disk_la_LIBADD += -lperfstat
 endif
 endif
 
+if BUILD_PLUGIN_DISKSTATS
+pkglib_LTLIBRARIES += diskstats.la
+diskstats_la_SOURCES = src/diskstats.c
+diskstats_la_CFLAGS = $(AM_CFLAGS)
+diskstats_la_CPPFLAGS = $(AM_CPPFLAGS)
+diskstats_la_LDFLAGS = $(PLUGIN_LDFLAGS)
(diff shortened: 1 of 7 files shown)


mrunge:
Thank you for your PR. I am a bit confused here, since this plugin basically adds another disk plugin? Wouldn't it make sense to include the metrics you are interested in in the disk plugin itself?

kwiatrox:
@mrunge Introducing those metrics to legacy Collectd Disk plugin could cause breaking backward compatibility. This plugin provides a very specific set of metrics, some of them averaged over moving window. On the other hand disk plugin is generic, verbose and flexible using many different input methods, while this plugin is limited to only one source of data. Merging those two approaches together would be very difficult. Thus we decided that more feasible scenario is to create new plugin.

faxm0dem:
One of the concerns I have is the computed part : often people compute those downstream. One often wants raw values, and other derived ones like moving average, but I think collectd's mission is especially on the raw side. What is your opinion and the rationale that drove you to implement those ?

PR #3846, opened 2021-02-16
[collectd 6] tail_csv: migration to v6.0

manuelluis:
ChangeLog: tail_csv plugin: migration to v6.0

**TODO:**

- [ ] maybe rename **Metric** block to **Result** to avoid confusion with _Metric_ keys


systemcrash:
Largely looks OK - have these changes been tested in operation?


Also, is it much more work to:
>     * rename **Metric** block to **Result** to avoid confusion with _Metric_ keys

would make for a clean jump to 6  😄

PR #3847, opened 2021-02-17
[collectd 6] sigrok: migration to v6.0

manuelluis:
ChangeLog: sigrok plugin: migration to v6.0

I have also set the minimum version of libsigrok to 0.5.0


systemcrash:
This largely looks OK - have these changes been tested in operation?

PR #3848, opened 2021-02-18
[collectd 6] irq: migration to v6.0

manuelluis:
ChangeLog: irq plugin: migration to v6.0


systemcrash:
Innocent question: but have these improvements been tested in action?

PR #3852, opened 2021-02-23
Add an Interval option to several plugins

efuss:
Add an `Interval` configuration option to all plugins that lacked one yet used `plugin_register_complex_read()`, i.e. _apache_, _dpdk_telemetry_, _dpdkevents_, _dpdkstat_, _intel_pmu_, _intel_rdt_, _ipmi_, _java_, _logparser_, _lua_, _mcelog_, _memcached_, _mysql_, _nut_, _openldap_, _openvpn_, _ovs_events_, _ovs_stats_, _pcie_errors_, _redfish_, _redis_, _routeros_, _varnish_, _virt_.

Note that _intel_rdt_'s documentation already claimed to have an `Interval` option despite not being actually implemented.

Most of this is not even compile-tested (I only needed it for _nut_) because I can't build the Linux-only plugins on NetBSD.

ChangeLog: multiple plugins: An "Interval" configuration option was added.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -978,11 +978,13 @@ Since its C<mod_status> module is very similar to Apache's, B<lighttpd> is
 also supported. It introduces a new field, called C<BusyServers>, to count the
 number of currently connected clients. This field is also supported.
 
-The configuration of the I<Apache> plugin consists of one or more
-C<E<lt>InstanceE<nbsp>/E<gt>> blocks. Each block requires one string argument
+The configuration of the I<Apache> plugin consists of an optional C<Interval>
+specification and one or more C<E<lt>InstanceE<nbsp>/E<gt>> blocks. Each
+C<E<lt>InstanceE<nbsp>/E<gt>> block requires one string argument
 as the instance name. For example:
 
  <Plugin "apache">
+   <Interval 60>
    <Instance "www1">
      URL "http://www1.example.com/mod_status?auto"
    </Instance>
@@ -3056,6 +3058,7 @@ be found here: http://dpdk.org/doc/guides/sample_app_ug/keep_alive.html
 B<Synopsis:>
 
  <Plugin "dpdkevents">
+   Interval 30
    <EAL>
      Coremask "0x1"
      MemoryChannels "4"
@@ -3078,6 +3081,14 @@ B<Synopsis:>
 
 B<Options:>
 
+=over 4
+
+=item B<Interval> I<Interval>
+
+Sets the interval (in seconds) in which values will be collected.
+By default the global B<Interval> setting will be used.
+
+=back
 
 =head3 The EAL block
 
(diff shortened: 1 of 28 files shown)


src/collectd.conf.pod
```diff
@@ -978,11 +978,13 @@ Since its C<mod_status> module is very similar to Apache's, B<lighttpd> is
...
 
-The configuration of the I<Apache> plugin consists of one or more
-C<E<lt>InstanceE<nbsp>/E<gt>> blocks. Each block requires one string argument
+The configuration of the I<Apache> plugin consists of an optional C<Interval>
+specification and one or more C<E<lt>InstanceE<nbsp>/E<gt>> blocks. Each
+C<E<lt>InstanceE<nbsp>/E<gt>> block requires one string argument
 as the instance name. For example:
 
  <Plugin "apache">
+   <Interval 60>
```
mrunge, line 987: Typo here?

efuss:
> Typo here?
Yes. And in two comments.
Fixed in 1a38b0cf.

mrunge:
Thank you! I would have bet there is an interval setting for each plugin and that it's configurable everywhere....

efuss:
> I would have bet there is an interval setting for each plugin and that it's configurable everywhere....
Well, sort of (as I discovered only after implementing it in the configuration): You can, instead of
```
LoadPlugin foo
```
do
```
<LoadPlugin foo>
	Interval 60
</LoadPlugin>
```

Note that there are still a lot of plugins that use `plugin_register_read()`, not `plugin_register_complex_read` and I didn't add an Interval option to those.

But I noticed all that only after patching all the plugins using `plugin_register_read()`, so well...

systemcrash:
Seems fit for purpose. 

Can you fix the CI errors?: 
- clang-format — Please run: contrib/format.sh src/dpdk_telemetry.c src/dpdkstat.c src/logparser.c src/mcelog.c src/nut.c src/openvpn.c src/pcie_errors.c sr… 
-

efuss:
Hm, I did run `clang-format --style=file -i` on all the files I touched, but CI still seems to choke on it.
Does CI use formatting rules different from the ones in `.clang-format`?

systemcrash:
I guess just run
``` Please run: contrib/format.sh src/dpdk_telemetry.c src/intel_pmu.c src/intel_rdt.c...`

like it asks. Take it from there.

efuss:
> Take it from there. 
Running the Web service doesn't (easily) work fpr me.
I do have clang-format installed plus the project provides a .clang-format file.
The project's CI should be consistent with the project's file.

systemcrash:
> Running the Web service doesn't (easily) work fpr me. I do have clang-format installed plus the project provides a .clang-format file. The project's CI should be consistent with the project's file.

It's a local shell script, in your checked out folder. And until your probably correct "should be" is implemented, then there's not much else we can do.

efuss:
> It's a local shell script, in your checked out folder.
Which runs a web service on collectd.org.

> then there's not much else we can do. 
The web service is most probably implemented using clang-format and a .clang-format file.
The git checkout contains another .clang-format file which is, I guess, inconsistent with the one used by the web service.

mrunge:
@efuss I feel your pain, I experienced the same. At some point, these issues were gone. IIRC I switched to running each file alone through the reformat and not running the command with all involved files in one go.

PR #3854, opened 2021-02-25
Introduce ReportProcs* options allowing NFS metrics to be limited to given set of procedures.

efuss:
For the NFS plugin, introduce new configuration options `ReportProcs` and `ReportProcsV[234]` to restrict metric collection to a given set of NFS procedures, either globally or for a specific NFS version.
Omitting the new options will collect data for all procedures as before.

This allows limiting the amount of data reported while still monitoring “interesting” procedures (_write_, _commit_) at full resolution.

This has only been tested on NetBSD and (backported to 5.7.1) on Debian Stretch.

While there, correct a typo in the nfs plugin documentation.

ChangeLog: nfs plugin: options ReportProcs and ReportProcsV[234] were added to allow limiting NFS metric collection to a given set of procedures.


mrunge:
Thank you for your PR @efuss 

CI is failing on this.
Could you also rebase your patch on main? This will fix the issue with CentOS 6 failing. 
Also please run: contrib/format.sh src/nfs.c

PR #3855, opened 2021-02-26
[collectd 6] ethstat: migration to v6.0

manuelluis:
ChangeLog: ethstat plugin: migration to v6.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3272,8 +3272,6 @@ B<Synopsis:>
 
  <Plugin "ethstat">
    Interface "eth0"
-   Map "rx_csum_offload_errors" "if_rx_errors" "checksum_offload"
-   Map "multicast" "if_multicast"
  </Plugin>
 
 B<Options:>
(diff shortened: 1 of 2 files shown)


bkotlowski:
Is removing all mapping functionality necessary for migration?

manuelluis:
I think that the mapping was a necessity on the old metrics model to map the metrics to know types.
With the new model I don't think that the renaming is necessary.

But, we can have some filtering like: 
```
<Plugin "ethstat">
  Interface "eth0"
  Value "collisions"
  Value "/^(rx|tx)_queue.*/"
  IgnoreSelected false
</Plugin>
```

bkotlowski:
I am currently working on developing the ethstats plugin and I can see an application, in some specific cases, for mapping, not related to selecting specific metrics, but just changing the name of the displayed counter.

manuelluis:
I don't really know if the ethstat plugin has an intrinsic need to rename the metrics, but other plugins may have that need too.
You can rename the metrics with chains, after the chains migration it's could be something like this: 

```
<Chain "Rename">
  <Rule "rename">
    <Match regex>
      Metric "^ethstat_tx_carrier_errors_total$"
    </Match>
    <Target "set">
      Metric "tx_carrier_errors"
    </Target>
  </Rule>
</Chain>
```

bkotlowski:
I've been working on extending the plugin to add standard interface statistics because they won't be available in ethtool soon, and I'd like to keep plugin backward compatible. For the same reason I would like to rename some metrics (the same metrics that were removed from ethtool may have a different names in sysfs, e.g. no additional underscore or something similar) using configuration. In this case, the mapping would help me create a example configuration that user could use after updating ethtool and still have the same metric list.

systemcrash:
The linked issue above mentions:

> * If possible, try to find a _Prometheus exporter_ reporting the same metric(s) and use the same naming schema.

> * Update the _ethstat plugin_ to produce a `metric_family_t` and call `plugin_dispatch_metric_famliy()`.

where `famliy` should be `family`, I assume. it looks like those thresholds have been met. How about the metric names @bkotlowski ?

eero-t:
What's the relation of this "ethstat" PR to the "ethstat" commit on "collectd-6.0" branch: https://github.com/collectd/collectd/commit/6d90bf4c30a3c9c9b34b70c8b61e26df5fb4d987 ?

(authored by @bkotlowski and committed by @mrunge)

systemcrash:
I would guess that that PR targets 6, and has been converted to v6 code.

PR #3859, opened 2021-03-08
[collectd 6] thermal: migration to v6.0

manuelluis:
ChangeLog: thermal plugin: migration to v6.0

### DIFF
--- a/src/thermal.c
+++ b/src/thermal.c
@@ -225,7 +225,6 @@ static int thermal_init(void) {
   } else if (access(dirname_procfs, R_OK | X_OK) == 0) {
     thermal_fs = THERMAL_PROCFS;
     ret = plugin_register_read("thermal", thermal_read);
-  } else {
   }
 
   return ret;
--- a/src/thermal.c
+++ b/src/thermal.c
@@ -38,26 +38,25 @@ static const char *const dirname_procfs = "/proc/acpi/thermal_zone";
 static bool force_procfs;
 static ignorelist_t *device_list;
 
+enum thermal_fs_e {
+  THERMAL_NONE = 0,
+  THERMAL_PROCFS,
+  THERMAL_SYSFS,
+};
+
+static enum thermal_fs_e thermal_fs;
+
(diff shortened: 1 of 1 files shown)


systemcrash:
Looks OK @mrunge

elfiesmelfie:
I tested this against CentOS/8 using the write_http lugint o compare the metrics, and everything looks reasonable. 
Here's the sample output for the two types of metrics (temperature of thermal zones and state of cooling device).

```
PUTVAL fbae30cc-2f20-11b2-a85c-819293100691/thermal-cooling_device1/gauge interval=10.000 1636380481.471:0
PUTMETRIC collectd_thermal_gauge type=GAUGE time=1636380481.646 interval=10.000 label:instance="fbae30cc-2f20-11b2-a85c-819293100691" label:thermal="cooling_device1" 0

PUTVAL fbae30cc-2f20-11b2-a85c-819293100691/thermal-thermal_zone2/temperature interval=10.000 1636380481.471:38.05
PUTMETRIC collectd_thermal_temperature type=GAUGE time=1636380481.646 interval=10.000 label:instance="fbae30cc-2f20-11b2-a85c-819293100691" label:thermal="thermal_zone2" 38.05
```

IMHO, it would be more useful to have the metric name changed for the cooling device from ``gauge`` to ``state``, to reflect the stat being exposed by sysfs [1].  I think this suggestion is within the scope of the 5->6 migration, given that name changes are expected for alignment to prometheus naming.


[1] https://www.kernel.org/doc/html/latest/driver-api/thermal/sysfs-api.html#cooling-device-attributes

elfiesmelfie:
> What are the repercussions of doing so? While state may well reflect the value's origin, `gauge` is more readily intuitive for the end-user at Prometheus, since gauge also represents a value which can increase or decrease (arbitrarily). That seems to be what this value does. Just my $0.02. ( So perhaps `cooling_state_gauge`? )

IINM, there's scope in the migration to update naming. IMHO, having cooling in the name is more intuitive. I'm not opposed to keeping gauge in there, so ``cooling_state_gauge`` would be fine by me.

PR #3862, opened 2021-03-10
[collectd 6] new plugin info

manuelluis:
ChangeLog: new plugin info to define static metrics


configure.ac
```diff
@@ -1,4 +1,4 @@
-dnl Process this file with autoconf to produce a configure script.
+dnl Proess this file with autoconf to produce a configure script.
```
mrunge, line 1: Process, there is a missing 'c' here

mrunge (changes requested):
There's a typo here, (pointed out), but otherwise this is good to go.

src/collectd.conf.pod
```diff
@@ -3824,6 +3824,47 @@ B<except> the ones matched.
 
 =back
 
+=head2 Plugin C<info>
+
+The C<info> plugin allow to define static metrics. By default sends the metric
```
systemcrash, line 3829: Should be:
```
The C<info> plugin allows to define static metrics. By default it sends the metric
B<collectd_info> with collectd version.
```

PR #3864, opened 2021-03-11
ubi: add percent calculation

feckert:
ChangeLog: ubi: Add percentage value calculation for the bad blocks

@pticon With this change, the percentage error of the flash can be calculated directly in the collectd ubi plugin.
Until now, this had to be done differently.

With this change, the threshold plugin can now triggers an event if there are too many bad erase blocks.
This means that the flash degenerates.

### DIFF
--- a/src/ubi.c
+++ b/src/ubi.c
@@ -84,9 +84,9 @@ static void ubi_submit(const char *dev_name, const char *type, gauge_t value) {
   plugin_dispatch_values(&vl);
 } /* void ubi_submit */
 
-static int ubi_read_dev_attr(const char *dev_name, const char *attr) {
+static int ubi_read_dev_attr(const char *dev_name, const char *attr,
+                             int *value) {
   FILE *f;
-  int val;
   char
       str[sizeof(SYS_PATH) + strlen(dev_name) + sizeof("/") + strlen(attr) + 1];
   int n;
@@ -98,7 +98,7 @@ static int ubi_read_dev_attr(const char *dev_name, const char *attr) {
     return -1;
   }
 
-  n = fscanf(f, "%d", &val);
+  n = fscanf(f, "%d", value);
   fclose(f);
 
   if (n != 1) {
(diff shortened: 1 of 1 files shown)


feckert:
@pticon thanks for review and feeback,
I have now added a check and emit an error if the value is zero.

pticon:
@feckert, thanks for the changes and the contribution.
From my side, the changes look fine but you still need to resolve the check errors.

feckert:
@pticon thanks for the hint.
I have all checks fixed except https://github.com/collectd/collectd/pull/3864/checks?check_run_id=2094242461
But I think this is not my fault, I do not know what I have todo.

feckert:
@pticon As I can see there is already an issue for this https://github.com/collectd/ci-docker/issues/15
And there is also a fix pending https://github.com/collectd/collectd/pull/3784
And the following line is criticised https://github.com/collectd/ci-docker/blob/a5b9d38b9fe0d5e55840a1beb32f0d863c413179/checks/check-built-plugins.sh#L14

That's all I found out. I hope it helps

pticon:
@feckert , thanks for the investigation.
I think it is okay then.

pticon (approved):

feckert:
@pticon @mrunge In OpenWrt, [this](https://github.com/openwrt/packages/pull/15235) has been merged for some time.
How does it look from the collectd side, should I do something?

The problem with the CI has nothing to do with my change.
I have already described this in this [comment](https://github.com/collectd/collectd/pull/3864#issuecomment-797774336).

I have [backported](https://github.com/openwrt/packages/commits/openwrt-21.02/utils/collectd) this already to 21.02 

If I still have to do something, I would like to backport the requested changes also to OpenWrt for the next upcoming stable release 21-02.
It has already reached rc1.

mrunge:
I'd love to be able to merge this. Unfortunately, the CI is broken and unfixable for me.

@octo would you be able to remove the travis config/integration into the collectd-5 branch? That is blocking is from fixing the CI for collectd 5.

@feckert collectd is migrating very slowly to the new version 6 (not released yet). This plugin will probably need small adjustments for the new format.

feckert:
> @feckert collectd is migrating very slowly to the new version 6 (not released yet). This plugin will probably need small adjustments for the new format.

@mrunge thanks for your reply.
I don't know much about the collectd development process for now.
But what do I have to do?
Will the plugin be ported automatically or do I have to do this in the new version branch with an additional merge request.

We have other plugins here in openwrt that are not yet upstream.

systemcrash:
Looks good - can merge this one now. Fix 6 migration in a separate step.

mrunge:
Thank you, I'm going to merge this. CentOS 6 has been fixed for some time now, the check is independent from the code here.

feckert:
Thank you :+1:

PR #3873, opened 2021-04-11
Adjust SFOS build scripts

rinigus:
This makes it possible to build on local SDK and OBS


rinigus:
Sorry, PR sent to wrong repository

mrunge:
Thank you for your PR. There are a few additional plugins included, which may be useful in collectd in general. Would you mind to propose them here as addition?

rinigus:
I could. which plugins are of interest? I suspect some of them were already proposed, but never got merged.

mrunge:
This was more an offer to get the additional plugins upstream, i.e. reduce the delta between your fork and upstream.

rinigus:
I will have to look through. Some plugins are Sailfish specific as well as rewrite of the main loop of the daemon. The latter was needed to allow mobile device to suspend/resume for collectd to get the stats as intended.

As for plugins that maybe of interest for others, that I have to check.

PR #3874, opened 2021-04-15
Fix #3472 - Correct LD flags for various libs

### LINKED ISSUE #3472: configure links to system netsnmp on MacOS, ignoring --with-libnetsnmp
*   Version of collectd: master
*   Operating system / distribution: MacOS Mojave
*   Kernel version (if applicable):

## Expected behavior

After building collectd as follows:

CFLAGS="-g -O0" ./configure --prefix=/tmp/collectd --without-java --with-librrd=/opt/local --enable-rrdtool --with-libssl=/tmp/collectd --with-libnetsnmp=/tmp/collectd

It is expected that snmp.so is linked to /tmp/collectd/lib/libnetsnmp.35.dylib

## Actual behavior

snmp.so is linked to /usr/lib/libnetsnmp.25.dylib instead:

Little-Net:collectd minfrin$ otool -L /tmp/collectd/lib/collectd/snmp.so 
/tmp/collectd/lib/collectd/snmp.so:
	/usr/lib/libnetsnmp.25.dylib (compatibility version 26.0.0, current version 26.2.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.250.1)

snmp plugin fails with an out of memory error, as the version 36 header files are used against the 25 library.
## Steps to reproduce

*   Build using ./configure as above.
*   Check library link with otool -L

craiggowing:
ChangeLog: Correct LD flags for some custom library paths

See #3472 for the original issue

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2071,7 +2071,7 @@ AC_ARG_WITH([cuda],
   [
     if test "x$withval" != "xno" && test "x$withval" != "xyes"; then
       with_cuda_cppflags="-I$withval/include"
-      with_cuda_ldflags="-I$withval/lib"
+      with_cuda_ldflags="-L$withval/lib"
       with_cuda="yes"
     else
       with_cuda="$withval"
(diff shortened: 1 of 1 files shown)


mrunge (approved):
Ugh, good catch. Thank you for your fix.

mrunge:
Unfortunately, the CI is known to be broken, and fixing it is out of my reach. 

Especially, failing CI is not because of this patch breaking it.

PR #3883, opened 2021-06-11
Add 'gpu_sysman' plugin for (Intel) GPU metrics

eero-t:
ChangeLog: gpu_sysman plugin: Add 'gpu_sysman' plugin for (Intel) GPU metrics

### New Sysman plugin for collectd

*Pre-production beta version*.

Plugin links against Level Zero loader:
* https://github.com/oneapi-src/level-zero

And provides GPU metrics using Level Zero Sysman API:
* https://spec.oneapi.com/level-zero/latest/sysman/api.html

### Metrics reporting

One of the uses for this plugin is providing GPU metrics for Prometheus in Kubernetes cluster environment.  However, some metric values can fluctuate between their possible min & max values much more frequently than those values can be reported in such an environment.

Because of this, plugin can be configured to query some GPU metrics (currently frequency & memory usage) internally at higher rate + cache the results internally, and provide aggregate metrics of the cached data (currently min/max for given period) at the reporting intervals.

If there are errors with some metric types when their values are queried, plugin disables that metrics type for corresponding GPU. Each of the GPU metrics type queries can also be disabled from plugin configuration.

### Metrics types

Plugin uses single helper function with GAUGE type to report all metrics. "error_count" GAUGE type is added to "types.db" because there was no pre-existing GAUGE type for errors.

### Supported metrics

Plugin provides most of the GPU metrics available through Sysman v1.1 API:
- Memory usage (free/used)
- Memory BW
- Frequency (request/actual)
- Frequency throttling
- Engine utilization
- Temperature
- Power usage
- Error counters (RAS)

Most of the metrics have been tested with data from real HW, but for few only unit testing was possible.

Metrics names are derived from Sysman API like this: <device/subdev[ID]>[-location][-category]<-type>[-min/max]

Resulting in for example following names:
- device-watts
- subdev00-local-HBM-used-max

(min/max postfixes are used for aggregated metrics sampled at higher internal frequency than they are reported at.)

### Possible other metrics

Support could be later added to other metrics available through Sysman API, such as:
- fabric port read/write BW
- PSU voltage, amperage, temp
- GPU voltage
- fan speed
- memory health

### Quality expectations

GCC gives no warnings for plugin when much stricter warnings are enabled than what collectd build used.

Over 80% unit-testing code coverage (95% with -O3 builds), with no warnings when unit-tests are run under Valgrind.

Among other things, unit-testing validates error handling for all Sysman functions used at run-time.  See "gpu_sysman_test.c" for more information.

Different versions of code have been sporadically tested on small set of Intel iGPU and discrete (Xe) GPUs development machines since 2020, but there has been no testing on production servers.

Testing has been done using Intel compute-runtime Sysman GPU backend (mostly with internal GPU SW stack supporting Xe HW).  In principle plugin should work also with other Sysman API GPU backend than Intel one, but such were not available for testing.

PS. Intel compute-runtime provides standalone Level Zero test program which can be used to check/validate metrics values:
https://github.com/intel/compute-runtime/blob/master/level_zero/tools/test/black_box_tests/zello_sysman.cpp

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1105,6 +1105,21 @@ gpu_nvidia_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(BUILD_WITH_CUDA_LDFLAGS)
 gpu_nvidia_la_LIBADD = $(BUILD_WITH_CUDA_LIBS)
 endif
 
+if BUILD_PLUGIN_GPU_SYSMAN
+pkglib_LTLIBRARIES += gpu_sysman.la
+gpu_sysman_la_SOURCES = src/gpu_sysman.c
+gpu_sysman_la_CPPFLAGS = $(AM_CPPFLAGS) $(PLUGIN_CPPFLAGS) $(BUILD_WITH_SYSMAN_CPPFLAGS)
+gpu_sysman_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(BUILD_WITH_SYSMAN_LDFLAGS)
+gpu_sysman_la_LIBADD = $(BUILD_WITH_SYSMAN_LIBS)
+
+test_plugin_gpu_sysman_SOURCES = src/gpu_sysman_test.c
+test_plugin_gpu_sysman_CPPFLAGS = $(AM_CPPFLAGS) $(PLUGIN_CPPFLAGS) $(BUILD_WITH_SYSMAN_CPPFLAGS)
+test_plugin_gpu_sysman_LDFLAGS = $(PLUGIN_LDFLAGS)
+test_plugin_gpu_sysman_LDADD =
+check_PROGRAMS += test_plugin_gpu_sysman
+TESTS += test_plugin_gpu_sysman
+endif
+
 if BUILD_PLUGIN_GRPC
 pkglib_LTLIBRARIES += grpc.la
 grpc_la_SOURCES = src/grpc.cc
--- a/README
+++ b/README
@@ -161,6 +161,9 @@ Features
     - gpu_nvidia
       Monitor NVIDIA GPU statistics available through NVML.
 
+    - gpu_sysman
+      Monitor GPU statistics available through Level-Zero Sysman API.
+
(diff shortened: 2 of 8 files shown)


eero-t:
Somebody should update project `.clang-format` file.  Running `./contrib/format.sh` after running `clan-format` locally, changed the code (changed indenting for few '//' comments that were after defines).

mrunge:
thank you for your PR!

You may have seen the CI system not reporting back/failing.
Also clang format has caused frustration before.

Unfortunately, I can not do anything about this, but maybe @octo would have the time to deactivate the travis checks so that we can move/fix CI.

eero-t:
Now that the project is moving again, can new plugins be still get integrated to future 5.x versions of collectd?

(I would like to post-pone porting of the plugin to v6 API until there's first pre-release of v6, and some schedule for releasing v6.0.)

eero-t:
To trigger fixed CI checks, I rebased my branch to latest main.  There are no changes to the commits themselves.

Main feedback I'm interested about, is for the new metrics type, and metrics naming.  See the description above.

eero-t:
@mrunge Any feedback / objections on merging this new GPU plugin?

It's disabled by default and as stated above, it adds "error_count" GAUGE type to "types.db".

---

Sysman plugin is built against Level Zero (L0) frontend API.  If there's no L0 backend, or GPU recognized by it at run-time, plugin initialization fails.  

Level Zero support is not yet in upstream distributions, so to test the plugin data output, one needs to build both from sources for now:
* L0 frontend (simple): https://github.com/oneapi-src/level-zero
* L0 backend (more involved): https://github.com/intel/compute-runtime/

All the metrics data is not available for older / integrated Intel GPUs, only for discrete ones.  Public driver stack does not yet support public kernel API for latter, but I have a bug about that: https://github.com/intel/compute-runtime/issues/453

eero-t:
Closing, superseded by: https://github.com/collectd/collectd/pull/3968

PR #3884, opened 2021-06-16
nginx plugin: Add support for UNIX domain sockets

basilfx:
Add support for UNIX domain sockets for NGINX servers that listen on a socket.

With this PR, one can serve the `stub_status` module on a socket, which can be more secure and faster than exposing it on a normal port. For example, I use the following NGINX configuration:

```
server {
  listen unix:/var/run/nginx.sock;

  location /stub_status {
    stub_status;
  }
}
````

ChangeLog: nginx plugin: Add support for UNIX domain sockets

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2322,6 +2322,12 @@ if test "x$with_libcurl" = "xyes"; then
       [have_curlopt_timeout="no"],
       [[#include <curl/curl.h>]]
     )
+
+    AC_CHECK_DECL(CURLOPT_UNIX_SOCKET_PATH,
+      [have_curlopt_unix_socket_path="yes"],
+      [have_curlopt_unix_socket_path="no"],
+      [[#include <curl/curl.h>]]
+    )
   fi
 fi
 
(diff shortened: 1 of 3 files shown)


mrunge:
Thank you for your PR!

You may have seen CI is defective currently, and unfortunately I can not do anything about it.

On your PR: what happens, if I provide the socket in addition to the traditional config? ... which setting takes precedence? Would it make sense to print out a warning during startup to give a hint? Otherwise a user may end up with unexpected behavior?

basilfx:
The setting is additional, and is only used if given. All other options keep working as-is. It goes alongside with the URL: https://curl.se/libcurl/c/CURLOPT_UNIX_SOCKET_PATH.html

So I don't think there is unexpected behavior to be expected.

mrunge:
Ah, thank you. It's great to confirm that.

mrunge (approved):

Unfortunately, merging is blocked due to CI failures (not related to this PR)

mrunge:
CI issues have been resolved. I can now finally merge this. Sorry for keep you waiting and thank you very much for your PR.

PR #3893, opened 2021-07-19
Fix debug print of excluderegex if NULL

sonertari:
ChangeLog: ExcludeRegex in logparser plugin is optional, hence can be NULL. This change fixes debug logging causing CRITICAL error for NULL excluderegex, if debug is enabled.

### DIFF
--- a/src/utils/match/match.c
+++ b/src/utils/match/match.c
@@ -204,7 +204,7 @@ match_create_callback(const char *regex, const char *excluderegex,
   int status;
 
   DEBUG("utils_match: match_create_callback: regex = %s, excluderegex = %s",
-        regex, excluderegex);
+        regex, excluderegex != NULL ? excluderegex : "");
 
   obj = calloc(1, sizeof(*obj));
   if (obj == NULL)


src/utils/match/match.c
```diff
@@ -204,7 +204,7 @@ match_create_callback(const char *regex, const char *excluderegex,
   int status;
 
   DEBUG("utils_match: match_create_callback: regex = %s, excluderegex = %s",
-        regex, excluderegex);
+        regex, excluderegex != NULL ? excluderegex : "");
```
kwiatrox, line 207: Personally I would change it to something more meaningful than empty string, which is not visible for the user, like:
`excluderegex != NULL ? excluderegex : "NULL"` or `excluderegex != NULL ? excluderegex : "''"`.

sonertari: I see that in similar cases the collectd source code prints the following:

- `"(null)"` in src/configfile.c
- `"<NULL>"` in src/utils_curl_stats.c
- `"(nil)"` in src/netlink.c
- `""` in src/nginx.c (as in my pr)

Your suggestion would be fine with me, but it will add yet another variation to the list. As you (the collectd contributors) wish.

kwiatrox: So we can choose existing one to not add another variation, I would choose between `"(null)"` or `"<NULL>"`.

kwiatrox (approved):

PR #3901, opened 2021-08-20
Fix-configure.ac: typing error for the CPPFLAGS of libredfish

matsKNY:
Changelog: typing error in configure.ac for the CPPFLAGS of libredfish.

Probably a copy/paste error: `LIBHIREDFISH_CPPFLAGS` was used instead of
`LIBREDFISH_CPPFLAGS`.


kwiatrox (approved):
Good catch. Thank you for your contribution, unfortunately it can not be merged now due to CI issue with el6. The issue is known and not related to this change. The change itself looks good.

matsKNY:
Thank you for your review and your work on collectd.
The mergeing of this PR does not seem to be critical, so there is no problem for it to be done when the CI will be fixed.

mrunge (approved):

Issue #3894, opened 2021-07-20
Small error due the unreleased lock before the exit


jenny-cheung:
Hi, developers, before normal exit, the mutex is unlocked. However, in the method `init_pi`, the lock `perl_threads->mutex` may still be held before the exit.It is not a very harmful bug( e.g., resource leak, deadlock) since the program exits and all of the program resources will be cleaned up by some common OS. However, other OS systems (e.g., embedded systems) do not automatically free some resources at the exit. It is a good manner for resource management. Adding the unlock statement adds symmetry, so the code looks better. Also, the debugger would not warn this case : )

https://github.com/collectd/collectd/blob/1da3305c10c8ff9a63081284cf3d4bb0f6daffd8/src/perl.c#L2446

```
static int init_pi(int argc, char **argv) {
  ...;
  pthread_mutex_lock(&perl_threads->mutex);

  perl_threads->head = c_ithread_create(NULL);
  perl_threads->tail = perl_threads->head;

  if (NULL == (perl_threads->head->interp = perl_alloc())) {
    log_err("init_pi: Not enough memory.");
    exit(3);
    }
   ...;
  }

}
```

Best,

jenny-cheung:
Hi, developers @octo @mrunge 

Any comments would be highly appreciated. Thank you.

Best Regards,

mrunge:
We appreciate pull requests.

PR #3903, opened 2021-08-23
Fix a small error due to the unreleased lock for 

jenny-cheung:
ChangeLog: Perl plugin: add unlock pthread mutex before exit.

Fix a small error due to the unreleased lock by add pthread_mutex_unlock(&perl_threads->mutex) before program exit. #3894


mrunge (approved):
Thank you, good catch!

PR #3905, opened 2021-08-23
Replace travis with githubaction

elfiesmelfie:
Repeats the steps that travis was using for building collectd

 * checks out branch
 * installs dependencies
 * runs the same script commands as travis ( pkg-config, confgure, make)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,13 +28,17 @@ jobs:
           # RedHat family
           - el7_x86_64
           - fedora28_x86_64
+        config_flags: ['']
         include:
           - container_tag: sid_amd64
             allow_failures: true
+            config_flags: "--disable-dpdkstat --disable-dpdkevents --disable-virt"
           - container_tag: fedora_rawhide_x86_64
             allow_failures: true
+            config_files: ""
     env:
       MAKEFLAGS: "-j 2"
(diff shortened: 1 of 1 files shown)


mrunge:
Thank you Emma. I'm still investigating where the travis-ci job is hooked into, ... where is it waiting on the job to report.

mrunge (dismissed):
Thank you, this looks good to me. Let's see if this also triggers running the action

elfiesmelfie:
@mrunge In order to trigger the githubactions (GHA), there may need to be a workflow merged into main first, which will enable GHA. I can either propose a very basic GHA, which will trigger on push and PRs, so we can see these ones working, or we can merge this and see what happens (which I don't agree with, because the contents may be up for debate).

mrunge:
@elfiesmelfie right, and thank you for getting this back to my attention. Enabling a basic action before this being really tested is probably the way to go here.

elfiesmelfie:
The remote I was using on github (collectd-mirror/collectd) was deleted, so I can't update this PR.

I have pushed a new update to elfiesmelfie/collectd, and opened a PR here: https://github.com/collectd/collectd/pull/3913

This new one has almost the same content - commits were squashed, and the replacement for jobs in .cirrus.yml was removed.

mrunge:
Tokay, thank you. I'm going to close this one here.

PR #3907, opened 2021-08-26
Feature-redfish: enabling libredfish logs when collectd debug mode is enforced

matsKNY:
Changelog: enabling libredfish logs when collectd debug mode is enforced

In order for `libredfish` to log internal information, which are more
than useful to debugging purposes, a printing function has to be
registered through a hook : a global function pointer named `gDebugFunc`.

The `plugin_log` function is registered as the printing function through
the aforementioned hook by the `redfish_init` function, when the
`--enable-debug` option is specified to the `configure` script.

### DIFF
--- a/src/redfish.c
+++ b/src/redfish.c
@@ -113,6 +113,10 @@ static int redfish_validate_config(void);
 static void *redfish_worker_thread(void __attribute__((unused)) * args);
 
 #if COLLECT_DEBUG
+/* Hook exposed by the libredfish library to define a printing function
+ * dedicated to logging purposes: */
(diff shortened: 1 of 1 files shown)


mrunge:
I really wonder why this still gets tested by centos 6, that has been deprecated and removed some time ago....

systemcrash:
Does CI rely on a specific branch in the repo?

mrunge (approved):

mrunge:
Thank you, looks good to me.

Re. CI, I believe this branch used a parent which still had the centos-6 test included.
All other tests passed here, thus I'm merging this.

PR #3913, opened 2021-09-10
Replace travis CI with GHA

elfiesmelfie:
ChangeLog: Add Github actions to replace travis CI

Repeats the steps that travis was using for building collectd

-     checks out branch
-     installs dependencies (either they are already in the container or they are explicitly installed)
-     runs the same script commands as travis ( pkg-config, confgure, make)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,18 +31,163 @@ jobs:
           - el7_x86_64
           - fedora28_x86_64
         config_flags: ['']
-        include:
-          - container_tag: sid_amd64
-            allow_failures: true
-            config_flags: "--disable-dpdkstat --disable-dpdkevents --disable-virt"
-          - container_tag: fedora_rawhide_x86_64
-            allow_failures: true
-            config_flags: ""
     env:
       MAKEFLAGS: "-j 2"
       CONFIGURE_FLAGS: ${{ matrix.config_flags }}
     steps:
     - uses: actions/checkout@v2
+    - run: type pkg-config
+    - run: pkg-config --list-all | sort -u
+    - name: Generate configure script
+      run:
+        ./build.sh
+    - name: configure
+      run: ./configure $CONFIGURE_FLAGS
+    - name: Make
+      run: make
+    - name: make check
+      run: make check
+    - name: install bzip2
+      continue-on-error: true
+      run: |
+        yum install -y bzip2 || apt install -y bzip2
+    - name: make
+      continue-on-error: true
+      run: |
+        make distcheck DISTCHECK_CONFIGURE_FLAGS="--disable-dependency-tracking --enable-debug"
+
+  experimental-debian:
+    runs-on: ubuntu-20.04
+    container: collectd/ci:${{ matrix.container_tag }}
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        # for tasks that are optional, use the continue-on-error option, to prevent a workflow from failing when the task fails
+        container_tag:
+          - sid_amd64
+        config_flags: ["--disable-dpdkstat --disable-dpdkevents --disable-virt"]
+    env:
+      MAKEFLAGS: "-j 2"
+      CONFIGURE_FLAGS: ${{ matrix.config_flags }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -6,21 +6,205 @@ on:
...
+        librdkafka-devel \
+        libsigrok-devel \
+        libudev-devel \
+        libvirt-devel \
+        libxml2-devel \
+        lm_sensors-devel \
+        lua-devel \
+        lvm2-devel \
+        mosquitto-devel \
+        mysql-devel \
```
mrunge, line 176: mariadb-devel, mysql-devel does not exist in Fedora 34+

mrunge: Not sure if that actually causes the node error, but it has to be fixed anyways.

elfiesmelfie: It is undetermined what is causing the error on fedora:rawhide, since the packages were not installed, due to being unable to resolve some link.

On both Fedora 34 and rawhide, ``dnf install mysql-devel`` will install ``mariadb-devel``, so this is most likely not causing the error. However, I have updated this dependency, because it is misleading anyone who casually looks at this.

mrunge:
The rawhide job fails with
```
Error: Error downloading packages:
  Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=rawhide&arch=x86_64 [getaddrinfo() thread failed to start]
Error: Process completed with exit code 1.
```

elfiesmelfie:
> The rawhide job fails with
> 
> ```
> Error: Error downloading packages:
>   Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=rawhide&arch=x86_64 [getaddrinfo() thread failed to start]
> Error: Process completed with exit code 1.
> ```

That might be due to:
https://www.mail-archive.com/devel@lists.fedoraproject.org/msg169919.html

.github/workflows/build.yml
```diff
@@ -6,21 +6,205 @@ on:
...
+    runs-on: ubuntu-20.04
+    container: ${{ matrix.container_image }}
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # for tasks that are optional, use the continue-on-error option, to prevent a workflow from failing when the task fails
+        container_image:
+          - fedora:34
+          - fedora:rawhide
```
mrunge, line 105: can we please skip fedora:rawhide for now? I've just added fedora:34 to the container registry.

mrunge, line 105:
```suggestion
```

elfiesmelfie: I'll add this in.

There are a few changes I'm going to make, since I don't have to add in the dependency install steps anymore:

* Move fedora34 into the ``build`` job
* Combine the ``experimental-debian`` and ``experimental-fedora`` jobs into an ``experimental`` job
* Fedora rawhide will be in the ``experimental`` job, once it works again.
* Add in the new containers from (collectd/ci-docker) to the build jobs

mrunge:
We may want to make the test of make check non-voting until it is working reliable?

elfiesmelfie:
> We may want to make the test of make check non-voting until it is working reliable?

I'm starting to realise I need to update the PR jobs on collectd/ci-docker not to do the make check either

I can add a single line here to still run make check, but also make it not cause the job to fail.

.github/workflows/build.yml
```diff
@@ -6,21 +6,102 @@ on:
...
+    - run: pkg-config --list-all | sort -u
+    - name: Generate configure script
+      run:
+        ./build.sh
+    - name: configure
+      run: ./configure $CONFIGURE_FLAGS
+    - name: Make
       run: make
     - name: make check
       run: make check
```
elfiesmelfie, line 98:
```suggestion
      # Make check is failing on a few newer distros, temporarily mark it as optional until that is resolved 
      continue-on-error: true
      run: make check
```

mrunge (approved):
Thank you, this looks good to me

PR #3914, opened 2021-09-11
add CounterGauge option in statsd

dctucker:
This PR introduces the `CounterGauge` option for the `statsd` plugin.

This enables compatibility with the reference StatsD implementation, aggregating the increase during the flushing interval, reporting the current value as a differential, and resetting the counter.

Originally this was written by @vmg as a patch against version 5.5.3 to make this the standard behavior. I've adapted it so this behavior is optional and does not interfere with reporting `derive` metrics, similarly to what I saw done with the Etsy option `CounterSum`. We found use for this while performing a Debian dist-upgrade from stretch to buster where we're using a patched collectd version 5.10.0.

ChangeLog: statsd plugin: The CounterGauge option has been added. Thanks to Casey Tucker and Vicent Marti.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1714,6 +1714,7 @@
 #  DeleteGauges   false
 #  DeleteSets     false
 #  CounterSum     false
+#  CounterGauge   false
 #  TimerPercentile 90.0
 #  TimerPercentile 95.0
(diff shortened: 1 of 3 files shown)


mrunge (approved):
Thank you for your patch, looks good to me.

PR #3917, opened 2021-09-29
Fix some gcc warnings with more strict checks

eero-t:
ChangeLog: Build: Fix some gcc warnings with more strict checks

Fixes are based on "-O3 -Werror -Wall -Wextra -Wformat-security" output:
* Properly initialize complex struct
* Tell compiler which args are expected to be unused with attributes
* Correct write_prometheus comments and use arg more like specified in MHD doc

Somebody may want to do similar thing for rest of collectd code, but I was interested only about core & write_prometheus plugin.

This is collected "main" (v5) branch version of (already merged) v6 PR: https://github.com/collectd/collectd/pull/3970

EDIT: added link to collectd v6 PR + dropped resolved comment about stuff discussed below.

### DIFF
--- a/src/daemon/utils_cache.c
+++ b/src/daemon/utils_cache.c
@@ -670,7 +670,7 @@ int uc_get_names(char ***ret_names, cdtime_t **ret_times, size_t *ret_number) {
   return 0;
 } /* int uc_get_names */
 
-int uc_get_state(const data_set_t *ds, const value_list_t *vl) {
+int uc_get_state(const value_list_t *vl) {
   char name[6 * DATA_MAX_NAME_LEN];
   cache_entry_t *ce = NULL;
   int ret = STATE_ERROR;
@@ -692,7 +692,7 @@ int uc_get_state(const data_set_t *ds, const value_list_t *vl) {
   return ret;
 } /* int uc_get_state */
 
-int uc_set_state(const data_set_t *ds, const value_list_t *vl, int state) {
+int uc_set_state(const value_list_t *vl, int state) {
   char name[6 * DATA_MAX_NAME_LEN];
   cache_entry_t *ce = NULL;
   int ret = -1;
@@ -775,19 +775,7 @@ int uc_get_history_by_name(const char *name, gauge_t *ret_history,
   return 0;
 } /* int uc_get_history_by_name */
 
-int uc_get_history(const data_set_t *ds, const value_list_t *vl,
-                   gauge_t *ret_history, size_t num_steps, size_t num_ds) {
-  char name[6 * DATA_MAX_NAME_LEN];
-
-  if (FORMAT_VL(name, sizeof(name), vl) != 0) {
-    ERROR("utils_cache: uc_get_history: FORMAT_VL failed.");
-    return -1;
-  }
-
-  return uc_get_history_by_name(name, ret_history, num_steps, num_ds);
-} /* int uc_get_history */
-
-int uc_get_hits(const data_set_t *ds, const value_list_t *vl) {
+int uc_get_hits(const value_list_t *vl) {
   char name[6 * DATA_MAX_NAME_LEN];
   cache_entry_t *ce = NULL;
   int ret = STATE_ERROR;
@@ -809,7 +797,7 @@ int uc_get_hits(const data_set_t *ds, const value_list_t *vl) {
   return ret;
 } /* int uc_get_hits */
 
-int uc_set_hits(const data_set_t *ds, const value_list_t *vl, int hits) {
+int uc_set_hits(const value_list_t *vl, int hits) {
   char name[6 * DATA_MAX_NAME_LEN];
   cache_entry_t *ce = NULL;
   int ret = -1;
@@ -832,7 +820,7 @@ int uc_set_hits(const data_set_t *ds, const value_list_t *vl, int hits) {
   return ret;
 } /* int uc_set_hits */
 
-int uc_inc_hits(const data_set_t *ds, const value_list_t *vl, int step) {
+int uc_inc_hits(const value_list_t *vl, int step) {
   char name[6 * DATA_MAX_NAME_LEN];
   cache_entry_t *ce = NULL;
   int ret = -1;
--- a/src/infiniband.c
+++ b/src/infiniband.c
@@ -118,7 +118,7 @@ static int ib_read_value_file_num_only(const char *device, const char *port,
   strstripnewline(buffer);
 
   // zero-out the first non-digit character
-  for (int i = 0; i < sizeof(buffer); i++) {
+  for (size_t i = 0; i < sizeof(buffer); i++) {
     if (!isdigit(buffer[i])) {
       buffer[i] = '\0';
       break;
@@ -327,7 +327,7 @@ static int infiniband_read(void) {
   char port_name[255];
 
   if (ib_glob_ports(&g) == 0) {
-    for (int i = 0; i < g.gl_pathc; ++i) {
+    for (unsigned i = 0; i < g.gl_pathc; ++i) {
       char *device = NULL, *port = NULL;
       if (ib_parse_glob_port(g.gl_pathv[i], &device, &port) == 0) {
         snprintf(port_name, sizeof(port_name), "%s:%s", device, port);
--- a/src/cgroups.c
+++ b/src/cgroups.c
@@ -52,7 +52,7 @@ cgroups_submit_one(char const *plugin_instance, char const *type_instance,
  * This callback reads the user/system CPU time for each cgroup.
  */
 static int read_cpuacct_procs(const char *dirname, char const *cgroup_name,
-                              void *user_data) {
+                              __attribute__((unused)) void *user_data) {
   char abs_path[PATH_MAX];
   struct stat statbuf;
   char buf[1024];
@@ -131,7 +131,7 @@ static int read_cpuacct_procs(const char *dirname, char const *cgroup_name,
  * read_cpuacct_procs callback on every folder it finds, such as "system".
  */
 static int read_cpuacct_root(const char *dirname, const char *filename,
-                             void *user_data) {
+                             __attribute__((unused)) void *user_data) {
   char abs_path[PATH_MAX];
   struct stat statbuf;
   int status;
--- a/src/conntrack.c
+++ b/src/conntrack.c
@@ -43,7 +43,8 @@ static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
 
 static int old_files;
 
-static int conntrack_config(const char *key, const char *value) {
+static int conntrack_config(const char *key,
+                            __attribute__((unused)) const char *value) {
   if (strcmp(key, "OldFiles") == 0)
     old_files = 1;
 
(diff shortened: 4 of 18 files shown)


systemcrash:
I'm a bit wary of committing `__attribute__((unused))` to code statically to satisfy one particular compiler. So I think a macro for this purpose would be far better: I have low expectations that these static changes would also satisfy a range of other compilers. Esp since you've done this for only two plugins. Otherwise the intention is welcome.

eero-t:
> I'm a bit wary of committing __attribute__((unused)) to code statically to satisfy one particular compiler. So I think a macro for this purpose would be far better: I have low expectations that these static changes would also satisfy a range of other compilers. Esp since you've done this for only two plugins. Otherwise the intention is welcome.

@systemcrash after looking into this, `git grep __attribute__` shows following attributes to be already used in collectd code:
* aligned
* force
* format
* malloc
* nonnull
* packed
* sentinel
* unused
* warn_unused_result

And of those, "unused" attribute seems to be already in use all over collectd code before my changes:
```
$ git grep -l '__attribute__((unused))' | wc -l
78
```
I.e. in more than every fifth C file in collectd:
```
$ git ls-files | grep '\.[ch]$' | wc -l
352
```

=> Your worry seems unwarranted

There are macro guards already, actually in several places:
```
$ git grep -C1 '# *define *__attribute'
src/collectd-nagios.c-#if !defined(__GNUC__) || !__GNUC__
src/collectd-nagios.c:#define __attribute__(x) /**/
src/collectd-nagios.c-#endif
--
src/collectd-tg.c-#if !__GNUC__
src/collectd-tg.c:#define __attribute__(x) /**/
src/collectd-tg.c-#endif
--
src/collectdmon.c-#if !defined(__GNUC__) || !__GNUC__
src/collectdmon.c:#define __attribute__(x) /**/
src/collectdmon.c-#endif
--
src/daemon/collectd.h-#if !defined(__clang__) && !defined(__GNUC__)
src/daemon/collectd.h:#define __attribute__(x) /**/
src/daemon/collectd.h-#endif
--
src/libcollectdclient/client.c-#if !defined(__GNUC__) || !__GNUC__
src/libcollectdclient/client.c:#define __attribute__(x) /**/
src/libcollectdclient/client.c-#endif
--
src/libcollectdclient/network_parse.c-#if !defined(__GNUC__) || !__GNUC__
src/libcollectdclient/network_parse.c:#define __attribute__(x) /**/
src/libcollectdclient/network_parse.c-#endif
--
src/libcollectdclient/server.c-#if !defined(__GNUC__) || !__GNUC__
src/libcollectdclient/server.c:#define __attribute__(x) /**/
src/libcollectdclient/server.c-#endif
```
=> These cover the files that I'm modifying so there's nothing extra to do in my PR.

---

Btw. As you can see from above, there are small differences in how the checks are done.  If all of those C files would include "collectd.h" instead of "config.h" / "globals.h" ("collectd.h" includes those other ones), then the slightly differing duplicate ifdefs could be removed.  Such fix would belong to some other PR though.

"unused" attribute is apparently not supported by MSVC:
https://stackoverflow.com/questions/52058457/visual-studio-equivelent-of-gcc-attribute-unused-in-c11-or-lower

But it is supported by LLVM/Clang, along with the other GCC attributes I think:
https://clang.llvm.org/docs/AttributeReference.html#maybe-unused-unused

Which is another reason (for somebody else) to consolidate the ifdefs e.g. to collectd.h....

eero-t:
Rebased to latest "main" and added compiler warning fixes also to some other plugins in addition to "write_prometheus".

eero-t:
I added separate bug for larger number of issues in some other plugins, as fixing them could mess up the existing collectd v6 PRs for the plugins.  I'll see which of the current commits apply to v6 branch as such, and add separate PR for that, maybe also fix few other issues from that.

eero-t:
Gah, I had accidentally rebased this on local "main", which was behind upstream "main".  As this PR does (minor) changes to many components, I rebased again to make sure checks get run also with bionic & focal.

eero-t:
* Rebased to main to get CI fix +  improved commit descriptions and write_prometheus plugin patch slightly (better comment, no var name change).

eero-t:
Only failure:
> debian_default_toolchain container:collectd/ci:xenial_amd64 Failing after 15m
> Failed to start an instance! CreateContainerError: context deadline exceeded

eero-t:
> Only failure:
> > debian_default_toolchain container:collectd/ci:xenial_amd64 Failing after 15m
> > Failed to start an instance! CreateContainerError: context deadline exceeded

@elfiesmelfie What's the preferred way to re-trigger tests?

eero-t:
Used "Fix" label because some of the warnings were actually visible to users as build errors.

src/daemon/types_list.c
```diff
@@ -104,7 +104,7 @@ static void parse_line(char *buf) {
   if (fields[0][0] == '#')
     return;
 
-  data_set_t ds = {{0}};
+  data_set_t ds = {{0}, 0, NULL};
```
octo, line 107: Why is this needed? When a struct is only partially initialized, the remainder is initialized to zero according to the C standard.

src/daemon/utils_cache.c
```diff
@@ -670,7 +670,8 @@ int uc_get_names(char ***ret_names, cdtime_t **ret_times, size_t *ret_number) {
   return 0;
 } /* int uc_get_names */
 
-int uc_get_state(const data_set_t *ds, const value_list_t *vl) {
```
octo, line 673: We could just remove the `ds` argument. There are only three call sites.

src/daemon/utils_cache.c
```diff
@@ -692,7 +693,8 @@ int uc_get_state(const data_set_t *ds, const value_list_t *vl) {
   return ret;
 } /* int uc_get_state */
 
-int uc_set_state(const data_set_t *ds, const value_list_t *vl, int state) {
+int uc_set_state(__attribute__((unused)) const data_set_t *ds,
```
octo, line 696: We could just remove the `ds` argument. There is only one call site.

src/daemon/utils_cache.c
```diff
@@ -775,8 +777,9 @@ int uc_get_history_by_name(const char *name, gauge_t *ret_history,
   return 0;
 } /* int uc_get_history_by_name */
 
-int uc_get_history(const data_set_t *ds, const value_list_t *vl,
-                   gauge_t *ret_history, size_t num_steps, size_t num_ds) {
+int uc_get_history(__attribute__((unused)) const data_set_t *ds,
```
octo, line 780: This function appears to be unused. If so, we should fix it by deleting it.

src/daemon/utils_cache.c
```diff
@@ -787,7 +790,8 @@ int uc_get_history(const data_set_t *ds, const value_list_t *vl,
   return uc_get_history_by_name(name, ret_history, num_steps, num_ds);
 } /* int uc_get_history */
 
-int uc_get_hits(const data_set_t *ds, const value_list_t *vl) {
+int uc_get_hits(__attribute__((unused)) const data_set_t *ds,
```
octo, line 793: We could just remove the `ds` argument. There are only two call sites.

src/daemon/utils_cache.c
```diff
@@ -809,7 +813,8 @@ int uc_get_hits(const data_set_t *ds, const value_list_t *vl) {
   return ret;
 } /* int uc_get_hits */
 
-int uc_set_hits(const data_set_t *ds, const value_list_t *vl, int hits) {
+int uc_set_hits(__attribute__((unused)) const data_set_t *ds,
```
octo, line 816: We could just remove the `ds` argument. There is only one call site.

src/daemon/utils_cache.c
```diff
@@ -832,7 +837,8 @@ int uc_set_hits(const data_set_t *ds, const value_list_t *vl, int hits) {
   return ret;
 } /* int uc_set_hits */
 
-int uc_inc_hits(const data_set_t *ds, const value_list_t *vl, int step) {
+int uc_inc_hits(__attribute__((unused)) const data_set_t *ds,
```
octo, line 840: We could just remove the `ds` argument. There is only one call site.

src/infiniband.c
```diff
@@ -118,7 +118,7 @@ static int ib_read_value_file_num_only(const char *device, const char *port,
   strstripnewline(buffer);
 
   // zero-out the first non-digit character
-  for (int i = 0; i < sizeof(buffer); i++) {
+  for (unsigned i = 0; i < sizeof(buffer); i++) {
```
octo, line 121: Prefer `size_t` when comparing with `sizeof()`.

octo (commented):
Thanks for your effort @eero-t! Overall this looks good. I think we should remove unused arguments form exported functions – keeping their argument list clean is possibly more important than for `static` functions.

octo:
When convenient, please rebase onto a recent `main` to pick up the CI fixes.

src/daemon/types_list.c
eero-t: `-Wextra` implies `-Wmissing-field-initializers` => I'll explicitly disable that from the tests, and drop this change.

src/daemon/utils_cache.c
eero-t: done

src/daemon/utils_cache.c
eero-t: done

src/daemon/utils_cache.c
eero-t: done

src/daemon/utils_cache.c
eero-t: done

eero-t:
Did the suggested changes.

Need still to check that GCC does not report new errors in the modified files (with additional -Werror warnings).

Then the updates to the original changes need to be ported to v6, as original ones were applied there as-is.

PS. These are just compile checked.  It's nearly 2 years since I last used v5 (instead of v6), and even before that I've never used the plugins modified here, except for Prometheus one.  Changes should be straight-forward enough though, that if they compile, they're fine.

eero-t:
Checking the modified files (after running configure so "config.h" + other generated headers are present):
```
$ for i in cgroups.c conntrack.c infiniband.c logparser.c statsd.c table.c turbostat.c \
  threshold.c write_riemann_threshold.c write_prometheus.c write_syslog.c write_tsdb.c \
  daemon/utils_cache.c libcollectdclient/network_parse.c utils/format_json/format_json.c; do
    echo "*** $i ***";
    gcc -I.. -I. -Idaemon -Ilibcollectdclient -DHAVE_CONFIG_H -O3 -Werror -Wall -Wextra -Wpedantic \
    -Wcast-align=strict -Wformat-security -Wnull-dereference -Wstrict-overflow=2 -Warray-bounds=2 \
    -D_FORTIFY_SOURCE=2 -Wno-aggressive-loop-optimizations -c -o /dev/null $i;
done
```

Showed couple of extra warnings in 2 modified files, for which I added separate fixes.

eero-t:
Added same cast correction to other files as what's in `write_prometheus` (similarly to #4243 one for v6).

octo:
> @octo Have you had any thoughts on consolidating the `__attribute__` handling ifdefs, discussed in [#3917 (comment)](https://github.com/collectd/collectd/pull/3917#issuecomment-936749909)?

You mean unifying how this is treated across binaries? I don't think that it's worth the effort switch that so that each binary inclused `"collectd.h"`, but if such a PR were to materialize I'd probably merge it. So mostly 🤷 ;)

octo (approved):
Great work, thanks @eero-t!

PR #3921, opened 2021-10-04
Switch to bullseye based testing

mrunge:
stretch has been EOLed in mid 2020.

ChangeLog: Update cirrus CI to use bullseye for non-standard toolchain testing.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -99,7 +99,7 @@ redhat_default_toolchain_task:
 # Some are expected to fail, others should always pass
 non_standard_toolchains_task:
   container:
-    image: collectd/ci:stretch_amd64
+    image: collectd/ci:bullseye_amd64
   only_if: $CIRRUS_PR == ''
 
   matrix:


elfiesmelfie:
I think I know why this is failing. 

It's because of the virt unit tests.
I marked the ``make check`` step as optional for the github actions. For debian sid (i.e. unstable), the ``--disable-virt`` configure flag is set, so it doesn't try to test virt (sid fails for other reasons though, but we don't expect that it'll always pass).
For bullseye and, the virt plugin is not disabled, and ``make check`` actually fails (but is hidden because that step is marked as ``continue-on-error: true``.

In order to use the ``bullseye_amd64`` container here, you need to pass the ``--disable-virt`` flag to the ./configure step in cirrus.

.cirrus.yml
```diff
@@ -27,10 +27,10 @@ release_ready_task:
...
     - container:
-        image: collectd/ci:buster_amd64
+        image: collectd/ci:bullseye_amd64
       allow_failures: false
     - container:
-        image: collectd/ci:stretch_amd64
+        image: collectd/ci:buster_amd64
       allow_failures: false
     - container:
         image: collectd/ci:stretch_i386
```
elfiesmelfie, line 36: You missed this one

elfiesmelfie (changes requested):

PR #3928, opened 2021-10-08
[collectd-6.0] ethstat plugin

bkotlowski:
ChangeLog: ethstat: ethstat plugin expose ethtool and NIC statistics from sysfs


kwiatrox (approved):
Reviewed and for me it looks good. It was tested and also should not interfere with existing functionality.

mrunge:
Thank you!

PR #3935, opened 2021-10-25
snmp plugin: Add support for SHA224, SHA256, SHA384 and SHA512

minfrin:
ChangeLog: snmp plugin: Add support for SHA224, SHA256, SHA384 and SHA512.

### DIFF
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -674,22 +674,97 @@ static int csnmp_config_add_host_auth_protocol(host_definition_t *hd,
   if (status != 0)
     return status;
 
+#ifdef NETSNMP_USMAUTH_HMACMD5
   if (strcasecmp("MD5", buffer) == 0) {
     hd->auth_protocol = usmHMACMD5AuthProtocol;
     hd->auth_protocol_len = sizeof(usmHMACMD5AuthProtocol) / sizeof(oid);
-  } else if (strcasecmp("SHA", buffer) == 0) {
+
+    DEBUG("snmp plugin: host = %s; host->auth_protocol = %s;", hd->name, "MD5");
+    return 0;
+  }
+#endif
+
+#ifdef NETSNMP_USMAUTH_HMACSHA1
+  if (strcasecmp("SHA", buffer) == 0) {
     hd->auth_protocol = usmHMACSHA1AuthProtocol;
     hd->auth_protocol_len = sizeof(usmHMACSHA1AuthProtocol) / sizeof(oid);
-  } else {
-    WARNING("snmp plugin: The `AuthProtocol' config option must be `MD5' or "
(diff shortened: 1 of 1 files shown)


minfrin:
> What happens when you if (strcasecmp("SHA224", buffer) == 0) and char buffer[4];? 🤔

Bad things - good catch, I've just fixed that.

> Also, does strcasecmp behave as expected with two strings, where one will potentially always match?
I'm thinking:
if (strcasecmp("SHA", buffer) == 0) {
and
if (strcasecmp("SHA224", buffer) == 0) {

strcasecmp works on whole strings rather than string prefixes, so the order doesn't matter in this case.

> Is it possible to IFDEF multiple HMACs? In which case, I propose not multiple IFDEF but chaining with ELSEIF.

Defining multiple HMACs is typical yes, you'll typically get SHA224, SHA256, SHA384 and SHA512 all at the same time.

> Does status get coerced to int? Some earlier definitions define int, but some don't. Probably safer to define.

This patch doesn't modify the behaviour of status, but can submit a separate PR for it? Good opportunity for a cleanup if you think it's needed.

systemcrash:
> > What happens when you if (strcasecmp("SHA224", buffer) == 0) and char buffer[4];? thinking
> 
> Bad things - good catch, I've just fixed that.
> 
> > Also, does strcasecmp behave as expected with two strings, where one will potentially always match?
> > I'm thinking:
> > if (strcasecmp("SHA", buffer) == 0) {
> > and
> > if (strcasecmp("SHA224", buffer) == 0) {
> 
> strcasecmp works on whole strings rather than string prefixes, so the order doesn't matter in this case.
> 
> > Is it possible to IFDEF multiple HMACs? In which case, I propose not multiple IFDEF but chaining with ELSEIF.
> 
> Defining multiple HMACs is typical yes, you'll typically get SHA224, SHA256, SHA384 and SHA512 all at the same time.
> 
> > Does status get coerced to int? Some earlier definitions define int, but some don't. Probably safer to define.
> 
> This patch doesn't modify the behaviour of status, but can submit a separate PR for it? Good opportunity for a cleanup if you think it's needed.

I think if we can make the behaviour consistent throughout the file, that would be worth striving for. Would it be unusual to do it in this PR?

systemcrash:
How about a (rough pseudocode) `int maxbuf = strlen( "SHA512" ) + 1;` and use that locally, instead of a global DEFINE? Then the local code is a bit more easily readable for the next programmer who comes along :smile: a strlen doesn't strike me as too big a cost. Your call.

systemcrash:
@mrunge Looks OK to me.

mrunge (approved):
Thank you, looks okay to me.

PR #3927, opened 2021-10-06
External ids buffer

supalik:
ChangeLog: ovs_stas plugin:These changes will enhance the feature of receiving the data in buffer.

By  including "ExternalIds true"   as a New Config parameter all the external ids will be populated in the buffer
O/P:  ovs_stats="vmId=3979a0b0-bddc-4cdd-ae87-0a4c2071d1f9,ifaceId=7e0607f6-9179-4852-aa36-b56279c3310,mac=fa:16:3e:18:a1:ff",instance=""}

which can provide the visibility of external ids if exists.

PR #3936, opened 2021-10-25
snmp plugin: Added SNMPv3 TSM support

minfrin:
ChangeLog: snmp plugin: Added SNMPv3 TSM support.

This patch adds support for SNMPv3 TSM security (certificates) in addition
to existing SNMPv3 USM.

Documentation added detailing all the key things to understand about net-snmp
before trying to deploy certificates.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -61,7 +61,7 @@ debian_default_toolchain_task:
   build_script:
     - make -j2 -sk
   tests_script:
-    - make -j2 -sk check
+    - VALGRIND_OPTS="--errors-for-leak-kinds=definite" make -j2 -sk check || cat ./test-suite.log || false
     - /checks/check-built-plugins.sh
 
 ###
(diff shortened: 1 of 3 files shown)


mrunge:
I have seen the capabilities plugin failing on Trusty on a few PRs now. There must be an issue with the test itself, not the PR here.

elfiesmelfie:
> I have seen the capabilities plugin failing on Trusty on a few PRs now. There must be an issue with the test itself, not the PR here.

As noted in https://github.com/collectd/collectd/pull/3927#issuecomment-953350598, this seems to be a glibc issue and the version is not expected to be updated there.
Since Ubuntu Trusty/14.04 is an old distro, and [out of standard support](https://wiki.ubuntu.com/Releases), we could probably pick a newer version for the `debian_default_toolchain` job.

systemcrash:
Agreed. Do you think testing is just
more sensitive than issues which are likely to arise for users on that
distro when they use collectd?

elfiesmelfie:
> Agreed. Do you think testing is just more sensitive than issues which are likely to arise for users on that distro when they use collectd?

It could be, I haven't tested capabilities with trusty, but depending on where the problematic call is, it could effect the core functionality. If that is the case, I think we'll have to update the dependency checks for the capabilities plugin to look for a higher version of glibc.

minfrin:
@mrunge Alas we still have two CI failures. Unrelated test test_plugin_smart is failing on sid, and yum coredumps in fedora_rawhide.

Is it possible to take a look?

minfrin:
Quick ping on this one - any news?

PR #3938, opened 2021-10-30
Some improvements for InfluxDB support

matwey:
Hi,

Please find support for InfluxDB HTTP endpoint in this PR. Two minor things are also included into this PR: unix socket support for write_http plugin and meta_data support for format_influx.


systemcrash:
Oh yummy! Did you test against v6 or v5? Also, is there a way to avoid two copies of  `format_influxdb` ?

matwey:
@systemcrash

I've tested for v5 branch. I would like this features be in both branches eventually, but I don't understand workflow for v6. Am I supposed to PR for v6 first and then back port changes?

I am not sure that I understand you correctly, but we could link `format_influxdb` to `collectd` binary.

Note, that I also not quite sure about the last commit. So comments on it are especially welcome. Some plugins use meta for storing `raw values` and this will silently breaks influxdb databases making the series cardinality near infinite. Yet this meta can be filtered out using `Filter`, but user has to know in advance that it has to be done.

mrunge:
Thank you for your PR. Ideally I'd get to look at the new request over the next week. I am currently quite busy after being on vacation.

The capabilities test failing on debian trusty is a common pattern lately. Someone would have to investigate why this is failing. I am quite certain this is not caused by this PR.

matwey:
> Green across the board here. You ready @matwey ?

Ready for what? I am waiting for the review.

systemcrash:
Just saw you pushed some new commit. I guess we just wait for @mrunge

matwey:
Yes, I decided to improve PR a little, since nobody have seen it yet.

mrunge:
Thank you. It is not clear to me if you consider this PR as done and ready for review?

matwey:
Yes it is ready for review, thank you.

mrunge:
Okay, thank you for confirming.

Unfortunately, I am very busy these days and it may take some time for me to get to this. Please bear with me. 

Good news is, there are others being able to merge code (reviews can be done by anybody).

matwey:
Sorry. I have to push new commits. Yesterday, I've tried this with InfluxDB 2.1 and this helped me to found a little bug. I think it is better to fix in later than never.

matwey:
Ping?

matwey:
So what do I do? No attention to this PR for three months or so.

mrunge:
As you see here from the pull requests, the project can use your help. People just drive by and drop code for inspection. Code reviews would definitely help, but this is rarely done by other people.

matwey:
Do I understand correctly, that my review for other PRs will make sense for you?

mrunge:
Yes, any help with reviews is greatly appreciated!

Please don't get me wrong: proposing changes is also welcome, for sure! I didn't want to sound negative, but please understand, carefully reviewing code takes a significant amount of time. Time has become a precious resource for me.

PR #3943, opened 2021-11-22
Redfish plugin: complete rework and enhancement

matsKNY:
ChangeLog: Redfish plugin: complete rework and enhancement

The Redfish plugin of Collectd was, in the facts, unusable.
Indeed, it suffered from several heavy limitations which made it impossible to
monitor standard "recent" IT components (such as HPC compute blades):

+ Only members of anonymous collections could be collected;
+ Those members had to have a `DeviceName` attribute (this criteria was
  hard-coded and, thus, not configurable).
  The value of this attribute was compulsorily the type instance associated
  with the metric;
+ There was no way to select/filter which members of an anonymous collection
  should be collected;
+ The plugin could generate segmentation errors, and induced memory leaks;
+ The test suite and the documentation associated with the plugin were greatly
  incomplete.

Consequently, the Redfish plugin of Collectd was refactored and extensively
enhanced to make it possible and easy to monitor metrics and collect information
exposed by Redfish interfaces.

Hereinbelow are the main features of this refined Redfish plugin, when compared
to its previous implementation:

+ Attributes of any Redfish resource, together with attributes of non-nested
  composite objects, and attributes of members of non-nested anonymous
  collections can be collected (without any constraint on their attributes);
+ Possibility to select members of anonymous collections to be monitored
  based on their IDs, the fact that they have specific attributes, and/or the
  values of those attributes;
+ Possibility to prefix the type instance of a member of an anonymous collection
  with its ID in the collection;
+ Possibility to use the value of a configurable attribute of the members of an
  anonymous collection as their type instances;
+ As a result, by combining the two previous features, it is even possible to
  use an attribute for which the values exhibited by the members of the
  anonymous collection are not unique to define the type instance of the
  associated metrics, without any risk of overwriting;
+ The source of segmentation errors and memory leaks within the plugin were
  corrected (though, there are still some memory leaks which seems unrelated to
  the plugin);
+ A complete and effective test suite, based on an up-to-date mockup Redfish
  interface developed by the DMTF, was implemented;
+ A thorough documentation of the plugin was written.

On top of that, the function `oconfig_print_tree`, which prints the `oconfig`
configuration tree which root was specified, was implemented.
The latter is effectively compiled in `liboconfig` only if the debug
mode of `Collectd` is enabled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,8 +92,9 @@ bindings/perl/pm_to_blib
 # lint stuff
 *.ln
 
-#ide stuff
+# IDE stuff:
 .vscode
+.ycm_extra_conf.py
 
 # cscope stuff
 cscope.*
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -1495,6 +1495,72 @@
 #	DB_Path "/var/lib/rasdaemon/ras-mc_event.db"
 #</Plugin>
 
+#<Plugin redfish>
+#  <Service "mock1U">
+#    Host "localhost:10000"
+#    User ""
+#    Passwd ""
+#    Queries "thermal" "voltages" "temperatures" "ps1_voltage" "storage"
+#  </Service>
+#  <Query "thermal">
+#    Endpoint "/Chassis[0]/Thermal"
+#    <Resource "Fans">
+#      <Property "Reading">
+#        PluginInstance "Fans"
+#        Type "fanspeed"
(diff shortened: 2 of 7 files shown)


matsKNY:
Oups, I forgot to run the formatting script on two files. I will make a new pull request once it is corrected.
Can you please delete this PR?
Thank you in advance.

PR #3950, opened 2021-12-03
[intel_rdt] Fix build for LIBPQOS < 2

elfiesmelfie:
ChangeLog: intel_rdt: Update intel_rdt to build on libpqos < 2

### DIFF
--- a/src/intel_rdt.c
+++ b/src/intel_rdt.c
@@ -526,6 +526,7 @@ static void rdt_free_ngroups(rdt_ctx_t *rdt) {
 
   rdt->num_ngroups = 0;
 }
+#endif /* LIBPQOS2 */
 
 /*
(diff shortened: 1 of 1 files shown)


mrunge:
Thank you for working on this.
These #ifdef LIBPQOS2 are spread across the whole file, followed by an #endif and within a few lines, there is another #ifdef LIBPQOS2. It looks like there is room for improvement.

paweltox (approved):
Looks good to me.

kwiatrox (approved):
It might look clearer if whole `rdt_config_events` function is moved after the `#ifdef LIBPQOS2` block. But the fix in general works and looks good.

mrunge:
Thanks again, let's get it merged.

PR #3939, opened 2021-11-05
[ci][cirrus] Replace trusty with bionic/focal in debian_default_toolchain

elfiesmelfie:
Ubuntu 14.04 (Trusty) is out of standard support [1].
``make check`` fails for test_capabilities, as noted in [2].
[3] indicates that the cause is glibc, but that updates are not expected
to the version in trusty.

This PR replaces trusty with Ubuntu 18.04 (Bionic) and 20.04 (Focal).

[1] https://wiki.ubuntu.com/Releases
[2] https://github.com/collectd/collectd/pull/3936
[3] https://github.com/collectd/collectd/pull/3927#issuecomment-953350598

PR #3960, opened 2022-01-12
[collectd 6] intel_pmu 2.0 - updated version of plugin

bkotlowski:
Make it possible to register for multiple read callbacks [read threads].
    Split cores grouping per multiple read callbacks.

### DIFF
--- a/src/types.db
+++ b/src/types.db
@@ -218,6 +218,7 @@ ping                    value:GAUGE:0:65535
 ping_droprate           value:GAUGE:0:1
 ping_stddev             value:GAUGE:0:65535
 players                 value:GAUGE:0:1000000
+pmu_counter             scaled:COUNTER:0:U, raw:COUNTER:0:U, enabled:COUNTER:0:U, running:COUNTER:0:U
 pools                   value:GAUGE:0:U
 power                   value:GAUGE:U:U
 pressure                value:GAUGE:0:U
--- a/Makefile.am
+++ b/Makefile.am
@@ -1168,6 +1168,17 @@ intel_pmu_la_SOURCES = \
 intel_pmu_la_CPPFLAGS = $(AM_CPPFLAGS) $(BUILD_WITH_LIBJEVENTS_CPPFLAGS)
 intel_pmu_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(BUILD_WITH_LIBJEVENTS_LDFLAGS)
 intel_pmu_la_LIBADD = $(BUILD_WITH_LIBJEVENTS_LIBS)
+
+test_plugin_intel_pmu_SOURCES = \
+	src/intel_pmu_test.c \
+	src/utils/config_cores/config_cores.c \
+	src/daemon/configfile.c \
+	src/daemon/types_list.c
+test_plugin_intel_pmu_CPPFLAGS = $(AM_CPPFLAGS) $(BUILD_WITH_LIBJEVENTS_CPPFLAGS)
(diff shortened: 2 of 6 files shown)


eero-t:
This failure looks odd as that plugin was not modified:
> FAIL: test_plugin_capabilities

(You may want to push again to trigger retesting, in case it was some temporary hiccup.)

Because this PR drops most of the "intel_pmu" configuration options, I'm wondering was using a different name for the plugin ever considered, e.g. "intel_pmu2"?   With the same plugin name being used, when people update their collectd version, it will fail (due to now unrecognized config options), they need to read docs, edit config and try again.  Is this explicitly intended?

src/types.db
```diff
@@ -218,6 +218,7 @@ ping                    value:GAUGE:0:65535
 ping_droprate           value:GAUGE:0:1
 ping_stddev             value:GAUGE:0:65535
 players                 value:GAUGE:0:1000000
+pmu_counter             scaled:COUNTER:0:U, raw:COUNTER:0:U, enabled:COUNTER:0:U, running:COUNTER:0:U
```
eero-t, line 221: Plugin code does not use "pmu_counter" => why this entry is added?

mrunge:
> This failure looks odd as that plugin was not modified:
> 
> > FAIL: test_plugin_capabilities
> 
> (You may want to push again to trigger retesting, in case it was some temporary hiccup.)
> 

We may be bit here by an ancient ubuntu release or testing on that. Trusty went out of support in 2019, and we should just drop it from CI. If I remember correctly, we did that for the collectd-5 branch already, but not for this branch here.

eero-t:
Rebase this after https://github.com/collectd/collectd/pull/3939 is in, to get rid of the "trusty" failure.

eero-t:
> Rebase this after #3939 is in, to get rid of the "trusty" failure.

That was for v5 main branch only.  Matthias added separate https://github.com/collectd/collectd/pull/3972 for fixing the issue in v6 branch too (where this PR goes to), but that shows an additional CI issue (virt unit tests failing on Focal) which also needs to be fixed.

kwiatrox:
@eero-t The diff for this PR is somehow misleading, because the majority of the changes are already merged in the main (v5) branch. Due to the fact that `collectd-6.0` was not rebased for some time this PR introduces everything from main and additionally  compatibility changes for collectd v6. The major changes are in function `pmu_submit_counters` only.
Please compare to [intel_pmu_main](https://github.com/collectd/collectd/blob/main/src/intel_pmu.c).
Maybe the 6.0 branch should be rebased at some time to main again, because there is bigger and bigger divergence from some time.

For the `pmu_counter` in types.db it was added to v5 version, but for v6 it seems to be not necessary indeed.

eero-t:
> The diff for this PR is somehow misleading, because the majority of the changes are already merged in the main (v5) branch

intel_pmu.c in v6 branch does not seem to be using v6 collect API.  Why it differs from v5 at all?

I mean, until plugin has been ported to v6 API, I think all the changes should be done to v5 branch.  v6 branch would then occasionally update the plugin as-is from v5 branch.  Once plugin supports v6 API, that would be switched around, the changes should go to v6 branch first, and only occasionally be backported to v5 (if that's still relevant and there are important bug fixes).

eero-t:
FYI: All v5 and v6 branches CI issues should now be fixed.

This would be easier to review & verify if there were 2 separate PRs (or at least 2 separate commits in this PR):
* One syncing v5 version to v6 branch
* Another updating intel_pmu to v6 API + any other changes relevant for v6

Changelog entry in the PR(s) should also reflect these changes.

And types.db update should be dropped as it's not relevant for v6 branch.

@bkotlowski ?

PR #3964, opened 2022-01-20
Mysql: add support for custom queries

marekbecka:
ChangeLog: Plugin mysql: add support for user specified queries

This patch adds support for user defined queries, which enable to collect data from MySQL tables or add special statistics provided by MySQL server.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1498,7 +1498,10 @@ endif
 
 if BUILD_PLUGIN_MYSQL
 pkglib_LTLIBRARIES += mysql.la
-mysql_la_SOURCES = src/mysql.c
+mysql_la_SOURCES = \
+	src/mysql.c \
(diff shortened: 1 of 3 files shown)


eero-t:
Rebase this after https://github.com/collectd/collectd/pull/3939 is in, to get rid of the "trusty" failure.

mrunge:
Thank you for your patch! The mysql plugin is intended to provide statistics from mysql. 

For running arbitrary queries on a database, there is already the dbi plugin: https://collectd.org/wiki/index.php/Plugin:DBI
I am hesitant to add another way of doing the same thing as dbi does.

eero-t:
In the PR this is derived from there's a comment that "It's inspired by dbi plugin and uses auxiliary functions from utils_db_query."

@marekbecka If there's some functionality missing from DBI plugin, why not add it there?

PR #3966, opened 2022-01-24
disk: Add support for discard and sync statistics in Linux

marekbecka:
Disk: Support for discard and sync statistics in Linux

Linux 4.18 appends 4 additional fields for discard and since 5.5 there are another two more fields for flush requests in /proc/diskstats.


mrunge:
Thank you for your proposal! Would you mind to rebase the patch on latest main branch?

PR #3967, opened 2022-01-25
disk: add utilization

marekbecka:
ChangeLog: Disk:  Added a block device utilization in Linux

An utilization is computed as percentage of elapsed time during which a device has some IO operations currently in progress.


mrunge (approved):

PR #3968, opened 2022-01-27
[collectd 6] Add 'gpu_sysman' plugin for (Intel) GPU metrics

eero-t:
ChangeLog: gpu_sysman plugin: Add collectd v6 'gpu_sysman' plugin for Intel GPU metrics

This is based on earlier v5 version PR of the plugin: https://github.com/collectd/collectd/pull/3883

### New Sysman plugin for collectd

*Pre-production beta version*.

Plugin links against Level Zero loader:
* https://github.com/oneapi-src/level-zero

And provides GPU metrics using Level Zero Sysman API:
* https://spec.oneapi.com/level-zero/latest/sysman/api.html

### Metric reporting

One of the uses for this plugin is providing GPU metrics for Prometheus in Kubernetes cluster environment.

Because some metric values can fluctuate between their possible min & max values much more frequently than those values can be reported in such an environment, plugin can be configured to query few of the GPU metrics (currently frequency & memory usage) internally at higher rate + cache the results internally, and provide aggregate metrics of the cached data (currently min/max for given period) at the reporting intervals.

If there are errors with some metric types when their values are queried, plugin disables that metrics type for corresponding GPU. Each of the GPU metrics type queries can also be disabled from the plugin configuration.

### Supported metrics

Plugin provides most of the GPU metrics available through Sysman v1.1 API:
* Temperature (Celsius)
* Power usage (ujoules counter / Watts)
* Memory usage (bytes used / usage ratio)
* Memory BW (byte counter / usage ratio)
* Frequency (requested + actual frequency)
* Frequency throttling (usec counter / throttle ratio)
* Engine utilization (usec counter / utilization ratio)
* RAS error counters (several correctable + uncorrectable categories)

What metrics are available depends on the underlying HW, FW and driver stack.  Which metrics get queried + reported, can be controlled with plugin options.

### Metric units

While Gauge values are in units recommended by Prometheus and Open Metrics spec (e.g. seconds), "raw" Counter values are not (they use ujoules / usecs).  This is collectd issue as it stores Counter values as integers, instead of using floating point also for those, as expected by the Open Metrics spec:
* https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units

### Metric labels

Plugin  helper function adds originating GPU identification label(s) to all metric values:
* "pci_bdf": PCI device address in BDF notation
* "dev_file": primary device node file name (enabled with ADD_DEV_FILE define, for use with Kubernetes Intel GPU plugin)

Values can also include following additional labels (when Sysman provides corresponding information for given metric type):
* "sub_dev": sub-device ID
* "location": location for given metric
* "direction": e.g. read or write for bandwidth metric
* "type": e.g. memory type for memory metric
* "function": aggregation function used to provide value for data that was internally sampled at higher rate (min / max)

### Possible other metrics

Support could be later added to other metrics available through Sysman API, such as:
* Fabric port read/write BW
* PSU voltage, amperage, temp
* GPU voltage
* Fan speed
* Memory health

### Quality expectations

GCC gives no warnings for plugin when much stricter warnings are enabled than what collectd build uses by default.

80-90% unit-testing code coverage (depending on used GCC optimization level), with no memory access nor leakage warnings when unit-tests are run under Valgrind.

Among other things, unit-testing validates error handling for all Sysman functions used at run-time. See `gpu_sysman_test.c` for more information.

In addition to unit tests, most of the metrics have been tested also with data from real HW.

Different versions of the code have been tested semi-regularly since 2020 in (a development) setup consisting of few Intel iGPU and discrete (Xe) GPUs, but there has been no testing on production servers.

Testing has been done using Intel "compute-runtime" Sysman GPU backend, both with the public and internal Intel GPU SW stacks supporting Xe HW (in theory plugin should work also with the other Sysman API GPU backends than Intel one).

### Validation

Comment at top of `gpu_sysman_test.c` tells how to run validation tests.

Intel compute-runtime provides standalone Level Zero test program which can be used to check / validate the metric values reported by the plugin:
* https://github.com/intel/compute-runtime/blob/master/level_zero/tools/test/black_box_tests/zello_sysman.cpp


eero-t:
> debian_default_toolchain container:collectd/ci:trusty_amd64 Failing after 1m — Task Summary
> FAIL: test_plugin_capabilities

@mrunge / @elfiesmelfie  please remove that broken Trusty config from CI, or replace it with still maintained / working distro version.

> clang-format — Please run: contrib/format.sh src/gpu_sysman.c 

Why this test fails?  `contrib/format.sh` does NOT show any changes (any more):
```
$ contrib/format.sh src/gpu_sysman.c
$ git diff
```

PS. What clang-format version and config file is used by the _https://format.collectd.org/_ called by `format.sh`?  It sorts includes in different order than latest (local) `clang-format`, which is run from collectd directory:
```
$ clang --version
clang version 13.0.0 (Fedora 13.0.0-3.fc35)

$ cat .clang-format 
---
BasedOnStyle: LLVM
IncludeCategories:
  - Regex: '"collectd.h"'
  - Priority: -1

$ clang-format -i src/gpu_sysman.c

$ git diff
diff --git a/src/gpu_sysman.c b/src/gpu_sysman.c
index a721b618..d731eea0 100644
--- a/src/gpu_sysman.c
+++ b/src/gpu_sysman.c
@@ -60,9 +60,9 @@
 #include <libgen.h>
 #endif
 
-#include "collectd.h"
 #include "plugin.h"
 #include "utils/common/common.h"
+#include "collectd.h"
 
 #define PLUGIN_NAME "gpu_sysman"
 #define METRIC_PREFIX "collectd_" PLUGIN_NAME "_"
```

eero-t:
@mrunge Thanks for merging the v6 "write_prometheus" PR (it's useful for testing this PR)!

Any idea how that merge could fix the earlier clang-format warning for the code in gpu_sysman.c, which it not touch?  Is the "clang-format" CI check flaky?

PS. There are some review comments for all the changes between the v5 and v6 PRs of this plugin here: https://github.com/eero-t/collectd/pull/1

eero-t:
No problem. This plugin and its test code have grown pretty large (line-wise it's now in plugins' top-10), so there's a lot to review.

And the "trusty" CI issue needs to be fixed first, both in main and v6 branch, as its blocking also other things than this.

After that is fixed, I could e.g. rebase this to "collectd-6.0", so that prometheus merge drops from changes list and tests get re-run.

----

If you want to test building & running the plugin and its test code, all you need is Level-Zero frontend: https://github.com/oneapi-src/level-zero/

However, for the plugin init to actually succeed, and to get metrics, you will need also Level-Zero backend, i.e. _Level-Zero enabled_ version of Intel compute stack.  Because that is not yet in distros, you either need to build it yourself (hard), or get it from Intel's driver packages repo.

Besides, most of the metrics come only from discrete GPUs, only few of them are available for Intel integrated GPUs (mainly frequency, possibly also engine utilization and RAS counters depending on compute-runtime and kernel version). I.e. to test most of the metrics you would need at least DG1 discrete GPU, very latest kernel and enable the dGPU support with kernel force-probe option.

eero-t:
I decided to rebase all my v6 based branches on "collectd-6.0" already (after rebasing first to help testing it, I could not really leave rest unrebased as these branches are stacked on each other, sorry about that).

eero-t:
@mrunge if there are some commits that you'd like to be squashed, dropped or added to this PR, just comment.

For example, last commit in the PR is useful only with Kubernetes Intel GPU plugin, so you might not want it in the first version: https://github.com/collectd/collectd/pull/3968/commits/f049b824a219b47599c87cf85bc5799b928e540a

And e.g. the later test code refactoring from my dev-branch could be nice: https://github.com/eero-t/collectd/commits/gpu_sysman-6.0-dev

(Extra metrics added in the dev-branch do not make sense at this point as publicly available SW stack does not provide data even for all of the metrics in this PR on currently publicly available HW.  I've filed bugs about latter: https://github.com/intel/compute-runtime/issues?q=is%3Aissue+is%3Aopen+DG1+)

src/gpu_sysman.c
```diff
@@ -0,0 +1,1939 @@
...
+}
+
+/* Allocate 'config.samples' sized array of 'count' sized arrays having 'size'
+ * sized items.  If given array is already allocated, it and its subarrays
+ * is freed first
+ */
+static void **gpu_subarray_realloc(void **mem, int count, int size) {
+  uint32_t i;
+  gpu_subarray_free(mem);
+  mem = malloc(config.samples * sizeof(void *));
```
mrunge, line 189: We use smalloc from https://github.com/collectd/collectd/blob/main/src/utils/common/common.c#L238 all over the code. Would you be able to adjust to that?

src/gpu_sysman.c
```diff
@@ -0,0 +1,1939 @@
...
+  assert(mem);
+  for (i = 0; i < config.samples; i++) {
+    mem[i] = calloc(count, size);
```
mrunge, line 192: same here?

src/gpu_sysman.c
```diff
@@ -0,0 +1,1939 @@
...
+
+/* log given UUID (without dashes):
+ * https://en.wikipedia.org/wiki/Universally_unique_identifier
+ */
+static void log_uuid(const char *prefix, const uint8_t *byte, int len) {
+  int offset = strlen(prefix);
+  char buf[offset + 2 * len + 1];
+  strcpy(buf, prefix);
+  while (len-- > 0) {
+    sprintf(buf + offset, "%02x", *byte++);
```
mrunge, line 327: https://github.com/collectd/collectd/blob/main/src/utils/common/common.c#L93

src/gpu_sysman.c
```diff
@@ -0,0 +1,1939 @@
...
+    case ZES_RAS_ERROR_TYPE_UNCORRECTABLE:
+      type = "uncorrectable";
+      break;
+    default:
+      type = "unknown";
+    }
+    char buf[8];
+    const char *subdev = NULL;
+    if (props.onSubdevice) {
+      snprintf(buf, sizeof(buf), "%d", props.subdeviceId);
```
mrunge, line 789: https://github.com/collectd/collectd/blob/main/src/utils/common/common.c#L93

src/gpu_sysman.c
```diff
@@ -0,0 +1,1939 @@
...
+    ok = true;
+  }
+  free(ras);
+  return ok;
+}
+
+static void metric_set_subdev(metric_t *m, bool onsub, uint32_t subid) {
+  if (onsub) {
+    char buf[8];
+    snprintf(buf, sizeof(buf), "%d", subid);
```
mrunge, line 872: https://github.com/collectd/collectd/blob/main/src/utils/common/common.c#L93

src/gpu_sysman.c
eero-t: I do not see calloc wrapper in common.c, only the smalloc wrapper for malloc?

mrunge (commented):
Thank you @eero-t . In general, this looks really good.
I have added a few nits here and there, the gist is, we have common memory usage functions. If you could adjust the code to use them: awesome.

src/gpu_sysman.c
mrunge: there is only the smalloc wrapper, no specific form for calloc.

src/gpu_sysman.c
eero-t: IMHO "ssnprintf" wrapper is pointless, it just adds extra overhead & confusion compared to using just snprintf() directly.

As to using snprintf(), "%02x" on a byte does have a fixed size, so it's already safe.  Giving "sizeof(buf) - offset" as size for snprintf() is also somewhat error prone for future modifications.  I would prefer keeping that as strcpy() so that it gets highlighted when people review code.

(Best option would be for collectd having helper function for converting UUIDs to strings.)

src/gpu_sysman.c
eero-t: Would you like me to include also commit adding scalloc() wrapper to common.c?

src/gpu_sysman.c
eero-t: Sure, but I'd like first an answer to how you want to resolve the calloc stuff.

src/gpu_sysman.c
```diff
@@ -0,0 +1,1939 @@
...
+  return RET_OK;
+}
+
+/* log given UUID (without dashes):
+ * https://en.wikipedia.org/wiki/Universally_unique_identifier
+ */
+static void log_uuid(const char *prefix, const uint8_t *byte, int len) {
+  int offset = strlen(prefix);
+  char buf[offset + 2 * len + 1];
+  strcpy(buf, prefix);
```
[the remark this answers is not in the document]
eero-t: Updated: https://github.com/collectd/collectd/pull/3968/commits/91ad2340e2093a1ade64259a6116461d8408da56

src/gpu_sysman.c
eero-t: Nope, as snprintf -> ssnprintf is pointless change, and because all used collectd functions need to be copied gpu_sysman_test.c.

Note: gpu_sysman_test.c needs to use modified simplified versions of most of the collectd functions used by gpu_sysman.c, for test validation purposes.  Linking it with normal collectd code would cause conflicts and collectd code in general does not even build with the GCC warning options I use (with -Werror) for testing the Sysman plugin.

(I have PRs both for v5.x and v6.x collectd branches to fix subset of those warnings.)

mrunge (approved):
*thank you*

This looks good to me now. If we find additional issues, we could fix them afterwards.

mrunge:
Oh, the PR would be great, yes!

eero-t:
@mrunge v5 PR for scalloc(): https://github.com/collectd/collectd/pull/4014

eero-t:
Note for anybody reading this... Names of metrics and their labels listed in the original description change with OpenTelemetry support being introduced to Sysman plugin for v6.0 release.

PR #3971, opened 2022-02-07
[gha] Add a test log when the tests fail

elfiesmelfie:


eero-t:
Shouldn't the test logs be shown already with this PR?

Cirrus jobs just do not show the log, and failing github workflow tests just show (new) "Dump test logs" step NOT being run: https://github.com/collectd/collectd/runs/5126592061?check_suite_focus=true

----

There's also a bunch of changes here...

github/workflows/build.yml:
* Switch obsolete "trusty" to "bionic" + "focal"
* Add "silent" + "keep going" options to MAKEFLAGS
* Add $MAKEFLAGS to all make invocations
* Update "name"s
  * Differentiating "Make" and "make" is good, but it might as well fix the inconsistent capitalization for rest of names
* Add debug log output

.cirrus.yml:
* Add debug log output

Shouldn't the other than logging output changes be in a separate PR?  (Which both would be then needed also for v6 branch.)

elfiesmelfie:
> Shouldn't the test logs be shown already with this PR?
> 
> Cirrus jobs just do not show the log, and failing github workflow tests just show (new) "Dump test logs" step NOT being run: https://github.com/collectd/collectd/runs/5126592061?check_suite_focus=true
> 

They show up on the cirrus-ci website, (click "View more details on Cirrus CI " at the bottom of the page)
https://cirrus-ci.com/task/5586713861947392

I'm not sure why it isn't showing up here.
I'll try  

``tests_script:
      - make -j2 -sk check || cat ./test-suite.log || false
 ``

and see if this issue is because catting the logs is included in that on_failure step

> There's also a bunch of changes here...
> 
> github/workflows/build.yml:
> 
>     * Switch obsolete "trusty" to "bionic" + "focal"
> 
>     * Add "silent" + "keep going" options to MAKEFLAGS
> 
>     * Add $MAKEFLAGS to all make invocations
> 
>     * Update "name"s
>       
>       * Differentiating "Make" and "make" is good, but it might as well fix the inconsistent capitalization for rest of names
> 
>     * Add debug log output
> 
> 
> .cirrus.yml:
> 
>     * Add debug log output
> 
> 
> Shouldn't the other than logging output changes be in a separate PR? (Which both would be then needed also for v6 branch.)

Regarding to the number of changes, perhaps they should be separated. However, the fact that the logs are not yet coming up, I think that might be something to look at once those are working. 


There's also an additional consideration: What do we do about the failing virt tests on all the newer distros? (That is, if it can't be resolved in a reasonable time)
I'm inclined to suggest that we disable virt on these distros for the main CI, as we've done in the experimenal workflows, so that jobs are not reporting failed for unrelated reasons (i.e. virt tests failed on a change to the cpu plugin). 
At the same time, add in "focal with virt plugin" to the experimental workflows, or something similar to indicate "yes, this is expected to fail for now, but we don't want it to block things".

eero-t:
> There's also an additional consideration: What do we do about the failing virt tests on all the newer distros? (That is, if it can't be resolved in a reasonable time) I'm inclined to suggest that we disable virt on these distros for the main CI, as we've done in the experimenal workflows, so that jobs are not reporting failed for unrelated reasons (i.e. virt tests failed on a change to the cpu plugin). At the same time, add in "focal with virt plugin" to the experimental workflows, or something similar to indicate "yes, this is expected to fail for now, but we don't want it to block things".

Looking at the log you pointed out, virtio plugin issues are:
* Unit tests running through Valgrind and erroring out on Valgrind memcheck tool leak checks, not just on memory access errors
* All those errors being of the possibly lost type, not definitely lost, i.e. something that should be definitely be output from CI tests, so developer can review them, but IMHO not fail the test
* They are all reported for "libgobject-2.0.so.0.6400.6" library constructor which is run by dynamic linker when the library gets loaded.  How the test is actually run, is e.g. the plugin dlclose()d so that library finalizers will be run?

Easiest way to fix the test error would be to add Valgrind "--errors-for-leak-kinds=definite" for CI runs, and e.g. error on possibly lost allocs only when unit-tests are run manually (there could be an option for that, which CI would use).

elfiesmelfie:
> Looking at the log you pointed out, virtio plugin issues are:
> 
>     * Unit tests running through Valgrind and erroring out on Valgrind memcheck tool leak checks, not just on memory access errors
> 
>     * All those errors being of the possibly lost type, not definitely lost, i.e. something that should be definitely be output from CI tests, so developer can review them, but IMHO not fail the test

So far, I've been unable to output this info without failing the tests. 
We might be able to add another `make check || cat test-suite.log || true ` step that should always pass, and will give more information.

> 
>     * They are all reported for "libgobject-2.0.so.0.6400.6" library constructor which is run by dynamic linker when the library gets loaded.  How the test is actually run, is e.g. the plugin dlclose()d so that library finalizers will be run?

Unfortunately, I don't know. Can you point me towards where I might get an answer for this?
 
> Easiest way to fix the test error would be to add Valgrind "--errors-for-leak-kinds=definite" for CI runs, and e.g. error on possibly lost allocs only when unit-tests are run manually (there could be an option for that, which CI would use).

Thank you for your analysis, I am not very experienced in C and the build/debug tools around it. Your insight is very helpful.

If I understand correctly, what you're suggesting is:

* CI will run unit tests and report ONLY on definite leaks.
* Manually running make check will still report on possible leaks (i.e. same behaviour as before)

We can pass valgrind options to make check via the VALGRIND_OPTS env variable:

`VALGRIND_OPTS="--errors-for-leak-kinds=definite" make check`

I am unsure whether your suggestions above also would lead to the conclusion about adding that additional step that re-runs the tests and reports on all check types in the CI? Or is my suggestion sufficient?

eero-t:
It's a Glib issue: https://gitlab.gnome.org/GNOME/glib/-/issues/2076

For which there's a valgrind workaround PR that went in only a bit over year ago:  https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1768

eero-t:
> [[ci] make Valgrind ignore possible memory leaks]

VALGRIND_OPTS seems to have worked.  I do not understand how though, as "grep -R VALGRIND_OPTS ." matches nothing in collectd source directory?

> Build / build (false, el8_x86_64) (pull_request) Failing after 3m

```
> install bzip2
Run yum install -y bzip2 || apt install -y bzip2
CentOS Linux 8 - AppStream                       70  B/s |  38  B     00:00    
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
/__w/_temp/fda41809-ae0a-47e9-97d7-aecd2af1ab39.sh: line 1: apt: command not found
Error: Process completed with exit code 127.

> make distcheck
Run make $MAKEFLAG distcheck DISTCHECK_CONFIGURE_FLAGS="--disable-dependency-tracking --enable-debug"
/bin/sh: bzip2: command not found
make[1]: *** [Makefile:10680: dist-bzip2] Error 127
```

Why bzip2 install step did not fail although it returned an error?

> Build Parsing Results — Parsing Results

`Error while parsing tasks: INVALID_ARGUMENT: yaml: line 126: did not find expected key!`

elfiesmelfie:
> > [[ci] make Valgrind ignore possible memory leaks]
> 
> VALGRIND_OPTS seems to have worked. I do not understand how though, as "grep -R VALGRIND_OPTS ." matches nothing in collectd source directory?

VALGRIND_OPTS is supported by valgrind directly when invoked. Valgrind takes config  from some config files, as well as this env variable. It is only used in the CI.

```
$ valgrind --help
...
Extra options read from ~/.valgrindrc, $VALGRIND_OPTS, ./.valgrindrc`
---
```
> 
> > Build / build (false, el8_x86_64) (pull_request) Failing after 3m
> 
> ```
> > install bzip2
> Run yum install -y bzip2 || apt install -y bzip2
> CentOS Linux 8 - AppStream                       70  B/s |  38  B     00:00    
> Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
> /__w/_temp/fda41809-ae0a-47e9-97d7-aecd2af1ab39.sh: line 1: apt: command not found
> Error: Process completed with exit code 127.
> 
> > make distcheck
> Run make $MAKEFLAG distcheck DISTCHECK_CONFIGURE_FLAGS="--disable-dependency-tracking --enable-debug"
> /bin/sh: bzip2: command not found
> make[1]: *** [Makefile:10680: dist-bzip2] Error 127
> ```
> 
> Why bzip2 install step did not fail although it returned an error?
In `.github/workflows/build.yaml, the step is has a ``continue-on-error: true`` option set.
This prevents the job from being marked as failed when a single task fails.
Unfortunately, it then reports the step as passed, rather than marking it as failed. It's essentially doing the same thing as:
``
yum install -y bzip2 || apt install -y bzip2 || true
``
> 
> > Build Parsing Results — Parsing Results
> 
> `Error while parsing tasks: INVALID_ARGUMENT: yaml: line 126: did not find expected key!`
> 
> What file this is trying to parse?
However, I don't see this issue anywhere.

I'm pretty sure it ISN'T  ``.github/workflows/build.yaml``, since a parsing error would mean no workflows ran.

I think the issue is that CentOS8 is EOL  which is why bzip2 can't be downloaded. I'll create a PR against the collectd/ci-docker repo to switch the el8 container over to CentOS Stream 8.

elfiesmelfie:
> > > [[ci] make Valgrind ignore possible memory leaks]
> > 
> > 
> > VALGRIND_OPTS seems to have worked. I do not understand how though, as "grep -R VALGRIND_OPTS ." matches nothing in collectd source directory?
> 
> VALGRIND_OPTS is supported by valgrind directly when invoked. Valgrind takes config from some config files, as well as this env variable. It is only used in the CI.
> 
> I noticed it by chance while reading the help text for valgrind.
> 
> ```
> $ valgrind --help
> ...
> Extra options read from ~/.valgrindrc, $VALGRIND_OPTS, ./.valgrindrc`
> ---
> ```
> 
> > > Build / build (false, el8_x86_64) (pull_request) Failing after 3m
> > 
> > 
> > ```
> > > install bzip2
> > Run yum install -y bzip2 || apt install -y bzip2
> > CentOS Linux 8 - AppStream                       70  B/s |  38  B     00:00    
> > Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
> > /__w/_temp/fda41809-ae0a-47e9-97d7-aecd2af1ab39.sh: line 1: apt: command not found
> > Error: Process completed with exit code 127.
> > 
> > > make distcheck
> > Run make $MAKEFLAG distcheck DISTCHECK_CONFIGURE_FLAGS="--disable-dependency-tracking --enable-debug"
> > /bin/sh: bzip2: command not found
> > make[1]: *** [Makefile:10680: dist-bzip2] Error 127
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Why bzip2 install step did not fail although it returned an error?
> 
> In `.github/workflows/build.yaml`, the step is has a `continue-on-error: true` option set. This prevents the job from being marked as failed when a single task fails. Unfortunately, it then reports the step as passed, rather than marking it as failed. It's essentially doing the same thing as: `yum install -y bzip2 || apt install -y bzip2 || true`
> 
> > > Build Parsing Results — Parsing Results
> > 
> > 
> > `Error while parsing tasks: INVALID_ARGUMENT: yaml: line 126: did not find expected key!`
> > What file this is trying to parse?
> 
> I don't see this issue anywhere. I'm pretty sure it ISN'T `.github/workflows/build.yaml`, since a parsing error would mean no workflows ran.
> 
> I think the issue is that CentOS8 is EOL which is why bzip2 can't be downloaded. I'll create a PR against the collectd/ci-docker repo to switch the el8 container over to CentOS Stream 8.

EDIT: PR created here: https://github.com/collectd/ci-docker/pull/55

eero-t:
> > > Build Parsing Results — Parsing Results
> > 
> > `Error while parsing tasks: INVALID_ARGUMENT: yaml: line 126: did not find expected key!`
> > What file this is trying to parse?
> 
> I don't see this issue anywhere.

In case it was not clear, I'm referring to the output pointed by the Details link: https://github.com/collectd/collectd/pull/3971/checks?check_run_id=5146076608

> I'm pretty sure it ISN'T `.github/workflows/build.yaml`, since a parsing error would mean no workflows ran.

You can try this in the relevant repo branches: `find -name '*.y*ml' | xargs yamllint -d relaxed`

(After: sudo dnf|apt install yamllint)

elfiesmelfie:
> I think it's enough to call Valgrind once in the CI for each test.
> 
> Valgrind seems to be added by "testwrapper.sh", and that does not support VALGRIND_OPTS variable yet, only VALGRIND one (which is set to the valgrind path if it exists):
Since the VALGRIND_OPTS env var is used, 

>
> ```
> $ git grep -i -e valgrind -e memcheck
> Makefile.am:    src/valgrind.FreeBSD.suppress \
> Makefile.am:    src/valgrind.suppress \
> Makefile.am:LOG_COMPILER = env VALGRIND="@VALGRIND@" $(abs_srcdir)/testwrapper.sh
> configure.ac:AC_PATH_PROG([VALGRIND], [valgrind])
> src/valgrind.FreeBSD.suppress:   Memcheck:Addr4
> src/valgrind.suppress:   Memcheck:Leak
> src/write_sensu.c:  // added to original: not optimized, but keeps valgrind happy.
> testwrapper.sh:# A wrapper script for running tests. If valgrind is available, memory
> testwrapper.sh:MEMCHECK=""
> testwrapper.sh:if test -n "$VALGRIND"; then
> testwrapper.sh: MEMCHECK="$VALGRIND --quiet --tool=memcheck --error-exitcode=1"
> testwrapper.sh: MEMCHECK="$MEMCHECK --trace-children=yes"
> testwrapper.sh: MEMCHECK="$MEMCHECK --leak-check=full"
> testwrapper.sh: MEMCHECK="$MEMCHECK --gen-suppressions=all"
> testwrapper.sh: for f in "valgrind.$( uname -s ).suppress" "valgrind.suppress"; do
> testwrapper.sh:                 # Valgrind supports up to 100 suppression files.
> testwrapper.sh:                 MEMCHECK="$MEMCHECK --suppressions=$filename"
> testwrapper.sh:exec $MEMCHECK "$@"
> ```
> 
> From above, there seems to be already a suppression file for libnl.so, but I do not think the libgobject should be added there, because its warnings happen with only specific version of libgobject (in specific distros).
> 
> I think this test stuff gets called by some autotools testing feature. I think the *_test.c code is called just as it is, and those #include the plugin code. I.e. there's no plugin loading, just a test binary exercising plugin functions. END_TEST macro used by virt_test.c calls just normal exit() (not _exit()). I.e. library destructor should have been run normally.
> 
> All the possible leaks are either for g_type_register_fundamental() or g_type_register_static() which are probably some kind of singletons to store some per-type information. It's possible that libgobject just does not care if it leaks those, as in total it's only couple of KB.



> > > > Build Parsing Results — Parsing Results
> > > 
> > > 
> > > `Error while parsing tasks: INVALID_ARGUMENT: yaml: line 126: did not find expected key!`
> > > What file this is trying to parse?
> > 
> > 
> > I don't see this issue anywhere.
> 
> In case it was not clear, I'm referring to the output pointed by the Details link: https://github.com/collectd/collectd/pull/3971/checks?check_run_id=5146076608
> 
> > I'm pretty sure it ISN'T `.github/workflows/build.yaml`, since a parsing error would mean no workflows ran.
> 
> You can try this in the relevant repo branches: `find -name '*.y*ml' | xargs yamllint -d relaxed`
> 
> (After:` sudo dnf|apt install yamllint`)

The cirrus configs are in .cirrus.yml
The issue was some copy-pasta errors. They should be fixed presently.

elfiesmelfie:
Current state is that the jobs are mostly passing.

The outstanding failure is el8, which is being fixed here: https://github.com/collectd/ci-docker/pull/55, and once the PR is approved and merged, a new container will be built. Manual verification has indicated that this will fix the bzip install and `make distcheck` failures.

I'll come back to this next week, and separate the components of the PR, per feedback from @eero-t 
Once they merge, I'll propose them to the collectd-6 branch.

Currently, there are no github actions on the collectd 6 branch. Perhaps we should consider whether to add it, and what would be gained from doing so (when I added the GHA jobs, I was doing so to replace the travis-ci jobs, but seemingly, there were jobs duplicated between cirrus-ci and travis-ci already).

eero-t:
I would suggest splitting this into three:
* CI failure fixes (as separate commits):
  * VALGRIND_OPTS
  * "trusty" => "bionic" + "focal"
* Misc github/workflows/build.yml improvements (to match cirrus)
  * Add "silent" + "keep going" options to MAKEFLAGS
  * Add $MAKEFLAGS to all make invocations
  * Update "name"s
* Test log output

Of which the first is obviously most critical to get in.

(If I understood correctly and log outputting still does not work as well it should, that part could be left here for further testing.)

elfiesmelfie:
> I would suggest splitting this into three:
> 
>     * CI failure fixes (as separate commits):
>       
>       * VALGRIND_OPTS
>       * "trusty" => "bionic" + "focal"
> 

here: https://github.com/collectd/collectd/pull/3975

>     * Misc github/workflows/build.yml improvements (to match cirrus)
>       
>       * Add "silent" + "keep going" options to MAKEFLAGS
>       * Add $MAKEFLAGS to all make invocations
>       * Update "name"s
> 

https://github.com/collectd/collectd/pull/3976

>     * Test log output

That's this PR.
> 
> Of which the first is obviously most critical to get in.
> 
> (If I understood correctly and log outputting still does not work as well it should, that part could be left here for further testing.)

I think log outputs are working okay. I had to combine the tests and log output steps in cirrus, but it worked eventually.
IINM, the test-suite.log is always created, so I can skip the check for whether the file exists.

mrunge:
Thank you for working on this, I think I've merged all the pending PRs here?

mrunge (approved):
thank you, looks good to me.

PR #3972, opened 2022-02-08
[ci][cirrus] Replace trusty with bionic/focal in debian_default_toolc…

mrunge:
…hain

Ubuntu 14.04 (Trusty) is out of standard support [1].
``make check`` fails for test_capabilities, as noted in [2].
[3] indicates that the cause is glibc, but that updates are not expected
to the version in trusty.

This PR replaces trusty with Ubuntu 18.04 (Bionic) and 20.04 (Focal).

[1] https://wiki.ubuntu.com/Releases
[2] https://github.com/collectd/collectd/pull/3936
[3] https://github.com/collectd/collectd/pull/3927#issuecomment-953350598

(cherry picked from commit b5d8ad13f4e300a7bfe342d27ad804003fcc9173)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -81,12 +81,15 @@ debian_default_toolchain_task:
         image: collectd/ci:stretch_i386
       allow_failures: false
     - container:
-        image: collectd/ci:trusty_amd64
-      allow_failures: false
-    - container:
         image: collectd/ci:xenial_amd64
       allow_failures: false
-    # debian/unstable is expected to fail
+    - container:
+        image: collectd/ci:bionic_amd64
+      allow_failures: false
+    - container:
+        image: collectd/ci:focal_amd64
+      allow_failures: false
+      # debian/unstable is expected to fail
     - allow_failures: true
       skip_notifications: true
       only_if: $CIRRUS_BRANCH == 'master'


eero-t:
> debian_default_toolchain container:collectd/ci:focal_amd64 
...
> FAIL: test_plugin_virt

Same issue with v6 branch as with v5 main branch I assume. Hopefully these become clearer once @elfiesmelfie makes failing CI test-cases to output failure log.

mrunge:
I can reproduce the virt failure on e.g Fedora 35 running the test suite locally.

eero-t:
> > FAIL: test_plugin_virt
> 
> Same issue with v6 branch as with v5 main branch I assume. Hopefully these become clearer once @elfiesmelfie makes failing CI test-cases to output failure log.

@mrunge It's due to that test (and others?) being run with valgrind using an option that makes even _possible_ leak warnings to be treated as errors, and for some reason libgobject library initializer allocs trigger those warnings.  See: https://github.com/collectd/collectd/pull/3971#issuecomment-1034742686

(IMHO only Valgrind memcheck tool's definitive leak warnings should be treated as errors.  Possible ones can be just code confusing Valgrind. At least earlier, that could happen e.g. with allocator using offsets instead of pointers to keep track of its allocations.)

elfiesmelfie:
@mrunge, This should pass since https://github.com/collectd/collectd/pull/3975 was merged

eero-t:
@mrunge Maybe it's better to update this v6 branch PR to include also rest of the CI fix commits from v5 #3975 PR?

elfiesmelfie (approved):
Looks good to me.

I suspect the failures are due to valgrind. 
https://github.com/collectd/collectd/pull/3977 should fix it. 
Unfortunately, it doesn't look like I can open a PR against your fork.

PR #3973, opened 2022-02-11
slurm: compatibility with Slurm >= 21.08.0

pllopis:
Add compatibility for Slurm versions >= 21.08.0

### DIFF
--- a/src/slurm.c
+++ b/src/slurm.c
@@ -40,6 +40,453 @@
 /* this function declaration is missing in slurm.h */
 extern void slurm_free_stats_response_msg(stats_info_response_msg_t *msg);
 
+#if SLURM_VERSION_NUMBER >= SLURM_VERSION_NUM(21, 8, 0)
+enum slurm_node_states {
+  INVAL,
+  MAINT_NONRESP,
+  MAINT,
+  REBOOT_REBOOT_ISSUED,
+  REBOOT_NONRESP,
+  REBOOT,
+  DRAINING_MAINT,
+  DRAINING_REBOOT_ISSUED,
+  DRAINING_REBOOT,
+  DRAINING_POWERUP,
+  DRAINING_POWERING_DOWN,
+  DRAINING_POWERED_DOWN,
+  DRAINING_POWERDOWN,
+  DRAINING_NONRESP,
+  DRAINING,
+  DRAINED_MAINT,
+  DRAINED_REBOOT_ISSUED,
+  DRAINED_REBOOT,
+  DRAINED_POWERUP,
+  DRAINED_POWERING_DOWN,
+  DRAINED_POWERED_DOWN,
+  DRAINED_POWERDOWN,
+  DRAINED_NONRESP,
+  DRAINED,
+  FAILING_NONRESP,
+  FAILING,
+  FAIL_NONRESP,
+  FAIL,
+  REBOOT_ISSUED,
+  CANCEL_REBOOT,
+  CLOUD,
+  POWER_DOWN,
+  POWER_UP,
+  POWERING_DOWN,
+  POWERED_DOWN,
+  POWERING_UP,
+  DOWN_MAINT,
+  DOWN_REBOOT_ISSUED,
+  DOWN_REBOOT,
+  DOWN_POWERUP,
+  DOWN_POWERING_DOWN,
+  DOWN_POWERED_DOWN,
+  DOWN_POWERDOWN,
+  DOWN_NONRESP,
+  DOWN,
+  ALLOCATED_MAINT,
+  ALLOCATED_REBOOT_ISSUED,
+  ALLOCATED_REBOOT,
+  ALLOCATED_POWERUP,
+  ALLOCATED_POWERING_DOWN,
+  ALLOCATED_POWERED_DOWN,
+  ALLOCATED_POWERDOWN,
+  ALLOCATED_NONRESP,
+  ALLOCATED_COMP,
+  ALLOCATED,
+  COMPLETING_MAINT,
+  COMPLETING_REBOOT_ISSUED,
+  COMPLETING_REBOOT,
+  COMPLETING_POWERUP,
+  COMPLETING_POWERING_DOWN,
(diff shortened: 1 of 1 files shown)


eero-t:
And are these nodes states not defined in some already existing Slurm header?

Cannot the new states be added to the existing enums, arrays and code, instead of ifdeffing the whole thing?

pllopis:
Hi Eero, thanks for taking the time to review this!

Your comments make a lot of sense, however I do not see an easy way around these ifdefs (without dropping support for past Slurm versions). Let me explain why I think this, and you can let me know your thoughts about it:

It seems by looking at the code that the node state arrays might be joinable, since the newer version only adds new states.
However, these states (the ones that are not defined as strings) depend on `#define`s in Slurm's provided `slurm.h`.
Hence, if we merged them, it would either not compile (for e.g. Slurm v20), or there would have to be a _lot_ of #ifdefs. (This single version change would be quite a few, but maintaining these would only get worse with future releases.)

There is no equivalent to this enum in slurm.h unfortunately. There are only base states (enum), and a separate set of (`#define`d) node state bit flags. The enum and function provided in this code are a comprehensive way to account for unique node states in Slurm.

Regarding the `slurm_node_state` function: this function is almost identical to an internal Slurm function named `node_state_string`, which is not part of the public API. But in any case, that Slurm internal function returns the states as strings, which is less useful for this code.

In terms of differences in that `#ifdef`ed function, there are not just additions here, but also lines that are removed/behave differently. The same issue as before applies here: the function would have to be filled with ifdef logic for the code to compile at all, which would get especially hard to maintain especially as new versions are released.

This is the reason I decided that having a single ifdef (or at least one per Slurm version) would be a good compromise, which preserves readability of the code, and makes it easier to maintain (especially for future versions). The trade-off is obviously having several very long arrays in there, but I think it's a positive trade-off considering you can still compile the plugin against a Slurm version that isn't necessarily the latest. 

It's worth noting that the Slurm folks make a major release per year, and every major release "breaks" the previous API (the corresponding `libslurm.so` major version changes). Meaning that these Slurm versions change ~yearly, and therefore I think it's useful to be able to support more than one version at a time.

eero-t:
> It's worth noting that the Slurm folks make a major release per year, and every major release "breaks" the previous API (the corresponding libslurm.so major version changes). Meaning that these Slurm versions change ~yearly, and therefore I think it's useful to be able to support more than one version at a time.

Ok, that's a good point.  As would be testing of different versions. If code is not changed for the older version, it does not need to be tested, only the newer version.

My suggestion is to do code splitting per Slurm API version differently though.

Put the per-version Slurm code into `slurm_{version}.h` header files, where _{version}_ is the Slurm version where the major API change was introduced.

"slurm.c" would then have just ifdefs on which of these headers it should include.

(As I'm not collectd maintainer myself, it's better to ask one for opinion... @mrunge please comment whether doing it that way sounds good to you too.)

That way:
* Comparing the changes between different Slurm versions would be trivial, just diff the relevant headers
* It's clear how to add and drop support for specific Slurm versions
* Code is more readable / reviewable

For minor API changes, those headers could include few ifdefs, with new header being added only with major API changes, to prevent increasing number of headers too much. But if there starts to be larger number of these headers, they could be put into their own subdirectory too.

PS. Add some version of your explanation also as comment to slurm.c (before the header inclusion), to state why things are done like that.

eero-t:
Btw. @pllopis I just noticed that you are the original author of the plugin.  Please consider migrating it to v6 API: https://collectd.org/wiki/index.php/V5_to_v6_migration_guide

And adding PR for "collectd-6.0" branch.

v6 is aimed to match better telemetry requirements for cloud, and does not any more require types.db updates.  One of the main advantages of the new API is that you can add arbitrary labels to each metric.

Wiki instructions are not very complete, so when doing the migration, it's best to use commits of the already migrated plugins as an example.

PS. There are few specs that are relevant for cloud telemetry.  If you want to match their recommendations, you need to change the current metric names a bit, at least add unit name suffixes to them:
* https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units
* https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#suffixes
* https://prometheus.io/docs/practices/naming/#metric-names
* https://prometheus.io/docs/practices/naming/#base-units

pllopis:
Hi Eero,

I think dividing these up in different slurm_{version}.h header files makes sense.

Regarding collectd6, I'm aware, just been lacking the time to port it, since I would do almost a full rewrite. Since the tag-based approach allows for arbitrary metrics, I'd make it report less aggregated and more fine-grained stats.

Cheers

pllopis:
> Would the parts you change in this PR, be impacted by such rewrite?
> 
> (If there are parts of such rewrite which would make sense also for v5, you might do separate PR for those, before starting the v6 migration for real.)

I think that the rewrite would be for v6 only. There were some compromises in the data types that I made due to the RRD format which would no longer apply in v6, and therefore the redesign for v6 would not make sense for v5.
(Having said that, I have not really started on a v6 version yet, but pretty sure the above would apply)

eero-t:
Is there old enough Slurm version that the older header does not support it?  If yes, might be better to add check for that too, and give `#error Unsupported (too old) Slurm version` otherwise.

eero-t:
@elfiesmelfie CI is acting still / again:
> Build / experimental (sid_amd64) (pull_request) Failing after 5m
>
>  FAIL: test_plugin_netlink
> PASS: test_plugin_network
> PASS: test_plugin_pcie_errors
> FAIL: test_plugin_smart

and:
> Build / experimental (fedora_rawhide_x86_64) (pull_request) Failing after 3m
>
> Run yum install -y bzip2 || apt install -y bzip2
> ...
> Module yaml error: Unexpected key in data: static_context [line 9 col 3]
> ....
> Error: GPG check FAILED

and:
> debian_default_toolchain container:collectd/ci:focal_amd64 Failing after 2m
>
> FAIL: test_plugin_virt

Can you see what are the actual errors with the failing unit-tests?

PR #3974, opened 2022-02-13
smart: use namespace id 0 for Identify Controller command

i-rinat:
Some drives don't like when .nsid field is set to 0xFFFFFFFF when asked 
to process Identify Controller command, and report that as an error.


i-rinat:
Although NVMe spec doesn't say how drives should process Identify Controller command when `CDW1.NSID` is set to `FFFFFFFFh`, at least Samsung 970 EVO treat that as an error. This cranks up "Error Information Log Entries" counter every time smart plugin takes another measurement. As far as I know, these errors have no negative effects, but increasing error counter is a bit annoying.

`nvme` command-line tool also uses namespace id 0 when performing `id-ctrl` command.

i-rinat:
`test_plugin_virt` seems to fail on CI, but it passes just fine when I try to run it locally. Are there any pointers on how to investigate failures on CI?

eero-t:
> `test_plugin_virt` seems to fail on CI, but it passes just fine when I try to run it locally. Are there any pointers on how to investigate failures on CI?

See discussion here on getting logs, reason for the virt plugin fail + workaround: https://github.com/collectd/collectd/pull/3971

eero-t:
> Although NVMe spec doesn't say how drives should process Identify Controller command when `CDW1.NSID` is set to `FFFFFFFFh`, at least Samsung 970 EVO treat that as an error. This cranks up "Error Information Log Entries" counter every time smart plugin takes another measurement. As far as I know, these errors have no negative effects, but increasing error counter is a bit annoying.

Looking at the latest/current (2.0b) NVM Express® Base Specification: https://nvmexpress.org/wp-content/uploads/NVM-Express-Base-Specification-2.0b-2021.12.18-Ratified.pdf

"Figure 87: Common Command Format" states following:

> Namespace Identifier (NSID): This field specifies the namespace that this command applies to. If the namespace identifier is not used for the command, then this field shall be cleared to 0h. The value FFFFFFFFh in this field is a broadcast value (refer to section 3.2.1.2), where the scope (e.g., the NVM subsystem, all attached namespaces, or all namespaces in the NVM subsystem) is dependent on the command. Refer to Figure 138 and Figure 390 for commands that support the use of the value FFFFFFFFh in this field.

> Specifying an inactive namespace identifier (refer to section 3.2.1.4) in a command that uses the namespace identifier shall cause the controller to abort the command with a status code of Invalid Field in Command, unless otherwise specified.  Specifying an invalid namespace identifier (refer to section 3.2.1.2) in a command that uses the namespace identifier shall cause the controller to abort the command with a status code of Invalid Namespace or Format, unless otherwise specified.

> If the namespace identifier is used for the command (refer to Figure 138), the value FFFFFFFFh is not supported for that command, and the host specifies a value of FFFFFFFFh, then the controller should abort the command with a status code of Invalid Field in Command, unless otherwise specified.

> If the namespace identifier is not used for the command and the host specifies a value from 1h to FFFFFFFFh, then the controller should abort the command with a status code of Invalid Field in Command, unless otherwise specified.

=> there are only specific, very few commands which accept all-namespace broadcast value.  And for all the other commands the controller should actually abort the command in that case.

"Figure 138" footnotes state that:
> A subset of commands use the Namespace Identifier (NSID) field. If the Namespace Identifier field is used, then the
value FFFFFFFFh is supported in this field unless otherwise indicated in footnotes in this figure that a specific command
does not support that value or supports that value only under specific conditions. When this field is not used, the field
is cleared to 0h as described in Figure 87.

Looking at where these are used in the smart.c code:
```
$ grep -e NVME_ADMIN_ -e NVME_NSID_ALL src/nvme.h src/smart.c 
src/nvme.h:#define NVME_NSID_ALL 0xffffffff
src/nvme.h:#define NVME_ADMIN_GET_LOG_PAGE 0x02
src/nvme.h:#define NVME_ADMIN_IDENTIFY 0x06
src/smart.c:              &(struct nvme_admin_cmd){.opcode = NVME_ADMIN_IDENTIFY,
src/smart.c:                                       .nsid = NVME_NSID_ALL,
src/smart.c:                 &(struct nvme_admin_cmd){.opcode = NVME_ADMIN_GET_LOG_PAGE,
src/smart.c:                                          .nsid = NVME_NSID_ALL,
src/smart.c:            &(struct nvme_admin_cmd){.opcode = NVME_ADMIN_GET_LOG_PAGE,
src/smart.c:                                     .nsid = NVME_NSID_ALL,
```

For the log command (0b10), namespace is supported and it can use broadcast.

For identify command (0b110) which you modified, there's a following footnote:
> Use of the Namespace Identifier field depends on the CNS value in the Identify Command as described in Figure 273.

When looking at "Identify command overview" section, it states:
> When issuing the Identify command, if the specified namespace is not associated with an I/O Command Set that supports the specified Identify CNS value (refer to Figure 273), then the controller shall abort the command with a status code of Invalid I/O Command Set.

(Then there are details about differences in the commands depending on what features and spec version the controller supports.)

On quick look of that table in figure 273, I was not able to decipher whether it's more OK to always use zero as namespace rather than broadcast value, or is that something specific to controller and whatever stuff it happens to support.

@bkotlowski as you added the original VNME support, could you look into this?

(I also wonder why libnvme was not used for these queries, as that has wrappers for them: https://github.com/linux-nvme/libnvme/blob/master/src/nvme/ioctl.c#L368)

eero-t:
> @bkotlowski as you added the original VNME support, could you look into this?

@bkotlowski?

Potentially related Debian collectd bugs and patches:
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=998166
* https://salsa.debian.org/debian/pkg-collectd/-/blob/master/debian/patches/fix-smart-test

vincele:
@eero-t : yes, both debian links looks related to this issue / PR.
@i-rinat : maybe you can try the debian fix with your NVMe drive ?

It would be nice to have this properly fixed...

i-rinat:
@vincele> maybe you can try the debian fix with your NVMe drive

I do use Debian, and so the patch "fix-smart-test" was already applied when I noticed the error count increase. I believe it's unrelated to the issue, since it only fixes an `ioctl()` mocking for the testing purposes, not actual talking to devices part of the code.

PhilippWendler:
I am affected by this problem with Samsung PM983 SSDs.

I tried the patch from this PR, but I only get `*** stack smashing detected ***: terminated` and collectd terminates. Maybe my SSD writes more data than expected into `&vid`?  I increased the size of `vid` to 4096 bytes (the NVME spec says that this is the size of the data structure returned by the identify command) and then it does not crash.

However, then I still get entries complaining about invalid namespace `0xffffffff` in the error log of the SSD. What I don't understand yet is that the error log complains about invalid namespaces for command is that are not 0x06 (identify).

Here is the error log:
```
Error Log Entries for device:nvme0 entries:64
.................
 Entry[ 0]   
.................
error_count	: 2426031
sqid		: 0
cmdid		: 0xf011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[ 1]   
.................
error_count	: 2426030
sqid		: 0
cmdid		: 0xd013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[ 2]   
.................
error_count	: 2426029
sqid		: 0
cmdid		: 0xe011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[ 3]   
.................
error_count	: 2426028
sqid		: 0
cmdid		: 0xc013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[ 4]   
.................
error_count	: 2426027
sqid		: 0
cmdid		: 0xd011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[ 5]   
.................
error_count	: 2426026
sqid		: 0
cmdid		: 0xb013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[ 6]   
.................
error_count	: 2426025
sqid		: 0
cmdid		: 0xc011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[ 7]   
.................
error_count	: 2426024
sqid		: 0
cmdid		: 0xa013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[ 8]   
.................
error_count	: 2426023
sqid		: 0
cmdid		: 0xb011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[ 9]   
.................
error_count	: 2426022
sqid		: 0
cmdid		: 0x9013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[10]   
.................
error_count	: 2426021
sqid		: 0
cmdid		: 0xa011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[11]   
.................
error_count	: 2426020
sqid		: 0
cmdid		: 0x8013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[12]   
.................
error_count	: 2426019
sqid		: 0
cmdid		: 0x9011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[13]   
.................
error_count	: 2426018
sqid		: 0
cmdid		: 0x7013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[14]   
.................
error_count	: 2426017
sqid		: 0
cmdid		: 0x8011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[15]   
.................
error_count	: 2426016
sqid		: 0
cmdid		: 0x6013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[16]   
.................
error_count	: 2426015
sqid		: 0
cmdid		: 0x7011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[17]   
.................
error_count	: 2426014
sqid		: 0
cmdid		: 0x5013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[18]   
.................
error_count	: 2426013
sqid		: 0
cmdid		: 0x6011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[19]   
.................
error_count	: 2426012
sqid		: 0
cmdid		: 0x4013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[20]   
.................
error_count	: 2426011
sqid		: 0
cmdid		: 0x5011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[21]   
.................
error_count	: 2426010
sqid		: 0
cmdid		: 0x3013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[22]   
.................
error_count	: 2426009
sqid		: 0
cmdid		: 0x4011
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[23]   
.................
error_count	: 2426008
sqid		: 0
cmdid		: 0x2013
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[24]   
.................
error_count	: 2426007
sqid		: 0
cmdid		: 0xe005
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[25]   
.................
error_count	: 2426006
sqid		: 0
cmdid		: 0x7007
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[26]   
.................
error_count	: 2426005
sqid		: 0
cmdid		: 0x4002
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[27]   
.................
error_count	: 2426004
sqid		: 0
cmdid		: 0x3000
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[28]   
.................
error_count	: 2426003
sqid		: 0
cmdid		: 0x1002
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[29]   
.................
error_count	: 2426002
sqid		: 0
cmdid		: 0
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[30]   
.................
error_count	: 2426001
sqid		: 0
cmdid		: 0xe002
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[31]   
.................
error_count	: 2426000
sqid		: 0
cmdid		: 0xd000
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[32]   
.................
error_count	: 2425999
sqid		: 0
cmdid		: 0xb002
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[33]   
.................
error_count	: 2425998
sqid		: 0
cmdid		: 0xa000
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[34]   
.................
error_count	: 2425997
sqid		: 0
cmdid		: 0x9002
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[35]   
.................
error_count	: 2425996
sqid		: 0
cmdid		: 0x8000
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[36]   
.................
error_count	: 2425995
sqid		: 0
cmdid		: 0x7002
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[37]   
.................
error_count	: 2425994
sqid		: 0
cmdid		: 0x6000
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[38]   
.................
error_count	: 2425993
sqid		: 0
cmdid		: 0xc018
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[39]   
.................
error_count	: 2425992
sqid		: 0
cmdid		: 0xa01a
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[40]   
.................
error_count	: 2425991
sqid		: 0
cmdid		: 0xa018
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[41]   
.................
error_count	: 2425990
sqid		: 0
cmdid		: 0x901a
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[42]   
.................
error_count	: 2425989
sqid		: 0
cmdid		: 0x9018
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[43]   
.................
error_count	: 2425988
sqid		: 0
cmdid		: 0x801a
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[44]   
.................
error_count	: 2425987
sqid		: 0
cmdid		: 0x8018
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[45]   
.................
error_count	: 2425986
sqid		: 0
cmdid		: 0x701a
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[46]   
.................
error_count	: 2425985
sqid		: 0
cmdid		: 0x7018
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[47]   
.................
error_count	: 2425984
sqid		: 0
cmdid		: 0x601a
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[48]   
.................
error_count	: 2425983
sqid		: 0
cmdid		: 0x5018
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[49]   
.................
error_count	: 2425982
sqid		: 0
cmdid		: 0x401a
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[50]   
.................
error_count	: 2425981
sqid		: 0
cmdid		: 0x4018
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[51]   
.................
error_count	: 2425980
sqid		: 0
cmdid		: 0x201a
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[52]   
.................
error_count	: 2425979
sqid		: 0
cmdid		: 0x101a
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[53]   
.................
error_count	: 2425978
sqid		: 0
cmdid		: 0x19
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[54]   
.................
error_count	: 2425977
sqid		: 0
cmdid		: 0xe01b
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[55]   
.................
error_count	: 2425976
sqid		: 0
cmdid		: 0xf019
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[56]   
.................
error_count	: 2425975
sqid		: 0
cmdid		: 0xc01b
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[57]   
.................
error_count	: 2425974
sqid		: 0
cmdid		: 0xd019
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[58]   
.................
error_count	: 2425973
sqid		: 0
cmdid		: 0xb01b
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[59]   
.................
error_count	: 2425972
sqid		: 0
cmdid		: 0xc019
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[60]   
.................
error_count	: 2425971
sqid		: 0
cmdid		: 0xa01b
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[61]   
.................
error_count	: 2425970
sqid		: 0
cmdid		: 0xb019
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[62]   
.................
error_count	: 2425969
sqid		: 0
cmdid		: 0x901b
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
 Entry[63]   
.................
error_count	: 2425968
sqid		: 0
cmdid		: 0xa019
status_field	: 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag	: 0
parm_err_loc	: 0x4
lba		: 0
nsid		: 0xffffffff
vs		: 0
trtype		: The transport type is not indicated or the error is not transport related.
cs		: 0
trtype_spec_info: 0
.................
```

These errors have to come from collectd's smart plugin, because if I remove this plugin or ignore the SSD in its config, no new error entries are added.

I am willing to try out further things, but right now I have no idea what to do. I am not familiar with NVMe unfortunately.

eero-t:
> Here is the error log:

that did not come out correctly.  Could you fix the code tagging for it?

(those backticks need to be at very start of the line, with nothing else, check with preview that it looks OK.)

PhilippWendler:
Sorry, fixed. Having the backticks on their own line is unfortunately not enough in combination with `<details>`. I had checked it in the beginning, but then added `<details>` and didn't check again.

eero-t:
Your log had errors for so many command IDs, that now I'm wondering whether there was _any_ command ID which _did accept_ broadcast (-1) namespace?

PhilippWendler:
I honestly have no idea. All I know is that these errors only appear if the smart plugin is enabled in collectd for this SSD. And as far as I understand `smart.c`, it doesn't even use any command except `NVME_ADMIN_IDENTIFY` and `NVME_ADMIN_GET_LOG_PAGE` (opcodes 0x06 and 0x02). I also don't know why the error log shown by the `nvme` tool has 16-bit command ids whereas the opcode in `smart.c` is just one byte. Reading through this thread was my first contact ever with such low-level details of NVMe... But I am happy to test something or provide more info if one can tell me what to do.

eero-t:
> But I am happy to test something or provide more info if one can tell me what to do.

@mrunge ?

(I'm myself just a contributor / maintainer for one plugin that's not yet in any collectd release, not one of project devs, so I cannot really do anything about this except a bit of review help.)

mrunge:
> > But I am happy to test something or provide more info if one can tell me what to do.
> 
> @mrunge ?
> 
> (I'm myself just a contributor / maintainer for one plugin that's not yet in any collectd release, not one of project devs, so I cannot really do anything about this except a bit of review help.)

Sorry, I am currently on the road at a conference + traveling. It may take me some time to get back to this.

In general: what helps me is a clear proof/hint that the proposed fix makes it behave in the correct way. I am not very fond of replacing predefined constants with integers without any additional comment. It makes the code harder to read, and the intention becomes less clear.

i-rinat:
Oh, great. I think there is a regression in the Linux kernel. Or a behavior change which is also not good.

In short, `NVME_IOCTL_ADMIN_CMD`:`NVME_ADMIN_IDENTIFY` now writes more than `.data_len` bytes, that's on kernel 6.1.27. I've tried kernel 4.19.249, and there only `.data_len` bytes are overwritten in the buffer. I haven't had a chance to do a bisection to find when and where the regression was introduced.

i-rinat:
The regression appeared around Linux 5.19, and there are multiple kernel versions out there with that behavior. Even if that gets fixed, it's better to be on the safe side in collectd too, so I've updated the patch to increase the buffer.

@PhilippWendler, thank you for notifying about the stack smashing. I was having the same issue on my machine too, but in my case it was silent. Probably lack of hardening during compilation, I don't know.

eero-t:
Please update "changeLog" entry also to something like:
> ChangeLog: smart: stop spurious increase of "Error Information Log Entries" NVMe counter and fix crashes with >=5.19 kernels

IMHO commit message could also refer the relevant spec, and the section which tells that:
* If nsid is not supported for given command, value 0h should be specified for it
* Only subset of commands supporting nsid, support the  (FFFFFFFFh) broadcast value
* In all other cases a valid nsid should be specified

(If you could also dit out from spec in which of the above 3 categories all commands used by smart plugin belong to, that would be great.)

PhilippWendler:
I tried the current state of this PR, but I still get stack smashing reported. I am applying this patch on top of the Ubuntu package with version 5.12.0-10ubuntu0.1 and running it on kernel 5.15.0-75-generic of Ubuntu. I am not sure what is different now compared to a few days ago when increasing the buffer size to 4096 bytes seemed to work for me. Sorry if this is not helpful, but I still wanted to leave this feedback here. If there is another patch to test or so I would be glad to do this.

eero-t:
Could you run it under valgrind to get backtrace where it's crashing?  Preferably with non-stripped version of the plugin + collectd.

If that doesn't trigger the issue, you could try running it under gdb, and provide backtrace from that ("bt" command).

PhilippWendler:
I now have a log, created with debug log enabled, valgrind, and debug symbols. I have disabled all plugins except for logging and smart, and configured the latter to ignore all SSDs except one. The log is here: [collectd.log](https://github.com/collectd/collectd/files/11866442/collectd.log)

In this system I have 4 SSDs, two from Samsung, and two from Intel. The Samsung SSDs are affected by what this PR is supposed to fix, and this fix seems to work. However, with f9f0f3d0534b87bac57610768e85eaaa97d47c1a I get the stack smashing, but only if the smart plugin reads from one of my Intel SSDs (in the log, this is `nvme3`, it is a SSDPE2KE064T8 model).

Valgrind points to `smart_read`, but this function is not changed by this PR. But it calls `get_vendor_id` via `smart_handle_disk`, and this is where the change is. Though to me the code still looks correct. But note that valgrind complains about some uninitialized values somewhere else.

PhilippWendler:
How can we proceed with this? I am not familiar with this code and by glancing over it I did not find any obvious problem.

i-rinat:
@eero-t,

I've updated changelog entry according to your suggestions, but tweaked "fix" into "prevent". Wrong NSID used in `get_vendor_id()` prevented further code execution, so there was never buffer overrun issues. They only surface up after the original fix in this pull request. I think "prevent" fits more.

There are now two patches, one for the buffer overrun workaround, and other for the NSID fix.
The workaround is to pass artificially not aligned buffers. This causes kernel to use what they call bounce buffer for the DMA. Data are then copied out to the user space with boundary checking. Ugly, but works. I think this is the best way to fix the bug, but I'm open to suggestions.

@PhilippWendler,

At some point I've realized that the bug in `get_vendor_id()` was a guard that prevented other NVMe ioctl calls from happening. That explains what you've seen.

I've added workaround to all three NVMe ioctl calls, and that should cover the specific code branch for Intel drives. Could you please check if the newer version still crashes? None of the NVMe drives I can borrow are manufactured by Intel, so it's hard for me to verify.

i-rinat:
> are there e.g. stack array accesses where the index value comes from outside the plugin? Which do not check that the access is within the array?

I believe it happened that way:
* user mode application calls nvme ioctl(), passes a buffer address and (relatively short) buffer length to the kernel;
* kernel maps user pages, obtains an address;
* kernel passes command with the buffer address to the NVMe drive;
* drive DMAs data into the buffer without checking for boundaries;
* kernel resumes user mode application;
* application find that data beyond buffer are overwritten.

It's not an application fault, it's the kernel fault.

i-rinat:
One of the CI jobs fails. Can reproduce fail locally, and will look into that.

i-rinat:
Applied fixes (initialize some stack variables), that stopped the test from failing.

i-rinat:
I've looked around and found a pull request with another way to fix crashes: https://github.com/collectd/collectd/pull/4124.
It's much simpler approach, but I still think the trick I used should be safer. Sadly, there is no objective metric for "safer".

PhilippWendler:
@i-rinat Thank you very much! I tested collectd 5.12 with 2e4e9c328eeceec5e368f15d2e13bb62e8c283b0 and 8269d287bbbedd6d21502c54a3f958d83c78c09b applied and it does not crash anymore (Linux 5.15).

However, it still produces entries in the error log of the Samsung SSD:
```
.................
 Entry[ 0]   
.................
error_count     : 2426295
sqid            : 0
cmdid           : 0x801b
status_field    : 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag       : 0
parm_err_loc    : 0x4
lba             : 0
nsid            : 0xffffffff
vs              : 0
trtype          : The transport type is not indicated or the error is not transport related.
cs              : 0
trtype_spec_info: 0
.................
 Entry[ 1]   
.................
error_count     : 2426294
sqid            : 0
cmdid           : 0x8019
status_field    : 0x200b(INVALID_NS: The namespace or the format of that namespace is invalid)
phase_tag       : 0
parm_err_loc    : 0x4
lba             : 0
nsid            : 0xffffffff
vs              : 0
trtype          : The transport type is not indicated or the error is not transport related.
cs              : 0
trtype_spec_info: 0
```

I even added another patch on top that also uses `nsid` 0 for the `NVME_ADMIN_GET_LOG_PAGE` ioctl, but still the same. I do not understand how this happens, because `NVME_NSID_ALL` is not used at all anymore, and still it occurs in the error log. Can anyone else with Samsung SSDs also reproduce this?

i-rinat:
> However, it still produces entries in the error log of the Samsung SSD

Is there any chance that you had two copies of `collectd` running?

PhilippWendler:
No. I only ever start collectd from a systemd service. And I just reconfirmed it (with the same binary from my last commit): no errors being produced - enable smart plugin for SSDs in config - restart collectd service - errors appear in log.

i-rinat:
I use Samsung SSD, but I have a machine with only single NVMe SSD in it. Maybe having more than one somehow changes things? I can't imagine how it can affect how the code works, though.

My hypothesis is that you are somehow running some other version of code. Maybe you have `PluginDir` defined in the configuration and it points to an older plugins?

You can verify that the right plugin is loaded by inspecting `maps` file for the collectd process:
```
# grep smart /proc/$(pgrep ^collectd$)/maps
```

PhilippWendler:
> My hypothesis is that you are somehow running some other version of code. Maybe you have `PluginDir` defined in the configuration and it points to an older plugins?

All my testing involves building a `.deb` package for collectd and using that to install and run collectd. On the system where I test it, I do nothing except (re-)installing the package and changing the plugin config, so there is certainly no old copy of the plugin lying around. And I just confirmed that by checking `maps` as you suggested. 

I also now tried again with a completely fresh build environment, built collectd 5.12 with the two patches, and installed it. Same symptoms. :-(

eero-t:
> The regression appeared around Linux 5.19, and there are multiple kernel versions out there with that behavior. Even if that gets fixed, it's better to be on the safe side in collectd too, so I've updated the patch to increase the buffer.

Were you able to narrow down the kernel commit introducing the issue?   Is there upstream kernel bug about that, and if not, could you file one?

octo:
I'm a bit hesitant to apply a work-around for what seems to be a quite severe kernel issue. How widespread is this problem? E.g. which kernel versions are affected and are the kernels of any large distro affected by this? [debian#998166](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=998166) mentions Sid – did this problem make it into Debian 12 ("Bookworm")? Has this been fixed upstream?

It'd be great if we could point to an issue on the Linux kernel side for tracking, or create one if it doesn't exist. Who could take care of this?

PR #3975, opened 2022-02-14
Fix CI failures caused by unsupported distros and updates to dependencies

elfiesmelfie:
* Remove Ubuntu/trusty
* Add Ubuntu/Focal and Ubuntu/Bionic
* Let ``make check`` fail on definite memory leaks only


mrunge (approved):
Awesome, thank you!

PR #3976, opened 2022-02-14
[ci][gha] Miscellaneous improvements and sync with cirrus

elfiesmelfie:
* Add "silent" + "keep going" options to MAKEFLAGS
       * Add $MAKEFLAGS to all make invocations
       * Update "name"s
       * Remove ``continue-on-error`` that was temporarily managing some failures

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,18 +44,18 @@ jobs:
     - name: Generate configure script
       run:
         ./build.sh
-    - name: configure
+    - name: Run configure script
       run: ./configure $CONFIGURE_FLAGS
-    - name: Make
+    - name: Build collectd
       run: make
-    - name: make check
+    - name: Run make check
       continue-on-error: true
       run: make check
-    - name: install bzip2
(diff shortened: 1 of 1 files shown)


mrunge (approved):
Thank you, looks good to me

.github/workflows/build.yml
```diff
@@ -48,24 +48,20 @@ jobs:
...
+      run: make $MAKEFLAGS
+    - name: Run make check
+      run: make $MAKEFLAGS check
     - name: Dump test logs
       run: |
           cat ./test-suite.log || true
-    - name: install bzip2
-      continue-on-error: true
-      run: |
+    - name: Install bzip2
```
mrunge, line 60: what about installing bzip2 in the used containers?

elfiesmelfie: That's a possibility. It'll take a while to update all the containers though, the changes are trivial, but the process is tedious)
But it should save a little bit of time in the CI runs.

I can update this PR to remove these, and then it'll become dependant on the new container changes.

PR #3984, opened 2022-03-15
smart: check udev_enumerate_scan_devices() return value and unify log messages

feckert:
This change checks the return value of the function and cancels the call
if the returned integer is not greater than or equal to 0.

ChangeLog: smart: check udev_enumerate_scan_devices() return value and unify log messages

### DIFF
--- a/src/smart.c
+++ b/src/smart.c
@@ -127,7 +127,12 @@ static int create_ignorelist_by_serial(ignorelist_t *il) {
   }
   udev_enumerate_add_match_subsystem(enumerate, "block");
   udev_enumerate_add_match_property(enumerate, "DEVTYPE", "disk");
-  udev_enumerate_scan_devices(enumerate);
+
+  if (udev_enumerate_scan_devices(enumerate) < 0) {
+    WARNING("smart plugin: udev scan devices failed");
+    return -1;
+  }
+
   devices = udev_enumerate_get_list_entry(enumerate);
   if (devices == NULL) {
     ERROR("udev returned an empty list deviecs");
@@ -611,7 +616,12 @@ static int smart_read(void) {
   }
   udev_enumerate_add_match_subsystem(enumerate, "block");
   udev_enumerate_add_match_property(enumerate, "DEVTYPE", "disk");
-  udev_enumerate_scan_devices(enumerate);
+
+  if (udev_enumerate_scan_devices(enumerate) < 0) {
+    WARNING("smart plugin: udev scan devices failed");
+    return -1;
+  }
+
   devices = udev_enumerate_get_list_entry(enumerate);
   if (devices == NULL) {
     ERROR("udev returned an empty list deviecs");
--- a/src/smart.c
+++ b/src/smart.c
@@ -47,6 +47,7 @@
 #define O_RDWR 02
 #define NVME_SMART_CDW10 0x00800002
 #define SHIFT_BYTE_LEFT 256
+#define PLUGIN_NAME "smart"
 struct nvme_admin_cmd {
   __u8 opcode;
   __u8 rsvd1[3];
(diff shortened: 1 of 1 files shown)


eero-t:
@feckert The change itself looks good, but as you are not using variable "r" for anything, you could drop it:
`if (udev_enumerate_scan_devices(enumerate) < 0) {`

@elfiesmelfie / @mrunge  Could you look into getting rid of the new CI issues before that?
```
Build / experimental (fedora_rawhide_x86_64) (pull_request) Failing after 3m
...
Run yum install -y bzip2 || apt install -y bzip2
Fedora 34 openh264 (From Cisco) - x86_64        335  B/s | 2.5 kB     00:07    
Fedora - Modular Rawhide - Developmental packag 204 kB/s | 1.6 MB     00:08    
Fedora - Rawhide - Developmental packages for t 6.9 MB/s |  64 MB     00:09    
Module yaml error: Document type modulemd-obsoletes unknown. [line 1629 col 1]
terminate called after throwing an instance of 'libdnf::ModulePackageContainer::ResolveException'
  what():  Failed to update from string: Unexpected YAML event in document stream [line 1630 col 1]
/__w/_temp/d0a1e83f-f19f-480f-bb84-0c3e1df7315b.sh: line 1: 154099 Aborted                 (core dumped) yum install -y bzip2
/__w/_temp/d0a1e83f-f19f-480f-bb84-0c3e1df7315b.sh: line 1: apt: command not found
Error: Process completed with exit code 127.
```
And:
```
Build / experimental (sid_amd64) (pull_request) Failing after 5m
...
FAIL: test_plugin_netlink
PASS: test_plugin_network
PASS: test_plugin_pcie_errors
FAIL: test_plugin_smart
```

feckert:
@eero-t thanks for your review. I have updated the first commit with your suggestion.
I also have push an other commit, which unifies the log messages.

eero-t:
> @eero-t thanks for your review. I have updated the first commit with your suggestion. I also have push an other commit, which unifies the log messages.

I see, you've added "smart_plugin" prefix to all log messages, fixed typos and dropped trailing dots, which all seem OK.

Btw.  Several plugins use a define for the plugin name, to avoid accidental typos in them (as compiler would catch mistyped defines):
```
src/logparser.c:#define PLUGIN_NAME "logparser"
...
src/logparser.c:      ERROR(PLUGIN_NAME ": Could not allocate memory");
...
src/logparser.c:      plugin_register_init(PLUGIN_NAME, logparser_init);
```

As I think you have changed all ERROR() messages, you might consider doing that change too.  But I'm fine either way.

feckert:
@eero-t Thanks for the advice, it is definitely better to do this via the preprocessor.

eero-t:
> it is definitely better to do this via the preprocessor

Register calls are still using "smart" string instead of a PLUGIN_NAME define which is a bit inconsistent, but probably best leave that as-is for now, as it's code churn to a different part of plugin functionality.

[1] Btw. according to Wiki doc, ChangeLog entry should be one line: https://github.com/collectd/collectd/blob/master/docs/CONTRIBUTING.md#changelog

@elfiesmelfie can ChangeLog handler accept multiple entries from the first comment, or does it need to be a **single** one-line ChangeLog comment?

In any case, probably best to have both PR title and single ChangeLog entry reflecting the new changes, something like: "smart: check udev_enumerate_scan_devices() return value and unify log messages".

mrunge:
The ChangeLog entry has to be one single line and it is also limited in length (70 chars)? Unfortunately, it just fails without response when it fails. Especially it doesn't tell why it's failing

feckert:
@eero-t @mrunge thanks for your review.
All suggestions were taken into account.

Except for the suggestion to use the preprocessor #define PLUGIN_NAME in the module_register() function. I looked in the [repository](https://github.com/collectd/collectd/search?q=plugin_register_read) and I saw that it is not clear to do it this way. Maybe in a next pullrequest.

mrunge (approved):
Thank you, looks good to me.

PR #3954, opened 2021-12-08
Install a dummy SIGPIPE handler instead of ignoring the signal

efuss:
Instead of ignoring `SIGPIPE`, install a dummy signal handler for it.
That way, on `exec()`, default behaviour will be restored which may be what external programs (e.g. those run via the _exec_ plugin) implicitly expect.
Most prominently, in case of collectd exiting ungracefully, long-running _exec_ plugins will receive `SIGPIPE` on the next try to send output to the now defunct collectd process.

ChangeLog: SIGPIPE handling has been changed from ignoring to installing a dummy handler so that exec plugins will be run with default SIGPIPE handling.

PR #3990, opened 2022-03-31
Add an ipfilter plugin

efuss:
Add an ipfilter plugin to monitor IPFilter statistics a la `ipfstat -s`.

Only tested on NetBSD.
ChangeLog: An ipfilter plugin to monitor IPFilter statistics a la ipfstat -s has been added.

### DIFF
--- a/src/ipfilter.c
+++ b/src/ipfilter.c
@@ -190,6 +190,7 @@ static int ipfilter_config(const char *key, const char *value) /* {{{ */ {
       for (r = report_tab, m = 1; r->name; r++, m <<= 1) {
         if (m == 0) {
           ERROR("ipfilter plugin: too many reports");
+          free(v);
           return 1;
         }
         if (strcmp(w, r->name) == 0) {
@@ -203,6 +204,7 @@ static int ipfilter_config(const char *key, const char *value) /* {{{ */ {
         WARNING("ipfilter plugin: unknown report %s", w);
       }
     }
+    free(v);
   } else
     return -1;
 


efuss:
> `Please run: contrib/format.sh src/ipfilter.c`

This will re-order the `#include`s so it won't compile (`netinet/in.h` needs `sys/time.h`) and makes the code much harder to read (removing TABs from `struct report report_tab[]`).

efuss:
Wow, it configures and builds on Solaris! Can someone test whether it works?

eero-t:
AFAIK new plugins are preferred for collectd-6.0 branch.  That is pretty far from release though...

efuss:
> AFAIK new plugins are preferred for collectd-6.0 branch.
A-ha.

As I'm using pkgsrc, I just have this and three more patches (SIGCHLD handling: #3526; SIGPIPE handling: #3954; NFS report narrowing: #3854) in LOCALPATCHES and tend to forget about them.

I guess quite a few people would benefit from the signal handler fixes, though.

> That is pretty far from release though...
I'm afraid IPFilter may be mostly out of business until then.

eero-t:
On a closer look, plugin sources are pretty small, so it could be OK for "main" too (I'm not collectd maintainer, just author of another plugin).

You say you've tested this only on NetBSD, but can you tell on which other OSes / OS versions ipfilter should work?  Should it e.g. work on all the other BSDs, include Mac?   If also on Linux, could you test it on some Linux version?

mrunge:
Thank you for proposing this. I'd say: let's get it in. It'd need another run on clang-format.
 Please run: contrib/format.sh src/ipfilter.c

eero-t:
Try disabling clang-format for the includes (with a comment explaining why), and re-enabling it after that:
https://stackoverflow.com/questions/24974919/how-to-mark-a-region-so-clang-format-wont-touch-it

efuss:
So I did run` clang-format --style=file`, but CI still complains. I guess CI is using a style file different from the one found in the repository.

efuss:
```
/usr/bin/docker pull collectd/ci:el9_x86_64 [16] (https://github.com/collectd/collectd/runs/6832077531?check_suite_focus=true#step:2:19)
Error response from daemon: manifest for collectd/ci:el9_x86_64 not found: manifest unknown: manifest unknown
```
I guess I'm not responsible for that one.

```
FAIL: test_plugin_virt
```
Or that one etc.

mrunge:
the el9 container test ran before the container was published on docker hub. That should be fixed by now, the container is uploaded.

the virt issue is known, the root-cause is not. 

Both issues are totally unrelated to this patch here.

PR #3995, opened 2022-04-05
snmp plugin: Allow FilterValues to match OIDs.

minfrin:
ChangeLog: snmp plugin: Allow FilterValues to match OIDs.

Where the value of a FilterOID is itself an OID, teach FilterValues
to parse the OID for comparison.

Add config example allowing RAM to be filtered from HOST-RESOURCES-MIB.

### DIFF
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -1264,12 +1264,15 @@ static int csnmp_strvbcopy(char *dst, /* {{{ */
     src = (char *)vb->val.string;
   else if (vb->type == ASN_BIT_STR)
     src = (char *)vb->val.bitstring;
-  else if (vb->type == ASN_IPADDRESS) {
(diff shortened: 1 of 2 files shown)


src/snmp.c
```diff
@@ -1264,12 +1264,17 @@ static int csnmp_strvbcopy(char *dst, /* {{{ */
     src = (char *)vb->val.string;
   else if (vb->type == ASN_BIT_STR)
     src = (char *)vb->val.bitstring;
-  else if (vb->type == ASN_IPADDRESS) {
```
vincele, line 1267: Don't remove the braces, yes there is only one statement, but it is multiline...
And also add braces to the below hunks too

minfrin: Can you confirm if this is ok?

minfrin:
Quick ping on this one, is this all ok?

PR #4006, opened 2022-05-25
mmc: add more vendor specific and generic data sources

hnez:
ChangeLog: mmc plugin: add support for generic and sandisk/micron health info

This PR extends the mmc plugin introduced by @feckert in #3887 with the following features:

- Support for the generic health metrics included in the eMMC 5.0 (and later) specification.
- Support for vendor-specific health metrics for micron and sandisk eMMCs.

The new vendor-specific health metrics are read out via `ioctl`s on the `/dev/mmcblk?` devices, while all other metrics are read out from the sysfs. Getting the respective device node for a path in sysfs is a non-trivial task that is solved quite nicely by `libudev`, which is why this PR also ports the plugin to use `libudev`.

I've tested the patch with sandisk and micon eMMCs, but did not have any swissbit devices at hand, so it would be nice if @feckert could tune in and check if everything still works there.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1491,8 +1491,9 @@ endif
 if BUILD_PLUGIN_MMC
 pkglib_LTLIBRARIES += mmc.la
 mmc_la_SOURCES = src/mmc.c
-mmc_la_LDFLAGS = $(PLUGIN_LDFLAGS)
-mmc_la_LIBADD = libignorelist.la
+mmc_la_CPPFLAGS = $(AM_CPPFLAGS) $(BUILD_WITH_LIBUDEV_CPPFLAGS)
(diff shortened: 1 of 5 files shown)


feckert:
Thanks for the hint, I'll have to have a look on Monday.

feckert:
I have pushed your commit on top of collectd 5.12 which is used in the [openwrt package master branch feed](https://github.com/openwrt/packages/tree/master/utils/collectd/patches). I have also [backported](https://github.com/collectd/collectd/commit/f76de1408cf3d61fbd5ce08363e641dbae384105) my initial mmc commit to this collected version .
So I could push your commit on top.

I have build this with openwrt master and tested this on an APU3 with my swissbit mmc card.
And from my point of view your changes have no effect on my swissbitt card.
It still works as expected.
Therefore :+1:

hnez:
@feckert very nice, thank you for testing!

mrunge (approved):
Thank you very much! Awesome!

PR #4009, opened 2022-05-31
modbus: add support for double values

LadyNamedLaura:
ChangeLog: Modbus plugin: support for double values.

Some devices (like the Siemens SENTRON PAC3200), expose some values as double (64-bit IEEE 754 floating point value).

This patch adds support for reading those.


eero-t:
@mrunge @elfiesmelfie CI seems broken again in "main" branch.  Rawhide test fails to package issues and Sid test fails with Smart plugin in multiple PRs (not touching Smart plugin).  Could you look what's the issue?

mrunge (approved):
thank you

PR #4019, opened 2022-06-15
Fix-configure.ac: define PREFIX in config.h

utoni:
ChangeLog: build: define PREFIX for later use.

ChangeLog: PREFIX was never defined and therefore set to the
default value `/opt/collectd`. collectd searched in
this path for desired files e.g. typesdb, plugins, etc
no matter if it was configured with `--prefix`.

Replaced `&(int){44}` with `(void *)((size_t)-1)` so
GCC-12 won't complain anymore.

Signed-off-by: lns <matzeton@googlemail.com>

Edit: @mrunge adjust ChangeLog (only one line and a limited number of characters allowed).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -145,6 +145,10 @@ if test "x$ac_system" = "xAIX"; then
   AC_DEFINE([_THREAD_SAFE_ERRNO], [1], [Define to use the thread-safe version of errno under AIX.])
 fi
 
+if test "x${prefix}" != xNONE; then
+  AC_DEFINE_UNQUOTED([PREFIX], ["${prefix}"], [Define the install prefix.])
+fi
+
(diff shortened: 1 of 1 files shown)


eero-t:
Connection state change is stale, that's been already fixed (in both v5 & v6 branch), and your commit include spurious white-space change.  Could you update the pull request?

eero-t:
You still have the obsolete connection_state change in your commit.  Please rebase to *latest* collectd code, otherwise your commit won't apply.

eero-t:
I think PR changelog entry needs to be single line.

@mrunge, any comments on this PREFIX change?

mrunge (approved):

mrunge:
Thank you!

PR #4020, opened 2022-07-08
Correct description of FQDNLookup in manpage

bionade24:
ChangeLog: collectd.conf manpage: Correct description of FQDNLookup option.

 ...of collectd.conf. FQDNLookup uses getaddrinfo() and not gethostname(). 

The `FQDNLookup false` check returns after the hostname was set by using `gethostname()`.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -330,8 +330,8 @@ hostname will be determined using the L<gethostname(2)> system call.
 
 If B<Hostname> is determined automatically this setting controls whether or not
 the daemon should try to figure out the "fully qualified domain name", FQDN.
-This is done using a lookup of the name returned by C<gethostname>. This option
-is enabled by default.
+This is achieved by using C<getaddrinfo()> to look up full web address of the
+first network interface that has one. This option is enabled by default.
 
 =item B<PreCacheChain> I<ChainName>
 


ChangeLog
```diff
@@ -1,4 +1,5 @@
 2020-09-03, Version 5.12.0
+	* collectd.conf manpage: Correct description of FQDNLookup option.
```
mrunge, line 2: The changelog file will be updated later for the release. You don't have to update it here. The next version would be 5.13.

src/collectd.conf.pod
```diff
@@ -330,7 +330,7 @@ hostname will be determined using the L<gethostname(2)> system call.
 
 If B<Hostname> is determined automatically this setting controls whether or not
 the daemon should try to figure out the "fully qualified domain name", FQDN.
-This is done using a lookup of the name returned by C<gethostname>. This option
+This is done using the value of struct addrinfo->ai_canonname returned by  C<getaddrinfo()>. This option
```
mrunge, line 333: I'd believe, the target audience for the config file won't understand, what this means exactly. Would you be able to rephrase? What does a struct addrinfo ... mean for the end user (that is probably someone using collectd to monitor something).

mrunge:
Thank you for picking this up. I have added a few thoughts in the review.

eero-t:
@bionade24 Could do the updates proposed by mrunge?    Changelog entry is needed in PR (not in ChangeLog file) to pass CI tests.

src/collectd.conf.pod
bionade24: I have rephrased it & also enhanced it to make clear it uses the first FQDN it can get from a network interface. I used the term "web address" as it's hard for me to describe behaviour without  repeating the previous sentence that it looks up the FQDN.

mrunge (approved):
Thank you

bionade24:
I guess the CI tests fail for a reason unrelated to my PR, don't they?

bionade24:
I guess the CI tests fail for a reason unrelated to my PR, don't they?

eero-t:
> I guess the CI tests fail for a reason unrelated to my PR, don't they?

Yes, your change cannot affect it.  CI breaking due to reasons outside of collectd project has been a recurring issue. :-/

Currently its virt plugin failing on RedHad distros, Smart plugin breaking on Debian testing (rolling distro) and multiple things breaking on Fedora Rawhide (rolling distro with new kernel headers + python includes).

eero-t:
getaddrinfo() is sure used in more files than gethostname():
```
$ git grep 'getaddrinfo *(' | grep -F .c: | wc -l
48
$ git grep 'gethostname *(' | grep -F .c:
src/collectdctl.c:    if (gethostname(hostname, sizeof(hostname)) != 0) {
src/daemon/collectd.c:  if (gethostname(hostname, hostname_len) != 0) 
```

@mrunge Few line doc change cannot be cause for build failures in plugins, so this is safe to merge.

mrunge:
Ah, you are right, I don't know why I didn't merge this long ago. Thank you for the patch!

PR #4022, opened 2022-07-12
EPICS plugin

matwey:
This is initial implementation for EPICS plugin. EPICS is Experimental Physics and Industrial Control System. It is widely used in particle physics and looks like message bus. In order to monitor and record the data distributed over the bus, the plugin is introduced.

Testing considerations:
There is package for debian sid with EPICS https://packages.debian.org/source/sid/epics-base so collectd images can be updated to test building of this plugin.

ChangeLog: Introduce EPICS plugin

Reference: https://epics-controls.org/

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3318,6 +3318,51 @@ at most B<16384> to prevent typos and dumb mistakes.
 
 =back
 
+=head2 Plugin C<epics>
+
+The I<epics plugin> collects data from EPICS (Experimental Physics and
+Industrial Control System) message bus.
+
+Note that in addition to the plugin configuration, the host OS has to be
+configured to be part of the message bus: caRepeater daemon is up and running
+and appropriate EPICS environment variables are set, if required. Consult EPICS
+documentation for details.
+
+B<Synopsis:>
+
+ <Plugin epics>
+   <Variable "Cavity1:Temp">
+     Type "gauge"
+   </Variable>
+   Label "Cavity1:Name"
+ </Plugin>
+
+B<Options:>
+
+=over 4
+
+=item E<lt>B<Variable> I<PVName>E<gt>
+
+The plugin configuration consists of multiple E<lt>B<Variable>E<gt> blocks,
+each per monitored EPICS Process Variable (PV). The variables are constantly
+monitored and their values are latched once per I<global Interval>.
+
+=item B<Type> I<Type>
+
+This is mandatory option to specify internal collectd type for this particular
+PV. PV data type announced by EPICS and the I<Type> specified here must be in a
+reasonable agreement.
+
+=item B<Label> I<PVName>
+
+Additionally, string-typed PV can be monitored as a B<Label>. For instance,
+this can potentially be useful for tracking an experiment ID. Note that in
+addition to configuring B<Label> the appropriate label-avare collectd writing
+plugin has to be enabled. For instance, I<write_prometheus plugin>,
+I<write_influxdb_udp plugin>, or I<write_http plugin> with InfluxDB B<Format>.
+
+=back
+
 =head2 Plugin C<ethstat>
 
 The I<ethstat plugin> collects information about network interface cards (NICs)
--- a/src/epics.c
+++ b/src/epics.c
@@ -1,0 +1,527 @@
+/**
+ * collectd - src/epics.c
+ * Copyright (C) 2022       Matwey V. Kornilov
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; only version 2 of the License is applicable.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ *
+ * Authors:
+ *   Matwey V. Kornilov <matwey.kornilov at gmail.com>
+ **/
+
+#include "collectd.h"
+#include "plugin.h"
+#include "utils/common/common.h"
+
+#include <cadef.h>
+
+struct pv {
+  char *name;
+
+  chtype ch_type;
+  chid id;
+  evid eid;
+
+  union {
+    value_list_t value;
+    char *label;
+  };
+
+  bool is_active;
+  bool is_label;
+};
+
+static struct {
+  struct pv *pvs;
+  int pvs_num;
+
+  pthread_mutex_t lock;
+  pthread_t thread_id;
+
+  bool thread_loop;
+} epics_plugin = {.lock = PTHREAD_MUTEX_INITIALIZER};
+
+static void free_pvs() {
+  for (int i = 0; i < epics_plugin.pvs_num; ++i) {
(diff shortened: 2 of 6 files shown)


eero-t:
I added bug about the CI errors (see above link) as they appear unrelated.

You're more likely to get these to v6 branch, as collectd developers have expressed preference for new plugins being added to the v6 branch instead of v5 one (avoids need for their later migration to v6, and keeping both in sync, as that's a problem already with current plugins).  @mrunge?

matwey:
@eero-t I would additionally provide 6.0 version of this plugin with great pleasure, but currently `ds_type` core machinery have not yet been ported to 6.0 and this prevents me to implement 6.0 version of the plugin.

eero-t:
> @eero-t I would additionally provide 6.0 version of this plugin with great pleasure, but currently `ds_type` core machinery have not yet been ported to 6.0 and this prevents me to implement 6.0 version of the plugin.

Would you consider looking into porting that?

(As you may have noticed, collectd is not very active any more, so any help would be appreciated.  I'm not myself collectd dev, just another person who first did v5 plugin, and had to port it to v6 to get it integrated.)

matwey:
> > @eero-t I would additionally provide 6.0 version of this plugin with great pleasure, but currently `ds_type` core machinery have not yet been ported to 6.0 and this prevents me to implement 6.0 version of the plugin.
> 
> Would you consider looking into porting that?

You are right. But I have no idea of how it is supposed to be in 6.0 version. Design decision is required here and then I can volunteer the implementation.

> 
> (As you may have noticed, collectd is not very active any more, so any help would be appreciated. I'm not myself collectd dev, just another person who first did v5 plugin, and had to port it to v6 to get it integrated.)

src/epics.c
```diff
@@ -0,0 +1,551 @@
...
+static int stop_thread(void) {
+  int ret = 0;
+
+  pthread_mutex_lock(&epics_plugin.lock);
+
+  if (epics_plugin.thread_loop == 0) {
+    goto epics_unlock;
+  }
+
+  epics_plugin.thread_loop = 0;
```
octo, line 384: You can skip the conditional and the goto. Just set the variable unconditionally, which has the same effect.

```suggestion
  epics_plugin.thread_loop = 0;
```

src/epics.c
```diff
@@ -0,0 +1,551 @@
...
+    handle_conn_down(p);
+    break;
+  }
+}
+
+static void *epics_thread(void *args) {
+  int i = 0;
+  long ret = 0;
+
+  ret = ca_context_create(ca_disable_preemptive_callback);
```
octo, line 305: Please declare variables as late as possible, ideally initializing them to their final value. Here and throughout.

```suggestion
  long ret = ca_context_create(ca_disable_preemptive_callback);
```

src/epics.c
```diff
@@ -0,0 +1,551 @@
...
+
+  if (ds == NULL) {
+    ERROR("epics plugin: Unknown type \"%s\" for channel \"%s\". See "
+          "types.db(5) for details.",
+          p->value.type, p->name);
+
+    return;
+  }
+
+  int i = 0;
```
octo, line 109: This declaration is out of place. Please declare `i` in the loop header, like so:

```c
for (size_t i = 0; i < values_len; ++i) {
```

Since `values_len` is a `size_t`, `i` should be of the same type. Otherwise compilers will complain about comparing signed and unsigned integers.

octo (commented):
Hey @matwey,

thank you so much for your PR! I've left some initial comments inline and would be happy to help you get this merged.

Best regards,
—octo

matwey:
> Hey @matwey,
> 
> thank you so much for your PR! I've left some initial comments inline and would be happy to help you get this merged.
> 
> Best regards, —octo

Thanks. I think, I've done with your comments and I've also rebased the PR on top of current `main`.

src/epics.c
```diff
@@ -0,0 +1,531 @@
...
+}
+
+static int epics_config(oconfig_item_t *ci) {
+  if (ci->children_num == 0) {
+    ERROR("epics plugin: No variables are specified");
+
+    return -1;
+  }
+
+  epics_plugin.pvs = calloc(ci->children_num, sizeof(struct pv));
```
octo, line 429: This should use `realloc` to allocate the memory, because the config callback can be called multiple times.

```c
struct pv *pvs = realloc(epics_plugin.pvs,
                         sizeof(*epics_plugin.pvs) * (epics_plugin.pvs_num + ci->children_num));
if (pvs == NULL) {
  return ENOMEM;
}
epics_plugin.pvs = pvs;
pvs = epics_plugin.pvs + epics_plugin.pvs_num;
memset(pvs, 0, sizeof(*pvs) * ci->children_num);
/* initialize elements in pvs[0 .. ci->children_num-1] here */
epics_plugin.pvs_num += ci->children_num;
```

octo (commented):
Hi @matwey, thanks again for your work. Overall this looks very good. I left a few more comments inline, mostly coding style and some minor memory management improvements. Once they are resolved this should be good to go.

Thanks and best regards
—octo

src/epics.c
```diff
@@ -0,0 +1,531 @@
...
+      ERROR("epics plugin: CA error %s occurred while trying to create channel "
+            "\"%s\"",
+            ca_message(ret), p->name);
+      ret = 1;
+      goto error;
+    }
+  }
+
+  while (epics_plugin.thread_loop != 0) {
+    ca_pend_event(2.0);
```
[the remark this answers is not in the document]
matwey: It is from the library examples. Basically it is a timeout similar to the timeout value for `select` syscall. So, this loop is iterated every two seconds when no events arrived.

src/epics.c
matwey: I didn't know about it. When does it happen? And what should be expected behavior when the childrens are different?

src/epics.c
```diff
@@ -0,0 +1,531 @@
...
+
+  return 0;
+
+error:
+  free_pvs();
+
+  return -1;
+}
+
+static int epics_init(void) { return start_thread(); }
```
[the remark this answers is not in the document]
matwey: It is clang-format who collapses the function.

matwey:
Ok, I think now I am done with the second part of comments.

src/epics.c
octo: Yeah, don't worry, this is not documented well and if it were, I think it's unreasonable to expect people to read the docs in this detail.

The callback is called multiple times when there are multiple `<Plugin epics>` blocks.

src/epics.c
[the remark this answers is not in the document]
octo: Okay, never mind then.

src/collectd.conf.pod
```diff
@@ -3318,6 +3318,18 @@ at most B<16384> to prevent typos and dumb mistakes.
 
 =back
 
+=head2 Plugin C<epics>
```
octo, line 3321: Can you please add 2-3 sentences on what this plugin does?

src/collectd.conf.pod
```diff
@@ -3318,6 +3318,18 @@ at most B<16384> to prevent typos and dumb mistakes.
...
+
+=over 4
+
+=item B<Variable> I<PVName>
+
+=item B<Type> I<Type>
+
+=item B<Label> I<PVName>
```
octo, line 3329: I'm not sure I'd be able to use the plugin based on this. Maybe copy the example from `collectd.conf.in` here as a "synopsis" and describe what effect that configuration would have? E.g. what would happen with the "label"?

octo (approved):
Code looks good now, thanks @matwey! I think the documentation needs a bit more content, at least an elevator pitch of what it does, then we're golden :grin:

matwey:
You are right. I added some documentation. I am not sure about wording though, so please say if something is missed or not clear enough.

octo (approved):
That's excellent, thank you so much!

Issue #2636, opened 2017-12-29
SNMP plugin needs a new schema


yellowpattern:
*   Version of collectd: 5.5.3
*   Operating system / distribution: CentOS 6

Sometimes it is possible to collect interface statistics using SNMP, in other instances it is done with collectd. Unfortunately the two methods do not present statistics in an aligned name space. To fix this, the SNMP plugin needs to be updated.

In the SNMP configuration, we can setup collecting octet counters like this:

    <Data "Octets">
        Table true
        Type "if_octets"
        #InstancePrefix "octets"
        Instance "MIBNAME::portName"
        Values "MIBNAME::ifInOctets" "MIBNAME::ifOutOctets"
    </Data>

What actually happens is that the SNMP plugin applies the "Instance" name to the "type_instance" field rather than the "instance" field.

What I'd like to suggest is the addition of a new per-Data boolean option "NewSchema" that when set to true, causes collectd to use "instance" rather than "type_instance", will cause "InstancePrefix" to be ignored and add a new configuration option, "TypeInstance" that allows the "type_instance" field to be set.

The goal here is to make it easier to re-use queries and graphs between interface and SNMP statistics. The "type_instance" change is to introduce another 'variable' to use with grafana when doing "GROUP BY".

faxm0dem:
we should probably add support for metadata to snmp

rpv-tomsk:
We can allow to override reported `plugin` value on per-Data basis,
and can add an option `PluginInstance` in addition to existing `Instance`.

All this allow to unify metrics from `interface` and `snmp` plugins.

rpv-tomsk:
This feature was implemented. If you still interested, you can do tests.
Please inform us if you will do this, we will wait for feedback then.

PR #4023, opened 2022-07-12
Update the format of contrib/snmp-data.conf

vitkabele:
ChangeLog: Update the format of snmp-data.conf file to not issue deprecation warning on startup.

s/InstancePrefix/TypeInstancePrefix/g
s/Instance/TypeInstanceOID/g

Signed-off-by: Vit Kabele <vit@kabele.me>

### DIFF
--- a/contrib/snmp-data.conf
+++ b/contrib/snmp-data.conf
@@ -6,37 +6,37 @@
     <Data "ifmib_if_octets32">
     	Type "if_octets"
 	Table true
-	Instance "IF-MIB::ifDescr"
+	TypeInstanceOID "IF-MIB::ifDescr"
 	Values "IF-MIB::ifInOctets" "IF-MIB::ifOutOctets"
     </Data>
     <Data "ifmib_if_octets64">
     	Type "if_octets"
 	Table true
-	Instance "IF-MIB::ifName"
+	TypeInstanceOID "IF-MIB::ifName"
 	Values "IF-MIB::ifHCInOctets" "IF-MIB::ifHCOutOctets"
     </Data>
     <Data "ifmib_if_packets32">
     	Type "if_packets"
 	Table true
-	Instance "IF-MIB::ifDescr"
+	TypeInstanceOID "IF-MIB::ifDescr"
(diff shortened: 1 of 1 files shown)


eero-t:
In which version the new name was added, and old one deprecated?    And are there still maintained distro versions, which would include old enough version not to supporte the new name?

vitkabele:
Hello,
the deprecation warning (and replacement options were added in b17fa8a7c71eb9fbc67702d8f1cf8a53fb2fe8a7, related to #2636. Released on June 13, 2019 as [collectd-5.9.0 (c829100a9f7ffc2dfa2c38489bc4df4b23316930)](https://github.com/collectd/collectd/releases/tag/collectd-5.9.0).

As for the maintained distro versions. Is there a way how to determine this other than going manually through all different repos?
I only know that Debian does not include this file in the basic installation anyway.

eero-t:
> As for the maintained distro versions. Is there a way how to determine this other than going manually through all different repos? I only know that Debian does not include this file in the basic installation anyway.

Sorry for the late reply.   There are some things from which checking the package versions is a bit more convenient:
* https://ftp.redhat.com/pub/redhat/linux/enterprise/  (RHEL6 - RHEL9 are ones currently maintained according to Wikipedia)
* https://www.debian.org/distrib/packages
* https://packages.ubuntu.com/
* https://src.fedoraproject.org/
* https://build.opensuse.org/
* https://svnweb.mageia.org/packages/
* https://packages.gentoo.org/

Oldest enterprise / LTS distro versions should have the oldest package versions, so I would check them first.

vitkabele:
Hello,
so apparently at least Ubuntu 18.04 is still out there using collectd 5.7. I however don't think that merging it to the head would somehow affect these distros right? It is not going to be installed there and if someone wants the version of the file related to the old release, it is always possible to take it from the respective tag.
The file on HEAD should however be in-sync with the code on HEAD i think.

mrunge (approved):
Thank you for your patch and sorry that this took so long

mrunge:
Any potential CI errors are not related to this patch

PR #4024, opened 2022-07-13
Fixed duplicate datarootdir in build.sh

bri-bear:
ChangeLog: In ./build.sh: fixed the --datarootdir duplicate in ./config section.


eero-t:
Are you sure you removed the correct duplicate?  I would have thought latter specified one to be in effect earlier, but you removed that one...

mrunge (approved):
thank you!

PR #4025, opened 2022-07-15
Use IOMainPort over deprecated IOMasterPort

ccaviness:
ChangeLog: Build system: Use IOMainPort over deprecated IOMasterPort

In macOS 12, `IOMasterPort`/`kIOMasterPortDefault` is deprecated in favor of `IOMainPort`/`kIOMainPortDefault`:

```
src/battery.c:250:7: error: 'kIOMasterPortDefault' is deprecated: first deprecated in macOS 12.0 [-Werror,-Wdeprecated-declarations]
      kIOMasterPortDefault, IOServiceNameMatching("battery"), &iterator);
      ^~~~~~~~~~~~~~~~~~~~
      kIOMainPortDefault
```

### DIFF
--- a/src/apple_sensors.c
+++ b/src/apple_sensors.c
@@ -51,20 +51,24 @@
 #include <IOKit/IOTypes.h>
 #endif
 
-static mach_port_t io_master_port = MACH_PORT_NULL;
+#if (MAC_OS_X_VERSION_MIN_REQUIRED < 120000) // Before macOS 12 Monterey
+#define IOMainPort IOMasterPort
+#endif
+
+static mach_port_t io_main_port = MACH_PORT_NULL;
 
 static int as_init(void) {
   kern_return_t status;
 
-  if (io_master_port != MACH_PORT_NULL) {
-    mach_port_deallocate(mach_task_self(), io_master_port);
-    io_master_port = MACH_PORT_NULL;
+  if (io_main_port != MACH_PORT_NULL) {
+    mach_port_deallocate(mach_task_self(), io_main_port);
+    io_main_port = MACH_PORT_NULL;
   }
 
-  status = IOMasterPort(MACH_PORT_NULL, &io_master_port);
+  status = IOMainPort(MACH_PORT_NULL, &io_main_port);
   if (status != kIOReturnSuccess) {
-    ERROR("IOMasterPort failed: %s", mach_error_string(status));
-    io_master_port = MACH_PORT_NULL;
+    ERROR("IOMainPort failed: %s", mach_error_string(status));
+    io_main_port = MACH_PORT_NULL;
     return -1;
   }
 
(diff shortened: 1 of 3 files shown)


eero-t:
Your change would be much smaller if you would do the define other way round as Monterey updates are still being released: https://en.wikipedia.org/wiki/MacOS_Monterey#Release_history

Any idea how low Apple still supports it?

ccaviness:
> Your change would be much smaller if you would do the define other way round as Monterey updates are still being released: https://en.wikipedia.org/wiki/MacOS_Monterey#Release_history

I was thinking that since the future, from macOS 12 on, the correct names are `IOMainPort`/`kIOMainPortDefault`, so I updated the code to to use the current correct names, and for machines running earlier versions just have the alias defined.

> Any idea how low Apple still supports it?

I don't follow - supports what? In macOS 12 and later, use `IOMainPort`; in macOS 11 and earlier, use `IOMasterPort`.

eero-t:
Sorry that was typo.  I meant: because next collectd v5 release may still be some time off, do you have any idea how long macOS 11 would still be supported by Apple?

If its official support is going away in year or two (I have no idea, as I have no mac), then moving to v12 names looks good, otherwise, smaller patch might be preferred.

(I'm not collectd developer, but as you can see by looking at open issue and PR list, reviewing them is a huge bottleneck.  Collectd lost a lot of steam when the original author left GitHub project unmaintained for long time before eventually granting other devs rights to fix things here.)

ccaviness:
> Sorry that was typo. I meant: because next collectd v5 release may still be some time off, do you have any idea how long macOS 11 would still be supported by Apple?
> 
> If its official support is going away in year or two (I have no idea, as I have no mac), then moving to v12 names looks good, otherwise, smaller patch might be preferred.

Apple doesn't really ever say when support has ended, they just ... stop providing any updates. Usually the current version gets feature updates over the first year, and then sporadic security updates for the next 2. But this is not guaranteed, nor does Apple actually document this anywhere. So the 3-year time is based on historical knowledge.

macOS 11 was released in November 2020, so it's likely that it will stop receiving updates in the next year and a half or so.

See https://eclecticlight.co/2021/09/22/how-long-does-apple-support-macos/ and  https://endoflife.date/macos for some details.

eero-t:
> macOS 11 was released in November 2020, so it's likely that it will stop receiving updates in the next year and a half or so.

Ok, based on that doing it this way seems fine.  @mrunge I think it would be good if collectd were to have Mac CI built test too.

eero-t:
> > macOS 11 was released in November 2020, so it's likely that it will stop receiving updates in the next year and a half or so.
> 
> Ok, based on that doing it this way seems fine. @mrunge I think it would be good if collectd were to have Mac CI built test too.

@mrunge ?

mrunge (approved):

mrunge:
Ah, sorry. This fell off my radar. Thank you for your PR.

Unfortunately, my time is quite limited right now, it would be great to have CI running on Macs, no question. But if it's blocked on me, it will probably wait for some time.

Issue #2873, opened 2018-07-24
network plugin swallowing samples?


mrunge:
*   Version of collectd: 5.8
*   Operating system / distribution: CentOS

## Actual behavior
We were using collectd-tg to create samples, like `collectd-tg -n 100 -p 1 -H 1 -i 300 -d dest_host -D 25826`
On the receivers side _dest_host_, there is the network plugin runing, and also the csv plugin configured to write out samples:

```
LoadPlugin csv

<Plugin csv>
    DataDir "/collectd/data/"
</Plugin>
```

Now, after sending the first batch of 100 samples, ctrl-c is pressed, to stop collectd-tg. Inspecting the output directory for collectd never shows the desired number of 100 samples/files. I usually get to 40 or 50 out of expected 100 samples. When using larger numbers, like 10k, you'll usually get 9970 (or so). 
If you wait a few intervals, the amount _n_ of values is reached, but not all csv files contain the same
number of samples.

For tests, make sure to stop collectd, clean out /collectd/data and start collectd again.
I've also tested with write_prometheus instead of write_csv, and did not reach the expected amount of metrics.

I'd be expecting some buffers to get filled and flushed when full; however, they should get flushed at some point anyways.

Any idea, what's causing this?

faxm0dem:
Remember the network plugin uses UDP. So at least two things can happen:

1. The datagrams reach dest_host not chronologically
2. Some datagrams get lost

You can easily test for 1. by searching the log for relevant messages. As for two, you could check the values of the `protocols` plugin on dest_host.

mrunge:
Right.
However in this test it doesn't matter, if datagrams are not getting in in chronological order, as long as they get in.
UDP is not that reliable, that is clear. I'd expect to loose one or the other package, but not ~ 50% when sending 100 samples.

faxm0dem:
It does matter because collectd will throw the late events away

mrunge:
`collectd-tg -n 100 -H 1 -p 100 -i 300 -d dest_host` creates in this test 63 directories,
some with 3 values, some with 2, some with 1, in total 96 files.

I would have expected 100 directories, all with 1 value.

faxm0dem:
The documentation states:

```
       -n num_vl
           Sets the number of unique value lists (VL) to generate. Defaults to 10000.
       -p num_plugins
           Sets the number of unique plugins to simulate. Defaults to 20.
```

I admit I'm not sure what to expect for instance when `-n = -p`

[...]

PR #4026, opened 2022-07-19
[collectd 6] src/daemon/plugin.c: Use one thread per write plugin

hnez:
ChangeLog: collectd: Use one write thread per write plugin

This PR is more or less an implementation of a changes proposed in #2486, but relevant discussion has also already happened in #1543, #2480, #2104 and #2873.
Our motivation behind implementing this draft is that we would like to use `collectd` network write plugins like MQTT over unreliable networks like LTE and, in addition to that, a write plugin that logs values to disk.
Ideally we would like to not loose values due to short network hickups and if that does not work due to overfull buffers, we want to at least rely on the values logged on disk.

The PR is marked as draft, as we:

1) Did not yet port the MQTT plugin to collectd 6 and peform actual tests in the wild.
2) Would like to discuss the design a bit more. One example is the handling of write callback return values, which are currently just ignored. Should e.g. `EAGAIN` be handled by waiting some time and calling the plugin again with the same value?

---

The previous write thread design used a single queue with a single read head from which one of the write threads would de-queue an element and would then sequentially call each registered write callback. This meant that all write plugins would have to cooperate in order to not drop values. If for example all write threads are stalled by the same write plugin's callback function not returning, the queue will start to fill up until elements start to be dropped, even though there are other plugins that could still make progress. In addition to that, all write callbacks have to be designed to be reentrant right now, which increases complexity.

This new design uses a single linked-list write queue with one read head per output plugin. Each output plugin is serviced in a dedicated write thread. Elements are freed based on a reference count, which is shown in the ASCII-Art below:

```
          +- Thread #1 Head       +- Thread #2 Head       +- Tail
          v                       v                       v
   +--+  +--+  +--+  +--+  +--+  +--+  +--+  +--+  +--+  +--+
   | 0|->| 1|->| 1|->| 1|->| 1|->| 2|->| 2|->| 2|->| 2|->| 2|->X
   +--+  +--+  +--+  +--+  +--+  +--+  +--+  +--+  +--+  +--+
    ^
    +- to be free()d
```

The changes introduced by this commit have some side-effects:

- The `WriteThreads` config option no longer exists, as a strict 1:1 ratio of  write plugins and write threads is used.
- The data flow has changed. The previous data flow was:
```
    (From one of the ReadThreads)
      plugin_dispatch_{values,multivalue}()
      plugin_dispatch_metric_family()
      enqueue_metric_family()
      write_queue_enqueue() -----{Queue}----+
                                            |
    (In one of the WriteThreads threads)    |
      plugin_write_thread()                 |
        ^- plugin_write_dequeue()         <-+
      plugin_dispatch_metric_internal()
        ^- fc_process_chain(pre_cache_chain)
      fc_process_chain(fc_process_chain)
      fc_bit_write_invoke()
      plugin_write(NULL) / plugin_write(plugin_name)
      plugin callback()
```
- … The data flow now is: …
```
    (From one of the ReadThreads)
      plugin_dispatch_{values,multivalue}()
      plugin_dispatch_metric_family()
      plugin_dispatch_metric_internal()
        ^- fc_process_chain(pre_cache_chain)
      fc_process_chain(post_cache_chain)
      fc_bit_write_invoke()
      plugin_write(NULL) / plugin_write(plugin_name)
      write_queue_enqueue() -----{Queue}----+
                                            |
    (In one of the WriteThreads threads)    |
      plugin_write_thread()               <-+
      plugin callback()
```
- …  One result of this change is, that the behaviour of `plugin_write` has changed from running the plugin callback immediately and in the same thread, to  always enqueueing the value and de-queing in the dedicated thread.
- The behaviour of the `WriteQueueLimitHigh` and `WriteQueueLimitLow` options has changed.  The Queue will be be capped to a length of `LimitHigh` by dropping random queue elements between the queue end and `LimitLow`. Setting `LimitLow` to a reasonably large value ensures that fast write plugins do not loose values, even in the vicinity of a slow plugin.  The diagram below shows the random element selected for removal (`###`) in  `Step 1` and the queue with the element removed in `Step 2`.
```
  Step 1:
    +- Thread #1 Head    |  +- Thread #2 Head       +- Tail
    v  |                 |  v                       v
   +--+| +--+  ####  +--+| +--+  +--+  +--+  +--+  +--+
   | 1|->| 1|-># 1#->| 1|->| 2|->| 2|->| 2|->| 2|->| 2|->X
   +--+| +--+  ####  +--+| +--+  +--+  +--+  +--+  +--+
       |                 |
       | LimitHigh       | LimitLow

  Step 2:
       |  +- Thread #1 Head +- Thread #2 Head       +- Tail
       |  v              |  v                       v
       | +--+  +--+  +--+| +--+  +--+  +--+  +--+  +--+
       | | 1|->| 1|->| 1|->| 2|->| 2|->| 2|->| 2|->| 2|->X
       | +--+  +--+  +--+| +--+  +--+  +--+  +--+  +--+
       |                 |
       | LimitHigh       | LimitLow
```


mrunge:
Thank you for working on this, it is an interesting proposal.

From experience with working with writes plugins over the network, I would like to consider all outputs over the net as being unreliable per se. (Some are more unreliable than others). Over time, you'll accumulate data which should have been sent out, but was not; ultimately, you may end up with storing data in a write queue, that will not be stored in the remote data store, because the timestamp is considered as being too old, and thus the data is discarded.

hnez:
Hi @mrunge,

> […] I would like to consider all outputs over the net as being unreliable per se. […]

Well, yes. It will always be kind of _best effort_ - there is nothing that can be done if the network is down long enough that output buffers start overflowing and UDP based plugins simply won't notice if measurements are being dropped.
But I would expect TCP based plugins to be mostly reliable over networks with reasonable down times and packet losses, like cellular internet.

> […] end up with storing data in a write queue, that will not be stored in the remote data store, because the timestamp is considered as being too old, and thus the data is discarded.

I think we may have different usecases in mind? We are currently developing two embedded devices and use collectd on both of them. The first one will be used indoors and will have a reliable ethernet connection to the storage backend, so we are fine with using the UDP based plugin and maybe loosing a measurement here and there.
The other one will however be deployed on masts around the city and have an LTE based uplink, which means a much less reliable/trustworthy network connection. For the second device we use the MQTT plugin via TLS, Telegraf and InfluxDB to store our data. For both devices we are not that interested in the most recent data, instead we want to be able to look at the data after a few months and decide if e.g. our thermal design is sound. Throwing away data because it arrived out of order or too late (or the clocks were out of sync) is not something we would do.

hnez:
Hi @eero-t,

> Are there some particular write plugins with which you have stall problems / are there issues in some of the plugins which you think should be separately fixed (e.g. in v5 branch) without your proposal?

I would not blame the write plugins. The MQTT plugin for example¹ tries to re-connect to the broker for each write callback call, if the connection was lost. This sounds like a reasonable thing to do to me, but means that each connection attempt will take some seconds in which the queue will not make progress.

¹ As far as I can tell, I did not look too deeply into it.

> […] but moving functions around (e.g. plugin_write) makes things slightly harder to review.

You are right, the diff is not that usable. The current `plugin_write` is based on a renamed `enqueue_metric_family`, hence the move. Now that I look at it I even forgot to update the function name in the end-of-function comment.
If there is enough interest in further developing this patch I will try and make the diff more readable in the next version.

> Build fails on Solaris due to: […] Undefined symbol stpcpy

Oh, I first assumed that that was a false positive. Will fix in the next version.

eero-t:
> Oh, I first assumed that that was a false positive. Will fix in the next version.

Note that there are few helper / wrapper functions in: https://github.com/collectd/collectd/blob/collectd-6.0/src/utils/common/common.h#L75

Btw. Have you checked whether your change has any measurable memory or performance impact on collectd?

mrunge:
> Hi @mrunge,
> 
> > […] I would like to consider all outputs over the net as being unreliable per se. […]
> 
> Well, yes. It will always be kind of _best effort_ - there is nothing that can be done if the network is down long enough that output buffers start overflowing and UDP based plugins simply won't notice if measurements are being dropped. But I would expect TCP based plugins to be mostly reliable over networks with reasonable down times and packet losses, like cellular internet.
> 
> > […] end up with storing data in a write queue, that will not be stored in the remote data store, because the timestamp is considered as being too old, and thus the data is discarded.
> 
> I think we may have different usecases in mind? We are currently developing two embedded devices and use collectd on both of them. The first one will be used indoors and will have a reliable ethernet connection to the storage backend, so we are fine with using the UDP based plugin and maybe loosing a measurement here and there. The other one will however be deployed on masts around the city and have an LTE based uplink, which means a much less reliable/trustworthy network connection. For the second device we use the MQTT plugin via TLS, Telegraf and InfluxDB to store our data. For both devices we are not that interested in the most recent data, instead we want to be able to look at the data after a few months and decide if e.g. our thermal design is sound. Throwing away data because it arrived out of order or too late (or the clocks were out of sync) is not something we would do.

Yes, for sure. I was mostly referring to use-cases, where you have collectd collecting metrics and events either in data centers and also on edge locations, both sending over the same transport mechanism.
However, and from experience, I have seen even in-house collectd growing to use 10th of gb of memory, because remote write endpoints were not as reliable as expected. That's what I meant: assume they can fail. If they don't: great, and if they do: you can already handle that.

hnez:
> Build fails on Solaris due to:
> 
> ```
> Undefined			first referenced
>  symbol  			    in file
> stpcpy                              src/daemon/collectd-plugin.o
> ld: fatal: symbol referencing errors. No output written to collectd
> ```

> Note that there are few helper / wrapper functions in: https://github.com/collectd/collectd/blob/collectd-6.0/src/utils/common/common.h#L75

It looks like none of those address the usecase of appending a string to an existing string without allocating new memory, but thank you for the tip. I've now replaced `stpcpy` with `strlen` and `memcpy`. That's also what `strjoin` in `common.c` does.
That seems to make the CI happy.

> Btw. Have you checked whether your change has any measurable memory or performance impact on collectd?

Not really. We can't do too much real world testing until we have ported the read/write plugins we use to collectd 6.
Do you have some particular test setup in mind that we should try to replicate?

The memory usage should stay mostly unchanged, as the maximum queue length is still bounded by `WriteQueueLimitHigh`. The queue element struct is slightly larger than before due to e.g. the reference count, but I do not expect that to have a relevant impact.

I see a potential performance impact if the difference between `WriteQueueLimitHigh` and `WriteQueueLimitLow` is large, as `write_queue_enqueue` traverses the linked list to find the element at `to_drop` to find the correct random element to drop.

hnez:
Hey @mrunge 

> Yes, for sure. I was mostly referring to use-cases, where you have collectd collecting metrics and events either in data centers and also on edge locations, both sending over the same transport mechanism. However, and from experience, I have seen even in-house collectd growing to use 10th of gb of memory, because remote write endpoints were not as reliable as expected. That's what I meant: assume they can fail. If they don't: great, and if they do: you can already handle that.

Well the main intention behind this PR is to reduce the impact of a failing write plugin on other plugins, not to make network based write plugins more reliable. We are fine with network write plugins loosing data if the queue grew too large. What we want to improve is the behavior of e.g. disk based plugins in the presence of a network write plugin that can not make any progress.
Currently one failing write plugin (one that just blocks when its callback function is called instead of returning immediately) can prevent all other write plugins from making progress, resulting in data loss for all plugins if the queue grew too large.

The PR should limit the impact of a failing write plugin to that plugin only, as each plugin has it's own queue head that it can make progress on.

eero-t:
>> Btw. Have you checked whether your change has any measurable memory or performance impact on collectd?
> 
> Not really. We can't do too much real world testing until we have ported the read/write plugins we use to collectd 6. Do you have some particular test setup in mind that we should try to replicate?
> 
> The memory usage should stay mostly unchanged, as the maximum queue length is still bounded by `WriteQueueLimitHigh`. The queue element struct is slightly larger than before due to e.g. the reference count, but I do not expect that to have a relevant impact.
>
> I see a potential performance impact if the difference between `WriteQueueLimitHigh` and `WriteQueueLimitLow` is large, as `write_queue_enqueue` traverses the linked list to find the element at `to_drop` to find the correct random element to drop.

I see.  That sounds OK.

To make sure the implementation is good, it would be nice if you could run tests under:
* ASAN (i.e. adding "-fsanitize=address" both to collectd compile and link flags):
  * https://github.com/google/sanitizers/wiki/AddressSanitizer
* Valgrind DRD/Helgrind (i.e. run collectd with "valgrind --tool=<drd|helgrind>") [1]:
  * https://valgrind.org/docs/manual/drd-manual.html
  * https://valgrind.org/docs/manual/hg-manual.html

For multiple writers, with use-cases where:
* None of the writers blocks
* One of the writers blocks
* All of the writers block

Do you have a setup, where you can test all of those 3 cases?

[1] Valgrind tools report _potential_ data races, so not all of them are real issues.  However, it's better to make locking simple enough that it does not fool tools, as automated checks for complex things like locking is really helpful in catching errors...

PS. If one suspects lock contention, only tools to analyze that without too much overhead (like Valgrind DRD would have) that I know of, is this:
* http://0pointer.de/blog/projects/mutrace.html
* https://packages.debian.org/buster/mutrace

eero-t:
@hnez Have you been able to run your changes under memory access & thread debugging tools?

hnez:
Hi @eero-t,

sadly not yet. I've been busy with other stuff the last couple of weeks and am currently at the Open Source summit in Dublin.
I do however plan to do so soon.

hnez:
Hi,

I've found some time to work on this PR again.

Setup
---

> Do you have a setup, where you can test all of those 3 cases?

Well now I do! I've hacked together a dummy plugin acts as both read and write plugin and produces measurements at a constant rate of one per second and consumes them at either a lower or greater average rate. In the former case this results in the queue slowly filling up until it overflows. In the latter case the queue will stay relatively short.
I've put the code in a hack commit [here](https://github.com/hnez/collectd/commits/topic/valgrind-thread-per-writer) (no intentions of merging it anywhere, just put it there so you can have a look).
I've configured collectd to only build the write_throttle and logfile plugins and tested three test scenarios under valgrind drd using the also checked in config files `throttle-noblock.conf`, `throttle-mix.conf` and `throttle-stall.conf` that register ten writers each that:

 1. All consume values at a greater rate than they are produced.
 2. Half of the writers consume at a greater rate, half at a lower rate.
 3. All consume values at a lower rate than they are produced.

### Valgrind


```
$ ./configure --disable-aggregation […] --disable-network --prefix=/tmp/collectd-valgrind
$ make && make install
$ valgrind --tool=drd /tmp/collectd-valgrind/sbin/collectd -C throttle-{noblock,mix,stall}.conf -f
```
Did not show any issues.

```
$ valgrind --tool=helgrind /tmp/collectd-valgrind/sbin/collectd -C throttle-{noblock,mix,stall}.conf -f
```

Did produce quite a bit of output that I am not too sure about how it should be interpreted. I've pasted one of them [here](https://gist.github.com/hnez/70cbb94fffaf9fedddf4b87b6413c12a). What makes me doubt the validity of the output are the [`Bug in libpthread` lines](https://gist.github.com/hnez/70cbb94fffaf9fedddf4b87b6413c12a#file-collectd-helgrind-log-txt-L50).

### ASAN

I've also created [a branch to do ASAN testing on](https://github.com/hnez/collectd/commits/topic/asan-thread-per-writer). That also uses the `write_throttle` plugin and adds some logging¹ on top of it and hacks in the -fsanitize=address flags.

¹ This was not included in the valgrind tests as printf could influence the concurrency.

The `write_throttle` plugin contains an option to trigger an invalid write to make sure that ASAN is configured correctly:

```
$ git clean -Xf
$ autoreconf
$ ./configure --disable-aggregation […] --disable-network --prefix=/tmp/collectd-asan
$ make && make install
$ /tmp/collectd-asan/sbin/collectd -C throttle-asan.conf -f
option = FQDNLookup; value = false;
option = Interval; value = 5.000000;
Created new plugin context.
[2022-09-21 10:28:10] plugin_load: plugin "logfile" successfully loaded.
[2022-09-21 10:28:10] plugin = logfile, key = LogLevel, value = notice
=================================================================
==109631==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffc096ee950 at pc 0x7f211682badb bp 0x7ffc096ee8c0 sp 0x7ffc096ee8b0
WRITE of size 4 at 0x7ffc096ee950 thread T0
    #0 0x7f211682bada in wl_config src/write_throttle.c:44
[…]
```

The log output shows that the Queues converge to the expected state (in this case half of the writers can keep up and keep their queues relatively short while the other half does not and values are dropped according to the configured limit (50 elemens)):

```
/tmp/collectd-asan/sbin/collectd -C throttle-mix.conf -f
option = FQDNLookup; value = false;
option = Interval; value = 5.000000;
option = WriteQueueLimitHigh; value = 50.000000;
option = WriteQueueLimitLow; value = 25.000000;
Created new plugin context.
[2022-09-21 10:33:13] plugin_load: plugin "logfile" successfully loaded.
[2022-09-21 10:33:13] plugin = logfile, key = LogLevel, value = notice
Spawning 5 write threads that will block
Spawning 5 write threads that will not block
[…]
Queue lengths:
  before enqueue:   0   0   0   0   0  47  46  46  44  47
  after enqueue:    1   1   1   1   1  48  47  47  45  48
Queue lengths:
  before enqueue:   1   1   0   0   1  48  47  47  45  48
  after enqueue:    2   2   1   1   2  49  48  48  46  49
Queue lengths:
  before enqueue:   2   1   1   1   2  49  48  48  46  49
  after enqueue:    3   2   2   2   3  50  49  49  47  50
Queue lengths:
  before enqueue:   3   2   2   2   3  50  49  49  47  50
  after enqueue:    4   3   3   3   4  50  49  49  47  50
```

I've let `collectd -C throttle-{noblock,mix,stall}.conf -f` run for a couple of hours and did not catch any errors.

eero-t:
Thanks, all that sounds really great!

[Hellgrind]
> Did produce quite a bit of output that I am not too sure about how it should be interpreted. I've pasted one of them [here](https://gist.github.com/hnez/70cbb94fffaf9fedddf4b87b6413c12a).

Function names are not shown for all glibc items in your backtraces.  Make sure that you have glibc debug symbols installed as Valgrind's suppression matching relies on function names (for ignoring known issues / false positives).

At least on Debian derived distros, "apt install valgrind" automatically pulls in glibc debug symbols.

> What makes me doubt the validity of the output are the [Bug in libpthread lines](https://gist.github.com/hnez/70cbb94fffaf9fedddf4b87b6413c12a#file-collectd-helgrind-log-txt-L50).

According to KDE bug discussion: https://bugs.kde.org/show_bug.cgi?id=445743

> This only happens when the mutex has the PTHREAD_PRIO_INHERIT attribute.
>
> The Helgrind code also implies that there could be a race condition between the intercept passing the call on to pthread_mutex_lock and Helgrind registering the return from pthread_mutex_lock. That could be affected by changing the priority.
>
> The other possibility is that DRD uses its own CV and mutex to ensure that the thread is joinable before it runs. There could be interference between the guest mutex with elevated priority and this CV/mutex.

See also Valgrind documentation: https://valgrind.org/docs/manual/hg-manual.html#hg-manual.effective-use

For ways how program or the libraries it uses, can potentially confuse Helgrind threading issue detection:
* Having memory usage errors (i.e. memcheck errors should be fixed before doing thread checks)
* Using memory recycling code which has not been annotated for Valgrind
* Using threading primitive syscalls directly instead of pthread functions
* Using unsupported libc/pthread version (not Glibc)
* Not joining threads
* Using TLS variables
* Using POSIX condition variables

I.e. one may unfortunately need to do quite a bit of investigation to find out whether reported issue is a (potential) false positive.

Best is to ignore stuff that is outside of the part you modified.

hnez:
I never would have guessed that it could have anything to do with missing debug symbols! I will make sure to install them and re-run my setups with helgrind.

Then I will record a before/after log to look out for any changes. I hope to get around to it tomorrow.

hnez:
> Then I will record a before/after log to look out for any changes.

Turns out that this is not necessary after all. All the errors seem to have come from the missing debug symbols.

Arch Linux does not seem to provide debug symbols as packages, but I can use `DEBUGINFOD_URLS` instead to fetch symbols from a server on demand.

The commands to run the test then look like this:

    $ ./configure --disable-aggregation […] --disable-network --prefix=/tmp/collectd-valgrind
    $ make && make install
    $ DEBUGINFOD_URLS="https://debuginfod.archlinux.org/" \
        valgrind --tool=helgrind /tmp/collectd-valgrind/sbin/collectd -C throttle-mix.conf -f
    ==252760== Helgrind, a thread error detector
    ==252760== Copyright (C) 2007-2017, and GNU GPL'd, by OpenWorks LLP et al.
    ==252760== Using Valgrind-3.19.0 and LibVEX; rerun with -h for copyright info
    ==252760== Command: /tmp/collectd-valgrind/sbin/collectd -f -C throttle-mix.conf
    ==252760==
    [2022-09-23 09:20:48] plugin_load: plugin "logfile" successfully loaded.
    Spawning 5 write threads that will block
    Spawning 5 write threads that will not block

I've let the three test setups run a again for a couple of hours and did not catch any data races.

eero-t:
Suppressions cannot match backtraces to issues using addresses, as those can change depending on address space randomization, linking order, compiler options, selected target architecture etc.  That's why they need symbols to work.

It's odd that you're not seeing *any* issues though.  At least when I was using Valgrind thread analysis tools on v5 couple of years ago, I was seeing quite a few things from collectd core, which I needed to manually check to be OK.  Do you see issues *without* your changes?   I.e. are your changes also making collectd easier to debug with Valgrind threading tools?

I'm asking because there can be buggy plugins that thread (e.g: https://github.com/collectd/collectd/issues/4050), and not having any warnings from collectd core would help with those.

hnez:
> Do you see issues _without_ your changes?

I've tested the current `collectd-6.0` branch with the `write_throttle` plugin (with a slight modification to make sure that an unique name is used for every `plugin_register_write`).

> I.e. are your changes also making collectd easier to debug with Valgrind threading tools?

I would love to be able to claim that, but helgrind did not catch any errors there as well - at least during runtime.
When I hit Ctrl+C it [outputs a bit of log](https://gist.github.com/hnez/f3531e417d604d84900d76eab28acc79), which does however contain possibly missing symbols again - so I would take these with a a grain of salt.

eero-t:
> > I.e. are your changes also making collectd easier to debug with Valgrind threading tools?
> 
> I would love to be able to claim that, but helgrind did not catch any errors there as well - at least during runtime.

I must have been remembering old collectd v5 results, as now that I tried my own plugin + "write_prometheus" with collectd v6, I saw only couple of issues in the libraries linked by the plugins [1], nothing in collectd core. This was with valgrind DRD/Helgrind from Ubuntu 22.04.

[1] "potential" race between what is done when plugin is initialized, and when it's shutdown i.e. not an issue in practice.

> When I hit Ctrl+C it [outputs a bit of log](https://gist.github.com/hnez/f3531e417d604d84900d76eab28acc79), which does however contain possibly missing symbols again - so I would take these with a a grain of salt.

`write_lock` is taken only by  `stop_write_threads()` and `plugin_write_dequeue()`, not by `plugin_write_thread()` which calls dequeue.  But setting `write_loop` bool to `false` at stop and checking for that in the other two functions,  does not really need to be protected.  At worst there's an extra round of checking for things to write (which is properly protected by `write_lock`).

PS. It's possible that you can rely on all backtraces relevant to runtime having been reported only at end.

hnez:
Hi,

some time has passed since we had the last activity here.
Are there still some open questions that need answering on my side, that I have missed?
How should we go forward with this PR?

Greetings
Leonard

src/collectd.conf.pod
```diff
@@ -304,19 +298,18 @@ metrics. For servers it is recommended to set this to a non-zero value, though.
...
-I<will> be enqueued. If the number of metrics currently in the queue is between
-I<LowNum> and I<HighNum>, the metric is dropped with a probability that is
-proportional to the number of metrics in the queue (i.e. it increases linearly
-until it reaches 100%.)
+queue. If there are I<HighNum> metrics in the queue a random element from the
+queue will be dropped. This results in the data becoming more and more sparse
+without dropping a lot of continguous values. An exception from the random
+selection of elements to drop are the I<LowNum> most recent elements which will
+never be dropped. This can be used to ensure that one miss-behaving write
+plugin does not result in dropped values for other well-behaved write plugins.
```
eero-t, line 306: Alternative phrasing:
> If there are `HighNum` metrics in the queue a random element from the queue will be dropped. An exception from the random selection of elements to drop are the `LowNum` most recent elements which will never be dropped.  Spreading the loss over the `LowNum` - `HighNum` range avoids large time gaps in the data on write hiccups. This way one miss-behaving write plugin is less likely to result in dropped values for well-behaved write plugins.

src/daemon/plugin.h
```diff
@@ -243,21 +243,19 @@ int plugin_shutdown_all(void);
...
+ *  Directly enqueues the given metric family without updating the cache or
+ *  calling the chain subsystem. The metric family may still be dropped from
+ *  the queue if collectd is stopped or the queue length surpasses the
+ *  configured limit.
  *
  * ARGUMENTS
  *  plugin     Name of the plugin. If NULL, the value is sent to all registered
  *             write functions.
- *  m   The actual value to be processed. Must not be NULL.
+ *  fam   The actual metric family to be processed.
```
eero-t, line 254: It still must not be NULL, right?

src/daemon/plugin.h
```diff
@@ -243,21 +243,19 @@ int plugin_shutdown_all(void);
...
  *
  * RETURN VALUE
- *  Returns zero upon success or non-zero if an error occurred. If `plugin' is
- *  NULL and more than one plugin is called, an error is only returned if *all*
- *  plugins fail.
+ *  Returns zero if the metric family was sucessfully enqueued or non-zero if
+ *  an error occurred.
```
eero-t, line 258: Do you have a write overflow test for both of the cases where plugin is specified, and where it is NULL (i.e. where family should go to all registered writers)?

eero-t:
I added few documentation comments.

> some time has passed since we had the last activity here. Are there still some open questions that need answering on my side, that I have missed? How should we go forward with this PR?

Of the collectd developers with commit access, I've seen only @mrunge be active in the past few months.  @elfiesmelfie was active a bit before that, but only on the CI side (I'm just a v6 plugin contributor myself).

=> I guess you need to wait until Matthias has again time to look into collectd and review this.  It's a largish change to core code, so it needs more careful review.

src/collectd.conf.pod
hnez: Sounds good to me, so I copied more or less verbatim into a new commit.
Thanks

src/daemon/plugin.h
hnez: The first thing `plugin_write` does is to check if `fam` is `NULL`, so nothing bad would happen if it was `NULL`.
I guess that's why I removed that part of the documentation.

The behavior between the previous `plugin_write` and the new `plugin_write` with regards to `fam` being `NULL` hasn't actually changed, so it makes sense to keep the documentation the same.
I've added a commit that re-adds the `Must not be NULL`.

src/collectd.conf.pod
eero-t: I thought the original "does not result in dropped values" sounded a bit too strong based on Matthias comments above. :-)

eero-t: You can resolve this now, I do not seem to have rights for that.

src/daemon/plugin.h
eero-t: This can be set as resolved now too.

src/daemon/plugin.h
hnez: Having `plugin == NULL` is actually the more common case (e.g. what usually happens when dispatching a metric family) and why I decided to implement this PR using a single queue with one tail per writer. Otherwise one queue per write would likely have been more efficient.

As far as I can tell calls to `plugin_write` _with_ a plugin name could happen via:

- Filter Chains
- Custom Python / Perl plugins (they have bindings for `plugin_write`)
- Custom C plugins that call `plugin_write` directly instead of going the usual dispatch route

I did however not play around with any of those.

hnez:
> Of the collectd developers with commit access, I've seen only @mrunge be active in the past few months. @elfiesmelfie was active a bit before that, but only on the CI side (I'm just a v6 plugin contributor myself).
> 
> => I guess you need to wait until Matthias has again time to look into collectd and review this. It's a largish change to core code, so it needs more careful review.

Yeah, maintainer time is always a scarce ressource and this change is not risk-free and needs a bit more review.
I am in no hurry here, as I can just carry the patch locally until a maintainer has time to look at it.
Just wanted to make sure noone is waiting for any activity on my side.

Thanks to @eero-t for the extensive review!

src/daemon/plugin.h
eero-t: Ah, right.  If filter chains are about aggregation, I do not think that has been ported to v6 API yet (see #4031).

If your earlier valgrind testing did not include "plugin != NULL" `plugin_write `case, could you check that just by using some specific plugin, e.g. `write_prometheus`?

IMHO `write_prometheus` is good test-case because MHD library used by it, also uses threads.  Testing would require also querying the write plugin, like this:
`http_proxy= curl --no-progress-meter http://127.0.0.1:<port>/metrics`

eero-t:
> I am in no hurry here, as I can just carry the patch locally until a maintainer has time to look at it.

Is this already running in a production setting on your side?  And if yes, what plugins are enabled in your collectd configuration?

hnez:
> Is this already running in a production setting on your side?

No, not yet. One of the projects we use collectd in has lost steam and the other one uses a lot of plugins, some of which we will have to port to collectd 6 first.

> And if yes, what plugins are enabled in your collectd configuration?

We currently load the following plugins: `syslog`, `chrony`, `contextswitch`, `cpu`, `df`, `disk`, `entropy`, `ethstat`, `exec`, `interface`, `load`, `memory`, `mmc`, `network`.
I did not check yet which of these are already ported to collectd 6, but we are planning to port the missing ones soon(ish), so we can use this patch.

eero-t:
> We currently load the following plugins: `syslog`, `chrony`, `contextswitch`, `cpu`, `df`, `disk`, `entropy`, `ethstat`, `exec`, `interface`, `load`, `memory`, `mmc`, `network`.  I did not check yet which of these are already ported to collectd 6,

Have you checked yet which of those need still porting?  And of those, which ones do not have a PR yet?

> but we are planning to port the missing ones soon(ish), so we can use this patch.

Knowing that you've been running this PR in production for some time would definitely help its merging.

hnez:
Hi,

I finally got around to glueing the relevant Pull Requests together and porting the missing plugins.
To replicate my setup:

    $ git clone "https://github.com/collectd/collectd.git"
    $ cd collectd
    $ git remote add manuelluis https://github.com/manuelluis/collectd.git
    $ git remote add hnez https://github.com/hnez/collectd.git
    $ git checkout -b cd6-octopus origin/collectd-6.0
    $ git merge --no-edit hnez/one-thread-per-writer \
        hnez/hnez/collectd6-{disk,exec,memory,mmc} \
        manuelluis/mlsr/collectd6-{chrony,contextswitch,df,entropy,load}
    $ ./configure --disable-aggregation --disable-barometer --disable-capabilities --disable-check_uptime \
        --disable-csv --disable-java --disable-lua --disable-network --disable-perl --disable-python \
        --disable-target_notification --disable-target_replace --disable-target_scale --disable-target_set \
        --disable-target_v5upgrade --disable-threshold --disable-uptime --disable-write_graphite \
        --without-libgrpc++ --disable-werror

I've also tested the plugins with [`-fsanitize=address` enabled](https://github.com/hnez/collectd/compare/f80e061...1bcdaef) and with `valgrind --tool=helgrind`. Apart from a false positive in `chrony` (added a comment in the repective PR) nothing out of the ordinary.

I've tested the following plugins with the octopus merge mentioned above:

- chrony - PR #3830 by @manuelluis is up to date and works fine, have commented there.
- contextswitch - PR #3812 by @manuelluis is up to date and works fine, have commented there.
- cpu - Is already in `collectd-6` and up to date with `main`
- df - PR #3810 by @manuelluis is up to date and works fine, have commented there.
- disk - PR #3809 by @manuelluis was behind `main` feature-wise, so I ported the changes and opened #4075.
- entropy - PR #3813 by @manuelluis is up to date and works fine, have commented there.
- ethstat - I somehow missed that one 😐. The version in `collectd-6` works but may have changes on `main` that are not in `collectd-6` yet. There is also an open PR #3855. So I can't say anything about it.
- exec - I've added a PR #4074 that wires the existing `PUTMETRIC` command through to it.
- interface - Is already in `collectd-6` and up to date with `main`.
- load - PR #3649 by @manuelluis is up to date and works fine, have commented there.
- memory - A port was already in `collectd-6` but behind `main` feature wise, so I've ported the features in PR #4076.
- mmc - This time I am responsible for this being behind `main` feature wise, so it seems fitting for me to open PR #4077 with a port.
- network - I've tried to port it without much success. I've documented my findings in Issue #3675.
- syslog - This one is easy as it is not affected by the API change.

I hope this helps in getting all the mentioned PRs merged (including this one, obviously 😉).

Greetings
Leonard

eero-t:
Wow, great work, thanks!

I did quick review on everything else except ethstat, mmc and network stuff.  I had few comments, but mainly on things in the original migrations PRs.

I would suggest using `--disable-all-plugins` configure option and enabling all used plugins explicitly.  Otherwise collectd configure can (try) enabling & building random set of plugins depending on what dev packages happens to be installed in the build environment, and one does not see what exactly you tested, or replicate it exactly.

(I'm myself doing builds just with: `--disable-all-plugins --enable-write_prometheus  --with-sysman=yes --enable-gpu_sysman`.)

eero-t:
@hnez Are you using this in production nowadays?

Now that Matthias merged this and few of the plugin PRs you linked, you could add e.g. a separate "Merging a basic subset of plugins to v6" ticket for tracking status for rest of the v6 plugins which PRs you've reviewed and are testing (in production?).

hnez:
Hi @eero-t,

> Are you using this in production nowadays?

sadly not. We will need to decide on some output plugin that is not the native `network` one and switch our infrastructure over to use it.
We are planning to do that, most likely with the MQTT plugin, but right now it has been pushed down the priority list, as we have finally (thanks chip-shortage!) received a new revision of our embedded hardware project that I am now busy bringing up and testing.

eero-t:
@mrunge / @hnez This seems to have broken Prometheus write plugin partially, maybe because the library used by it also threads: https://packages.debian.org/bullseye/libmicrohttpd-dev

While Prometheus plugin seems to be still reporting values ok, collectd logs now constantly these kind of warnings:
```
uc_update: Value too old: name = collectd_gpu_sysman_temperature_celsius{dev_file="card2",location="global-max",pci_bdf="0000:93:00.0",pci_dev="0x56c0"}; value time = 0.000; last cache update = 0.000;
uc_update: uc_update_metric failed: Error #-1; Additionally, strerror_r failed.
uc_update: Value too old: name = collectd_gpu_sysman_temperature_celsius{dev_file="card2",location="gpu-max",pci_bdf="0000:93:00.0",pci_dev="0x56c0"}; value time = 0.000; last cache update = 0.000;
uc_update: uc_update_metric failed: Error #-1; Additionally, strerror_r failed.
```

This is with collect v6 configured using:
`--disable-all-plugins --enable-write_prometheus  --enable-<some read plugin>`

(I used Sysman GPU metrics read plugin, but I would image that it would happen with any other read plugin.)

eero-t:
No problem, for me it was just a log spam issue, data still got through and nothing crashed (like was the case with `write_http`).

PR #4035, opened 2022-08-15
Fix anonymous connection_state pointer

ismaell:
Returning an anonymous pointer (which is limited to the scope of the function) causes an error when building with the `-Werror` flag.

Tested with: GCC 12.1.0
Test case:
```sh
echo 'void test(int **r) { *r = &(int){44}; }' | gcc -Werror=dangling-pointer -x c -c -o /dev/null -
```
Error:
```
<stdin>:1:25: error: storing the address of local variable '({anonymous})' in '*r' [-Werror=dangling-pointer=]
```


mrunge (approved):

mrunge:
Thank you.

The CI failures are not connected with this patch.

Issue #3897, opened 2021-08-04
Google OAuth 2.0 token size is now 2048


tonypelle:
*   Version of collectd: 5.12.0
*   Operating system / distribution: Ubuntu 18.04

## Expected behavior

The write_stackdriver plugin should be able to successfully create metric descriptors.

## Actual behavior

From the logs:

```
write_stackdriver plugin: getting access token failed with
write_stackdriver plugin: POST https://monitoring.googleapis.com/v3/projects/<redacted>/metricDescriptors failed
write_stackdriver plugin: wg_call_metricdescriptor_create failed with status -1
```

Visual inspection of the code in `write_stackdriver.c` implicates that the 256 byte buffer allocated in `wg_get_authorization_header` is too small, now that the access_token can be up to 2048 bytes.

tonypelle:
Google OAuth documentation: https://developers.google.com/identity/protocols/oauth2#size

PR #4037, opened 2022-08-19
Increase buffer for OAuth access token.  Another fix for #3897.

tonypelle:
ChangeLog: increase oauth access token buffer size.

Previous PR was insufficient in fixing #3897.  Increase another OAuth access token buffer to 2048 to match the updated spec from Google.


mrunge (approved):

mrunge:
thank you

PR #4045, opened 2022-09-08
require systemd-devel to rpmbuild SMART plugin

quartsize:
Changelog: RPM specfile: systemd-devel needed to build SMART plugin

I'm including some additions to the build instructions that might have helped me.


mrunge (approved):
Thank you.

I never caught that before since I always had the disk plugin included in the build as well.

PR #4047, opened 2022-09-15
[collectd 6.0] gpu_sysman: Add more metric output variants and better control for them

eero-t:
ChangeLog: gpu_sysman: Add more metric output variants and better control for them

Changes:
* Add "rate" variant flag (3rd one) for MetricsOutput option, and change that option to a bitmask
* Rename MetricsOutput "raw" + "derived" flag names to "counter" + "ratio" ones matching metric names [1]
* Add "ratio" variant to all metric types where L0 API provides limit / max value for the metric (frequency, power, temperature)
* Add "rate" variant for memory BW metric type, in addition to already existing "counter" and "ratio" variants
* All metric types that submit multiple variants, check now these flags, and if metric type is enabled, but none of its output variants are enabled, warning is logged
* Tests for the new metrics + related improvements to the generic sysman plugin test code

Because this adds quite a bit of code to the plugin (231 added, 71 removed lines), main question is whether providing multiple variants for same metric is fine (e.g. for memory bandwidth: increasing counter, bytes per second value, and a ratio for how much that is from max BW)?

(Things like Prometheus can provide rate data based on counter data only, but if one is consuming metric data directly, providing rate directly from collectd would be preferred.)

[1] As there has been no release with this plugin, this API change should be fine.

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -796,7 +796,7 @@
 #<Plugin gpu_sysman>
 #   Samples 1
 #   LogGpuInfo false
-#   MetricsOutput both
+#   MetricsOutput "counter:rate:ratio"
 #   DisableMemory false
 #   DisableMemoryBandwidth false
 #   DisableFrequency false
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3736,10 +3736,10 @@ Several of the metric types support multiple variants for their
 values. This option specifies which ones of them are to be reported.
 
 Default is to report all variants ("counter:rate:ratio"). To reduce
-amount of data, it is better to configure just the relevant one for
-given use. Note that some of the metric types support only two of
-these variants, whereas metrics supporting only single variant ignore
-this option.
+amount of data, it is better to configure just the relevant ones for
+given use (e.g. "counter:ratio" or "rate:ratio"). Note that some of
+the metric types support only two of these variants, whereas metrics
+supporting only single variant ignore this option.
 
 Counter metric variant (e.g. HW energy usage as Joules counter) is
 preferred by Prometheus as doing rate calculations in Prometheus is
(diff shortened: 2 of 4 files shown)


mrunge (approved):

eero-t:
Review for my dev branch (https://github.com/eero-t/collectd/pull/2) resulted in few minor things that apply to these changes too.  I'll cherry-pick them here later this week, so merging this could be post-poned until that.

eero-t:
@mrunge I've cherry-picked the applicable 2 review commits from dev branch, and this PR can be merged now.

Would you like me to squash fix commits to the earlier changes in the PR before merge, to get nicer description generated for it?

PS. Plugin code in PR is checked both with Klocwork (static analysis) and Valgrind memcheck tool (dynamic analysis for running tests[1]).

[1] Gcov coverage stats for test run, when code is built with -O3 [2]:
* `gpu_sysman.c`: 92%
  * Some error checks and part of gpu info option output at startup are not covered
* `gpu_sysman_test.c`: 88%
  * Some error checks and tracing code for debugging testing failures are not covered by valgrinded test runs

[2] Higher optimization level increases coverage, because code has a lot of switch-cases, and gcc can turn most of the cases into lookup tables (so all case values in a switch do not need to be separately tested, for related code to be covered).

eero-t:
> Would you like me to squash fix commits to the earlier changes in the PR before merge, to get nicer description generated for it?

@mrunge?

mrunge:
A squash would be much easier to cherry-pick. But that's not going to happen. Thank you for the offer, I'd think this is good to go as is.

eero-t:
> Thank you for the offer, I'd think this is good to go as is.

@mrunge Merge?

mrunge (approved):
Thank you, and sorry to keep you waiting. This looks good to me.

PR #4051, opened 2022-09-21
Set X-Content-Type-Options header For Prometheus plugin

jsaroja66:
ChangeLog: Write_Prometheus plugin:  "X-Content-Type-Options : nosniff" response header has been added.

### DIFF
--- a/src/write_prometheus.c
+++ b/src/write_prometheus.c
@@ -51,6 +51,9 @@
   "encoding=delimited"
 #define CONTENT_TYPE_TEXT "text/plain; version=0.0.4"
 
+#define MHD_HTTP_HEADER_CONTENT_TYPE_OPTIONS "X-Content-Type-Options"
+#define CONTENT_TYPE_OPTIONS_TEXT "nosniff"
+
 #if MHD_VERSION >= 0x00097002
 #define MHD_RESULT enum MHD_Result
 #else
@@ -274,6 +277,8 @@ static MHD_RESULT http_handler(void *cls, struct MHD_Connection *connection,
 #endif
   MHD_add_response_header(res, MHD_HTTP_HEADER_CONTENT_TYPE,
                           want_proto ? CONTENT_TYPE_PROTO : CONTENT_TYPE_TEXT);
+  MHD_add_response_header(res, MHD_HTTP_HEADER_CONTENT_TYPE_OPTIONS,
+                          CONTENT_TYPE_OPTIONS_TEXT);
 
   MHD_RESULT status = MHD_queue_response(connection, MHD_HTTP_OK, res);
 


eero-t:
Why this is needed?  (And could you add it first to v6 branch?)

jsaroja66:
This is a vulnerability identified by a security scan conducted on collectd service. The report mentions this:
"The Anti-MIME-Sniffing header X-Content-Type-Options was not set to 'nosniff'. This allows older versions of Internet Explorer and Chrome to perform MIME-sniffing on the response body, potentially causing the response body to be interpreted and displayed as a content type other than the declared content type. Current (early 2014) and legacy versions of Firefox will use the declared content type (if one is set), rather than performing MIME-sniffing.
Solution. We have to ensure that the application/web server sets the Content-Type header appropriately, and that it sets the X-Content-Type-Options header to 'nosniff' for web pages.
Add 'X-Content-Type-Options: nosniff' header to the responses, to avoid browser mime sniffing and incorrectly assuming file types, as well as enable Cross Origin Read Blocking (CORB) for specific mime types."
This is the reason I added it.
Regarding v6 branch, could you please let me know how I add this change to v6 branch? I don't see any v6 branch in collectd project. I could only see main branch. Please help me since this is my first time contributing to collectd.

src/write_prometheus.c
```diff
@@ -274,6 +277,8 @@ static MHD_RESULT http_handler(void *cls, struct MHD_Connection *connection,
 #endif
   MHD_add_response_header(res, MHD_HTTP_HEADER_CONTENT_TYPE,
                           want_proto ? CONTENT_TYPE_PROTO : CONTENT_TYPE_TEXT);
+  MHD_add_response_header(res, MHD_HTTP_HEADER_CONTENT_TYPE_OPTIONS,
```
mrunge, line 280: you are dropping the want_proto conditional.

mrunge:
Thank you for proposing this patch.  I have commented inline.

Personally, I disagree with this adding any security (why should I care about nosniff?), also: how is the information provided by the exporter secure at all? What I could agree here on is: this adds expected content type headers and provides better cooperation with prometheus. Does it?

The collectd-6 branch is this one: https://github.com/collectd/collectd/tree/collectd-6.0

jsaroja66:
Hi, This commit does not add content type to the response, it rather adds x-content-type-options header, which when set to nosniff, makes sure that the content type header in the response is set to what we declare it to be (plain/text in case of write_prometheus) and that it can't be and wont be changed by others. Since this commit does not add content type header, I have skipped want_proto conditional, which is only meant for content type header and it is already in place. Please let me know if you have any further queries. If not, I will make pull request to collectd-6 project.

eero-t:
> This is a vulnerability identified by a security scan conducted on collectd service. The report mentions this: "The Anti-MIME-Sniffing header X-Content-Type-Options was not set to 'nosniff'. This allows older versions of Internet Explorer and Chrome to perform MIME-sniffing on the response body, potentially causing the response body to be interpreted and displayed as a content type other than the declared content type.

Prometheus plugin endpoint is expected to be read by Prometheus.   In which real world use-case it would be read by Chrome, or IE?

(Only case where something else than Prometheus should be used, is debugging why data is not getting there, by checking whether the endpoint is providing any data, but wget/curl is fine for that.)

> Current (early 2014) and legacy versions of Firefox will use the declared content type (if one is set), rather than performing MIME-sniffing. Solution. We have to ensure that the application/web server sets the Content-Type header appropriately, and that it sets the X-Content-Type-Options header to 'nosniff' for web pages.

Although Prometheus plugin uses MHD, it is not a web server.  It does not server web pages, just single metrics endpoint.

> Add 'X-Content-Type-Options: nosniff' header to the responses, to avoid browser mime sniffing and incorrectly assuming file types, as well as enable Cross Origin Read Blocking (CORB) for specific mime types."

Plugin does not provide any user-generated content or anything where external parties could inject extra content which could be mis-interpreted as some other mime-type.

=> This is false positive from the security scan

PR #4053, opened 2022-09-27
mmc: cache open file descriptors to block devices

hnez:
ChangeLog: mmc plugin: cache file descriptors to reduce udev events generated as side-effect

Hi,

there is a slight annoyance when using the eMMC health info support for Micron and Sandisk chips introduced by me in #4006: an udev event is generated on every call to the read callback.

This is due to to the `watch` option that can be used in udev rules, which is described in the man page as:

>  Watch the device node with inotify; when the node is closed after being opened for writing, a change uevent is synthesized.

This watch option is [enabled by default](https://github.com/systemd/systemd/blob/ef2ad30aee9fa99b0fdb8fe7efda397513cec6af/rules.d/60-block.rules) for all block devices. The intention behind this is to be notified about changes to the partition table. The mmc plugin does however also need to open the block device for writing, even though it never modifies its content, in order to be able to issue `ioctl`s with vendor defined MMC-commands.

Reduce the amount of generated change events from one per read to one per collectd runtime by caching the open file descriptor.

Fixes: 2f15c704 (mmc: add more vendor specific and generic data sources (#4006))

### DIFF
--- a/src/mmc.c
+++ b/src/mmc.c
@@ -49,6 +49,18 @@ static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
 
 static ignorelist_t *ignorelist = NULL;
 
+// Cache of open file descriptors for /dev/mmcblk? block devices.
+// The purpose of caching the file descriptors instead of open()/close()ing for
+// every read is to prevent the generation of udev events for every close().
+typedef struct dev_cache_entry_s {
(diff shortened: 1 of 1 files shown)


mrunge:
Thank you. The failed CI checks are not related to this PR.

PR #4054, opened 2022-10-04
Use power_average as a fallback for subfeature power_input if isn't a…

marekbecka:
ACPI power meter reports a computer consumption with SENSORS_SUBFEATURE_POWER_AVERAGE. Which is probably more common for power sensors.

For backward compatibility this is used for power features as a fallback only if SENSORS_SUBFEATURE_POWER_INPUT is not available.

### DIFF
--- a/src/sensors.c
+++ b/src/sensors.c
@@ -241,6 +241,8 @@ static int sensors_load_conf(void) {
     while ((feature = sensors_get_features(chip, &feature_num)) != NULL) {
       const sensors_subfeature *subfeature;
       int subfeature_num = 0;
+      bool have_power_input = false;
+      featurelist_t *fl_power = NULL;
 
       /* Only handle voltage, fanspeeds and temperatures */
       if ((feature->type != SENSORS_FEATURE_IN) &&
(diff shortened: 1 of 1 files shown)


eero-t:
From which version onward `SENSORS_SUBFEATURE_POWER_AVERAGE` is supported (i.e. should that code be inside ifdef)?

Changes to sensors plugin cannot cause virt, smart and netlink plugin fails in CI, so those CI issues can be ignored.  Same thing with Fedora rawhide `mount.h` header related conflicts.

Single-line "ChangeLog:" line is needed for this PR though.

eero-t:
> The changes in this PR are working for me (applied as a patch to 5.12.0). FWIW Before I found this PR I used a much simpler patch that just added a check for `SENSORS_SUBFEATURE_POWER_AVERAGE` here
> 
> https://github.com/collectd/collectd/blob/050877ed952ffc15c849803a9b3e77c5cec15f81/src/sensors.c#L277
>
> and that also worked :)

Are there any downsides on having both enabled? 

(If not, please file PR for that, and refer this PR.  Simpler solution would be preferred.)
 
> @eero-t Regarding your question
> 
> > From which version onward `SENSORS_SUBFEATURE_POWER_AVERAGE` is supported (i.e. should that code be inside ifdef)?
> 
> It's supported since 3.0.2 (0x401), see https://github.com/hramrach/lm-sensors/blob/c297368a0e4b0442fca06bae58a0bd3386629af0/doc/libsensors-API.txt#L143 This is an older version than what's even required for `SENSORS_SUBFEATURE_POWER_INPUT` (the only currently supported power measurement) which was added in lm-sensors 3.1.0 (0x410), see https://github.com/hramrach/lm-sensors/blob/c297368a0e4b0442fca06bae58a0bd3386629af0/doc/libsensors-API.txt#L118

Ok, thanks, even RHEL8 seems to have already v3.4 (and oldest distro versions for which Ubuntu & Debian list packages, have v3.6).

PR #3934, opened 2021-10-25
check_presence plugin: Add new plugin to detect new and missing hosts

minfrin:
ChangeLog: check_presence plugin: Add new plugin to detect new and missing hosts.

This PR adds a new plugin that keeps track of hosts that have been seen by collectd, and
sends notifications when a new host is seen, or an existing is no longer seen for a
period of time, respectively.

It depends on liblmdb to provide the underlying index, chosen for low cost.

Example notification when a host is seen:

Notification: severity = OKAY, host = little-net.local, plugin = host, type = host, type_instance = found, message = Host is found: little-net.local

Example notification when a host goes missing:

Notification: severity = FAILURE, host = little-net.local, plugin = host, type = host, type_instance = lost, message = Host not seen for 45 seconds: little-net.local

Issue #3675, opened 2020-09-24
[collectd 6] Migrate the network plugin


octo:
This is a tracking issue for migrating the *network plugin* to *collectd 6*. Progress is tracked in the [collectd&nbsp;6 project](https://github.com/orgs/collectd/projects/1). Technical guidance for migrating plugins to *collectd 6* is provided in the [V5 to V6 migration guide](https://collectd.org/wiki/index.php/V5_to_v6_migration_guide) topic in the wiki.

**Build status:** Failure

The *network plugin* currently doesn't compile and has been disabled in the automatic builds. Please see below for instructions to re-enable the plugin in the CI systems.

### Next steps

This is an automatically generated todo list. If one of the following steps doesn't make sense, blame the automation. :see_no_evil:

*   [ ] Assign this issue to yourself.
*   [ ] Determine whether the *network plugin* should be migrated in the first place. If this is not an obvious "yes", discuss on the [mailing list](https://collectd.org/wiki/index.php/Mailing_list) or at the next [community call](https://collectd.org/wiki/index.php/Community_calls).
*   [ ] Re-enable the *network plugin* in the CI systems by removing the appropriate "disable" line from `.cirrus.yml`, `.travis.yml`, and `solaris-ci.flags`.
*   [ ] Change the *network plugin's* write callback to accept a `metric_family_t*` instead of a `value_list_t*`. For technical guidance, refer to the [V5 to V6 migration guide](https://collectd.org/wiki/index.php/V5_to_v6_migration_guide).
*   [ ] Ensure that the plugin handles *distribution metrics* correctly, i.e. metrics of type `METRIC_TYPE_DISTRIBUTION` that contain a `distribution_t*` inside the `value_t`.
*   [ ] Create a pull request referencing this issue, based on the [collectd-6.0](https://github.com/collectd/collectd/tree/collectd-6.0) branch.

eero-t:
> The native [collectd network protocol](https://collectd.org/wiki/index.php/Binary_protocol) is, as far as I can tell, quite closely coupled to the value_list_t based plugin API. This makes it hard to serialize metric_family_ts into network packets, so implementing the sending side would be difficult.
>
> The value_list_t based API has separate fields for every field used in the network plugin:
...
> 3. Agree on standard label names for the value_list_t fields used in the network plugin. E.g. encourage all plugins to set host, plugin and plugin_instance labels and serialize those into the network packets like before.

[...]

PR #4055, opened 2022-10-10
check_presence plugin: Add new plugin to detect new and missing hosts

minfrin:
ChangeLog: check_presence plugin: Add new plugin to detect new and missing hosts.

This PR adds a new plugin that keeps track of hosts that have been seen by collectd, and sends notifications when a new host is seen, or an existing is no longer seen for a period of time, respectively.

It depends on liblmdb to provide the underlying index, chosen for low cost.

Example notification when a host is seen:

Notification: severity = OKAY, host = little-net.local, plugin = host, type = host, type_instance = found, message = Host is found: little-net.local

Example notification when a host goes missing:

Notification: severity = FAILURE, host = little-net.local, plugin = host, type = host, type_instance = lost, message = Host not seen for 45 seconds: little-net.local

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -189,6 +189,9 @@ Florian Eckert
 Gergely Nagy <algernon at madhouse-project.org>
  - Write-Riemann plugin.
 
+Graham Leggett <minfrin at sharp.fm>
+ - check_presence plugin.
+
 Hari TG <hari.tg at intel.com>
  - dcpmm plugin.
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -826,6 +826,14 @@ chrony_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 chrony_la_LIBADD = -lm
 endif
 
+if BUILD_PLUGIN_CHECK_PRESENCE
+pkglib_LTLIBRARIES += check_presence.la
+check_presence_la_SOURCES = src/check_presence.c
+check_presence_la_CPPFLAGS = $(AM_CPPFLAGS) $(BUILD_WITH_LIBLMDB_CPPFLAGS)
+check_presence_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(BUILD_WITH_LIBLMDB_LDFLAGS)
+check_presence_la_LIBADD = -llmdb
+endif
+
 if BUILD_PLUGIN_CHECK_UPTIME
 pkglib_LTLIBRARIES += check_uptime.la
 check_uptime_la_SOURCES = src/check_uptime.c
--- a/README
+++ b/README
@@ -708,6 +708,11 @@ Features
       Selects multiple value lists based on patterns or regular expressions
       and creates new aggregated values lists from those.
 
+    - check_presence
+      Detects whether a host has been seen for the first time, or whether a
+      host has not been seen for a specified interval, and creates
+      notifications for each case.
+
     - threshold
       Checks values against configured thresholds and creates notifications if
       values are out of bounds. See collectd-threshold(5) for details.
--- a/configure.ac
+++ b/configure.ac
@@ -3193,6 +3193,55 @@ AC_SUBST([BUILD_WITH_LIBLDAP_CPPFLAGS])
 AC_SUBST([BUILD_WITH_LIBLDAP_LDFLAGS])
 # }}}
 
+# --with-liblmdb {{{
+AC_ARG_WITH([liblmdb],
+  [AS_HELP_STRING([--with-liblmdb@<:@=PREFIX@:>@], [Path to liblmdb.])],
+  [
+    if test "x$withval" = "xyes"; then
+      with_liblmdb="yes"
+    else if test "x$withval" = "xno"; then
+      with_liblmdb="no"
+    else
+      with_liblmdb="yes"
+      LIBLMDB_CPPFLAGS="$LIBLMDB_CPPFLAGS -I$withval/include"
+      LIBLMDB_LDFLAGS="$LIBLMDB_LDFLAGS -L$withval/lib"
+    fi; fi
+  ],
+  [with_liblmdb="yes"]
+)
+
+SAVE_CPPFLAGS="$CPPFLAGS"
+SAVE_LDFLAGS="$LDFLAGS"
+
+CPPFLAGS="$CPPFLAGS $LIBLMDB_CPPFLAGS"
+LDFLAGS="$LDFLAGS $LIBLMDB_LDFLAGS"
+
+if test "x$with_liblmdb" = "xyes"; then
+  AC_CHECK_HEADERS([lmdb.h],
+    [with_liblmdb="yes"],
+    [with_liblmdb="no ('lmdb.h' not found)"]
+  )
+fi
+
+if test "x$with_liblmdb" = "xyes"; then
+  AC_CHECK_LIB([lmdb], [mdb_version],
+    [with_liblmdb="yes"],
+    [with_liblmdb="no (symbol 'mdb_version' not found)"]
+  )
+fi
+
+CPPFLAGS="$SAVE_CPPFLAGS"
+LDFLAGS="$SAVE_LDFLAGS"
+
+if test "x$with_liblmdb" = "xyes"
+then
+  BUILD_WITH_LIBLMDB_CPPFLAGS="$LIBLMDB_CPPFLAGS"
+  BUILD_WITH_LIBLMDB_LDFLAGS="$LIBLMDB_LDFLAGS"
+fi
+AC_SUBST([BUILD_WITH_LIBLMDB_CPPFLAGS])
+AC_SUBST([BUILD_WITH_LIBLMDB_LDFLAGS])
+# }}}
+
 # --with-liblua {{{
 AC_ARG_VAR([LIBLUA_PKG_CONFIG_NAME], [Name of liblua used by pkg-config])
 if test "x$LIBLUA_PKG_CONFIG_NAME" != "x"
@@ -7226,6 +7275,7 @@ AC_PLUGIN([capabilities],        [$plugin_capabilities],      [Platform static c
 AC_PLUGIN([ceph],                [$plugin_ceph],              [Ceph daemon statistics])
 AC_PLUGIN([cgroups],             [$plugin_cgroups],           [CGroups CPU usage accounting])
 AC_PLUGIN([chrony],              [yes],                       [Chrony statistics])
+AC_PLUGIN([check_presence],      [$with_liblmdb],             [Host presence checking plugin])
 AC_PLUGIN([check_uptime],        [yes],                       [Notify about uptime reset])
 AC_PLUGIN([connectivity],        [$plugin_connectivity],      [Network interface up/down events])
 AC_PLUGIN([conntrack],           [$plugin_conntrack],         [nf_conntrack statistics])
@@ -7602,6 +7652,7 @@ AC_MSG_RESULT([    libjvm  . . . . . . . $with_java])
 AC_MSG_RESULT([    libkstat  . . . . . . $with_kstat])
 AC_MSG_RESULT([    libkvm  . . . . . . . $with_libkvm])
 AC_MSG_RESULT([    libldap . . . . . . . $with_libldap])
+AC_MSG_RESULT([    liblmdb . . . . . . . $with_liblmdb])
 AC_MSG_RESULT([    liblua  . . . . . . . $with_liblua])
 AC_MSG_RESULT([    libmemcached  . . . . $with_libmemcached])
 AC_MSG_RESULT([    libmicrohttpd . . . . $with_libmicrohttpd])
@@ -7675,6 +7726,7 @@ AC_MSG_RESULT([    capabilities  . . . . $enable_capabilities])
 AC_MSG_RESULT([    ceph  . . . . . . . . $enable_ceph])
 AC_MSG_RESULT([    cgroups . . . . . . . $enable_cgroups])
 AC_MSG_RESULT([    chrony. . . . . . . . $enable_chrony])
+AC_MSG_RESULT([    check_presence  . . . $enable_check_presence])
 AC_MSG_RESULT([    check_uptime. . . . . $enable_check_uptime])
 AC_MSG_RESULT([    connectivity. . . . . $enable_connectivity])
 AC_MSG_RESULT([    conntrack . . . . . . $enable_conntrack])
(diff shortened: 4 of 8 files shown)


minfrin:
Replaces https://github.com/collectd/collectd/pull/3934

eero-t:
How this differs from the original version in https://github.com/collectd/collectd/pull/3357?

Nowadays new plugins are preferred to be added to `collectd-6.0` branch.  While there is no release of the v6 yet, there are few upsides to using v6 branch:
* Support for labels e.g. in `write_prometheus` output plugin
* Fewer CI issues than in `main` branch

(I'm not a collectd developer, just a developer of another new plugin.)

mrunge:
I can only fully support @eero-t s comments. There's nothing to add, other than the CI failures don't look connected to the PR here.

minfrin:
> How this differs from the original version in #3357?

Original version was submitted in 2019, patch has been reapplied as the underlying codebase moved and conflicted.

> Nowadays new plugins are preferred to be added to `collectd-6.0` branch. While there is no release of the v6 yet, there are few upsides to using v6 branch:
> 
> * Support for labels e.g. in `write_prometheus` output plugin
> * Fewer CI issues than in `main` branch
> 
> (I'm not a collectd developer, just a developer of another new plugin.)

This was written before collectd 6 existed.

I took a look a while back for details of what was involved in supporting collectd 6, but I couldn't find anything the described the changes needed. Does anyone know?

Alternatively if there is a "hello world" collectd 6 plugin that shows the required structure I can rewrite it again.

eero-t:
Collectd documentation migration guide is a bit sparse: https://collectd.org/wiki/index.php/V5_to_v6_migration_guide

But you can use some of the simpler plugins that are in process of being, or have already been migrated to v6, as examples.

You'll find them from the open & closed pull requests list: https://github.com/collectd/collectd/pulls?q=is%3Apr+%22collectd+6%22+is%3Aopen

minfrin:
> Collectd documentation migration guide is a bit sparse: https://collectd.org/wiki/index.php/V5_to_v6_migration_guide

Am I right in understanding that metric_t is going away, and metric_family_t is replacing it?

> But you can use some of the simpler plugins that are in process of being, or have already been migrated to v6, as examples.
> 
> You'll find them from the open & closed pull requests list: https://github.com/collectd/collectd/pulls?q=is%3Apr+%22collectd+6%22+is%3Aopen

I could find only one notification plugin in the list above, and that was new so didn't show what changes were needed.

The check_presence plugin does not make use of metric_t. Can you confirm if any changes are needed so I can make them? So far it looks like the notification mechanism is unchanged?

eero-t:
> Am I right in understanding that metric_t is going away, and metric_family_t is replacing it?

Yes.
 
> The check_presence plugin does not make use of metric_t. Can you confirm if any changes are needed so I can make them?

I'm just an author of one v6 metric plugin, and did not participate in v6 API design or discussions (they happened before I got involved).   Maybe @mrunge can answer?

> So far it looks like the notification mechanism is unchanged?

Best may be if you just checkout `collectd-6.0` branch and try building your plugin against it, and whether the resulting plugin works.

minfrin:
> Best may be if you just checkout `collectd-6.0` branch and try building your plugin against it, and whether the resulting plugin works.

Is collectd-6.0 older or newer than the main branch?

Alas the collectd-6.0 branch currently fails to build on both MacOS and Linux:

```
  CC       src/aggregation.lo
src/aggregation.c:718:22: error: unknown type name 'metric_single_t'; did you mean 'metric_list_t'?
 static int agg_write(metric_single_t const *m, /* {{{ */
                      ^~~~~~~~~~~~~~~
                      metric_list_t
src/aggregation.c: In function 'module_register':
src/aggregation.c:744:40: error: 'agg_write' undeclared (first use in this function); did you mean 'agg_read'?
   plugin_register_write("aggregation", agg_write, /* user_data = */ NULL);
                                        ^~~~~~~~~
                                        agg_read
src/aggregation.c:744:40: note: each undeclared identifier is reported only once for each function it appears in
make[1]: *** [Makefile:7862: src/aggregation.lo] Error 1
make[1]: Leaving directory '/home/minfrin/src/collectd/collectd-6.0'
make: *** [Makefile:5528: all] Error 2
```

eero-t:
> Is collectd-6.0 older or newer than the main branch?

Both, since it was branched couple of years ago.  Many of the plugins have not been updated since that, as people have made their PRs against v5 (main branch).  Before v6 is released, those changes need to be synched to v6.

For now it's better to to use `--disable-all-plugins` configure option and enable just the plugins you're explicitly interested about.

minfrin:
I am stuck looking for detailed examples of write plugins. Specifically I have this code, because I need the name of the value_list_t host:

```
key.mv_size = sizeof(vl->host);
key.mv_data = (void *)vl->host;
```

What function would I need to use to extract a host from a metric?

eero-t:
> I need the name of the value_list_t host:
> ```
> key.mv_size = sizeof(vl->host);
> key.mv_data = (void *)vl->host;
> ```
> 
> What function would I need to use to extract a host from a metric?

I'm not sure I understand you question.  `host` member in `value_list_t` is just a string: https://github.com/collectd/collectd/blob/collectd-6.0/src/daemon/plugin.h#L93 ?

Or do you mean function for getting value of a given label for a metric: https://github.com/collectd/collectd/blob/collectd-6.0/src/daemon/metric.h#L115 ?

minfrin:
> Or do you mean function for getting value of a given label for a metric: https://github.com/collectd/collectd/blob/collectd-6.0/src/daemon/metric.h#L115 ?

What I'm after specifically is, given a metric_family_t*, how do I extract the host from that?

There are no examples the existing code for use of metric_label_get(), and this function operates on a single metric, not the whole metric_family.

I'm stuck as to what the intended way is to gain access to explicit values. All the examples loop through all values, and don't care about the content.

eero-t:
I'm not collectd developer myself, but this seems fairly obvious from the `metric.h` header.

Just loop through the family (metric type) metric array and query host label for each of the items:
https://github.com/collectd/collectd/blob/collectd-6.0/src/daemon/metric.h#L121 ?

Something like (completely untested):
```
const char* has_family_label_value(family_t *fam, const char *label, *const char *check) {
  assert(fam && label && check);
  metric_list_t list = fam.metric;
  for (int i = 0; i < list.num; i++) {
     const char *value = metric_label_get(list.ptr[i], label);
     if (!value) continue;
     if (strcmp(value, check) == 0) return true;
  }
  return false;
}
```

If your intention is checking whether some host value is new, you obviously need to think carefully how you store your items (should you e.g. qsort items so you can bsearch) and do the queries & checks (e.g. is your inner loop for the list of hosts, or for metrics in given family) to avoid quadratic behavior when number of hosts and metrics grows.

minfrin:
When what is previously a trivial one liner is now nine lines of code, that's a red flag that something is being done incorrectly, or the function call I'm supposed to use is being missed.

eero-t:
There were no labels in v5, it's a completely new feature that matches to what modern telemetry components need (see e.g. OpenTelemetry spec), so comparing it to v5 is a bit pointless.

Because of your question, I was assuming that something in v6 would be setting host info to a label in metrics, and you would then want to get them from there, but maybe you want something else, I don't know...  It all is in headers though, just look at them, and if that's not enough, check the corresponding .c file(s).  They are all pretty straightforward.

PS. Adding one more database dependency to collectd seems a bit dubious, when existing plugins already depend on several different ones...

minfrin:
> There were no labels in v5, it's a completely new feature that matches to what modern telemetry components need (see e.g. OpenTelemetry spec), so comparing it to v5 is a bit pointless.
> 
> Because of your question, I was assuming that something in v6 would be setting host info to a label in metrics, and you would then want to get them from there, but maybe you want something else, I don't know... It all is in headers though, just look at them, and if that's not enough, check the corresponding .c file(s). They are all pretty straightforward.

The reason I'm lost is that this piece of information no longer exists, containing the host that sent the metric:

https://github.com/collectd/collectd/blob/main/src/daemon/plugin.h#L108

What I don't see is any coherent mapping from these old values in collectd v5 to collectd v6. There are no #defines coming up in any searches corresponding to "host". Is the "host" now a user defined string that should be made configurable?

Is "host" optional now, or can I still get the name of the host that logged the metric?

> PS. Adding one more database dependency to collectd seems a bit dubious, when existing plugins already depend on several different ones...

I covered this a number of years ago. The plugin needs to do a read on every metric write, and so the database used to do this must do extremely cheap lock-free reads. The database has been carefully chosen to meet this requirement. The database is purely a library, has no configuration, and takes up very little space. The collectd plugin mechanism already ensures that dependencies are limited in scope to the plugins that need them.

eero-t:
The host member is still there in the values list as can be seen from link above.

However, when looking at the code:
* Values lists seem to be used only with the legacy metric APIs
* When they're dispatched, they get converted to family metrics: https://github.com/collectd/collectd/blob/collectd-6.0/src/daemon/plugin.c#L2196
* And host item is added to "instance" label: https://github.com/collectd/collectd/blob/collectd-6.0/src/utils/common/common.c#L1172

So yeah, it seems to be available only as label now, and by default, only for metrics coming from plugins that have not been converted yet.

However, already in v5:
* `host` item was initialized from global `hostname_g` variable: https://github.com/collectd/collectd/blob/main/src/daemon/plugin.c#L711
* Which was set only **once** at startup: https://github.com/collectd/collectd/blob/main/src/daemon/collectd.c#L117

So your plugin is indirectly tracking changes in *non-changing* global variable value by constantly inspecting (all?) metrics???  WTF?

That's something that makes sense to do only once.  And you don't need collectd metrics for that, you can just query it directly.


Also, in https://github.com/collectd/collectd/pull/3357, you claim that "we have to detect hosts that have disappeared during times when collectd is not running, for example during restarts of collectd or reboots of the machine running collectd".

But this plugin is instead tracking changes in collectd configuration `Hostname` setting, and when that is not set, changes in the host hostname (hardcoded in /etc/ or coming from DNS).  Which could change independently of whether the machine is running or not.

If you want to actually identify different host machines, use their ID: https://www.freedesktop.org/software/systemd/man/machine-id.html

Additionally, doing this in collectd, does not detect host going away.  That's something that you would need to do outside of the host itself.  So this whole plugin seems kind of bogus...

minfrin:
> * And host item is added to "instance" label: https://github.com/collectd/collectd/blob/collectd-6.0/src/utils/common/common.c#L1172

That's what I was looking for - host maps to "instance". Ideally this should be a #define and not a string buried in the code, as the risk is that plugins might start choosing their own constants.

> However, already in v5:
> 
> * `host` item was initialized from global `hostname_g` variable: https://github.com/collectd/collectd/blob/main/src/daemon/plugin.c#L711
> * Which was set only **once** at startup: https://github.com/collectd/collectd/blob/main/src/daemon/collectd.c#L117
> 
> So your plugin is indirectly tracking changes in _non-changing_ global variable value by constantly inspecting (all?) metrics??? WTF?
> 
> That's something that makes sense to do only once. And you don't need collectd metrics for that, you can just query it directly.
> 
> Also, in #3357, you claim that "we have to detect hosts that have disappeared during times when collectd is not running, for example during restarts of collectd or reboots of the machine running collectd".
> 
> But this plugin is instead tracking changes in collectd configuration `Hostname` setting, and when that is not set, changes in the host hostname (hardcoded in /etc/ or coming from DNS). Which could change independently of whether the machine is running or not.
> 
> If you want to actually identify different host machines, use their ID: https://www.freedesktop.org/software/systemd/man/machine-id.html
> 
> Additionally, doing this in collectd, does not detect host going away. That's something that you would need to do outside of the host itself. So this whole plugin seems kind of bogus...

Please tone down the aggressive responses like "WTF" and "kind of bogus", they make working on this unnecessarily difficult.

For the vast majority of the plugins in collectd, this host is the source of the data, and so the host value will be picked up by a static mechanism, like a global static variable. Obviously this plugin is not relevant for these cases.

For network plugins which receive data from other hosts using protocols like the collectd protocol, SNMP, etc, the hostname of the remote host supplying the data is now important. Given that none of the network plugins appear to have been upgraded to collectd-6.0 yet, it's easy to miss this detail.

When network plugins are receiving data from other hosts, the host field is now no longer our host but the remote host instead, and the check_presence plugin kicks in. It will generate a notification the very first time some remote host was seen, and will generate a notification for you as soon as your remote host is not longer being seen.

_Crucially you do not have to configure anything at all to get this information, it is completely automatic and zero effort._

We do not want a flood of false alerts when collectd is stopped and restarted. Nobody deserves a heart attack. For this reason we keep track of hosts across a collectd restart, so if the host was there when we stopped, and is still there when we started again, we don't say anything and panic anybody unnecessarily.

eero-t:
Sorry about the tone.   Your clarifications should really have been in the initial PR description and/or the plugin documentation. :-/

(I.e. plugin being intended to extract host info from metrics provided by network plugins that collect data from other hosts, and which metric data includes remote host names.)

Other things that the PR description / docs should state is this being a write plugin.  And like with other write plugins, I think that should also be reflected in its name.

Additionally, I don't think you should be registering your functions anywhere else than in module_register() function.  There might be locking problems otherwise.  If init fails, write function should not get called.  I'm not sure about shutdown, but to be sure, your shutdown code should handle partial init.

Because module is threading internally... Have you run collectd with your plugin under Valgrind `--tool=memcheck`, `--tool=drd` and `--tool=helgrind` options?


> For the vast majority of the plugins in collectd, this host is the source of the data, and so the host value will be picked up by a static mechanism, like a global static variable. Obviously this plugin is not relevant for these cases.
>
> For network plugins which receive data from other hosts using protocols like the collectd protocol, SNMP, etc, the hostname of the remote host supplying the data is now important.

So this is intended for collectd configurations that include *only* networking plugins?

> Given that none of the network plugins appear to have been upgraded to collectd-6.0 yet, it's easy to miss this detail.

Guilty.  There's some recent discussion for migrating "network" plugin, see: https://github.com/collectd/collectd/issues/3675

eero-t:
Summary of the issues:
* It's a new plugin, so it should be filed against `collectd-6.0` branch, and use the new v6 APIs
* PR description should state that it's a write plugin investigating metrics that collectd network read plugins _collect from remote hosts_, and reporting what hosts are producing those metrics
* Plugin name should better reflect this (I don't have good suggestion though)

With the v6 switch to OpenTelemetry metrics, host name should be included (only) to metric family resource attributes: https://opentelemetry.io/docs/specs/semconv/resource/host/

See the related pull requests for more info:
* https://github.com/collectd/collectd/pull/4187
* https://github.com/collectd/collectd/pull/4199

So it may be enough to just call `label_set_get(fam->resource, "host.name")` on each metric family.

However, I'm not sure whether plugins that have not yet been updated for v6, and require converting their (deprecated) value lists to v6 metrics, also provide relevant (default) resource attributes.  If not, you may need to call also `metric_label_get(metric, "<labelname>")` on each metric under metric families.

PR #4057, opened 2022-10-25
[collectd 6.0] cpython: fix build with Python 3.11

sgn:
ChangeLog: cpython: fix build with Python 3.11

Python 3.11 moves longintrepr.h into cpython sub-directory. However, in this version, longintrepr.h is always included.


mrunge (approved):
Thank you!

PR #4060, opened 2022-11-06
fix for 'rrdtool plugin: illegal attempt to update ... (minimum one second step)'

sorcky:
ChangeLog: collectd: 'rrdtool plugin: illegal attempt to update ... (minimum one second step)'

When rrdtool is used on collectd server for data storage it is crucial for clients to send metrics with strictly increasing timestamps. Equal adjacent timestamps cause error similar to this:
`rrdtool plugin: rrd_update_r (/opt/collectd/rrd/testhost1/GenericJMX-jetty/gauge-idleThreads.rrd) failed: /opt/collectd/rrd/testhost1/GenericJMX-jetty/gauge-idleThreads.rrd: illegal attempt to update using time 1667670455 when last update time is 1667670455 (minimum one second step)`

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -583,12 +583,12 @@ static void *plugin_read_thread(void __attribute__((unused)) * args) {
      * should be called. */
     rf->rf_next_read += rf->rf_effective_interval;
 
-    /* Check, if `rf_next_read' is in the past. */
-    if (rf->rf_next_read < now) {
-      /* `rf_next_read' is in the past. Insert `now'
-       * so this value doesn't trail off into the
-       * past too much. */
-      rf->rf_next_read = now;
+    /* Check, if `rf_next_read' is on the same second as 'now' or in the past */
+    if (CDTIME_T_TO_TIME_T(rf->rf_next_read) <= CDTIME_T_TO_TIME_T(now)) {
+      /* `rf_next_read' is on the same second as `now' or earlier
+       * it can cause 'rrdtool plugin: illegal attempt to update ... (minimum one second step)' on server
+       * so move it to the next second. */
+      rf->rf_next_read = now + rf->rf_effective_interval;
     }
 
     DEBUG("plugin_read_thread: Next read of the `%s' plugin at %.3f.",


src/daemon/plugin.c
```diff
@@ -583,12 +583,12 @@ static void *plugin_read_thread(void __attribute__((unused)) * args) {
...
     rf->rf_next_read += rf->rf_effective_interval;
 
-    /* Check, if `rf_next_read' is in the past. */
-    if (rf->rf_next_read < now) {
-      /* `rf_next_read' is in the past. Insert `now'
-       * so this value doesn't trail off into the
-       * past too much. */
-      rf->rf_next_read = now;
+    /* Check, if `rf_next_read' is on the same second as 'now' or in the past */
+    if (CDTIME_T_TO_TIME_T(rf->rf_next_read) <= CDTIME_T_TO_TIME_T(now)) {
```
eero-t, line 587: Why the addition of `CDTIME_T_TO_TIME_T()`?

src/daemon/plugin.c
```diff
@@ -583,12 +583,12 @@ static void *plugin_read_thread(void __attribute__((unused)) * args) {
...
+      /* `rf_next_read' is on the same second as `now' or earlier. it can cause
+       * 'rrdtool plugin: illegal attempt to update...(minimum one second step)'
+       * on server so move it to the next second. */
+      rf->rf_next_read = now + rf->rf_effective_interval;
```
eero-t, line 591: Effective interval could be temporarily up to max interval, so use of `rf_interval` could be better.

src/daemon/plugin.c
sorcky: CDTIME_T_TO_TIME_T() to round timestamp to nearest integer second.
E.g.:
now = 1667983304.2
rf->rf_next_read = 1667983304.4
It means, that current metric is collected at 1667983304.2 and next one will be collected at 1667983304.4
Technically, condition `rf->rf_next_read < now` is false but if we don't increment rf->rf_next_read, server gets metrics with these timestamps in sequence: 1667983304.2, 1667983304.4. They both round to 1667983304, but rrdtool storage "resolution" is 1 second minimum and rrd rejects whole batch of data, causing gap on graph.

src/daemon/plugin.c
sorcky: As I understand `rf->rf_effective_interval` is calculated dynamically on each iteration of read loop according to callback status. It reflects current state of plugin more accurately. Thus I decided to use it.

sorcky:
Should I take any actions or make changes for approval?

src/daemon/plugin.c
mrunge: Collectd supports intervals smaller than one second. With this change, we'd cut that off.

mrunge:
Thank you for your PR!

I would prefer to round the timestamps down in the rrd write plugin[1]; other write targets do support timestamps in sub-seconds resolution.

[1] https://github.com/collectd/collectd/blob/main/src/rrdtool.c

sorcky:
How are you going to handle situation if `rrd write plugin` gets two timestamps within one second?
This pull request avoids this situation on client side.

sorcky:
Current behavior of rrd_write is to ignore whole bunch of data (not only data with timestamps within the same second). In our clients we flush data to server once per 120 secs. So it leads to gaps of up to 120 secs of data.
It would be nice indeed to drop only data with timestamps that rounds to the same second.

eero-t:
@sorcky, @mrunge Any further updates / comments on this?

(I'm asking because apparently e.g. WRT has quite a bit of usage for this plugin.)

sorcky:
@eero-t I'm not sure what comment/update you are expecting from me.
This patch works fine for me. Apparently, it is not that good for the code mainainers.

eero-t:
> @eero-t I'm not sure what comment/update you are expecting from me. This patch works fine for me. Apparently, it is not that good for the code mainainers.

Not expecting anything, I was just wondering whether there have been any changes/updates since 2022 (that were not pushed to this PR).

sorcky:
No changes since then.
But we don't update collectd very often at our side. This PR is associated with the problem we encounered during our infrastructure upgrade to OL8 2 yars ago.

PR #4062, opened 2022-11-25
plugin.c: fix drop probability

carlospeon:
ChangeLog: plugin.c: fix drop probability

The higher is the position p in the window between min and max thresholds the higher must be the probability to drop. (ie, p=0, returns false and p=1 returns true).


mrunge (approved):
thank you!

PR #4063, opened 2022-11-25
format_influxdb: add config option to write metadata

carlospeon:
ChangeLog: format_influxdb: add config option to write metadata

In #3938 the feature to write meta data to influxdb was added. This adds more labels and indexes to influxdb and could not be always desirable. 

This PR adds a configuration option to enable this feature, disabled by default to keep the previous versions of write_influxdb_udp plugin behavior.


mrunge (approved):
Thank you!

PR #4065, opened 2022-12-02
evaluate flush condition with oldest value in the buffer

carlospeon:
ChangeLog: network.c, write_influx_udp.c evaluate flush condition with oldest value in the buffer.

Plugin flush condition should be evaluated with the oldest value in the buffer not with the latest one.


octo:
Rebased on `main` so fixes to the CI system are picked up.

carlospeon:
Run format.sh in network.c

Issue #4066, opened 2022-12-06
"network plugin: getnameinfo failed: Non-recoverable failure in name resolution" on NetBSD


sparked435:
*   Version of collectd: 5.12.0
*   Operating system / distribution: NetBSD 9.1
*   Kernel version (if applicable): default with release (NetBSD 9.1 (GENERIC) #0: Sun Oct 18 19:24:30 UTC 2020  mkrepro@mkrepro.NetBSD.org:/usr/src/sys/arch/amd64/compile/GENERIC amd64)

## Expected behavior

For collectd to receive data from network plugin

## Actual behavior

Network plugin logs the message "network plugin: getnameinfo failed: Non-recoverable failure in name resolution"; data from remote instance is not processed

## Steps to reproduce

*   Install collectd 5.12.0nb2 via pkgin
*   Enable network plugin

This looks identical to #3477 except a different (and current) flavor of BSD.

sparked435:
So I've found NetBSD is extremely picky about the 'salen' parameter when calling getnameinfo(). Specifically, it must match the exact size of the sockaddr struct that's being handled to it - the program cannot simply provide sizeof(struct sockaddr_storage), even if that's the size of the buffer being passed.

Pull request coming soon.

PR #4072, opened 2022-12-30
Pass precise size of struct sockaddr to getnameinfo() on NetBSD.

sparked435:
This resolves issue #4066.

I cannot imagine the if/else block will cause problems on any other OS, but wrapped it in an ifdef just to be sure.

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -443,8 +443,17 @@ static int network_dispatch_values(value_list_t *vl, /* {{{ */
 
   if (address != NULL) {
     char host[48];
-    status = getnameinfo((struct sockaddr *)address,
-                         sizeof(struct sockaddr_storage), host, sizeof(host),
+    size_t len = sizeof(struct sockaddr_storage);
+
+#ifdef __NetBSD__
+    if (address->ss_family == AF_INET) {
+      len = sizeof(struct sockaddr_in);
+    } else if (address->ss_family == AF_INET6) {
+      len = sizeof(struct sockaddr_in6);
+    }
+#endif
+
+    status = getnameinfo((struct sockaddr *)address, len, host, sizeof(host),
                          NULL, 0, NI_NUMERICHOST | NI_NUMERICSERV);
     if (status != 0) {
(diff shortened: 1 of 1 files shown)


mrunge:
Thank you for your PR. Please run `contrib/format.sh src/network.c` in your branch, it will create something like the following, which I could not push to your branch myself.

```
diff --git a/src/network.c b/src/network.c
index 39e534a2..e30d1b0e 100644
--- a/src/network.c
+++ b/src/network.c
@@ -1142,7 +1142,7 @@ static int parse_part_sign_sha256(sockent_t *se, /* {{{ */

   return 0;
 } /* }}} int parse_part_sign_sha256 */
-  /* #endif HAVE_GCRYPT_H */
+/* #endif HAVE_GCRYPT_H */

 #else  /* if !HAVE_GCRYPT_H */
 static int parse_part_sign_sha256(sockent_t *se, /* {{{ */
@@ -1297,7 +1297,7 @@ static int parse_part_encr_aes256(sockent_t *se, /* {{{ */

   return 0;
 } /* }}} int parse_part_encr_aes256 */
-  /* #endif HAVE_GCRYPT_H */
+/* #endif HAVE_GCRYPT_H */

 #else  /* if !HAVE_GCRYPT_H */
 static int parse_part_encr_aes256(sockent_t *se, /* {{{ */
@@ -1706,7 +1706,7 @@ static int network_set_interface(const sockent_t *se,
       ERROR("network plugin: setsockopt (bind-if): %s", STRERRNO);
       return -1;
     }
-      /* #endif HAVE_IF_INDEXTONAME && SO_BINDTODEVICE */
+    /* #endif HAVE_IF_INDEXTONAME && SO_BINDTODEVICE */

 #else
     WARNING("network plugin: Cannot set the interface on a unicast "
```

sparked435:
> Thank you for your PR. Please run `contrib/format.sh src/network.c` in your branch, it will create something like the following, which I could not push to your branch myself.

Done, thank you!

mrunge (approved):
thank you

PR #4074, opened 2023-01-11
[collectd 6] src/exec.c make the existing PUTMETRIC command available to scripts

hnez:
ChangeLog: exec plugin: Make PUTMETRIC available as command

This PR adds the `PUTMETRIC` command to the list of commands available in the `eval`plugin.
This means you can now have scripts like these that natively uses the `metric_*` types introduced with collectd 6.

    #!/bin/bash

    while true
    do
        echo "PUTMETRIC unit_test 42"
        echo "PUTMETRIC gauge type=GAUGE 42"
        echo "PUTMETRIC counter type=Counter 42"
        echo "PUTMETRIC untyped type=untyped 42"
        echo "PUTMETRIC quoted_gauge type=\"GAUGE\" 42"
        echo "PUTMETRIC with_interval interval=10.0 42"
        echo "PUTMETRIC with_time time=1594806526 42"
        echo "PUTMETRIC with_label label:unquoted=bare 42"
        echo "PUTMETRIC with_label label:quoted=\"with space\" 42"
        echo "PUTMETRIC multiple_label label:foo=1 label:bar=2 42"
        sleep $COLLECTD_INTERVAL
    done

Existing `exec` scripts can keep using the `PUTVAL` command via the backward compatibility layer implemented in `plugin_dispatch_values`.

I've checked if there are any changes to the `exec` plugin in the `main` branch that did not land in `collectd-6` yet, which is not the case.

Issue #3629 tracks the exec plugin migration progress. As far as I can tell this PR should tick the _Update the exec plugin to produce a `metric_family_t`_ check mark.


eero-t:
Looks OK.  How much did you test it?

hnez:
> How much did you test it?

Not much more than running the script in the initial PR comment for some iterations.
I will have to port the script we currently use via the `exec` plugin to `PUTMETRIC` to get some real-world experience.
To get said real-world collectd-6 experience we will however also have to change the plugin we use to get data out of the device away from the native `network` plugin, as that turned out to be a bit stubborn to port #3675.
So it could be some time before we see it in production.

PR #4077, opened 2023-01-11
[collectd 6] src/mmc.c port to collectd 6

hnez:
ChangeLog: mmc plugin: Port to collectd 6

The `mmc` plugin was added to the `main` branch after the `collectd-6` branch was split off.
This PR:

  - Cherry-picks the commit that added the `mmc` plugin to the `main` branch (see #3887)
  - Cherry-picks my commits that added support for more eMMCs (#4006 and #4053)
  - Ports the plugin to collectd 6 API.

This brings the plugin to feature parity with the `main` branch.

Issue #4034, opened 2022-08-10
Doesn't compile - possibly an issue due to a recent glibc update?


WarmChocolateCake:
*   Version of collectd: 5.12.0
*   Operating system / distribution: Arch Linux
*   Kernel version (if applicable): 5.18.16

## Expected behavior

Clean compile

## Actual behavior

Compilation errors:

```
/usr/include/linux/mount.h:95:6: error: redeclaration of enum fsconfig_command
  95 | enum fsconfig_command {
     |    ^----------------
In file included from .src/utils/mount/mount.h:57,
[...]
```
## Steps to reproduce
.configure / make / make install

This appears to be something to do with glibc-2.36? (Based purely on web searches which shows a similar issue with Gentoo:
https://bugs.gentoo.org/863683

So, commenting out:

/src/utils/mount/mount.h:

```
#if HAVE_SYS_MOUNT_H
#include <sys/mount.h>
#endif

```
Resulted in a good compile.

FYI, glibc was recently upgraded in Arch linux repos (local device upgraded 3rd Aug), but I've never installed collectd before, so this may all be coincidence ;)

ConiKost:
Yes, this is related to glibc-2.36. I am also hitting this error.

eero-t:
It's also breaking Fedora Rawhide tests in CI.  @mrunge?

eero-t:
This Glibc issue is best explained here: https://lore.kernel.org/all/20220802183409.GB2040@redhat.com/T/

The issue is kernel `linux/mount.h` and Glibc 2.36 `sys/mount.h` having (partially) same content, so they conflict, and  `linux/mount.h` getting also included indirectly through `linux/fs.h` (which provides some useful additional defines/macros).

Direct conflict is documented here:
* https://sourceware.org/glibc/wiki/Release/2.36#Packaging_Changes
* https://sourceware.org/glibc/wiki/Synchronizing_Headers

But how both end up being used in collectd `mount.c` happens very indirectly:
* `HAVE_XFS_XQM_H`: `mount.c` -> `xfs/xqm.h` -> `xfs/xfs.h` -> `xfs/linux.h` -> `linux/fs.h` -> `linux/mount.h`
* `HAVE_SYS_MOUNT_H`: `mount.c` -> `mount.h` -> `sys/mount.h`

[...]

Issue #4027, opened 2022-07-20
CI errors for unrelated plugins


eero-t:
*   Version of collectd: main branch
*   Operating system / distribution: collectd CI
*   Kernel version (if applicable):

## Expected behavior

CI tests pass in PRs for plugins that have not been touched.

## Actual behavior

See CI results for: https://github.com/collectd/collectd/pull/4022

@mrunge added already bug for "test_plugin_virt" fails for RedHat: https://github.com/collectd/collectd/issues/4018

But additionally "test_plugin_netlink" + "test_plugin_smart" fail for Debian Sid in that PR.

And on Rawhide, plugin builds fail to compile warnings:
```
src/nut.c: In function 'nut_add_ups':
src/nut.c:89:66: error: passing argument 4 of 'upscli_splitname' from incompatible pointer type [-Werror=incompatible-pointer-types]
   89 |   status = upscli_splitname(name, &ups->upsname, &ups->hostname, &ups->port);
      |                                                                  ^~~~~~~~~~
      |                                                                  |
      |                                                                  int *
In file included from src/nut.c:33:
/usr/include/upsclient.h:109:35: note: expected 'uint16_t *' {aka 'short unsigned int *'} but argument is of type 'int *'
  109 |                         uint16_t *port);
      |                         ~~~~~~~~~~^~~~
src/nut.c: In function 'nut_read':
src/nut.c:281:66: error: passing argument 4 of 'upscli_list_next' from incompatible pointer type [-Werror=incompatible-pointer-types]
  281 |   while ((status = upscli_list_next(ups->conn, query_num, query, &answer_num,
      |                                                                  ^~~~~~~~~~~
      |                                                                  |
      |                                                                  unsigned int *
/usr/include/upsclient.h:100:25: note: expected 'size_t *' {aka 'long unsigned int *'} but argument is of type 'unsigned int *'
  100 |                 size_t *numa, char ***answer);
      |                 ~~~~~~~~^~~~
cc1: all warnings being treated as errors
```

And errors:
```
In file included from src/python.c:36:
src/cpython.h:29:10: fatal error: longintrepr.h: No such file or directory
   29 | #include <longintrepr.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[1]: *** [Makefile:8647: src/python_la-python.lo] Error 1
  CC       src/python_la-pyvalues.lo
In file included from src/pyconfig.c:34:
[...]

Issue #4018, opened 2022-06-13
virt plugin fails in checks


mrunge:
*   Version of collectd: recent git checkout
*   Operating system / distribution: recent centos, fedora
*   Kernel version (if applicable):

## Expected behavior

make check passes

(Description of the behavior / output that you expected)

## Actual behavior
https://github.com/collectd/collectd/runs/6825588247?check_suite_focus=true

mrunge:
worth to mention, it doesn't fail on a real machine, only with the container.

PR #4084, opened 2023-01-30
[collectd 6] Cirrus CI / GitHub Workers - sync config with main branch

hnez:
Hi,

we've found some differences in the way the CI behaves between the `main` and the `collect-6.0` branches in #4077.
This PR cherry-picks all changes to `.cirrus.yml` and `.github/workflows/build.yml` from `main` on top of `collectd-6.0`.

Changes introduced here that were not in `main`/`collectd-6.0` are:

- Usage of `on_failure:` instead of `|| cat … || false` in the `.cirrus.yml` to print a log on failure. This matches the what 1b040d7cd0a8d8fcabdd8952a48e97045de9d508 already did in the `collectd-6.0` branch and is imho prettier.
- Trigger on changes to the `collectd-6.0` branch in the GitHub action config instead of `main`.

My only experience with these CI runners is them flagging my pull requests as bad, so feel free to comment if something looks wrong.

The GitHub worker config will likely need to be extended to disable plugins that are not yet buildable.

ChangeLog: CI: Update collectd-6.0 branch to match main

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -118,10 +118,6 @@ debian_default_toolchain_task:
 #
 redhat_default_toolchain_task:
   matrix:
-    - allow_failures: false
-      skip_notifications: false
-      container:
-        image: collectd/ci:el6_x86_64
     - allow_failures: true
       skip_notifications: true
       container:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -178,6 +178,9 @@ non_standard_toolchains_task:
         - make -j2 -sk
       tests_script:
         - VALGRIND_OPTS="--errors-for-leak-kinds=definite" make -j2 -sk check
+      on_failure:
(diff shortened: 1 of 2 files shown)


hnez:
I've now added enough `configure` flags to the GitHub worker config to make most jobs build.
The remaining build errors look like valid (but small) issues to me that should be solved by fixing them and not by modifying the job config to work around them.

eero-t:
This cause v6 CI to fail to same code issues in plugins as v5 CI already fails, for which there are several open tickets (https://github.com/collectd/collectd/issues/4034, https://github.com/collectd/collectd/issues/4027, https://github.com/collectd/collectd/issues/4018).

mrunge:
Thank you! While there are undeniable CI failures, I still think this is a good step forward. I am inclined to merge it anyways.

mrunge (approved):

PR #4086, opened 2023-01-31
Fix md5 compile issue

watologo1:
ChangeLog: snmp plugin: Compile issue fixed if net-snmp has NETSNMP_DISABLE_MD5 set


mrunge:
Thank you for your patch!

mrunge (approved):

Issue #3993, opened 2022-04-01
hugepages: cannot open /sys/devices/system/node/node0/hugepages/hugepages-1048576kB/demote


ConiKost:
*   Version of collectd: 5.12.0
*   Operating system / distribution: Gentoo
*   Kernel version (if applicable): 5.17.1

## Expected behavior
Do not try to open demote, as it seems write only, but no read.

## Actual behavior
Collectd tries to read /sys/devices/system/node/node0/hugepages/hugepages-1048576kB/demote, but it's a write-only file.

```
$ ls -la /sys/devices/system/node/node0/hugepages/hugepages-1048576kB/demote
--w------- 1 root root 4096  1. Apr 23:10 /sys/devices/system/node/node0/hugepages/hugepages-1048576kB/demote
```

## Steps to reproduce

*   Enable hugepages
*   Start Collectd
*   Check logs

lnatapov:
Any update on this issue ? Was it a misconfiguration ?

ConiKost:
No. This is not a missconfiguration. As the interface changed, the plugin needs to be updated.

PR #4091, opened 2023-02-22
Hugepages plugin skips reading write-only file

yadneshk:
ChangeLog: hugepages plugin: Skip reading write-only file `demote` in hugepages directory

Since 'demote'[1] is a write-only file do not attempt to read it. This also prevents the plugin from generating incessant logs about the failure to open it.

Fixes: #3993

[1] https://docs.kernel.org/admin-guide/mm/hugetlbpage.html

### DIFF
--- a/src/hugepages.c
+++ b/src/hugepages.c
@@ -125,6 +125,14 @@ static int read_hugepage_entry(const char *path, const char *entry,
   struct entry_info *info = e_info;
   double value;
 
+  /**
+   * Since demote is a write-only file, do not attempt to open it.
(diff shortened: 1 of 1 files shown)


eero-t:
Thanks for the fix @yadneshk !  It looks otherwise good, but could you move the check before snprintf() call?

(CI failures with virt, smart, mount plugins on rolling distros  are known issues with tickets, netlink fail is likely similar, at least it cannot come from this change.)

eero-t:
Thanks, this looks fine now!

@mrunge merge?

mrunge:
Thank you Yadnesh, this looks good to me.

mrunge (approved):

PR #4093, opened 2023-03-01
vmem: add metrics start with "pgscan_" in Kernel Linux

zzzyhtheonly:
ChangeLog: Vmem plugin: A feature  supporting metrics start with "pgscan_" in Kernel Linux was added.

Some Linux Kernel versions have metrics start with "pgscan_" in /proc/vmstat, for instance:
```
cat /proc/vmstat | grep pgscan
pgscan_kswapd 0
pgscan_direct 0
pgscan_direct_throttle 0
```

### DIFF
--- a/src/vmem.c
+++ b/src/vmem.c
@@ -211,10 +211,6 @@ static int vmem_read(void) {
       char *inst = key + strlen("pgsteal_");
       value_t value = {.derive = counter};
       submit_one(inst, "vmpage_action", "steal", value);
-    } else if (strncmp("pgscan_", key, strlen("pgscan_")) == 0) {
-      char *inst = key + strlen("pgscan_");
-      value_t value = {.derive = counter};
-      submit_one(inst, "vmpage_action", "scan", value);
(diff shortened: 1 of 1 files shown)


src/vmem.c
```diff
@@ -211,6 +211,10 @@ static int vmem_read(void) {
       char *inst = key + strlen("pgsteal_");
       value_t value = {.derive = counter};
       submit_one(inst, "vmpage_action", "steal", value);
+    } else if (strncmp("pgscan_", key, strlen("pgscan_")) == 0) {
```
kwiatrox, line 214: This will hide the later checks for `pgscan_kswapd_` and `pgscan_direct_`. Imho it should be placed after those two.

eero-t:
The new `pgscan_` check covers previous `pgscan_kswapd_` and `pgscan_direct_` cases, so patch should remove those two.

@mrunge Looking at `man proc`, apparently earlier only the latter ones were provided, so I do not think omitting rest of `pgscan_`  items was deliberate decision from @octo.  I.e. after above change, PR should be fine.

eero-t:
Thanks for the update, it looks good now.

mrunge (approved):
thank you

PR #4094, opened 2023-03-03
Hugepages plugin skips reading write-only file

yadneshk:
ChangeLog: hugepages plugin: Skip reading write-only file `demote` in hugepages directory

Since 'demote' is a write-only file do not attempt to to read it. This also prevents the plugin from generating incessant logs about the failure to open it.

Fixes: #3993

https://docs.kernel.org/admin-guide/mm/hugetlbpage.html

### DIFF
--- a/src/hugepages.c
+++ b/src/hugepages.c
@@ -125,6 +125,14 @@ static int read_hugepage_entry(const char *path, const char *entry,
   struct entry_info *info = e_info;
   double value;
 
+  /**
+   * Since demote is a write-only file, do not attempt to open it.
+   * https://docs.kernel.org/admin-guide/mm/hugetlbpage.html
+   */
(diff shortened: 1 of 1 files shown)


mrunge:
@eero-t 
The 5.13 branch was opened a while ago, but did not receive the love it desired. My proposal would be to merge the main branch to 5.13 and to then cut a release. At this stage, the branch is just old and stale.

mrunge (approved):

eero-t:
> @eero-t The 5.13 branch was opened a while ago, but did not receive the love it desired. My proposal would be to merge the main branch to 5.13 and to then cut a release. At this stage, the branch is just old and stale.

@mrunge that sounds good to me!   Earlier there were multiple releases per year, but now it has been several years without one...

PR #4101, opened 2023-03-23
Fix a memory leak in the `write_stackdriver` plugin

agrue:
ChangeLog: write_stackdriver plugin: A memory leak has been fixed

The `payload` was never freed. This was affecting potentially every call to both the "write" and "flush" callbacks.


eero-t:
CI fails are for unrelated plugins, not related to this.

`sd_output_reset()` returns strdup()ed JSON buffer as `payload`:
https://github.com/collectd/collectd/blob/main/src/utils/format_stackdriver/format_stackdriver.c#L620

`wg_call_timeseries_write()` just gives it to `do_post()` which sets it for libcurl, and asks it  to do the transfer, after which payload is not needed any more.

=> @mrunge this is fine to merge.

octo (approved):

mrunge:
thank you @octo , this was on my todo list.

agrue:
Thanks, much appreciated.

Issue #4098, opened 2023-03-10
[collectd 6] one thread per write plugin breaks write_http (and write_prometheus) behavior


acmiyaguchi:
*   Version of collectd: v6.0-55efb56a8d56f5b37bfdad430801e731210ecd1f
*   Operating system / distribution: linux, ubuntu 20.04
*   Kernel version (if applicable):

## Expected behavior

The write_http can write to an endpoint (service running socat) and write_prometheus metrics are non-empty.

## Actual behavior

`write_http` plugin will segfault. `write_prometheus` plugin does not register new metrics.

## Steps to reproduce

I bisected the change to the following commit: https://github.com/collectd/collectd/commit/55efb56a8d56f5b37bfdad430801e731210ecd1f

I built collectd in a container the following options:

```dockerfile
WORKDIR /opt
ARG GIT_REV=55efb56a8d56f5b37bfdad430801e731210ecd1f
RUN git clone https://github.com/collectd/collectd.git collectd-src && \
    cd collectd-src && \
    git checkout ${GIT_REV}

WORKDIR /opt/collectd-src
ARG PREFIX=/opt/collectd
RUN ./build.sh && \
    ./configure \
        --prefix ${PREFIX} \
        --enable-debug \
        --disable-all-plugins \
        --enable-statsd \
        --enable-write_prometheus \
        --enable-write_http \
        --enable-logfile \
        --enable-unixsock \
        && \
    make all install
```

I use the following configuration with collectd in the foreground (actually a slightly different configuration but the shape is the same):

```
Hostname "localhost"
FQDNLookup false
Interval 3

LoadPlugin logfile
<Plugin logfile>
    LogLevel "debug"
    File STDOUT
    Timestamp true
    PrintSeverity true
</Plugin>

LoadPlugin statsd
<Plugin statsd>
  Host "0.0.0.0"
  Port "8125"
</Plugin>

LoadPlugin write_prometheus

<Plugin "write_prometheus">
  Port "9103"
</Plugin>

<LoadPlugin "write_http">
  FlushInterval 3
</LoadPlugin>

<Plugin "write_http">
  <Node "HttpWriter">
    URL "http://http-mock:9000"
  </Node>
</Plugin>
```

I configure a mock service reachable at `http://http-mock:9000` inside of a docker container running the following script:

```bash
bash -c "
  echo listening on 0.0.0.0:9000;
  while true; do
[...]

PR #4102, opened 2023-03-24
src/daemon/plugin.c: don't store references to stack allocated values

hnez:
The changes in commit 55efb56a.

wrongly assume that the references passed in to `plugin_register_write()` somehow outlive the spawned write thread.
While this is true for some plugins that pass statically allocated strings and global `user_data_t` structs to `plugin_register_write()` it is not correct for all plugins.

See #4098 for an example of a plugin (`write_http`) mis-behaving due to this.

Store owned versions of the passed values instead. For `user_data` this means the content of the struct and for the name it means a `strdup()`ed version of the string.

This should fix the `write_http` plugin crashing in #4098.

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -105,9 +105,9 @@ typedef struct write_queue_elem_s {
 typedef struct write_queue_thread_s {
   bool loop;
   long queue_length;
-  const char *name;
+  char *name;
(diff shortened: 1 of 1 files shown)


mrunge (approved):

mrunge:
thanks!

PR #4103, opened 2023-03-24
[collectd 6] src/daemon/plugin.c: init fam.{time,interval} before uc_update(fam) on dispatch

hnez:
ChangeLog: fix uc_update: Value too old errors on metric dispatch

The changes in commit 55efb56a changed the order in which the time and interval on a to-be-dispatched metric family were set and `uc_update()` was called on the metric family.

The time and interval inside a metric family have to be set before calling `uc_update()` on it, otherwise warnings [like these][1] will be generated en masse:

    uc_update: Value too old: … value time = 0.000; last cache update = 0.000;

And the "missing" handlers for the metric family will be called resulting in plugins like `write_prometheus` dropping the value, resulting in it not showing up when queried, as mentioned in #4098.

[1]: https://github.com/collectd/collectd/pull/4026#issuecomment-1463884930

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -1936,6 +1936,8 @@ EXPORT int plugin_write(const char *plugin, metric_family_t const *fam) {
     return EINVAL;
   }
 
+  /* Create a copy of the metric_family_t so we can metric_family_free() it
+   * ourself once it is processed. */
   metric_family_t *fam_copy = metric_family_clone(fam);
   if (fam_copy == NULL) {
     int status = errno;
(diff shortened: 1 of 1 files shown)


eero-t:
@mrunge I've tested this and it fixes Prometheus plugin log spam.

To me the changes look fine.

I'm of two minds whether the description should be in code or in commit message like it's now (in code it can get obsolete, but one can find it easier than from a commit message) ,

There are few changes that could be done for the commit message though:
* inroduction -> introduction
* Add fix tags for the bug report and the commit introducing the bug
```
Fixes: https://github.com/collectd/collectd/commit/55efb56a8d56f5b37bfdad430801e731210ecd1f
Indicated commit changed the order ...
```

mrunge (approved):
thank you

PR #4104, opened 2023-03-24
[collectd 6] src/write_http.c: use reference counting to decide when to free user_data

hnez:
ChangeLog: Use reference counting in write_http to make plugin teardown more robust

The teardown code for the `wh_callback_t` struct previously relied on the order in which the different callback functions are de-initialized to be known and to never change.

This is prone to failure and is indeed currently broken, leading to a segmentation fault on `collectd` exit.

Fix this by counting the active references to the user data and freeing it once it reaches zero.

### DIFF
--- a/src/write_http.c
+++ b/src/write_http.c
@@ -92,6 +92,8 @@ struct wh_callback_s {
   char *metrics_prefix;
 
   char *unix_socket_path;
+
+  size_t reference_count;
 };
 typedef struct wh_callback_s wh_callback_t;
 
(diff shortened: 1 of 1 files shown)


hnez:
I've had a second look at this PR (mostly because I was unsure about maybe requiring locks around the reference counting).
While investigating I found out that the previous behavior of `write_http.c` (only setting `user_data.free_func` in the `plugin_register_write()` call and relying on the teardown order to prevent a double-free) is actually recommended in `plugin.c`:

https://github.com/collectd/collectd/blob/8286d53f7cab2457ddb491153345700451603ae3/src/daemon/plugin.c#L1941-L1949

And multiple other plugins do it this way:

- [`write_riemann`](https://github.com/collectd/collectd/blob/collectd-6.0/src/write_riemann.c#L807)
- [`write_stackdriver`](https://github.com/collectd/collectd/blob/collectd-6.0/src/write_stackdriver.c#L689)
- [`write_syslog`](https://github.com/collectd/collectd/blob/collectd-6.0/src/write_syslog.c#L604)
- [`write_tsdb`](https://github.com/collectd/collectd/blob/collectd-6.0/src/write_tsdb.c#L615)

I am preparing a pull request that restores the original teardown order between write and flush to fix the double free in all affected plugins.

Closing this PR as it only fixes part of the problem.

PR #4106, opened 2023-04-06
Fix glibc feature macro handling for timegm

fweimer-rh:
ChangeLog: Avoid an implicit declaration of timegm in src/bind.c, and build failures with future compilers

The way strptime is activated using feature macros, _DEFAULT_SOURCE (successor to _BSD_SOURCE) is disabled implicitly, so timegm is hidden.  Defining _DEFAULT_SOURCE at the same time as the other feature macros solves this, and removes the need for the TIMEGM_NEEDS_BSD configure macro.

This avoids an implicit declaration of timegm in src/bind.c, and build failures with future compilers.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

[edit mrunge: added a changelog line]

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -974,6 +974,12 @@ if test "x$have_strptime" = "xyes" && test "x$c_cv_have_strptime_default" = "xno
               #ifndef _XOPEN_SOURCE
               # define _XOPEN_SOURCE 500
               #endif
+              # ifndef _BSD_SOURCE
+              #  define _BSD_SOURCE
+              # endif
+              # ifndef _DEFAULT_SOURCE
+              #  define _DEFAULT_SOURCE
+              # endif
               #include <time.h>
             ]],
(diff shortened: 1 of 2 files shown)


mrunge (approved):
Thank you!

eero-t:
Your changes concern two functions used *only* in bind plugin:
```
$ git grep -l -e strptime -e timegm
configure.ac
src/bind.c
```

Whereas virt and mount plugins have known issues with newer RedHat distros, and smart & netlink plugins with newer Debian versions, so CI failures are not a problem.

@mrunge?

mrunge:
@fweimer-rh you don't need to do anything; the CI is in a miserable state and has known failures; they are not caused by this PR.

PR #4109, opened 2023-04-20
[collectd 6] Sysman plugin improvements

eero-t:
ChangeLog: gpu_sysman plugin: Misc improvements

Improvements:
* Much improved logging of enabled/disabled metrics [1]
* Log device (memory) ECC state at start
* Add memory types from L0 v1.3 spec
* Fix error value for power limit failure log message
----

[1] Plugin supports large number of Sysman provided metrics.  It tries each metric, logs failure for each missing one, and disables further queries for that.  However, it could be hard to determine which metrics are actually enabled in the end, especially for older integrated GPUs which provide data only for couple of metrics.

After this change, plugin will log a list of the enabled (and disabled) metrics at end of query round, if the enabled metric set changed (which should happen only few times start startup).

### DIFF
--- a/src/gpu_sysman.c
+++ b/src/gpu_sysman.c
@@ -280,6 +280,8 @@ static unsigned int list_gpu_metrics(const gpu_disable_t *disabled) {
     if (!disabled) {
       INFO("- none");
     }
+  }
+  if (config.gpuinfo) {
     INFO("Enabled metrics:");
   }
   unsigned int enabled = 0;
@@ -290,9 +292,9 @@ static unsigned int list_gpu_metrics(const gpu_disable_t *disabled) {
       }
       enabled++;
     }
-    if (config.gpuinfo && !enabled) {
-      INFO("- none");
-    }
+  }
+  if (config.gpuinfo && !enabled) {
+    INFO("- none");
   }
   return enabled;
 }
--- a/src/gpu_sysman.c
+++ b/src/gpu_sysman.c
@@ -249,6 +249,54 @@ static int gpu_config_free(void) {
   return RET_OK;
 }
 
+/* list GPU metric options that can get disabled at run-time */
+static unsigned int list_gpu_metrics(const gpu_disable_t *disabled) {
+  struct {
+    bool disabled;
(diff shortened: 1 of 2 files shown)


src/gpu_sysman.c
```diff
@@ -249,6 +249,54 @@ static int gpu_config_free(void) {
...
+    for (i = 0; i < STATIC_ARRAY_SIZE(names); i++) {
+      if (names[i].disabled) {
+        INFO("- %s%s", names[i].name, names[i].submetric ? " (submetric)" : "");
+        disabled++;
+      }
+    }
+    if (!disabled) {
+      INFO("- none");
+    }
+    INFO("Enabled metrics:");
```
octo, line 283: Please move this out of this `if` block. The enabled metrics are printed unconditionally, and so should this line.

src/gpu_sysman.c
```diff
@@ -249,6 +249,54 @@ static int gpu_config_free(void) {
...
+  }
+  unsigned int enabled = 0;
+  for (i = 0; i < STATIC_ARRAY_SIZE(names); i++) {
+    if (!(names[i].disabled || names[i].submetric)) {
+      if (config.gpuinfo) {
+        INFO("- %s", names[i].name);
+      }
+      enabled++;
+    }
+    if (config.gpuinfo && !enabled) {
```
octo, line 293: Please move this out of the `for` block, otherwise `- none` will be printed for each disabled metric at the beginning of the list. Ironically, it will *not* be printed when there are no metrics, because the loop is never entered.

src/gpu_sysman.c
eero-t: All metrics printing *is* conditional on `config.gpuinfo`.  But I moved it out, as code is clearer that way.

src/gpu_sysman.c
eero-t: Thanks, I fixed this.

("none" _was_ printed when there were no metrics enabled though, as loop is over whole fixed size array.)

mrunge (approved):

Issue #3980, opened 2022-02-28
build.sh: duplicate '--datarootdir'


Brown-LiuZan:
./configure \
      --prefix="${INSTALL_DIR}" \
      --libdir="${LIBDIR}" \
      --bindir="${BINDIR}" \
      --sbindir="${SBINDIR}" \
      --sysconfdir="${SYSCONFDIR}" \
      --localstatedir="${LOCALSTATEDIR}" \
      --datarootdir="${DATAROOTDIR}" \
      --datarootdir="${DATADIR}" \    <--- Should this second datarootdir be datadir?
      --disable-all-plugins \
      --host="mingw32" \
      --enable-logfile \
      --enable-match_regex \
      --enable-target_replace \
      --enable-target_set

mrunge:
Oh, good catch! Thank you for reporting it.

ardi0x01:
Hi @mrunge, Is anyone working on this? I would like to start my contribution to collectd. Can I pick this up?

mrunge:
Sure, go ahead. I vaguely recall seeing a PR though, maybe you'd check it yourself?

ardi0x01:
umm ok.. there any communication we use? like slack or etc?  Can you please help me to reproduce this?

Daksh011:
is it a work in progress?

ilovethensa:
maybe close this?

PR #4112, opened 2023-04-25
duplicated --datarootdir

mogulcoder26:
### With reference to #3980 
ChangeLog: Foo plugin: A specific issue people had has been fixed.
> Created a duplicate root of the directory tree and initialized it to ${DATADIR}\ as mentioned.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -139,7 +139,7 @@ build_cygwin()
       --sysconfdir="${SYSCONFDIR}" \
       --localstatedir="${LOCALSTATEDIR}" \
       --datarootdir="${DATAROOTDIR}" \
-      --datarootdir="${DATADIR}"\
+      --datarootdir="${DATADIR}" \
       --disable-all-plugins \
       --host="mingw32" \
       --enable-logfile \


build.sh
```diff
@@ -139,6 +139,7 @@ build_cygwin()
       --sysconfdir="${SYSCONFDIR}" \
       --localstatedir="${LOCALSTATEDIR}" \
       --datarootdir="${DATAROOTDIR}" \
+      --datarootdir="${DATADIR}"\
```
vincele, line 142:
```suggestion
      --datarootdir="${DATADIR}" \
```
keep the spaces consistent...

vincele (approved):
I approve the whitespace fix, but I don't really know why the second `--datarootdir` parameter is needed, maybe you can add more context in the commit message explaining the problem this fixes...

mrunge:
IMHO the DATADIR option should go away. It looks like that is a duplication of the DATAROOTDIR option.
```
  --datarootdir=DIR       read-only arch.-independent data root [PREFIX/share]
  --datadir=DIR           read-only architecture-independent data [DATAROOTDIR]
```

Issue #319, opened 2013-05-11
Add a way to collect HTTP status codes


jan-matejka:
possibly with curl pugin?

PR #4113, opened 2023-04-26
processes: add contextswitches a configurable option & use buffer to read /proc/stat 's content only once

zzzyhtheonly:
Changelog: add contextswitches a configurable option & use buffer to read /proc/stat 's content only once

We already have a plugin "contextswitch" to collect "ctxt" in "/proc/stat" in Linux. But since we also read from "/proc/stat" from plugin "processes". We add an option to enable "contextswitch" in plugin "processes", default disabled. For this feature, please see 58a1e09e88f2cce1640a4072f7527c54832901d6 

Also, we find that we open and read "/proc/stat" repeatedly when we read "procs_running" and "fork_rate", so we also patch this by only reading "/proc/stat" once into a buffer and making "procs_running", "fork_rate" and "contextswitch" read from this buffer. For this feature, please see de2d033ba5ccd1531caab2e3772d471c219671eb

Tested on my local Linux machines; please help review. Thanks a lot!

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -8037,6 +8037,7 @@ B<Synopsis:>
    CollectFileDescriptor  true
    CollectContextSwitch   true
    CollectDelayAccounting false
+   CollectSystemContextSwitch false
    Process "name"
    ProcessMatch "name" "regex"
    <Process "collectd">
@@ -8094,6 +8095,13 @@ Disabled by default.
 Collect the number of memory mapped files of the process.
 The limit for this number is configured via F</proc/sys/vm/max_map_count> in
 the Linux kernel.
+
+=item B<CollectSystemContextSwitch> I<Boolean>
+
+Collect the number of context switches at the system level.
+Collect ctxt fields from /proc/stat in linux systems.
+The option can only be configured outside the B<Process> and B<ProcessMatch>
+blocks.
 
 =back
 
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -8100,7 +8100,7 @@ the Linux kernel.
 
 Collect the number of context switches at the system level.
 Collect ctxt fields from /proc/stat in linux systems.
-The option can only be configured outside the B<Process> and B<ProcessMatch>
+Can be configured only outside the B<Process> and B<ProcessMatch>
 blocks.
 
 =back
--- a/src/processes.c
+++ b/src/processes.c
@@ -303,7 +303,7 @@ static bool report_ctx_switch;
 static bool report_fd_num;
 static bool report_maps_num;
 static bool report_delay;
-static bool report_sys_ctx_switch;
+static bool report_sys_ctxt_switch;
 
 #if HAVE_THREAD_INFO
 static mach_port_t port_host_self;
@@ -735,8 +735,8 @@ static int ps_config(oconfig_item_t *ci) {
       WARNING("processes plugin: The plugin has been compiled without support "
               "for the \"CollectDelayAccounting\" option.");
 #endif
-    } else if(strcasecmp(c->key, "CollectSystemContextSwitch") ==0) {
-      cf_util_get_boolean(c, &report_sys_ctx_switch);
+    } else if (strcasecmp(c->key, "CollectSystemContextSwitch") == 0) {
+      cf_util_get_boolean(c, &report_sys_ctxt_switch);
     } else {
       ERROR("processes plugin: The `%s' configuration option is not "
             "understood and will be ignored.",
@@ -977,7 +977,10 @@ static void ps_submit_proc_list(procstat_t *ps) {
 } /* void ps_submit_proc_list */
 
 #if KERNEL_LINUX || KERNEL_SOLARIS
-static void ps_submit_global_stat(const char* type_name, derive_t value) {  // change: ps_submit_fork_rate -> ps_submit_global_stat. In addtion, add a argument: char* type_name.
+static void ps_submit_global_stat(
+    const char *type_name,
+    derive_t value) { // change: ps_submit_fork_rate -> ps_submit_global_stat.
+                      // In addtion, add a argument: char* type_name.
   value_list_t vl = VALUE_LIST_INIT;
 
   vl.values = &(value_t){.derive = value};
@@ -1452,19 +1455,11 @@ static int ps_read_process(long pid, process_entry_t *ps, char *state) {
   return 0;
 } /* int ps_read_process (...) */
 
-static int procs_running(void) {
-  char buffer[65536] = {};
+static int procs_running(const char *buffer) {
   char id[] = "procs_running "; /* white space terminated */
   char *running;
   char *endptr = NULL;
   long result = 0L;
-
-  ssize_t status;
-
-  status = read_file_contents("/proc/stat", buffer, sizeof(buffer) - 1);
-  if (status <= 0) {
-    return -1;
-  }
 
   /* the data contains :
    * the literal string 'procs_running',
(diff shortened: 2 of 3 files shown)


src/processes.c
```diff
@@ -1574,42 +1572,58 @@ static char *ps_get_cmdline(long pid, char *name, char *buf, size_t buf_len) {
...
-  proc_stat = fopen("/proc/stat", "r");
-  if (proc_stat == NULL) {
-    ERROR("processes plugin: fopen (/proc/stat) failed: %s", STRERRNO);
+  int status;
+  char *fields[2];
+  int fields_num;
+
+  processes = strstr(buffer, id);
+  if (!processes) {
+    WARNING("processes not found");
```
vincele, line 1586:
```suggestion
    WARNING("'processes ' not found in /proc/stat");
```
Maybe a bit uglier, but conveys more info.

If you think it is better like that, there are other occurrences of the same kind of warning.

src/processes.c
```diff
@@ -1574,42 +1572,58 @@ static char *ps_get_cmdline(long pid, char *name, char *buf, size_t buf_len) {
...
+  fields_num = strsplit(ctxt, fields, STATIC_ARRAY_SIZE(fields));
+  if (fields_num != 2)
+    return -1;
+
+  status = parse_value(fields[1], &value, DS_TYPE_DERIVE);
+  if (status != 0)
     return -1;
 
-  ps_submit_fork_rate(value.derive);
+  if (report_sys_ctxt_switch)
```
vincele, line 1625: Naive question: why not exit earlier if the value is not to be reported ?

src/processes.c
```diff
@@ -2139,6 +2151,10 @@ static int ps_read(void) {
 
   closedir(proc);
 
+  if (read_file_contents("/proc/stat", buffer, sizeof(buffer) - 1) <= 0) {
+    ERROR("Cannot open `/proc/stat`");
```
vincele, line 2155:
```suggestion
    ERROR("Cannot read `/proc/stat`");
```

src/processes.c
```diff
@@ -2159,7 +2175,8 @@ static int ps_read(void) {
   for (procstat_t *ps_ptr = list_head_g; ps_ptr != NULL; ps_ptr = ps_ptr->next)
     ps_submit_proc_list(ps_ptr);
 
-  read_fork_rate();
+  read_fork_rate(buffer);
+  read_sys_ctxt_switch(buffer);
```
vincele, line 2179: maybe this is where the `if (report_sys_ctxt_switch)` should be ?

vincele (commented):
LGTM

src/processes.c
vincele: see my other comment below...

zzzyhtheonly:
@vincele Thank you! All resolved, please help review again.

src/collectd.conf.pod
```diff
@@ -8095,6 +8096,13 @@ Collect the number of memory mapped files of the process.
 The limit for this number is configured via F</proc/sys/vm/max_map_count> in
 the Linux kernel.
 
+=item B<CollectSystemContextSwitch> I<Boolean>
+
+Collect the number of context switches at the system level.
+Collect ctxt fields from /proc/stat in linux systems.
+The option can only be configured outside the B<Process> and B<ProcessMatch>
```
eero-t, line 8103: Simpler as:
```suggestion
Can be configured only outside the B<Process> and B<ProcessMatch>
```

src/processes.c
```diff
@@ -1574,42 +1572,57 @@ static char *ps_get_cmdline(long pid, char *name, char *buf, size_t buf_len) {
...
+  ctxt = strstr(buffer, id);
+  if (!ctxt) {
+    WARNING("'ctxt ' not found in /proc/stat");
+    return -1;
   }
-  fclose(proc_stat);
 
-  if (!value_valid)
+  fields_num = strsplit(ctxt, fields, STATIC_ARRAY_SIZE(fields));
+  if (fields_num != 2)
```
eero-t, line 1618: Seems odd to hard code value on other line.  Maybe do it like this (untested):
```suggestion
  const int expected = STATIC_ARRAY_SIZE(fields);
  fields_num = strsplit(ctxt, fields, expected);
  if (fields_num != expected)
```

src/processes.c
```diff
@@ -1469,7 +1469,7 @@ static int procs_running(const char *buffer) {
    */
   running = strstr(buffer, id);
   if (!running) {
-    WARNING("procs_running not found");
+    WARNING("'processes ' not found in /proc/stat");
```
vincele, line 1472:
```suggestion
    WARNING("'procs_running ' not found in /proc/stat");
```

src/processes.c
```diff
@@ -2176,8 +2175,9 @@ static int ps_read(void) {
     ps_submit_proc_list(ps_ptr);
 
   read_fork_rate(buffer);
-  read_sys_ctxt_switch(buffer);
-  /* #endif KERNEL_LINUX */
+  if (report_sys_ctxt_switch)
+    read_sys_ctxt_switch(buffer);
+    /* #endif KERNEL_LINUX */
```
vincele, line 2182: Is this (commented `#endif`) useful ?

vincele (changes requested):
I think this still needs some small fixes

src/processes.c
```diff
@@ -974,14 +977,17 @@ static void ps_submit_proc_list(procstat_t *ps) {
 } /* void ps_submit_proc_list */
 
 #if KERNEL_LINUX || KERNEL_SOLARIS
-static void ps_submit_fork_rate(derive_t value) {
+static void ps_submit_global_stat(
+    const char *type_name,
+    derive_t value) { // change: ps_submit_fork_rate -> ps_submit_global_stat.
+                      // In addtion, add a argument: char* type_name.
```
vincele, line 983: Why is that comment needed ?

src/processes.c
zzzyhtheonly: I guess this makes it clearer where the end of the "KERNEL_LINUX" part is. Since this kind of comment exists in other places, e.g., #319, #786, maybe we should keep it for this PR?

BTW, the starting spaces are caused by format.sh

src/processes.c
zzzyhtheonly: Removed

src/processes.c
zzzyhtheonly: Sorry for my carelessness, fixed.

vincele (commented):
This diff has become hard to read, but it LGTM...

src/processes.c
```diff
@@ -2159,8 +2173,10 @@ static int ps_read(void) {
...
     ps_submit_proc_list(ps_ptr);
 
-  read_fork_rate();
-  /* #endif KERNEL_LINUX */
+  read_fork_rate(buffer);
+  if (report_sys_ctxt_switch)
+    read_sys_ctxt_switch(buffer);
+    /* #endif KERNEL_LINUX */
 
 #elif HAVE_LIBKVM_GETPROCS && HAVE_STRUCT_KINFO_PROC_FREEBSD
```
eero-t, line 2181: Does "format.sh" work better if that comment is associated with the code it refers to?
```suggestion
    read_sys_ctxt_switch(buffer);

  /* #endif KERNEL_LINUX */
#elif HAVE_LIBKVM_GETPROCS && HAVE_STRUCT_KINFO_PROC_FREEBSD
```

PS. this this is not the only place where such comments are wrongly associated & indented with the last code line, instead of the #elif/#else/#endif that actually terminates the preceeding #if/#elif/#else block.

Also, IMHO having `#endif` in the comments is misleading as it looks like commented-code instead of just a comment.  Esing e.g. "KERNEL_LINUX ends" would be clearer.

zsfz: I agree with your point, but it seems like this annotation style is prevalent in many plugins such as cpu plugin. Should we strive for consistency?

eero-t: IMHO:
* If that fixes the indenting done by `format.sh`, that would be nice, but that should be a separate PR, and some comment from current project maintainers would be nice too
* Above KERNEL_LINUX "define end" comment does not need to be changed in this PR, as it's how rests of the plugin code is written currently

zzzyhtheonly: Agreed, so for now I'll leave this comment unchanged here for this PR.

eero-t:
@mrunge review feedback so far has been handled, do you want something still to be done before you merge?

mrunge:
thank you, this looks good to me, and thank you for the effort following up with this.

PR #4116, opened 2023-05-26
vmem: add metrics starting with `pgrefill`

zzzyhtheonly:
Changelog: add metrics starting with `pgrefill`

Same issue as I submit here: https://github.com/collectd/collectd/pull/4093
In newer Linux Kernel, there is only `pgrefill` instead of `pgrefill_*`, so we should add `pgrefill` into consideration. 

But maybe we should discuss if we have to keep the previous `pgrefill_` logic? Since `pgrefill` would also add all `pgrefill_*` into `refill`. 

Please help review, thanks a lot!


mrunge (approved):
thank you!

The ci failures have nothing to do with this PR.

PR #4117, opened 2023-06-02
[collectd 6] Fix callback freeing order in plugin_shutdown_all()

hnez:
ChangeLog: fix segfaults in plugin_shutdown_all() due to use-after-free

This fixes an issue I've introduced in #4026, which changed the order in which plugin callbacks are freed in `plugin_shutdown_all()`.

Without these changes some plugins will segfault during `plugin_shutdown_all()` e.g. when stopping collectd:

```
[2023-06-02 14:37:34] [debug] plugin_read_thread: Next read of the `flush/write_http/HttpWriter' plugin …
^C
[2023-06-02 14:37:35] [info] Exiting normally.
[2023-06-02 14:37:35] [info] collectd: Stopping 5 read threads.
[2023-06-02 14:37:35] [debug] plugin: stop_read_threads: Signalling `read_cond'
[2023-06-02 14:37:35] [debug] plugin_write_thread(write_prometheus): teardown
[2023-06-02 14:37:35] [debug] plugin_write_thread(write_http/HttpWriter): teardown
fish: “./install/sbin/collectd -f -C /…” terminated by signal SIGSEGV (Address boundary error)
```

This is due to the plugins assuming a specific order in which callbacks (and associated `user_data`) are freed, in accordance with the following comment:

https://github.com/collectd/collectd/blob/8286d53f7cab2457ddb491153345700451603ae3/src/daemon/plugin.c#L1941-L1949

I found at least the following plugins using this trick to prevent double-frees during teardown. Which means they will likely currently produce a SIGSEGV when stopping collectd:

 - `write_http`
 - `write_riemann`
 - `write_stackdriver`
 - `write_syslog`
 - `write_tsdb`

This should fix use-after-frees during teardown of these plugins by restoring the freeing order to what it was before 55efb56.

I've made a previous attempt to fix this issue for `write_http`  (#4104) that I've replaced with this PR.

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -2024,6 +2024,7 @@ EXPORT int plugin_flush(const char *plugin, cdtime_t timeout,
 } /* int plugin_flush */
 
 EXPORT int plugin_shutdown_all(void) {
+  write_queue_thread_t *write_threads_to_free = NULL;
   llentry_t *le;
   int ret = 0; // Assume success.
 
@@ -2038,8 +2039,11 @@ EXPORT int plugin_shutdown_all(void) {
 
   destroy_read_heap();
 
-  /* blocks until all write threads have shut down. */
-  plugin_unregister_write(NULL);
+  /* blocks until all write threads have stopped.
+   * Do not free the resources associated with this thread yet
+   * (the `user_data` provided via plugin_register_write in particular),
+   * as plugins rely on the order in which callback's user data is freed. */
+  stop_write_threads(NULL, &write_threads_to_free);
 
   /* ask all plugins to write out the state they kept. */
   plugin_flush(/* plugin = */ NULL,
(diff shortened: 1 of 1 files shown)


eero-t:
Ok, so plugin resource freeing is now done (using new helper functions) when either `plugin_shutdown_all()` or `plugin_unregister_write()` is called, instead of (directly) at the end of `plugin_write_thread()`.   That seems correct to me, but I'm not really familiar with that code.  @mrunge?

Btw. @hnez Have you done any other work / more testing on the `collecd-6.0` branch since this?

eero-t:
> Ok, so plugin resource freeing is now done (using new helper functions) when either `plugin_shutdown_all()` or `plugin_unregister_write()` is called, instead of (directly) at the end of `plugin_write_thread()`. That seems correct to me, but I'm not really familiar with that code. @mrunge?

@mrunge ?

octo:
I feel like this is working around broken plugins. The "only specify the `free` callback in one specific callback" approach was already hacky (a leaking abstraction), and adding further hacks to retain this behavior is not the best path forward.

I think the proper approach is to use reference counting and pass free callback to each callback registration. This way the callbacks can get shut down in any order and the plugin doesn't make any assumptions about the inner workings of the daemon.

I have implemented a demonstration of how this could work in #4176. What do you think?

octo:
Meh, I only now saw #4104 – sorry for duplicating the effort :(

octo:
Closing this in favor of #4176

PR #4122, opened 2023-06-16
processes: collect PSS memory usage for each process

TeknoVenus:
ChangeLog: Processes: Collect PSS memory usage per process

If available, will use the smaps_rollup file for performance reasons. Otherwise will use the smaps file and sum up all the PSS fields (which is just what the smaps_rollup file is doing internally).

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -783,9 +783,9 @@ static int ps_init(void) {
   // smaps_rollup requires kernel >= 4.14
   struct stat sb;
   if (stat("/proc/1/smaps_rollup", &sb) == 0) {
-      use_smaps_rollup = true;
+    use_smaps_rollup = true;
   } else {
-      use_smaps_rollup = false;
+    use_smaps_rollup = false;
   }
 
 #if HAVE_LIBTASKSTATS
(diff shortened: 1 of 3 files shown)


src/processes.c
```diff
@@ -1226,6 +1244,55 @@ static int ps_count_fd(int pid) {
...
+
+    numfields = strsplit(buffer, fields, STATIC_ARRAY_SIZE(fields));
+    if (numfields < 2) {
+      continue;
+    }
+
+    errno = 0;
+    endptr = NULL;
+    tmp = strtoul(fields[1], &endptr, 10);
+    if ((errno == 0) && endptr != fields[1]) {
```
eero-t, line 1282: This should chould `endptr` for 'nil' character, as according to `strtoul()` manual page:
> if *nptr is not '\0' but **endptr is '\0' on return, the entire string is valid

And apparently all implementations might not set errno:
> If there were no digits at all, strtoul() stores the original value of nptr in *endptr (and returns 0).
> ...
> The implementation **may** also set errno to EINVAL in case no conversion was performed (no digits seen, and 0 returned).

src/processes.c
```diff
@@ -775,6 +780,14 @@ static int ps_init(void) {
   pagesize_g = sysconf(_SC_PAGESIZE);
   DEBUG("pagesize_g = %li; CONFIG_HZ = %i;", pagesize_g, CONFIG_HZ);
 
+  // smaps_rollup requires kernel >= 4.14
+  struct stat sb;
+  if (stat("/proc/1/smaps_rollup", &sb) == 0) {
```
eero-t, line 785: You could add link to doc to the comment:
https://www.kernel.org/doc/Documentation/ABI/testing/procfs-smaps_rollup

And skip dummy variable use by using `access()` instead of `stat()`.

src/processes.c
```diff
@@ -1226,6 +1244,55 @@ static int ps_count_fd(int pid) {
...
+    if (numfields < 2) {
+      continue;
+    }
+
+    errno = 0;
+    endptr = NULL;
+    tmp = strtoul(fields[1], &endptr, 10);
+    if ((errno == 0) && endptr != fields[1]) {
+      pss += tmp;
+    }
```
eero-t, line 1284: In rollup case, loop can be once the value has been parsed.

eero-t (commented):
Looks fine except for few nit-picks, but I would suggest parsing also (at least) `Private_Dirty` from Smaps file at the same time, as all this functionality just for `Pss` value seem a bit waste.

I assume this to (at least) double data parsing overhead (both user-space + kernel side), but `Pss` and `Private_Dirty` are _much_ more useful data than the earlier provided values, so it's a good compromise.

src/processes.c
```diff
@@ -1226,6 +1244,55 @@ static int ps_count_fd(int pid) {
   return (count >= 1) ? count : 1;
 } /* int ps_count_fd (pid) */
 
+static int ps_get_pss(process_entry_t *ps) {
+  FILE *fh;
+  char buffer[1024];
+  char filename[64];
+  char *fields[4];
+  int numfields;
+  unsigned long pss = 0;
```
octo, line 1253: Please define variables as late as possible.

https://github.com/collectd/collectd/blob/main/docs/review_comments.md#define-variables-on-first-use

src/processes.c
```diff
@@ -1411,6 +1478,12 @@ static int ps_read_process(long pid, process_entry_t *ps, char *state) {
     return 0;
   }
 
+  if (ps_get_pss(ps) != 0) {
+    /* no pss data */
+    ps->vmem_pss = -1;
```
octo, line 1483: `vmem_pss` is an `unsigned long`. Assigning -1 here will cause problems with compilers. I suggest to use zero here instead.

src/processes.c
```diff
@@ -843,6 +856,11 @@ static void ps_submit_proc_list(procstat_t *ps) {
   vl.values_len = 1;
   plugin_dispatch_values(&vl);
 
+  sstrncpy(vl.type, "ps_pss", sizeof(vl.type));
+  vl.values[0].gauge = ps->vmem_pss;
```
octo, line 860: Below, `vmem_pss` is set to a well-known value on error. Add a check for this well-known value here and set the gauge to `NAN` if there was an error:

```c
vl.values[0].gauge = ps->vmem_pss;
if (ps->vmem_pss == 0) {
  vl.values[0].gauge = NAN;
}
```

octo (commented):
Hi @TeknoVenus, thank you very much for your PR!

I think this looks quite good overall and with @eero-t and my comments addressed this should be good to go.

Thanks and best regards
—octo

src/processes.c
eero-t: I first thought there could be some tasks with zero PSS, but kernel tasks apparently do not have `smaps_rollup` files, and their `smaps` files are empty, so zero PSS seems OK indicator for missing data (otherwise I would have suggested `ULONG_MAX`).

PR #4124, opened 2023-06-22
Fix for smart plugin crashing on Intel NVMe drives

PeterSurda:
I have a server with an Intel NVMe, model SSDPE2MX450G7. The smart plugin crashes and takes down whole collectd with it. The error message I get is

```
*** stack smashing detected ***: terminated
Aborted (core dumped)
```

It looks like there is inadequate data allocated for the return values of an ioctl. This patch adds padding to the data, padding even one byte less will result in the crash mentioned above.

ChangeLog: Smart plugin: Fix crash on Intel NVMe


PhilippWendler:
This problem was also detected in the discussion of #3974 and https://github.com/collectd/collectd/pull/3974/commits/2e4e9c328eeceec5e368f15d2e13bb62e8c283b0 is an alternative fix.

octo:
Closing this in favor of #3974 which has more analysis and context.

Issue #780, opened 2014-10-27
nginx plugin: support several instances


landryb:
more a feature request than an actual bug.. there's been a patch a while ago on the collectd mailing list to support multiple nginx instances in the nginx plugin (see https://www.mail-archive.com/collectd@verplant.org/msg02497.html) and it seems to have been missed/ignored. The proposal was to use the same layout as the apache plugin with <instance> xml childs.

What would it take to get it properly integrated ?
What should be done for a migration from the old-style to the new-style (move rrds around) ?

pauvos:
No updates since 2014? I agree that nginx plugin should provide roughly the same functionality/syntax as the apache plugin.

There is also some free-of-context proposal at stackoverflow containing just 3 lines: http://stackoverflow.com/questions/20739397/adding-multiple-nginx-to-collectds-monitoring-plugin

Unfortunately I'm not an expert c-coder. :/

cvle:
That's rather disappointing :-(

octo:
Patches are always welcome. That snippet from StackOverflow doesn't implement this feature though.

octo:
Closing ancient feature request due to lack of interest.

PR #4125, opened 2023-06-29
Initial support for gathering metrics from multiple nginx instances.

arrnorets:
Several days ago I faced with a case in which I needed to gather /status metrics from multiple nginx instances with a single collectd instance, something like https://github.com/collectd/collectd/issues/780.

I created a patch to the current nginx plugin implementation that covers this above mentioned case. The main idea of the patch is to create a number of structures that store configured "InstanceName" and a CURL object containing options for curl-ing of each nginx instance separately. The config parser implementation was inspired by the one for ceph plugin. The only difference between them is a static limitation on a total number of nginx instances which can be served by plugin simultaneously. It's defined with a constant value:
```c
#define MAX_NGINX_INSTANCES 1000     // Max nginx instances that plugin can serve simultaneously
```

### DIFF
--- a/src/nginx.c
+++ b/src/nginx.c
@@ -33,25 +33,21 @@
 
 #include <curl/curl.h>
 
-static char *url;
-static char *user;
-static char *pass;
-static char *verify_peer;
-static char *verify_host;
-static char *cacert;
-static char *timeout;
-static char *sock;
+#define MAX_NGINX_INSTANCES 1000     // Max nginx instances that plugin can serve simultaneously
+static int instance_num_counter = 0; // Counts number of configured nginx instances
 
-static CURL *curl;
+/* Stores name and curl options of the nginx instance */
+
+static struct nginx_instance {
+    char name[DATA_MAX_NAME_LEN];
+    CURL *curl;
+} ngx_inst[MAX_NGINX_INSTANCES];
+
+/******************/
 
 static char nginx_buffer[16384];
 static size_t nginx_buffer_len;
 static char nginx_curl_error[CURL_ERROR_SIZE];
-
-static const char *config_keys[] = {"URL",        "User",       "Password",
-                                    "VerifyPeer", "VerifyHost", "CACert",
-                                    "Timeout",    "Socket"};
-static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
 
 static size_t nginx_curl_callback(void *buf, size_t size, size_t nmemb,
                                   void __attribute__((unused)) * stream) {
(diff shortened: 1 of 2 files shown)


src/nginx.c
```diff
@@ -73,108 +71,243 @@ static size_t nginx_curl_callback(void *buf, size_t size, size_t nmemb,
...
+   * plugin options */
+
+  char u[DATA_MAX_NAME_LEN] = {'\0'};
+  char p[DATA_MAX_NAME_LEN] = {'\0'};
+  char url[DATA_MAX_NAME_LEN] = {'\0'};
+  char verify_peer[DATA_MAX_NAME_LEN] = {'\0'};
+  char verify_host[DATA_MAX_NAME_LEN] = {'\0'};
+  char cacert[DATA_MAX_NAME_LEN] = {'\0'};
+  char timeout[DATA_MAX_NAME_LEN] = {'\0'};
+  char socket[DATA_MAX_NAME_LEN] = {'\0'};
```
eero-t, line 154: This function is nearly two hundred lines.  This could be good place to split it (at least) into two.

Single char variables should be used only as loop counters (i, j etc), anything else should have more descriptive names (especially in function this large).

`socket` overrides standard C-lib function name.  Maybe "sockname"?

Otherwise this looks pretty OK on quick glance.

eero-t (changes requested):

src/nginx.c
arrnorets: hank you for review and for your notes @eero-t .
Definitely you are right about nginx_add_daemon_config() function size and not very obvious names for temporary arrays. So in order to solve the second issue instead of using predefined number of static arrays I decided to return back ```static char *config_keys[]``` setting list and use dynamic ```**parsed_values_for_nginx_instance``` string list for parsing configuration file and further curl options settings. Additionally in terms of the first issue curl options were splitted into default ones ( that cannot be changed from configuration file ) and optional ones that are managed with config.

src/nginx.c
```diff
@@ -73,108 +85,261 @@ static size_t nginx_curl_callback(void *buf, size_t size, size_t nmemb,
...
+
+static int nginx_config(oconfig_item_t *ci) {
+  int ret;
+
+  parsed_values_for_nginx_instance =
+      (char **)calloc(config_keys_num, sizeof(char *));
+  for (int i = 0; i < config_keys_num; i++) {
+    parsed_values_for_nginx_instance[i] =
+        (char *)calloc(DATA_MAX_NAME_LEN, sizeof(char));
+  }
```
eero-t, line 306: You have deinit function for freeing the allocs, so there should be also corresponding init function doing the allocs (close together).

src/nginx.c
```diff
@@ -73,108 +85,261 @@ static size_t nginx_curl_callback(void *buf, size_t size, size_t nmemb,
...
+/******************/
+
+/* Cleanup per nginx instance temporary stringlist for storing settings from
+ * collectd config for each nginx instance  */
+static void deinit_nginx_instance_settings_array(void) {
+  for (int i = 0; i < config_keys_num; i++) {
+    if (parsed_values_for_nginx_instance[i] != NULL) {
+      sfree(parsed_values_for_nginx_instance[i]);
+      parsed_values_for_nginx_instance[i] = NULL;
+    }
```
eero-t, line 106: While using an array makes certain things easier, I do not really like how you did it.  You have no names for the indexes, they're just magic values, when you use them later.  I.e. it's really annoying to check whether they e.g. match.

I'd rather have  `parsed_values_for_nginx_instance` a struct, with named members matching to config keys, so you can refer to them by name in `set_optional_curl_params()`.

Helper functions doing init + deinit for those struct members, could have auxiliary array though, like this:
`static const char *nginx_inst_str_ptrs[config_keys_num] = { &parsed_values_for_nginx_instance.username, ... };`

(Such helper could also include the names of the keys, so you can use it in parsing too.)

src/nginx.c
```diff
@@ -73,108 +85,261 @@ static size_t nginx_curl_callback(void *buf, size_t size, size_t nmemb,
...
+  }
+  val = item->values[0].value.string;
+  if (snprintf(dest, dest_len, "%s", val) > (dest_len - 1)) {
+    ERROR("nginx plugin: configuration parameter '%s' is too long.\n",
+          item->key);
+    return -ENAMETOOLONG;
+  }
   return 0;
-} /* void init */
+}
```
eero-t, line 239: Earlier code did not use strings with fixed lengths, but took the strings from config as-is in `config_set()`.  I think the new code should do the same.

eero-t (commented):
While the function sizes are now OK, I don't really approve how the strings related to each instance are handled.  Unlike current NIGX code, new code has "magic" assumptions about indexes, and uses fixed string sizes.

PR #4130, opened 2023-08-02
add questions metric for mysql plugin

lurkr2:
mysql plugin: add "questions" metric
ChangeLog: mysql plugin: add "questions" metric


mrunge (approved):
thank you.

Today I learned, there are both: queries and questions in mysql.

PR #4133, opened 2023-08-14
plugin mysql: add multisource repliacation support

lurkr2:
mysql plugin: add multisource repliacation support
ChangeLog: mysql plugin: add multisource repliacation support

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -5487,6 +5487,7 @@ Synopsis:
       Host "localhost"
       Socket "/var/run/mysql/mysqld.sock"
       SlaveStats true
+      MariaDb false
       SlaveNotifications true
     </Database>
 
@@ -5561,6 +5562,10 @@ Disabled by default.
 Enable the collection of primary / replica statistics in a replication setup. In
 order to be able to get access to these statistics, the user needs special
 privileges. See the B<User> documentation above. Defaults to B<false>.
+
+=item B<MariaDb> I<true|false>
+
+Enable Multisource replication Mariadb support.
 
 =item B<SlaveNotifications> I<true|false>
 
--- a/src/mysql.c
+++ b/src/mysql.c
@@ -60,6 +60,7 @@ struct mysql_database_s /* {{{ */
 
   bool primary_stats;
   bool replica_stats;
+  bool mariadb;
   bool innodb_stats;
   bool wsrep_stats;
 
@@ -198,6 +199,8 @@ static int mysql_config_database(oconfig_item_t *ci) /* {{{ */
       status = cf_util_get_boolean(child, &db->replica_stats);
     else if (strcasecmp("SlaveNotifications", child->key) == 0)
       status = cf_util_get_boolean(child, &db->replica_notif);
+    else if (strcasecmp("MariaDb", child->key) == 0)
+      status = cf_util_get_boolean(child, &db->mariadb);
     else if (strcasecmp("InnodbStats", child->key) == 0)
       status = cf_util_get_boolean(child, &db->innodb_stats);
     else if (strcasecmp("WsrepStats", child->key) == 0)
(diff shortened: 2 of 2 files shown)


src/collectd.conf.pod
```diff
@@ -5487,6 +5487,7 @@ Synopsis:
       Host "localhost"
       Socket "/var/run/mysql/mysqld.sock"
       SlaveStats true
+      MariaDb false
```
mrunge, line 5490: Can we please name the option more descriptive? Like `MultiSourceReplication`?

src/mysql.c
```diff
@@ -60,6 +60,7 @@ struct mysql_database_s /* {{{ */
 
   bool primary_stats;
   bool replica_stats;
+  bool mariadb;
```
mrunge, line 63: same here then

src/mysql.c
```diff
@@ -433,121 +436,204 @@ static int mysql_read_primary_stats(mysql_database_t *db, MYSQL *con) {
 
 static int mysql_read_replica_stats(mysql_database_t *db, MYSQL *con) {
   MYSQL_RES *res;
+  MYSQL_FIELD *fields;
   MYSQL_ROW row;
 
-  const char *query;
-  int field_num;
+  const char *query, *channel_name_field;
+  unsigned char field_num, field_req;
```
eero-t, line 443: Why 8-bit variable instead of e.g. (unsigned) int?

eero-t (changes requested):
`mysql_read_replica_stats()` is now way too large / complex:
```
$ complexity mysql.c 
NOTE: proc mysql_read_replica_stats in file mysql.c line 429
	nesting depth reached level 5
Complexity Scores
Score | ln-ct | nc-lns| file-name(line): proc-name
   40     228     189   mysql.c(803): mysql_read
   53     193     177   mysql.c(429): mysql_read_replica_stats
```
It should be split into multiple functions.

(`mysql_config_database()` has also fairly high complexity score.  One can see it by using `pmccabe` instead of GNU `complexity`, or lowering latter's default threshold.)

src/collectd.conf.pod
lurkr2: the variable includes mariadb dialect, multisource-replication on the vanilla version is determined automatically

src/mysql.c
lurkr2: reduce usage ram, unsigned char - 1 byte, AFAIK

src/collectd.conf.pod
lurkr2: maybe MariaDbMultiSourceReplication?, or let's leave it as it is for the future.

src/mysql.c
lurkr2: [test.zip](https://github.com/collectd/collectd/files/12426226/test.zip)
docker-compose scenario for test mariadb/vanilla

src/collectd.conf.pod
mrunge: I am very in favor of having a more speaking name. MariaDb sounds very generic and may lead to misunderstandings.

lurkr2: maybe MariadbDialect, like telegraf (https://github.com/influxdata/telegraf/blob/master/plugins/inputs/mysql/mysql.go) ?

src/collectd.conf.pod
```diff
@@ -5563,9 +5563,9 @@ Enable the collection of primary / replica statistics in a replication setup. In
 order to be able to get access to these statistics, the user needs special
 privileges. See the B<User> documentation above. Defaults to B<false>.
 
-=item B<MariaDb> I<true|false>
+=item B<MariadbDialect> I<true|false>
 
-Enable Multisource replication Mariadb support.
+Enable Multisource replication MariadbDialect support.
```
mrunge, line 5568: This may sound like splitting hairs or overly critical, but what should "MariadbDialect" be? I think this needs a little bit more explanation, what exactly does this switch do?

src/mysql.c
eero-t: Performance claims should be backed up by evidence.  :-)

What e.g. "file collecd" states for compiled collectd code (text section) sizes when you use unsigned int vs. unsigned char?

src/collectd.conf.pod
lurkr2: ok, added like in telegraf :)

PR #4138, opened 2023-08-25
Change AMQP queue drops from DEBUG to WARNING

csibbitt:
Debug messages are only available if collectd is compiled with debug enabled, making it hard to troubleshoot the situation where the amqp queue is being overrun.

I think dropping metrics merits a "warning", but could be negotiated down to pretty much any level above debug :)


mrunge (approved):
thank you.


CI failures are not related to this change.

PR #4144, opened 2023-10-27
Improve the performance of the postgres insertion.

schorsch1976:
- The first version commited 7c5ce9f250aafbb6ef89769d7543ea155618b2ad was working ok but as the tables grew the postgres worker for that connection hit constantly 100% cpu usage on my small machine and also data was lost. So the functions needed to be optimized.

- The function collectd_insert() was split into 2 parts. One part just collecting the data, create the instances and realign the timestamps. Second part: move_data_to_table() that must be called periodically. It also creates the plugin tables and the asociated indices.

I also removed the readme part from the sql script and updated the README.md.

Development of the cpu load:
cpuload

More detail with the high load
cpuload1

With this setup in place
cpuload2

### DIFF
--- a/contrib/postgresql/README.md
+++ b/contrib/postgresql/README.md
@@ -18,10 +18,10 @@ The first database layout, from Sebastian 'tokkee' Harl, is like this:
 ```
 
 The ID connects the two tables. The plugin, plugin_inst, type and type_inst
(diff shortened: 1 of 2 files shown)


schorsch1976:
Need more work

PR #4148, opened 2023-11-07
FreeBSD's PF has a new interface so leverage libpfctl to access it so…

brd:
… the right interface is used depending on the version

I am interested in feedback if there is possibly a better way to do this.


src/pf.c
```diff
@@ -77,46 +80,84 @@ static void pf_submit(char const *type, char const *type_instance, uint64_t val,
 } /* void pf_submit */
 
 static int pf_read(void) {
+#ifdef __FreeBSD__
```
octo, line 83: This reads like two interleaved functions. It almost feels like one of those "Magic Eye" books ;)

Joking aside, this is a readability vs. maintainability trade-off. I think I would just duplicate the function, one for FreeBSD and one for other systems. This errs on the side of readability. Maintainability suffers because now the two functions could diverge, but I think they're simple enough that this is not a huge risk to take.

```c
#ifdef __FreeBSD__
static int pf_read(void) {
  // …
}
#else
static int pf_read(void) {
  // …
}
#endif
```

octo (changes requested):
Hi @brd, thanks for the contribution!

My overall concern is that this is making the code more complicated but doesn't provide additional utility. Is there anything that this new interface allows us to do that we were unable to do previously? If not, I think it would be preferable to keep the code simple.

src/pf.c
brd: Thanks for the feedback, it is exactly what I was looking for!  I will make that change.

brd:
OK, I have split it into two separate functions, let me know what you think!

octo:
This looks much better now, thank you! I'm still not quite clean on why we need special code for FreeBSD if is does the exact same as the generic code. What's the benefit?

brd:
libpfctl is only for FreeBSD, as the interface to access pf info is changing over time.. it was ioctl, then nvlist, and now it is netlink.  libpfctl abstracts that away so collectd does not need to have ifdef for the different versions of FreeBSD.

octo:
Hi @brd, do you need any help making the above changes?

brd:
@octo sorry for the delay, I kept meaning to get back to this and kept forgetting!

I am struggling with getting the configure script to work.. I tried this:

`
diff --git a/Makefile.am b/Makefile.am
index 1d3c69f2..71821bd0 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -1744,7 +1744,7 @@ endif
 if BUILD_PLUGIN_PF
 pkglib_LTLIBRARIES += pf.la
 pf_la_SOURCES = src/pf.c
-pf_la_LDFLAGS = $(PLUGIN_LDFLAGS)
+pf_la_LDFLAGS = $(PLUGIN_LDFLAGS) $(PF_LDFLAGS)
 endif
 
 if BUILD_PLUGIN_PINBA
diff --git a/configure.ac b/configure.ac
index 5383e3e2..f95fa1a4 100644
--- a/configure.ac
+++ b/configure.ac
@@ -693,6 +693,9 @@ AC_CHECK_HEADERS([net/pfvar.h],
     # include <netinet/in.h>
     #endif
   ]]
+  AC_CHECK_LIB([libpfctl], [pfctl_status_counter],
+    PF_LDFLAGS="$PF_LDFLAGS -lpfctl"
+  )
 )
`

octo:
Hey @brd, no worries. I think the following should work. Note: it's untested!

```diff
diff --git a/Makefile.am b/Makefile.am
index 64c719c48..6b0e4b4ea 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -1761,6 +1761,9 @@ if BUILD_PLUGIN_PF
 pkglib_LTLIBRARIES += pf.la
 pf_la_SOURCES = src/pf.c
 pf_la_LDFLAGS = $(PLUGIN_LDFLAGS)
+if BUILD_WITH_LIBPFCTL
+pf_la_LDFLAGS += -lpfctl
+endif
 endif

 if BUILD_PLUGIN_PINBA
diff --git a/configure.ac b/configure.ac
index dad27f53b..ae77bf3b5 100644
--- a/configure.ac
+++ b/configure.ac
@@ -2040,6 +2040,13 @@ if test "x$with_kvm_openfiles" = "xyes"; then
   with_libkvm="yes"
 fi

+AC_CHECK_HEADERS([libpfctl.h])
+AC_CHECK_LIB([libpfctl], [pfctl_status_counter],
+  [with_libpfctl="yes"],
+  [with_libpfctl="no"]
+)
+AM_CONDITIONAL([BUILD_WITH_LIBPFCTL], [test "x$with_libpfctl" = "xyes"])
+
 # --with-cuda {{{
 AC_ARG_WITH([cuda],
   [AS_HELP_STRING([--with-cuda@<:@=PREFIX@:>@], [Path to cuda.])],
diff --git a/src/pf.c b/src/pf.c
index 9681d366a..2dbcd31cc 100644
--- a/src/pf.c
+++ b/src/pf.c
@@ -36,6 +36,10 @@

 #include <net/pfvar.h>

+#if HAVE_LIBPFCTL_H
+#include <libpfctl.h>
+#endif
+
 #ifndef FCNT_NAMES
 #if FCNT_MAX != 3
 #error "Unexpected value for FCNT_MAX"
```

brd:
Hi @octo I think I got it working right.. have a look and let me know what you think.

configure.ac
```diff
@@ -2040,6 +2040,20 @@ if test "x$with_kvm_openfiles" = "xyes"; then
...
 fi
 
+AC_CHECK_HEADERS([libpfctl.h], [], [],
+[
+  #include <sys/queue.h>
+  #include <sys/types.h>
+  #include <netinet/in.h>
+  #include <net/if.h>
+  #include <net/pfvar.h>
+])
```
octo, line 2050: This check is inconsequential: neither the success nor the failure case have any side effects.

Option 1: If you simplify this to

```
AC_CHECK_HEADERS([libpfctl.h])
```

then autoconf's default behavior will be active, which will create a `HAVE_LIBPFCTL_H` define if the header is present. You can use it instead of `__FreeBSD__` in the .c file.

Option 2: In the unlikely case that you have to specify the other headers, skipping the arguments entirely will trigger the default behavior (I think):

```
AC_CHECK_HEADERS([libpfctl.h],,, # <-- empty arguments
[
  #include <sys/queue.h>
  #include <sys/types.h>
  #include <netinet/in.h>
  #include <net/if.h>
  #include <net/pfvar.h>
])
```

Option 3: remove entirely.

src/pf.c
```diff
@@ -76,6 +79,56 @@ static void pf_submit(char const *type, char const *type_instance, uint64_t val,
   plugin_dispatch_values(&vl);
 } /* void pf_submit */
 
+#ifdef __FreeBSD__
+static int pf_read(void) {
+  struct pfctl_status *state;
+  int fd;
+
+  fd = open(pf_device, O_RDONLY);
```
octo, line 87: Optional coding style improvement.

```suggestion

  int fd = open(pf_device, O_RDONLY);
```

octo (commented):
This looks much better, thank you @brd! Once we fix the `AC_CHECK_HEADERS` check this is good to go.

configure.ac
octo: Case in point, the build currently fails with:

```
src/pf.c:39:10: fatal error: 'libpfctl.h' file not found
#include <libpfctl.h>
         ^~~~~~~~~~~~
1 error generated.
--- src/pf.lo ---
*** [src/pf.lo] Error code 1
```

This is despite the configure script technically checking for the header's existence:

```
checking for libpfctl.h... no
checking for pfctl_status_counter in -llibpfctl... no
```

brd: Hmm, the build works for me where I have libpfctl present:

```
checking for libpfctl.h... yes
checking for pfctl_status_counter in -lpfctl... yes
```

I had to add the other includes so that the test worked.. but if we don't need it, I am happy to remove it.

Where are you seeing that build failure? It looks like the checks on this PR are successful?

octo: I see some failed checks below. The detailed view of the FreeBSD one is here: https://cirrus-ci.com/task/6128106226843648

brd: ah, thank you!  I am not very familiar with the github actions and didn't see where I could get that detail.. the in github link didn't show any errors, just looked like it built successfully.

In the FreeBSD port, we set `CPPFLAGS= -I/usr/local/include` :
https://cgit.freebsd.org/ports/tree/net-mgmt/collectd5/Makefile#n84

Looks like that is needed in the .cirrus.yml...

.cirrus.yml
```diff
@@ -232,10 +232,11 @@ freebsd_task:
       autotools
       bison
       flex
+      libpfctl
       pkgconf
   configure_script:
     - ./build.sh
-    - ./configure --disable-perl
+    - ./configure --disable-perl CPPLAGS='-I/usr/local/include'
```
octo, line 239:
```suggestion
    - ./configure --disable-perl CPPFLAGS='-I/usr/local/include'
```

configure.ac
octo: There is a typo in the flag definition (`CPPLAGS` instead of `CPPFLAGS`).

configure.ac
```diff
@@ -2040,6 +2040,20 @@ if test "x$with_kvm_openfiles" = "xyes"; then
...
+AC_CHECK_LIB([libpfctl], [pfctl_status_counter],
```
octo, line 2051: I think you need to drop the `lib` prefix: https://www.freshports.org/net/libpfctl/ says the shared object is called `lib/libpfctl.so`, so we need to link using `-lpfctl`.

```suggestion
AC_CHECK_LIB([pfctl], [pfctl_status_counter],
```

brd: yep, doing that now

src/pf.c
```diff
@@ -35,6 +35,9 @@
 #endif
 
 #include <net/pfvar.h>
+#ifdef __FreeBSD__
```
octo, line 38: We cannot unconditionally include `<libpfctl.h>` on FreeBSD, since the header is only available when the "libpfctl" package is installed. Please use `HAVE_LIBPFCTL_H` instead.

configure.ac
octo: We definitely need `HAVE_LIBPFCTL_H` and cannot rely on `__FreeBSD__`, see comment below. Please try if

```
AC_CHECK_HEADERS([libpfctl.h])
```

works, that would be the easiest.

octo (changes requested):
The build succeeds now. The remaining issue is in `src/ras.c` which is unrelated.

src/pf.c
brd: OK, let me try that.

brd:
I think that addresses all the build issues.. I kept the #includes for the `AC_CHECK_HEADERS` since the test failed for me without them.

octo (approved):
LGTM, thanks @brd!

PR #4153, opened 2023-11-21
write_prometheus: don't use AI_ADDRCONFIG for resolving bind address

### LINKED ISSUE #4150: write_prometheus fails to bind when the host has no non-loopback IP addresses
*   Version of collectd: 5.12.0
*   Operating system / distribution: Debian 11
*   Kernel version (if applicable): Linux 5.10.0-20-amd64 Debian 5.10.158-2

Collectd config contains:
```
LoadPlugin write_prometheus

[...]

<Plugin "write_prometheus">
  Port "9103"
</Plugin>
```

## Expected behavior

When collectd starts during the boot sequence, the write_prometheus plugin binds to a wildcard address (v4, v6, or sth in between) and port 9103, even if the network network interfaces haven't come up yet, and logs contain:
```
collectd[$pid]: write_prometheus plugin: Listening on [::]:9103.
```

## Actual behavior

When collectd starts before dhcp has obtained an address, and only loopback addresses are available, the write_prometheus plugin fails to bind. Collectd keeps running without that plugin. Logs say:
```
collectd[$pid]: write_prometheus plugin: Opening a listening socket for [::]:9103 failed.
collectd[$pid]: Initialization of plugin `write_prometheus' failed with status -1. Plugin will be unloaded.
```

If I manually restart collectd afterwards, it works fine.

Another workaround is to `ip addr add 127.0.0.2/32 dev lo` early in the boot, before collectd starts.

## Steps to reproduce

* install collectd form Debian 11 repository
* write a config containing the write_prometheus snippets mentioned earlier
* do either of these two:
   * method A
      * `systemctl enable collectd`
      * reboot
   * method B
      * log in from a local tty
      * ifdown all network interfaces except `lo`
      * make sure there are no extra IPs assigned to `lo`
      * systemctl (re)start collectd
* observe that collectd is not listening on port 9103 (check collectd logs, `ss -tlnp`, or try `curl http://localhost:9103`)

## Investigation

Looks like write_prometheus (both in 5.12.0 and on master) uses `getaddrinfo` to obtain a sockaddr to bind to
https://github.com/collectd/collectd/blob/3f935d017e81b7eaf84c9df421490230845dc5c0/src/write_prometheus.c#L752-L764

since https://github.com/collectd/collectd/commit/e11851791d3e82572ea2d9580a533f623116781d, apparently because on FreeBSD it's not feasible to construct a sockaddr_in6 by hand.

It passes the `AI_ADDRCONFIG` flag to `getaddrinfo`, which, according to the [Linux manpage][Debian manpage], ensures that

> IPv4 addresses are returned in the list pointed to by res only if the local system has at least one IPv4 address configured
> and IPv6 addresses are returned only if the local system has at least one  IPv6 address configured.
> The loopback address is not considered for this case as valid as a configured address.

([FreeBSD manpage] has a similar snippet without the "loopback address not considered" part)

IOW, it makes getaddrinfo try to guess if external IPv4 or IPv6 connectivity is available by looking through all network interfaces and checking if any of them have a non-loopback IP address configured. If it finds neither, then no `AF_INET` or `AF_INET6` lookup with that flag will succeed. (Note that at least with glibc, an `AF_UNSPEC` lookup will succeed and return both IPv4 and IPv6 results, even though per manpage it shouldn't.)

I have no idea why the `AI_ADDRCONFIG` flag is used here. It's also used in other uses of getaddrinfo in collectd codebase, including when looking up bind addresses, but all the other places that I've seen use `AF_UNSPEC` instead of explicitly trying IPv6 first then IPv4.

## Proposed fixes

Here are some of my ideas how to fix this. Please let me know which would be preferred, and I'll try to submit it as a pull request.

### Option 1

```diff
--- collectd-5.12.0.orig/src/write_prometheus.c
+++ collectd-5.12.0/src/write_prometheus.c
@@ -757,7 +757,7 @@ static int prom_open_socket(int addrfami
   struct addrinfo *res;
   int status = getaddrinfo(httpd_host, service,
                            &(struct addrinfo){
-                               .ai_flags = AI_PASSIVE | AI_ADDRCONFIG,
+                               .ai_flags = AI_PASSIVE,
                                .ai_family = addrfamily,
                                .ai_socktype = SOCK_STREAM,
                            },
```

This is the fix I'm running right now.
I guess I'll have to rebase it onto master before submitting a PR, but that should be trivial.

It's not very consistent with the rest of collectd codebase, and I don't know how it will behave on systems other than Linux+glibc,
but it's a small change, and according to the [two][Debian manpage] [manpages][FreeBSD manpage] I looked up, it's correct.

The only problem I can think of would be on systems where sockets bound to IPv6 wildcard can't receive IPv4 connections (eg. Linux with `sysctl net.ipv6.bindv6only=1`), but should be easily worked around by explicitly specifying `Host 127.0.0.1` in the config.
IMO an explicitly configured bind address is a better way to decide that trying to guess based on addresses assigned to interfaces.

### Option 2

If we want to keep `AI_ADDRCONFIG`, another fix would be to use `getaddrinfo` with AF_UNSPEC, similar to here:
https://github.com/collectd/collectd/blob/cf965e24cd37c5b2a9109b0e143bafc18985a2b5/src/gmond.c#L213-L218
and then filter the results checking if their addres family matches the one passed as an argument to `prom_open_socket` to preserve the try-IPv6-first behavior.

I haven't tested that, I feel like it has more room for things to go wrong, but it'd be more consistent with how `getaddrinfo` is used elsewhere in the project.

### Option 3

Same as Option 2 but get rid of the `addrfamily` argument and just let getaddrinfo decide which family we bind.
This would be the most consistent with how it's done in gmond.c, but also the biggest change.

Haven't tried this one either.

[FreeBSD manpage]: https://man.freebsd.org/cgi/man.cgi?query=getaddrinfo&sektion=3&apropos=0&manpath=FreeBSD+13.2-RELEASE+and+Ports
[Debian manpage]: https://manpages.debian.org/bookworm/manpages-dev/getaddrinfo.3.en.html

Wolf480pl:
Fixes #4150

write_prometheus uses getaddrinfo to resolve the bind address. The AI_ADDRCONFIG flag causes getaddrinfo to refuse to resolve 0.0.0.0 when the system has no non-loopback IPv4 addresses configured and refuse to resolve :: when the system has no non-loopback IPv6 configured.

We want binding to a wildcard address (0.0.0.0 or ::) to always work, even if the network is down.

To achieve that, don't pass the AI_ADDRCONFIG flag when resolving a bind address.

Changelog: write_prometheus: fixed failing to listen on wildcard address before network is up

### DIFF
--- a/src/write_prometheus.c
+++ b/src/write_prometheus.c
@@ -757,7 +757,7 @@ static int prom_open_socket(int addrfamily) {
   struct addrinfo *res;
   int status = getaddrinfo(httpd_host, service,
                            &(struct addrinfo){
-                               .ai_flags = AI_PASSIVE | AI_ADDRCONFIG,
+                               .ai_flags = AI_PASSIVE,
                                .ai_family = addrfamily,
                                .ai_socktype = SOCK_STREAM,
                            },


mrunge (approved):
Thank you

PR #4156, opened 2023-11-24
curl_stats: fix compatibility with new versions of cURL.

octo:
Use integer based keys for metrics if available.

cURL ≥ 7.55.0 provides additional keys that allow getting certain metrics as integers rather than doubles, e.g. content length. In some newer versions of cURL, the original keys (using doubles) are marked as deprecated.

ChangeLog: cURL, cURL-JSON, cURL-XML, Write HTTP plugins: fix compatibility with new versions of cURL.


octo:
The cURL build is fixed, though other build errors remain.

PR #4161, opened 2023-11-24
Intel RDT plugin: use `pqos_mon_start_pids2()` if available.

octo:
Newer versions of the pqos library deprecate the previous function, so we have to migrate even though it makes memory management quite ugly.

This should fix the build on Debian Sid ("unstable").

ChangeLog: Intel RDT plugin: support for PQOS ≥5.0.0 has been added.


mrunge (approved):
thank you!

PR #4111, opened 2023-04-24
Fix compile warning for amqp

mrunge:
warning "amqp.h is deprecated, use rabbitmq-c/amqp.h instead."

PR #4162, opened 2023-11-24
Build workflow: remove old distros.

octo:
This removes old and unsupported Linux distros:

*   Ubuntu 16 ("Xenial", support ended April 2021)
*   Ubuntu 18 ("Bionic Beaver", support ended June 2023)
*   Enterprise Linux 7 (technically still supported, but nine years old
    at this point)
*   Fedora 34 (support ended June 2022)
*   Fedora 35 (support ended December 2022)

Fedora 36 is also unsupported. We will keep the builder until we have never versions of Fedora available (or at least Rawhide builds again).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,14 +29,10 @@ jobs:
           # Ubuntu
           - focal_amd64
           - bionic_amd64
-          - xenial_amd64
           # RedHat family
           - el9_x86_64
           - el8_x86_64
-          - el7_x86_64
           - fedora36_x86_64
-          - fedora35_x86_64
-          - fedora34_x86_64
         config_flags: ['']
     env:
       MAKEFLAGS: "-j2 -sk"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,6 @@ jobs:
           - bionic_amd64
           # RedHat family
           - el9_x86_64
-          - el8_x86_64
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -29,14 +29,10 @@ jobs:
           # Ubuntu
           - focal_amd64
           - bionic_amd64
-          - xenial_amd64
           # RedHat family
           - el9_x86_64
           - el8_x86_64
```
mrunge, line 34: It seems we missed CentOS 8 EOL 
```CentOS Linux 8 will reach End Of Life (EOL) on December 31st, 2021. ```, see 
https://www.centos.org/centos-linux-eol/

mrunge:
Thank you for doing this! 
For the Fedora rawhide build issue, I have https://github.com/collectd/collectd/pull/4111

.github/workflows/build.yml
octo: I didn't realize the "EL" builders were actually using CentOS. With CentOS being discontinued, what would be the best way forward here? Are there free to use variants of RHEL that we could test instead?

octo:
Just merged #4111, thank you!

.github/workflows/build.yml
mrunge: I would start with CentOS stream (that's basically current RHEL). Or we could use almalinux here?
https://hub.docker.com/_/almalinux

Issue #4152, opened 2023-11-20
collectd tells the init system it's ready before it actually is


josephcsible:
*   Version of collectd: Git cfdfe4a
*   Operating system / distribution: Debian bookworm
*   Kernel version (if applicable): N/A

## Expected behavior

collectd should notify upstart or systemd that it's ready after `do_init` is finished, here: https://github.com/collectd/collectd/blob/cfdfe4aa3f6dae65d9b7c9265993c74610d85c62/src/daemon/collectd.c#L426

## Actual behavior

collectd notifies upstart or systemd that it's ready at the beginning of `main`, well before `do_init`, here: https://github.com/collectd/collectd/blob/cfdfe4aa3f6dae65d9b7c9265993c74610d85c62/src/daemon/cmd.c#L183

## Steps to reproduce

* Start collectd with upstart or systemd
* Immediately after, spike the CPU for a few milliseconds
* Observe that collectd didn't notice the CPU usage

octo:
Thanks for reporting this @josephcsible! I have a patch for this in #4164. Do you have the means to test it?

PR #4164, opened 2023-11-24
Notify upstart/systemd after initialization is complete.

octo:
ChangeLog: Daemon: The interaction with Upstart and Systemd has been improved.

Fixes: #4152

### DIFF
--- a/src/daemon/cmd.c
+++ b/src/daemon/cmd.c
@@ -83,42 +83,45 @@ static int pidfile_remove(void) {
 #endif /* COLLECT_DAEMON */
 
 #ifdef KERNEL_LINUX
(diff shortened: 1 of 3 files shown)


eero-t:
Changes look OK.

Btw. I wondered whether it still makes sense to continue supporting Upstart, as last LTS release to support that was Ubuntu 14.04 LTS, which standard support ended 4 years ago, and even its extended (commercial) 10 year support will end in few months (on 2024-04-25).

However, then I found out that Chrome/iumOS still uses Upstart (although there have been discussions about switching to systemd since 2014): https://www.chromium.org/chromium-os/chromiumos-design-docs/boot-design/

PR #4173, opened 2023-11-27
[Collectd 6] gpu_sysman: improve metric variants handling

eero-t:
ChangeLog: gpu_sysman: improve metric variants handling

In https://github.com/collectd/collectd/pull/4047 PR, Sysman plugin added option for controlling which variant(s) for the same base metric value are reported (e.g. joules counter, its rate as watts, or ratio of that compared to max power).  Reason being that while they each are useful in different situations, having all of them enabled (=default) can on some HW produce overwhelming amount of data (unless one disables some metric types).

This PR extends that support to cover all metrics:
  * Memory usage, frequency, and temperature metrics had only partial support earlier
  * I changed "counter" variant to a "base" one that includes both "counter" and "gauge" style base metrics (I did not want to have multiple variant names for Sysman base types)

Other changes:
* Rename GPU struct "counter" member to better reflect its purpose (to differentiate it from metric "counter" type)
* Add unit (s) name to (optional) plugin output at startup
* Fix few unrelated typos in plugin documentation

EDIT: split "LogMetrics" part to separate PR.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3686,8 +3686,8 @@ to read GPU information.
 
 Options below give overview of what information it could provide, but
 the information actually available through it depends on what is
-supported by the underlying GPU HW, its kernel driver, and Sysman
-backend user-space (compute) driver implementation.
+supported by the underlying GPU HW, its kernel driver, and Level-Zero
+Sysman backend (compute) driver implementation.
 
 =over 4
 
@@ -3706,7 +3706,7 @@ per second. When Samples is larger than 1, min + max are calculated
 for the sampled values and submitted instead of the read values
 themselves.
 
-Most of other metrics values are either counters, or change much
+Most of the other metric values are either counters, or change much
 slower, and are read only at submit intervals.  If collecting of the
 sampled values is disabled, it is better to set Samples to 1
 (default).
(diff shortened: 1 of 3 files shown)


octo:
I don't think that implementing metric logging in a read plugin as a one-off is a scalable strategy. We have the `write_log` for logging all metrics, mostly for development and debugging. Does that work for your use-case?

The other changes look good to me.

octo:
Those a great points, thanks for elaborating!

The first few are easy:

* The *write_log plugin* already supports two formats and we can rather easily add another "human readable" format.
* The extra "write_log values:" line doesn't add utility and we should just drop it.
* The [LogFile plugin](https://github.com/collectd/collectd/wiki/Plugin-LogFile) has a `Timestamp` option that will print the time at which the line is printed. Alternatively, the potential new "human readable" format could print the metric time in a similar fashion.

The label selection seems quite specific to the *gpu_sysman plugin*. We could add an *IgnoreLabel* option to *write_log*, but I get that configuring this every time you want to test during debugging can be annoying. We might be able to implement this in a generic way while delegating the decision which labels to skip to the plugins by adding some metadata key to the metric. Not sure whether that's worth it though.

Printing some labels as a prefix and others as part of the metric is also something that doesn't lend itself to a generic approach.

How about this way forward: let's move the non-metric-printing changes to a separate PR and submit them and continue the debugging/printing conversation here. Does that work for you?

octo (approved):
Thanks @eero-t!

eero-t:
Merge? (I cannot)

All CI errors are from cpu / intel_rdt / nut / virt plugins, or capabilties / curl helpers, i.e. unrelated to sysman plugin...

octo:
I think the merge is blocked because the branch cannot be merged cleanly. Happy to take care of this later tonight.

PR #4174, opened 2023-11-28
[collectd 6.0] Fix the build

octo:
*   Disable all unsupported plugins in the `configure` script. They can still be force-enabled with `--enable-foo`.
*   Remove the lists of unsupported plugins from the CI definitions (Cirrus CI and GitHub Actions).
*   Cherry-picks the following fixes from `main`:
    *   #4155
    *   #4156
    *   #4161
    *   #4165
    *   #4167
    *   #4169
    *   #4172
*   Fix minor issues in the Capabilities and CPU plugins.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2649,6 +2649,25 @@ if test "x$with_libganglia" = "xyes"; then
   fi
 fi
 
+AC_CHECK_HEADERS([rpc/types.h],
+  [have_rpc_types_h="yes"],
+  [have_rpc_types_h="no"]
+)
+AS_UNSET([ac_cv_header_rpc_types_h])
+if test "x$have_rpc_types_h" = "xno"; then
+  # SunRPC has been removed from glibc.
+  # Replacement headers are in the "tirpc" subdirectory.
+  SAVE_CPPFLAGS="$CPPFLAGS"
+  CPPFLAGS="$CPPFLAGS -I/usr/include/tirpc"
+
+  AC_CHECK_HEADERS([rpc/types.h],
+    [GANGLIA_CPPFLAGS="$GANGLIA_CPPFLAGS -I/usr/include/tirpc"],
+    [with_libganglia="no (rpc/types.h not found)"]
+  )
+
+  CPPFLAGS="$SAVE_CPPFLAGS"
+fi
+
 SAVE_CPPFLAGS="$CPPFLAGS"
 SAVE_LDFLAGS="$LDFLAGS"
 CPPFLAGS="$CPPFLAGS $GANGLIA_CPPFLAGS"
--- a/configure.ac
+++ b/configure.ac
@@ -5873,9 +5873,10 @@ fi
 if test "x$with_libupsclient" = "xyes"; then
   SAVE_CPPFLAGS="$CPPFLAGS"
   CPPFLAGS="$CPPFLAGS $with_libupsclient_cflags"
+  someUPSCONN_FOUND=false
 
   AC_CHECK_TYPES([UPSCONN_t, UPSCONN],
-    [],
+    [someUPSCONN_FOUND=true],
     [],
     [[
       #include <stdlib.h>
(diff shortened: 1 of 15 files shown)


configure.ac
```diff
@@ -6567,88 +6696,186 @@ AC_COLLECTD([werror],    [disable], [feature], [building with -Werror])
...
+plugin_write_riemann="$with_libriemann_client"
+plugin_write_sensu="yes"
 plugin_write_stackdriver="no"
+plugin_write_syslog="yes"
+plugin_write_tsdb="yes"
 plugin_xencpu="no"
 plugin_zfs_arc="no"
 plugin_zone="no"
 plugin_zookeeper="no"
+plugin_zookeeper="yes"
```
eero-t, line 6878: "no", but "yes"???

And why some plugins are first declared with  `="$with_something"`, and right after that either as `="no"` or `="yes"`, but some plugins in above list have just `="$with_something"`?

src/capabilities.c
```diff
@@ -206,7 +206,7 @@ static MHD_RESULT cap_http_handler(void *cls, struct MHD_Connection *connection,
    * round. The docs are not very specific on the issue. */
   if (*connection_state == NULL) {
     /* set to a random non-NULL pointer. */
-    *connection_state = &(int){44};
+    *connection_state = (void *)1;
```
eero-t, line 209: To me it would be clearer if this used same as `write_prometheus.c` plugin code:
```
    /* keep track of connection state */
    *connection_state = (void *)"called";
```

And the comment above stating "The docs are not very specific on the issue" is not really on-spot either.  MHD docs explicitly state that first call for each connection is after headers have been parsed, and should be used only for reporting errors (like is stated in  `write_prometheus.c`).

eero-t (commented):
* `nut` plugin change here is different from the one indicated?
* I do not see corresponding `smart` plugin change in `main` branch?
* while `curl` & `gmond` changes here look identical to main branch PRs, those were merged without reviews, and I do not know enough of their APIs to do that here
* I wonder why fedora39 and debian unstable images in CI files do not have arch suffix like the rest of the image...
* For rest of the commits I added couple of comments, but in general they seem OK

configure.ac
octo: \*lol\* the additions have been generated from the arguments that have previously been passed  the `AC_PLUGIN` macro. Apparently `plugin_zookeeper="no"` was defined and never used, hence the redefinition. I'll fix that.

eero-t:
`mdevents_test.c` fails on redhat:
```
In file included from src/mdevents_test.c:30:
src/testing.h:43:35: warning: 'test_compile_regex' defined but not used [-Wunused-function]
   43 | #define DEF_TEST(func) static int test_##func(void)
      |                                   ^~~~~
src/mdevents_test.c:49:1: note: in expansion of macro 'DEF_TEST'
   49 | DEF_TEST(compile_regex) {
      | ^~~~~~~~
```

And the new `freebsd` target does not like kairodb:
```
src/utils/format_kairosdb/format_kairosdb.c:66:28: error: unknown type name 'yajl_gen'
static int json_add_string(yajl_gen g, char const *str) /* {{{ */

<lots of other errors>
```

Rest of errors are complaining about `expected plugins missing`.

octo:
Thanks for the thorough review!

* The NUT plugin fix is a cherry-pick of 4a469251a67487814d44a171087f6ac9d5e8291d
* The SMART plugin fix is part of #4165, sorry for not separating this better.
* I want to move to a different naming schema for these contains and we're in an inconsistent state due to the incomplete transition. I find that the architecture suffix is irrelevant if *all* images use the same architecture. Noting when the architecture is _not_ x86-64 improves the readability of the result list. I also want to move away from codenames (e.g. Debian Bookworm) and towards release versions (e.g. Debian 12). Hence the move from `sid_amd64` to `debian_unstable`.
* Thanks for pointing out the `expected plugins missing` issue. The list of expected pugins is currently defined in the container image, i.e. we cannot differentiate between `main` and `collectd-6.0`. I'll think of something …

octo:
I still need to investigate the FreeBSD / KairosDB issue. It looks like a missing dependency / configure issue.

octo:
Okay, all the GitHub Actions based builders are happy. I think this is a good point to merge this PR. I'll look at the rest in separate PRs.

Issue #4050, opened 2022-09-19
write_riemann: riemann-client v1.8+ timeout support causes incomplete events & segfaults


node13h:
*   Version of collectd: 5.12.0-16.fc36
*   Operating system / distribution: Fedora 36
*   Kernel version (if applicable): 5.19.9-200.fc36.x86_64
*   SELinux in permissive mode

I'm using collectd with a `write_riemann` plugin for output, both installed from Fedora repositories. 
To reproduce this issue I'm running Riemann 0.3.8 on localhost, with a default config modified to print collectd events.

The following issues have been reproduced on multiple hosts (all running Fedora 36). I've also tried running collectd 5.12.0 compiled from source - it behaves the same as the rpm version.

**1. With a small number of read plugins enabled some of the events sent to Riemann are missing field values.**
Example:

collectd config:
```
Interval 10
FQDNLookup false
Hostname "host.example.com"

LoadPlugin syslog
LoadPlugin write_riemann
LoadPlugin load
LoadPlugin memory
LoadPlugin swap

<Plugin "write_riemann">
    <Node "local">
        Host "127.0.0.1"
        Protocol TCP
        BatchFlushTimeout 1
        TTLFactor 3.0
    </Node>
    Tag "collectd"
</Plugin>

<Target "write">
Plugin "write_riemann/local"
</Target>
```

Events captured on the Riemann side (note the missing service value in the first event, and the missing host value in the second event:

```
{:host "host.example.com", :service "", :state nil, :description nil, :metric 3.99548416E8, :tags ["collectd"], :time 1.663619928593841E9, :ttl 30.0, :plugin "memory", :type "memory", :ds_name "value", :type_instance "slab_unrecl", :ds_type "gauge", :ds_index "0"}
{:host "", :service "memory/memory-slab_recl", :state nil, :description nil, :metric 2.93199872E8, :tags ["collectd"], :time 1.663619928593841E9, :ttl 30.0, :plugin "memory", :type "memory", :ds_name "value", :type_instance "slab_recl", :ds_type "gauge", :ds_index "0"}
{:host "host.example.com", :service "swap/swap-used", :state nil, :description nil, :metric 0.0, :tags ["collectd"], :time 1.663619928593964E9, :ttl 30.0, :plugin "swap", :type "swap", :ds_name "value", :type_instance "used", :ds_type "gauge", :ds_index "0"}
[...]

PR #4176, opened 2023-11-29
[collectd 6] add reference counting to plugins that register multiple callbacks with the same `user_data_t`.

octo:
This outlines an alternative fix to the problem described in #4117.

The idea is to fix the plugins rather than make the plugin's memory management a problem of the shutdown code in `src/daemon/plugin.c`. By adding a reference count, it doesn't matter in which order the various callbacks are shut down in the daemon code.

I hope that the same approach can be used to fix other affected plugins (write_riemann, write_stackdriver, write_syslog, write_tsdb), but I have not verified that yet.

### DIFF
--- a/src/write_riemann.c
+++ b/src/write_riemann.c
@@ -617,15 +617,14 @@ static void wrr_free(void *p) /* {{{ */
 
   pthread_mutex_lock(&host->lock);
   host->reference_count--;
-  int reference_count = host->reference_count;
-  pthread_mutex_unlock(&host->lock);
-
-  if (reference_count > 0) {
+  if (host->reference_count > 0) {
+    pthread_mutex_unlock(&host->lock);
     return;
   }
 
   wrr_disconnect(host);
 
+  pthread_mutex_unlock(&host->lock);
   pthread_mutex_destroy(&host->lock);
   sfree(host);
 } /* }}} void wrr_free */
--- a/src/write_riemann.c
+++ b/src/write_riemann.c
@@ -616,16 +616,16 @@ static void wrr_free(void *p) /* {{{ */
     return;
 
   pthread_mutex_lock(&host->lock);
-
   host->reference_count--;
-  if (host->reference_count > 0) {
-    pthread_mutex_unlock(&host->lock);
+  int reference_count = host->reference_count;
+  pthread_mutex_unlock(&host->lock);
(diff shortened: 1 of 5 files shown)


src/write_riemann.c
```diff
@@ -616,16 +616,16 @@ static void wrr_free(void *p) /* {{{ */
...
 
   pthread_mutex_lock(&host->lock);
-
   host->reference_count--;
-  if (host->reference_count > 0) {
-    pthread_mutex_unlock(&host->lock);
+  int reference_count = host->reference_count;
+  pthread_mutex_unlock(&host->lock);
+
+  if (reference_count > 0) {
```
eero-t, line 623: What's the reason for this change, doesn't it introduce small chance for race condition?

PS.  See also https://github.com/collectd/collectd/issues/4050

eero-t (commented):
On quick look this seems OK, I just have a question about one Rieman plugin change.

I would also suggest running collectd with these plugins under Valgrind memcheck (default valgrind tool) and ones discussed here: https://github.com/collectd/collectd/pull/4026#issuecomment-1214894922.

src/write_riemann.c
octo: I wanted to keep this code consistent across plugins. This looks like an oversight. Fixed.

src/write_riemann.c
```diff
@@ -625,7 +624,7 @@ static void wrr_free(void *p) /* {{{ */
 
   wrr_disconnect(host);
 
-  pthread_mutex_lock(&host->lock);
+  pthread_mutex_unlock(&cb->curl_lock);
   pthread_mutex_destroy(&host->lock);
```
eero-t, line 628: Changing lock to unlock looks like bug fix, but why the `host->lock` to  `cb->curl_lock` change, when only `host->lock` is destroyed (and other plugins destroy `cb` lock)?

eero-t (approved):
Looks OK to me now, but I would really recommend doing also init/deinit testing under  Valgrind/ASAN to verify at least riemann plugin validity.

octo:
I could not test the *write_riemann plugin* because it hasn't been ported to collectd 6 yet. I did run the *write_http plugin* with Valgrind (both memcheck and helgrind) without any findings.

PR #4181, opened 2023-12-05
Update standard system metric names in accordance with OpenTelemetry.

octo:
[OpenTelemetry Enhancement Proposal \#119](https://github.com/open-telemetry/oteps/blob/main/text/0119-standard-system-metrics.md) outlines names for a few commonly tracked system metrics, such as CPU, disk, and memory usage. This change intends to change the emitted metrics of the appropriate plugins to match this naming schema.

This includes merging metric families in the disk and df plugins, which have previously used separate metric families for the read and write "direction".

ChangeLog: CPU, DF, Disk, and Memory plugins: Metric names have been updates to conform to OTEP 119.

### DIFF
--- a/src/cpu.c
+++ b/src/cpu.c
@@ -442,7 +442,7 @@ static void aggregate(gauge_t *sum_by_state) /* {{{ */
 static void cpu_commit_one(int cpu_num, /* {{{ */
                            gauge_t rates[static COLLECTD_CPU_STATE_MAX]) {
   metric_family_t fam = {
-      .name = "cpu_usage_percent",
+      .name = "system.cpu.utilization",
       .type = METRIC_TYPE_GAUGE,
   };
 
@@ -450,8 +450,8 @@ static void cpu_commit_one(int cpu_num, /* {{{ */
   if (cpu_num == -1) {
     metric_label_set(&m, "cpu", "total");
   } else {
-    char cpu_num_str[16];
-    snprintf(cpu_num_str, sizeof(cpu_num_str), "%d", cpu_num);
+    char cpu_num_str[32];
+    ssnprintf(cpu_num_str, sizeof(cpu_num_str), "%d", cpu_num);
     metric_label_set(&m, "cpu", cpu_num_str);
   }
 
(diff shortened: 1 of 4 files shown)


mrunge (approved):
thank you!

eero-t:
Was that proposal already accepted?

I think some of the earlier names matched better e.g. with Prometheus metric naming conventions: https://prometheus.io/docs/practices/naming/

And OpenMetrics standard recommendations: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily

Which include unit (bytes, seconds etc) in the metric family name.

eero-t:
> How could I tell if the proposal was not accepted yet?

Apparently naming is "experimental": https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-naming


> To make the "unit" field useful, it needs to be machine readable, i.e. we stick to naming conventions for units (https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units). IMHO an authoritative enum would have been preferable over a string field + naming convention, but that's what we have.
>
> collectd should add a separate unit field. When emitting OpenMetrics, we can conform to their naming schema by adding appropriate suffixes if necessary.

Uh, that link says to use eg. "Cel" to represent Celsius.   That's really user **un**friendly unit name...

octo:
> Apparently naming is "experimental": https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-naming

For system metrics, https://opentelemetry.io/docs/specs/semconv/system/system-metrics/ is more relevant – which is also "experimental".

> Uh, that link says to use eg. "Cel" to represent Celsius. That's really user unfriendly unit name...

Yeah, not disagreeing. OpenTelemetry defers to a "Unified Code for Units of Measure", which IMHO is a good call in principle (better than making up your own). But that means that most units relevant for collectd have "uncommon" names. "Cel" for "°Celsius" is not great, but "MBy" for "megabyte" is just outright confusing.

eero-t:
Ok, so UCUM c/s names will need a mapping table(s) for the write plugin, and I guess collectd structs will just store unit index to those mapping tables.

eero-t:
This breaks Prometheus integration. Dot is invalid in metric and label names. Double colon is also invalid for labels, but could be used as separator for metric names. See: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

PR #4182, opened 2023-12-05
configure: Disable the write_http plugin if libyajl2 is not found.

octo:
This fixes the build on FreeBSD.


mrunge (approved):
looks good to me

PR #4183, opened 2023-12-05
network.c: allow notification forwarding

carlospeon:
ChangeLog: network.c allow notification forwarding.

Allow notification forwarding checking notification metadata to avoid loops.


carlospeon:
Tagged as WIP: checks do not build

octo:
High level feedback: the problem we need to solve here is when two instances of collectd are configured to send notifications to each other. Because we can't currently tell whether we've seen a notification before, collectd would potentially bounce a notification back and forth indefinitely.

To get around that, we need to "remember" notifications somehow. We could, for example, keep a copy of the notifications of the last 60 seconds and discard notifications that are older than that. Or we keep the last 100 notifications, which would make it possible to implement a "give me the last *n* notifications" API. Or a combination of the two.

src/network.c
```diff
@@ -3059,6 +3026,32 @@ static int network_config(oconfig_item_t *ci) /* {{{ */
   return 0;
 } /* }}} int network_config */
 
+static bool check_send_notification(const notification_t *n) {
```
eero-t, line 3029: Please keep the changed code in same place. Moving the code around makes diffs larger & reviewing the actual changes harder.

carlospeon: Ok, done.
Thank you.

eero-t: Thanks!

eero-t:
All checks fail due to network test changes:
```
 In file included from src/network_test.c:25:
src/daemon/plugin.c: In function 'set_thread_name':
src/daemon/plugin.c:622:16: error: implicit declaration of function 'pthread_setname_np'; did you mean 'pthread_setcanceltype'? [-Werror=implicit-function-declaration]
  622 |   int status = pthread_setname_np(tid, n);
      |                ^~~~~~~~~~~~~~~~~~
      |                pthread_setcanceltype
```

carlospeon:
Finally I manage to run checks. Removing WIP tag.

octo (changes requested):
Hi @carlospeon, thanks again for your work!

> In the case of two instances of collectd configured to send notifications (and values) to each other, "forwarding" should be "false" and only local notifications (and values) would be send to the other instance.

I know several people who have configured their collectd servers this way, and I'm sure that's just the tip of the iceberg. These setups would be massively disrupted with the PR in its current form.

In order to forward notifications, we need a way to deduplicate them. E.g. add notifications to a list of notifications of the last 300 seconds (with some upper bound) and dismiss received notifications if they are in that list.

In the interest of avoiding miscommunication, let me be super direct: Duplicate detection is not optional. Without it, I won't merge this PR.

carlospeon:
Hello @octo, thank you for your review!

> I know several people who have configured their collectd servers this way, and I'm sure that's just the tip of the iceberg. These setups would be massively disrupted with the PR in its current form.

In my experience if you have a configuration with a loop, values are resent continuously and collectd crashs, so there should not be such setups working properly. I checked the code and although there is a de-duplication detection for values:

https://github.com/collectd/collectd/blob/4cc3426c90b8dd9eb294cf446a50289332e3605e/src/daemon/utils_cache.c#L350

the return value of uc_update is not checked:

https://github.com/collectd/collectd/blob/4cc3426c90b8dd9eb294cf446a50289332e3605e/src/daemon/plugin.c#L2162

and duplicated values are dispatched to the write plugins. Or maybe I'm missing something.

We are running this code for a year without issues, our rate of notifications is something above a thousand per day and with a proper configuration we don't have any specific issue regarding loops. 

> In the interest of avoiding miscommunication, let me be super direct: Duplicate detection is not optional. Without it, I won't merge this PR.

I appreciate that, fair enough. This solution -maybe suboptimal- "works for me" and I just want to share it with the community. In this particular moment I'm not ready to code and test a more complex/complete solution.

src/daemon/plugin.h
```diff
@@ -43,13 +43,11 @@
...
-      : (t == DS_TYPE_GAUGE)                                                   \
-            ? "gauge"                                                          \
-            : (t == DS_TYPE_DERIVE)                                            \
-                  ? "derive"                                                   \
-                  : (t == DS_TYPE_ABSOLUTE) ? "absolute" : "unknown"
+  (t == DS_TYPE_COUNTER)    ? "counter"                                        \
+  : (t == DS_TYPE_GAUGE)    ? "gauge"                                          \
+  : (t == DS_TYPE_DERIVE)   ? "derive"                                         \
+  : (t == DS_TYPE_ABSOLUTE) ? "absolute"                                       \
+                            : "unknown"
```
eero-t, line 50: Unrelated code formatting change.  Drop it (it's something that changes in v6 any way).

carlospeon: Sorry, I ran format.sh over all files. Dropped.

eero-t: Cr*p, I had not realized clang-format rules had changed/updated since this file had last been updated.

In that case it's better to add back.  Sorry about that! :-/

eero-t:
CI failures are unrelated to these changes:
* Debian Unstable: due to `epics` plugin dependency change
* FreeBSD: `The resource 'projects/freebsd-org-cloud-dev/global/images/family/freebsd-13-2' was not found`
  * Maybe because there's now v14, and latest v13.x release is v13.3?

src/daemon/plugin.h
carlospeon: OK, no problem! Undone.

eero-t:
It's really up to @octo to accept this, as he NAK'ed it.

carlospeon:
> It's really up to @octo to accept this, as he NAK'ed it.

OK sorry, I missunderstood the intention of your previous review.

PR #4185, opened 2023-12-06
.cirrus.yml: Read expected plugins from branch specific config file.

octo:
This is an attempt to configure different expected plugins based on the branch.


eero-t:
Branch specific lists of plugins could be added within this PR too, so we could see whether tests pass then.

eero-t:
Option 1. sounds fine to me.

Is there some base set of plugins that should always succeed regardless of branch or distro?    I.e. would it make sense to split the list to 2 parts?

eero-t:
@octo are you waiting on comment from @mrunge?

octo:
Closing in favor of #4208

octo:
@eero-t @mrunge Thanks for your input! Option 1 it is.

PR #4187, opened 2023-12-13
[collectd 6] RFC: Add resource attributes.

octo:
The internal data structure in collectd 6 is closely modeled after [OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md). This is, for example, where the term "metric family" originates.

In the mean time, [OpenTelemetry](https://opentelemetry.io/) has emerged as the de-facto industry standard for metrics, in particular in "cloud native" environments. I think it is prudent to ensure collectd 6 is compatible with OpenTelemetry and integrates well into this ecosystem.

Most concepts translate well and don't require changes to our internal data structures. However, there is one important differentiation:

OpenTelemetry specifies two sets of "attributes" (aka. labels): "metric attributes" and "resource attributes". For former is equivalent to the existing `label` field in `metric_t`. [Resource attributes](https://opentelemetry.io/docs/specs/semconv/resource/) describe which resource is being monitored and currently lacks an equivalent in collectd 6. In collectd 5, the `host` field played a comparable role. I believe other systems use this information for sharding, i.e. ensure data locality for metrics originating from the same "resource".

I propose the following changes:

* Associate a new "resource attributes" `label_set_t` with metrics, either directly or indirectly. I see a few ways we could implement this:
    1. Create a `resource_metrics_t` struct, wrapping one or more `metric_family_t`.
        * :+1: Pro: closely follows the OpenTelemetry data structures, avoiding future incompatibilities.
        * :-1: Con: requires lots of code changes because `metric_family_t` is used as *the* metric exchange struct throughout the existing code.
    2. Create a `resource_metrics_t` struct and add a pointer to it to `metric_family_t`.
        * :+1: Pro: conceptually this is similar to how a `metric_t` refers back to its `metric_family_t`. In other words, we'd be kind of "wrapping" the metric family without changing which struct we're passing around (→ fewer code changes).
        * :-1: Con: metrics from different "resources" have to go into separate metric families.
    3. Add a `label_set_t resource` field to `metric_t`.
        * :+1: Pro: most information identifying a metric is concentrated in one place (the metric name is in the metric family though). Schema is relatively simple to understand.
        * :-1: Con: resource attributes have to be stored with each metric, consuming a lot of memory, particularly when one resource is exporting a large number of metrics.

    I have initially implemented the last option (iii.). A later commit changes this to store the resource attributes in `metric_family_t` (similar to option ii.).
* Write infrastructure that auto-populates the "resource attributes" with reasonable defaults. I think following the [Host](https://opentelemetry.io/docs/specs/semconv/resource/host/) semantics for compute instances would be a good MVP. More attributes for specific cloud environments could be added at a later time and the infrastructure should be able to accommodate these extensions.
* Allow users to overwrite the resource attributes, either partially or in full.

ChangeLog: Resource attributes have been added to the core data structures.

### DIFF
--- a/src/daemon/metric.c
+++ b/src/daemon/metric.c
@@ -214,6 +214,10 @@ void label_set_reset(label_set_t *labels) {
 }
 
 int label_set_clone(label_set_t *dest, label_set_t src) {
+  if (dest == NULL || dest->num != 0) {
+    return EINVAL;
+  }
+
   if (src.num == 0) {
     return 0;
   }
--- a/src/daemon/metric.c
+++ b/src/daemon/metric.c
@@ -38,9 +38,13 @@
 #define VALID_LABEL_CHARS                                                      \
   "abcdefghijklmnopqrstuvwxyz"                                                 \
   "ABCDEFGHIJKLMNOPQRSTUVWXYZ"                                                 \
-  "0123456789_"
+  "0123456789_.-"
+
 /* Metric names must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*` */
-#define VALID_NAME_CHARS VALID_LABEL_CHARS ":"
+// instrument-name = ALPHA 0*254 ("_" / "." / "-" / "/" / ALPHA / DIGIT)
+#define VALID_NAME_CHARS VALID_LABEL_CHARS "/"
+
+#define RESOURCE_LABEL_PREFIX "resource:"
 
 int value_marshal_text(strbuf_t *buf, value_t v, metric_type_t type) {
   switch (type) {
@@ -60,7 +64,8 @@ static int label_pair_compare(void const *a, void const *b) {
                 ((label_pair_t const *)b)->name);
 }
 
-static label_pair_t *label_set_read(label_set_t labels, char const *name) {
+static label_pair_t *label_set_read(label_set_t const *labels,
+                                    char const *name) {
   if (name == NULL) {
     errno = EINVAL;
     return NULL;
@@ -73,8 +78,8 @@ static label_pair_t *label_set_read(label_set_t labels, char const *name) {
       .name = name,
   };
 
-  label_pair_t *ret = bsearch(&label, labels.ptr, labels.num,
-                              sizeof(*labels.ptr), label_pair_compare);
+  label_pair_t *ret = bsearch(&label, labels->ptr, labels->num,
+                              sizeof(*labels->ptr), label_pair_compare);
   if (ret == NULL) {
     errno = ENOENT;
     return NULL;
@@ -83,8 +88,7 @@ static label_pair_t *label_set_read(label_set_t labels, char const *name) {
   return ret;
 }
 
-static int label_set_create(label_set_t *labels, char const *name,
-                            char const *value) {
+int label_set_add(label_set_t *labels, char const *name, char const *value) {
   if ((labels == NULL) || (name == NULL) || (value == NULL)) {
     return EINVAL;
   }
@@ -99,7 +103,7 @@ static int label_set_create(label_set_t *labels, char const *name,
     return EINVAL;
   }
 
-  if (label_set_read(*labels, name) != NULL) {
+  if (label_set_read(labels, name) != NULL) {
     return EEXIST;
   }
   errno = 0;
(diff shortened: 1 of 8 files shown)


octo:
@eero-t @mrunge I'd love to hear your thoughts on this (see description above).

src/daemon/metric.c
```diff
@@ -207,70 +245,70 @@ int metric_reset(metric_t *m) {
...
-    status = status || strbuf_print(buf, "=\"");
-    status = status || strbuf_print_escaped(buf, m->label.ptr[i].value,
-                                            "\\\"\n\r\t", '\\');
-    status = status || strbuf_print(buf, "\"");
-  }
+  bool first_label = true;
+  status = status || format_label_set(buf, &m->resource, RESOURCE_LABEL_PREFIX,
+                                      first_label);
+
+  first_label = (m->resource.num == 0);
```
eero-t, line 291: IMHO this is odd way of writing:
```suggestion
  if (m->resource.num == 0) {
    status = status || format_label_set(buf, &m->resource, RESOURCE_LABEL_PREFIX,
                                        first_label);
    first_label = false;
  }
```

src/daemon/metric.h
```diff
@@ -70,6 +70,19 @@ typedef struct {
   size_t num;
 } label_set_t;
 
+/* label_set_clone copies all the laels in src into dest. If dest contains
```
eero-t, line 73:
```suggestion
/* label_set_clone copies all the labels in src into dest. If dest contains
```

src/daemon/metric.h
```diff
@@ -70,6 +70,19 @@ typedef struct {
...
+ * any labels prior to calling label_set_clone, the associated memory is
+ * leaked. */
+int label_set_clone(label_set_t *dest, label_set_t src);
```
eero-t, line 76: That seems rather nasty way of introducing accidental leaks.

What if it would return e.g. `EEXISTS` if there are already labels in it, like `label_set_add()` does?

(Maybe also log it, or assert on it?)

src/daemon/metric.h
```diff
@@ -108,6 +122,12 @@ metric_t *metric_parse_identity(char const *s);
  * label that does not exist is *not* an error. */
 int metric_label_set(metric_t *m, char const *name, char const *value);
 
+/* metric_resource_attribute_update adds, updates, or deleted a resource
+ * attribute. If "value" is NULL or the empty string, the attribute is removed.
```
eero-t, line 126:
```suggestion
/* metric_resource_attribute_update adds, updates, or deletes a resource
 * attribute. If "value" is NULL or an empty string, the attribute is removed.
```

eero-t (commented):
Except for the (documented) leakage thing (and few typos), this seems fine to me.

src/daemon/metric.h
octo: Changed the semantic to return `EINVAL` if `dest` is not empty.

eero-t:
It leaks in tests:
```
==35362== 63 (32 direct, 31 indirect) bytes in 1 blocks are definitely lost in loss record 3 of 4
==35362==    at 0x484ABC0: realloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==35362==    by 0x1099F3: UnknownInlinedFun (metric.c:116)
==35362==    by 0x1099F3: UnknownInlinedFun (metric.c:91)
==35362==    by 0x1099F3: label_set_update.part.0 (metric.c:184)
==35362==    by 0x10AB29: UnknownInlinedFun (stdio2.h:86)
==35362==    by 0x10AB29: UnknownInlinedFun (metric.c:307)
==35362==    by 0x10AB29: test_metric_identity (metric_test.c:174)
==35362==    by 0x1093C9: main (metric_test.c:298)
...
==35362== 512 bytes in 1 blocks are definitely lost in loss record 4 of 4
==35362==    at 0x484ABC0: realloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==35362==    by 0x109B32: strbuf_resize (strbuf.c:86)
==35362==    by 0x109C05: strbuf_print.part.0 (strbuf.c:153)
==35362==    by 0x10AC7C: UnknownInlinedFun (strbuf.c:149)
==35362==    by 0x10AC7C: UnknownInlinedFun (metric.c:284)
==35362==    by 0x10AC7C: test_metric_identity (metric_test.c:183)
==35362==    by 0x1093C9: main (metric_test.c:298)
```

src/daemon/metric.c
```diff
@@ -280,18 +279,19 @@ int metric_identity(strbuf_t *buf, metric_t const *m) {
...
     return status;
   }
 
   status = status || strbuf_print(buf, "{");
 
   bool first_label = true;
-  if (m->resource.num == 0) {
-    status = status || format_label_set(buf, &m->resource,
-                                        RESOURCE_LABEL_PREFIX, first_label);
+  if (resource->num != 0) {
```
eero-t, line 292: Sorry about that!

(I think I had ">" there originally, but force-push caused my pending suggestion to disappear and I messed when quickly re-writing it.  Maybe you could consider leaving force pushes until reviews are about finished?)

eero-t (approved):
Tests pass now (except for unrelated "expected plugins are MISSING" fails), and having resource labels for metric family instead of individual metrics is clear improvement => approved.

src/daemon/metric.c
octo: Oh, sorry about that! Yeah, I can rebase and clean up the commits later in the process, or just accept that my stupid mistakes will be forever enshrined in the commit history ;)

octo:
Thanks @eero-t!

PR #4180, opened 2023-12-04
[collectd 6] Write OpenTelemetry: New plugin to export metrics via OTLP over gRPC.

octo:
This adds a new plugin that exports metrics to an [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/) using the *OpenTelemetry Protocol* (OTLP) over gRPC.

This is intended as an MVP and comes with many limitations, e.g.:

* Communication is synchronous and blocks the write thread
* TLS is not yet supported
* ~The [Resource field](https://opentelemetry.io/docs/concepts/resources/) is unset (→ see #4187)~

ChangeLog: Write OpenTelemetry: A new plugin for exporting metrics using the OpenTelemetry Protocol over gRPC has been added.

PR #4188, opened 2023-12-14
[collectd 6] Add OTLP support to the Write HTTP plugin.

octo:
The protocol is described here: https://opentelemetry.io/docs/specs/otlp/#otlphttp


src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,278 @@
...
+    if (status != 0) {                                                         \
+      ERROR("format_json: %s failed with status %d", #f, status);              \
+      return status;                                                           \
+    }                                                                          \
+  } while (0)
+
+static int json_add_string(yajl_gen g, char const *str) /* {{{ */
+{
+  if (str == NULL) {
+    CHECK(yajl_gen_null(g));
```
eero-t, line 45: All callers of `json_add_string()` do it also with `CHECK()`, so error will be logged twice.

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,278 @@
...
+
+  CHECK(yajl_gen_string(g, (const unsigned char *)str,
+                        (unsigned int)strlen(str)));
+  return 0;
+} /* }}} int json_add_string */
+
+static int key_value(yajl_gen g, label_pair_t label) {
+  CHECK(yajl_gen_map_open(g)); /* BEGIN KeyValue */
+
+  CHECK(json_add_string(g, "key"));
```
eero-t, line 57: Both of the `key_value()` callers do it with `CHECK()`, so same error could be logged up to 3 times.

=> Please add tests for these error cases (and check the output).

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,278 @@
...
+  CHECK(yajl_gen_map_close(g)); /* END KeyValue */
+  return 0;
+}
+
+static int number_data_point(yajl_gen g, metric_t const *m) {
+  CHECK(yajl_gen_map_open(g)); /* BEGIN NumberDataPoint */
+
+  CHECK(json_add_string(g, "attributes"));
+  for (size_t i = 0; i < m->label.num; i++) {
+    CHECK(key_value(g, m->label.ptr[i]));
```
eero-t, line 75: Both of the `number_data_point()` callers do it with` CHECK()`, so same error could be logged up to 4 times.

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,278 @@
...
+  return 0;
+}
+
+static int gauge(yajl_gen g, metric_family_t const *fam) {
+  CHECK(yajl_gen_map_open(g)); /* BEGIN Gauge */
+
+  CHECK(json_add_string(g, "dataPoints"));
+  CHECK(yajl_gen_array_open(g));
+  for (size_t i = 0; i < fam->metric.num; i++) {
+    CHECK(number_data_point(g, fam->metric.ptr + i));
```
eero-t, line 106: `gauge()` caller does it with `CHECK()`, so same error could be logged up to 5 times.

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,278 @@
...
+  return 0;
+}
+
+static int sum(yajl_gen g, metric_family_t const *fam) {
+  CHECK(yajl_gen_map_open(g)); /* BEGIN Sum */
+
+  CHECK(json_add_string(g, "dataPoints"));
+  CHECK(yajl_gen_array_open(g));
+  for (size_t i = 0; i < fam->metric.num; i++) {
+    CHECK(number_data_point(g, fam->metric.ptr + i));
```
eero-t, line 120: `sum()` caller does it with `CHECK()`, so same error could be logged up to 5 times.

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  switch (fam->type) {
+  case METRIC_TYPE_COUNTER:
+    CHECK(json_add_string(g, "sum"));
+    CHECK(sum(g, fam));
+    break;
+  case METRIC_TYPE_GAUGE:
+    CHECK(json_add_string(g, "gauge"));
+    CHECK(gauge(g, fam));
```
eero-t, line 157: `metric()` caller does it with `CHECK()`, so same error could be logged up to 6 times.

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,278 @@
...
+
+static int scope_metrics(yajl_gen g, metric_family_t const *fam) {
+  CHECK(yajl_gen_map_open(g)); /* BEGIN ScopeMetrics */
+
+  CHECK(json_add_string(g, "scope"));
+  CHECK(instrumentation_scope(g));
+
+  CHECK(json_add_string(g, "metrics"));
+  CHECK(yajl_gen_array_open(g));
+  CHECK(metric(g, fam));
```
eero-t, line 189: `scope_metric()` caller does it with `CHECK()`, so same error could be logged up to 7 times.

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,278 @@
...
+
+  // TODO
+  CHECK(yajl_gen_map_open(g)); /* BEGIN ExportMetricsServiceRequest */
+  CHECK(json_add_string(g, "resourceMetrics"));
+  CHECK(yajl_gen_array_open(g));
+
+  unsigned char const *out = NULL;
+  for (size_t i = 0; i < families_num; i++) {
+    metric_family_t const *fam = families[i];
+    add_resource_metric(g, fam);
```
eero-t, line 251: As the error message is always same, error value is always propagated up, and this is the only place where those other (static) functions are called, it would be best to drop logging from `CHECK()`, and do logging for the error values only in this function.

eero-t (changes requested):
> ChangeLog: Write HTTP plugin: Support for the OpenTelemetry protocol (OTLP) has been added.

There's no caller for `format_json_open_telemetry()`, and the otlp functions in the write_http plugin are just stubs?

=> For now I'm just commenting on the error logging...

octo:
Hey @eero-t, thanks for your comments! I'll address them shortly.

This PR is still a draft and work in progress. Current state:

* The changes in `src/utils/format_json/` are ready to be reviewed.
* Changes to the *write_http plugin* are largely missing, see below.

The existing formats encode each metric family individually and append the encoded metric family to a buffer. We could do the same for the JSON output but we can't do that for the binary protobuf output.

What I'd prefer to do is what the *write_open_telemetry* plugin is also doing: stage metric families and convert them all to the desired format inside the *flush* callback/function. This allows us to (eventually) group metric families by their resource attributes, reducing the size of the generated payload.

src/utils/format_json/open_telemetry.c
octo: I just pushed two changes:

* Add specialized logging to `json_add_string()` so it includes the string that it failed to add to the buffer.
* Modify the macro to only log an error if the function name begins with `yajl_gen_`.

I think this should address the concerns you have raised.

src/utils/format_json/open_telemetry.c
octo: Fixed in 1bd6d8c7220fe9f00e9155f4a8dc4570b200ba08

src/utils/format_json/open_telemetry.c
octo: Fixed in 1bd6d8c7220fe9f00e9155f4a8dc4570b200ba08

src/utils/format_json/open_telemetry.c
octo: Fixed in 1bd6d8c7220fe9f00e9155f4a8dc4570b200ba08

src/utils/format_json/open_telemetry.c
octo: Fixed in 1bd6d8c7220fe9f00e9155f4a8dc4570b200ba08

src/utils/format_json/open_telemetry.c
octo: Fixed in 1bd6d8c7220fe9f00e9155f4a8dc4570b200ba08

src/utils/format_json/open_telemetry.c
octo: Fixed in 1bd6d8c7220fe9f00e9155f4a8dc4570b200ba08

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,285 @@
...
+
+#include <yajl/yajl_gen.h>
+
+#define CHECK(f)                                                               \
+  do {                                                                         \
+    int status = (f);                                                          \
+    if (status != 0) {                                                         \
+      if (strncmp(#f, "yajl_gen_", strlen("yajl_gen_")) == 0) {                \
+        ERROR("format_json: %s failed with status %d", #f, status);            \
+      }                                                                        \
```
eero-t, line 39: This whole thing is quite a lot of extra code for every CHECK line.  At least stop replicating this part by using something like:
```
static void log_yajl_gen_error(const char *fname) {
  if (strncmp(fname, "yajl_gen_", strlen("yajl_gen_")) == 0) {
    ERROR("format_json: %s failed with status %d", fname, status);
  }
}
```
and:
```suggestion
      log_yajl_gen_error(#f); \
```

PS. If the return value is relevant only for whether it's non-zero, another option would have been just to return NULL for success, and the failing function name for error.  Then print only in top level function that failing function name...

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,285 @@
...
+  CHECK(yajl_gen_integer(g, CDTIME_T_TO_NS(m->time)));
+
+  switch (m->family->type) {
+  case METRIC_TYPE_COUNTER:
+    CHECK(json_add_string(g, "asInt"));
+    CHECK(yajl_gen_integer(g, m->value.counter));
+    break;
+  case METRIC_TYPE_GAUGE:
+    CHECK(json_add_string(g, "asDouble"));
+    CHECK(yajl_gen_integer(g, m->value.gauge));
```
eero-t, line 96: generate _integer_?

src/utils/format_json/open_telemetry.c
octo: Great suggestion, thank you! Done.

src/utils/format_json/open_telemetry.c
octo: Eek! Great catch! Thank you

src/utils/format_json/open_telemetry.c
eero-t: Changes look good!

octo:
This is now ready for review.

This PR and  #4180 share commits that introduce the `resource_metrics` utility. I'm happy to create a separate PR just for that utility to make the review simpler. Just let me know.

eero-t:
> This PR and #4180 share commits that introduce the `resource_metrics` utility. I'm happy to create a separate PR just for that utility to make the review simpler. Just let me know.

Yes, please do that.

octo:
> Yes, please do that.

Done: #4196

eero-t:
Btw. according to this: https://github.com/influxdata/telegraf/issues/6205

OpenTelemetry **input** plugin would mean _collectd_ supporting also _Jaeger_ clients (applications providing cluster request tracing data).

octo:
OpenTelemetry knows three distinct types: metrics, traces, and logs. At the moment we only have support for metrics. I think collectd could potentially send its logs to OpenTelemetry, but I'm not sure if collectd would provide value as traces and/or logs broker.

eero-t:
Hm.  Traces appear to be _hierarchical_ / structured "logs": https://opentelemetry.io/docs/concepts/signals/traces/

=> Yeah, that could indeed be awkward to support.  Better to advance in smaller steps and get initial support ready first, and a release of it out.

(If given application already has network port open, I guess it could as well offer it to outside the node, not just to a node-local trace broker like collectd.)

octo:
Hey @eero-t, I'd like to merge this. Is there anything else you'd like to see changed before that?

src/utils/format_json/open_telemetry.c
```diff
@@ -0,0 +1,291 @@
...
+
+static int metric(yajl_gen g, metric_family_t const *fam) {
+  CHECK(yajl_gen_map_open(g)); /* BEGIN Metric */
+
+  CHECK(json_add_string(g, "name"));
+  CHECK(json_add_string(g, fam->name));
+
+  // TODO(octo): populate the "unit" field.
+  // CHECK(json_add_string(g, "unit"));
+  // CHECK(json_add_string(g, "1"));
```
eero-t, line 155: Unit field can be used now?

src/write_http.c
```diff
@@ -457,6 +523,13 @@ static int wh_write(metric_family_t const *fam, user_data_t *user_data) {
   case WH_FORMAT_INFLUXDB:
     status = wh_write_influxdb(fam, cb);
     break;
+  case WH_FORMAT_OTLP_PROTO:
```
eero-t, line 526: Nothing is setting/using `WH_FORMAT_OTLP_PROTO`, so IMHO it would be better to drop references to that completely from this PR.

eero-t (changes requested):
Looks OK on quick look, but I have few minor comments.

src/utils/format_json/format_json_test.c
```diff
@@ -288,10 +288,65 @@ DEF_TEST(metric_family_append) {
...
+      "\"version\":\"" PACKAGE_VERSION "\"},\"metrics\":[{\"name\":\"unit."
+      "test.count\",\"description\":\"Example counter "
+      "metric\",\"sum\":{\"dataPoints\":[{\"attributes\":[{\"key\":\"metric."
+      "label\",\"value\":{\"stringValue\":\"bar\"}}],\"timeUnixNano\":0,"
+      "\"asInt\":31337}],\"aggregationTemporality\":\"2\",\"isMonotonic\":true}"
+      "},{\"name\":\"unit.tests\",\"description\":\"Example gauge "
+      "metric\",\"gauge\":{\"dataPoints\":[{\"attributes\":[{\"key\":\"metric."
+      "label\",\"value\":{\"stringValue\":\"test "
+      "label\"}}],\"timeUnixNano\":0,\"asDouble\":42.0}]}}]}]}]}",
+      buf.ptr);
```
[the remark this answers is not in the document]
octo: Yes, metrics are sorted by the `resource_metrics` package to allow binary search. Metrics are sorted by metric family name first, then metric labels, then time stamp.

src/write_http.c
octo: That's fair. Done.

src/utils/format_json/open_telemetry.c
octo: Good point, added.

src/collectd.conf.pod
```diff
@@ -11267,14 +11268,36 @@ Define which SSL protocol version must be used. By default C<libcurl> will
...
-, will create output in the KairosDB format.
+=item B<KairosDB>
 
-Defaults to B<Command>.
+KairosDB is a Time Series Database based on Cassandra. This produces its JSON
+format.
+
+=item B<InfluxDB>
+
+InfluxDB is an open-source time series database. This produces its line
```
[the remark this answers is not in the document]
octo: Done

octo:
Thanks @eero-t!

PR #4191, opened 2023-12-14
[collectd 6] Load plugin: Migrate to collectd 6 data structures.

octo:
ChangeLog: Load plugin: Migrated to collectd 6 data structures.

### DIFF
--- a/src/load.c
+++ b/src/load.c
@@ -118,7 +118,7 @@ static int load_read(void) {
   else {
     WARNING("load: getloadavg failed: %s", STRERRNO);
   }
-    /* #endif HAVE_GETLOADAVG */
+  /* #endif HAVE_GETLOADAVG */
 
 #elif defined(KERNEL_LINUX)
   char buffer[64] = {0};
--- a/src/load.c
+++ b/src/load.c
@@ -73,7 +73,8 @@ static int load_config(const char *key, const char *value) {
   }
   return -1;
 }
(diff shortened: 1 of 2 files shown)


src/load.c
```diff
@@ -83,30 +84,29 @@ static void load_submit(gauge_t snum, gauge_t mnum, gauge_t lnum) {
...
 
-  vl.values = values;
-  vl.values_len = STATIC_ARRAY_SIZE(values);
-
-  sstrncpy(vl.plugin, "load", sizeof(vl.plugin));
-  sstrncpy(vl.type, "load", sizeof(vl.type));
-
   if (cores > 0) {
-    sstrncpy(vl.type_instance, "relative", sizeof(vl.type_instance));
+    fam.name = "system.load.scaled";
```
eero-t, line 95: IMHO reporting differently named metric should be a user decision, not one done by plugin itself.

eero-t (commented):
Changes look fine, but I have different opinion on metric naming decision.

src/load.c
octo: This is a bit hidden by the diff, but it *is* a user decision: line 81 evaluates `report_relative_load` and only sets `cores` if it is true. `report_relative_load` represents a configuration option.

eero-t: Ah, that's why it was doing `cores > 0` which I was wondering a bit about... What machine has zero CPUs? :-)

eero-t (approved):

PR #4192, opened 2023-12-15
 [collectd 6] IRQ plugin: Migrate to collectd 6 data structures. 

octo:
ChangeLog: IRQ plugin: Migrated to collectd 6 data structures.
Fixes: #3647

### DIFF
--- a/src/irq.c
+++ b/src/irq.c
@@ -69,6 +69,25 @@ static int irq_config(const char *key, const char *value) {
 }
 
 #if KERNEL_LINUX
+/* irq_strsplit is a special split function for Linux' /proc/interrupts file. It
+ * uses two or more spaces to separate fields. */
+static int irq_strsplit(char *string, char **fields, size_t fields_num) {
+  for (size_t i = 0; i < fields_num; i++) {
+    while (string[0] != 0 && isspace(string[0])) {
+      string[0] = 0;
+      string++;
+    }
+
+    fields[i] = string;
+    string = strstr(string, "  ");
+    if (string == NULL) {
+      return (int)i + 1;
+    }
+  }
+
+  return (int)fields_num;
+}
+
 static int irq_read_data(metric_family_t *fam) {
   /*
    * Example content:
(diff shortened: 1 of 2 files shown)


src/irq.c
```diff
@@ -68,153 +68,167 @@ static int irq_config(const char *key, const char *value) {
...
 
-    irq_value = 0;
-    for (i = 1; i <= irq_values_to_parse; i++) {
+    for (int i = 1; i <= irq_values_to_parse; i++) {
       /* Per-CPU value */
       value_t v;
-      int status;
-
-      status = parse_value(fields[i], &v, DS_TYPE_DERIVE);
+      int status = parse_value(fields[i], &v, DS_TYPE_COUNTER);
```
eero-t, line 179: DERIVE -> COUNTER?

src/irq.c
```diff
@@ -68,153 +68,167 @@ static int irq_config(const char *key, const char *value) {
...
+    }
+    /* strip colon */
+    irq_name[irq_name_len - 1] = 0;
+    irq_name_len--;
 
-    /* Check if irq name ends with colon.
-     * Otherwise it's a header. */
-    if (irq_name[irq_name_len - 1] != ':')
+    /* Is it the the ARM fast interrupt (FIQ)? */
+    if (strcmp(irq_name, "FIQ") == 0)
```
eero-t, line 153: IMHO this could be in user config, but that could be regression to old users.

eero-t (commented):
Looks otherwise OK, but I wonder about need for `irq_strsplit()`.

src/irq.c
```diff
@@ -68,153 +68,167 @@ static int irq_config(const char *key, const char *value) {
...
-
-    fields_num = strsplit(buffer, fields, STATIC_ARRAY_SIZE(fields));
+    strstripnewline(buffer);
+
+    /* There is one column with the interrupt ID before the CPU counters and
+     * there may be up to three columns after the counters. */
+    char *fields[cpu_count + 8];
+    memset(fields, 0, sizeof(fields));
+
+    int fields_num = irq_strsplit(buffer, fields, STATIC_ARRAY_SIZE(fields));
```
eero-t, line 131: irq_strsplit() is basically same as strsplit(), except for using isspace() & manual space removal, instead of strtok().  Why it's needed?

(Gah, my comment disappeared again somewhere.)

src/irq.c
octo: Yeah, I don't think this belongs here, but I don't fully understand why the code is there yet and so didn't want to remove it in the middle of everything else.

src/irq.c
octo: The difference is in the use of `strstr()`, which is used to look for two consecutive spaces. You cannot achieve this with `strtok()`, because it takes a list of characters that should act as separators, i.e. each character is considered individually rather than a substring to look for.

src/irq.c
eero-t: FIQ is just other type of interrupt:
* https://en.wikipedia.org/wiki/Fast_interrupt_request
* https://stackoverflow.com/questions/973933/what-is-the-difference-between-fiq-and-irq-interrupt-system

That code was added for: https://github.com/collectd/collectd/issues/971

But IMHO that was wrong fix.  It should check for the error condition, not IRQ name.

src/irq.c
```diff
@@ -68,153 +68,168 @@ static int irq_config(const char *key, const char *value) {
...
-    return;
-
-  vl.values = &(value_t){.derive = value};
-  vl.values_len = 1;
-  sstrncpy(vl.plugin, "irq", sizeof(vl.plugin));
-  sstrncpy(vl.type, "irq", sizeof(vl.type));
-  sstrncpy(vl.type_instance, irq_name, sizeof(vl.type_instance));
+#if KERNEL_LINUX
+/* irq_strsplit is a special split function for Linux' /proc/interrupts file.
+ * It uses two or more spaces to separate fields. Returns number of parsed
```
eero-t, line 73:
```suggestion
 * Unlike strsplit(), this requires at fields to be separated by at least two spaces. Returns number of parsed
```

src/irq.c
eero-t: Ah, I'll suggest another update to irq_strsplit() comment.

src/irq.c
octo: Yeah, don't know why this change was made. I think for version 6 we should move away from unsigned integers with overflow semantic. Reverted.

src/irq.c
```diff
@@ -68,153 +68,167 @@ static int irq_config(const char *key, const char *value) {
...
-    return;
-
-  vl.values = &(value_t){.derive = value};
-  vl.values_len = 1;
-  sstrncpy(vl.plugin, "irq", sizeof(vl.plugin));
-  sstrncpy(vl.type, "irq", sizeof(vl.type));
-  sstrncpy(vl.type_instance, irq_name, sizeof(vl.type_instance));
+#if KERNEL_LINUX
+/* irq_strsplit is a special split function for Linux' /proc/interrupts file. It
+ * uses two or more spaces to separate fields. */
```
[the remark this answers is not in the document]
octo: Applied.

src/irq.c
octo: Applied.

src/irq.c
eero-t: Problem with metric types is counters being integers (regardless of whether they're unsigned or signed), because _Prometheus_ / _OpenMetrics_ wants base units to be used for reporting.

Reporting usage seconds accurately (to differentiate between complete idle vs. very little usage) would require counters to be floating point, but `collectd` does not support that, so sysman plugin needs to report microseconds instead...

src/irq.c
octo: Okay, I'm pretty sure that the current code does not try to parse `NULL` as a value, and if it does we can fix that code instead. I'll revert that change.

There are a few rows in Linux that have a single column, e.g. `ERR`. They are currently attributed to "CPU 0", which is probably not 100% correct. I suspect FIQ was a similar case.

eero-t (approved):
Approved.

("single whitespace" would be less ambiguous term than "individual spaces", but it's good enough.)

octo:
Thanks for all your feedback @eero-t!

PR #4194, opened 2023-12-17
[collectd 6] Add a `unit` field to the `metric_family_t` struct.

octo:
Both, OpenMetrics and OpenTelemetry require that metrics specify their unit, such as "bytes per second". See for example OpenTelemetry's [Metrics Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units).

This adds a `unit` field to the `metric_family_t` field and automatically populates it based on the metric family name, if possible.

ChangeLog: daemon: A unit field has been added to the metric family struct.

### DIFF
--- a/src/daemon/unit.c
+++ b/src/daemon/unit.c
@@ -40,6 +40,8 @@ static bool string_has_suffix(char const *s, char const *suffix) {
 }
 
 static char const *default_unit_static(metric_family_t const *fam) {
+  // Determine units for some well-known name suffixes, see:
+  // https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-naming
   if (string_has_suffix(fam->name, ".utilization")) {
     return "1";
   }
--- a/src/daemon/unit.c
+++ b/src/daemon/unit.c
@@ -1,0 +1,70 @@
+/**
+ * collectd - src/daemon/unit.c
+ * Copyright (C) 2023       Florian "octo" Forster
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ * Authors:
+ *   Florian octo Forster <octo at collectd.org>
+ **/
+
+#include "collectd.h"
+#include "daemon/unit.h"
+
+static bool string_has_suffix(char const *s, char const *suffix) {
+  size_t s_len = strlen(s);
+  size_t suffix_len = strlen(suffix);
+
+  if (s_len < suffix_len) {
+    return false;
+  }
+
+  s += (s_len - suffix_len);
+  return strcmp(s, suffix) == 0;
+}
+
+static char const *default_unit_static(metric_family_t const *fam) {
+  if (string_has_suffix(fam->name, ".utilization")) {
+    return "1";
+  }
+  if (string_has_suffix(fam->name, ".time")) {
+    return "s";
+  }
+  if (string_has_suffix(fam->name, ".io")) {
+    return "By";
+  }
+  if (string_has_suffix(fam->name, ".operations")) {
+    return "{operation}";
+  }
+
+  return NULL;
+}
+
+char *default_unit(metric_family_t const *fam) {
+  if (fam->unit != NULL) {
+    return fam->unit;
+  }
+
+  char const *unit = default_unit_static(fam);
+  if (unit == NULL) {
+    return NULL;
+  }
+
+  return strdup(unit);
+}
--- a/src/daemon/unit.c
+++ b/src/daemon/unit.c
@@ -26,18 +26,7 @@
 
 #include "collectd.h"
 #include "daemon/unit.h"
-
-static bool string_has_suffix(char const *s, char const *suffix) {
-  size_t s_len = strlen(s);
-  size_t suffix_len = strlen(suffix);
-
-  if (s_len < suffix_len) {
-    return false;
-  }
-
-  s += (s_len - suffix_len);
-  return strcmp(s, suffix) == 0;
-}
+#include "utils/common/common.h"
 
 static char const *default_unit_static(metric_family_t const *fam) {
   // Determine units for some well-known name suffixes, see:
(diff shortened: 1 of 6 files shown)


src/cpu.c
```diff
@@ -440,33 +440,33 @@ static void aggregate(gauge_t *sum_by_state) /* {{{ */
...
 
   metric_t m = {0};
-  if (cpu_num == -1) {
-    metric_label_set(&m, "cpu", "total");
-  } else {
+  if (cpu_num != -1) {
     char cpu_num_str[32];
     ssnprintf(cpu_num_str, sizeof(cpu_num_str), "%d", cpu_num);
-    metric_label_set(&m, "cpu", cpu_num_str);
+    metric_label_set(&m, "system.cpu.logical_number", cpu_num_str);
```
eero-t, line 457: IMHO using hierarchical dot notation also in metric labels, in addition to the name of the metric itself, is redundant.

src/daemon/unit.c
```diff
@@ -0,0 +1,70 @@
...
+    return "1";
+  }
+  if (string_has_suffix(fam->name, ".time")) {
+    return "s";
+  }
+  if (string_has_suffix(fam->name, ".io")) {
+    return "By";
+  }
+  if (string_has_suffix(fam->name, ".operations")) {
+    return "{operation}";
```
eero-t, line 53: Using "1" (max value) instead of e.g. "ratio", "By" instead of "bytes", and curly braces for unit names, need reasoning why they are used (pointers to specs etc).  Maybe both in the docs and code itself.

src/cpu.c
```diff
@@ -440,33 +440,33 @@ static void aggregate(gauge_t *sum_by_state) /* {{{ */
...
+        &fam, "system.cpu.state", cpu_state_names[COLLECTD_CPU_STATE_ACTIVE],
+        (value_t){.gauge = rates[COLLECTD_CPU_STATE_ACTIVE] / global_rate_sum},
+        &m);
   } else {
     for (size_t state = 0; state < COLLECTD_CPU_STATE_ACTIVE; state++) {
-      m.value.gauge = 100.0 * rates[state] / sum;
-      metric_label_set(&m, "state", cpu_state_names[state]);
-      metric_family_metric_append(&fam, m);
+      metric_family_append(&fam, "system.cpu.state", cpu_state_names[state],
+                           (value_t){.gauge = rates[state] / global_rate_sum},
```
eero-t, line 468: This is changing public collectd API, so both the name and range changes require documentation updates.

eero-t (changes requested):

src/cpu.c
octo: Yeah, I agree with you. Labels are limited to the metric they are attached to, I don't see the need for introducing namespaces there either. Unfortunately this is what OpenTelemetry has [documented](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#processor-metrics) as "recommended".

src/daemon/unit.c
octo: Done.

src/cpu.c
eero-t: At the very top of that document is this statement:
> The system.* namespace SHOULD be exclusively used to report hosts’ metrics. The system.* namespace SHOULD only be used when the metrics are collected from within the target system. (physical servers, virtual machines etc). Metrics collected from technology-specific, well-defined APIs (e.g. Kubelet’s API or container runtimes) should be reported under their respective namespace (e.g. k8s., container.). Resource attributes related to a host, SHOULD be reported under the host.* namespace.

=> `collectd` should **not** be **hard-coding** `system.` prefix to anything

Component knowing in which environment `collectd` is running, can add it instead.

(In Kubernetes / Prometheus case it would be done with `metricRelabelings` / `relabelings` rule in `serviceMonitor` object for `collectd`.)

src/cpu.c
octo: I feel that the description (line 447) sufficiently explains the metric. Where else would you like to document it?

Yes, this is a change in the metric, just like any other name change also is. I'm trying to get this into shape before we do a release, so we don't have to change this later. My hope is that we get the core plugins (cpu, memory, disk, df) into a shape where we don't have to make any changes to the metrics, even after a "release candidate".

src/cpu.c
octo: collectd is reading these metrics directly from the operating system, e.g. from `/proc` on Linux. This is exactly what the `system` namespace is for, per the quote you have provided:

> The system.\* namespace SHOULD only be used when the **metrics are collected from within the target system**.

Other metrics, such as the HTTP requests served by Apache, SHOULD NOT go into the `system` namespace.

src/cpu.c
octo: You have a point, we should start writing a "migration guide" that summarizes the changes between collectd 5 and 6, and gives guidance on how to upgrade an existing setup.

re no demand: I know of two companies that are phasing out their use of collectd in favor of an OpenTelemetry based system. One company took a detour over Prometheus, the other didn't.

re too much change: putting the plugin updates in with the new `unit` field is not ideal. I'm happy to make separate PRs if you prefer.

re backwards compatibility: collectd 6 is **not** meant to be a drop-in replacement for collectd 5. It is a new major version specifically so that we are free to change data structures, behavior, integration points, etc. as needed to get to a base that we can build off of for the foreseeable future.

I think it is key to also provide interoperability between collectd 5 and 6. At the very least, a collectd 6 server should be able to accept metrics using the *network plugin*, so that not the entire fleet has to be upgraded in lockstep.

src/cpu.c
eero-t: Ah, should have read it closer.  Prefix was about used APIs, rather than whether that info is from host, VM or container.

(Whole /sys, most of /proc, and few devices from /dev, are mounted to all containers by default.)

src/cpu.c
eero-t: Prometheus does not accept metrics or labels with dots in their names, so this breaks `write_prometheus`.  At least for now, I would think working (well) with Prometheus to be more important than OpenTelemetry spec support. :-/

> re no demand: I know of two companies that are phasing out their use of collectd in favor of an OpenTelemetry based system.

Do you know whether their reasons were related to OpenTelemetry spec support, or those systems (which ones?) just being more actively developed & having releases in past few years?

eero-t: Ah, Google.  Then it's probably also "metrics collector has open network sockets and managed languages like Golang are better for security" + "let's avoid GPL by rewriting whole thing as we have resources for it".

(Downside of those Golang projects is that they do not allow linking to a non-Golang library, so *everything* required by a plugin really needs to be written from scratch in that language.   Which is not really a feasible option for something like Sysman functionality.)

eero-t: > You have a point, we should start writing a "migration guide" that summarizes the changes between collectd 5 and 6, and gives guidance on how to upgrade an existing setup.

Yes, but I guess I cannot ask that in this PR. :-)

> re too much change: putting the plugin updates in with the new unit field is not ideal. I'm happy to make separate PRs if you prefer.

Having one plugin in that PR as example of the usage for the new APIs, would be fine (e.g. memory plugin as changes in that seem smallest), with rest of plugins coming in other PR(s).

octo: > Having one plugin in that PR as example of the usage for the new APIs, would be fine (e.g. memory plugin as changes in that seem smallest), with rest of plugins coming in other PR(s).

Done, plugin updates moved to: #4211

src/daemon/unit.h
```diff
@@ -0,0 +1,44 @@
...
+#define DAEMON_UNIT_H 1
+
+#include "daemon/metric.h"
+
+/* default_unit tries to guess a metric family's unit.
+ *
+ * If fam->unit is not NULL, that pointer is returned. Otherwise, the function
+ * tries to heuristically determine a unit for th metric family. If successful,
+ * a new string is allocated on the heap and returned. This string must be freed
+ * using free(). If unsuccessful, NULL is returned.
```
eero-t, line 37: Fix typo and note that heuristics are based on OTEL metric names (i.e. units are not recognized for OpenMetrics / Prometheus metric names):
```suggestion
 * If fam->unit is not NULL, that pointer is returned. Otherwise, the function
 * tries to heuristically determine a unit for the metric family, based on known
 * OpenTelemetry metric names:
 * https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-naming
 * If successful, a new string is allocated on the heap and returned. This string must
 * be freed using free(). If unsuccessful, NULL is returned.
```

eero-t (changes requested):
Looks fine except for the commented issue.

octo:
> Looks fine except for the comment issue.

Suggestion committed. Thanks for the review and happy holidays! :christmas_tree:

eero-t (approved):

PR #4195, opened 2023-12-18
[collectd 6] Move legacy code into its own package.

octo:
This moves code related to `value_list_t` into `src/utils/value_list/`, and code related to `data_set_t` into `src/daemon/data_set.[ch]`. Because data sets are a global resource, they unfortunately need to remain within the daemon.

At the moment, `src/daemon/plugin.h` includes `src/utils/value_list/value_list.h` unconditionally. This allows plugins to compile without modification. I'm thinking about changing this, so that the include is guarded by a preprocessor define and legacy plugins need to do something like `#define LEGACY_PLUGIN_SUPPORT` to get access to the `value_list_t` data structure. It would prevent accidental use of this legacy infrastructure.

ChangeLog: The legacy "value_list_t" data structure and supporting functions have been moved into a separate component.


eero-t:
> I'm thinking about changing this, so that the include is guarded by a preprocessor define and legacy plugins need to do something like #define LEGACY_PLUGIN_SUPPORT to get access to the value_list_t data structure. It would prevent accidental use of this legacy infrastructure.

IMHO before that, code for those legacy plugins should be synced (copied again)  from `main` to 6.0 branch, as 6.0 branch specific changes in them will complicate later syncing.

eero-t (approved):
On very quick and casual glance of the changes, they looked OK.

eero-t:
> [Link value_list into libplugin_mock.](https://github.com/collectd/collectd/pull/4195/commits/bb7cdc7198649699d5b2c9717822b3fb666ec46c)

`curl_json` needs it still after that?

octo:
> `curl_json` needs it still after that?

Uhm, no. No idea how that ended up there, I was only trying to touch up tests … :shrug: Fixed.

PR #4196, opened 2023-12-18
[collectd 6] New utility: `resource_metrics` to group metrics by resource.

octo:
OpenTelemetry groups metric families by their `resource` field. This is going to be used by two upcoming changes:

* #4180 
* #4188

### DIFF
--- a/src/utils/resource_metrics/resource_metrics.c
+++ b/src/utils/resource_metrics/resource_metrics.c
@@ -167,32 +167,42 @@ static bool metric_exists(metric_family_t const *fam, metric_t const *m) {
 }
 
 static int insert_metrics(metric_family_t *fam, metric_list_t metrics) {
+  int skipped = 0;
   for (size_t i = 0; i < metrics.num; i++) {
-    if (metric_exists(fam, metrics.ptr + i)) {
-      return EEXIST;
-    }
-  }
+    metric_t const *m = metrics.ptr + i;
 
-  int last_err = 0;
-  for (size_t i = 0; i < metrics.num; i++) {
-    int status = metric_family_metric_append(fam, metrics.ptr[i]);
+    if (metric_exists(fam, m)) {
+#if COLLECT_DEBUG
+      strbuf_t buf = STRBUF_CREATE;
+      metric_identity(&buf, m);
+      DEBUG("resource_metrics: Skipping duplicate of metric %s", buf.ptr);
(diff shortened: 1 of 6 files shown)


src/utils/resource_metrics/resource_metrics.c
```diff
@@ -0,0 +1,240 @@
...
+    return ret;
+  }
+
+  int status = insert_resource(set, resource);
+  if (status != 0) {
+    ERROR("resource_metrics: insert_resource failed: %s", STRERROR(status));
+    return NULL;
+  }
+
+  ret = lookup_resource(set, resource);
```
eero-t, line 81: Maybe insert could return the inserted pointer, to avoid need for 2nd lookup?

src/utils/resource_metrics/resource_metrics.c
```diff
@@ -0,0 +1,240 @@
...
+    return ret;
+  }
+
+  int status = insert_family(rm, fam);
+  if (status != 0) {
+    ERROR("resource_metrics: insert_family failed: %s", STRERROR(status));
+    return NULL;
+  }
+
+  ret = lookup_family(rm, fam);
```
eero-t, line 142: Maybe insert could return the inserted pointer, to avoid need for 2nd lookup?

eero-t (approved):
Code and its test coverage looks fine.  I have minor optimization suggestion(s), but approved.

src/utils/resource_metrics/resource_metrics.c
octo: Yeah, I see where you're coming from. The problem with that is that *bsearch(3)* does not tell us where in the array the new item would be expected, and so we append and run *qsort(3)*. With the information where in the array the element should go, we could implement this more efficiently.

However, that requires us to write our own binary search, do more memory management, etc. which makes the code a good deal harder to read and maintain. I think we should only go down that road after profiling has shown us that this is actually a bottleneck. "Premature optimization is the root of all evil" is one of my guiding principles. (Not that I'm consistently succeeding, there are plenty of instances in the collectd code base where I failed …)

src/utils/resource_metrics/resource_metrics.c
octo: Heh, this is very tempting ;) In the interest of not getting nerd-sniped at 23:30 I'm going to stick with "maybe later". :wink:

PR #4197, opened 2023-12-18
[collectd 6] Generate better names and labels for legacy metrics.

octo:
Legacy metrics, i.e. metrics reported with the `value_list_t` data structure, need to be upgraded to `metric_family_t`.

The new code puts all metrics into the `collectd.v5` namespace. The metric family name is:

* `collectd.v5.${plugin}.utilization` for metrics with `type` "percent".
* `collectd.v5.${plugin}.io` for *directional metrics* ending in `_octets`.
* `collectd.v5.${type}.${data_source}` for metrics with more than one data source (except *directional metrics*).
* `collectd.v5.${type}` for metrics with one data source.

If the `hostname` field is set, it is translated into the `host.name` resource attribute. Otherwise resource attributes are omitted to trigger the default behavior (like an empty `hostname` field does).

The labels depend on whether plugin instance, type instance, or both are set:

* Both: `${plugin}=${plugin_instance}`, `type=${type_instance}`
* Plugin instance: `${plugin}=${plugin_instance}`
* Type instance: `${plugin}=${type_instance}`
* Neither: `plugin=${plugin}`

This does a decent job of mapping metrics to the new schema, but I do wonder if we should simply put those fields into labels 1-to-1: `collectd.v5.plugin=${plugin}`, `collectd.v5.plugin_instance=${plugin_instance}`, `collectd.v5.type_instance=${type_instance}`. What do you think?

**Directional metrics** are metrics with two data sources called "read" and "write", or "rx" and "tx". Despite having more than one data source, the data source name is added to a `direction` label instead of the metric family name.

ChangeLog: The schema with which metrics are translated from v5 to v6 has been improved.

### DIFF
--- a/src/utils/value_list/value_list.c
+++ b/src/utils/value_list/value_list.c
@@ -408,6 +408,9 @@ plugin_value_list_to_metric_family(value_list_t const *vl, data_set_t const *ds,
     }
     status = status || metric_label_set(&m, name, vl->type_instance);
   }
+  if (strlen(vl->plugin_instance) == 0 && strlen(vl->type_instance) == 0) {
+    status = status || metric_label_set(&m, "plugin", vl->plugin);
+  }
 
   status = status || metric_family_metric_append(fam, m);
   if (status != 0) {
--- a/src/utils/value_list/value_list.c
+++ b/src/utils/value_list/value_list.c
@@ -187,30 +187,6 @@ plugin_dispatch_multivalue(value_list_t const *template, /* {{{ */
   return failed;
 } /* }}} int plugin_dispatch_multivalue */
 
-static int metric_family_name(strbuf_t *buf, value_list_t const *vl,
-                              data_source_t const *dsrc) {
-  int status = strbuf_print(buf, "collectd");
-
-  if (strcmp(vl->plugin, vl->type) != 0) {
-    status = status || strbuf_print(buf, "_");
-    status = status || strbuf_print(buf, vl->plugin);
-  }
-
-  status = status || strbuf_print(buf, "_");
-  status = status || strbuf_print(buf, vl->type);
-
-  if (strcmp("value", dsrc->name) != 0) {
-    status = status || strbuf_print(buf, "_");
-    status = status || strbuf_print(buf, dsrc->name);
-  }
-
-  if ((dsrc->type == DS_TYPE_COUNTER) || (dsrc->type == DS_TYPE_DERIVE)) {
-    status = status || strbuf_print(buf, "_total");
-  }
-
-  return status;
-}
-
 int parse_identifier(char *str, char **ret_host, char **ret_plugin,
                      char **ret_type, char **ret_data_source,
                      char *default_host) {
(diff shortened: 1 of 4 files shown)


src/utils/value_list/value_list.c
```diff
@@ -416,6 +430,9 @@ plugin_value_list_to_metric_family(value_list_t const *vl, data_set_t const *ds,
     }
     status = status || metric_label_set(&m, name, vl->type_instance);
   }
+  if (strlen(vl->plugin_instance) == 0 && strlen(vl->type_instance) == 0) {
```
eero-t, line 433:
```suggestion
  else if (strlen(vl->plugin_instance) == 0) {
```

eero-t (commented):
Code looks OK.  Have just small optimization suggestion(s).

eero-t:
> This does a decent job of mapping metrics to the new schema, but I do wonder if we should simply put those fields into labels 1-to-1: `collectd.v5.plugin=${plugin}`, `collectd.v5.plugin_instance=${plugin_instance}`, `collectd.v5.type_instance=${type_instance}`. What do you think?

No need for hierarchical labels, metric name being such is enough.

src/utils/value_list/value_list.c
```diff
@@ -369,10 +345,37 @@ metric_t *parse_legacy_identifier(char const *s) {
...
+    strbuf_printf(buf, ".%s.utilization", vl->plugin);
+  } else if (is_directional_metric(ds) &&
+             string_has_suffix(ds->type, "_octets")) {
+    strbuf_printf(buf, ".%s.io", vl->plugin);
+  } else {
+    strbuf_printf(buf, ".%s", vl->type);
+  }
+
+  if (ds->ds_num > 1 && !is_directional_metric(ds)) {
+    status = status || strbuf_printf(buf, ".%s", ds->ds[index].name);
```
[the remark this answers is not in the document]
octo: I submitted something very similar. The changes to your suggestion are:

* Switch to `strbuf_print` (no "f") when dropping the format string. Otherwise we could potentially be susceptible to to format string exploits.
* We need to keep the period in line 369, since it adds to the metric name.

src/utils/value_list/value_list.c
octo: The two previous `if` blocks can both be executed. The last block is only executed if *neither* of the first two blocks is executed. Using an `else if` here makes it seem like the last two blocks are more closely related than they actually are, so I prefer to use just an `if` here.

octo: On second thought, I think the control flow can be improved and I have done this in 215eee3f791833408a5e43c9b5682481bfbdff93.

eero-t: Much clearer, thanks!

src/utils/value_list/value_list.c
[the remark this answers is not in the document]
eero-t: Good catch, I would have been embarrassed to introduce bug. :-)

eero-t (approved):

eero-t:
This breaks Prometheus integration.  Dot is invalid in metric and label names: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

PR #4199, opened 2023-12-19
[collectd 6] Add an option to configure resource attributes.

octo:
ChangeLog: daemon: The new "Resource" option allows users to configure resource attributes manually if desired.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -328,6 +328,51 @@ the daemon should try to figure out the "fully qualified domain name", FQDN.
 This is done using a lookup of the name returned by C<gethostname>. This option
 is enabled by default.
 
+=item B<Resource> I<Type>
+
+The B<Resource> block defined the I<Resource Attributes> to use for local
+metrics such as CPU usage, i.e. metrics that were not received from a remote
+system or service.
+
+The recommended way to configure resource attributes is by following the
+OpenTelemetry Semantic Conventions,
+L<https://opentelemetry.io/docs/specs/semconv/resource/>.
+
+The I<Type> argument determines which default attributes are automatically
+populated. B<Attribute> options inside the block can be used to add additional
+attributes or overwrite existing ones.
+
+Valid values for I<Type> are:
+
+=over 4
+
+=item B<Host> (default)
+
+For use with computing instances (e.g. physical and virtual machines). Will
+automatically populate the C<service.name>, C<host.name>, and C<host.id>
+attributes.
+
+For more information, refer to
+L<https://opentelemetry.io/docs/specs/semconv/resource/host/>.
+
+=item B<Generic>
+
+Minimal defaults for maximum flexibility. Only the C<service.name> attribute is
+populated.
+
+=back
+
+Inside the B<Resource> block, the following configuration options are valid:
+
+=over 4
+
+=item B<Attribute> I<Name> I<Value>
+
+Adds a resource attribute or updates an existing one. If I<Value> is the empty
+string, the attribute is deleted.
+
+=back
+
 =item B<PreCacheChain> I<ChainName>
 
 =item B<PostCacheChain> I<ChainName>
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -362,6 +362,13 @@ L<https://opentelemetry.io/docs/specs/semconv/resource/host/>.
 Minimal defaults for maximum flexibility. Only the C<service.name> attribute is
 populated.
 
+To also remove the C<service.name> resource attribute, set it to an empty
+string:
+
+  <Resource Generic>
+    Attribute "service.name" ""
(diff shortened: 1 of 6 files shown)


src/collectd.conf.pod
```diff
@@ -328,6 +328,51 @@ the daemon should try to figure out the "fully qualified domain name", FQDN.
 This is done using a lookup of the name returned by C<gethostname>. This option
 is enabled by default.
 
+=item B<Resource> I<Type>
+
+The B<Resource> block defined the I<Resource Attributes> to use for local
+metrics such as CPU usage, i.e. metrics that were not received from a remote
```
eero-t, line 334: Reads better to me when tenses are switched:
```suggestion
B<Resource> block defines I<Resource Attributes> to be used for local
metrics such as CPU usage, i.e. metrics that were not received from a remote
```

eero-t (changes requested):
Code changes look fine.   But I think documentation needs some improvements:
* Name is too generic, so some short explanation how this "Resource" differs from all the other "Resource" items already mentioned in the document, would be good
* Text is on the abstract side and IMHO would need some example of how it changes collectd output, so users unfamiliar with OpenTelemetry know what it's about

eero-t:
There should be also `None` option for disabling resource attribute labels completely, because resource label names use characters that are not valid for Prometheus metrics, see https://github.com/collectd/collectd/issues/4204 and: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

eero-t:
> There should be also `None` option for disabling resource attribute labels completely, because resource label names use characters that are not valid for Prometheus metrics, see #4204 and: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

That issue is now fixed, but need for disabling the resource labels has gone nowhere, because they just duplicate information that Prometheus **already** adds to the metrics.

octo:
My thinking was that the following config should give you the same effect as a `None` resource:

```
<Resource Generic>
  Attribute "service.name" ""
</Resource>
```

Can you tell me more how the resource attributes and the Prometheus export clash? It's important to me to get this right.

src/collectd.conf.pod
octo: Fixed the typo. I prefer the active voice though.

eero-t:
> My thinking was that the following config should give you the same effect as a `None` resource:

True, but I'm assuming that you're going to be adding more attributes in the future?

`None` option would be a more user-friendly way of making sure that all extra labels are disabled.

> Can you tell me more how the resource attributes and the Prometheus export clash? It's important to me to get this right.

Problem is that extra labels (on every metric) can noticeably increase Prometheus resource usage in larger clusters, so it's important that one can easily disable anything that is not needed, especially _when Prometheus already provides corresponding information_ in (more relevant kubernetes namespace) labels.

octo:
> > My thinking was that the following config should give you the same effect as a `None` resource:
> 
> True, and it's a useful feature, but I'm assuming that you're going to be adding more attributes in the future?

My design idea is this:

* Metrics that already have resource attributes, e.g. because they were received from somewhere else via gRPC or something similar, are simply passed through. No resource attributes are added, removed, or modified.
* Metrics that are passed to `plugin_dispatch_metric_family` without resource attributes are considered to be "local", i.e. as originating from this collectd instance. The resource attributes associated with this instance of collectd are attached to the metric family.
* Strike a good balance between auto-discovery and user control. Users likely don't want to manually configure the host name. If you're managing a large fleet, having a different config file for each host is quite annoying.
    * Provide presets that users can chose by passing an appropriate argument to the `Preset` block. At the moment, `Generic` and `Host` are the only available presets. Other presets, e.g. for GCE instances, K8s pods, etc. might follow.
    * Allow users to amend or overwrite the defaults, using the `Attribute` option inside the `Resource` block.

> `None` option would be a more user-friendly way of making sure that _all_ extra labels are disabled.

I think you're coming from a position where these labels cause problems with the *write_prometheus plugin*. Once these are properly integrated (I'm working on this in #4213) I think that the existing `Generic` preset provides more utility for you than a `None` preset, because they allow you to set the `job` and `instance` labels appropriately. 

> Problem is that extra labels (on every metric) can noticeably increase Prometheus resource usage in larger clusters, so it's important that one can easily disable anything that is not needed, especially _when Prometheus already provides corresponding information_ in (more relevant kubernetes namespace) labels.

Gotcha, thanks for the info! IMHO you have two options already:

1. Use the `Generic` preset and unset the one resource attribute it installs if required, as outlined above.
2. Use metric relabeling in Prometheus to discard unwanted labels.

I'm sure collectd will gain additional features in the future, but we have to start somewhere, and this seems like quite a good position to be in for a first step.

eero-t (approved):
Ok, seems reasonable, especially considering https://github.com/collectd/collectd/pull/4213.

PR #4201, opened 2023-12-19
[collectd 6] .github/workflows/build.yml: Remove Ubuntu Bionic.

octo:
It is too old to work with `actions/checkout@v4`.

This is a prerequisite for #4180

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,6 @@ jobs:
           - mantic_amd64
           - jammy_amd64
           - focal_amd64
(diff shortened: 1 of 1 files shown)


eero-t (approved):
Approved, Bionic dropped from free (5 year) security support already earlier this year.

eero-t:
The expected plugins list would really be good to fix, to avoid CI failures...

octo:
> The expected plugins list would really be good to fix, to avoid CI failures...

ACK, I'll spend some time on this today.

PR #4202, opened 2023-12-20
[collectd 6] liboconfig: Apply automatic fixes.

octo:
Command:

```
bison --update src/liboconfig/parser.y
```

ChangeLog: liboconfig: Automatic fixes have been applied to the parser.


eero-t:
Really annoying.  `release_ready` does not warn about this any more:
```
/tmp/cirrus-ci-build/src/liboconfig/parser.y:78.1-14: warning: deprecated directive: '%error-verbose', use '%define parse.error verbose' [-Wdeprecated]
   78 | %error-verbose
      | ^~~~~~~~~~~~~~
      | %define parse.error verbose
```

But now e.g. `jammy` build gives:
```
 /__w/collectd/collectd/src/liboconfig/parser.y:78.1-7: warning: POSIX Yacc does not support %define [-Wyacc]
   78 | %define parse.error verbose
      | ^~~~~~~
```

Are different YACC versions and options used on different distros?

octo:
`make dist` creates a tar archive containing the generated `.c` and `.h` files. Users that download a release are no going to run into these issues.

Unfortunately, our CI system always starts from scratch and needs to generate the code, and thus we run into these incompatibilities.

What we might be able to do is run the `make dist` step first, and upload the tar archive as a build artifact. The other builders can then depend on this and build from a tar archive, like normal users.

Pro:
* Closer to user experience
* Solves this immediate compatibility problem
* Uses fewer CPU resources, particularly when `make dist` already fails
* Support for git submodules need only exist for the `make dist` run, others don't access Git at all

Con:
* Increases the wall-clock time needed to run all builders because they block on the first `make dist` run
* Dependencies between workflows introduce additional breakage points

Overall, my feeling is that this would be a net positive change.

eero-t:
I wonder how relevant it's to test that `make dist` succeeds on different distros versions.  At least that could use a smaller subset of them...

octo:
We run `make distcheck` (a superset of `make dist`) on only one distro, Debian 12 ("Bookworm", aka. Debian Stable).

The other builders need to generate the `.c` file because they build from a Git repo, not because they run `make dist` (they don't).

eero-t:
> We run `make distcheck` (a superset of `make dist`) on only one distro, Debian 12 ("Bookworm", aka. Debian Stable). I have changed this somewhat recently, previously we ran `make distcheck` on every distro, which took a long time …

I guess it would be enough to do every-distro `make distcheck` only before releases.  But I wonder whether it would make sense run run it regularly both on the oldest & newest distro, not just one stable distro, as both very old, and very new  tooling versions could cause issues for `collectd`.

(E.g. debian oldstable + latest fedora, or oldest rhel + debian testing.)

PR #4207, opened 2023-12-21
[collectd 6] Link more things into `libplugin_mock.la`.

octo:
With this change `libplugin_mock` contains more of the symbols provided by the collectd daemon, making it easier to link unit tests, especially for plugins.


eero-t (approved):
Approved
* `libmetadata` comes through `libmetric`
* `libmetric` | `liboconfig` | `libcommon` | `libignorelist` through `libmock`
* `libavltree` is added to things that were earlier using `avltree.c` directly

PR #4208, opened 2023-12-21
.cirrus.yml: Provide expected plugin list to `check-built-plugins.sh` if present.

octo:


.github/workflows/build.yml
```diff
@@ -77,7 +77,7 @@ jobs:
       run: exit $(< make-check.status)
 
   experimental:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
eero-t, line 80: The other `runs-on` statement is also for `ubuntu-latest`, but where that name actually comes from?  Wouldn't it make sense to use "fixed" distro version name here too?

.expected-plugins/centos9
```diff
@@ -0,0 +1,113 @@
...
+chrony
+connectivity
+conntrack
+contextswitch
+cpu
+cpufreq
+cpusleep
+curl
+curl_json
+dbi
```
eero-t, line 20: Other lists have also `curl_xml`.

eero-t (commented):
There are much more enabled plugins than disabled (commented out) ones.

=> Maybe it would make more sense to invert the lists purpose, i.e. list only plugins that should NOT be tested on given distro?

(If there are plugins that should not be tested on any distro, that could be one additional list.)

PS. Some lists for distros (listed in workflows) are still empty, I guess they'll get filled later on?

.expected-plugins/centos9
octo: I have created these lists in slightly different ways. For CentOS 9, I have created an empty file, and then imported all the plugins there were built by CirrusCI. For Debian 12, I have imported the list from `main` and then commented out all the non-working plugins. The *curl_xml plugin* is one that is commented out.

.github/workflows/build.yml
octo: `ubuntu-latest` is a tag on the Ubuntu Docker image: https://hub.docker.com/_/ubuntu

Apparently it refers to the latest LTS version, i.e. currently 22.04. This is a rather conservative choice, IMHO.

I think this is preferable to pinning to a specific version, like `ubuntu-20.04`, because these tend to get outdated quickly and we need to spend more time than necessary keeping the CI system alive.

.expected-plugins/centos9
eero-t: Any thoughts on those files listing only plugins that should NOT be tested?

It should be a much smaller set, and therefore much easier to review & maintain in future...

octo: An inverted list wouldn't provide the same value IMHO. The value this system provides is that it alerts us to the fact if a plugin that used to build on, say, Debian Unstable stops being built there.

With an inverted list, we would have a list of plugins that must not be built. This would trigger when a contributor ports a plugin or migrates a plugin to a more modern dependency. These are not really cases we need to be *alerted* on.

I do see the point about maintenance, though. We could expand the check script to accept multiple files. We could use that to move all the "will build on all Linux distros" plugins to a central location and only maintain lists of plugins in which the distros differ. What do you think?

eero-t: > I do see the point about maintenance, though. We could expand the check script to accept multiple files. We could use that to move all the "will build on all Linux distros" plugins to a central location and only maintain lists of plugins in which the distros differ. What do you think?

I was thinking that there would a common "do-build" list, and per-distro "do-not-build" _exceptions_ for plugins on that list:
`grep -v -f $DISTRO-exceptions.txt all-plugins-to-build.txt > plugins-to-build-on-$DISTRO.txt`

In general same plugins should be buildable on all distros, right?  So exceptions should be really rare.

Having explicit  "do-not-build" lists would IMHO both make the lists easier to maintain, **and** very clearly document what those distro specific exceptions are.


> With an inverted list, we would have a list of plugins that must not be built. This would trigger when a contributor ports a plugin or migrates a plugin to a more modern dependency. These are not really cases we need to be *alerted* on.

Unless new plugin is added to common "do-build" list, it would not be built.

octo: I think this is a very similar approach. The important aspect (to me) is that it breaks out a "base line", with per-distro specialization. Whether that specialization is an allow or deny list is not a huge difference IMHO.

There is a rather big difference between operating systems. Many plugins should build on any Linux distro, but won't build on FreeBSD. That's why I thought composition would be a good way of reducing duplication while allowing us to have different base lines for Linux and FreeBSD, for example.

eero-t: > There is a rather big difference between operating systems. Many plugins should build on any Linux distro, but won't build on FreeBSD.

There are more Linux distros in CI than the others, so I would think per-distro deny lists deducted from "all-plugins", would in total be less lines than per-distro allow lists added to "base-plugins" one, but I have no idea how significant the difference is...

PR #4213, opened 2023-12-22
[collectd 6] Improve Write Prometheus' handling of resource attributes.

octo:
This PR aims to implement the "target info" metrics described in the [guidance on push based systems from the OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems). It also heavily relies on guidance provided in https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/.

* Calculate a set of all resources, called `target_info_t`.
* Emit a `target_info` gauge metric family.
* Emit one `target_info` metric for each "resource" in the set. The resource attributes `service.name` and `service.instance.id` are mapped to the special Prometheus labels `job` and `instance` respectively. Resource attributes are only emitted once, in the special "target info metric". Other metrics are exported with `job` and `instance` labels (if available) but no other resource attributes.
* Re-implement the logic for sanitizing metric names and label names based on the OpenTelemetry integration.

ChangeLog: Write Prometheus plugin: Resource attributes have been integrated into the plugin. "service.name" and "service.instance.id" are exported as "job" and "instance" labels. All other resource attributes are exported once in the `target_info` metric family.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -130,8 +130,8 @@ noinst_LTLIBRARIES = \
 	libavltree.la \
 	libcmds.la \
 	libcommon.la \
-	libformat_influxdb.la \
 	libformat_graphite.la \
+	libformat_influxdb.la \
 	libheap.la \
 	libignorelist.la \
 	liblatency.la \
@@ -141,6 +141,7 @@ noinst_LTLIBRARIES = \
 	libmetric.la \
 	libmount.la \
 	liboconfig.la \
+	libresource_metrics.la \
 	libstrbuf.la
 
 
@@ -386,13 +387,6 @@ test_utils_message_parser_SOURCES = \
 test_utils_message_parser_CPPFLAGS = $(AM_CPPFLAGS)
 test_utils_message_parser_LDADD = libplugin_mock.la -lm
 
-test_utils_resource_metrics_SOURCES = \
-	src/utils/resource_metrics/resource_metrics_test.c \
-	src/utils/resource_metrics/resource_metrics.c \
-	src/utils/resource_metrics/resource_metrics.h \
-	src/testing.h
-test_utils_resource_metrics_LDADD = libplugin_mock.la
-
 test_utils_time_SOURCES = \
 	src/daemon/utils_time_test.c \
 	src/testing.h
@@ -579,6 +573,15 @@ test_utils_mount_LDADD = \
 if BUILD_WITH_LIBKSTAT
 test_utils_mount_LDADD += -lkstat
 endif
+
+libresource_metrics_la_SOURCES = src/utils/resource_metrics/resource_metrics.c \
+				 src/utils/resource_metrics/resource_metrics.h
+libresource_metrics_la_LIBADD = libcommon.la
+
+test_utils_resource_metrics_SOURCES = \
+	src/utils/resource_metrics/resource_metrics_test.c \
+	src/testing.h
+test_utils_resource_metrics_LDADD = libresource_metrics.la libplugin_mock.la
 
 libstrbuf_la_SOURCES = \
 		       src/utils/strbuf/strbuf.c \
--- a/src/daemon/metric.c
+++ b/src/daemon/metric.c
@@ -64,8 +64,8 @@ static int label_name_compare(void const *a, void const *b) {
                 ((label_pair_t const *)b)->name);
 }
 
-static label_pair_t *label_set_read(label_set_t const *labels,
-                                    char const *name) {
+static label_pair_t *label_set_lookup(label_set_t const *labels,
+                                      char const *name) {
   if (name == NULL) {
     errno = EINVAL;
     return NULL;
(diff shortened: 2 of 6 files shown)


octo:
[Supporting Target Metadata in both Push-based and Pull-based Systems](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems) says:

> An exposition from a general purpose monitoring system may contain metrics from many individual targets, and thus may expose multiple target info Metrics. The metrics may already have had target metadata added to them as labels as part of ingestion. The metric names MUST NOT be varied based on target metadata. For example it would be incorrect for all metrics to end up being prefixed with staging_ even if they all originated from targets in a staging environment).

There also is a "Prometheus exporter" for OpenTelemetry. We should probably aim to provide a similar functionality to that.

octo:
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/translator/prometheus has a more comprehensive list of translations that are applied to OpenTelemetry metrics.

octo:
Next steps:

1.  Add a feature to `utils/resource_metrics/` allowing us to specify whether a new instance of a `metric_t` should be added to the metric family (current behavior, geared towards plugins that are flushed periodically and that should not lose any data point), or whether the new metric replaces the previous one (for write_prometheus, which is only interested in the latest state of each metric).
2.  Change *write_prometheus* to use `resource_metrics`, so we can group metric families by resource attributes.
3.  Emit one target into metric for reach resource.

Further steps (depend on the `unit` field, #4194):

* Add a unit specific suffix, as documented here: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheus/README.md

octo:
Currently, the code produces this:

```
# TYPE target info
# HELP target Target metadata
target_info{job="collectd-testing",instance="tmp",host_id="911c4cdca88ae8847dbaab5b6358febf",host_name="example.com"} 1
# HELP system_load System load average over a given duration
# TYPE system_load gauge
system_load{job="collectd-testing",instance="tmp",duration="15m"} 0.38916015625 1703363892613
system_load{job="collectd-testing",instance="tmp",duration="1m"} 0.35693359375 1703363892613
system_load{job="collectd-testing",instance="tmp",duration="5m"} 0.34033203125 1703363892613

# collectd/write_prometheus 5.12.0.438.g6ef83a7+ at example.com
```

In the target info metric, the metric name doesn't seem to match (`target` vs. `target_info`). According to open-telemetry/opentelemetry-specification#3700 this is correct for OpenMetrics, but Prometheus does something different (gauge metric + naming convention ("_info" suffix)).

octo:
From https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#resource-attributes-1:

> If info-typed metric families are not yet supported by the language Prometheus client library, a gauge-typed metric family named target_info with a constant value of 1 MUST be used instead.

It's unclear to me how this works with multiple resource metrics though.

octo:
Prometheus 2.48.1 does not (yet?) support metric families with type "info". Changed to a "gauge" metric, which is the convention for Prometheus.

eero-t (commented):
Code looks fine, but I'm wondering about two things:
* AFAIK resources are (normally?) set only once at collectd startup, so re-parsing them on every metric output seems a bit excessive
* Metric name coercion part is relevant also for v5 collectd Prometheus plugin.  Any thoughts on extracting relevant pieces for v5?

octo:
At the moment I focus on exporting metrics, as that is IMHO the core of what collectd does. So for the time being you're right, all metrics will have the same resource attributes. I expect this to change eventually, though. Let me give you two examples to illustrate where metrics with other resource attributes may come from:

1. Pushing external metrics to collectd. For example, I'd like to change the existing gRPC plugin to implement the `MetricsService.Export` method, allowing collectd to receive metrics from elsewhere, possibly other instances of collectd. Prometheus calls this "federation".
2. collectd pulling external metrics. For example, the *SNMP plugin* can be configured to query a large number of devices. The data provided by SNMP follows a pre-defined schema (called "MIB"), so it is very likely that two identical metrics from multiple devices.

I haven't looked at the v5 code of the *Write Prometheus plugin* yet. For v5, we guarantee backwards compatibility, so we can't freely change the way we form metric names. Any new logic would need to be implemented in addition to the existing logic, and guarded by a config option – not sure if that is worth the trouble. What aspect do you think we need to improve about the v5 behavior?

eero-t (approved):
Approved.

PS. I really like that you're adding also test-cases for the new code!

octo:
Thanks @eero-t! I'm trying out the TDD paradigm with some of these changes, and so far I like its influence on the design and code :D

eero-t:
Checked quickly resource usage of the new version.  There's no change in the (very small) CPU load, and collectd RSS (with just  sysman + prometheus plugins enabled) decreased marginally (~2%), so all is good from perf point of view.

eero-t:
> * [x]   Add a feature to `utils/resource_metrics/` allowing us to specify whether a new instance of a `metric_t` should be added to the metric family (current behavior, geared towards plugins that are flushed periodically and that should not lose any data point), or whether the new metric replaces the previous one (for write_prometheus, which is only interested in the latest state of each metric).

This part seems to be broken.  It should consider metrics with different resources (label set label values) as unique, but  adding multiple metrics with same name & labels and only differing through separate metric family resource label sets, causes all other metrics, other than the ones for the first resource, to be ignored.

(Noticed while implementing #4267, and moving per-GPU metric labels to a resource label set.)

eero-t:
Prometheus plugin unit tests check for formatting of multiple resources working fine, so I've started going through collectd core code starting from `plugin_dispatch_metric_family()` onward, but haven't found yet where the metric families given to it get lost (that differ only in their family resource labels, instead of metric labels).

PR #4214, opened 2023-12-26
[collectd 6] src/daemon/metric.c: Fix memory corruption in `label_set_delete`.

octo:
The size parameter passed to `memmove` was in number of pairs, not bytes.

ChangeLog: Daemon: Memory corruption in the "label_set_delete" function has been fixed.

### DIFF
--- a/src/daemon/metric.c
+++ b/src/daemon/metric.c
@@ -257,8 +257,8 @@ int metric_reset(metric_t *m) {
   return 0;
 }
 
-static int format_label_set(strbuf_t *buf, label_set_t const *labels,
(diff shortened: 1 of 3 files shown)


eero-t (approved):
Looks fine -> approved.

(I'm only wondering whether `format_label_set()` would be more general useful, than just for `metric_test.c`.)

octo:
> (I'm only wondering whether `format_label_set()` would be more general useful, than just for `metric_test.c`.)

There is another implementation in `src/daemon/metric.c`, which I have made public and used in the test now.

PR #4217, opened 2023-12-29
[collectd 6] disk plugin: Align metrics with OpenTelemetry recommendations.

octo:
* Rename labels to `system.device` and `disk.io.direction`.
* Rename `system.disk.time` to `system.disk.operation_time`.
* Add descriptions and units to all metric families.
* Add the "utilization" metric to FreeBSD.

ChangeLog: Disk plugin: The metric schema has been aligned with OpenTelemetry semantic conventions.

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -1092,14 +1092,6 @@ static int disk_read(void) {
     metric_family_append(&fam_ops, direction_label, write_direction,
                          (value_t){.counter = kio.KIO_WOPS}, &m);
 
-    if (strncmp(ksp[i]->ks_class, "disk", 4) == 0) {
-      /* FIXME: Convert this to microseconds if necessary */
-      metric_family_append(&fam_ops_time, direction_label, read_direction,
-                           (value_t){.counter = kio.KIO_RTIME}, &m);
-      metric_family_append(&fam_ops_time, direction_label, write_direction,
-                           (value_t){.counter = kio.KIO_WTIME}, &m);
-    }
-
     metric_reset(&m);
   }
   /* #endif defined(HAVE_LIBKSTAT) */
--- a/src/disk.c
+++ b/src/disk.c
@@ -849,12 +849,6 @@ static int disk_read(void) {
     value_to_rate(&io_time_rate_ms, (value_t){.counter = io_time_ms},
                   DS_TYPE_COUNTER, cdtime(), &ds->io_time_state);
 
-    /* Skip first cycle for newly-added disk */
-    if (ds->poll_count == 0) {
-      DEBUG("disk plugin: (ds->poll_count = 0) => Skipping.");
-      ds->poll_count = poll_count;
-      continue;
-    }
     ds->poll_count = poll_count;
 
     if ((read_ops == 0) && (write_ops == 0)) {
--- a/src/disk.c
+++ b/src/disk.c
@@ -394,9 +394,9 @@ static int disk_read(void) {
   };
   metric_family_t fam_ops = {
       .name = "system.disk.operations",
-      .help = "Read and write operations performed by the disk. If multiple "
-              "operations are merged into one by the disk, they will be "
-              "accounted for separately.",
+      .help = "Read and write operations performed by the disk. Operations are "
+              "reported separately even if they were merged into one operation "
+              "by the I/O subsystem.",
       .unit = "{operation}",
       .type = METRIC_TYPE_COUNTER,
   };
@@ -421,10 +421,10 @@ static int disk_read(void) {
   metric_family_t fam_disk_io_weighted_time = {
       .name = "system.disk.weighted_io_time",
       .help = "This metric is incremented at each I/O start, I/O completion, "
-              "or I/O merge by the number of I/Os in progress times the number "
-              "of milliseconds spent doing I/O since the last update of this "
-              "field.  This can provide an easy measure of both I/O completion "
-              "time and the backlog that may be accumulating.",
+              "or I/O merge by the number of I/Os in progress, times the "
+              "number of milliseconds spent doing I/O since the last update of "
+              "this field. This can provide an easy measure of both I/O "
+              "completion time and the backlog that may be accumulating.",
       .unit = "ms",
       .type = METRIC_TYPE_COUNTER,
   };
--- a/src/disk.c
+++ b/src/disk.c
@@ -75,6 +75,9 @@
 #include <sys/protosw.h>
 #endif
 
+static char const *const device_label = "system.device";
+static char const *const direction_label = "disk.io.direction";
+
 #if (MAC_OS_X_VERSION_MIN_REQUIRED < 120000) // Before macOS 12 Monterey
 #define IOMainPort IOMasterPort
 #endif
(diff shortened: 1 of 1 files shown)


src/disk.c
```diff
@@ -981,12 +1006,8 @@ static int disk_read(void) {
       metric_family_metric_append(&fam_disk_io_weighted_time, m);
 
       long interval = CDTIME_T_TO_MS(plugin_get_interval());
-      if (interval == 0) {
-        DEBUG("disk plugin: got zero plugin interval");
-      }
-
-      m.value.gauge = ((diff_io_time / (double)interval) * 100.0);
-      metric_family_metric_append(&fam_disk_utilization, m);
+      m.value.gauge = ((gauge_t)diff_io_time) / ((gauge_t)interval);
```
eero-t, line 1009: div-by-zero if interval is zero?

src/disk.c
```diff
@@ -1217,20 +1238,20 @@ static int disk_read(void) {
...
+    metric_family_append(&fam_ops, direction_label, read_direction,
                          (value_t){.counter = drives[i].rxfer}, &m);
-    metric_family_append(&fam_ops, "direction", "write",
+    metric_family_append(&fam_ops, direction_label, write_direction,
                          (value_t){.counter = drives[i].wxfer}, &m);
 
-    m.value.counter = drives[i].time_sec * 1000 + drives[i].time_usec / 1000;
-    metric_family_metric_append(&fam_disk_io_time, m);
+    m.value.counter = drives[i].time_sec + drives[i].time_usec / 1000000;
+    metric_family_metric_append(&fam_ops_time, m);
```
eero-t, line 1254: This metric family name change does not correspond with changes at start of the `disk_read()`.

eero-t (commented):
I think it would be better to handle https://github.com/collectd/collectd/pull/4242 in `main`, and merge all `main` changes to `disk.c` before this PR.

Reporting (some) time units as seconds (`s`) is IMHO not good while collectd counters are integers i.e. do not support subsecond timings.  Meaning that one does not see from counters very small usage spikes i.e. is disk completely idle or not.

Also, in general time manipulation for values appended to `fam_ops_time` is confusing as they can be multiplied or divided by different values, and variable names not having any indication of indication of their time units.

src/disk.c
octo: I think we can assume that the interval is non-zero. Otherwise I think a lot will break in the daemon. We divide by the interval somewhat frequently to calculate rates.

src/disk.c
```diff
@@ -382,54 +388,70 @@ static signed long long dict_get_value(CFDictionaryRef dict, const char *key) {
...
       .name = "system.disk.io",
+      .help = "Bytes read from and written to disk.",
+      .unit = "By",
       .type = METRIC_TYPE_COUNTER,
   };
   metric_family_t fam_ops = {
       .name = "system.disk.operations",
+      .help = "Read and write operations performed by the disk. If multiple "
+              "operations are merged into one by the disk, they will be "
+              "accounted for separately.",
```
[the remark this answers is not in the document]
octo: I like your wording. I have changed "disk" to "I/O subsystem" because I'm not sure if the kernel, the disk, or both merge operations.

src/disk.c
```diff
@@ -382,54 +388,70 @@ static signed long long dict_get_value(CFDictionaryRef dict, const char *key) {
...
       .type = METRIC_TYPE_COUNTER,
   };
   metric_family_t fam_disk_io_weighted_time = {
-      .name = "disk_io_weighted_time_total",
+      .name = "system.disk.weighted_io_time",
+      .help = "This metric is incremented at each I/O start, I/O completion, "
+              "or I/O merge by the number of I/Os in progress times the number "
+              "of milliseconds spent doing I/O since the last update of this "
+              "field.  This can provide an easy measure of both I/O completion "
+              "time and the backlog that may be accumulating.",
```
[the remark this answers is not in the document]
octo: Yeah, this quite a never-ending sentence. I copied it from the Kernel docs. I added some commas to make it easier to digest.

src/disk.c
octo: Not sure I understand you correctly. Did you mean that this is in seconds while other systems report in us? I have changed all implementations to report us.

octo:
Hey @eero-t, thanks for the review!

>     * I think it would be better to handle [disk: add /proc/diskstats fields 15 to 20 in KERNEL_LINUX #4242](https://github.com/collectd/collectd/pull/4242) in `main`, and merge related `main` changes to `disk.c` before this PR.

The v6 code is so different that I don't think resolving the resulting merge conflict is any easier than reimplementing the feature. Writing the metric families is more effort than plumbing the counters through.

>     * Reporting (some) time units as seconds (`s`) is IMHO not good while collectd counters are integers i.e. do not support subsecond timings.  Meaning that one does not see from counters very small usage spikes i.e. is disk completely idle or not.

That was a mistake, thanks for your diligence! I have changed all platforms to report microseconds, and changed the metric family unit to reflect that.

>     * Time manipulation for values appended to `fam_ops_time` is confusing as they can be multiplied or divided by different values, and variable names do not indicate their time units (like they IMHO should).

I have added appropriate suffixes to all variable and field names.

src/disk.c
eero-t: I mean that this change is:
* `fam_disk_io_time` -> `fam_ops_time`

Whereas other changes in the file were:
* `fam_disk_io_time` ->  `fam_disk_io_time`  (no fam name change)
* `fam_time` -> `fam_ops_time`

src/disk.c
```diff
@@ -1059,26 +1080,18 @@ static int disk_read(void) {
...
                          (value_t){.counter = kio.KIO_WOPS}, &m);
 
-    if (strncmp(ksp[i]->ks_class, "disk", 4) == 0) {
-      /* FIXME: Convert this to microseconds if necessary */
-      metric_family_append(&fam_time, "direction", "read",
-                           (value_t){.counter = kio.KIO_RTIME}, &m);
-      metric_family_append(&fam_time, "direction", "write",
-                           (value_t){.counter = kio.KIO_WTIME}, &m);
-    }
-
```
eero-t, line 1081: You commented in commit removing these Solaris metrics, them being run & wait, instead of read & write => maybe they should now use `fam_disk_io_time`?

Note: wait time is important disk metric.  It tells how much things are stalling because there's too much to write / disk is too slow.

eero-t (changes requested):
Value wrap-up calculations in `disk.c` are broken, see: `grep UINT_MAX disk.c`

They deduct 64-bit (signed) `derive_t` value from 32-bit (unsigned) `UINT_MAX` value.

Maybe collectd should have `DERIVE_T_MAX` define for this (both in v5 & v6 branches)?

src/disk.c
octo: I have added a few metrics for Solaris based on the kstat docs.

eero-t (approved):
Approved.  I'll file separate bug about  the UINT_MAX vs derive_t issue, as it's already in the v5 version of disk plugin, and not introduced by these changes.

octo:
> Value wrap-up calculations in `disk.c` are broken, see: `grep UINT_MAX disk.c`
> 
> They deduct 64-bit (signed) `derive_t` value from 32-bit (unsigned) `UINT_MAX` value.
> 
> Maybe collectd should have `DERIVE_T_MAX` define for this (both in v5 & v6 branches)?

That's an excellent callout, thank you! We actually have a function that is doing this right, `counter_diff()`. I ended up refactoring a lot of the Linux code; being one of the oldest pieces of code in collectd it was quite in need of a facelift.

After a lot of fiddling I figured out that we calculated the average time per operation … just for that variable never to be read. I think the existing metrics are much better at conveying useful information about the disk, so I remove the calculcation and made the plugin a good deal simpler.

We also unnecessarily stored a bunch of stuff in global state, which I have now removed.

eero-t:
Gah, read Linux diskstats documentation, and counters use _native_ unsigned long (see the new bug).

I.e. they are indeed 64-bit, but only on 64-bit host.  For 32-bit hosts, UINT_MAX was actually the correct maximum value.

eero-t:
Nice, `counter_diff()` already handles transparently unsigned 32-bit vs 64-bit wrap up => it will fix #4245 for v6.

The very best changes are ones that remove lots of code, and still manage retain the (relevant) functionality. :-)

On (very) quick check of the changes, they seemed OK, i.e. from my side this still seems fine for merging.

eero-t:
As to adding tests for `disk.c` (in another PR)...

I think that could be done for the Linux functionality by extracting that from `disk_read()` to another function, one that takes path to `/proc/diskstats` as an argument.   Test code can then ask it to read pre-prepared regular file(s) instead, to verify that reported metric updates are as expected, and to verify that invalid file content is handled gracefully.

(CPU plugin test coverage could be similarly extended to parsing of `/proc/stats` file.  At that point some helper functions for dealing with temporary test files could be useful.)

octo:
> As to adding tests for `disk.c` (in another PR)...

I think we need to separate I/O and parsing. Maybe a schema along the lines of:

```c
int parse(char const *data, metric_family_t families[]);

int read_callback(void) {
  char *buffer = read_text_file_contents();
  parse(buffer);
}
```

This will allow us to test `parse()` without using I/O.

We should also split plugins into multiple files, one for each operating system we support. E.g. `disk.c`, `disk_linux.c`, `disk_freebsd.c` etc. The current approach is hard to read and hard to maintain.

PR #4076, opened 2023-01-11
[collectd 6] src/memory.c: bring to feature parity with main branch

hnez:
ChangeLog: memory plugin: bring to feature parity with main branch

The `memory` plugin is already ported to the v6 API but some new features have been added in the `main` branch in the meantime.

This PR should bring the `memory` plugin in the collectd-6 branch to feature parity with the `main` branch by:

  - Replicating the feature introduced in #3916 (Report MemAvailable when present in meminfo)
  - Cherry-Picking #3880 (do not account reclaimable slab as used)
  - Cherry-Picking #3962 (add laundry and user wired pages)

PR #4219, opened 2023-12-29
[collectd 6] memory plugin: Align metrics with OpenTelemetry recommendations.

octo:
* The "state" label has been renamed to "system.memory.state".
* Metric descriptions and units have been added.

ChangeLog: Memory plugin: The metric schema has been aligned with OpenTelemetry semantic conventions.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -168,6 +168,8 @@ static int memory_config(oconfig_item_t *ci) /* {{{ */
 static int memory_dispatch(gauge_t values[COLLECTD_MEMORY_TYPE_MAX]) {
   metric_family_t fam_absolute = {
       .name = "system.memory.usage",
+      .help = "Reports memory in use by state",
+      .unit = "By",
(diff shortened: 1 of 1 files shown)


eero-t (approved):
Looks fine, but it maybe better to merge changes from main first, as v5 and v6 versions have diverged.

eero-t:
> Looks fine, but it maybe better to merge changes from main first, as v5 and v6 versions have diverged.

Ahem, Leonard already merged memory plugin changes from `main` year ago (in PR https://github.com/collectd/collectd/pull/4076) i.e. memory plugin is already up to date as last commit in `main` is from Feb 2022 => merging further changes to it is fine.

PS. @octo You already created table of changes between plugins in different branches.  Would it make sense to have separate ticket with that table, tracking plugins having changes that still need to be synced from `main` to v6, which would be updated updated whenever changes have been checked and synced?

PR #4223, opened 2024-01-03
[collectd 6] Use real collectd metric functions in gpu_sysman plugin tests

eero-t:
ChangeLog: Use real collectd metric functions in "gpu_sysman" plugin tests

There have been lot of changes in metric handling code, so it's better to use real metric functions rather than mocked ones.

This allows removing the mock functions, but makes it much easier for unrelated changes to break `gpu_sysman_test.c` compilation (e.g .utf-8 support already broke earlier version of this).

### DIFF
--- a/src/gpu_sysman_test.c
+++ b/src/gpu_sysman_test.c
@@ -1,7 +1,7 @@
 /**
  * collectd - src/gpu_sysman_test.c
  *
- * Copyright(c) 2020-2022 Intel Corporation. All rights reserved.
+ * Copyright(c) 2020-2024 Intel Corporation. All rights reserved.
  *
  * Licensed under the same terms and conditions as src/gpu_sysman.c.
  *
@@ -10,16 +10,27 @@
  *
  * Testing for gpu_sysman.c Sysman API usage and error handling.
  *
- * See:
+ * Sysman API:
  * - https://spec.oneapi.com/level-zero/latest/sysman/PROG.html
  * - https://spec.oneapi.io/level-zero/latest/sysman/api.html
  *
- * Building unit-tests:
- *   gcc -I. -Idaemon  -I/path/to/level-zero -O3 -g --coverage -Werror \
- *       -Wall -Wextra -Wpedantic -Wcast-align=strict -Wformat-security \
- *       gpu_sysman_test.c -o test_plugin_gpu_sysman
+ * Installing build dependencies:
+ * - Fedora: sudo dnf install oneapi-level-zero-devel
+ * - Debian: sudo apt install libze-dev
  *
- * Running unit-units:
+ * Normal build / testing:
+ * - build: make test_plugin_gpu_sysman
+ * - test: make test_plugin_gpu_sysman.log
+ *
+ * Direct builds for whole-program analysis, coverage etc (for configured
+ * sources, so that "config.h" + other generated includes are present):
+ *	gcc -I. -Idaemon -DHAVE_CONFIG_H -DBUILD_STANDALONE --coverage -Werror \
+ *	 -O3 -g -Wall -Wextra -Wpedantic -Wcast-align=strict -Wformat-security \
+ *	 -Wnull-dereference -Wstrict-overflow=2 -Warray-bounds=2 \
+ *	 -D_FORTIFY_SOURCE=2 -Wno-aggressive-loop-optimizations \
+ *	 gpu_sysman_test.c -o test_plugin_gpu_sysman
+ *
+ * Running unit-units directly:
  *	./test_plugin_gpu_sysman
  *
  * Testing for memory leakage:
@@ -37,10 +48,7 @@
  *   optimizations have significantly lower coverage due to each (trivial
  *   and build-time verifiable) switch-case being considered separately
  *
- *
  * Mock up functionality details:
- * - Allocation helpers assert instead of exiting on alloc failure,
- *   to get a backtrace
  * - All functions return only a single property or metric item,
  *   until hitting earlier set call limit, after which they return error
  * - All metric property functions report them coming from subdevice 0
@@ -64,9 +72,19 @@
  * - Plugin init, shutdown and re-init works without problems
  */
 
-#define KERNEL_LINUX 1
-#define FP_LAYOUT_NEED_NOTHING 1
 #include "gpu_sysman.c" /* test this */
+
+/* include metric functions + their dependencies directly, instead of
+ * building & linking libcommon.a (like normal collectd builds do)?
+ */
+#ifdef BUILD_STANDALONE
+/* utilities needed from collectd core */
+#include "daemon/metric.c"
+#include "utils/common/common.c"
+#include "utils/metadata/meta_data.c"
+#include "utils/strbuf/strbuf.c"
+#include "utils/utf8/utf8.c"
+#endif
 
 /* logging check bit, and per-phase logging bits enabling it */
 #define VERBOSE_CALLS 1
@@ -842,120 +860,6 @@ int plugin_dispatch_metric_family(metric_family_t const *fam) {
   return 0;
 }
 
-#define MAX_LABELS 16
-
-/* mock function uses just one large enough metrics array (for testing)
- * instead of increasing it one-by-one, like the real collectd metrics
- * code does
- */
-int metric_label_set(metric_t *m, char const *name, char const *value) {
-  assert(m && name);
-  size_t num = m->label.num;
-  label_pair_t *pair = m->label.ptr;
-  if (num) {
-    assert(num < MAX_LABELS);
-    assert(pair);
-  } else {
-    assert(!pair);
-    pair = scalloc(MAX_LABELS, sizeof(*pair));
-    m->label.ptr = pair;
-  }
-  int i;
-  for (i = 0; i < MAX_LABELS; i++) {
-    if (!pair[i].name) {
-      /* not found -> new label */
-      pair[i].name = sstrdup(name);
-      m->label.num++;
-      break;
-    }
-    if (strcmp(name, pair[i].name) == 0) {
-      break;
-    }
-  }
-  assert(value); /* removing label with NULL 'value' is not supported */
-  free(pair[i].value);
-  pair[i].value = sstrdup(value);
-  return 0;
-}
-
-int metric_reset(metric_t *m) {
-  assert(m);
-  size_t num = m->label.num;
-  label_pair_t *pair = m->label.ptr;
-  if (!num) {
-    assert(!pair);
-    return 0;
-  }
-  assert(pair);
-  for (int i = 0; i < MAX_LABELS; i++) {
-    if (!pair[i].name) {
-      break;
-    }
-    free(pair[i].name);
-    free(pair[i].value);
-    pair[i].value = pair[i].name = NULL;
-    num--;
-  }
-  assert(!num);
-  free(pair);
-  m->label.ptr = NULL;
-  m->label.num = 0;
-  return 0;
-}
-
-#define MAX_METRICS 8
-
-/* mock function uses just one large enough metrics array (for testing)
- * instead of increasing it one-by-one, like the real collectd metrics
- * code does
- */
-int metric_family_metric_append(metric_family_t *fam, metric_t m) {
-  assert(fam);
-  size_t num = fam->metric.num;
-  metric_t *metric = fam->metric.ptr;
-  if (num) {
-    assert(num < MAX_METRICS);
-    assert(metric);
-  } else {
-    assert(!metric);
-    metric = scalloc(MAX_METRICS, sizeof(*metric));
-    fam->metric.ptr = metric;
-  }
-  /* copy metric and pointers to its labels */
-  metric[num] = m;
-  label_pair_t *src = m.label.ptr;
-  if (src) {
-    /* alloc max size as labels can be added also to family metrics copies */
-    label_pair_t *dst = scalloc(MAX_LABELS, sizeof(*src));
-    metric[num].label.ptr = dst;
-    for (size_t i = 0; i < m.label.num; i++) {
-      dst[i].name = sstrdup(src[i].name);
-      dst[i].value = sstrdup(src[i].value);
-    }
-  }
-  fam->metric.num++;
-  m.family = fam;
-  return 0;
-}
-
-int metric_family_metric_reset(metric_family_t *fam) {
-  metric_t *metric = fam->metric.ptr;
-  for (size_t m = 0; m < fam->metric.num; m++) {
-    label_pair_t *pair = metric[m].label.ptr;
-    for (size_t i = 0; i < metric[m].label.num; i++) {
-      free(pair[i].name);
-      free(pair[i].value);
-    }
-    free(pair);
-    metric[m].label.ptr = NULL;
-    metric[m].label.num = 0;
-  }
-  free(fam->metric.ptr);
-  fam->metric.ptr = NULL;
-  fam->metric.num = 0;
-  return 0;
-}
-
 /* ------------------------------------------------------------------------- */
 /* mock up of collectd plugin API */
 
@@ -1038,26 +942,22 @@ void plugin_log(int level, const char *format, ...) {
   fprintf(stderr, "%s (%s)\n", msg, log_levels[level].name);
 }
 
-/* safe function wrapper from utils/common/common.c */
-char *sstrncpy(char *dest, const char *src, size_t n) {
-  strncpy(dest, src, n);
-  dest[n - 1] = '\0';
-  return dest;
+/* ------------------------------------------------------------------------- */
+/* Dummies for unused collectd core functions, needed for linking
+ * because "utils/common/common.c" refers both "daemon/utils_cache.c"
+ * and "daemon/plugin.c" functionality.
+ *
+ * Therefore "libcommon.la" needs these too for linking to work.
+ */
+
+void daemon_log(__attribute__((unused)) int level,
+                __attribute__((unused)) const char *format, ...) {
+  assert(0);
 }
-void *scalloc(size_t nmemb, size_t size) {
-  void *p = calloc(nmemb, size);
-  assert(p);
-  return p;
-}
-void *smalloc(size_t size) {
-  void *p = malloc(size);
-  assert(p);
-  return p;
-}
-char *sstrdup(const char *s1) {
-  char *s2 = strdup(s1);
-  assert(s2);
-  return s2;
+
+int uc_get_rate(__attribute__((unused)) metric_t const *m,
+                __attribute__((unused)) gauge_t *ret_value) {
+  return ENOTSUP;
 }
 
 /* ------------------------------------------------------------------------- */
--- a/Makefile.am
+++ b/Makefile.am
@@ -1200,7 +1200,7 @@ gpu_sysman_la_LIBADD = $(LEVEL_ZERO_LIBS)
 test_plugin_gpu_sysman_SOURCES = src/gpu_sysman_test.c
 test_plugin_gpu_sysman_CPPFLAGS = $(AM_CPPFLAGS) $(PLUGIN_CPPFLAGS) $(LEVEL_ZERO_CFLAGS)
 test_plugin_gpu_sysman_LDFLAGS = $(PLUGIN_LDFLAGS)
-test_plugin_gpu_sysman_LDADD =
+test_plugin_gpu_sysman_LDADD = libcommon.la
 check_PROGRAMS += test_plugin_gpu_sysman
 TESTS += test_plugin_gpu_sysman
 endif


eero-t:
@octo CI fails to unrelated error:
```
  Submodule 'gnulib' (git://git.savannah.gnu.org/gnulib.git) registered for path 'gnulib'
  Submodule 'opentelemetry-proto' (https://github.com/open-telemetry/opentelemetry-proto/) registered for path 'opentelemetry-proto'
  Cloning into '/__w/collectd/collectd/gnulib'...
  Cloning into '/__w/collectd/collectd/opentelemetry-proto'...
  Error: fatal: read error: Connection reset by peer
  Fetched in submodule path 'gnulib', but it did not contain 2f8140bc8ce5501e31dcc665b42b5df64f84c20c. Direct fetching of that commit failed.
  Error: The process '/usr/bin/git' failed with exit code 1
```

eero-t:
> @octo CI fails to unrelated error on "debian10" (although "debian11" passes):

Oddly, everything passed in https://github.com/collectd/collectd/pull/4177 that was rebased on top of this. Maybe that error was just transitory network error.

src/gpu_sysman_test.c
```diff
@@ -68,6 +68,25 @@
 #define FP_LAYOUT_NEED_NOTHING 1
 #include "gpu_sysman.c" /* test this */
 
+/* utilities needed from collectd core */
+#include <unistd.h>
+/* common.c fails to -Werror=unused-parameter, so
```
octo, line 73: Can you post the error you're getting? This sounds like it would be easy to fix.

src/gpu_sysman_test.c
```diff
@@ -68,6 +68,25 @@
...
+ * string helpers still need mocked versions.
+ */
+#if 0
+#include <dirent.h>
+#include <math.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+/* comment to avoid contrib/format.sh breaking code */
+#include "utils/common/common.c"
+#endif
```
octo, line 83: Is this left-over from development or the preferred solution should we be able to resolve the issues with `common.c`?

src/gpu_sysman_test.c
```diff
@@ -842,120 +861,6 @@ int plugin_dispatch_metric_family(metric_family_t const *fam) {
...
-    metric[m].label.num = 0;
-  }
-  free(fam->metric.ptr);
-  fam->metric.ptr = NULL;
-  fam->metric.num = 0;
-  return 0;
-}
-
 /* ------------------------------------------------------------------------- */
 /* mock up of collectd plugin API */
```
octo, line 865: Would it make sense to move this to `src/daemon/plugin_mock.c`?

octo (approved):

octo:
> > @octo CI fails to unrelated error on "debian10" (although "debian11" passes):
> 
> Oddly, everything passed in #4177 that was rebased on top of this. Maybe that error was just transitory network error.

git.savannah.gnu.org had transient errors all day. The "Actions" tab has a "re-run failed workflows" button that is quite useful in that situation.

src/gpu_sysman_test.c
eero-t: I added commit fixing includes in `common.c` & `strbuf.c`, and removed these from the test code.

src/gpu_sysman_test.c
eero-t: `getpwnam_r()` implementation in `common.c` does not use all of its parameters.  I worked around that by using `#define HAVE_GETPWNAM_R 1`.

After that, and fixing the includes, there are still following warnings:
```
gcc -Werror -Wall -Wextra -Wpedantic -Wcast-align=strict -Wformat-security -Wnull-dereference -Wstrict-overflow=4 -Warray-bounds=2 -Wno-aggressive-loop-optimizations -I. -Idaemon -O3 -g -D_FORTIFY_SOURCE=2 gpu_sysman_test.c -o gpu_sysman_test
In file included from gpu_sysman_test.c:74:
utils/common/common.c: In function ‘check_create_dir’:
utils/common/common.c:97:3: error: ‘__builtin_strncpy’ specified bound 256 equals destination size [-Werror=stringop-truncation]
   97 |   strncpy(dest, src, n);
      |   ^
utils/common/common.c:97:3: error: ‘__builtin_strncpy’ specified bound 256 equals destination size [-Werror=stringop-truncation]
utils/common/common.c: In function ‘notification_init’:
utils/common/common.c:97:3: error: ‘__builtin_strncpy’ specified bound 256 equals destination size [-Werror=stringop-truncation]
utils/common/common.c:97:3: error: ‘__builtin_strncpy’ specified bound 128 equals destination size [-Werror=stringop-truncation]
utils/common/common.c:97:3: error: ‘__builtin_strncpy’ specified bound 128 equals destination size [-Werror=stringop-truncation]
utils/common/common.c:97:3: error: ‘__builtin_strncpy’ specified bound 128 equals destination size [-Werror=stringop-truncation]
utils/common/common.c:97:3: error: ‘__builtin_strncpy’ specified bound 128 equals destination size [-Werror=stringop-truncation]
utils/common/common.c:97:3: error: ‘__builtin_strncpy’ specified bound 128 equals destination size [-Werror=stringop-truncation]
utils/common/common.c: In function ‘notification_init_metric’:
utils/common/common.c:97:3: error: ‘__builtin_strncpy’ specified bound 256 equals destination size [-Werror=stringop-truncation]
utils/common/common.c: In function ‘walk_directory’:
utils/common/common.c:1142:5: error: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Werror=strict-overflow]
 1142 | int walk_directory(const char *dir, dirwalk_callback_f callback,
      |     ^~~~~~~~~~~~~~
```

src/gpu_sysman_test.c
eero-t: I need the information passed to that API, to validate correct working of the plugin's config and metric handling. What is checked from metric and logging parts in the mocked functions is specific to sysman plugin, config options less so, but I think it makes sense to keep it all plugin specific.

octo: Sure, works for me.

octo (approved):
LGTM

src/gpu_sysman_test.c
octo: Which version of GCC is that? I have been unable to reproduce with GCC on Debian 12:

```
$ make V=1 src/utils/common/common.lo
depbase=`echo src/utils/common/common.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
/bin/bash ./libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I./src  -I./src -I./src/daemon -DPREFIX='"/tmp/collectd"' -DCONFIGFILE='"/tmp/collectd/etc/collectd.conf"' -DLOCALSTATEDIR='"/tmp/collectd/var"' -DPKGLOCALSTATEDIR='"/tmp/collectd/var/lib/collectd"' -DPLUGINDIR='"/tmp/collectd/lib/collectd"' -DPKGDATADIR='"/tmp/collectd/share/collectd"'   -Wall -Werror -Werror -Wall -Wextra -Wpedantic -Wcast-align=strict -Wformat-security -Wnull-dereference -Wstrict-overflow=4 -Warray-bounds=2 -Wno-aggressive-loop-optimizations -MT src/utils/common/common.lo -MD -MP -MF $depbase.Tpo -c -o src/utils/common/common.lo src/utils/common/common.c &&\
mv -f $depbase.Tpo $depbase.Plo
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I./src -I./src -I./src/daemon -DPREFIX=\"/tmp/collectd\" -DCONFIGFILE=\"/tmp/collectd/etc/collectd.conf\" -DLOCALSTATEDIR=\"/tmp/collectd/var\" -DPKGLOCALSTATEDIR=\"/tmp/collectd/var/lib/collectd\" -DPLUGINDIR=\"/tmp/collectd/lib/collectd\" -DPKGDATADIR=\"/tmp/collectd/share/collectd\" -Wall -Werror -Werror -Wall -Wextra -Wpedantic -Wcast-align=strict -Wformat-security -Wnull-dereference -Wstrict-overflow=4 -Warray-bounds=2 -Wno-aggressive-loop-optimizations -MT src/utils/common/common.lo -MD -MP -MF src/utils/common/.deps/common.Tpo -c src/utils/common/common.c  -fPIC -DPIC -o src/utils/common/.libs/common.o
$ gcc --version
gcc (Debian 12.2.0-14) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

octo: #4233 attempts to fix this.

eero-t: This was with the  default GCC version in Ubuntu 22.04 LTS:
`gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04)`

However, to get GCC to actually analyze the code more thoroughly, you also need to use high optimization levels (I used `-O3`, your command line does not seem to  specify any optimizations?).

And your command line was missing `-D_FORTIFY_SOURCE=2`: https://wiki.ubuntu.com/Security/Features#Built_with_Fortify_Source

eero-t:
Rebased to to latest `collectd-6.0`, fixed new/remaining GCC warnings in `collectd` core code, and ordered commits so that every one of them compiles with stricter compiler options.

(Sorry for the extra pushes, annoyingly my local and CI `clang-format` disagree on include order.)

----

@octo I had to downgrade one warning in my own tests due to this:
```
$ gcc -v
gcc version 13.2.1 20231011 (Red Hat 13.2.1-4) (GCC)

$ gcc -O2 -I src -I src/daemon -DFP_LAYOUT_NEED_NOTHING=1 -Werror -Wstrict-overflow=4 src/utils/common/common.c 
src/utils/common/common.c: In function ‘walk_directory’:
src/utils/common/common.c:1146:5: error: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Werror=strict-overflow]
 1146 | int walk_directory(const char *dir, dirwalk_callback_f callback,
      |     ^~~~~~~~~~~~~~
```
`-Wstrict-overflow` takes values from 1-5, and any value above 2 gives above warning when GCC optimization level is >1.

I do not see what GCC could be complaining about `common.c` though.

eero-t:
I hadn't understood how collectd code handles system includes (= nearly all code relies on them being included from collectd.h, but that happens only when HAVE_CONFIG_H is defined, i.e. configure has generated config.h etc).

=> Reverted my include addition commits, cleaned up those from test-code, and documented how tests should be run manually.

src/gpu_sysman_test.c
```diff
@@ -64,9 +69,15 @@
...
-#define KERNEL_LINUX 1
-#define FP_LAYOUT_NEED_NOTHING 1
-#include "gpu_sysman.c" /* test this */
+/* utilities needed from collectd core */
+#include "daemon/metric.c"
+#include "daemon/utils_cache_mock.c"
+#include "utils/common/common.c"
+#include "utils/metadata/meta_data.c"
+#include "utils/strbuf/strbuf.c"
+#include "utils/utf8/utf8.c"
```
octo, line 78: Why doesn't `gpu_sysman.c` include these?

src/gpu_sysman_test.c
```diff
@@ -10,14 +10,21 @@
...
+ * - Fedora: sudo dnf install oneapi-level-zero-devel
+ * - Debian: sudo apt install libze-dev
+ *
+ * Building unit-tests (after running "configure", so that
+ * "config.h" + other generated includes are present):
+ *	gcc -I. -Idaemon -DHAVE_CONFIG_H -O3 -g --coverage -Werror \
+ *	 -Wall -Wextra -Wpedantic -Wcast-align=strict -Wformat-security \
+ *	 -Wnull-dereference -Wstrict-overflow=2 -Warray-bounds=2 \
+ *	 -D_FORTIFY_SOURCE=2 -Wno-aggressive-loop-optimizations \
+ *	 gpu_sysman_test.c -o test_plugin_gpu_sysman
```
octo, line 27: Why do you need to document this?

The tests should automatically be built and run as part of `make check`. If you need to build this test individuall, run `make test_plugin_gpu_sysman`. If you want to run only this test, you can use `make test_plugin_gpu_sysman.log` (only really useful during development).

octo (approved):

src/gpu_sysman_test.c
eero-t: Because `gpu_sysman.c` is a collect plugin [1], whereas  `gpu_sysman_test.c` is a separate binary.

[1] i.e. collectd provides the symbols:
```
$ strip -s collectd
$ readelf -sW collectd | grep -v UND | awk '/FUNC *GLOBAL/{print $8}' | cut -b-7 | sort | uniq -c | sort -nr
     58 plugin_
     19 meta_da
     14 metric_
     13 uc_meta
     12 uc_get_
      9 strbuf_
...
```

The reasons why test binary includes all the necessary code directly, instead of linking to archives built during collectd build, are several:
* With all used code being visible to GCC, it can do proper whole-program analysis
* I'm testing the code in question with more strict options i.e. I will improve quality also in them
* I run these tests separately from the collectd build

src/gpu_sysman_test.c
eero-t: > Why do you need to document this?

* Comment documents everything that I'm testing with that file (used compiler options are just one of those items)
* It lists (stricter) compiler options that included collectd core files should IMHO pass (reason for the previous fix commits to them)

> The tests should automatically be built and run as part of `make check`. If you need to build this test individually, run `make test_plugin_gpu_sysman`. If you want to run only this test, you can use `make test_plugin_gpu_sysman.log` (only really useful during development).

Thanks!

Often I will (build and) run it separately though, because I want that to be done under Valgrind (maybe later compile tests with ASAN too), and I'm interested also in test coverage:
```
Optimized build coverage info:

File 'gpu_sysman_test.c'
Lines executed:86.50% of 511

File 'gpu_sysman.c'
Lines executed:87.46% of 1132

File 'daemon/metric.c'
Lines executed:27.10% of 369

File 'utils/strbuf/strbuf.c'
Lines executed:0.00% of 126

File 'utils/utf8/utf8.c'
Lines executed:71.43% of 7

File 'utils/metadata/meta_data.c'
Lines executed:1.29% of 309

File 'utils/common/common.c'
Lines executed:3.41% of 499

File 'daemon/utils_cache_mock.c'
Lines executed:0.00% of 19
```

(As you can see, some of the includes are needed just duo to another collectd core file referring symbols in them, nothing is actually used from them during Sysman plugin tests.)

src/gpu_sysman_test.c
octo: Oh! That is unexpected. I'd actually much prefer to consistently build tests by linking to modules that are not under test.

* Whole program analysis: Why is this needed? My understanding is that whole program analysis is primarily done to optimize the code, and I don't particularly care about the performance of unit tests.
* Stricter cflags: We can add an additional build to the CI system that builds everything with stricter flags. We can probably also run tests with ASAN instead of Valgrind.
* Alternative build method: I'm not keen on maintaining alternative, bespoke way of building parts of the tree. In other words, maintainability is more important to me than supporting such one-off use cases.

eero-t: > * Whole program analysis: Why is this needed? My understanding is that whole program analysis is primarily done to optimize the code, and I don't particularly care about the performance of unit tests.

Warnings analysis is part of compiler optimization passes.  If you do not enable optimizations, code checks can miss a lot of things.  And with contents of all functions visible to compiler, it can do better checks (along with those optimizations), for things actually used in the test.


> * Stricter cflags: We can add an additional build to the CI system that builds everything with stricter flags.

Enabling all the flags I'm interested about for whole collectd, would require a lot of fixes to rest of the code first.  Now I've needed to fix only things that are included directly to the tester.

(I can work on fixing those for collectd core, but I do not feel comfortable doing it for plugins that I'm not using, unless my changes are really trivial/obvious improvements.)

> We can probably also run tests with ASAN instead of Valgrind.

Yes, that would be really nice!


> * Alternative build method: I'm not keen on maintaining alternative, bespoke way of building parts of the tree. 

Sorry, I forgot to mention most important reason of including just the needed C-files instead of using the archives.

To do the checks requires by tests, I need to intercept (have special versions of) few collectd functions:
* metric submit: what metrics were reported or not reported, and with what exact values?
* logging: were warnings/errors generated as expected, or unexpectedly?
* config: were config option parsing results expected?

Unless you have a good suggestion on how to get access to submitted metrics and logs when using real collectd functions (without needing to implement both write & log plugins functionality), I think using build-internal archives would mess the tester.

> In other words, maintainability is more important to me than supporting such one-off use cases.

Yes that's understandable.

Skipping this PR is one option (for avoiding direct inclusion potentially causing later test breakages with unrelated collectd changes).

I would then need to update the mocked metric functions (removed in this PR) instead though.

octo: I thought about this a bit, because I believe that we need to improve our test coverage, particularly for frequently used plugins. My rough idea so far was:

* (Optional) Rename `plugin_mock.c` to `plugin_fake.c`
* In `src/daemon/plugin_mock.c` write a test-only implementation of `plugin_dispatch_metric_family()` that clones all metric families and stores the copy somewhere.
* Add functions for retrieving and resetting the stored metric families. Make these functions accessible to tests via `src/daemon/plugin_mock.h`.

This should allow (read) plugins to actually call the dispatch function successfully. Once the read function returns the test can inspect what metric families were dispatched, if any.

Possible alternatives:

* Implement a fake `plugin_register_write()` so that the fake `plugin_dispatch_metric_family()` call write callbacks without all the thread start-up required by the real `plugin.c`. Then provide a "write plugin for testing" implementation with the "copy and store in memory" behavior outlines above.
* Provide a test harness to do the above, but with the "real" `plugin.c` implementation. This would probably give the best test coverage, but is also the most complex to pull off.

We already have a fake logging implementation, we'd just need to also store log messages to make them available to the tests. This should be rather easy.

If the plugin is written so that it doesn't so any I/O in the config callback, the config parsing should be easy to test. 
#4238 is doing it, although with a temporary file for now. If we add the ability to parse a string to the config file code, then we should be able to test this without I/O also.

eero-t: **Config**

Sysman plugin test code calls registered config function for each key.   If real plugin code is used, it could call plugin config function directly instead.

**Logging**

Logs should not be much of a problem either.  The main thing (besides logs being printed for debugging) is able to get counts of things logged above given level, and being able to reset those counts.  With that, test can verify that failing things fail (with log output), and ones that should not fail, don't do that.

**Metrics**

This is the issue. Propsed thing will make debugging metric issues much harder.   My own fake code uses asserts on all possible errors, instead of just logging something, or exiting, like collectd code does.  Sysman plugin tester does a *lot* of things and backtraces from asserts right in the submit path, give a huge help for tracking down issues.  Backtraces do not help if metrics are collected first, and only then verified.

=> Better would be to have a test callback for metric submit

eero-t:
Current version is failing because `config.h` apparently gets configured with HAVE_CAPABILITY in some CI targets, and therefore:
```
/__w/collectd/collectd/src/utils/common/common.c:1485: undefined reference to `cap_get_bound'
/usr/bin/ld: /__w/collectd/collectd/src/utils/common/common.c:1488: undefined reference to `cap_get_proc'
/usr/bin/ld: /__w/collectd/collectd/src/utils/common/common.c:1493: undefined reference to `cap_get_flag'
/usr/bin/ld: /__w/collectd/collectd/src/utils/common/common.c:1498: undefined reference to `cap_free'
/usr/bin/ld: /__w/collectd/collectd/src/utils/common/common.c:1495: undefined reference to `cap_free'
```

=> Trying to include real `*.c` files directly is too much whack-a-mole game.

I guess I'll next need to try how badly testing breaks if tester is instead linked with `libplugin_mock.la`, and calls Sysman plugin's plugin-API functions directly.

Looking at `Makefile.am`, there does not seem to be any archive that I could use to link all relevant real functions, as closest to that seemed to be what `collectd` daemon itself links.

octo:
> Looking at `Makefile.am`, there does not seem to be any archive that I could use to link all relevant real functions, as closest to that seemed to be what `collectd` daemon itself links.

That is was libplugin_mock.la needs to become IMHO – a drop-in replacement for the daemon to test plugins. That's for example why I recently threw out the mocked/faked "cache" implementation and linked with the real one instead. By linking with the "real" implementation tests can now cover plugins that require the cache to function.

src/gpu_sysman_test.c
eero-t: I could avoid those includes by linking to `libcommon.la` and adding couple of dummy functions (for things that requires, but does not link).

Those includes are still in the code, but behind `BUILD_STANDALONE` ifdef, so that my own testing can do whole program analysis.

src/gpu_sysman_test.c
eero-t: Documented this a bit better in the comment.

eero-t:
@octo This works OK again, and your feedback should be handled.

As to my removal of (apparently redundant) `-lm` lines from `Makefile.am`, should I split that to separate PR?

eero-t:
Dropped the extra `Makefile.am` change, rebased to latest & squashed commits to a single one.  Will merge this later today (based on the earlier approval,  unless there are comments/objections/CI-check fails).

PR #4224, opened 2024-01-04
[collectd 6] Memory plugin: Add "shared", drop "slab" and "available", report utilization as ratio.

octo:
Another round of improvements for the *memory plugin*:

* Report the `system.memory.utilization` metric as a fraction. Previously, the metric family's unit was `1` but the reported values were `%`. This fixes the discrepancy by changing how values are reported.

    Side note: We could report a percentage, but there is no "official" OpenTelemetry guidance for it. If we want to do this (I'm not convinced that we should), we need to add a config option (allowing users to explicitly opt-in to this behavior) and create a new metric family, maybe `system.memory.utilization_percent`.
* Create two new config options, `ReportUsage` and `ReportUtilization`. The previous config options have been kept for compatibility.
* Utilization reporting has been enabled by default, following the OT semantic convention.
* The `system.memory.limit` metric has been added. As per semantic convention, reporting of this metric is guarded by a config option and disabled by default.
* The "slab" and "available" states have been removed from the Linux implementation. The reported values are already contained in other states, resulting in double reporting of some bytes. This leads to the memory total fluctuating over time.
* The "shared" state has been added for Linux. This is in line with the OT semantic conventions and the *free(1)* command line utility.

ChangeLog: Memory plugin: The `system.memory.limit` metric has been added. The "slab" and "available" states have been removed, the "shared" state was added (Linux only). Utilization reporting has been changed to report a ratio (fraction of 1) rather than a percentage.

### DIFF
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -5269,7 +5269,7 @@ The I<memory plugin> provides the following configuration options:
 
 =over 4
 
-=item B<ValuesAbsolute> B<true>|B<false>
+=item B<ReportUsage> B<true>|B<false>
 
 Enables or disables reporting of physical memory usage in absolute numbers,
 i.e. bytes. Defaults to B<true>.
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -5282,6 +5282,11 @@ memory, e.g. the fraction of physical memory used. Defaults to B<true>.
 This is useful for deploying I<collectd> in a heterogeneous environment in
 which the sizes of physical memory vary.
 
+=item B<ReportLimit> B<false>|B<true>
+
+Controls reporting of the total amount of physical memory available to the
+system. Defaults to B<true>.
+
 =back
 
 =head2 Plugin C<modbus>
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -5274,10 +5274,10 @@ The I<memory plugin> provides the following configuration options:
 Enables or disables reporting of physical memory usage in absolute numbers,
 i.e. bytes. Defaults to B<true>.
 
-=item B<ValuesPercentage> B<false>|B<true>
+=item B<ReportUtilization> B<false>|B<true>
 
-Enables or disables reporting of physical memory usage in percentages, e.g.
-percent of physical memory used. Defaults to B<false>.
+Enables or disables reporting of physical memory usage as a ratio of total
+memory, e.g. the fraction of physical memory used. Defaults to B<false>.
 
 This is useful for deploying I<collectd> in a heterogeneous environment in
 which the sizes of physical memory vary.
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -5274,10 +5274,10 @@ The I<memory plugin> provides the following configuration options:
 Enables or disables reporting of physical memory usage in absolute numbers,
 i.e. bytes. Defaults to B<true>.
 
-=item B<ReportUtilization> B<false>|B<true>
+=item B<ReportUtilization> B<true>|B<false>
 
 Enables or disables reporting of physical memory usage as a ratio of total
-memory, e.g. the fraction of physical memory used. Defaults to B<false>.
+memory, e.g. the fraction of physical memory used. Defaults to B<true>.
 
 This is useful for deploying I<collectd> in a heterogeneous environment in
 which the sizes of physical memory vary.
--- a/src/memory.c
+++ b/src/memory.c
@@ -67,6 +67,7 @@ static char const *const label_state = "system.memory.state";
 typedef enum {
   STATE_USED,
   STATE_FREE,
+  STATE_SHARED,
   STATE_BUFFERS,
   STATE_CACHED,
   STATE_WIRED,
@@ -82,9 +83,9 @@ typedef enum {
 } memory_type_t;
 
 static char const *memory_type_names[STATE_MAX] = {
-    "used",     "free",      "buffers", "cached", "wired",
-    "active",   "inactive",  "kernel",  "locked", "arc",
-    "unusable", "user_wire", "laundry",
+    "used",  "free",     "shared",    "buffers", "cached",
+    "wired", "active",   "inactive",  "kernel",  "locked",
+    "arc",   "unusable", "user_wire", "laundry",
 };
 
 /* vm_statistics_data_t */
@@ -444,6 +445,9 @@ static int memory_read_internal(gauge_t values[STATE_MAX]) {
       mem_not_used += v;
     } else if (strcmp(fields[0], "Cached:") == 0) {
       values[STATE_CACHED] = v;
+      mem_not_used += v;
+    } else if (strcmp(fields[0], "Shmem:") == 0) {
+      values[STATE_SHARED] = v;
       mem_not_used += v;
     }
   }
--- a/src/memory.c
+++ b/src/memory.c
@@ -149,15 +149,16 @@ static int pagesize;
 #include <uvm/uvm_extern.h>
 #endif
 
-static bool values_absolute = true;
+static bool report_usage = true;
 static bool values_percentage;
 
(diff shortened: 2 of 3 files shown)


src/memory.c
```diff
@@ -65,44 +65,27 @@
...
-    "locked",
-    "arc",
-    "unusable",
-    "available",
-    "user_wire",
-    "laundry",
+static char const *memory_type_names[STATE_MAX] = {
+    "used",  "free",     "shared",    "buffers", "cached",
+    "wired", "active",   "inactive",  "kernel",  "locked",
+    "arc",   "unusable", "user_wire", "laundry",
```
eero-t, line 88: IMHO this would be easier to read / check against each other if they would be sorted alphabetically.

src/memory.c
```diff
@@ -149,87 +132,112 @@ static int pagesize;
...
+    };
+    metric_family_metric_append(&fam_limit, m);
+
+    int status = plugin_dispatch_metric_family(&fam_limit);
+    if (status != 0) {
+      ERROR("memory plugin: plugin_dispatch_metric_family failed: %s",
+            STRERROR(status));
+    }
+    ret = ret ? ret : status;
+  }
```
eero-t, line 213: Installed memory amount is more like cluster node property, rather than its metric, because it cannot change (without taking the node down).  If usage + utilization ratio are not enough, report free memory instead (user can then derive total from usage+free, if s/he wants).

eero-t (commented):
Code changes in general look OK, but I'm not so sure about providing total (instead of e.g. free).

src/memory.c
```diff
@@ -149,87 +132,112 @@ static int pagesize;
...
     oconfig_item_t *child = ci->children + i;
-    if (strcasecmp("ValuesAbsolute", child->key) == 0)
-      cf_util_get_boolean(child, &values_absolute);
-    else if (strcasecmp("ValuesPercentage", child->key) == 0)
-      cf_util_get_boolean(child, &values_percentage);
+    if (strcasecmp("ReportUsage", child->key) == 0 ||
+        strcasecmp("ValuesAbsolute", child->key) == 0)
+      cf_util_get_boolean(child, &report_usage);
+    else if (strcasecmp("ReportUtilization", child->key) == 0 ||
+             strcasecmp("ValuesPercentage", child->key) == 0)
```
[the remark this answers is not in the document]
octo: Done

src/memory.c
octo: Sound good; done.

src/memory.c
octo: I think the target audience here are Cloud workloads, e.g. containers. The runtime may do automatic resizing and the question "how much memory is allocated to my service" is both interesting and not static.

For physical machines and VMs this is not very interesting, and so I think the default of disabling the reporting of this metric is correct.

eero-t: > I think the target audience here are Cloud workloads, e.g. containers. The runtime may do automatic resizing and the question "how much memory is allocated to my service" is both interesting and not static.

Sure, but that's not what the plugin provides.

Containers' memory usage is limited by memory cgroup: https://docs.kernel.org/admin-guide/cgroup-v2.html#memory

Whereas `/proc/meminfo` read by the memory plugin lists the whole system memory statistics, which total does **not** change.

octo: That's a good point. I also agree that this is not really a "metric" – I have argued against the *uptime plugin* along similar lines. So if you feel strongly about this, we can revert 413112fe93888e13eb8b4c11fc09ecd1159e710f.

On the other hand, many people don't have an inventory system alongside their metrics, and fewer yet can merge data between the two. I think there are situations where having the number of CPUs, the amount of memory, etc. available *as a metric* can be useful for some people. With the metric being disabled by default, I don't think we negatively impact users that don't need it. So if you don't feel strongly about this, I'd say let's keep the metric, it doesn't impact the user experience unless activated.

src/collectd.conf.pod
```diff
@@ -5269,19 +5269,24 @@ The I<memory plugin> provides the following configuration options:
...
+Enables or disables reporting of physical memory usage as a ratio of total
+memory, e.g. the fraction of physical memory used. Defaults to B<true>.
 
 This is useful for deploying I<collectd> in a heterogeneous environment in
 which the sizes of physical memory vary.
 
+=item B<ReportLimit> B<false>|B<true>
+
+Controls reporting of the total amount of physical memory available to the
+system. Defaults to B<true>.
```
eero-t, line 5288: `conf.in` says this is `false`.

src/memory.c
eero-t: Good point about inventory.  Do not feel strongly about this / it's OK being disabled by default.

src/memory.c
```diff
@@ -149,87 +145,114 @@ static int pagesize;
...
+        .value = (value_t){.derive = (derive_t)total},
+    };
+    metric_family_metric_append(&fam_limit, m);
+
+    int status = plugin_dispatch_metric_family(&fam_limit);
+    if (status != 0) {
+      ERROR("memory plugin: plugin_dispatch_metric_family failed: %s",
+            STRERROR(status));
+    }
+    ret = ret ? ret : status;
```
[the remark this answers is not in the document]
octo: Done.

octo:
Thanks for the review, @eero-t!

octo:
> Please remember to update both metric name changes & slab metric removal to v6 metric changes Wiki page.

Done

eero-t:
@octo after https://github.com/collectd/collectd/pull/3916#issuecomment-1894085292 I checked how this plugin calculates ratio...  Code sums all values together at dispatch and uses that as a total, which is just wrong.

When fetching memory info from the OS, It should store total separately, and use that when calculating ratios, anything else is bogus.

Also, you ignored my comment about `ReportLimit` default in doc mismatching what's set in `conf.in`?

octo:
> @octo after [#3916 (comment)](https://github.com/collectd/collectd/pull/3916#issuecomment-1894085292) I checked how this plugin calculates ratio... Code sums all values together at dispatch and uses that as a total, which is just wrong.

Systems typically report total memory and free memory, but not memory that is in use. We calcualte "used" as "total - free" in most (all?) cases. For Linux, this is done [here](https://github.com/collectd/collectd/blob/main/src/memory.c#L391):

```c
    mem_used =
        mem_total - (mem_free + mem_buffered + mem_cached + mem_slab_total);
```

That means that `free + buffered + cached + slab + used` will equal `total`. The problem is/was that "slab" overlaps with "cached" (I think, might be "buffered") and so we under-report "used".

> Also, you ignored my comment about `ReportLimit` default in doc mismatching what's set in `conf.in`?

Oh, sorry about that. I'll fix that soon.

eero-t:
> ```c
>     mem_used =
>         mem_total - (mem_free + mem_buffered + mem_cached + mem_slab_total);
> ```

That's wrong in two ways.  From `man proc`:
```
              SReclaimable %lu (since Linux 2.6.19)
                     Part of Slab, that might be reclaimed, such as caches.
```

I.e. only part of slab memory is reclaimable, and part of that overlaps with cached/buffers which you're already counting.

> That means that `free + buffered + cached + slab + used` will equal `total`.

That's wrong, besides slab overlapping buffered/cached, it misses e.g. all the memory that kernel does not even report to user-space, which can both vary, and be a very large amount.   Just report what kernel directly tells to be the total!

=> I think plugin should just do same as what `man free` says:
```
       total  Total installed memory (MemTotal and SwapTotal in /proc/meminfo)
       used   Used memory (calculated as total - free - buffers - cache)
       free   Unused memory (MemFree and SwapFree in /proc/meminfo)
```
But please report swap separately, as disk is not RAM.


From initial PR description:
> The "slab" and "available" states have been removed from the Linux implementation. The reported values are already contained in other states, resulting in double reporting of some bytes. This leads to the memory total fluctuating over time.

I hadn't realized that "available" was removed because of this bogus way of calculating total (instead of just reporting the kernel provided correct value).  Reporting "available" instead of calculated "free", would be another alternative...

PR #4227, opened 2024-01-10
[collectd 6, draft] Add the `VALUE_T()` generic macro.

octo:
`VALUE_T(x)` is a generic macro that initializes and returns a `value_t`. Depending on the type of the argument `x`, it will either initialize the `value_t` as a `counter_t`, `derive_t`, or `gauge_t`.

This means that the previous expression:

```c
value_t x = (value_t){.derive = y};
```

can now be written as:

```c
value_t x = VALUE_T(y);
```

Passing in an incompatible type, such as an `int`, will cause a compile time error:

```
src/cpu.c: In function 'module_register':
./src/daemon/metric.h:56:12: error: '_Generic' selector of type 'int' is not compatible with any association
   56 |   _Generic((x),                                                                \
      |            ^
src/cpu.c:929:17: note: in expansion of macro 'VALUE_T'
  929 |   value_t foo = VALUE_T((int){42});
      |                 ^~~~~~~
```

**Draft:** I want to gauge how widely supported this C11 feature is.

### DIFF
--- a/src/cpu.c
+++ b/src/cpu.c
@@ -366,8 +366,7 @@ static cpu_state_t *get_cpu_state(size_t cpu_num, size_t state) /* {{{ */
 static int total_rate(gauge_t *sum_by_state, size_t state, derive_t d,
                       value_to_rate_state_t *conv, cdtime_t now) {
   gauge_t rate = NAN;
-  int status =
-      value_to_rate(&rate, (value_t){.derive = d}, DS_TYPE_DERIVE, now, conv);
+  int status = value_to_rate(&rate, VALUE_T(d), DS_TYPE_DERIVE, now, conv);
   if (status != 0)
     return status;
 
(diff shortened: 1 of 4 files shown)


octo:
Solaris (`/opt/solarisstudio12.4/bin/cc`) apparently does not implement type-generic macros.

eero-t:
> Solaris (`/opt/solarisstudio12.4/bin/cc`) apparently does not implement type-generic macros.

So what's the plan with this PR, give also type as arg for now?  Give type as another arg, or give up as value of that macro was automatic inferencing of the type to avoid mistakes with it?

octo:
I'm undecided. I think the macro makes a common use case nicer and safer. It may be reasonable to break compatibility with an ancient compiler for this.

However I don't know how commonly the platform is still used and whether any alternative compilers are available. With this lack of knowledge is hard to make an informed decision.

PR #4228, opened 2024-01-11
[collectd 6] interface plugin: Align metrics with OpenTelemetry recommendations.

octo:
https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#network-metrics

ChangeLog: Interface plugin: The metric schema has been aligned with OpenTelemetry semantic conventions.

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -293,12 +293,12 @@ static int if_read_internal(metric_family_t fams[static FAM_MAX]) {
       value_t value;
     } metrics[] = {
         // clang-format off
-        {&fam[FAM_IO],      rx_direction, VALUE_T(if_data->IFA_RX_BYTES)},
-        {&fam[FAM_PACKETS], rx_direction, VALUE_T(if_data->IFA_RX_PACKT)},
-        {&fam[FAM_ERRORS],  rx_direction, VALUE_T(if_data->IFA_RX_ERROR)},
-        {&fam[FAM_IO],      tx_direction, VALUE_T(if_data->IFA_TX_BYTES)},
-        {&fam[FAM_PACKETS], tx_direction, VALUE_T(if_data->IFA_TX_PACKT)},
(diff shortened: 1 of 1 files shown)


src/interface.c
```diff
@@ -97,47 +97,19 @@ static size_t numif;
...
-    .name = "node_network_transmit_errs_total",
-    .help = "Network device statistic transmit_errs.",
-    .type = METRIC_TYPE_COUNTER,
-};
-metric_family_t transmit_packets = {
-    .name = "node_network_transmit_packets_total",
-    .help = "Network device statistic transmit_packets.",
-    .type = METRIC_TYPE_COUNTER,
-};
+static char const *const direction_label = "network.io.direction";
```
eero-t, line 108: Could have link to spec as comment.

eero-t (commented):
Code looks OK, but IMHO there should be some migration doc where all these metric name/label changes from v5 are documented, at the same time they are changed in the plugins.

I don't think anybody's going to go through all the plugins to document it, and catch all the changes afterwards, at release time.

But such documentation is absolutely necessary because users are going to need to update metric names in all their collectd integrations (Prometheus monitors, Grafana dashboards, SQL injection scripts etc)!

src/interface.c
octo: Done.

octo:
> there should be some migration doc where all these metric name/label changes from v5 are documented, at the same time they are changed in the plugins.

I started to document this here: https://github.com/collectd/collectd/wiki/collectd-6

eero-t:
> I started to document this here: https://github.com/collectd/collectd/wiki/collectd-6

Great!    (Noticed typo there:  /ReportUtilizatoin//ReportUtilization/)

eero-t (approved):
On quick review, looks OK.

octo:
> Noticed typo there: /ReportUtilizatoin/ReportUtilization/

Thanks for flagging this! It's a wiki. You should also have access to fix it.

PR #4230, opened 2024-01-11
Import changes to `.github/workflows/` from the `collectd-6.0` branch.

octo:


.cirrus.yml
```diff
@@ -49,7 +55,12 @@ debian_default_toolchain_task:
     - make -j$(nproc) -sk
   tests_script:
     - VALGRIND_OPTS="--errors-for-leak-kinds=definite" make -j$(nproc) -sk check || (cat ./test-suite.log && false)
-    - /checks/check-built-plugins.sh
+    - |
+      if [[ -e ".expected-plugins/$CONTAINER_IMAGE" ]]; then
+        /checks/check-built-plugins.sh ".expected-plugins/$CONTAINER_IMAGE"
+      else
+        /checks/check-built-plugins.sh
+      fi
```
eero-t, line 63: Would it make sense for this logic to be instead part of `check-built-plugins.sh`?
```suggestion
    - /checks/check-built-plugins.sh ".expected-plugins/$CONTAINER_IMAGE"
```

.cirrus.yml
```diff
@@ -77,7 +94,12 @@ redhat_default_toolchain_task:
...
```
eero-t, line 102: ditto.

.github/workflows/build.yml
```diff
@@ -76,9 +75,14 @@ jobs:
         path: "*.log"
     - name: Report check status
       run: exit $(< make-check.status)
+    - name: Check expected plugins
+      run: |
+        if [[ -e ".expected-plugins/${{ matrix.container_tag }}" ]]; then
+          /checks/check-built-plugins.sh ".expected-plugins/${{ matrix.container_tag }}"
+        fi
```
eero-t, line 82: For some containers, no plugin checks are done?

.github/workflows/build.yml
```diff
@@ -130,3 +134,8 @@ jobs:
...
```
eero-t, line 141: For some containers, no plugin checks are done?

.cirrus.yml
octo: This is backwards compatibility behavior. If called without an argument, `check-built-plugins.sh` will fall back on the environment variable set in the Docker containers. Once the transition to a file based plugin list is complete, we can get rid of this.

.github/workflows/build.yml
octo: `.expected-plugins/` does not yet exist in the `main` branch. If you want, we can adopt the CirrusCI behavior for GitHub Actions, but I'd like to make that change in `collectd-6.0` first and then cherry-pick to here, so that the PR remains a "pure" import (i.e. no novel changes).

.github/workflows/build.yml
octo: See above.

.cirrus.yml
octo: See above.

.github/workflows/distcheck.yml
```diff
@@ -16,7 +16,7 @@ defaults:
 jobs:
   make_distcheck:
     runs-on: ubuntu-latest
-    container: collectd/ci:bullseye_amd64
+    container: collectd/ci:debian11
```
eero-t, line 19: In v6, `ubuntu-latest` is `debian12`?

eero-t (commented):
Other than the distcheck change, this corresponds to v6 files.

eero-t (approved):

PR #4231, opened 2024-01-11
[collectd 6] .gitmodules: Mark `gnulib` as inactive.

octo:
This submodule is only needed to build on Windows, and it is causing issues for our automatic builders because the Git repository has many transient errors.

This also instructs Git to use "shallow" clones, because we don't need history in these submodules.


eero-t (approved):
Approved.  Only thing I'm wondering about is `--add safe.directory "$(pwd)"`, otherwise this looks fine.

octo:
> Approved. Only thing I'm wondering about is `--add safe.directory "$(pwd)"`, otherwise this looks fine.

git is unhappy without it. I copied this from the `git checkout` code used by the upstream GitHub action. You can see it in the checkout output.

PR #4232, opened 2024-01-11
write_riemann plugin: Terminate `riemann_event_set` arguments with `RIEMANN_EVENT_FIELD_NONE`.

octo:
`riemann_event_set` is a variadic function, that means it accepts a variable number of arguments. That means it needs some way to determine – at runtime – how many arguments there are. It appears to be doing so by using `RIEMANN_EVENT_FIELD_NONE` to indicate the last element in the argument list. Unfortunately I was unable to find the library's documentation and code and could not verify this.

That means that the argument list passed to `riemann_event_set` was not always terminated, causing it to read past where it was supposed to and adding random crap into the message it crafted.

ChangeLog: Write Riemann plugin: An out of bounds memory read, resulting in garbled messages, was fixed.
Issue: #4050

### DIFF
--- a/src/write_riemann.c
+++ b/src/write_riemann.c
@@ -214,7 +214,7 @@ static riemann_message_t *wrr_notification_to_message(notification_t const *n) {
 
 #if RCC_VERSION_NUMBER >= 0x010A00
   riemann_event_set(event, RIEMANN_EVENT_FIELD_TIME_MICROS,
-                    (int64_t)CDTIME_T_TO_US(n->time));
+                    (int64_t)CDTIME_T_TO_US(n->time), RIEMANN_EVENT_FIELD_NONE);
 #endif
 
   if (n->host[0] != 0)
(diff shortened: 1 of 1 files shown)


octo:
Found the relevant code: https://git.madhouse-project.org/algernon/riemann-c-client/src/tag/riemann-c-client-1.10.4/lib/riemann/event.c#L60

The function is sane (apart from using variadic arguments, that is), returning an error when a field type is unknown (which collectd ignores 🙈). This is probably why it takes a while before the bug causes a crash.

eero-t (approved):
Approved.

PS. Because finding the correct riemann code is a bit hard, some comment in the plugin code listing which riemann library functions are variadic (along with names of terminators for them) could be good.

octo:
Cherry-picked:

* `collectd-5.12`: cfdf526f1b4867d8952fc6d37b85f0d426a78fc7
* `collectd-6.0`: f6036685694b23032f53be8a37d26c7b33e670b3

Issue #4222, opened 2024-01-03
[collectd 6] RFC: Automatic release process


octo:
I think for *collectd 6* we should move away from large and manual releases and towards frequent automatic releases, ideally full CI/CD.

Proposed workflow:

* Trigger the release pipeline on a regular basis (daily? weekly?).
* Check whether there have been any changes since the last run and exit if not.
* Extract the new pull requests and from them the `ChangeLog:` lines.
    * Future improvement: Add changes to the `ChangeLog` file and commit.
* Add a release tag (this is required for creating a "release" next)
* Create a new draft "release" on GitHub with the extracted ChangeLog lines.
* Set `PACKAGE_VERSION` to something sane, such as `6.2024.01.03`
* Run `make dist` to create a `.tar.bz2` file.
* Attach the tar archive to the release and mark the release as published.
* Trigger a workflow in the `collectd.github.io` repo to reference the new release from the website.

What do you think about frequent, automatic releases? What are your thoughts on dropping the differentiation between feature releases and patch releases?

eero-t:
> What do you think about frequent, automatic releases?

Actual releases normally have a manually curated list of major changes, and public pre-releases for extra testing.

**Release testing**

Because collectd has so many plugins, of which many require specific setups to test & generate data, and some expertise to validate that their data still makes sense, IMHO pre-releases do not work so well for collectd (now that many plugins are lacking active maintainers).  Getting info on issues with rarely used plugins could take years, as users may try them only after upgrading their enterprise distro...

From that point,  doing more frequent releases that explicitly state what is actually tested, and what is not tested, sounds OK.

=> Listing tests coverage information for the automated releases would be great, but at least which plugins do not have tests, which have tests, and which of those tests were ran/passing should be listed.  This information needs to be per-distro, as tested plugins differ between them.

I think this info would look good in table format, with plugins being listed in row headings, distros in column headings and MISSING / SKIP / PASS info in the table cells. 

(Hopefully that will also encourage people to submit PRs for improving test coverage of the plugins they use or maintain.)

**Release change notes**

Typically people do not update their SW to completely new versions daily or even weekly.  Not even rolling distros, I think.

While there can be separate automated tagging which is done more frequently (to support testing & discussing issues), and which do not need change notes, I think it makes sense to have automated releases more rarely, e.g. monthly, to "encourage" distros and individuals to converge on testing same releases.  That way it's easier for them to compare their results.

[...]

PR #4079, opened 2023-01-16
[collectd 6] Misc "gpu_sysman" plugin fixes & improvements

eero-t:
ChangeLog: Misc "gpu_sysman" plugin fixes & improvements

Changes:
* Fix memory corruption / crashes with latest L0 backend versions
  * Resulting from backend implementation API change to a previously under-specified behavior, that is documented now in L0 spec
* Improve error logging by reporting the related Sysman API error code
   * This is by far the largest change, but helps debugging metric reporting errors a lot
* Avoid (unlikely) issues of:
   * Reporting incorrect metric labels
   * Disabling power ratio reporting unnecessarily
* Add support for memory "health", frequency "throttled_by", and metric "pci_dev" labels
  * First 2 are especially useful for GPU issue monitoring / alerts on servers
* Minor code consistency improvements
* Slight increase in tests code coverage
* Code comment fixes

PR #4234, opened 2024-01-12
docs: Document important labels and how to interact with them.

octo:


docs/CONTRIBUTING.md
```diff
@@ -51,13 +51,42 @@ the mailing list have a tendency to fall through the cracks.
     entries to `src/types.db`, you should talk to us early in the design
     process.
 
+### Labels
+
+PRs need to be categorized into one of three categories:
+
+*   **Feature**
+
+    The code has new behavior. Performance and documentation improvements are
```
eero-t, line 60:
```suggestion
    The code has new behavior. Also performance and documentation improvements (other than bug fixes) are
```

docs/CONTRIBUTING.md
```diff
@@ -51,13 +51,42 @@ the mailing list have a tendency to fall through the cracks.
...
+    considered features. These changes would typically go into a "feature
+    release".
+*   **Bug**
+
+    The code fixes a bug, regression, or otherwise unintended behavior. These
+    changes would typically go into a "patch" release.
+*   **Unlisted Change**
```
eero-t, line 67: I'd like more descriptive name for this category, maybe "User invisible change"?

docs/CONTRIBUTING.md
```diff
@@ -51,13 +51,42 @@ the mailing list have a tendency to fall through the cracks.
...
+### Labels
+
+PRs need to be categorized into one of three categories:
+
+*   **Feature**
+
+    The code has new behavior. Performance and documentation improvements are
+    considered features. These changes would typically go into a "feature
+    release".
+*   **Bug**
```
eero-t, line 63:
```suggestion
*   **Fix**
```

docs/CONTRIBUTING.md
```diff
@@ -51,13 +51,42 @@ the mailing list have a tendency to fall through the cracks.
...
+
+    The code fixes a bug, regression, or otherwise unintended behavior. These
```
eero-t, line 65:
```suggestion
    The code fixes a bug, regression, otherwise unintended behavior, or incorrect documentation. These
```

(incorrect, misleading, wrong... could not decide on best word.)

docs/CONTRIBUTING.md
```diff
@@ -66,11 +95,11 @@ ChangeLog: Foo plugin: A specific issue people had has been fixed.
...
 The summary must be on a line of its own, with a "ChangeLog:" prefix at the
-beginning of the line. The text should start with "Foo plugin" to give the
-reader context for the information. Other common contexts are "collectd" for
-the core daemon, "Build system", and "Documentation". Use past tense and
-passive voice the for remainder, e.g. "a bug has been fixed", "a feature has
-been added".
+beginning of the line. The text should start with "Foo plugin" or "collectd"
+(when making changes to the daemon itself) to give the reader context for the
+information. Other common contexts are "collectd" for the core daemon, "Build
+system", and "Documentation". Use past tense and passive voice the for
```
eero-t, line 101: "collectd" context is mentioned twice, maybe:
```suggestion
beginning of the line. To give the reader context for the information, the text
should start with "<name of> plugin" when making changes to a plugin, and
"collectd" for the core daemon changes.  Other common contexts are "Build
system" and "Documentation". Use past tense and passive voice the for
```

eero-t (changes requested):

docs/CONTRIBUTING.md
eero-t: Patch releases with "Fixes" or "Bug fixes" is fine, one with "Bugs" is not. :-)

docs/CONTRIBUTING.md
eero-t: "Not user visible" would be marginally shorter.

octo: The name was initially chosen to say "not listed in the change log". What do you think about "Internal"?

docs/CONTRIBUTING.md
octo: Fixed and bot has been updated.

docs/CONTRIBUTING.md
octo: This is not really helpful when trying to differentiate "Feature" and "Fix", because it essentially says "it's a feature, unless it's a fix".

I guess it's probably unlikely that we'll increase the "minor version" just for a documentation change. We can also document doc updates as "fixes" and trivial changes (e.g. fixing a typo) as "unlisted" / "internal" … WDYT?

docs/CONTRIBUTING.md
octo: Great, thanks!

docs/CONTRIBUTING.md
eero-t: I tried to clarify that:
* Some performance and documentation changes can be labeled as fixes (e.g. to large performance regression, or grossly misleading documentation)
* A given change (within PR) should be tagged only with one label [1]

Would there be some better wording for that?

[1] If PR contains multiple types of changes; features, fixes, and some non-user-visible changes, I assume that PR should include both "Feature" and "Fix" labels, but maybe "Internal" one could be skipped.

=> If PR has multiple labels, including "Fix", patch release should cherry-pick commits from the PR, instead of taking the whole PR?

=> Review should  reject commits that include multiple types, and ask PR to be split if it has commits of multiple types?

(I have one PR with 2 labels, but as Sysman plugin is new, these distinctions aren't really relevant for it as the whole thing is a new "Feature": https://github.com/collectd/collectd/pull/4079)

> I guess it's probably unlikely that we'll increase the "minor version" just for a documentation change. We can also document doc updates as "fixes" and trivial changes (e.g. fixing a typo) as "unlisted" / "internal" … WDYT?

Hm.  If PR fixes documentation _that relates to an earlier code change PR_ (not yet in release), IMHO it should have the same label as the code change, so that it gets picked to the same branch / release type.

Other documentation updates should be safe to include also into patch/fix releases, and typo fixes _are_ fixes...

Note: I'm not sure what's the role of "Internal" changes in deciding whether something goes to a patch/fix release or not.  Will they be included into those or not?

octo: > * Some performance and documentation changes can be labeled as fixes (e.g. to large performance regression, or grossly misleading documentation)

I have addressed this by using the word "typically", e.g. *"Performance improvements are **typically** considered features."* My goal is to give guidance to contributors that are unsure. If there's a good reason for labeling a performance improvement as "fix", that is perfectly fine.

> * A given change (within PR) should be tagged only with one label [1]

IMHO the doc already says this:
*"PRs need to be categorized into one of three categories"*

If that is not clear enough, we could add "exactly":
*"PRs need to be categorized into **exactly** one of three categories"*

> => Review should reject commits that include multiple types, and ask PR to be split if it has commits of multiple types?

This. Don't mix concerns; open separate PRs instead. I have added a note to this effect to the doc.

> (I have one PR with 2 labels, but as Sysman plugin is new, these distinctions aren't really relevant for it as the whole thing is a new "Feature": #4079)

The main problem, in my opinion, is that this code hasn't been released yet: is it really a "fix" if nobody is using the code yet? I think the best answer is to get the code into the hands of users. With regard to the label, either way is fine. When we create a release and there is at least one PR that is marked as a feature, the entire release is going to be considered a feature release.

> Hm. If PR fixes documentation _that relates to an earlier code change PR_ (not yet in release), IMHO it should have the same label as the code change, so that it gets picked to the same branch / release type.

As mentioned above, I don't think this matters enough that we should be arguing about it. I'd just use a "docs" prefix in the PR title, which currently maps to "Fix", but your approach also works for me.

> Note: I'm not sure what's the role of "Internal" changes in deciding whether something goes to a patch/fix release or not. Will they be included into those or not?

As mentioned (or possibly implied) in #4222 I'd like to move towards a "trunk based" development model. So the next release will include all new PRs and we don't cherry-pick into release branches.

docs/CONTRIBUTING.md
octo: I have seen "Maintenance" on a different project and think this is probably the best term overall. Changed.

eero-t: Yes, "Maintenance" is definitely best!

docs/CONTRIBUTING.md
```diff
@@ -66,16 +99,16 @@ ChangeLog: Foo plugin: A specific issue people had has been fixed.
...
-beginning of the line. The text should start with "Foo plugin" to give the
-reader context for the information. Other common contexts are "collectd" for
-the core daemon, "Build system", and "Documentation". Use past tense and
-passive voice the for remainder, e.g. "a bug has been fixed", "a feature has
-been added".
+beginning of the line. To give the reader context for the information, the text
+should start with "${name} plugin" when making changes to a plugin, and
+"collectd" for the core daemon changes. Other common contexts are "Build
+system", and "Documentation". Use past tense and passive voice the for
+remainder, e.g. "a bug has been fixed", "a feature has been added".
```
eero-t, line 106: To avoid anybody copying these too-generic suggestions literally:
```suggestion
remainder, e.g. "bug <id|description> has been fixed", "<id|description>
feature has been added".
```

eero-t (approved):
Approved.   Changelog entry examples could be improved a bit before merging though.

docs/CONTRIBUTING.md
octo: Great suggestion, thanks!

eero-t (approved):

PR #4236, opened 2024-01-15
[collectd 6] feat: Add the `--enable-compatibility-mode` flag to the `configure` script.

octo:
By default, plugins that use the "compatibility mode" (i.e. they are still using `value_list_t` / `plugin_dispatch_values()`) are disabled.

ChangeLog: Build system: the `--enable-compatibility-mode` has been added to control whether or not to build plugins using the compatibility mode. Such plugins are considered "unstable" and the metrics reported by these plugins will change in the future.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,14 +34,14 @@ jobs:
           - centos9
           - fedora39
           - fedora38
-        configure_flags: ['']
+        configure_flags: ['--enable-compatibility-mode']
         include:
           - container_tag: debian12
-            configure_flags: '--enable-debug'
+            configure_flags: '--enable-compatibility-mode --enable-debug'
           - container_tag: debian12
             # By default clang emits DWARF v5, which Valgrind cannot read yet.
             # https://github.com/llvm/llvm-project/issues/56550
-            configure_flags: 'CC=clang CXX=clang++ CFLAGS=-gdwarf-4'
+            configure_flags: '--enable-compatibility-mode CC=clang CXX=clang++ CFLAGS=-gdwarf-4'
     env:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -34,14 +34,14 @@ jobs:
...
+        configure_flags: ['--enable-compatibility-mode']
         include:
           - container_tag: debian12
-            configure_flags: '--enable-debug'
+            configure_flags: '--enable-compatibility-mode --enable-debug'
           - container_tag: debian12
             # By default clang emits DWARF v5, which Valgrind cannot read yet.
             # https://github.com/llvm/llvm-project/issues/56550
-            configure_flags: 'CC=clang CXX=clang++ CFLAGS=-gdwarf-4'
+            configure_flags: '--enable-compatibility-mode CC=clang CXX=clang++ CFLAGS=-gdwarf-4'
```
eero-t, line 44: Btw. Why there are 2 `same` tags, with different configure flags, does GitHub do separate build for each?

eero-t (approved):
Looks OK, I have just one question about the inlude section container_tags.

.github/workflows/build.yml
octo: What's happening here is that we set up a "matrix" build with two dimensions (line 23): `container_tag` and `configure_flags`. In the standard matrix, we have a bunch of values for `container_tag` but `configure_flags` has only one value (line 37).

The `include:` on line 38 tells GitHub actions that in addition to the matrix, we want two bespoke configurations: one builds with `--enable-debug` and the other with `CC=clang`. Both use the same `container_tag`, Debian 12.

HTH

PR #4235, opened 2024-01-13
 lua: add missing notification callback

kenhys:
ChangeLog: lua: add missing notification callback

* register_notification

In this version, it support callback function something like:

```
      function notification(notif)
         print(inspect(notif))
         return 0
      end

      collectd.register_notification(notification)
```

It accepts PUTNOTIF event via plugin_dispatch_notification then you can  handle it.

PR #4238, opened 2024-01-17
lua: add missing test case for #3779, #4235

kenhys:
ChangeLog: lua: add missing test case for #3779, #4235

* luaC_pushoconfigitem
* luaC_pushnotification

In the previous versions, there was no test code for Lua plugin. Thus this is first step to improve such a situation.

### DIFF
--- a/src/lua_test.c
+++ b/src/lua_test.c
@@ -1,0 +1,117 @@
+/**
+ * Copyright (C) 2024 Kentaro Hayashi
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ * Authors:
+ *   Kentaro Hayashi <kenhys at xdump.org>
+ **/
+
+#include "collectd.h"
+#include "lua.c"
+#include "testing.h"
+#include "utils_lua.c"
+
+static char config_path[PATH_MAX] = {0};
+const char *module_key_value_config = "<Plugin lua>\n"
+                                      "  Script \"example.lua\"\n"
+                                      "  <Module>\n"
+                                      "    Key \"Value\"\n"
+                                      "  </Module>\n"
+                                      "</Plugin>\n";
+
+void test_setup_config(const char *config) {
+  int fd = 0;
+  FILE *fp = NULL;
+  {
+    /* for simplicity, use lua.conf.XXXXXX here instead of .conf */
+    strncpy(config_path, "lua.conf.XXXXXX", sizeof(config_path));
+    fd = mkstemp(config_path);
+    fp = fdopen(fd, "w+");
+    fprintf(fp, "%s", config);
+    fclose(fp);
+  }
+  {
+    /* create empty Lua script which is loaded via lua_config */
+    fp = fopen("example.lua", "w+");
+    fclose(fp);
+  }
+}
+
+void test_teardown_config(void) {
+  lua_close(scripts->lua_state);
+  unlink(config_path);
+  unlink("example.lua");
+}
+
+lua_State *setup_lua(void) {
+  lua_State *L = luaL_newstate();
+  luaL_openlibs(L);
+  return L;
+}
+
+void teardown_lua(lua_State *L) { lua_close(L); }
+
+DEF_TEST(lua_config) {
+  test_setup_config(module_key_value_config);
+
+  oconfig_item_t *root = oconfig_parse_file(config_path);
+  oconfig_item_t *plugin = root->children;
+
+  /*
+    lua_config internally calls luaC_pushoconfigitem
+    and lua_State context is stored in each script.
+    <Module> will be mapped to Lua table:
+    { Key => Value } so check the value of "Key".
+  */
+  lua_config(plugin);
+  lua_State *L = scripts->lua_state;
+  lua_getglobal(L, "example.lua");
+  lua_getfield(L, -1, "Key");
+  EXPECT_EQ_STR("Value", luaL_checkstring(L, -1));
+
+  oconfig_free(root);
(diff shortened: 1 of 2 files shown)


eero-t:
Labeled as "Unlisted Change" to indicate that this is not directly user-visible change.

octo (commented):
Thanks for adding these tests @kenhys!

src/lua_test.c
```diff
@@ -0,0 +1,151 @@
...
+      </Plugin>
+    </root>
+  */
+  test_setup();
+  test_setup_config(module_key_value_config);
+  oconfig_item_t *root = oconfig_parse_file(TEST_LUA_PLUGIN_CONF);
+  oconfig_item_t *plugin, *module, *ci;
+  char buf[PATH_MAX];
+  CHECK_NOT_NULL(root);
+  EXPECT_EQ_INT(0, root->values_num);
```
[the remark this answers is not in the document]
kenhys: It is added to ensure oconfig_parse_file behavior, but also redundant in this context. 
I've removed it.

src/lua_test.c
```diff
@@ -0,0 +1,151 @@
...
+  oconfig_free(root);
+  test_teardown();
+  return 0;
+}
+
+DEF_TEST(luaC_pushnotification) {
+  notification_t notify = {.severity = NOTIF_OKAY, .plugin = "lua"};
+
+  test_setup();
+  test_setup_config(module_key_value_config);
```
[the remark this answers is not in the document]
kenhys: It should be removed, thanks.

src/lua_test.c
```diff
@@ -0,0 +1,151 @@
...
+  /*
+    <Module> configuration will be mapped to the following Lua table:
+
+    {
+      Key => Value
+    }
+  */
+  lua_newtable(L);
+  for (int i = 0; i < ci->children_num; i++) {
+    luaC_pushoconfigitem(L, &ci->children[i]);
```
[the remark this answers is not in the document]
kenhys: It may be better to simplify it, thanks.

src/lua_test.c
```diff
@@ -0,0 +1,112 @@
...
+                                      "    Key \"Value\"\n"
+                                      "  </Module>\n"
+                                      "</Plugin>\n";
+
+void test_setup_config(const char *config) {
+  int fd = 0;
+  FILE *fp = NULL;
+  {
+    /* create temporary .conf */
+    strncpy(config_path, "lua-XXXXXX.conf", strlen("lua-XXXXXX.conf") + 1);
```
octo, line 45: I think it would be easier to just initialize `config_path` when declaring it:

```
static char config_path[] = "/tmp/lua-XXXXXX.conf";
```

Alternatively, if you decide to keep the `strncpy()` call, please use `sizeof(config_path)-1` as the size argument. The size argument should be the size of the buffer to be written to, not the length of the string.

src/lua_test.c
```diff
@@ -0,0 +1,112 @@
...
+    fd = mkstemps(config_path, /* .conf */ 5);
```
octo, line 46: At your option: pass `strlen(".conf")` here instead of the `5` literal.

I don't think that `oconfig_parse_file()` actually cares about the suffix, so you could make your life easier by using e.g. `/tmp/lua.conf.XXXXXX` as the template.

src/lua_test.c
```diff
@@ -0,0 +1,112 @@
...
+  */
+  lua_config(plugin);
+  L = scripts->lua_state;
+  lua_getglobal(L, "example.lua");
+  lua_getfield(L, -1, "Key");
+  EXPECT_EQ_STR("Value", luaL_checkstring(L, -1));
+
+  oconfig_free(root);
+  unlink(config_path);
+  unlink("example.lua");
```
octo, line 85: Please move this to `test_teardown()`.

src/lua_test.c
```diff
@@ -0,0 +1,112 @@
...
+  return 0;
+}
+
+DEF_TEST(luaC_pushnotification) {
+  notification_t notify = {.severity = NOTIF_OKAY, .plugin = "lua"};
+
+  test_setup();
```
octo, line 92: I don't think the setup/teardown is required for this test.

src/lua_test.c
```diff
@@ -0,0 +1,112 @@
...
+}
+
+void test_setup(void) {
+  L = luaL_newstate();
+  luaL_openlibs(L);
+}
+
+void test_teardown(void) { lua_close(L); }
+
+DEF_TEST(luaC_pushoconfigitem) {
```
octo, line 65: This is a great test, thank you!

```suggestion
DEF_TEST(lua_config) {
```

octo (commented):
This looks much better now, thanks @kenhys!

src/lua_test.c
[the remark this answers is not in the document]
eero-t: > As it is, this test verifies that the config parser works as advertised. This is very valuable – the config parsing is woefully undertested. Untested, really.

@octo, for Sysman plugin I have a directory with invalid collectd config files, which are parsed with collectd under Valgrind, with the expectation that collectd should return error for each config, but Valgrind should not report any errors.

Similar approach works as well for testing basic collectd config handling and any other plugin:
```
# test configs are supposed to be in same dir as this script
confdir=${0%/*}

failexit=1
memexit=33

fail=0
unknown=0
memcheck=0
output=$(mktemp)
testconf=$(mktemp)
echo "Validating invalid config handling..."
for conf in "$confdir"/*.cnf; do
        echo
        echo "$conf:"
        cat "$baseconf" "$conf" > "$testconf"
        valgrind --error-exitcode=$memexit "$collectd" -T -C "$testconf" > "$output" 2>&1
        ret=$?
        if [ $ret -eq $failexit ]; then
                echo "  PASS: exit=$ret"
                fail=$((fail+1))
        elif [ $ret -eq $memexit ]; then
                echo "  ERROR: unexpected valgrind error"
                memcheck=$((memcheck+1))
        else
                echo "  ERROR: unexpected collectd exit=$ret"
                unknown=$((unknown+1))
        fi
        echo $line
        cat "$output"
        echo $line
done
rm "$baseconf" "$output"

echo
echo "Config parsing results:"
echo "- $fail expected fails"
echo "- $unknown unexpected exits"
echo "- $memcheck unexpected valgrind failures"
if [ $((unknown+memcheck)) -gt 0 ] ; then
        echo "=> ERROR"
        exit 1
fi
echo "=> PASS"
```

PS. Instead of using Valgrind, collectd could be built with ASAN as that's much faster than Valgrind, and checks also stack accesses.  In general, it might be good to run also other tests with ASAN built code: https://github.com/google/sanitizers/wiki/AddressSanitizer

src/lua_test.c
eero-t: Pattern can be given directly for `mkstemps()`, no need to have separate variable for it.

kenhys: Fixed.

src/lua_test.c
kenhys: It looks reasonable to use mkstemp. if suffix is not limited to .conf.

src/lua_test.c
kenhys: Thanks, fixed.

src/lua_test.c
kenhys: These are outcome of test_setup_config, so I doubt it should be.

octo: That's my point: the environment that is set up in the setup() function, should be cleaned up in the teardown() function. The test code itself shouldn't have to worry about whether files were created and where.

It looks like there are multiple separate concerns here:

* Creating the config file -> `test_setup_config()`
* Setting up the Lua interpreter -> `test_setup()`

I suggest that:

1.  We rename `test_setup()` and `test_teardown()` to `setup_lua()` and `teardown_lua()` respectively.
2.  We pair `setup_config()` with a `teardown_config()` function.

octo (approved):
Thanks @kenhys!

eero-t:
`release_ready` check fails to:
`FAIL: test_plugin_lua`

octo:
I have merged #4259 which should help debugging this problem. Can you rebase or pull in those changes?

kenhys:
Rebased with main.

kenhys:
all checks have passed! :+1:

octo:
Thanks @kenhys!

PR #4239, opened 2024-01-17
gpu_sysman: enable plugin automatically when `pkg-config` finds `level-zero` dependency

eero-t:
ChangeLog: Build system: Automatically enable the *gpu_sysman plugin* when `level-zero` is available.

While GPU plugin uses just Sysman API subset of the Level-Zero API family, name of the used library / loader is `level-zero` => renamed configure variables for the dependency to use level-zero/LEVEL_ZERO names.

Now that Level-Zero packages are in the distributions, I thought Sysman API GPU plugin can be enabled by default when `pkg-config` finds `level-zero`.  I did not think it that useful for user to be able to specify `level-zero` path with a configure option instead of suitable `PKG_CONFIG_PATH` env variable => dropped `--with-sysman` option.

(There was no good example of doing that in addition to using `pkg-config` in reasonable amount of configure.ac code.)

PS. Sysman plugin test does much more thorough checking than other tests, so running it takes >10x more time than the other tests.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1177,12 +1177,12 @@ endif
 if BUILD_PLUGIN_GPU_SYSMAN
 pkglib_LTLIBRARIES += gpu_sysman.la
 gpu_sysman_la_SOURCES = src/gpu_sysman.c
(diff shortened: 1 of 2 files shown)


octo (approved):
LGTM

PR #4242, opened 2024-01-19
disk: add /proc/diskstats fields 15 to 20 in KERNEL_LINUX

zzzyhtheonly:
Kernel 4.18+ appends four more fields for discard tracking putting the total at 18.
Kernel 5.5+ appends two more fields for flush requests. 
Details see: https://www.kernel.org/doc/Documentation/ABI/testing/procfs-diskstats

This patch makes collectd available for those fields.

Tested on kernel 5.15* (Ubuntu 22.04), e, g, 'write_log' writes things like:
"[{"values":[9993,5746],"dstypes":["derive","derive"],"dsnames":["complete","time"],"time":1705653958.743,"interval":10.000,"host":"7643","plugin":"disk","plugin_instance":"nvme0n1","type":"disk_flush","type_instance":""}]"

ChangeLog: disk plugin: supports /proc/diskstats fields 15 to 20 in Linux

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -112,6 +112,8 @@ typedef struct diskstats {
   bool has_merged;
   bool has_in_progress;
   bool has_io_time;
+  bool has_discard;
+  bool has_flush;
 
   struct diskstats *next;
 } diskstats_t;
@@ -307,6 +309,19 @@ static int disk_shutdown(void) {
 #endif /* KERNEL_LINUX */
   return 0;
 } /* int disk_shutdown */
+
+static void disk_submit_single(const char *plugin_instance, const char *type,
+                               derive_t value) {
+  value_list_t vl = VALUE_LIST_INIT;
+
+  vl.values = &(value_t){.derive = value};
+  vl.values_len = 1;
+  sstrncpy(vl.plugin, "disk", sizeof(vl.plugin));
+  sstrncpy(vl.plugin_instance, plugin_instance, sizeof(vl.plugin_instance));
+  sstrncpy(vl.type, type, sizeof(vl.type));
+
+  plugin_dispatch_values(&vl);
+} /* void disk_submit_single */
 
 static void disk_submit(const char *plugin_instance, const char *type,
                         derive_t read, derive_t write) {
@@ -740,10 +755,17 @@ static int disk_read(void) {
   derive_t io_time = 0;
   derive_t weighted_time = 0;
   derive_t diff_io_time = 0;
+  derive_t discard_ops = 0;
+  derive_t discard_merged = 0;
+  derive_t discard_sectors = 0;
+  derive_t discard_time = 0;
+  derive_t flush_ops = 0;
+  derive_t flush_time = 0;
   int is_disk = 0;
 
   diskstats_t *ds, *pre_ds;
 
+  /* https://www.kernel.org/doc/Documentation/ABI/testing/procfs-diskstats */
   if ((fh = fopen("/proc/diskstats", "r")) == NULL) {
     ERROR("disk plugin: fopen(\"/proc/diskstats\"): %s", STRERRNO);
     return -1;
@@ -788,6 +810,21 @@ static int disk_read(void) {
       write_sectors = atoll(fields[6]);
     } else {
       assert(numfields >= 14);
+
+      /* Kernel 4.18+, Discards */
+      if (numfields >= 18) {
+        discard_ops = atoll(fields[14]);
+        discard_merged = atoll(fields[15]);
+        discard_sectors = atoll(fields[16]);
+        discard_time = atoll(fields[17]);
+      }
+
+      /* Kernel 5.5+, Flush */
+      if (numfields >= 20) {
+        flush_ops = atoll(fields[18]);
+        flush_time = atoll(fields[19]);
+      }
+
       read_ops = atoll(fields[3]);
       write_ops = atoll(fields[7]);
 
@@ -882,6 +919,13 @@ static int disk_read(void) {
       if (io_time)
         ds->has_io_time = true;
 
+      ds->has_discard =
+          discard_ops || discard_merged || discard_sectors || discard_time;
+
+      /* There is chance 'has_flush' is true while 'has_discard' remains false
+       */
+      ds->has_flush = flush_ops || flush_time;
+
     } /* if (is_disk) */
 
     /* Skip first cycle for newly-added disk */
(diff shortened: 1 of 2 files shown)


eero-t:
Changes themselves look reasonable, but I think it would be high time for functionality in that monster-sized `disk_read()` to be split into several helper functions (not in this PR, but maybe in another one before merging this one).

I also wonder whether it would make sense there to be an option to disable these new metrics.

@octo?

src/disk.c
```diff
@@ -740,6 +778,12 @@ static int disk_read(void) {
   derive_t io_time = 0;
   derive_t weighted_time = 0;
   derive_t diff_io_time = 0;
+  derive_t discard_complete = 0;
+  derive_t discard_merged = 0;
+  derive_t discard_sectors = 0;
+  derive_t discard_time = 0;
+  derive_t flush_complete = 0;
+  derive_t flush_time = 0;
```
eero-t, line 786: You could add link to doc above the `fopen("/proc/diskstats", "r")` call:
`// https://www.kernel.org/doc/Documentation/ABI/testing/procfs-diskstats`

src/types.db
```diff
@@ -57,7 +57,9 @@ df_inodes               value:GAUGE:0:U
 dilution_of_precision   value:GAUGE:0:U
 disk_allocation         value:GAUGE:0:U
 disk_capacity           value:GAUGE:0:U
+disk_discard            complete:DERIVE:0:U, merged:DERIVE:0:U, sectors:DERIVE:0:U, time:DERIVE:0:U
```
octo, line 60: Please don't combine multiple metrics / data sources into one "data set". Instead, use a single data source called "value" and put "complete", "merged", … into the `type_instance` field.

src/types.db
```diff
@@ -57,7 +57,9 @@ df_inodes               value:GAUGE:0:U
...
 disk_error              value:GAUGE:0:U
+disk_flush              complete:DERIVE:0:U, time:DERIVE:0:U
```
octo, line 62: Dito.

src/disk.c
```diff
@@ -882,6 +941,14 @@ static int disk_read(void) {
       if (io_time)
         ds->has_io_time = true;
 
+      if (discard_complete || discard_merged || discard_sectors || discard_time)
+        ds->has_discard = true;
```
octo, line 945:
```suggestion
      ds->has_discard = discard_complete || discard_merged || discard_sectors || discard_time;
```

src/disk.c
```diff
@@ -882,6 +941,14 @@ static int disk_read(void) {
...
+
+      /* There is chance 'has_flush' is true while 'has_discard' remains false
+       */
+      if (flush_complete || flush_time)
+        ds->has_flush = true;
```
octo, line 950:
```suggestion
      ds->has_flush = flush_complete || flush_time;
```

octo (commented):
Hi @zzzyhtheonly, thank you so much for your change!

We need to split the metric definitions up so that each "value list" contains a single value: https://www.collectd.org/documentation/faq.html#faq:split_up_data_sets

Once that's done I think this is good to go.

octo:
> Changes themselves look reasonable, but I think it would be high time for functionality in that monster-sized `disk_read()` to be split into several helper functions (not in this PR, but maybe in another one before merging this one).

Yeah, some of these plugins are in desperate need of a refactor. Unfortunately the test coverage is very bad, making this a brittle exercise.

> I also wonder whether it would make sense there to be an option to disable these new metrics.

That would probably make sense to add.

src/types.db
zzzyhtheonly: I put them into the `type` field to align with other metrics in disk.c (almost all of them put them into `type`, e, g, "disk_ops", "disk_time"...)

zzzyhtheonly: /* this feature */
[{"values":[0],"dstypes":["derive"],"dsnames":["value"],"time":1705908765.727,"interval":10.000,"host":"7643","plugin":"disk","plugin_instance":"nvme0n1","type":"disk_discard_merged","type_instance":""}]
write_log values:

/* other old metric */
[{"values":[94271,10518372],"dstypes":["derive","derive"],"dsnames":["read","write"],"time":1705908775.727,"interval":10.000,"host":"7643","plugin":"disk","plugin_instance":"nvme0n1","type":"disk_merged","type_instance":""}]

zzzyhtheonly:
> > Changes themselves look reasonable, but I think it would be high time for functionality in that monster-sized `disk_read()` to be split into several helper functions (not in this PR, but maybe in another one before merging this one).
> 
> Yeah, some of these plugins are in desperate need of a refactor. Unfortunately the test coverage is very bad, making this a brittle exercise.
> 
> > I also wonder whether it would make sense there to be an option to disable these new metrics.
> 
> That would probably make sense to add.

Sounds good, please let me work on this in a separate PR.  Maybe after this one is done?

src/types.db
```diff
@@ -57,7 +57,13 @@ df_inodes               value:GAUGE:0:U
 dilution_of_precision   value:GAUGE:0:U
 disk_allocation         value:GAUGE:0:U
 disk_capacity           value:GAUGE:0:U
+disk_discard_complete   value:DERIVE:0:U
+disk_discard_merged     value:DERIVE:0:U
+disk_discard_sectors    value:DERIVE:0:U
+disk_discard_time       value:DERIVE:0:U
 disk_error              value:GAUGE:0:U
+disk_flush_complete     value:DERIVE:0:U
+disk_flush_time         value:DERIVE:0:U
```
octo, line 66: Oh, I didn't realize that these are actually different type metrics. Putting the read and write metrics into one "type", e.g. for `disk_ops`, bites us again :(

I see two options:

1.  Just add new metrics for these fields, like the PR currently does.

    If we go down the route, I'd like to rename `…_complete` to `…_ops` though, to be consistent with other disk metrics.
1.  Change existing metrics to use single-value types. Because this is a breaking change, it must be guarded by a config option that's off by default. In this option, we'd use types like `disk_ops_complex` and set the `type_instance` to "read", "write", "discard", "flush".

Overall I think that 1. is less disruptive to users. What do you think?

zzzyhtheonly: IMHO, I guess we should first let those new metrics align with old ones, it might be an easier way to use those metrics for now, like what Option 1 does.

Option 2, if you mean we change things into e.g.  "type: 'disk_ops', type_instance: 'read'", "type: 'disk_flush', type_instance: 'ops'"? If I understand correctly, this also sounds reasonable to me.  
But since we need a new conf option to implement this, maybe it is inconvenient to only let new metrics work on new options (which makes users confused I guess). 

So maybe we should go on this PR for Option 1 to let it work first. Then start a new PR for adding a conf option to let `disk_ops_complex` with `type_instance` filled to substitute existing metrics if necessary.  And it is okay for me to try working on this (Option 2). ;)

octo: SGTM, thanks for making those changes :D

octo (approved):

Issue #4241, opened 2024-01-18
[collectd 6] Compiler warnings for collectd core with stricter compilation options


eero-t:
*   Version of collectd: collectd-6.0 git
*   Operating system / distribution: Fedora 38

## Steps to reproduce

*  Prepare `config.h`:  `./configure `
*  build (most of) collectd core files with stricter warnings:
```
for i in $(find src/ -mindepth 3 -name '*.c' | grep -v -e /dpdk/ -e format_json.c -e format_kairosdb.c -e _mock.c -e _test.c -e _windows.c); do
  echo "*** $i ***";
  gcc -I. -Isrc -Isrc/daemon -Isrc/libcollectdclient -DHAVE_CONFIG_H -O3 -Werror -Wall -Wextra -Wno-missing-field-initializers -Wcast-align=strict -Wformat-security -Wnull-dereference -Wstrict-overflow=2 -Warray-bounds=2 -D_FORTIFY_SOURCE=2 -Wno-aggressive-loop-optimizations -c -o /dev/null $i;
done
```

(I do not have compatible include files for the files I've left out, and warnings from mock/test/windows files were not that relevant.  Proper warning analysis requires use of high optimization level.)

## Actual output

Output shows few warnings:
```
*** src/utils/cmds/listval.c ***
*** src/utils/cmds/putnotif.c ***
*** src/utils/cmds/putval.c ***
*** src/utils/cmds/putmetric.c ***
*** src/utils/cmds/flush.c ***
*** src/utils/cmds/parse_option.c ***
*** src/utils/cmds/getthreshold.c ***
*** src/utils/cmds/getval.c ***
*** src/utils/cmds/cmds.c ***
*** src/utils/curl_stats/curl_stats.c ***
*** src/utils/resource_metrics/resource_metrics.c ***
*** src/utils/value_list/value_list.c ***
*** src/utils/format_json/open_telemetry.c ***
*** src/utils/ignorelist/ignorelist.c ***
*** src/utils/crc32/crc32.c ***
*** src/utils/format_influxdb/format_influxdb.c ***
*** src/utils/gce/gce.c ***
*** src/utils/avltree/avltree.c ***
*** src/utils/oauth/oauth.c ***
*** src/utils/taskstats/taskstats.c ***
*** src/utils/utf8/utf8.c ***
*** src/utils/ovs/ovs.c ***
src/utils/ovs/ovs.c: In function ‘ovs_yajl_gen_val’:
[...]

PR #4243, opened 2024-01-19
[collectd 6] Fix most -Wpedantic warnings in collectd core

eero-t:
ChangeLog: Fix most `-Wpedantic` warnings in collectd core

I think first commit makes code more readable (and fixes comment typo).

Second commit should probably be dropped, as I do not see any reliable way to detect compilers that do not support C99 (`__func__` is variable in GCC, so one cannot use `#ifdef` check for it). It's included in case C89 compiler support is something that we do not actually care about (then few plugins could have similar change too)?

### DIFF
--- a/src/daemon/utils_cache.c
+++ b/src/daemon/utils_cache.c
@@ -909,7 +909,7 @@ static meta_data_t *uc_get_meta(metric_t const *m) /* {{{ */
 /* Sorry about this preprocessor magic, but it really makes this file much
  * shorter.. */
 #define UC_WRAP(wrap_function)                                                 \
-  {                                                                            \
+  do {                                                                         \
     pthread_mutex_lock(&cache_lock);                                           \
     errno = 0;                                                                 \
     meta_data_t *meta = uc_get_meta(m);                                        \
@@ -920,23 +920,23 @@ static meta_data_t *uc_get_meta(metric_t const *m) /* {{{ */
     int ret = wrap_function(meta, key);                                        \
     pthread_mutex_unlock(&cache_lock);                                         \
     return ret;                                                                \
-  }
+  } while (0)
 
-int uc_meta_data_exists(metric_t const *m, const char *key)
-    UC_WRAP(meta_data_exists);
-
-int uc_meta_data_delete(metric_t const *m, const char *key)
-    UC_WRAP(meta_data_delete);
-
+int uc_meta_data_exists(metric_t const *m, const char *key) {
+  UC_WRAP(meta_data_exists);
+}
+int uc_meta_data_delete(metric_t const *m, const char *key) {
+  UC_WRAP(meta_data_delete);
+}
 /* The second argument is called `toc` in the API, but the macro expects
  * `key`. */
-int uc_meta_data_toc(metric_t const *m, char ***key) UC_WRAP(meta_data_toc);
+int uc_meta_data_toc(metric_t const *m, char ***key) { UC_WRAP(meta_data_toc); }
 #undef UC_WRAP
 
 /* We need a new version of this macro because the following functions take
- * two argumetns. gratituous semicolons added for formatting sanity*/
+ * two arguments. Gratuitous semicolons added for formatting sanity. */
 #define UC_WRAP(wrap_function)                                                 \
-  {                                                                            \
+  do {                                                                         \
     pthread_mutex_lock(&cache_lock);                                           \
     errno = 0;                                                                 \
     meta_data_t *meta = uc_get_meta(m);                                        \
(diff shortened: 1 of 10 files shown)


eero-t:
Changed casts in several files to match what's expected.  With this, all `-Wpedantic` warnings listed in https://github.com/collectd/collectd/issues/4241 are fixed.

In `resource_metrics.c`, another option would have been changing `resource_metrics_compare()` `compare_family_by_name()` and `compare_metrics()` signatures to matched what `bsearch()` and `qsort()` expect, and cast their arguments instead.

eero-t:
> Second commit should probably be dropped, as I do not see any reliable way to detect compilers that do not support C99 (`__func__` is variable in GCC, so one cannot use `#ifdef` check for it). It's included in case C89 compiler support is something that we do not actually care about (then few plugins could have similar change too)?

@octo any comments on this?

While `-Wpendantic` so far has not detected any actual issues, I think the code passing its checks is more readable (casts are correct etc).   If C89 support is needed, then I do not pursue this further though, as that would prevent eventually enabling `-Wpedantic` for all the code.

octo:
> > Second commit should probably be dropped, as I do not see any reliable way to detect compilers that do not support C99 (`__func__` is variable in GCC, so one cannot use `#ifdef` check for it). It's included in case C89 compiler support is something that we do not actually care about (then few plugins could have similar change too)?
> 
> @octo any comments on this?

We're using many C99 features, such as `<stdbool.h>`, variable length arrays, and compound literals. If `__func__` is specified in C99 and `__FUNCTION__` is not, we should use `__func__`.

octo (approved):
LGTM. Not merging yet due to the "WIP" in the title, but feel free to go ahead and merge yourself.

eero-t:
Dropped the `__FUNCTION__` change (and WIP from title).

While C compilers should be fine with it, if somebody compiles collectd with C++ one, there can be trouble, because `__func__` came to C++ only in C++11, and MSVC messed that up, so one cannot reliably check C++ version. See:
* https://stackoverflow.com/questions/7008485/func-or-function-or-manual-const-char-id
* https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

octo:
> Changed casts in several files to match what's expected. With this, all `-Wpedantic` warnings listed in #4241 are fixed.
> 
> In `resource_metrics.c`, another option would have been changing `resource_metrics_compare()` `compare_family_by_name()` and `compare_metrics()` signatures to matched what `bsearch()` and `qsort()` expect, and cast their arguments instead.

Sorry, didn't see this earlier. I'd actually prefer this, like we did in the write_prometheus plugin: https://github.com/collectd/collectd/blob/76cd6e3230d966d0edef4ba4d883abed280b8be1/src/write_prometheus.c#L119

eero-t:
> Sorry, didn't see this earlier. I'd actually prefer this, like we did in the write_prometheus plugin:

Done.  Changed the function signatures instead of their casts, and fixed almost all of the other `-Wpedantic` issues in collectd core.

This gives now only single (unavoidable) warning for the checked 90 files:
```
$ for i in $(find src/ -mindepth 2 -name '*.c' | grep -v -e dpdk.c -e format_kairosdb.c -e ovs.c -e _windows.c); do
    echo "*** $i ***";
    gcc -I. -Isrc -Isrc/daemon -Isrc/libcollectdclient -DHAVE_CONFIG_H -O3 -Werror -Wpedantic -c -o /dev/null $i;
  done
...
src/daemon/plugin.c: In function ‘plugin_load_file’:
src/daemon/plugin.c:433:30: error: ISO C forbids initialization between function pointer and ‘void *’ [-Werror=pedantic]
  433 |   void (*reg_handle)(void) = dlsym(dlh, "module_register");
      |
...
```

src/daemon/utils_cache.c
```diff
@@ -920,23 +920,23 @@ static meta_data_t *uc_get_meta(metric_t const *m) /* {{{ */
...
+}
 /* The second argument is called `toc` in the API, but the macro expects
  * `key`. */
-int uc_meta_data_toc(metric_t const *m, char ***key) UC_WRAP(meta_data_toc);
+int uc_meta_data_toc(metric_t const *m, char ***key) { UC_WRAP(meta_data_toc); }
 #undef UC_WRAP
 
 /* We need a new version of this macro because the following functions take
- * two argumetns. gratituous semicolons added for formatting sanity*/
+ * two arguments. Gratuitous semicolons added for formatting sanity. */
```
octo, line 937: With the use of do … while in the macros, the semicolons are now required. Please remove the last sentence.

octo (approved):
Thanks @eero-t!

src/daemon/utils_cache.c
eero-t: `collectd-bot` is still not working correctly, as it merged this PR after you added `automerge` tag, although this comment was still unresolved. :-/

I could could do another PR fixing some other GCC warnings, and fix that comment in it...

octo: Yeah, I'll look into the bot. I think GitHub now has a native automerge feature. Maybe we should enable that instead.

Don't worry about it; I don't think this is worth a separate PR.

eero-t:
Argh. My `-Wpendantic` changes for `plugin.c` conflict with `-Wcast-function-type` warnings (included to `-Wextra`, which I'm testing now).

While `-Wpendantic` does not accept `(void *)` (data pointer) for function pointers, `-Wcast-function-type` ignores function cast type difference when that is used, but not with real function types.

@octo, I could fix that by changing `plugin_log()` to return 0 instead of `void` (and changing placeholder callback type in `plugin.c` to `int (*)()`).   Does that sound acceptable?

(Or should I consider reverting `plugin.c` cast changes from this PR?)

octo:
Truly living up to its name, `-Wpedantic` is *technically* correct. The C standard says:

> 6.3.2.3 Pointers, 8:
> A pointer to a function of one type may be converted to a pointer to a function of another type and back again; the result shall compare equal to the original pointer. If a converted pointer is used to call a function whose type is not compatible with the referenced type, the behavior is undefined.

=> casting between (compatible) function pointers is okay.

> 6.3.2.3 Pointers, 1:
> A pointer to `void` may be converted to or from a pointer to any object type. A pointer to any object type may be converted to a pointer to void and back again; the result shall compare equal to the original pointer.

Where "object" is defined as:

> 3.15, 1
> object:
> region of data storage in the execution environment, the contents of which can represent values

=> casting between `void *` and any other pointer to a data type is okay

=> (by omission) casting between object pointers and function pointers is not okay

Based on the GCC documentation, I think we should use `void (*) (void)`:

>  The function type `void (*) (void)` is special and matches everything, which can be used to suppress this warning.

An alternative might be to define a union for holding the function pointers. This might allow us to avoid casting function pointers entirely:

```c
typedef union {
  plugin_init_cb init;
  plugin_read_cb read;
  plugin_log_cb log;
  // …
} plugin_callback_t;
```

eero-t:
I tried all 3 alternatives in https://github.com/collectd/collectd/pull/4261, but `plugin_log()` return value change caused CI failures, so dropped that.

Please tell which solution you'd rather prefer:
* `void (*) (void)`: "plugin.c: use callback pointer type allowing signature mismatches"
* using `enum`: "plugin.c: use union as generic callback type"
?

octo:
> Please tell which solution you'd rather prefer:

I really like the union approach you now have in #4261. Using `callback.log` is so much nicer than casting function pointers from one type to another.

Issue #4240, opened 2024-01-18
interface plugin no values returned on Solaris 11.4 (Patch included)


hergla:
*   Version of collectd:  collectd 5.12.0.287.g5dba56b
*   Operating system / distribution:  Solaris / 11.4
*   Kernel version (if applicable):  11.4.62.151.3

## Expected behavior

should return values for the network interfaces.

## Actual behavior

No values for network interfaces returned.

## Steps to reproduce

Compiler will not reach the HAVE_LIBKSTAT.
Reason is:  HAVE_GETIFADDRS is set to TRUE from configure script. 

github gives an error when i try to add a pull request.
Patch:

```diff
diff --git a/src/interface.c b/src/interface.c
index 4e7c5739..6c1ea251 100644
--- a/src/interface.c
+++ b/src/interface.c
@@ -67,6 +67,10 @@
 #endif /* !COLLECT_GETIFADDRS */
 #endif /* KERNEL_LINUX */

+#if HAVE_LIBKSTAT
+#undef HAVE_GETIFADDRS
+#endif
+
 #if HAVE_PERFSTAT
 static perfstat_netinterface_t *ifstat;
 static int nif;
@@ -85,6 +89,7 @@ static const char *config_keys[] = {
     "Interface",
     "IgnoreSelected",
     "ReportInactive",
+    "UniqueName",
 };
 static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
 
```

octo:
Thanks for reporting this @hergla!

The patch you suggest would probably break FreeBSD, but we can fix this by giving kstat preference over *getifaddrs()*.

What kind of problem are you facing when trying to open a pull request?

PR #4247, opened 2024-01-22
interface plugin: Reorder implementations to put kstat above getifaddrs.

octo:
ChangeLog: Interface plugin: Prefer `kstat` over `getifaddrs` to fix the plugin on Solaris.
Fixes: #4240

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -85,6 +85,7 @@ static const char *config_keys[] = {
     "Interface",
     "IgnoreSelected",
     "ReportInactive",
+    "UniqueName",
 };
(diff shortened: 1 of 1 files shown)


eero-t (approved):
E.g. BSDs can also have a libary implementing SunOS `libkstat` interfaces installed:
https://man.freebsd.org/cgi/man.cgi?query=libkstat&sektion=3LIB&apropos=0&manpath=SunOS+5.10

In addition to `getifaddrs` one:
https://man.freebsd.org/cgi/man.cgi?getifaddrs

I doubt many would have such compat things installed, especially their devel headers, so approving this.

PR #4250, opened 2024-01-23
.cirrus.yml: statgrab: Print `ldd` of failing plugins. Continue on error.

octo:
Investigating this error:

```
plugin cpu NOT linked against libstatgrab
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -153,7 +153,7 @@ non_standard_toolchains_task:
             if ! $(ldd ".libs/${plugin}.so" 2>/dev/null | grep -q 'libstatgrab.so'); then
               echo "plugin ${plugin} is NOT linked against libstatgrab:"
               ldd ".libs/${plugin}.so" | sed -e 's/^/    /'
-              err=$(($err + 1))
+              err=$((err + 1))
             fi
           done
           exit $err
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -145,7 +145,7 @@ non_standard_toolchains_task:
           CPPLAGS="$(dpkg-buildflags --get CPPFLAGS)"
           LDFLAGS="$(dpkg-buildflags --get LDFLAGS)"
       build_script:
-        - make -j$(nproc) -sk
+        - make -j$(nproc) -sk V=1
       tests_script:
         - |
           err=0
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -147,7 +147,7 @@ non_standard_toolchains_task:
       build_script:
         - make -j$(nproc) -sk
       tests_script:
-        - >
+        - |
           err=0
           for plugin in cpu disk interface load memory swap users; do
             if ! $(ldd ".libs/${plugin}.so" 2>/dev/null | grep -q 'libstatgrab.so'); then
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -110,10 +110,8 @@ redhat_default_toolchain_task:
 non_standard_toolchains_task:
   container:
     image: collectd/ci:debian12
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -150,12 +148,15 @@ non_standard_toolchains_task:
...
-              exit 1
+          err=0
+          for plugin in cpu disk interface load memory swap users; do
+            if ! $(ldd ".libs/${plugin}.so" 2>/dev/null | grep -q 'libstatgrab.so'); then
+              echo "plugin ${plugin} is NOT linked against libstatgrab:"
+              ldd ".libs/${plugin}.so" | sed -e 's/^/    /'
+              err=$(($err + 1))
             fi
           done
+          exit $err
```
eero-t, line 159: Because of this error:
```
err=0 for plugin in cpu disk interface load memory swap users; do
/tmp/scriptsb51677938757ea6f822ffb83f91a272f.sh: line 5: syntax error near unexpected token `do'
```

I think this needs semicolons to separate statements:
```suggestion
          err=0;
          for plugin in cpu disk interface load memory swap users; do
            if ! $(ldd ".libs/${plugin}.so" 2>/dev/null | grep -q 'libstatgrab.so'); then
              echo "plugin ${plugin} is NOT linked against libstatgrab:";
              ldd ".libs/${plugin}.so" | sed -e 's/^/    /';
              err=$((err + 1));
            fi;
          done;
          exit $err
```
(And `$` is redundant for variable names within `$(())`, as `shellcheck` likes to often remind me.)

eero-t (changes requested):

.cirrus.yml
octo: I think the problem is the `>` multi-line indicator, because it replaces newlines with spaces. With `|` this should work. Now I don't know how this could ever have worked previously.

octo:
The plugins are indeed not linked with libstatgrab:

```
plugin cpu is NOT linked against libstatgrab:
    	linux-vdso.so.1 (0x00007ffd4bd31000)
    	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000078dbdd4ce000)
    	/lib64/ld-linux-x86-64.so.2 (0x000078dbdd6c9000)
```

My current hypothesis is that the linker is clever enough to notice that the plugins are not actually _using_ any symbols from libstatgrab and therefore omits the link.

octo:
Indeed:

```
libtool: link:  gcc -shared  -fPIC -DPIC  src/.libs/cpu_la-cpu.o   -lstatgrab -lpthread  -g -O2 -fstack-protector-strong -Wl,-z -Wl,relro   -Wl,-soname -Wl,cpu.so -Wl,-version-script -Wl,.libs/cpu.ver -o .libs/cpu.so
                                                                   ^^^^^^^^^^
```

I think our options are:

1.  Re-order the implementations in `src/cpu.c` (and all other plugins) to prefer libstatgrab over our native implementation.
2. Remove this check.

Personally, I'm in favor of 2. Any objections?

eero-t:
Does the order in which items are given on link line affect only linker symbol search, not also what gets linked to final binary?

As to which option to use, normally I prefer using code maintained by somebody, but at least the libstatgrab GitHub seems a bit dead, last merged PR was from 2021, and there are not really other commits: https://github.com/libstatgrab/libstatgrab/commits/master/

=> option 2 seems fine

eero-t (approved):
Approved, the change itself is fine.

Even if this check is eventually removed, I think it makes sense merge this improved version first, in case we want similar check back again, for this, or some other thing.

octo:
Okay, let's do that then.

PR #4251, opened 2024-01-23
[collectd 6] ci: Remove the "statgrab" task from CirrusCI.

octo:
The "statgrab" job builds collectd with the "statgrab" library and then expects certain plugins to be linked against it. However, since the build runs on Linux, the plugins are actually using the "native" implementation. The linker, noticing that the plugin is not referring to any symbol in the statgrab library, does not actually produce a link, causing the test to fail.

The native implementations provide much richer metrics than the libstatgrab based implementations, so we'll drop the test.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -131,33 +131,6 @@ non_standard_toolchains_task:
       tests_script:
         - VALGRIND_OPTS="--errors-for-leak-kinds=definite" make -j$(nproc) -sk check || (cat ./test-suite.log && false)
 
-    # build against libstatgrab, should always pass
-    - env:
-        LABEL: statgrab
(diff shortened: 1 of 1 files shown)


eero-t (changes requested):
Please add the reason for removing this to the commit description.

(For one-commit PRs, GH would have added that also as default PR description which is rather nice.)

octo:
> Please add the reason for removing this to the commit description.

Done, updated the PR description also.

eero-t:
Btw. although `Maintenance` tags do not require `ChangeLog` entry, I think those would be good to have, in case project ever wants to generate also log of maintenance change PRs.   Or is the intention then to use PR titles instead?

PR #4252, opened 2024-01-24
[collectd 6] fix: df plugin: Fix an array out of bounds crash.

octo:
The second argument of the `for` loop always evaluated to true, causing the code to read past the end of the array.

ChangeLog: df plugin: An out of bounds read, causing segmentation faults, has been fixed.

### DIFF
--- a/src/df.c
+++ b/src/df.c
@@ -332,7 +332,7 @@ static int df_read(void) {
       &fam_inode_utilization,
   };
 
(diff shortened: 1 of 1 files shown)


eero-t (approved):
Approved.   Commit message could include:
`Fixes: c6f5e418674f49`

PS. sorry for missing this!

PR #4256, opened 2024-01-25
disk.c: add config options to support diskstats 15-20 in KERNEL_LINUX

zzzyhtheonly:
According to feedback by
#4242
This PR lets discard and flush metrics to be disabled by default and able to enabled by config options.

ChangeLog: disk plugin: options to control discard and flush metrics

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -119,6 +119,9 @@ typedef struct diskstats {
 } diskstats_t;
 
 static diskstats_t *disklist;
+
+static bool report_discard;
+static bool report_flush;
 /* #endif KERNEL_LINUX */
 #elif KERNEL_FREEBSD
 static struct gmesh geom_tree;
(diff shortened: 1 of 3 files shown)


src/disk.c
```diff
@@ -814,15 +830,15 @@ static int disk_read(void) {
       assert(numfields >= 14);
 
       /* Kernel 4.18+, Discards */
-      if (numfields >= 18) {
+      if (numfields >= 18 && report_discard) {
```
octo, line 833: Please handle the config option just before calling `disk_submit_single` and always populate the `discard_*` and `flush_*` variables. That means line 1000 would look something like this:

```c
if (report_discard && ds->has_discard) {
```

Dito for `report_flush`.

src/disk.c
```diff
@@ -164,8 +167,9 @@ static char *conf_udev_name_attr;
 static struct udev *handle_udev;
 #endif
 
-static const char *config_keys[] = {"Disk", "UseBSDName", "IgnoreSelected",
-                                    "UdevNameAttr"};
+static const char *config_keys[] = {"Disk",           "UseBSDName",
+                                    "IgnoreSelected", "UdevNameAttr",
+                                    "ReportDiscard",  "ReportFlush"};
```
octo, line 172: Please add the new config options to the documentation:

* `src/collectd.conf.pod` (the *collectd.conf(5)* manpage)
* `src/collectd.conf.in` (the default config; add options with their default value there)

octo (changes requested):
Thank you so much @zzzyhtheonly! I've left some comments inline.

octo (approved):
✅ LGTM, thank you so much @zzzyhtheonly! 🎉

Issue #4200, opened 2023-12-19
collectd-6.0 regression, logs "uc_update" errors


eero-t:
*   Version of collectd: `collectd-6.0` branch HEAD
*   Operating system / distribution: Ubuntu 22.04

## Expected behavior

Like with earlier `collectd-6.0` code, no errors logged from collectd with `gpu_sysman` plugin input being queried through `write_prometheus` plugin HTTP endpoint.

## Actual behavior

Collectd logs errors like this:
```
uc_update: Value too old: name = collectd_gpu_sysman_frequency_mhz{resource:host.name="5e22ae14c686",resource:service.name="unknown_service:collectd",dev_file="card2",function="max",location="gpu",pci_bdf="0000:93:00.0",type="request"}; value time = 1703013964.867; last cache update = 1703013964.867;
uc_update: uc_update_metric failed: Error #-1; Additionally, strerror_r failed.
```

### Notes

I'm not sure when this regressed.  Last time I'm sure it worked, was after this fix: https://github.com/collectd/collectd/pull/4103

For some reason I get these only for frequency metric, not for the others.  Frequency is the first metric submitted on each read cycle.

If I increase interval to >1s, this log spam seems to go away.

This is the config I'm using:
```
Interval 10

LoadPlugin write_prometheus
<LoadPlugin gpu_sysman>
  Interval 0.5
</LoadPlugin>

<Plugin write_prometheus>
  Port 9103
</Plugin>

<Plugin gpu_sysman>
  Samples 8
</Plugin>
```

But getting enough metrics requires discrete Intel GPU, and Sysman plugin to be built & running with latest kernel + distro user-space drivers, or drivers installed according to instructions here (for LTS distros): https://dgpu-docs.intel.com/

eero-t:
@octo Any comments on these `uc_update` issues (previously fixed for 6.x in https://github.com/collectd/collectd/pull/4103)?

Issue https://github.com/collectd/collectd/issues/4119 for 5.x is potentially related.

eero-t:
@octo Thanks, you found it really fast!  (From my own code which was embarrassing, but I guess one is somewhat blind to own code)

I'll close this and updated the title as it was not in collectd core code.

octo:
You're welcome :) Such bugs are much easier to find with a fresh pair of eyes.

eero-t:
This: `uc_update: uc_update_metric failed: Error #-1; Additionally, strerror_r failed.`

[...]

PR #4257, opened 2024-01-26
[collectd 6] fix: Handle metric families with zero metrics gracefully.

octo:
This allows plugins to call `plugin_dispatch_metric_family()` unconditionally.

ChangeLog: Daemon: The `plugin_dispatch_metric_family()` function has been changed to handle metric families with zero metrics gracefully.

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -2105,6 +2105,9 @@ EXPORT int plugin_dispatch_metric_family(metric_family_t const *fam) {
     return EINVAL;
   }
   if (fam->metric.num == 0) {
+    DEBUG("plugin_dispatch_metric_family: Metric family \"%s\" contains zero "
+          "metrics.",
+          fam->name);
     return 0;
   }
 
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -2100,8 +2100,12 @@ static void set_default_resource_attributes(metric_family_t *fam) {
 }
 
 EXPORT int plugin_dispatch_metric_family(metric_family_t const *fam) {
-  if ((fam == NULL) || (fam->metric.num == 0)) {
+  if (fam == NULL) {
+    ERROR("plugin_dispatch_metric_family: fam == NULL");
     return EINVAL;
+  }
+  if (fam->metric.num == 0) {
+    return 0;
   }
 
   /* Create a copy of the metric_family_t so we can modify the time and
--- a/src/chrony.c
+++ b/src/chrony.c
@@ -1105,14 +1105,12 @@ static int chrony_read(void) {
   }
 
   for (size_t i = 0; i < FAM_CHRONY_MAX; i++) {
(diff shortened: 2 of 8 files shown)


src/daemon/plugin.c
```diff
@@ -2094,9 +2094,13 @@ static void set_default_resource_attributes(metric_family_t *fam) {
...
 
 EXPORT int plugin_dispatch_metric_family(metric_family_t const *fam) {
-  if ((fam == NULL) || (fam->metric.num == 0)) {
+  if (fam == NULL) {
+    ERROR("plugin_dispatch_metric_family: fam == NULL");
     return EINVAL;
   }
+  if (fam->metric.num == 0) {
+    return 0;
+  }
```
eero-t, line 2103: While I'm fine with removing `if (fams[i].metric.num > 0)` checks from the _plugins_, I'm against collectd core just silently ignoring that.

In which situations such plugin behavior (trying to report 0 metrics) would not be a bug?

PS. checks for `metric.num` against zero are in:
```
$ grep "if.*metric.num .* 0" *.c
chrony.c:    if (fams[i].metric.num > 0) {
df.c:    if (families[i]->metric.num == 0) {
disk.c:    if (fams[i]->metric.num == 0) {
ipc.c:    if (fams[i].metric.num > 0) {
irq.c:  if (fam.metric.num > 0) {
olsrd.c:    if (metric_num > 0)
ping.c:    if (fams[i]->metric.num > 0) {
write_prometheus.c:  if (prom_fam->metric.num == 0)
write_prometheus.c:  if (fam->metric.num == 0) {
write_prometheus.c:    if (prom_fam->metric.num == 0) {
```

eero-t (changes requested):

src/daemon/plugin.c
octo: The *disk plugin* is a good example: it defines a list of 8 metric families. Then architecture specific code adds metrics to some or all of these metric families. At the end, architecture agnostic code is calling `plugin_dispatch_metric_family` for all metric families. Previously it needed to check whether the family contained any metrics. Now this check is done centrally.

Overall I don't think that calling a function that accpets a list with an empty list is an error.

octo: The locations you called out are either simplified in thie PR (most), false positives (olsrd), or working as intended (write_prometheus).

eero-t: > Overall I don't think that calling a function that accepts a list with an empty list is an error.

At least it should not be silently ignored, it should be logged, as it could be a plugin error.

Duplicate metrics are logged (e.g. https://github.com/collectd/collectd/issues/4200), so IMHO missing metrics should also...

octo: I added a debug message. I really don't think we should log at any higher severity, this is working as intended.

eero-t (approved):
Approved, debug message is fine.

PR #3970, opened 2022-02-04
[collectd 6] Fix some gcc warnings with more strict checks

eero-t:
ChangeLog: Build: Fix some gcc warnings with more strict checks

Fixes are based on "-O3 -Werror -Wall -Wextra -Wformat-security" output, and partly shared with the v5 fixes in https://github.com/collectd/collectd/pull/3917:

* Properly initialize complex struct
* Fix signed vs unsigned comparison warnings
* Correct "write_prometheus" comment and fix metric family freeing in a loop
* Remove unused functions and function arguments, and when latter cannot be removed, tell compiler that they're not used

PR #4261, opened 2024-01-29
[collectd 6]: Fix some GCC "-Wextra" warnings in collectd core

eero-t:
Continuation for PR https://github.com/collectd/collectd/pull/4243, fixing more warnings from ticket https://github.com/collectd/collectd/issues/4241.

(Related also to v5 branch PR https://github.com/collectd/collectd/pull/3970.)

### DIFF
--- a/src/daemon/plugin.c
+++ b/src/daemon/plugin.c
@@ -60,14 +60,24 @@
 
 #include <dlfcn.h>
 
-/* except for plugin_log_cb, all callback types return an int */
-typedef int (*placeholder_cb_t)(void);
+typedef int (*plugin_read_simple_cb)(void);
+/* all possible callback types */
+typedef union {
+  plugin_init_cb init;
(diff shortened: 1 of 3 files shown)


octo (approved):
I really like how the code looks with the new union. We should have done this sooner :)

P.S.: Leaving merging to you, since the PR is still marked "WIP".

eero-t:
> I really like how the code looks with the new union. We should have done this sooner :)

Ok, I squashed away the other alternative away.

> P.S.: Leaving merging to you, since the PR is still marked "WIP".

Removed WIP and merged.  I'll work next on the other warnings (+ adding units support to sysman plugin).

@octo in the meanwhile it would be good if you can come up with a nice way of adding extra compiler warning options to `Makefile.am` for building the collectd core files, and maybe also rest of sources for `collectd*` binaries (but not plugins, as those will not pass stricter options).

PR #4266, opened 2024-01-31
[collectd 6] Introduce a floating point counter type.

octo:
**Note:** This PR is a super-set of #4265, which should be merged first.

This change introduces a new metric type, called "FPCOUNTER". This is useful when the resolution provided by integers is not sufficient. A good example is the `system.cpu.time` metric, which should be measured in seconds. However, with a ten second interval, the counters of all states combined will only increase by ten, resulting in very poor resolution.

One difference to integer counters is that floating point counters don't overflow. We therefore treat every situation in which the previous value is larger than the current value as a counter reset.

ChangeLog: Daemon: The new `FPCOUNTER` (floating point counter) metric type has been added.


src/daemon/metric.h
```diff
@@ -32,21 +32,24 @@
...
+#define METRIC_ATTR_CUMULATIVE 0x02
 
 typedef enum {
-  METRIC_TYPE_COUNTER = 0,
-  METRIC_TYPE_GAUGE = 1,
-  METRIC_TYPE_UNTYPED = 2,
+  METRIC_TYPE_UNTYPED = 0,
+  METRIC_TYPE_COUNTER = METRIC_ATTR_CUMULATIVE,
+  METRIC_TYPE_FPCOUNTER = METRIC_ATTR_DOUBLE | METRIC_ATTR_CUMULATIVE,
+  METRIC_TYPE_GAUGE = METRIC_ATTR_DOUBLE,
```
eero-t, line 42: IMHO clearer to define them in numerical order:
```suggestion
  METRIC_TYPE_UNTYPED = 0,
  METRIC_TYPE_GAUGE = METRIC_ATTR_DOUBLE,
  METRIC_TYPE_COUNTER = METRIC_ATTR_CUMULATIVE,
  METRIC_TYPE_FPCOUNTER = METRIC_ATTR_DOUBLE | METRIC_ATTR_CUMULATIVE,
```

src/daemon/plugin.h
```diff
@@ -40,10 +40,6 @@
 #include <inttypes.h>
 #include <pthread.h>
 
-#define DS_TYPE_COUNTER 0
-#define DS_TYPE_GAUGE VALUE_TYPE_GAUGE
-#define DS_TYPE_DERIVE VALUE_TYPE_DERIVE
-
 #define DS_TYPE_TO_STRING(t)                                                   \
   (t == DS_TYPE_COUNTER)  ? "counter"                                          \
   : (t == DS_TYPE_GAUGE)  ? "gauge"                                            \
```
eero-t, line 45: Move also `DS_TYPE_TO_STRING` macro to `value_list.h`?

eero-t (commented):
Reviewed just `daemon/` dir stuff. I'll check rest once the dependent PR is merged.

src/daemon/plugin.h
octo: Yes! Done.

src/daemon/metric.h
octo: 🤷  I was going for alphabetic order, but really don't care. Changed.

src/utils/format_open_telemetry/format_open_telemetry.cc
```diff
@@ -66,19 +66,22 @@ static void metric_to_number_data_point(NumberDataPoint *dp,
...
-    break;
+    return;
   case METRIC_TYPE_UNTYPED:
-    // never reached, only here to make the compiler happy
+    // Fall through. This case signals the compiler that we're checking all
+    // values of the enum. We report an error outside of the switch to also
+    // cover other values.
     break;
   }
+  // never reached, only here to make the compiler happy
```
eero-t, line 83: IMHO this would be more appropriate in `add_metric()`, maybe just drop it from here?
```suggestion
```

src/utils/format_stackdriver/format_stackdriver.c
```diff
@@ -134,16 +134,24 @@ static int format_typed_value(yajl_gen gen, metric_t const *m,
 
   switch (m->family->type) {
   case METRIC_TYPE_GAUGE: {
-    int status = json_string(gen, "doubleValue");
-    if (status != 0)
+    int status = json_string(gen, "doubleValue") ||
+                 (int)yajl_gen_double(gen, m->value.gauge);
+    if (status != 0) {
```
eero-t, line 139: Although system `yajl_gen.h` header seems to declare `yajl_gen_status_ok` also as zero, IMHO this change does _not_ improve the code.

Additionally, this function seems to leak `yajl_gen_map` on errors => set error value and break from switch, only then return error value?

src/utils/value_list/value_list.h
```diff
@@ -30,7 +30,17 @@
 #define UTILS_VALUE_LIST_H 1
 
 #include "daemon/data_set.h"
-#include "daemon/plugin.h"
+#include "daemon/metric.h" // for value_t
```
eero-t, line 33:
```suggestion
#include "daemon/metric.h" // for value_t + metric types
```

eero-t (commented):
Noticed leak in `format_stackdriver.c` error cases, other comments are cosmetic.

src/utils/format_open_telemetry/format_open_telemetry.cc
octo: Done.

src/utils/format_stackdriver/format_stackdriver.c
octo: This is how the rest of the file avoids to check each call for error individually.

What do you think about the `ERR_COMBINE()` macro in `src/utils/format_influxdb/format_influxdb.c`? That would be a possible alternative.

src/utils/value_list/value_list.h
octo: I dropped the comment.

src/utils/format_stackdriver/format_stackdriver.c
octo: `yajl_gen_map_open` basically pushes a `{` into the `yajl_gen` generator, which is freed later. So there is not leak. But …

… the early return leaves the generator in an unknown and likely broken state, such as an opened map that we will never close. This happens not just here, but all throughout the file. I just pushed a change that will reset the generator state in case of an error so that we're back in a well defined state.

octo: Moved this into  #4268, because we need to cherry-pick this into 5.12 / main also.

src/daemon/utils_cache.c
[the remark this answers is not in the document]
eero-t: Removing code is best fix! :-)

src/utils/format_stackdriver/format_stackdriver.c
eero-t: > This is how the rest of the file avoids to check each call for error individually.

Why the variable is `int`?  `||` operator reduces result to boolean value, i.e. looses the actual integer return value.

> What do you think about the ERR_COMBINE() macro in src/utils/format_influxdb/format_influxdb.c? That would be a possible alternative.

As that macro is only given ints, it could be function instead (`static inline` one in a header), e.g:
```
/* return non-zero one of the given two return values */
static int return_nonzero(int ret1, int ret2);
/* or */
static int nonzero_error(int ret1, int ret2);
```

octo: I removed the `int` cast.

Using a function does result in the same behavior: With the macro, the second argument is evaluated lazily, i.e. the function is only called if `!err`. With the `nonzero_error` function above, both arguments need to be evaluated before the function is called.

eero-t: > I removed the int cast.

My point was that result of `int || int` = `bool`, not `int`:
```
$ cat test.c 
#include <stdio.h>
int main() {
	int a = 2, b = 3;
	printf("%d || %d = %d\n", a, b, a||b);
	return 0;
}
$ gcc -O test.c -o test && ./test
2 || 3 = 1
```

> With the macro, the second argument is evaluated lazily, i.e. the function is only called if !err. With the nonzero_error function above, both arguments need to be evaluated before the function is called.

Macro is OK.  [1]

----

[1] I thought that function to be small enough that GCC could inline it, but but that does not seem to be the case:
```
$ cat test.c 
#include <stdio.h>
int print1(char *s) { return printf("1: %s\n", s); }
int print2(char *s) { return printf("2: %s\n", s); }
int print3(char *s) { return printf("3: %s\n", s); }
static inline int nonzero_error(int ret1, int ret2) { return ret1 || ret2; }
int main() {
	return nonzero_error(print1("always") || print2("direct-OR"), print3("indirect-OR"));
}
$ gcc -O3 -Wall test.c -o test && ./test
3: indirect-OR
1: always
```
Quick googling did not answer it, but maybe C evaluation rules require all arguments to be evaluated _even_ when compiler has full visibility to the called function?

octo: > Quick googling did not answer it, but maybe C evaluation rules require all arguments to be evaluated _even_ when compiler has full visibility to the called function?

Correct.

src/utils/format_stackdriver/format_stackdriver.c
```diff
@@ -134,16 +134,23 @@ static int format_typed_value(yajl_gen gen, metric_t const *m,
 
   switch (m->family->type) {
   case METRIC_TYPE_GAUGE: {
-    int status = json_string(gen, "doubleValue");
-    if (status != 0)
+    int status =
+        json_string(gen, "doubleValue") || yajl_gen_double(gen, m->value.gauge);
```
eero-t, line 138: Because `||` operation results in `status` value being always a `bool`, I think in all instances of this pattern, it could actually declared as `bool` to help compiler better catch mistakes:
```suggestion
    bool status =
        json_string(gen, "doubleValue") || yajl_gen_double(gen, m->value.gauge);
```

Other alternative would be using something like the `ERR_COMBINE()` macro to preserve also other return values than 0/1.

This is large enough change that it should be in another PR though.

eero-t (approved):
Approved.

PR #4268, opened 2024-02-01
[collectd 6] format_stackdriver: When JSON generation fails, reset the generator …

octo:
… because it is in an unknown state.

Whenever the JSON generation fails, the code returns early, leaving the generator in an unknown and likely broken state, such as an opened map that we will never close. This happens all throughout the file. This change will reset the generator state in case of an error so that we're back in a well defined state.

Additionally, a free/alloc cycle was replaced with clear+reset, which are much cheaper operations.

ChangeLog: format_stackdriver: An error path that could leave the JSON generator in an undefined state has been fixed.

### DIFF
--- a/src/utils/format_stackdriver/format_stackdriver.c
+++ b/src/utils/format_stackdriver/format_stackdriver.c
@@ -447,6 +447,15 @@ static void sd_output_reset_staged(sd_output_t *out) /* {{{ */
     sfree(key);
 } /* }}} void sd_output_reset_staged */
 
+static void reset(sd_output_t *out) {
+  sd_output_reset_staged(out);
+
+  yajl_gen_clear(out->gen);       /* empty generator buffer */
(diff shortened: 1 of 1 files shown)


eero-t (approved):
Approved.

In addition to calling the new reset function when create fails, `out->gen` free+alloc is now replaced with clear+reset, I guess on assumption that it's faster?

octo:
> * `sd_output_initialize()` can fail and return error without closing map, but its callers do not even check its return value

Good call: #4270 (and #4269)

> * `sd_format_metric_descriptor()` calls internal `format_*()` functions (that open maps, and return on errors without closing them).  When an error is returned, `sd_format_metric_descriptor()` calls yajl_gen_free(gen), is that enough?

Yes, that's fine. That function does not use the generator state in `sd_output_t` that is shared across calls, but uses a generator that is allocated for this one function only.

PR #4269, opened 2024-02-01
format_stackdriver: When JSON generation fails, reset the generator b…

octo:
… because it is in an unknown state.

Whenever the JSON generation fails, the code returns early, leaving the generator in an unknown and likely broken state, such as an opened map that we will never close. This happens all throughout the file. This change will reset the generator state in case of an error so that we're back in a well defined state.

Additionally, a free/alloc cycle was replaced with clear+reset, which are much cheaper operations.

This is a cherry-pick of #4268 for collectd 5.12.

ChangeLog: format_stackdriver: An error path that could leave the JSON generator in an undefined state has been fixed.

### DIFF
--- a/src/utils/format_stackdriver/format_stackdriver.c
+++ b/src/utils/format_stackdriver/format_stackdriver.c
@@ -623,15 +623,32 @@ int sd_output_register_metric(sd_output_t *out, data_set_t const *ds,
 
 char *sd_output_reset(sd_output_t *out) /* {{{ */
 {
-  sd_output_finalize(out);
+  int err = sd_output_finalize(out);
+  if (err) {
+    goto handle_err;
+  }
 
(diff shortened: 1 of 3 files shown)


eero-t:
I noticed some other error paths that might also need handling and commented on them in https://github.com/collectd/collectd/pull/4268.  I'll review this after that has been clarified (i.e. should this cherry-pick further commits).

eero-t:
`virt` plugin tests failed for `debian_default_toolchain container:collectd/ci:focal_amd64`?

octo:
Also cherry-picked #4270

eero-t (approved):
Approved.

(`sd_output_reset()` could also have been split to static one returning error, and public one handling the returned error. Needing `errno` & return string separately would not have made that noticeably simpler though.)

PR #4272, opened 2024-02-03
[collectd 6] Scale time metrics to seconds per second using floating point counters.

octo:
This converts the following metrics to seconds per second, in order to align with OpenTelemetry Semantic Conventions:

* `system.cpu.time` ([link](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemcputime))
* `system.disk.operation_time` ([link](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemdiskoperation_time))
* `system.disk.io_time` ([link](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemdiskio_time))
* `system.disk.weighted_io_time`

The CPU plugin relies on `rate_to_value()` and `value_to_rate()`, which were lacking support for floating point counters still. Apparently I missed them in #4266. This PR fixes them.


src/utils/common/common_test.c
```diff
@@ -299,53 +299,285 @@ DEF_TEST(strunescape) {
...
+          .v0 = {.counter = 4294967238ULL},
+          .v1 = {.counter = 42},
+          .want = 10.0,
+      },
+      {
+          .name = "counter_t 64bit wrap-around",
+          .t0 = 30,
+          .t1 = 40,
+          .type = METRIC_TYPE_COUNTER,
+          .v0 = {.counter = 18446744073709551558ULL},
```
eero-t, line 533: Wrap-around initial values would IMHO be clearer as `0xffffffff-57` and `0xffffffffffffffff-57`.

eero-t (commented):
Looks nice, improves the plugin code.

PR #4274, opened 2024-02-06
.cirrus.yml: Add task for Mac OS X, cleanup and improvements.

octo:
* Add a new task for building on Mac OS X.
* Disable `CIRRUS_CLONE_SUBMODULES` and instead call `git submodules` as part of the script. This avoids cloding the gnulib repository, reducing build times by ~4 minutes.
* Fix the clang build by adding `-gdwarf-4` to the CFLAGS.
* Remove the "default toolchain" tasks. Debian, Ubuntu, and Fedora are sufficiently tested by the GitHub Actions. On CirrusCI these tasks congest the available worker pool.


.cirrus.yml
```diff
@@ -1,192 +1,61 @@
...
-    - gcc --version
-    - ./configure CC=gcc $DEFAULT_CONFIG_OPTS CFLAGS="$(rpm --eval '%optflags')"
+      ./configure CC=clang CXX=clang++
+      $DEFAULT_CONFIG_OPTS
+      CFLAGS='-gdwarf-4 -Wall
+      -Wno-error
+      -Wextra
+      -Wformat=2
+      -Wformat-security
+      -Wformat-nonliteral
```
eero-t, line 32: At least with GCC, latter 2 are implied by the first one (according to `man gcc`).

.cirrus.yml
```diff
@@ -1,192 +1,61 @@
...
+      -Wmissing-include-dirs
+      -Wold-style-definition
+      -Wpointer-arith
+      -Winit-self
+      -Wmissing-prototypes
+      -Wimplicit-function-declaration
```
eero-t, line 38: Implied by `-Wall`  at least with GCC.

.cirrus.yml
```diff
@@ -1,192 +1,61 @@
...
-      only_if: $CIRRUS_BRANCH == 'main'
-      env:
-        CONTAINER_IMAGE: fedora_main
-  configure_script:
-    - ./build.sh
-    - gcc --version
-    - ./configure CC=gcc $DEFAULT_CONFIG_OPTS CFLAGS="$(rpm --eval '%optflags')"
+      ./configure CC=clang CXX=clang++
+      $DEFAULT_CONFIG_OPTS
+      CFLAGS='-gdwarf-4 -Wall
```
eero-t, line 27: With GCC, analysis is done along with optimization passes.  I.e. for better warning analysis, optimizations would need to be enabled.  Is that not the case with Clang?

PS.  it would be nice for warning flags to be in sorted order.

eero-t (commented):
This is already merged, but I thought to comment a bit on the used stricter warning options.

PR #4276, opened 2024-02-07
ras plugin: Remove unnecessary includes.

octo:
ChangeLog: ras plugin: Unnecessary includes have been removed, improving portability.


eero-t (commented):
Can't mark this as approved afterwards, but it looks fine.

(Logging prefix could IMHO have been `RAS_PLUGIN` though).

PR #4277, opened 2024-02-07
[collectd 6] cpufreq plugin: migration to collectd 6.

octo:
ChangeLog: cpufreq plugin: Migration to v6.0.

### DIFF
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -1,6 +1,7 @@
 /**
  * collectd - src/cpufreq.c
  * Copyright (C) 2005-2007  Peter Holik
+ * Copyright (C) 2024       Florian Forster
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the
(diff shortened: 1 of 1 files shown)


eero-t:
Time-in-state stats are not enabled by default in kernel, and not available for CPUs which frequency / power management is done by firmware (instead of kernel being in control of changing the frequencies by itself), so dropping that info is OKish.

However, Changelog entry should mention that time-in-state info is dropped, as that's the actually useful information, whereas scaling frequency value is just a sample of a value that can fluctuate between its min & max values at few millisecond intervals...

octo:
Reworded the ChangeLog entry. I agree that the time-in-state info is much more meaningful than the point-in-time information. I'll see if I can find the cpufreq-stats kernel module on my development machine.

eero-t:
For unit tests, you could change plugin code to take prefix for the sysfs stats directory, and point the code to a directory with fake time-in-state files (which content could be e.g. copied from the linked kernel doc).

Those files could be either generated at run-time to randomized dir, or pre-created e.g. under `src/test-files/cpufreq/{stats1,stats2,invalid}/`.

PR #4278, opened 2024-02-07
[collectd 6] swap plugin: Add the `system.paging.faults` metric.

octo:
ChangeLog: swap plugin: The `system.paging.faults` metric has been added.


eero-t (approved):
Looks OK.

Issue #4237, opened 2024-01-17
memory plugin: Metrics to not sum up to a stable physical memory size.


octo:
*   Version of collectd: since  #3916
*   Operating system / distribution: Linux
*   Kernel version (if applicable): dunno

## Expected behavior

The sum of the `memory/memory-*` metrics should be total amount of physical memory, i.e. the value should not fluctuate.

## Actual behavior

The "available" and "slab" metrics count memory that is already accounted for by other metrics. The "shared" memory metric is missing.

## Steps to reproduce

See graphs in #3916. Some more information is available in #4224.

octo:
I just recently ran into this issue as part of #4224 and ended up removing the "available" and "slab" metrics for collectd 6.

The idea is that all the `memory/memory-*` metrics can be grouped in one graph, and these two metrics were violating that assumption. After reading [the kernel sources](https://kernel.googlesource.com/pub/scm/linux/kernel/git/tip/tip/+/refs/heads/x86/urgent/fs/proc/meminfo.c) I ended up adding the "shared" metric, because it also originates in the `struct sysinfo` in the kernel.

eero-t:
Grouping "random" memory values reported by OS to get "total" is bogus, as some of them can overlap (like shared), and you can miss multiple relevant types of memory (for example system RAM used by DRM subsystem for GPU allocations, which could in some situations be fairly huge amount).

OS reports memory total separately, and _that_ is what should be reported as total, _and_ used for ratio calculations.

eero-t:
By (correctly) using OS reported value for total memory amount, slap & available values could be brought back.

(They are somewhat useful and removing them is bound to disrupt somebody's workflow: https://m.xkcd.com/1172/)

eero-t:
> By (correctly) using OS reported value for total memory amount, slap & available values could be brought back.

Btw. Available is defined in OTEL spec: https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemlinuxmemoryavailable

zzzyhtheonly:
Hi, 
I'm thinking of adding "Dirty" and "Writeback" (personally use, also willing to send a PR then) in Linux for “memory” plugin, going through I find the same issue in #4224.  

I'm trying to resolve this issue by following the two ways @octo says, just come up with some thoughts, not sure understand the whole thing correctly :-)

> * Put both, "available" and "slab" behind a flag, if they're not already.

[...]

PR #4279, opened 2024-02-08
memory.c: workaround to let ValuesPercentage works again without "available"

zzzyhtheonly:
This PR is trying to fix issue #4237 by splitting "available" off from counting the "sum". 

We use flags with names to set all metrics with a `plugin_instance`, the logic is to let only metrics with the same `plugin_instance` should aggregate to count sum by `MEMORY_SUBMIT`.  Except for metrics `plugin_instance` is "OTHERS" (we add a function `memory_submit_single ` to deal with this situation).

With this commit, we can add new metrics easily. For instance, if we want to add a single metric like "Dirty", we can treat it the same way as "available" in this commit. And if we want to add a subset of metrics like "Vmalloc\*", we can add a new flag in `plugin_instance_flags` and then use `MEMORY_SUBMIT(plugin_instance_names[NEW_FLAG], ...`, it is also convenient.

Tests:

```
$cat log

[{"values":[40.0481980524077],"dstypes":["gauge"],"dsnames":["value"],"time":1707375884.824,"interval":5.000,"host":"cpubox1","plugin":"memory","plugin_instance":"basic","type":"percent","type_instance":"cached"}]
write_log values:
[{"values":[2.16065039544852],"dstypes":["gauge"],"dsnames":["value"],"time":1707375884.824,"interval":5.000,"host":"cpubox1","plugin":"memory","plugin_instance":"basic","type":"percent","type_instance":"used"}]
write_log values:
[{"values":[0.302166379016608],"dstypes":["gauge"],"dsnames":["value"],"time":1707375884.824,"interval":5.000,"host":"cpubox1","plugin":"memory","plugin_instance":"basic","type":"percent","type_instance":"slab_unrecl"}]
[{"values":[54.713116822235],"dstypes":["gauge"],"dsnames":["value"],"time":1707375884.824,"interval":5.000,"host":"cpubox1","plugin":"memory","plugin_instance":"basic","type":"percent","type_instance":"free"}]
write_log values:
[{"values":[1.15392490431425],"dstypes":["gauge"],"dsnames":["value"],"time":1707375884.824,"interval":5.000,"host":"cpubox1","plugin":"memory","plugin_instance":"basic","type":"percent","type_instance":"buffered"}]
write_log values:
[{"values":[1.62194344657793],"dstypes":["gauge"],"dsnames":["value"],"time":1707375884.824,"interval":5.000,"host":"cpubox1","plugin":"memory","plugin_instance":"basic","type":"percent","type_instance":"slab_recl"}]
write_log values:
[{"values":[91406610432],"dstypes":["gauge"],"dsnames":["value"],"time":1707375884.824,"interval":5.000,"host":"cpubox1","plugin":"memory","plugin_instance":"others","type":"memory","type_instance":"available"}]

$ sum=0;for item in \`cat log | grep -vE "available|write_log" | awk -F ":" '{print $2}' | cut -d "," -f 1 | sed -r "s/\[|\]//g" | xargs\`;do sum=\`bc <<< $item+$sum\`;done;echo $sum
100.000000000000008
```

ChangeLog: memory plugin: split "available" from counting on the "sum" of other metrics and add a `plugin_instance` for all metrics

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -185,6 +185,21 @@ static int memory_init(void) {
       plugin_dispatch_multivalue(vl, true, DS_TYPE_GAUGE, __VA_ARGS__, NULL);  \
   } while (0)
 
+#if KERNEL_LINUX
+static void memory_submit_available(gauge_t value) {
+  value_list_t vl = VALUE_LIST_INIT;
+
+  vl.values = &(value_t){.gauge = value};
+  vl.values_len = 1;
+
+  sstrncpy(vl.plugin, "memory", sizeof(vl.plugin));
+  sstrncpy(vl.type, "memory", sizeof(vl.type));
+  sstrncpy(vl.type_instance, "available", sizeof(vl.type_instance));
+
+  plugin_dispatch_values(&vl);
+}
+#endif
+
 static int memory_read_internal(value_list_t *vl) {
 #if HAVE_HOST_STATISTICS
   kern_return_t status;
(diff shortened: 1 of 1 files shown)


zzzyhtheonly:
Hi @octo  thanks for the reviewing. I'm out of office and will move on this in about a week :)

octo (approved):
Much better! Thanks @zzzyhtheonly!

PR #4281, opened 2024-02-18
epics: Fix potential deadlock in handle_label_event

matwey:
ChangeLog: epics plugin: Fix potential deadlock in the plugin

### DIFF
--- a/src/epics.c
+++ b/src/epics.c
@@ -179,8 +179,7 @@ static void handle_label_event(struct pv *p, evargs args) {
   p->label = strdup(value);
   if (p->label == NULL) {
     ERROR("epics plugin: Cannot allocate memory for \"%s\" value", p->name);
-
-    return;
+    // fall-through
   }
 
   pthread_mutex_unlock(&epics_plugin.lock);


src/epics.c
```diff
@@ -180,9 +180,10 @@ static void handle_label_event(struct pv *p, evargs args) {
   if (p->label == NULL) {
     ERROR("epics plugin: Cannot allocate memory for \"%s\" value", p->name);
 
-    return;
+    goto error;
   }
 
+error:
```
eero-t, line 186: Return can be just removed.
```suggestion
  }

```

PS. The error message could also state that it was label event handling that failed.

octo: I think there are two options here:

1. Ensure that the error path ends in a `return` statement.
   Pro: simple and consistent control flow

   ```c
   if (p->label == NULL) {
     ERROR("epics plugin: Cannot allocate memory for \"%s\" value", p->name);
     pthread_mutex_unlock(&epics_plugin.lock);
     return;
   }

   pthread_mutex_unlock(&epics_plugin.lock);
   // end of function
   ```
1. Log the error and continue, since we're at the end of the function. Add a comment to document that the missing `return` is intentional. This is what @eero-t suggests.
   Pro: fewer lines of code

   ```c
   if (p->label == NULL) {
     ERROR("epics plugin: Cannot allocate memory for \"%s\" value", p->name);
     // fall-through
   }

   pthread_mutex_unlock(&epics_plugin.lock);
   // end of function
   ```

matwey: I've redone in the second way since I really do like single lock / single unlock in function code, otherwise it is too easy to miss something.

octo (approved):
Thanks @matwey!

src/epics.c
octo: That's also a good point. Thanks!

PR #4286, opened 2024-02-20
[collectd 6] Enable `-Wstrict-prototypes`.

octo:
In pre-ANSI K&R-style C, functions can use an "initializer list" instead of a "parameter type list". Compilers still support this for backwards compatibility, unfortunately.

One result is that compilers will accept the following prototype:

```c
void foo();
```

Note that this probabaly does not mean what you think it means: this is a function with *an unknown number of arguments* and compilers will happily let you pass anything to it. In other words, the following would be legal:

```c
foo("Hello, World!");
```

This PR changes all (unintentional) uses of empty initializer lists to use `(void)` instead, which expresses what is intended: "this function takes no arguments". It also adds the `-Wstrict-prototypes` CFLAG to prevent backsliding.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,15 +35,19 @@ jobs:
           - fedora39
           - fedora38
         configure_flags: ['--enable-compatibility-mode']
+        cflags: ['-Wstrict-prototypes']
         include:
           - container_tag: debian12
             configure_flags: '--enable-compatibility-mode --enable-debug'
+            cflags: '-Wstrict-prototypes'
           - container_tag: debian12
+            configure_flags: '--enable-compatibility-mode CC=clang CXX=clang++'
             # By default clang emits DWARF v5, which Valgrind cannot read yet.
             # https://github.com/llvm/llvm-project/issues/56550
-            configure_flags: '--enable-compatibility-mode CC=clang CXX=clang++ CFLAGS=-gdwarf-4'
+            cflags: '-gdwarf-4 -Wstrict-prototypes'
     env:
       CONFIGURE_FLAGS: ${{ matrix.configure_flags }}
+      CFLAGS: ${{ matrix.cflags }}
       # this env var picked up by valgrind during make check phase
       VALGRIND_OPTS: "--errors-for-leak-kinds=definite"
     steps:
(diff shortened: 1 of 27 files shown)


eero-t:
> Note that this probabaly does not mean what you think it means: this is a function with _an unknown number of arguments_ and compilers will happily let you pass anything to it.

FYI: I filed GCC bug 13 years ago about that not giving a warning (even when large number of generic GCC warning options are enabled), and it's still open: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=48091

eero-t (approved):
Approved, but both of these failures happen on multiple CI checks:
* configure: error: `strtok_r` is not available. Please file a bugreport!
* expected plugins are MISSING: `bind`

First error happens on older OSes, latter one on newer ones.

PS. Please apply similar compiler option fix also to `main`!

eero-t:
```
src/bind.c:420:15: error: implicit declaration of function 'strptime'; did you mean 'strftime'? [-Werror=implicit-function-declaration]
  420 |   char *tmp = strptime(str_ptr, "%Y-%m-%dT%T", &tm);
      |               ^~~~~~~~
      |               strftime
src/bind.c:420:15: error: initialization of 'char *' from 'int' makes pointer from integer without a cast [-Werror=int-conversion]
```

octo:
The problem is that the configure script relies heavily on K&R style functions, for example:

```
configure:13281: checking for dlopen
configure:13281: gcc -o conftest -O0 -g -Wstrict-prototypes   conftest.c  >&5
conftest.c:63:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
   63 | char dlopen ();
      | ^~~~
configure:13281: $? = 0
configure:13281: result: yes
```

.github/workflows/build.yml
```diff
@@ -64,12 +68,14 @@ jobs:
       with:
         name: ${{ matrix.container_tag }}
         path: config.log
+    # The configure script relies on the old syntax to check whether functions
+    # exist, so we can't specify `-Wstrict-prototypes` globally.
```
eero-t, line 72: Maybe better place for this comment is in the configure step:
```suggestion
    # Configure scripts rely on the K&R syntax to check whether functions
    # exist, so `-Wstrict-prototypes` can be specified only for make.
```

eero-t:
> It is absolutely incomprehensible to me why `-Wall` doesn't include `-Wstrict-prototypes`.

Especially when it's not enabled even if one adds e.g. `-Wextra -Wpedantic -std=c99`...

PR #4288, opened 2024-02-25
ci: Trigger a website update after uploading a release artifact.

octo:


.github/workflows/release.yml
```diff
@@ -33,3 +33,13 @@ jobs:
         [[ -e "${GITHUB_REF_NAME:?}.tar.bz2" ]] && gh release upload "${GITHUB_REF_NAME:?}" "${GITHUB_REF_NAME:?}.tar.bz2"
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Trigger website update
+      run: |
+        if [[ -e "${GITHUB_REF_NAME:?}.tar.bz2" ]]; then
+          declare -r VERSION="${GITHUB_REF_NAME#collectd-}"
+          declare -r DOWNLOAD_URL="$(gh release view "${GITHUB_REF_NAME}" -R collectd/collectd --json assets --jq '.assets[0].url')"
+          declare -r SHA256SUM="$(sha256sum "${GITHUB_REF_NAME}.tar.bz2" | awk '{ print $1; }')"
+          gh workflow run release.yml -R collectd/collectd.github.io --ref main -F "version=${VERSION}" -F "download_url=${DOWNLOAD_URL}" -F "sha256sum=${SHA256SUM}"
```
eero-t, line 42: I'm not familiar with `gh` command, but this has few odd things:
* Recursion: command in `release.yaml` runs something from `release.yaml`?
* As this is v6 branch update, shouldn't it refer to `collecd-6.0` instead of `main`?

(I'm not that familiar with Bash specific syntax either, but the variable assignments / use seems logical.)

octo: This triggers a workflow in another repository.

Not the `-R` argument: it points to the https://github.com/collectd/collectd.github.io/ repository. Since the homepage is always served from `main`, that's basically the only branch that repository has.

* Recursion: no, that `release.yaml` is in a different repo.
* Branch: no, website doesn't have a `collectd-6.0` branch.

`declare -r` might be Bash specific. It declares a variable as "read-only". It's an old habbit to define upper case variables as "constants". It doesn't really make a difference here.

eero-t (approved):
Approved with the explanations.

.github/workflows/release.yml
eero-t: Btw. by Bash syntax,  I was referring not just to `declare`, but also to `[[` and `:?`.

(I guess workflows specifically use Bash as shell, instead of some non-interactive, POSIX one.)

PR #4290, opened 2024-02-26
[collectd 6] ci: Update list of expected plugins.

octo:


eero-t:
"Debian unstable": `FAIL: test_plugin_virt`

"Fedora main" is not configured with some of the added plugins:
```
egrep: warning: egrep is obsolescent; using grep -E
## The following expected plugins are MISSING:
 *  dpdkevents
 *  dpdkstat
 *  virt
```

PR #4291, opened 2024-02-27
[collectd 6] gpu_sysman: add PSU and fan metrics support

eero-t:
ChangeLog: gpu_sysman: add PSU and fan metrics support

I'd rather get these (old commits) in first, and change the names of the new metrics to OTEL naming conventions later as part of PR #4267.

(As name changes have impact also to things outside collectd, and I have not done those other changes yet, due to some of the OTEL name changes still not being final, and relevant parts of the OTEL spec still being in experimental stage...)

### DIFF
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -805,6 +805,7 @@
 #   DisableMemory false
 #   DisableMemoryBandwidth false
 #   DisableFabric false
+#   DisableFan false
 #   DisableFrequency false
 #   DisableThrottleTime false
 #   DisableTemperature false
--- a/src/collectd.conf.in
+++ b/src/collectd.conf.in
@@ -809,6 +809,7 @@
 #   DisableThrottleTime false
 #   DisableTemperature false
 #   DisablePower false
+#   DisablePSU false
 #   DisableEngine false
 #   DisableEngineSingle false
 #   DisableErrors false
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -3835,6 +3835,10 @@ Disable frequency throttling time metric collection.
 
 Disable power usage metrics collection.
 
+=item B<DisablePSU>
+
+Disable power supply related metrics collection.
+
 =item B<DisableTemperature>
 
 Disable temperature metrics collection.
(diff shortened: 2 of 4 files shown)


eero-t:
@octo Debian unstable: `FAIL: test_plugin_virt`?

eero-t:
@mrunge could you update CI tests?
* Most of them fail due to:
`Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/`
* BSD test fails to: `The resource 'projects/freebsd-org-cloud-dev/global/images/family/freebsd-13-2' was not found`
   * Should be updated to 13.4 (still supported), preferably to 14.2 (current)
* Tested Fedora releases (38+39) are not supported any more => replace with 40+41
* Tested Ubuntu 23.10 is not supported any more => replace with 24.04

eero-t:
Hm.  Both of these produce same result, but CI does not accept it:
* `contrib/format.sh src/gpu_sysman.c`
* `clang-format -i src/gpu_sysman.c` (with Clang v18.1)

eero-t:
Marked this as ready for review after rebasing it to latest upstream.

Fan metrics support has been in upstream kernel since 6.12, and current Intel L0 backend vesions support it.

AFAIK PSU metrics are still supported only in L0 frontend, there's no backend + upstream kernel support yet.  They could be split to a separate PR if needed.

mrunge:
Hi @eero-t as you may have witnessed, I have little to no time to work on collectd unfortunately. Let me see what I can do here, it may take some time though.

eero-t:
Rebased to latest upstream to check current CI status.  `centos9` and `debian_unstable` checks still fail due to unrelated issue (`dpdkevents.c` ignoring return value), and latter also due to `virt.c` compilation issue.

AFAIK there still is nothing providing PSU metrics, but new enough kernels to provide fan metrics are nowadays available even as LTS distro HW-enabling packages.

eero-t:
Dropped PSU metrics as there's no L0 backend support for them yet.

PR #4292, opened 2024-02-29
memory.c: add "Dirty" and "Writeback" with conf option

zzzyhtheonly:
"Dirty" and "Writeback" help debug IO delay issues, e.g.,  when we see io-await getting high, with the help of those two metrics it becomes easier to learn if our IOs are stuck on flushing Dirty pages. 
So it might be useful to let Collectd include those two metrics. 
 
Changelog: add "Dirty" and "Writeback" with conf option in the memory plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Features
       Various sensors in the Aquaero 5 water cooling board made by Aquacomputer.
 
     - ascent
-      Statistics about Ascent, a free server for the game `World of Warcraft'.
+      Statistics about Ascent, a free server for the game "World of Warcraft".
 
     - barometer
       Reads absolute barometric pressure, air pressure reduced to sea level and
@@ -55,7 +55,7 @@ Features
       batteries.
 
     - bind
-      Name server and resolver statistics from the `statistics-channel'
+      Name server and resolver statistics from the `statistics-channel`
       interface of BIND 9.5, 9,6 and later.
 
     - buddyinfo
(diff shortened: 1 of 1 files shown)


eero-t:
Code looks OK.

`ReportDirty` option name seems too specific to a single Linux-specific meminfo metric though. IMHO it should be a more generic one telling user that option is about disk IO / usage related memory metrics.  Any suggestions?

zzzyhtheonly:
> Code looks OK.
> 
> `ReportDirty` option name seems too specific to a single Linux-specific meminfo metric though. IMHO it should be a more generic one telling user that option is about disk IO / usage related memory metrics. Any suggestions?

Change to `ReportMem4IOwrite`.

zzzyhtheonly:
By the way, do not know why `format.sh` check fails...

eero-t:
> By the way, do not know why `format.sh` check fails...

I've seen local `clang-format` to order includes differently form the remote version used by `format.sh`.  Just run the suggested command, and commit the result.

If it is indeed due to different include ordering, see: https://github.com/collectd/collectd/pull/4177/commits/ce8a26c9ad07ecd63aff2296d46aa3e7893727db

zzzyhtheonly:
`format.sh` is resolved by another computer re-run. Guess it caused by the alignment of`    /* #endif KERNEL_LINUX */`

But this time `freebsd` fails :(

eero-t:
Only failing check is for `epics` plugin, nothing to do with this.  Changes are small & fine, only question is about the option name. I'm not completely sure about the new one either, but do not have a good suggestion myself.  @octo?

PR #4293, opened 2024-03-01
[collectd 6] gpu_sysman: switch from zeInit() to zesInit()

eero-t:
ChangeLog: gpu_sysman: switch from zeInit() to zesInit() for Xe KMD support

LevelZero spec v1.5 introduced `zesInit()`, and deprecated `zeInit()`, for initializing Sysman part of L0: https://spec.oneapi.io/level-zero/latest/sysman/PROG.html#initialization

Another motivation for introducing this, is L0 Intel backend supporting the new Xe KMD (for the future GPUs [1]), in addition to earlier i915 KMD (for current Intel GPUs), only when Sysman is initialized with `zesInit()`.

Changing to `zesInit()` requires also dropping use of L0 core API functions, and using only L0 Sysman API ones, which is not an exact match.  Therefore:
* Some of the information logged at plugin init (when GpuInfo option is used) needed to be dropped
* `pci_dev` (PCI device ID) label value needs to be digged from sysfs
* There are also few other minor changes to the logged GPU info

`dev_name` label (matching one used by Intel XPU Manager) is added for user-friendly device name.

(Its name can be changed to OTEL compatible one along with other device specific labels when switching to resources attributes.)

This is on top of another PR, which should be merged before this: #4177.

----

[1] I've tested this with current HW using Xe KMD experimental support for them: https://docs.kernel.org/gpu/rfc/xe.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -163,6 +163,7 @@ AC_CHECK_HEADERS_ONCE([ \
   fs_info.h \
   fshelp.h \
   grp.h \
+  glob.h \
   kstat.h \
   kvm.h \
   libgen.h \
--- a/configure.ac
+++ b/configure.ac
@@ -2111,6 +2111,10 @@ PKG_CHECK_MODULES([LEVEL_ZERO], [level-zero],
   [with_level_zero="no (pkg-config could not find 'level-zero')"]
 )
 
+if ! pkg-config --atleast-version=1.9.4 level-zero; then
+  with_level_zero="no (version $(pkg-config --version level-zero) < 1.9.4)"
+fi
+
 if test "x$with_level_zero" = "xyes"; then
   SAVE_CPPFLAGS="$CPPFLAGS"
   CPPFLAGS="$CPPFLAGS $LEVEL_ZERO_CFLAGS"
@@ -2126,9 +2130,9 @@ if test "x$with_level_zero" = "xyes"; then
   SAVE_LDFLAGS="$LDFLAGS"
   LDFLAGS="$LDFLAGS $LEVEL_ZERO_LIBS"
 
-  AC_CHECK_LIB([ze_loader], zeInit,
+  AC_CHECK_LIB([ze_loader], zesInit,
     [with_level_zero="yes"],
-    [with_level_zero="no (libze_loader symbol 'zeInit' not found)"]
+    [with_level_zero="no (libze_loader symbol 'zesInit' not found)"]
   )
   LDFLAGS="$SAVE_LDFLAGS"
 fi
--- a/src/gpu_sysman.c
+++ b/src/gpu_sysman.c
@@ -447,6 +447,25 @@ static const char *map_mem_type(zes_mem_type_t type) {
   }
 }
 
+/* returns NULL if health is unknown, to avoid adding useless labels */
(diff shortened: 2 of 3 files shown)


eero-t:
L0 dependencies:
* L0 spec v1.5 has been supported by the L0 frontend since v1.9.x (Jan 2023): https://github.com/oneapi-src/level-zero/releases/tag/v1.9.4
* But backend had full support for it only since 23.26.26690.12 (Jul 2023): https://github.com/intel/compute-runtime/issues/650

Distro versions of them:
* Fedora:
  - https://src.fedoraproject.org/rpms/oneapi-level-zero
  - https://src.fedoraproject.org/rpms/intel-compute-runtime
* Ubuntu:
  - https://packages.ubuntu.com/lunar/libze-dev
  - https://packages.ubuntu.com/source/focal/intel-compute-runtime
* Debian
  - https://packages.debian.org/bookworm/libze-dev
  - https://packages.debian.org/source/bookworm/intel-compute-runtime

Debian 12 (Stable / bookworm) versions of both are too old, that's why those CI checks fail.  Trixie (Testing) versions are both new enough though.

All Fedora versions have both new enough frontend & backend, so for Fedora & Debian it would be enough to check for >= 1.9 frontend version.

Ubuntu 23.10 is problematic, it has new enough frontend (1.12), but backend is way too old (22.x).  Only in forthcoming 24.04 LTS, both are new enough.  I.e. in 23.10, L0 backend would return "not-implemented" error when Sysman plugin calls frontend `zesInit()` function.

I think Ubuntu 23.10 can be ignored, as it's soon out of support, and frontend >= 1.9 check will at least allow plugin to be built when it's enabled (although its init would fail with too old backend).

Only way to support older backends would be having fallback intialization using the old L0 core functions, but I'd rather not go there.

@octo, any comments?

eero-t:
CI checks pass now, Debian unstable fail is unrelated:
```
PASS: test_plugin_gpu_sysman
...
FAIL: test_plugin_virt
```

Sysman plugin is not built any more for the included stable Debian versions, so installing of the related L0 packages for them could be dropped from CI configs once this is merged.

(Those packages could be added for Ubuntu 24.04 LTS once it's released, and added to CI test matrix.)

eero-t:
> Only way to support older backends would be having fallback intialization using the old L0 core functions, but I'd rather not go there.

Investigated Level-Zero frontend implementation.  It will return "unitialized" error for `zesInit()` _regardless_ of whether `zesInit()` support is missing from backend, or backend function returned some error (e.g. because there was no GPU).  I.e. plugin _cannot_ differentiate between these cases (and fallback to older initialization, or just log a warning, if newer backend would be needed).

eero-t:
> Simplified sysfs reading code a bit by using read_text_file_contents()

New code had off-by-one check bug which caused above not to do what it's supposed, which is now fixed (I'm not sure how I did not notice that before, maybe I accidentally tested old version?).

@octo The only CI failure is for `test_plugin_virt` i.e. nothing related to this PR, but for some reason summary lists 2 fails, although log shows only the virt plugin one: https://github.com/collectd/collectd/actions/runs/8345888360/job/22841843932?pr=4293 ?

PR #4294, opened 2024-03-07
irq.c: add per CPU data collection with conf option

zzzyhtheonly:
Reasons for doing this:
  1. help user better understand interrupts CPU distribution to debug scheduling issues (especially for Linux CONFIG_IRQ_TIME_ACCOUNTING not set and do not have CPU.interrupt statistics).
  2. learn if there are interrupts balancing issues

Changelog: add per CPU data collection with conf option in Linux for irq Plugin


src/collectd.conf.pod
```diff
@@ -4380,6 +4380,11 @@ few ones. This option enables you to do that: By setting B<IgnoreSelected> to
 I<true> the effect of B<Irq> is inverted: All selected interrupts are ignored
 and all other interrupts are collected.
 
+=item B<ReportCPUnr> B<true>|B<false>
```
octo, line 4383: `ReportCPU` fits in better with other config options and is easier to grasp.

src/irq.c
```diff
@@ -80,6 +91,9 @@ static void irq_submit(const char *irq_name, derive_t value) {
   sstrncpy(vl.type, "irq", sizeof(vl.type));
   sstrncpy(vl.type_instance, irq_name, sizeof(vl.type_instance));
 
+  if (cpu_num != -1)
```
octo, line 94: Please use a constant when dealing with special literals. E.g.:

```c
static int const cpu_sum = -1;
// …
if (cpu_num != cpu_sum) {
// …
irq_submit(irq_name, irq_value, cpu_sum);
```

octo (changes requested):
Hi @zzzyhtheonly, thanks for the PR! Overall this looks good, just two small changes before this can go in.

Best regards,
—octo

PR #4296, opened 2024-03-25
feat: [intel_rdt] Added support for LLC misses monitoring event

aleksinx:
Added llc_miss metric

ChangeLog: intel_rdt: Added support for LLC misses monitoring event

### DIFF
--- a/src/intel_rdt.c
+++ b/src/intel_rdt.c
@@ -47,11 +47,12 @@
 #if PQOS_VERSION >= 40400
 #define RDT_EVENTS                                                             \
   (PQOS_MON_EVENT_L3_OCCUP | PQOS_PERF_EVENT_IPC | PQOS_MON_EVENT_LMEM_BW |    \
-   PQOS_MON_EVENT_TMEM_BW | PQOS_MON_EVENT_RMEM_BW | PQOS_PERF_EVENT_LLC_REF)
+   PQOS_MON_EVENT_TMEM_BW | PQOS_MON_EVENT_RMEM_BW | PQOS_PERF_EVENT_LLC_REF | \
+   PQOS_PERF_EVENT_LLC_MISS)
 #else
 #define RDT_EVENTS                                                             \
   (PQOS_MON_EVENT_L3_OCCUP | PQOS_PERF_EVENT_IPC | PQOS_MON_EVENT_LMEM_BW |    \
-   PQOS_MON_EVENT_TMEM_BW | PQOS_MON_EVENT_RMEM_BW)
+   PQOS_MON_EVENT_TMEM_BW | PQOS_MON_EVENT_RMEM_BW | PQOS_PERF_EVENT_LLC_MISS)
 #endif
 
 #define RDT_MAX_SOCKETS 8
@@ -89,6 +90,7 @@ struct rdt_ctx_s {
 #if PQOS_VERSION >= 40400
   bool mon_llc_ref_enabled;
 #endif
+  bool mon_llc_miss_enabled;
   core_groups_list_t cores;
   enum pqos_mon_event events[RDT_MAX_CORES];
   struct pqos_mon_data *pcgroups[RDT_MAX_CORES];
@@ -163,6 +165,18 @@ static void rdt_submit(const struct pqos_mon_data *group) {
       rdt_submit_gauge(desc, "bytes", "llc_ref", value);
   }
 #endif
+
+  if (events & PQOS_PERF_EVENT_LLC_MISS) {
+#if PQOS_VERSION >= 40400
+    uint64_t value;
+
+    int ret = pqos_mon_get_value(group, PQOS_PERF_EVENT_LLC_MISS, &value, NULL);
+    if (ret == PQOS_RETVAL_OK)
+      rdt_submit_gauge(desc, "bytes", "llc_miss", value);
+#else
+    rdt_submit_gauge(desc, "ipc", NULL, values->llc_misses);
+#endif
+  }
 
   if (events & PQOS_MON_EVENT_LMEM_BW) {
     const struct pqos_monitor *mon = NULL;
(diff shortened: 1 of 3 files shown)


src/intel_rdt.c
```diff
@@ -164,6 +166,18 @@ static void rdt_submit(const struct pqos_mon_data *group) {
...
 #endif
 
+  if (events & PQOS_PERF_EVENT_LLC_MISS) {
+#if PQOS_VERSION >= 40400
+    uint64_t value;
+
+    int ret = pqos_mon_get_value(group, PQOS_PERF_EVENT_LLC_MISS, &value, NULL);
+    if (ret == PQOS_RETVAL_OK)
+      rdt_submit_gauge(desc, "bytes", "llc_miss", value);
+#else
```
eero-t, line 176: Looking at intel-cmt-cat Git history...

Whereas  `pqos_mon_get_value()` was added 2 years ago for in 4.04, same time with LLC refs support:
https://github.com/intel/intel-cmt-cat/commit/6ee8ee7fddce835e3e387a40b027c7d0f34d4462

LLC misses counter support was added alread >8 years ago, for version 1.03:
https://github.com/intel/intel-cmt-cat/commit/53261fa6ce2be876a424e27c95b7e8941b40b98b

As the other code in this function just takes numbers directly from `values` (i.e. bypasses locking used by  `pqos_mon_get_value()`), and delta parameter is not used, what value there's from having this ifdef for calling  `pqos_mon_get_value()`?

src/intel_rdt.c
```diff
@@ -47,11 +47,12 @@
...
   (PQOS_MON_EVENT_L3_OCCUP | PQOS_PERF_EVENT_IPC | PQOS_MON_EVENT_LMEM_BW |    \
-   PQOS_MON_EVENT_TMEM_BW | PQOS_MON_EVENT_RMEM_BW | PQOS_PERF_EVENT_LLC_REF)
+   PQOS_MON_EVENT_TMEM_BW | PQOS_MON_EVENT_RMEM_BW | PQOS_PERF_EVENT_LLC_REF | \
+   PQOS_PERF_EVENT_LLC_MISS)
 #else
 #define RDT_EVENTS                                                             \
   (PQOS_MON_EVENT_L3_OCCUP | PQOS_PERF_EVENT_IPC | PQOS_MON_EVENT_LMEM_BW |    \
-   PQOS_MON_EVENT_TMEM_BW | PQOS_MON_EVENT_RMEM_BW)
+   PQOS_MON_EVENT_TMEM_BW | PQOS_MON_EVENT_RMEM_BW | PQOS_PERF_EVENT_LLC_MISS)
 #endif
```
eero-t, line 56: Would IMHO be easier to read if base flags were set separately, and inside `#if PQOS_VERSION >= 40400` only extra flags added to those. Something like this:
```
#define RDT_EVENTS_BASE                                                        \
  (PQOS_MON_EVENT_L3_OCCUP | PQOS_PERF_EVENT_IPC | PQOS_MON_EVENT_LMEM_BW |    \
   PQOS_MON_EVENT_TMEM_BW | PQOS_MON_EVENT_RMEM_BW | PQOS_PERF_EVENT_LLC_MISS)

#if PQOS_VERSION >= 40400
#define RDT_EVENTS (RDT_EVENTS_BASE | PQOS_PERF_EVENT_LLC_REF)
#else
#define RDT_EVENTS RDT_EVENTS_BASE
#endif
```

src/intel_rdt.c
aleksinx: No update will be required if pqos_mon_data structure will change

eero-t: Update is still needed for the `#else` branch in that case...

eero-t: I was suggesting using just the code in `#else` branch, similarly to other code in here, unless there's compelling reason to do otherwise.

If lock checks done in them is one, it would be better to have **all** the values requested with `pqos_*()` getters inside `#if PQOS_VERSION >= 40400`, and `#else` branch doing the same for all values (that are available before 4.04) by accessing the struct values directly.

src/intel_rdt.c
```diff
@@ -164,6 +166,18 @@ static void rdt_submit(const struct pqos_mon_data *group) {
   }
 #endif
 
+  if (events & PQOS_PERF_EVENT_LLC_MISS) {
+#if PQOS_VERSION >= 40400
+    uint64_t value;
+
+    int ret = pqos_mon_get_value(group, PQOS_PERF_EVENT_LLC_MISS, &value, NULL);
+    if (ret == PQOS_RETVAL_OK)
+      rdt_submit_gauge(desc, "bytes", "llc_miss", value);
```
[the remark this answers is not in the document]
eero-t: Hm.  I've never used collectd with derived/delta metrics, only gauges & counters.    Do collectd write plugins delta values to totals?

Otherwise there's a problem...

Write plugins do not know whether same client queried the value earlier, so they could report the latest available value on every query, meaning that one cannot add them up in case one got same metric multiple times.  Or if plugin reports given metric only once, and multiple clients are querying for them, then some of the delta values are missed...

@octo Any advice on using derive delta vs. total counter values for metrics?

src/intel_rdt.c
```diff
@@ -158,12 +159,24 @@ static void rdt_submit(const struct pqos_mon_data *group) {
...
 
+  if (events & PQOS_PERF_EVENT_LLC_MISS) {
+#if PQOS_VERSION >= 40400
+    uint64_t value;
+
+    int ret = pqos_mon_get_value(group, PQOS_PERF_EVENT_LLC_MISS, NULL, &value);
+    if (ret == PQOS_RETVAL_OK)
+      rdt_submit_derive(desc, "bytes", "llc_miss", value);
+#else
+    rdt_submit_derive(desc, "bytes", "llc_miss", values->llc_misses_delta);
```
eero-t, line 176: I've never used collectd with derived/delta metrics, only gauges & counters. Do collectd write plugins delta values to totals?

Otherwise there could be a problem...

Write plugins do not know whether same client queried the value earlier, so they could report the latest available value on every query, meaning that one cannot add them up in case one got same metric multiple times. Or if plugin reports given metric only once, and multiple clients are querying for them, then some of the delta values are missed, and client cannot construct total/avg for given (longer) time interval.

@octo Any advice on using derive delta vs. total counter values for metrics?

aleksinx: llc_misses is a counter and llc_misses_delta is difference from previous pull. Counter value should be reported as derive

src/intel_rdt.c
aleksinx: done

eero-t (approved):
Thanks!   Approved.

mkobyli (approved):
Thanks, looks good to me!

PR #4299, opened 2024-04-17
[collectd 6] gpu_sysman: use base units + other metric handling fixes

eero-t:
ChangeLog: gpu_sysman: use base units + other metric handling fixes

This split the fixes from #4267 to their own PR (that should go in first).

Changes are IMHO straightforward and easiest to review from individual commits.

(Because switching to base units means metric names changing, `Feature` label could have been used instead of a `Fix` one.)

PR #4300, opened 2024-04-17
[collectd 6] gpu_sysman: change GPU metric labels to resource attributes

eero-t:
ChangeLog: gpu_sysman: change GPU metric labels to resource attributes

Besides changing existing per-GPU metric labels to corresponding resource attributes, this also provides additional GPU attributes recommended by the OTEL spec.


eero-t:
Hm. On the two HWs / kernels I just tested, one did not report any FW info, and the other reported 2 of them.  This PR reports info for just the last FW, as OTEL spec seems to assume there only being single FW => I'll file bug against OTEL spec.

(FW info support is almost half of the changes in PR.)

eero-t:
This is in draft state because it makes querying metrics from Prometheus much more complicated.

Prometheus adds `exported_` prefix to `job` & `instance` labels (which can probably be remapped with `serviceMonitor`), and if one wants any of the resource attributes with the queried metrics, it seems that one needs to:
* specify each of the labels separately in the query
* for each metric

I.e. instead of getting all metrics for given `$NODE` from `$SERVICE` with query like this:
`{node="$NODE", service="$SERVICE"}`

One needs to now query each of them separately with a query like this:
`collectd_gpu_sysman_temperature_celsius{node="$NODE"} * on (exported_instance) group_left($LABELS) target_info`

Which gives items like this (when `LABELS`=`model,pci_bdf,pci_dev`):
```
      {
        "metric": {
          "container": "collectd",
          "endpoint": "http",
          "exported_instance": "0000:0a:00.0",
          "exported_job": "unknown_service:collectd",
          "instance": "10.32.130.125:9103",
          "job": "collectd",
          "location": "gpu-max",
          "model": "Intel(R) Iris(R) Xe MAX Graphics",
          "namespace": "monitoring",
          "node": "testnode",
          "pci_bdf": "0a:00.0",
          "pci_dev": "0x4905",
          "pod": "collectd-d9m6z",
          "service": "collectd"
        },
        "value": [
          1713363392.117,
          "30"
        ]
      }
```

eero-t:
Looking at the OTEL HW spec: https://opentelemetry.io/docs/specs/semconv/system/hardware-metrics/#common-hardware-attributes

It seems that all HW devices should provide also `id` and `name` attrributes, in addition to similar GPU `model` and `serial` attributes?

FYI: I've raised these and the earlier noticed issues in following OTEL spec ticket: https://github.com/open-telemetry/opentelemetry-specification/issues/3999

eero-t:
Updated code according to feedback given in OTEL bug tracker, which dealt with the issues I raised above.

Will keep this still as "Draft" though, until #4299 is merged, the discussion in OTEL ticket has winded down and I've tested this a bit more.

eero-t:
Questions answered in OTEL ticket and code changed to match => dropped Draft status.

Rebased GCC warning WA & FW attribute commits to be last ones, in case they should be dropped from this PR.

eero-t:
No changes, just rebased to latest `collectd-6.0` state.

PR #4303, opened 2024-04-26
notify_telegram plugin

YanikusGG:
This is initial implementation for **notify_telegram** plugin. Later I will create PR for collectd 6 also.

The **notify_telegram** plugin uses _libcurl_ and Telegram Bot API to send notifications to a configured telegram users.
It also can send config instruction with ChatID in response to all users by long polling or webhooks using _libcurl_, _libyajl_, _libmicrohttpd_.

ChangeLog: Introduce notify_telegram plugin

Reference: https://core.telegram.org/bots/api

### DIFF
--- a/src/notify_telegram.c
+++ b/src/notify_telegram.c
@@ -1,6 +1,6 @@
 /**
  * collectd - src/notify_telegram.c
- * Copyright (C) 2023  Yan Anikiev
+ * Copyright (C) 2024  Yan Anikiev
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
@@ -55,8 +55,8 @@
 #define MAX_PARAMS_SIZE 2048
 #define MAX_INPUT_MESSAGES_COUNT 30
 
-#define PARSE_OK 1
-#define PARSE_FAIL 0
+#define YAJL_CB_PARSE_OK 1
+#define YAJL_CB_PARSE_FAIL 0
 
 struct response_buffer_t {
   char *response;
@@ -76,6 +76,19 @@ struct parse_context_t {
   char *chat_id[MAX_INPUT_MESSAGES_COUNT];
 };
 
+static struct plugin_config_t {
+  char *bot_token;
+  char *proxy_url;
+  bool disable_getting_updates;
+  char *webhook_url;
+  char *webhook_host;
+  char *webhook_port;
+  char *mhd_daemon_host;
+  unsigned short mhd_daemon_port;
+  char **recipients;
+  int recipients_len;
+} plugin_config;
+
 #if HAVE_YAJL_V2
 typedef size_t yajl_len_t;
 #else
@@ -88,18 +101,8 @@ static const char *config_keys[] = {
     "MHDDaemonHost", "MHDDaemonPort", "RecipientChatID"};
 static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
 
-static char *bot_token;
-static char *proxy_url;
-static bool disable_getting_updates = false;
-static char *webhook_url;
-static char *webhook_host;
-static char *webhook_port;
-static char *mhd_daemon_host;
-static unsigned short mhd_daemon_port;
-static char **recipients;
-static int recipients_len;
-
-const char *CONFIG_HELP_TEXT_TEMPLATE =
+static const char *DEFAULT_PROXY_URL = "https://api.telegram.org/bot";
+static const char *CONFIG_HELP_TEXT_TEMPLATE =
     "Here is the collectd configuration with your chat id:\n"
     "```\n"
     "<Plugin notify_telegram>\n"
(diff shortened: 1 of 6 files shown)


src/notify_telegram.c
```diff
@@ -0,0 +1,871 @@
...
+      return -1;
+    }
+    recipients_len++;
+  } else if (strcasecmp(key, "BotToken") == 0) {
+    sfree(bot_token);
+    bot_token = strdup(value);
+    if (bot_token == NULL) {
+      ERROR("notify_telegram: strdup failed.");
+      return -1;
+    }
```
eero-t, line 732: Code would be easier to read with a table containing the string and variable (address) pair, and a loop handling all of those.

src/notify_telegram.c
```diff
@@ -0,0 +1,871 @@
...
+  } else if (strcasecmp(key, "WebhookPort") == 0) {
+    sfree(webhook_port);
+    webhook_port = strdup(value);
+    if (webhook_port == NULL) {
+      ERROR("notify_telegram: strdup failed.");
+      return -1;
+    }
+  } else if (strcasecmp(key, "MHDDaemonHost") == 0) {
+#if MHD_VERSION >= 0x00090000
```
eero-t, line 764: Checks for earlier MHD version keys could be done separately from the strdup() loop.  Or the looped table could have member indicating for which MHD version give key is.

src/notify_telegram.c
```diff
@@ -0,0 +1,871 @@
...
+
+static int notify_telegram_notification(const notification_t *n,
+                                        user_data_t __attribute__((unused)) *
+                                            user_data) {
+  char buf[MAX_BUF_SIZE] = "";
+  char *buf_ptr = buf;
+  int buf_len = sizeof(buf);
+  int status;
+
+  status = snprintf(
```
eero-t, line 800: While it's both count and error status, I think here `count` would be more appropriate name.

src/notify_telegram.c
```diff
@@ -0,0 +1,871 @@
...
+  }
+
+#define APPEND(bufptr, buflen, key, value)                                     \
+  if ((buflen > 0) && (strlen(value) > 0)) {                                   \
+    status = snprintf(bufptr, buflen, ",\n%s = %s", key, value);               \
+    if (status > 0) {                                                          \
+      bufptr += status;                                                        \
+      buflen -= status;                                                        \
+    }                                                                          \
+  }
```
eero-t, line 819: Use static function instead.  Compiler can then optimize it as it sees best (inline it like you did here, or not).

src/notify_telegram.c
```diff
@@ -0,0 +1,871 @@
...
+      buf_ptr, buf_len, "<b>Notification:</b>\nseverity = %s",
+      (n->severity == NOTIF_FAILURE)
+          ? "FAILURE"
+          : ((n->severity == NOTIF_WARNING)
+                 ? "WARNING"
+                 : ((n->severity == NOTIF_OKAY) ? "OKAY" : "UNKNOWN")));
+  if (status > 0) {
+    buf_ptr += status;
+    buf_len -= status;
+  }
```
eero-t, line 810: You could assign severity to a separate string variable, use the append function also for this, and add `,\n` to the latter key values.

PS. I wonder though why both `\n` and `,` separators are needed, isn't one of them enough?

eero-t:
> This is initial implementation for **notify_telegram** plugin. Later I will create PR for collectd 6 also.

New plugins should go to `collectd-6.0` branch, only existing plugins are updated in v5.x code (`main`) .

src/notify_telegram.c
```diff
@@ -0,0 +1,871 @@
...
+#endif
+
+static int notify_telegram_init(void) {
+  curl_global_init(CURL_GLOBAL_SSL);
+
+  if (disable_getting_updates) {
+    return 0;
+  }
+
+  if (webhook_host && !httpd) {
```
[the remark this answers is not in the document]
YanikusGG: thank you, I've moved the common parts of the code into separate functions

src/notify_telegram.c
```diff
@@ -0,0 +1,871 @@
...
+}
+
+static int tg_parse_number_callback(void *ctx, const char *number,
+                                    yajl_len_t number_len) {
+  struct parse_context_t *parse_context = (struct parse_context_t *)ctx;
+
+  if (parse_context->inside_key_update_id) {
+    if (parse_context->max_update_id == NULL) {
+      parse_context->max_update_id = strndup(number, number_len);
+      if (parse_context->max_update_id == NULL) {
```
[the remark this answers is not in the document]
YanikusGG: I tried early-exit code style and at this case it got worse (in my opinion). So I want to leave it as it is.

src/notify_telegram.c
YanikusGG: I don't think it's worth it, because not all keys in the config are handled the same way

YanikusGG:
> > This is initial implementation for **notify_telegram** plugin. Later I will create PR for collectd 6 also.
> 
> New plugins should go to `collectd-6.0` branch, only existing plugins are updated in v5.x code (`main`) , but please handle review feedback first.

Thank you for your helpful issues. I've fixed most of them.

src/notify_telegram.c
eero-t: Most keys are, but indeed there would be few that would need separate handling outside of the proposed loop.  I'll resolve this.

src/notify_telegram.c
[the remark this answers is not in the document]
eero-t: Thanks for testing it!

The early-exit change you did for `notify_telegram_read()` IMHO really improved / was important for it, but this function is small enough that if/else ladders are fine => resolved.

eero-t:
Thanks, fixes look good!    Next step is review from Florian (@octo) for the v6 version of the PR.

(He hasn't been active for a while, but hopefully you don't need to wait too long...)

eero-t:
With the v6 PR (https://github.com/collectd/collectd/pull/4311) and new plugins not being accepted to v5, closing this.

Issue #4305, opened 2024-05-05
Build fails on MacOS


jvsg:
*   Version of collectd: `main` branch on commit `ff8cd79667fa379f3088a927e75e989bc996fccc`
*   Operating system / distribution: MacOS
*   Kernel version (if applicable): Darwin 23.1.0 arm64
* Configuration:
```
  Build:
    Platform  . . . . . . Darwin
    Compiler vendor . . . clang
    CC  . . . . . . . . . gcc
    CFLAGS  . . . . . . . -Wall -Werror -g -O2
    CXXFLAGS  . . . . . . -Wall -Werror -g -O2
    CPP . . . . . . . . . gcc -E
    CPPFLAGS  . . . . . .
    GRPC_CPP_PLUGIN . . .
    LD  . . . . . . . . . /Library/Developer/CommandLineTools/usr/bin/ld
    LDFLAGS . . . . . . .
    PROTOC  . . . . . . .
    YACC  . . . . . . . . bison -y
    YFLAGS  . . . . . . .

```
Note: It may seem above that the compiler is gcc, but it actually is an Apple thing where they ship clang by the name of gcc

```
$ gcc --version
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: arm64-apple-darwin23.1.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

## Expected behavior

We expect `make` to do its job successfully.

## Actual behavior

```
src/bind.c:473:7: error: variable 'num_entries' set but not used [-Werror,-Wunused-but-set-variable]
  int num_entries = 0;
      ^
src/bind.c:552:7: error: variable 'num_entries' set but not used [-Werror,-Wunused-but-set-variable]
  int num_entries = 0;
      ^
src/bind.c:612:7: error: variable 'num_entries' set but not used [-Werror,-Wunused-but-set-variable]
  int num_entries = 0;
      ^
```

## Steps to reproduce

*   `./configure`
*   `make`

eero-t:
Odd.  When testing on Linux: `gcc|clang -c -O3 -Isrc/ -Isrc/daemon/ -I/usr/include/libxml2 -Wunused-but-set-variable src/bind.c`

That warning indeed is reported only by Clang (v16.0), not GCC (v13.2).   I thought they were supposed to be comparable in this regard...

PR #4306, opened 2024-05-06
bind: fix warning for variable used only in DEBUG() calls

eero-t:
ChangeLog: bind: fix warning for variable used only in DEBUG() calls

Fixes: https://github.com/collectd/collectd/issues/4305


eero-t:
The only check fail is due to `epics` plugin `#include` failing, nothing to do with changes here.

mrunge (approved):
thank you!

PR #4307, opened 2024-05-08
Clang: Fix build failure

jvsg:
Clang complains about a variable being unused because in the non-debug builds, we don't end up using its value. This commit fixes it.

Addresses: #4305


eero-t:
> Addresses: #4305

As you can see from that bug, it was fixed before you even filed this PR.

Issue #4308, opened 2024-05-08
Perl.c fails to compile with clang


jvsg:
*   Version of collectd: `main` on `ff8cd79667fa379f3088a927e75e989bc996fccc`
*   Operating system / distribution: MacOS with clang

## Expected behavior

`make` should compile cleanly

## Actual behavior

```
src/perl.c:1081:5: error: '(' and '{' tokens introducing statement expression appear in different macro expansion contexts [-Werror,-Wcompound-token-split-by-macro]
    XPUSHs(sv_2mortal(newSVpv(va_arg(ap, char *), 0)));
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Perl/5.30/darwin-thread-multi-2level/CORE/pp.h:462:19: note: expanded from macro 'XPUSHs'
#define XPUSHs(s)       STMT_START { EXTEND(sp,1); *++sp = (s); } STMT_END
                        ^~~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Perl/5.30/darwin-thread-multi-2level/CORE/perl.h:544:29: note: expanded from macro 'STMT_START'
#   define STMT_START   (void)( /* gcc supports "({ STATEMENTS; })" */
                              ^
src/perl.c:1081:5: note: '{' token is here
    XPUSHs(sv_2mortal(newSVpv(va_arg(ap, char *), 0)));
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Perl/5.30/darwin-thread-multi-2level/CORE/pp.h:462:30: note: expanded from macro 'XPUSHs'
#define XPUSHs(s)       STMT_START { EXTEND(sp,1); *++sp = (s); } STMT_END
                                   ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]
```

## Steps to reproduce

*   `./configure`
*   `make`

PR #4298, opened 2024-04-09
fix(epics): Fix include path -- at least on Debuan unstable.

octo:
ChangeLog: Epics plugin: The include path for header files has been adapted to new versions of the *epics-base* package.

PR #4309, opened 2024-05-08
perl plugin: fix Clang compilation by ignoring a warning

jvsg:
ChangeLog: perl plugin: fix Clang compilation by ignoring a warning

Fixes: #4308

### DIFF
--- a/src/perl.c
+++ b/src/perl.c
@@ -33,6 +33,12 @@
 /* do not automatically get the thread specific Perl interpreter */
 #define PERL_NO_GET_CONTEXT
 
+#if defined(__clang__) && defined(__clang_major__)
+#if __clang_major__ >= 12
+#pragma clang diagnostic ignored "-Wcompound-token-split-by-macro"
+#endif
+#endif
+
 #include "collectd.h"
 #include <stdbool.h>
 


eero-t:
Changed title to better describe what is being fixed, and added the required (Fix) label & "ChangeLog" entry.

src/perl.c
```diff
@@ -33,6 +33,10 @@
 /* do not automatically get the thread specific Perl interpreter */
 #define PERL_NO_GET_CONTEXT
 
+#if defined(__APPLE__)
+#pragma clang diagnostic ignored "-Wcompound-token-split-by-macro"
+#endif
```
eero-t, line 38: According to:
* https://reviews.llvm.org/D86751
* https://github.com/Perl/perl5/issues/18780

It's a **new** option that came with Clang-12 around 2021.

Also, Clang is not Apple specific, and one could be using GCC on Apple.

=> I think this should be changed to ignore the diagnostic for Clang >= 12, not for Apple.

eero-t (changes requested):
If the warned macro issue cannot be fixed (I know very little of Perl, so I have no idea how hard that would be), at least the `#if` guard should be fixed to check for correct compiler version.

eero-t:
@mrunge, do you have Linux setup for checking whether Perl plugin has same issue with current Linux Clang / or any other comments on this?

(I don't have any Perl dev stuff on my setup, and I'd rather keep it out of it, as I'm rather a Python user. :-))

eero-t:
Commit title should be also be fixed, to indicate that it's the perl plugin that is being fixed (not clang one).

jvsg:
@eero-t please re-review

eero-t (approved):
Thanks, this looks fine now!  => Triggered checks & approved.

eero-t:
The only failure is unrelated (https://github.com/collectd/collectd/pull/4298).

Before merging, I would a comment from somebody who uses (or at least builds) Perl plugin.  @mrunge?

mrunge:
I am not using the perl plugin, and I'm also not building collectd on a regular basis anymore, so can't add test cases or feedback to this.

mrunge:
@jvsg thank you for your PR, it looks good to me.

PR #4311, opened 2024-05-11
[collectd 6] notify_telegram plugin

YanikusGG:
This is collectd-6.0 implementation for notify_telegram plugin.
PR for collectd-5: https://github.com/collectd/collectd/pull/4303

The notify_telegram plugin uses libcurl and Telegram Bot API to send notifications to a configured telegram users.
It also can send config instruction with ChatID in response to all users, either by long polling, or by using webhooks that utilize libcurl, libyajl, libmicrohttpd.

ChangeLog: Introduce notify_telegram plugin

Reference: https://core.telegram.org/bots/api

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1688,6 +1688,15 @@ notify_nagios_la_SOURCES = src/notify_nagios.c
 notify_nagios_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 endif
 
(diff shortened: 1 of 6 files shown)


eero-t:
Thanks for filing the v6 version of the PR!

I haven't used or written any notify plugins, so next step would be for Florian to review this.

@octo, I checked that this is identical to v5 PR (which fixed all my earlier review comments), and that it builds on all platforms, without warnings.

PR #4317, opened 2024-07-08
feat(mesh): Add a new Wifi Mesh plugin

pticon:
ChangeLog: Mesh plugin: add a new plugin to collect statistics about WiFi Mesh networks.

Collects various metrics related to the WiFi Mesh, such as dropped frames, forwarded frames, established links, etc.

### DIFF
--- a/README
+++ b/README
@@ -261,6 +261,12 @@ Features
       Memory utilization: Memory occupied by running processes, page cache,
       buffer cache and free.
 
+    - mesh
+      Collects various metrics related to Wifi mesh performance (e.g. established
+      peer links, dropped frames, forwarded frames). The statistics are read
+      from the Linux sysfs and require the Linux kernel to be compiled with
+      CONFIG_MAC80211_DEBUGFS=y.
+
     - mic
       Collects CPU usage, memory usage, temperatures and power consumption from
       Intel Many Integrated Core (MIC) CPUs.
--- a/src/collectd.conf.pod
+++ b/src/collectd.conf.pod
@@ -5121,6 +5121,36 @@ setting is given, the B<Address> and B<Port> settings are ignored.
 
 =back
 
+=head2 Plugin C<mesh>
+
+The I<Mesh plugin> collects some statistics about the Wifi Mesh performance.
+The Linux kernel requires to be compile with CONFIG_MAC80211_DEBUGFS=y.
+
+B<Synopsis:>
+
+ <Plugin mesh>
+   <PhysicalDevice "phy0">
+     NetworkDevice "wlan0"
+   </PhysicalDevice>
+   <PhysicalDevice "phy1">
+     NetworkDevice "wlan1"
+   </PhysicalDevice>
+ </Plugin>
+
+=over 4
+
+=item B<PhysicalDevice> I<Name>
+
+Specifies a block containing the physical WiFi device name found under
+I</sys/kernel/debug/ieee80211/>.
+
+=item B<NetworkDevice> I<Name>
+
+Specifies the network interface associated with the WiFi mesh, typically
+found in the system's network configuration.
+
+=back
+
 =head2 Plugin C<mic>
 
 The B<mic plugin> gathers CPU statistics, memory usage and temperatures from
--- a/src/mesh.c
+++ b/src/mesh.c
@@ -1,0 +1,365 @@
+/**
+ * collectd - src/mesh.c
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; only version 2 of the License is applicable.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ *
+ * Authors:
+ *   Pierre Lebleu <pierre.lebleu at pile-engineering.com>
+ */
+
+#include "collectd.h"
+#include <inttypes.h>
+
+#include "plugin.h"
+#include "utils/common/common.h"
+
+#if !KERNEL_LINUX
+#error "No applicable input method."
+#endif
+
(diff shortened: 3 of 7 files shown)


src/mesh.c
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  return mesh_submit(m, DROPPED_FRAMES_TTL, value);
+} /* int mesh_read_dropped_frames_ttl */
+
+static int mesh_read_fwded_frames(const mesh_list_t *m) {
+  int ret;
+  value_t value;
+
+  DEBUG(PLUGIN_NAME ": Reading attribute " FWDED_FRAMES " for device %s",
```
eero-t, line 277: fwded/FWDED -> forwarded/FORWARDED

src/mesh.c
```diff
@@ -0,0 +1,365 @@
...
+    ERROR(PLUGIN_NAME " : Unable to read " FWDED_MCAST);
+    return ret;
+  }
+
+  return mesh_submit(m, FWDED_MCAST, value);
+} /* int mesh_read_fwded_mcast */
+
+typedef int (*mesh_reader_t)(const mesh_list_t *);
+
+static const mesh_reader_t g_mesh_readers[] = {
```
eero-t, line 326: This array is only used inside `mesh_read()`, so you can just as well move it there.

src/mesh.c
```diff
@@ -0,0 +1,365 @@
...
+  values[0] = value;
+
+  vl.values = values;
+  vl.values_len = 1;
+  sstrncpy(vl.plugin, PLUGIN_NAME, sizeof(vl.plugin));
+  sstrncpy(vl.type_instance, m->net_name, sizeof(vl.type_instance));
+  sstrncpy(vl.type, type, sizeof(vl.type));
+
+  return plugin_dispatch_values(&vl);
+} /* int mesh_submit */
```
eero-t, line 141: Plugins using deprecated value lists will be disabled by default and need `--enable-compatibility-mode` option: https://github.com/collectd/collectd/releases/tag/collectd-6.0.0.rc0

Use metric families instead: https://github.com/collectd/collectd/blob/collectd-6.0/src/daemon/metric.h

Example on their use: https://github.com/collectd/collectd/blob/collectd-6.0/src/cpusleep.c

(Benefits are that labels can be added to metrics, common labels can be given as OpenTelemetry resource attributes/labels and better metric types.)

src/types.db
eero-t, line 1: None of this is needed with metrics families.

As to metric naming, please try to follow OpenTelemetry semantics:
https://opentelemetry.io/docs/specs/semconv/system/hardware-metrics/#hwnetwork---network-adapter-metrics

eero-t (changes requested):
Code and plugin integration looks fairly OK, but before merging it needs to move to using non-deprecated metrics types and naming should use OTEL semantics.

src/mesh.c
pticon: Got it.

src/types.db
pticon: The metrics proposed by OpenTelemetry do not have the right name for the propose of this plugin.

What do you think about those ?
* hw.network.peers
* hw.network.packets.dropped.congestion
* hw.network.packets.dropped.no_route
* hw.network.packets.dropped.ttl
* hw.network.packets.forwarded
* hw.network.packets.forwarded.unicast
* hw.network.packets.forwarded.multicast

eero-t: Looks good to me!   (IMHO much more readable than the ones in initial PR)

Note that you could have the drop reason (congestion, no_route, ttl) also as e.g. "reason" metric label:
`hw.network.packets.dropped{reason="ttl"}`

And forwarded frame type (broadcast, multicast, unicast) e.g. as "type" label:
`hw.network.packets.forwarded{type="multicast"}`

That way it's easier to get total count and do drill down to specifics in the receiving end / GUIs like Grafana.

src/collectd.conf.pod
```diff
@@ -5121,6 +5121,36 @@ setting is given, the B<Address> and B<Port> settings are ignored.
 
 =back
 
+=head2 Plugin C<mesh>
+
+The I<Mesh plugin> collects some statistics about the Wifi Mesh performance.
+The Linux kernel requires to be compile with CONFIG_MAC80211_DEBUGFS=y.
```
eero-t, line 5127: You could mention that as data is in debugfs, `collectd` would need to be run as `root` to access.  Maybe also that doing that with network facing service is a bit risky...

src/mesh.c
```diff
@@ -0,0 +1,426 @@
...
+#if !KERNEL_LINUX
+#error "This module only supports the Linux implementation of ieee8021"
+#endif
+
+#define PLUGIN_NAME "mesh"
+
+#define SYS_PATH "/sys/kernel/debug/ieee80211/"
+
+/*
+ * Device attributes
```
eero-t, line 36:
```suggestion
 * File names for device attributes under SYSPATH in debugfs
```

src/mesh.c
```diff
@@ -0,0 +1,426 @@
...
+  MESH_FAM_FORWARDED,
+  MESH_FAM_PEERS,
+  MESH_FAM_MAX,
+} mesh_family_t;
+
+static metric_family_t g_families[MESH_FAM_MAX] = {
+    [MESH_FAM_DROPPED] =
+        {
+            .name = "hw.network.packets.dropped",
+            .help = "",
```
eero-t, line 85: Either have a help description, or drop the `.help` members.   Write plugins should handle NULL members just fine (Prometheus plugin does, if others do not, that's a bug).

README
```diff
@@ -261,6 +261,12 @@ Features
       Memory utilization: Memory occupied by running processes, page cache,
       buffer cache and free.
 
+    - mesh
+      Collects various metrics related to Wifi mesh performance (e.g. established
+      peer links, dropped frames, forwarded frames). The statistics are read
+      from the Linux sysfs and require the Linux kernel to be compiled with
```
eero-t, line 267: It's read from `debugfs`, not `sysfs` (while distros do mount `debugfs` under `sysfs`, it's a separate fs):
```suggestion
      from the Linux debugfs and require the Linux kernel to be compiled with
```

eero-t (commented):
On quick check, looks good.   Just a couple of minor comments.

Btw. How much testing you've done for this?

pticon:
> On quick check, looks good. Just a couple of minor comments.
> 
> Btw. How much testing you've done for this?

Thanks for the review.

It has been fully tested  with Collectd v5 and I prefer to keep it as a draft for the moment.
Indeed, I'm using it with Yocto and it requires a bit more work to have Collectd v6 running on it.

Issue #4186, opened 2023-12-10
Any release plans? 5.12.1 or 5.13.0 or 6.0.0 ?


hnyman:
It is now over 3 years since the last release, 5.12.0.

Looking at the current branches, I am totally confused on how the development is going on and if there are any release plans?

From what I have followed collectd from "use for OpenWrt perspective" for over a decade, the development is currently somewhat stuck, and a "big change"(?) for 6.0 will likely take quite long ?

There are currently four branches that have seen commits this year:

* main branch. Last commit 6 December 2023.
* 6.0 branch: 341 commits ahead and 242 commits behind main. Last commit 6 December 2023.
* 5.13 branch: 2 commits ahead and 240 commits behind main. Last commit 24 March 2023.
* 5.12 branch: 6 commits ahead and 258 commits behind main. Last commit on 23 March 2023.

Four differing branches are a lot. 
Especially when only 5.12 has been released and both 5.13 and 6.0 are unreleased release branches.  

Differences between 5.12 and 5.13 are minimal, so the role of 5.13 branch is a puzzle.

Could there be a 5.12.1 maintenance release or a 5.13 release while waiting for 6.0?
Are both 5.12 and 5.13 really needed, if 5.13 is never going to be released?

octo:
Hi @hnyman,

that's an excellent question and I think we should have a community discussion about next steps and where we focus our energy.

* `collectd-5.12` branch

    Historically, we have committed bug fixes to "version branches" and merged them back into `main` to avoid cherry picking and duplicate commits. My suspicion is that recently bug fixes have simply been committed to `main` instead. My suggestion would be to package up a version 5.12.1 with the few fixes we have in there, since it's not much work. I wouldn't expect much more development to happen here.
* `collectd-5.13` branch

    Sounds like a release branch that got abandoned. I think we should rebase the few commits on there onto `main` (unless they are there already) and delete the `collectd-5.13` branch.
* `main` branch

   This is the current development branch for the `collectd-5.*` line. I think it contains numerous bug fixes and new features that we haven't released yet, and we should publish that as 5.13.0 at some point. I don't have a time line for that in mind at the moment. What do others think?
* `collectd-6.0` branch

[...]

PR #4319, opened 2024-07-25
processes: add delay-irq for latest taskstats

zzzyhtheonly:
Changelog: support IRQ delay accounting metric for processes plugin

Linux taskstats version has updated to VERSION 14 while collectd only support VERSION 8 at the moment.
This commit only add delay-irq in VERSION 14.
Other metrics after VERSION 8 should be added in later commits. If this PR make sense. 

Reference: 
https://docs.kernel.org/accounting/delay-accounting.html
https://github.com/torvalds/linux/blob/c33ffdb70cc6df4105160f991288e7d2567d7ffa/include/uapi/linux/taskstats.h#L202

Test on my local machine:
`write_log values:
[{"values":[0.000701929083335049],"dstypes":["gauge"],"dsnames":["value"],"time":1721907943.155,"interval":1.000,"host":"xavier2","plugin":"processes","plugin_instance":"collectd","type":"delay_rate","type_instance":"delay-irq"}]`


eero-t (changes requested):
I added few comments, but I think new features should go to v6 (`collectd-6.0`) branch instead of legacy v5 (`main`) branch.

Especially in `processes` plugin case, as @octo rewrote v6 version 1/2 year ago, to be more maintainable, and to support all the new `collectd` features, see: https://github.com/collectd/collectd/commits/collectd-6.0/src/processes.c

zzzyhtheonly:
Thanks for review. 

> I added few comments, but I think new features should go to v6 (`collectd-6.0`) branch instead of legacy v5 (`main`) branch.

Is there any plan for stable releases of v6? Now we only have rc versions which may not give users big motivation switching to v6 in production environment.

// Personally I'm using v5 releases in my production, so the fastest way for me to push new features into upstream is going on `main` branch :)

> 
> Especially in `processes` plugin case, as @octo rewrote v6 version 1/2 year ago, to be more maintainable, and to support all the new `collectd` features, see: https://github.com/collectd/collectd/commits/collectd-6.0/src/processes.c

I'll take a look and try to move on this.

eero-t:
> Is there any plan for stable releases of v6? Now we only have rc versions which may not give users big motivation switching to v6 in production environment.
> 
> // Personally I'm using v5 releases in my production, so the fastest way for me to push new features into upstream is going on `main` branch :)

See: https://github.com/collectd/collectd/issues/4186#issuecomment-1914389937

PR #4327, opened 2024-08-23
plugin.h:Fixed some punctuation errors in comments

LeavesMountains:
plugin.h:Fixed some punctuation errors in comments


mrunge:
Thank you for your PR. The punctuation in this file is consistent with all the other .h files at the moment.

eero-t:
Is there some problem with the quotes when using API document extraction tool like Doxygen?

Issue #4323, opened 2024-08-17
MariaDB Support?


robertsilen:
Hello, I noticed that the [documentation](https://www.collectd.org/documentation/) and the github [README](https://github.com/collectd/collectd/blob/main/README.md) only mention MySQL. Considering most distros run MariaDB by default, is there any particular reason MariaDB couldn't be mentioned? Only seems fair, and would remove any compatibility doubts. I could make a PR suggestion with some changes to include MariaDB?

eero-t:
Yes, PR would be appreciated!

(Along with checking that `mysql` plugin works with MariaDB as expected, if you are not already doing that.)

robertsilen:
Hello, MariaDB has been default in distros, e.g. Debian since 2017 and Fedora since 2015, that effectively any compatibility issues with collectd would already have turned up already (and been fixed). I suggested a PR https://github.com/collectd/collectd/pull/4329 to add MariaDB to the README, and I added a mention of MariaDB to the Wiki about [Plugin MySQL](https://github.com/collectd/collectd/wiki/Plugin-MySQL). Fairly simple additions, it could be explained more, but I think this already works so people understand the compatibility. I did no find the location of the manpages in the source, where is that?

eero-t:
> I did no find the location of the manpages in the source, where is that?

They're built from the `*.pod` files:
https://github.com/collectd/collectd/blob/main/Makefile.am#L2446

eero-t:
Merged the PR.  Sorry for the late response!

PR #4329, opened 2024-08-27
add MariaDB to README

robertsilen:
As per discussion in https://github.com/collectd/collectd/issues/4323 . 

I doubt there are any problems with MariaDB. It is the default running in most distros - in Debian since 2017 and on Fedora since 2015 (if I remember correctly). If you are not getting bug reports from Debian/Fedora users, then collectd probably does not have any particular issues with MariaDB. 

ChangeLog: Update the README to include MariaDB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,7 +296,7 @@ Features
       Information provided by serial multimeters, such as the `Metex M-4650CR`.
 
     - mysql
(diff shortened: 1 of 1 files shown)


mrunge:
Thank you for your PR! This looks good to me.

eero-t (approved):

PR #4331, opened 2024-10-23
Fix lcc_receive

WinstonMDP:
ChangeLog: collectd: The unit has been substracted to pass gcc stringop-truncation warning in lcc_receive function

The warning has been observed with gcc v14.2.1

### DIFF
--- a/src/libcollectdclient/client.c
+++ b/src/libcollectdclient/client.c
@@ -309,7 +309,7 @@ static int lcc_receive(lcc_connection_t *c, /* {{{ */
     ptr++;
 
   /* Now copy the message. */
-  strncpy(res.message, ptr, sizeof(res.message));
+  strncpy(res.message, ptr, sizeof(res.message) - 1);
   res.message[sizeof(res.message) - 1] = '\0';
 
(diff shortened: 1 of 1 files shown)


src/libcollectdclient/client.c
```diff
@@ -309,7 +309,7 @@ static int lcc_receive(lcc_connection_t *c, /* {{{ */
     ptr++;
 
   /* Now copy the message. */
-  strncpy(res.message, ptr, sizeof(res.message));
+  strncpy(res.message, ptr, sizeof(res.message) - 1);
   res.message[sizeof(res.message) - 1] = '\0';
```
matwey, line 313: I think that it is reasonable even though this seems to be false-positive alarm.
`res.message[sizeof(res.message) - 1]` is overwritten by zero, so there is no sense to copy anything to this place just one line before.

matwey (approved):

matwey:
@mrunge can we merge this too? I can confirm that build is really broken on gcc 14 due to `-Werror`.

mrunge (approved):
Makes sense. Yes, let's get this done.

PR #4332, opened 2024-10-30
configure.ac: fix protoc version check

jameshilliard:
The current version check for `3.0.0+` fails to handle cases where protoc is above version `3.0.0` but does not start with a `3`, for example protoc may have a version of `28.1`.

To fix this use `AX_COMPARE_VERSION` to properly compare the version.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4897,13 +4897,16 @@ AC_PATH_PROG([PROTOC], [protoc])
 have_protoc3="no"
 if test "x$PROTOC" != "x"; then
   AC_MSG_CHECKING([for protoc 3.0.0+])
-  if $PROTOC --version | $EGREP libprotoc.3 >/dev/null; then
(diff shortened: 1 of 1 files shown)


eero-t:
> The current version check for `3.0.0+` fails to handle cases where protoc is above version `3.0.0` but does not start with a `3`, for example protoc may have a version of `28.1`.
>
> To fix this use AX_COMPARE_VERSION to properly compare the version.

Will protoc v3 assumptions work if actual `protoc` version is much higher?  (I've never used protoc myself)

@mrunge any comments?

jameshilliard:
> Will protoc v3 assumptions work if actual `protoc` version is much higher? (I've never used protoc myself)

I think so, the versioning scheme changed to use much higher numbers or something but it seems compatible with the normal assumptions regarding the version being greater than 3.

PR #4333, opened 2024-11-15
dpdk: do not ignore return value from rte_eth_link_get_nowait

bluca:
In version 24.11 it sets __rte_warn_unused_result so compilation fails unless the return value is checked


bluca:
Debian 10 is EOL, and uses DPDK 18.11 which is also EOL, please consider removing it from the CI

eero-t:
> Debian 10 is EOL, and uses DPDK 18.11 which is also EOL, please consider removing it from the CI

@octo do you have time to look into this?

PR #4338, opened 2025-01-11
Collectd 6.0

danilabokhanov:
Prometheus plugin implementation with flex and bison code generation


danilabokhanov:
PR is ready for review

src/collectd.conf.pod
```diff
@@ -8425,6 +8425,62 @@ regular expression will be monitored for EXECs and EXITs.
...
 =back
 
+=head2 Plugin C<prometheus>
+
+This plugin collects metrics generated by a prometheus webserver.
+To see full description of prometheus format, please visit
+L<https://prometheus.io/docs/instrumenting/exposition_formats/>
+
+Collectd inner state doesn't support histogram and summary metric types. Thus,
+these types stored as gauge metrics.
```
eero-t, line 8435:
```suggestion
This plugin collects metrics generated by a prometheus webserver.
To see full description of Prometheus format, please visit
L<https://prometheus.io/docs/instrumenting/exposition_formats/>

Collectd supports neither histogram nor summary metric types so those
are converted to gauge metrics.
```

src/collectd.conf.pod
```diff
@@ -8425,6 +8425,62 @@ regular expression will be monitored for EXECs and EXITs.
...
+
+This plugin has the following config options: 
+
+=over 4
+
+=item B<URL> I<Prometheus_instance_URL>
+
+Sets the URL of the prometheus instance output.
```
eero-t, line 8443:
```suggestion
Sets the URL of the Prometheus instance output.
```

src/prometheus/ast.h
```diff
@@ -0,0 +1,126 @@
...
+} pr_metric_type_t;
+
+typedef enum {
+  PR_METRIC_ENTRY,
+  PR_COMMENT_ENTRY,
+  PR_TYPE_ENTRY,
+  PR_HELP_ENTRY
+} pr_entry_type_t;
+
+typedef enum { PR_METRIC_FAMILY_ITEM, PR_COMMENT_ITEM } pr_item_type_t;
```
eero-t, line 22: Nicer to have all typedefs layed out similarly:
```suggestion
typedef enum {
  PR_METRIC_FAMILY_ITEM,
  PR_COMMENT_ITEM
} pr_item_type_t;
```

Or was this something the format checker insisted on?

src/prometheus/prometheus.c
```diff
@@ -0,0 +1,358 @@
...
+                                    "Timeout",    "Socket",     "JWTToken"};
+static int config_keys_num = STATIC_ARRAY_SIZE(config_keys);
+
+static size_t prometheus_write_callback(void *ptr, size_t size, size_t nmemb,
+                                        void *stream) {
+  size_t len = size * nmemb;
+
+  if ((prometheus_buffer_len + len) >= sizeof(prometheus_buffer)) {
+    assert(sizeof(prometheus_buffer) > prometheus_buffer_len);
+    len = (sizeof(prometheus_buffer) - 1) - prometheus_buffer_len;
```
eero-t, line 61: Shouldn't `len` be aligned to member `size`?

src/prometheus/prometheus.c
```diff
@@ -0,0 +1,358 @@
...
+  }
+#endif
+
+#ifdef HAVE_CURLOPT_UNIX_SOCKET_PATH
+  if (sock != NULL) {
+    curl_easy_setopt(curl, CURLOPT_UNIX_SOCKET_PATH, sock);
+  }
+#endif
+
+  static char jwt_header[512];
```
eero-t, line 174: Some comment about the reasons for different buffer sizes (`prometheus_buffer`, `credentials`, `jwt_header`), e.g. pointer to relevant doc, wold be nice.

src/prometheus/prometheus.c
```diff
@@ -0,0 +1,358 @@
...
+  }
+  curl_easy_setopt(curl, CURLOPT_HTTPHEADER,
+                   curl_slist_append(NULL, jwt_header));
+
+  return 0;
+}
+
+static metric_t *convert_pr_metric_to_metric(pr_metric_t *pr_metric,
+                                             metric_type_t metric_type) {
+  metric_t *metric = malloc(sizeof(*metric));
```
eero-t, line 189: Allocs here and elsewhere in code could use collectd `scalloc()`, `smalloc()` and `sstrdup()` helpers which log an error and exit on allocation failures.

(There's rarely any point in continuing after alloc fails, as functionality on such error conditions is rarely tested to work correctly, and rest of system is pretty hosed at that point.)

src/prometheus/prometheus.c
```diff
@@ -0,0 +1,358 @@
...
+  case (PR_SUMMARY): {
+    fam->type = METRIC_TYPE_UP_DOWN_FP;
+    break;
+  }
+  case (PR_HISTOGRAM): {
+    fam->type = METRIC_TYPE_UP_DOWN_FP;
+    break;
+  }
+  default: {
+    ERROR("Unknown prometheus metric family type");
```
eero-t, line 255: If there's an error, log message should give details on the problem, i.e. what the unknown received type was.

(And please properly capitalize Prometheus name.)

src/prometheus/prometheus.c
```diff
@@ -0,0 +1,358 @@
...
+static pthread_mutex_t pr_metrics_lock;
+
+static int prometheus_read(void) {
+  pthread_mutex_lock(&pr_metrics_lock);
+  if (get_metrics() != 0) {
+    ERROR("Failed to fetch metrics from Prometheus.");
+    pthread_mutex_unlock(&pr_metrics_lock);
+    return -1;
+  }
+  pthread_mutex_unlock(&pr_metrics_lock);
```
eero-t, line 347: I don't see this threading, why it needs locking?

src/prometheus/ast.h
```diff
@@ -0,0 +1,126 @@
...
+  pr_metric_t *metric_list;
+} pr_metric_family_t;
+
+typedef pr_comment_entry_t pr_comment_t;
+
+typedef struct pr_item_s {
+  pr_item_type_t tp;
+  union {
+    pr_metric_family_t *metric_family;
+    pr_comment_t *comment;
```
eero-t, line 90: Why this struct is special?  It could use `pr_comment_entry_t` like the other structs do:
```suggestion
typedef struct pr_item_s {
  pr_item_type_t tp;
  union {
    pr_metric_family_t *metric_family;
    pr_comment_entry_t *comment;
```

src/prometheus/ast.h
```diff
@@ -0,0 +1,126 @@
...
+  char *name;
+  pr_metric_type_t tp;
+} pr_type_entry_t;
+
+typedef struct pr_help_entry_s {
+  char *name;
+  char *hint;
+} pr_help_entry_t;
+
+typedef struct pr_entry_s {
```
eero-t, line 56: Drop the redundant/unused struct names.

eero-t (commented):
It's so long time since I last touched C-based lexers, that somebody else can review that part. Here are few comments on the plain C-code & doc update.

src/prometheus/prometheus.c
danilabokhanov: This is the case when the curl data does not fit into the Prometheus buffer and should be truncated. This could potentially happen. I have deleted all fixed size buffers.

src/prometheus/ast.h
danilabokhanov: Unfortunately, clang-format repository check insisted on one-line enum

src/prometheus/prometheus.c
danilabokhanov: Jwt token as well as credentials doesn't have fixed size.
https://stackoverflow.com/questions/26033983/what-is-the-maximum-size-of-jwt-token
https://stackoverflow.com/questions/4027825/is-there-a-maximum-length-for-a-http-basic-authentication-username#:~:text=According%20to%20http%3A%2F%2Fhttpd,longer%20it%20will%20be%20truncated.
I've replace them with dynamically allocated buffers.

src/prometheus/prometheus.c
danilabokhanov: Sorry, this is a part of draft code. I've deleted it

matwey:
Can you please rebase the PR on top of `collectd-6.0`?

danilabokhanov:
> Can you please rebase the PR on top of `collectd-6.0`?

Done

src/prometheus/prometheus.c
```diff
@@ -0,0 +1,352 @@
...
+    break;
+  }
+  case (PR_SUMMARY): {
+    fam->type = METRIC_TYPE_UP_DOWN_FP;
+    break;
+  }
+  case (PR_HISTOGRAM): {
+    fam->type = METRIC_TYPE_UP_DOWN_FP;
+    break;
+  }
```
eero-t, line 258: Prometheus histogram types have total sum, total item count and per-bucket item counts count.

On quick browse I don't see which of them is converted. I assume it's sum, as nothing else would make sense, but why it's converted to Gauge instead of (FP) Counter?

danilabokhanov: All lines of the histogram (as well as summary data) are converted into a single metric family . Since the compiled metric system does not support additional suffixes, they are saved in the *m_suff* label (this part of the plugin [here](https://github.com/danilabokhanov/collectd/blob/collectd-6.0/src/prometheus/ast.c#L361)). Example of a processed histogram:
***
\# HELP duration_ms Histogram of the request duration distribution .
\# TYPE duration_ms **gauge**
duration_ms {le="50",**m_suff** ="bucket", ... } 1000 1751751751751
duration_ms {le="100 ", **m_suff**="bucket", ...} 2200 1751751751751
duration_ms {le="200 ", **m_suff** ="bucket", ...} 5000 1751751751751
duration_ms {le="500 ", **m_suff** ="bucket", ...} 17000 1751751751751
duration_ms {le="+Inf", **m_suff** ="bucket", ...} 19000 1751751751751
duration_ms {**m_suff** ="sum", ...} 65550 1751751751751
duration_ms {**m_suff**="count", ...} 19000 1751751751751
***
Gauge was chosen because I had noticed that there is something familiar to histogram that [minio](https://min.io/) server [produces](https://github.com/danilabokhanov/collectd_prometheus_parser/blob/main/tests/test_minio.in#L82)

I can change this part if necessary

src/prometheus/prometheus.c
```diff
@@ -120,15 +121,20 @@ static int prometheus_init(void) {
     curl_easy_setopt(curl, CURLOPT_USERNAME, user);
     curl_easy_setopt(curl, CURLOPT_PASSWORD, (pass == NULL) ? "" : pass);
 #else
-    static char credentials[1024];
-    int status = ssnprintf(credentials, sizeof(credentials), "%s:%s", user,
+    size_t credentials_size = strlen(user) + sizeof(":") + strlen(pass);
```
matwey, line 128: I think that `sizeof(":")` would return 8 since `":"` is a `const char *` and size of the pointer is 8. Is it what you what?

eero-t: Because it's a constant, at least GCC treats it same as:
`const char dcolon[] = ":";`
Which `sizeof()` is 2.

eero-t: When optimizations are enabled, the code generated for `2`, `sizeof(":")` and `strlen(":")+1"` are all identical.  Using just `strlen(":")` would IMHO be clearest though.

PR #4339, opened 2025-01-17
Systemd plugin

WinstonMDP:
The systemd plugin collects systemd metrics of specified services and slices.


configure.ac
```diff
@@ -7584,6 +7621,7 @@ AC_PLUGIN([swap],                [$plugin_swap],              [Swap usage statis
 AC_PLUGIN([synproxy],            [$plugin_synproxy],          [Synproxy stats plugin])
 AC_PLUGIN([sysevent],            [$plugin_sysevent],          [rsyslog events])
 AC_PLUGIN([syslog],              [$plugin_syslog],            [Syslog logging plugin])
+AC_PLUGIN([systemd],             [$plugin_systemd],           [Systemd plugin hey])
```
eero-t, line 7624: `hey`?

src/collectd.conf.pod
```diff
@@ -9555,6 +9555,40 @@ B<FAILURE> will only send failures to syslog.
...
+
+=over 4
+
+=item B<Service> I<service>
+
+Sets the service to collect metrics from it.
+
+=item B<Slice> I<slice>
+
+Sets the slice to collect metrics from it.
```
eero-t, line 9588: IMHO clearer as:
```suggestion
Collect metrics from specified service.

=item B<Slice> I<slice>

Collect metrics from specified slice.
```

src/systemd.c
```diff
@@ -0,0 +1,512 @@
...
+} systemd_metric_group;
+
+static systemd_metric_group service_groups[] = {
+    {
+        .accounting_flag = "MemoryAccounting",
+        .metrics =
+            (systemd_metric[]){
+                {
+                    .name = "MemoryAvailable",
+                    .dbus_type = "t",
```
eero-t, line 29: Add defines for these types and use them.

src/systemd.c
```diff
@@ -0,0 +1,512 @@
...
+                    .name = "NRestarts",
+                    .dbus_type = "u",
+                    .collectd_type = METRIC_TYPE_COUNTER,
+                },
+                {.collectd_type = METRIC_TYPE_UNTYPED},
+            },
+    },
+};
+
+static systemd_metric_group slice_groups[] = {
```
eero-t, line 152: Most of these duplicate the service groups.  It seems it would be better to split these into common ones, and service & slice specific ones?

src/systemd.c
```diff
@@ -0,0 +1,512 @@
...
+
+static sd_bus *bus = NULL;
+
+static bool was_service = false;
+static bool was_slice = false;
+
+static int introspect_prop(xmlXPathContextPtr xpath_ctx,
+                           char const interface[static 1], char **prop) {
+  char query[128];
+  sprintf(query,
```
eero-t, line 287: Use `snprintf()` to avoid compiler  etc. warnings.

src/systemd.c
```diff
@@ -0,0 +1,512 @@
...
+                             "org.freedesktop.DBus.Introspectable",
+                             "Introspect", &sd_bus_err, &m, "");
+  if (r < 0) {
+    ERROR("Can't introspect %s: %s {%s}, %s", unit->path, sd_bus_err.name,
+          sd_bus_err.message, strerror(-r));
+    return EXIT_FAILURE;
+  }
+  char const *xml;
+  r = sd_bus_message_read(m, "s", &xml);
+  xmlDocPtr doc = xmlReadMemory(xml, strlen(xml), "noname.xml", NULL, 0);
```
eero-t, line 318: XML is rather inefficient protocol.  Systemd has recently started to support JSON-based varlink interfaces, and those support also introspection: https://www.freedesktop.org/software/systemd/man/255/varlinkctl.html

Is it not possible to get same info through that?

src/systemd.c
```diff
@@ -0,0 +1,512 @@
...
+          metric_t m;
+          switch (metrics_it->collectd_type) {
+          case METRIC_TYPE_COUNTER:
+            m = (metric_t){.value.counter = val};
+            break;
+          case METRIC_TYPE_GAUGE:
+            m = (metric_t){.value.gauge = val};
+            break;
+          default:
+            ERROR("Unimplemented collectd type");
```
eero-t, line 477: Please log the failing type.

eero-t (commented):
This was not a proper review, just a quick look with few comments, but in general monitoring systemd group (cgroup) stats seems very useful functionality addition.

eero-t:
@mrunge Any comments on this functionality?

src/systemd.c
WinstonMDP: Done

src/systemd.c
WinstonMDP: Done

configure.ac
WinstonMDP: Done

matwey:
Can you please rebase the PR on top of `collectd-6.0` to fix the CI?

minfrin:
I have a need for a systemd plugin, was about to write one and found this pull request :)

What I am after specifically is to get notifications of when state changes occur (ActiveState) so that I can take actions based on this. For example, if a backup fails and enters the auto-restart state, I want to know about it, just in case the problem is caused by something I need to fix (eg full disk).

What I plan to do is use this PR as a base, and add the notifications to it.

matwey:
> I have a need for a systemd plugin, was about to write one and found this pull request :)
> 
> What I am after specifically is to get notifications of when state changes occur (ActiveState) so that I can take actions based on this. For example, if a backup fails and enters the auto-restart state, I want to know about it, just in case the problem is caused by something I need to fix (eg full disk).
> 
> What I plan to do is use this PR as a base, and add the notifications to it.

Cool. Since the PR is still on a review, it would be great if you could report us whether the code works for you.

PR #4347, opened 2025-02-20
chrony: Switch to libchrony

mlichvar:
This patch reworks the chrony plugin to use https://gitlab.com/chrony/libchrony.  It's a new library which makes it easy to implement monitoring clients like the collectd plugin.

The advantages are:
- less code to maintain
- compatible with different chronyd versions
- support for Unix domain sockets (specified as Host starting with '/')
- easy to add more data from other chronyd reports

The disadvantage is that the library is not available in any distros yet.

The patch is not finished and fully tested yet. It's just an RFC to see if it makes sense to continue the work. @bbczeuz, what do you think?


matwey:
> The disadvantage is that the library is not available in any distros yet.

To make it clear. Is `libcrhony` required for `chrony` package itself?

mlichvar:
No, chrony doesn't depend on libchrony. It's only for monitoring and doesn't support all the commands that chronyd and chronyc support.

matwey:
Ok. I wrongly guessed that `chrony` uses `libchrony`. I just wanted to say that relation to main `chrony` package would speed up `libchony` spreading over Linux distributions.

bbczeuz:
Hi mlichvar,


> The patch is not finished and fully tested yet. It's just an RFC to see if it makes sense to continue the work. @bbczeuz, what do you think?


happy to see that the code gets some attention and enhancements to more recent developments! I'm not using collectd in any setups now, so I cannot talk from a user's perspective. In general, I'm supporting the development; one thing that I'm not 100% sure about is the added dependency to a library that seemingly isn't very wide spread yet. And if the glue code to attach to the library is more maintenance than the current code. But the potential abstraction from the `chrony` version is good.
I haven't reviewed the code so.


> The disadvantage is that the library is not available in any distros yet.

Not *available* is a big con for me; *not installed by default* would be OK for me.


Happy coding!

mrunge:
Thank you for your work here! The CI requires the library to be present in the CI systems, i.e. in Debian or Fedora.

matwey:
Do I understand correctly that `libcrhony` has no releases currently?

mrunge:
@matwey that looks correct to me https://gitlab.com/chrony/libchrony/-/tree/main

mlichvar:
Thanks for the feedback.

I'm planning to make a libchrony release soon. I think I just need to add some documentation to the header. Then I'll submit a new package for Fedora.

My motivation is collecting data from other chronyd reports like serverstats. That is accessible only over the Unix domain socket and this part of the protocol is changing quite frequently. With libchrony it will be easy to support different versions.

bbczeuz:
Two things that came to me during the weekend:
1. I would wait with applying the PR until libchrony is added to all major distributions (IMHO at least Centos/Fedora/RHEL/Ubuntu/Debian/Arch/Mint/openSUSE); the PR would otherwise kill chrony monitoring. Optional: Add soft dependency and use libchrony if available, or directly connect to chronyd if not.
2. I don't really see a big benefit of the change; the current code does not depend on any external libs or binaries and seemingly ran stable for 10 years, so I wonder if cleaning up the current code (or add a chrony version check) would have the same benefit?

mlichvar:
It worked for so long because it used only the few UDP-accessible commands, which are intentionally kept stable. That doesn't apply to the commands accessible over the Unix domain socket. The serverstats command I mentioned is at its [fourth version](https://gitlab.com/chrony/libchrony/-/blob/main/reports.h?ref_type=heads#L334).

I can keep the original code wrapped in an `#ifdef` as a fallback when libchrony is not available, if you think that would help. I'd expect (at least some) distros to package libchrony when it shows up as a new dependency of collectd.

matwey:
> I'd expect (at least some) distros to package libchrony when it shows up as a new dependency of collectd.

I actually tend to agree with it, but `libchrony` still should have a released version and meaningful soversion to pass guidelines of some distros.

mlichvar:
libchrony-0.1 has been released and it's now available in EPEL9. How can I add it to the centos9 CI image?

mlichvar:
Here is a PR to add libchrony-devel to the centos9 image: https://github.com/collectd/ci-docker/pull/82

It could be also added to Fedora rawhide and Fedora 42 if it was present.

PR #4349, opened 2025-03-08
Added note about sparse data being late in large packets.

un-clouded:
Just a note for the `collectd.conf` man page (in the section for the Network plugin) that data can be late and cause FAILURE notifications if there isn't enough data to fill a packet in time.  It took me a while to figure this out!


matwey:
Is it valid for collect-6.0 branch?

PR #4353, opened 2025-04-05
fix smart: Fix stack smashing when use incorrect NVME_SMART_CDW10 value

xenols:
ChangeLog:  smart: Fix stack smashing when use incorrect NVME_SMART_CDW10 value

I'm faced that a collectd crash when trying to get a Smart from an NVMe disk
on an arm64 machine. The problem is due to the wrong amount of information being
requested in the ioctl call.

/etc/collectd/collectd.conf

```
<Plugin "smart">
    Disk "nvme0n1"
    IgnoreSelected false
</Plugin>
```

valgrind collectd -f

```
*** stack smashing detected ***: terminated
==861185==
==861185== Process terminating with default action of signal 6 (SIGABRT)
==861185==    at 0x4970A60: __pthread_kill_implementation (pthread_kill.c:44)
==861185==    by 0x492A72B: raise (raise.c:26)
==861185==    by 0x491747B: abort (abort.c:79)
==861185==    by 0x4964AAB: __libc_message (libc_fatal.c:156)
==861185==    by 0x49E7857: __fortify_fail (fortify_fail.c:26)
==861185==    by 0x49E7823: __stack_chk_fail (stack_chk_fail.c:24)
==861185==    by 0x6722117: smart_read_nvme_disk (smart.c:375)
==861185==    by 0x6722897: smart_handle_disk (smart.c:566)
==861185==    by 0x6722897: smart_read (smart.c:628)
==861185==    by 0x11936F: plugin_read_thread (plugin.c:533)
==861185==    by 0x496EE9F: start_thread (pthread_create.c:442)
==861185==    by 0x49D7B1B: thread_start (clone.S:79)
==861185==
```

Fix based according information:

NVM Express® Base Specification, Revision 2.1
Figure 197: Get Log Page – Command Dword 10

### DIFF
--- a/src/smart.c
+++ b/src/smart.c
@@ -320,9 +320,10 @@ static int smart_read_nvme_disk(const char *dev, char const *name) {
 
   /**
    * Prepare Get Log Page command
-   * Fill following fields (see NVMe 1.4 spec, section 5.14.1)
-   * - Number of DWORDS (bits 27:16) - the struct that will be passed for
-   *   filling has 512 bytes which gives 128 (0x80) DWORDS
+   * Fill following fields (see NVMe 2.1 spec, section 5.1.12)
+   * - Number of DWORDS (bits 31:16) - 0's based value.
+   *   This means that the value will be 0x7F for a 512 byte structure (0x80
+   * DWORD)
    * - Log Page Indentifier (bits 7:0) - for SMART the id is 0x02
    */
 
--- a/src/smart.c
+++ b/src/smart.c
@@ -45,7 +45,7 @@
 #endif
 
 #define O_RDWR 02
-#define NVME_SMART_CDW10 0x00800002
+#define NVME_SMART_CDW10(numdl) (((((numdl) >> 2) - 1) << 16) | 0x00000002)
 #define SHIFT_BYTE_LEFT 256
 #define PLUGIN_NAME "smart"
 struct nvme_admin_cmd {
@@ -331,7 +331,7 @@ static int smart_read_nvme_disk(const char *dev, char const *name) {
                                           .nsid = NVME_NSID_ALL,
                                           .addr = (unsigned long)&smart_log,
                                           .data_len = sizeof(smart_log),
-                                          .cdw10 = NVME_SMART_CDW10});
+                                          .cdw10 = NVME_SMART_CDW10(sizeof(smart_log))});
   if (status < 0) {
     ERROR(PLUGIN_NAME ": ioctl for NVME_IOCTL_ADMIN_CMD failed with %s\n",
           strerror(errno));
--- a/src/smart.c
+++ b/src/smart.c
@@ -326,12 +326,13 @@ static int smart_read_nvme_disk(const char *dev, char const *name) {
    * - Log Page Indentifier (bits 7:0) - for SMART the id is 0x02
    */
 
-  status = ioctl(fd, NVME_IOCTL_ADMIN_CMD,
-                 &(struct nvme_admin_cmd){.opcode = NVME_ADMIN_GET_LOG_PAGE,
-                                          .nsid = NVME_NSID_ALL,
-                                          .addr = (unsigned long)&smart_log,
-                                          .data_len = sizeof(smart_log),
-                                          .cdw10 = NVME_SMART_CDW10(sizeof(smart_log))});
+  status = ioctl(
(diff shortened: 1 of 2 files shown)


src/smart.c
```diff
@@ -45,7 +45,7 @@
 #endif
 
 #define O_RDWR 02
-#define NVME_SMART_CDW10 0x00800002
+#define NVME_SMART_CDW10(numdl) (((((numdl) >> 2) - 1) << 16) | 0x00000002)
```
matwey, line 48: Where did it come from? Could you please left a reference?

matwey:
Could you also verify if collectd-6.0 branch also needs this patch?

src/smart.c
xenols: https://nvmexpress.org/wp-content/uploads/NVM-Express-Base-Specification-Revision-2.1-2024.08.05-Ratified.pdf

```
Figure 197: Get Log Page – Command Dword 10

Bits 31:16 
Number of Dwords Lower (NUMDL): This field specifies the least significant 16 bits of the number of
dwords to return unless otherwise specified. If host software specifies a size larger than the log page
requested, the controller returns the complete log page with undefined results for dwords beyond the end
of the log page. The combined NUMDL and NUMDU fields form a 0’s based value.

Bits 07:00
Log Page Identifier (LID): This field specifies the identifier of the log page to retrieve (refer to Figure
202).

5.1.12.1.3 SMART / Health Information (Log Page Identifier 02h)
```

xenols:
> Could you also verify if collectd-6.0 branch also needs this patch?

xenols:
> Could you also verify if collectd-6.0 branch also needs this patch?

yes, as far as i can see, collectd-6.0 branch also needs this patch.

src/smart.c
matwey: Sorry, I still doesn't get it. size of `nvme_smart_log` is 512 bytes or 128 double words, while your equation evaluates to 127.

xenols: "The combined NUMDL and NUMDU fields form a 0’s based value."
0 - means a request for one dword.

xenols: Yes, I'm also looking at this code now)

src/smart.c
```diff
@@ -331,7 +331,7 @@ static int smart_read_nvme_disk(const char *dev, char const *name) {
                                           .nsid = NVME_NSID_ALL,
                                           .addr = (unsigned long)&smart_log,
                                           .data_len = sizeof(smart_log),
```
matwey, line 333: Why does `NVME_IOCTL_ADMIN_CMD` not respect `.data_len`?

matwey:
Could you please fix failing checks?

src/smart.c
xenols: I don't know, maybe some dma issue in driver. Anyway, after specifying the correct size in cdw10  the problem went away.

matwey: To make it clear, I do agree that `NVME_SMART_CDW10` should be corrected to `smartmontools` equation, i.e `0x80` -> `0x7f`. But you also need to report this issue to glibc/kernel, since `ioctl` is not expected to override more than `.data_len` bytes.

matwey:
Also, I think that this comment should be rephrased as well: https://github.com/collectd/collectd/blob/41f6ad4c6f8d0723dcc1da0558494051525035c7/src/smart.c#L321

xenols:
> Also, I think that this comment should be rephrased as well:
> 
> https://github.com/collectd/collectd/blob/41f6ad4c6f8d0723dcc1da0558494051525035c7/src/smart.c#L321

Yes, thank you

matwey (approved):
@octo @mrunge I believe this commit makes sense and is ready to merge

eero-t (approved):
Looks OK (thanks @matway!), approved.  Could you rebase to `main` to get tests re-run?

user318:
Hi, I have the same problem with Intel NVME, but the patch does not fully help, because Intel has additional ioctl request. I fixed it in the same manner and it works for me. I also changed the proposed patch to embed the struct size in the constant define itself. It seems not right to pass struct size every time to it. At least it is not used with different structs.
[collectd-nvme-ioctl.patch](https://github.com/user-attachments/files/24529533/collectd-nvme-ioctl.patch)

PR #4355, opened 2025-04-23
configure.ac: use pkg-config to find libgcrypt

### LINKED ISSUE #4310: Please use pkg-config instead of libgcrypt-config to locate libgcrypt
Forwarding Debian Bug#1070905:

Please switch from using libgcrypt-config to pkg-config to locate
libgcrypt. libgcrypt-config will be dropped in the next libgcrypt major
release.

(Please do not switch from collectd's handwritten detection code to
AM_PATH_LIBGCRYPT() unless you invoke AM_PATH_GPG_ERROR() first. The
former rcurrently requires the latter to work without libgcrypt-config.
Just go for pkg-config instead.)

atsampson:
libgcrypt has provided a `.pc` file since 2018, and as of 2024 it no longer provides a `libgcrypt-config` program by default, so the existing test won't find the library any more.

Fixes #4310.

This changes the help message for `--with-libgcrypt` to reflect that it can't take a path as an argument any more - if you want to override pkg-config, you can instead do this by setting environment variables as normal with `PKG_CHECK_MODULES`. It might be worth making the same change to the other `--with` options for pkg-config libs, since they also ignore their argument now...

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -581,8 +581,7 @@ libcollectdclient_la_LDFLAGS += -shared -no-undefined
 libcollectdclient_la_LIBADD += -lgnu -lws2_32 -liphlpapi
 endif
 if BUILD_WITH_LIBGCRYPT
-libcollectdclient_la_CPPFLAGS += $(GCRYPT_CPPFLAGS)
-libcollectdclient_la_LDFLAGS += $(GCRYPT_LDFLAGS)
+libcollectdclient_la_CPPFLAGS += $(GCRYPT_CFLAGS)
 libcollectdclient_la_LIBADD += $(GCRYPT_LIBS)
 endif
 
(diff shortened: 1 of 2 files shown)


eero-t:
Thanks for the PR!  Please add the required one-liner `ChangeLog:` line to the PR description.

Changes look otherwise OK, but I'm not sure whether non-`pkg-config` checks would still be needed for CentOS/RHEL, maybe also on non-Linux distros.

=> @mrunge, any comments on how old libgcrypt versions oldest (still maintained) enterprise distro versions have?

atsampson:
ChangeLog entry added, and I've gone back through the tarballs and added the version numbers above. 

All maintained versions of Debian have the .pc file. For Red Hat, it looks like RHEL 9 was the first to include the .pc file (RHEL 8 has 1.8.5 but they deliberately remove the .pc file).

mrunge:
> Changes look otherwise OK, but I'm not sure whether non-`pkg-config` checks would still be needed for CentOS/RHEL, maybe also on non-Linux distros.
> 
> => @mrunge, any comments on how old libgcrypt versions oldest (still maintained) enterprise distro versions have?

Let's target more recent versions with this.

mrunge (approved):
Thank you!

eero-t:
Debian unstable test fails due to missing `cadef.h` include for `epics` plugin, FreeBSD test due to obsolete OS version, which need separate fixes => merged this.

@atsampson Could you provide similar PR also for collectd-6.0 branch?

PR #4379, opened 2025-10-08
feat: mqtt notifications in plain text or json

markferry:
ChangeLog: mqtt plugin: publish notifications. Publish as PLAIN or JSON.

This adds:
 - mqtt notification publishing
 - mqtt config `SendNotifications` (default: `false`)
 - mqtt publishing as either plain text or JSON.
 - mqtt config `Format` (default `PLAIN`)

Tested notifications with `threshold` and `connectivity` plugins.
The latter produces *embedded* https://www.onap.org/ formatted json which is correctly escaped.

# Plain vs JSON metrics

Topic `collectd/canard.local/interface-dummy0/if_dropped`:

## Plain
```
1760120317.152:0:0
```

## JSON

> :warning: note the redundant host, plugin, plugin_instance, type and type_instance
>

```json
[
  {
    "values": [
      0,
      0
    ],
    "dstypes": [
      "derive",
      "derive"
    ],
    "dsnames": [
      "rx",
      "tx"
    ],
    "time": 1760120317.152,
    "interval": 10.000,
    "host": "canard.local",
    "plugin": "interface",
    "plugin_instance": "dummy0",
    "type": "if_dropped",
    "type_instance": ""
  }
]
```

# Plain vs JSON notifications

Topic: `collectd/event/canard.local/interface-dummy0/if_octets`

```
Time: 1760218183.205
Severity: OKAY
DataSource: tx
CurrentValue: nan
WarningMin: nan
WarningMax: nan
FailureMin: nan
FailureMax: 100

Message: Host canard.local, plugin interface (instance dummy0) type if_octets: All data sources are within range again. Current value of "tx" is nan.

```

## JSON
```json
[
    {
        "labels": {
            "alertname": "collectd_interface_if_octets",
            "instance": "canard.local",
            "interface": "dummy0",
            "service": "collectd"
        },
        "annotations": {
            "severity": "OKAY",
            "summary": "Host canard.local, plugin interface (instance dummy0) type if_octets: All data sources are within range again. Current value of \"tx\" is nan.",
            "DataSource": "tx",
            "CurrentValue": "nan",
            "WarningMin": "nan",
            "WarningMax": "nan",
            "FailureMin": "nan",
            "FailureMax": "100"
        },
        "startsAt": "2025-10-11T21:29:43.204853637Z",
        "endsAt": "2025-10-11T21:29:43.204853637Z"
    }
]
```

### DIFF
--- a/src/mqtt.c
+++ b/src/mqtt.c
@@ -523,8 +523,6 @@ static int mqtt_write(const data_set_t *ds, const value_list_t *vl,
   mqtt_client_conf_t *conf;
   char topic[MQTT_MAX_TOPIC_SIZE];
   char payload[MQTT_MAX_MESSAGE_SIZE];
-  size_t offset = 0;
-  size_t bfree = sizeof(payload);
   int status = 0;
 
   if ((user_data == NULL) || (user_data->data == NULL))
@@ -538,6 +536,9 @@ static int mqtt_write(const data_set_t *ds, const value_list_t *vl,
   }
 
   if (conf->format == MQTT_FORMAT_JSON) {
+    size_t offset = 0;
+    size_t bfree = sizeof(payload);
+
     format_json_initialize(payload, &offset, &bfree);
     format_json_value_list(payload, &offset, &bfree, ds, vl, conf->store_rates);
     status = format_json_finalize(payload, &offset, &bfree);
@@ -593,18 +594,18 @@ static int format_notification_topic(char *buf, size_t buf_len,
   return 0;
 } /* format_notification_topic */
 
-static int format_plain_notification(char *buffer, size_t buffer_size,
+static int format_plain_notification(char *buf, size_t buf_len,
                                      notification_t const *n) {
   int status = 0;
 
 #define APPEND_VA(format, ...)                                                 \
-  if (buffer_size > 0) {                                                       \
-    status = ssnprintf(buffer, buffer_size, format "\r\n", __VA_ARGS__);       \
+  if (buf_len > 0) {                                                           \
+    status = ssnprintf(buf, buf_len, format "\r\n", __VA_ARGS__);              \
     if (status >= 0) {                                                         \
-      buffer += status;                                                        \
-      buffer_size -= status;                                                   \
+      buf += status;                                                           \
+      buf_len -= status;                                                       \
     } else {                                                                   \
-      ERROR("mqtt plugin: notification buffer too large.");                    \
+      ERROR("mqtt plugin: notification buf too small.");                       \
       return ENOMEM;                                                           \
     }                                                                          \
   }


src/mqtt.c
```diff
@@ -514,6 +523,8 @@ static int mqtt_write(const data_set_t *ds, const value_list_t *vl,
   mqtt_client_conf_t *conf;
   char topic[MQTT_MAX_TOPIC_SIZE];
   char payload[MQTT_MAX_MESSAGE_SIZE];
+  size_t offset = 0;
+  size_t bfree = sizeof(payload);
```
eero-t: Move these under MQTT_FORMAT_JSON block, so they are declared where they're used.

src/mqtt.c
```diff
@@ -526,20 +537,188 @@ static int mqtt_write(const data_set_t *ds, const value_list_t *vl,
...
+    return ENOMEM;
+
+  while ((c = strchr(buf, '#')) || (c = strchr(buf, '+'))) {
+    *c = '_';
+  }
+
+  return 0;
+} /* format_notification_topic */
+
+static int format_plain_notification(char *buffer, size_t buffer_size,
```
eero-t: For consistency with other functions:
```suggestion
static int format_plain_notification(char *buf, size_t buf_len,
```

src/mqtt.c
```diff
@@ -526,20 +537,188 @@ static int mqtt_write(const data_set_t *ds, const value_list_t *vl,
...
+  int status = 0;
+
+#define APPEND_VA(format, ...)                                                 \
+  if (buffer_size > 0) {                                                       \
+    status = ssnprintf(buffer, buffer_size, format "\r\n", __VA_ARGS__);       \
+    if (status >= 0) {                                                         \
+      buffer += status;                                                        \
+      buffer_size -= status;                                                   \
+    } else {                                                                   \
+      ERROR("mqtt plugin: notification buffer too large.");                    \
```
eero-t:
```suggestion
      ERROR("mqtt plugin: notification buffer too small.");                    \
```

src/mqtt.c
markferry: Done. (I take it then that collectd does not target C89.)

src/mqtt.c
markferry: Done!

I note, for future refactorings...

```
$ ack --cc 'buffer_size'  src | wc -l
507
$ ack --cc 'buf_len'  src | wc -l
85
$ ack --cc 'buf_size'  src | wc -l
58
```

src/mqtt.c
markferry: Done

src/mqtt.c
eero-t: Thanks, I did not check rest of collectd.  While it would be good to be consistent within whole project, IMHO it's more important to be consistent with same plugin (I think most people deal only with a single one).

src/mqtt.c
eero-t: Collectd requires C99, but that's fine even with C89. It only requires variables to be in start of a scope, not in start of a function.

src/mqtt.c
eero-t: Err... Log message could still say "buffer" for readability, that's more important than using exact variable name...

eero-t:
@markferry maybe you could add PR for the same change in `collectd-6.0`?

(You can just copy `main` version over, that will provide also previous fix commit from 2022 that seems to be missing from v6 branch.)

> telegraf mqtt_consumer config to parse the important bits:

@mrunge Would it make sense to include such example with collectd?  That can be separate PR though.

src/mqtt.c
markferry: Doh!

src/mqtt.c
markferry: Right you are. It's been a while.

markferry:
> @markferry maybe you could add PR for the same change in `collectd-6.0`?
> 

*If* mqtt were building for collectd-6.0, I'd do that but...

./configure:
> plugin_mqtt="no (collectd 6 support missing)"

PR #4381, opened 2025-10-26
Fix lcc_receive

matwey:
Issue #4331 is still present in `main` branch. Here we backport fixing commit from 6.0.

ChangeLog: collectd: The unit has been substracted to pass gcc stringop-truncation warning in lcc_receive function


mrunge (approved):
Thank you

PR #4383, opened 2025-10-26
dpdk: check rte_eth_link_get_nowait return value

matwey:
Update link status only if rte_eth_link_get_nowait return success otherwise link may not be filled properly.

This also fixes the following build issue with up-to-date gcc version:

```
src/dpdkevents.c: In function 'dpdk_helper_link_status_get': src/dpdkevents.c:439:7: error: ignoring return value of 'rte_eth_link_get_nowait' declared with attribute 'warn_unused_result' [-Werror=unused-result]
  439 |       rte_eth_link_get_nowait(i, &link);
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ChangeLog: Fix dpdk build for newer gcc versions


mrunge (approved):
Thank you!

PR #4388, opened 2025-11-16
fix: for TypeInstanceOID, PluginOID or HostOID being outside of the direct scope of Values. Fixes bug where collectd discards metrics collected from SNMP tables augmenting another table.

### LINKED ISSUE #4387: SNMP: Collectd collects but does not dispatch metrics when snmp tables augments other tables
Collectd currently collects but does not dispatch snmp metrics in a table when using a TypeInstanceOID, PluginOID or HostOID residing anywhere but in direct same OID-scope as the collected values. More info is in the pull request.

cekstam:
Fixes: #4387 

csnmp_dispatch_table() only considers rows where every collected column shares the exact same suffix. In a case where the type-instance column produces .1.23 while Value-column produces .1.23.1, the comparison marks every row as “skipped” and nothing is dispatched, even though the agent returned valid data.
Instead of assuming metadata/value suffixes are identical, introduce a helper that treats two suffixes as aligned when either one is a prefix of the other. The driver loop now advances pointers only when the suffix is both “behind” and not aligned, and accepts rows when the shorter index matches the beginning of the longer one. This lets instance columns live elsewhere in the MIB tree while still labelling values correctly.

ChangeLog: SNMP plugin: Dispatch table values when using augmented tables.

### DIFF
--- a/src/snmp.c
+++ b/src/snmp.c
@@ -179,8 +179,7 @@ static bool csnmp_suffix_within(oid_t const *prefix, oid_t const *full) {
 }
 
 static bool csnmp_suffixes_align(oid_t const *left, oid_t const *right) {
-  return csnmp_suffix_within(left, right) ||
-         csnmp_suffix_within(right, left);
+  return csnmp_suffix_within(left, right) || csnmp_suffix_within(right, left);
 }
 
 static int csnmp_oid_to_string(char *buffer, size_t buffer_size,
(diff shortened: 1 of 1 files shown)


cekstam:
Any chance this would be considered for the 5.1-train aswell? To get adoption in all major distributions before they go collectd-6.

cekstam:
@octo sorry to bother you, but I cannot find `collectd/trusted-contributors` so I do not know whom to reach out to. Do you know anyone that could and have the time to review this?

mrunge:
Hi @cekstam thank you for your patch!

You should be able to find the list here https://github.com/orgs/collectd/teams/trusted-contributors

That being said, I'd hope to get to it at some point; however, my focus has moved and my time for collectd is very limited.

cekstam:
> https://github.com/orgs/collectd/teams/trusted-contributors

I get a 404 when trying to access that url, may something permissions-related?

PR #4400, opened 2026-03-25
feat(ipnat): Add an ipnat plugin

efuss:
ChangeLog: An `ipnat` plugin for monitoring NAT statistics (a la `ipnat -s`) has been added.

This PR adds an `ipnat` plugin for monitoring ipnat (part of Darren Reed's IPFilter) statistics akin to `ipnat -s`.

The branch is based on `ipfilter`, not `main`, because it was developed that way and changes parts of `configure.ac` and `Makefile.am` very close to the ipfilter changes which thus are part of the patch context.

It has only been tested on NetBSD.


efuss:
Unfortunately, compilation fails on Solaris and I don't have access to a Solaris machine. Can someone please provide `<netinet/ipnat.h>` from the build machine plus, if available, the source to `ipnat`, e.g. `ipnat.c`, wherever that lives in the source tree? It may also help to know which version of `IPFilter` that Solaris uses, e.g., the `IPL_VERSION`/`IPFILTER_VERSION` values from `<netinet/ipl.h>`.

Otoh, from the logs, it looks like my `ipfilter` plugin _does_ build on Solaris. Someone able to check whether it actually _works_?

efuss:
OK, someone sent me the relevant header files from Solaris, now it builds on that platform. Would still be nice if someone could check it (plus the `ipfilter` plugin) actually _works_ on Solaris.

PR #4401, opened 2026-03-26
Fix new qpos library call missing an arguement

watologo1:
ChangeLog: Fixes compile error in rdt_init.c

src/intel_rdt.c: In function ‘rdt_init_cores_monitoring’: src/intel_rdt.c:1406:48: error: passing argument 5 of \ ‘pqos_mon_start_cores’ from incompatible pointer type \ [-Wincompatible-pointer-types]

It looks as if in qpos.h they introduce C&P typo
in the code description of pqos_mon_start and
pqos_mon_start_cores.
Cmp with latest /usr/include/pqos.h lines 1044-1105.
The new pqos_mon_start_cores has a new arguement at position 5
and now takes 6 arguments, while the description
(and intel_rdt.c in collectd) wrongly has the same arguments.

int pqos_mon_start_cores(const unsigned num_cores,
                         const unsigned *cores,
                         const enum pqos_mon_event event,
                         void *context,
----->
                         struct pqos_mon_mem_region *mem_region,
----->
                         struct pqos_mon_data **group);

### DIFF
--- a/src/intel_rdt.c
+++ b/src/intel_rdt.c
@@ -1400,7 +1400,11 @@ static void rdt_init_cores_monitoring() {
   for (size_t i = 0; i < g_rdt->cores.num_cgroups; i++) {
     core_group_t *cg = g_rdt->cores.cgroups + i;
 
-#if PQOS_VERSION >= 40600
+#if PQOS_VERSION >= 60001
+    int mon_start_result =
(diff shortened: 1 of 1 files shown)


watologo1:
The patch should be correct, I expect that debian experimental has an intel-cmt-cat (qpos) version in between 6001 and the API breaking change.

PR #4403, opened 2026-04-22
fix: virt: Fix GCC 16's `-Wunused-but-set-variable`

sergiodj:
GCC 16 started treating these are errors, which leads to a build failure.

ChangeLog: virt: Fix GCC 16's `-Wunused-but-set-variable` warning

### DIFF
--- a/src/virt.c
+++ b/src/virt.c
@@ -2343,11 +2343,15 @@ static int persistent_domains_state_notification(void) {
   } else {
     DEBUG(PLUGIN_NAME " plugin: getting state of %i persistent domains", n);
     /* Fetch each persistent domain's state and notify it */
+#if COLLECT_DEBUG
     int n_notified = n;
+#endif
     for (int i = 0; i < n; ++i) {
       status = get_domain_state_notify(domains[i]);
       if (status != 0) {
+#if COLLECT_DEBUG
         n_notified--;
+#endif
         ERROR(PLUGIN_NAME " plugin: could not notify state of domain %s",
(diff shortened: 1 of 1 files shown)


kwiatrox (approved):
Looks good, thank you.

src/virt.c
```diff
@@ -2343,11 +2343,15 @@ static int persistent_domains_state_notification(void) {
   } else {
     DEBUG(PLUGIN_NAME " plugin: getting state of %i persistent domains", n);
     /* Fetch each persistent domain's state and notify it */
+#if COLLECT_DEBUG
     int n_notified = n;
+#endif
```
eero-t, line 2348: IMHO cleaner could be to have another macro in `plugin.h`:
```
 #if COLLECT_DEBUG
+#define DEBUG_VAR(x) x
 #define DEBUG(...) plugin_log(LOG_DEBUG, __VA_ARGS__)
 #else              /* COLLECT_DEBUG */
+#define DEBUG_VAR(x)
 #define DEBUG(...) /* noop */
 #endif             /* ! COLLECT_DEBUG */
```

And then use that variable macro on debug logging variables:
```suggestion
DEBUG_VAR(int n_notified = n);
```

Thoughts? / Do you see any problems with using something like that?

PS. C23 would allow use of `maybe_unused` annotation, but `collectd` cannot switch to that yet: https://en.wikipedia.org/wiki/C23_(C_standard_revision)#C++_compatibility

mrunge:
Thank you. @eero-t suggestions could be a follow-up

PR #4404, opened 2026-05-28
preserve const-ness when calling strrchr()

hector-cao:
ChangeLog: cleanup: strrchr() preserves the const-ness of the argument

, the argument is a const char* so the return value should also be a const.


mrunge (approved):
Thank you

PR #4411, opened 2026-09-01
fix(apache): count lighttpd response-end scoreboard entries

darkdi:
`submit_scoreboard` tests `buf[i] == 'S'` twice in the same chain. The first arm counts it as an Apache starting worker, the second as a lighttpd response end, and the second is unreachable.

That makes `response_end` a metric that is submitted for lighttpd and can never be anything but zero, while the `S` characters that should feed it land in `starting`, which is only submitted when the server type is Apache. So for lighttpd those events are counted into a value nobody reads and reported as zero in the one that people do.

The letter genuinely means two different things depending on the server, and `st->server_type` is already available here and used a few lines below, so I branched on it rather than picking one meaning. The other lighttpd letters do not collide: `q`, `Q`, `s`, `r`, `h` and `E` are all unused by the Apache arms.

I have not run this against a live lighttpd, the reasoning is from the parser and the two submit blocks.

ChangeLog: Apache plugin: A bug that made the lighttpd "response_end" scoreboard metric always report zero has been fixed.


darkdi:
Ran `clang-format` on the touched block and added the `ChangeLog:` line to the description. The Labels check needs `Fix` set, which I cannot do without triage rights on the repo.

darkdi:
The Labels check now passes after adding the documented `fix(apache):` title prefix. The remaining clang-format status says `Post "https://format.collectd.org/": context deadline exceeded`; it is a service error, not a reported formatting difference. Local clang-format accepts the changed block, and both Solaris builds pass. Could you rerun the formatter check when the service is available?

